PR #8729, opened 2021-08-18
Quick Action Menu: Add event tracking

### LINKED ISSUE #8689: Quick Action Menus: Add Tracking
## Feature Description

Add tracking to the actions in quick action menus so we can see how often users click them.

samwhale:
## Context

Right click menu got tracking but quick actions needed some love too.

## Summary

Adds event tracking to the quick action menu.

## Relevant Technical Choices

Refactored a bit so that we didn't pass in the selected element's data, but rather pulled it from the hook scope. This value should be constant since clicking these actions will not change the selected element in the canvas.

## To-do

n/a

## User-facing changes

None

## Testing Instructions

All quick actions should still work the same, but now we have analytics!

### QA

All quick actions in the quick action menu should still work the same but now will have analytics through google analytics.

no

no

no

## Checklist

- [x] This PR contains automated tests (unit, integration, and/or e2e) to verify the code works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))

---

Fixes #8689

### DIFF
--- a/packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
+++ b/packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
@@ -36,7 +36,7 @@ import {
   ELEMENT_TYPES,
 } from '../../story';
 import { getResetProperties } from './utils';
-import { ACTION_TEXT, RESET_PROPERTIES, RESET_DEFAULTS } from './constants';
+import { ACTIONS, RESET_PROPERTIES, RESET_DEFAULTS } from './constants';
 
 const {
   Bucket,
@@ -145,11 +145,6 @@ const useQuickActions = () => {
             /* commitValues */ true
           ),
       });
-
-      trackEvent('quick_action', {
-        name: 'reset_properties',
-        element: elementType,
-      });
     },
     [selectedElementAnimations, updateElementsById]
   );
@@ -172,7 +167,15 @@ const useQuickActions = () => {
         dismissable: false,
         message: __('Element properties have been reset', 'web-stories'),
         // Don't pass a stale version of `undo`
-        onAction: () => undoRef.current(),
+        onAction: () => {
+          undoRef.current();
+
+          trackEvent('quick_action', {
+            name: `undo_${ACTIONS.RESET_ELEMENT.trackingEventName}`,
+            element: elementType,
+            isBackground: true,
+          });
+        },
       });
     },
     [handleResetProperties, showSnackbar]
@@ -192,21 +195,13 @@ const useQuickActions = () => {
    * @param {Event} ev The triggering event.
    */
   const handleFocusPanel = useCallback(
-    (highlight) =>
-      ({ elementId, elementType } = {}) =>
-      (ev) => {
-        ev.preventDefault();
-        setHighlights({
-          elementId: elementId || selectedElement?.id,
-          highlight,
-        });
-
-        trackEvent('quick_action', {
-          name: 'focus_panel',
-          element: elementType || selectedElement?.type,
-          panel: highlight,
-        });
-      },
+    (highlight) => (elementId) => (ev) => {
+      ev.preventDefault();
+      setHighlights({
+        elementId: elementId || selectedElement?.id,
+        highlight,
+      });
+    },
     [setHighlights, selectedElement]
   );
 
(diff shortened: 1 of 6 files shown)


packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
```diff
@@ -157,18 +171,43 @@ const useQuickActions = () => {
...
+        setHighlights({
+          elementId: elementId || selectedElement?.id,
+          highlight,
+        });
+
+        trackEvent('quick_action', {
+          name: 'focus_panel',
+          element: elementType || selectedElement?.type,
+          panel: highlight,
+        });
```
samwhale: Most quick actions focus a panel in the library or the inspector. The `selectedElement` usually does not change when clicking these actions.

The one case when the selected element needs to change is when there is no selected element (default state). Clicking the quick actions should select the background element and still highlight the relevant panel.

---

There's one place where we need to specify that the background _should_ be selected if it already isn't, so we pass in an override when we need to specify which element to highlight. This makes sure that the inspector has the correct panel to highlight.

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
```diff
@@ -223,12 +257,15 @@ const useQuickActions = () => {
...
+  const noElementSelectedActions = useMemo(() => {
     return [
       {
         Icon: Bucket,
         label: ACTION_TEXT.CHANGE_BACKGROUND_COLOR,
-        onClick: handleFocusPageBackground(backgroundElement?.id),
+        onClick: handleFocusPageBackground({
+          elementId: backgroundElement?.id,
+          elementType: backgroundElement?.type,
+        }),
```
samwhale: Only spot where we need to pass in the element to select.

BrittanyIRL (approved):
Really helpful comments, I like the refactor!

spacedmonkey (approved):

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
swissspidy: What's an example value of `highlight` here?

To me it seems a bit confusing that the quick action being clicked on is "Change color", but in telemetry it's a "focus_panel" event. Can't we track "change_color" instead?

It's especially confusing since the "Change color" and "Edit text" actions highlight the same Text panel in the Design Panel. How are these being distinguished here?

samwhale: `highlight` would be the name of the panel as defined here: https://github.com/google/web-stories-wp/blob/main/packages/story-editor/src/app/highlights/states.js#L34-L55

Yeah I agree it's a bit confusing. I can update that so that the action that's clicked is what is recorded rather than the result of the action 👍 

>It's especially confusing since the "Change color" and "Edit text" actions highlight the same Text panel in the Design Panel. How are these being distinguished here?

Good shout: it's not being distinguished here since they both end up doing the same thing. I can add some granularity here.

swissspidy: > Yeah I agree it's a bit confusing. I can update that so that the action that's clicked is what is recorded rather than the result of the action 👍

I think that would be more valuable for us to have in analytics. And should also help with the "Change color" and "Edit text" distinction.

samwhale: Should be fixed now 😄

packages/story-editor/src/components/canvas/karma/quickActions.karma.js
```diff
@@ -502,7 +502,7 @@ describe('Quick Actions integration', () => {
       // The "Replace background media" quick action also triggers a help center tip which steals focus.
       // Open the tip now to avoid the focus stealing after the quick menu button is clicked.
       const { actions } = await fixture.renderHook(() => useHelpCenter());
-      actions.openToUnreadTip.text(KEYS.ADD_BACKGROUND_MEDIA);
+      actions.openToUnreadTip(KEYS.ADD_BACKGROUND_MEDIA);
```
samwhale: Looks like this was failing. Looking at the help center provider it seems that `openToUnreadTip` is a function without a `.text` property.

swissspidy (approved):

PR #8638, opened 2021-08-10
Media Hotlinking: Add initial version

miina:
## Context
Adds the possibility to add media by inserting a link.

## To-do

- [ ] Some refactoring
- [x] Feature flag
- [x] Tests

## User-facing changes
Screenshot 2021-08-11 at 18 38 58
Screenshot 2021-08-11 at 18 38 53

## Testing Instructions
Feature flag for hotlinking must be enabled first.
1. Click on the new "Insert by link" button.
2. Try inserting images/videos via link.
3. verify that invalid links show an error.
4. Verify that valid links create a new media element correctly.
5. If the user has permissions to upload media, verify that for a video inserted by a link, a poster image is created, too.
6. Verify that everything looks as before when disabling the hotlinking feature.

Note: Not all sites allow linking externally like this, in that case, an error starting with "Media failed to load" will be displayed under the input.

### QA

- [ ] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1.

## Checklist

- [ ] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [ ] I have tested this code to the best of my abilities
- [ ] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [ ] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [ ] This PR contains automated tests (unit, integration, and/or e2e) to verify the code works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [ ] I have added documentation where necessary
- [ ] I have added a matching `Type: XYZ` label to the PR

---

Fixes #471

PR #8731, opened 2021-08-19
Media Hotlinking: add REST API endpoint for validation

### LINKED ISSUE #8666: Media Hotlinking: server-side validation
## Feature Description

For hotlinking, let's add some server-side validation to verify that the content type is actually video or image, by doing a `HEAD` request to the given URL. The REST API also will nicely validate the URL itself for us.

This should be done upon insertion in the editor, the user should get notified when the URL seems invalid.

See `Link_Controller` / element links for inspiration.

## Additional Context

---

spacedmonkey:
## Summary

A new REST Controller that pings the URL, to ensure that url is valid. 

If url is valid, type, size, mime type and file name are returned. Otherwise an error state is returned. 
There are checks for, 

- Valid mime type
- Empty host
- Empty path
- Invalid response from server. 

How to test

```
await wp.apiFetch({path: 'web-stories/v1/hotlink?url=https://www.learningcontainer.com/wp-content/uploads/2020/05/sample-webm-file.webm'})
```

```
await wp.apiFetch({path: 'web-stories/v1/hotlink?url=https://www.learningcontainer.com/wp-content/uploads/2020/05/sample-avi-file.avi'})
```

Example response 
Screenshot 2021-08-23 at 14 14 49

Screenshot 2021-08-19 at 15 03 21

## Testing Instructions

1.

## Checklist

---

Fixes #8666


includes/REST_API/URL_Controller.php
```diff
@@ -0,0 +1,245 @@
...
+use WP_REST_Request;
+use WP_REST_Response;
+use WP_REST_Server;
+
+/**
+ * API endpoint to allow parsing of metadata from a URL
+ *
+ * Class URL_Controller
+ */
+class URL_Controller extends REST_Controller {
```
swissspidy: This is too similar to `Link_Controller`.

Let's rename to `Media_Hotlinking_Controller` or something.

includes/REST_API/URL_Controller.php
```diff
@@ -0,0 +1,245 @@
...
+		$cache_key = 'web_stories_url_data_' . md5( $url );
+
+		$data = get_transient( $cache_key );
+		if ( ! empty( $data ) ) {
+			$response = $this->prepare_item_for_response( json_decode( $data, true ), $request );
+
+			return rest_ensure_response( $response );
+		}
+
+		$response = wp_safe_remote_get( $url );
```
swissspidy: Since we're not actually using the body and are only interested in the headers, let's make sure this is a HEAD request accordingly.

```suggestion
		$response = wp_safe_remote_head( $url );
```

includes/REST_API/URL_Controller.php
```diff
@@ -0,0 +1,245 @@
...
+
+		if ( is_wp_error( $response ) && 'http_request_failed' === $response->get_error_code() ) {
+			return new WP_Error( 'rest_invalid_url', __( 'Invalid URL', 'web-stories' ), [ 'status' => 404 ] );
+		}
+
+		$code = wp_remote_retrieve_response_code( $response );
+		if ( ! in_array( $code, [ WP_Http::OK, WP_Http::MOVED_PERMANENTLY, WP_Http::FOUND ], true ) ) {
```
swissspidy: `wp_safe_remote_get()`/`wp_safe_remote_head()` follows redirects, so I don't think this will be anything else than `WP_Http::OK`.

So `WP_Http::OK !== wp_remote_retrieve_response_code( $response )` like in `Link_Controller` should be enough.

And if there are way more redirects than what `wp_remote_*` follows, the mime type check below would catch this anyway.

includes/REST_API/URL_Controller.php
```diff
@@ -0,0 +1,245 @@
...
+		$path = wp_parse_url( $url, PHP_URL_PATH );
+		$exts = $this->get_file_type_exts( [ $mime_type ] );
+		$ext  = end( $exts );
+
+		$file_name = basename( $path );
+		$data      = [
+			'ext'       => $ext,
+			'file_name' => $file_name,
+			'mime_type' => $mime_type,
+		];
```
swissspidy: Let's also add something like `'size' => $headers['content-length'],` in case we might need this in the future

includes/REST_API/URL_Controller.php
```diff
@@ -0,0 +1,245 @@
...
+	/**
+	 * Checks if current user can process links.
+	 *
+	 * @since 1.11.0
+	 *
+	 * @return true|WP_Error True if the request has read access, WP_Error object otherwise.
+	 */
+	public function parse_url_permissions_check() {
+		if ( ! $this->get_post_type_cap( Story_Post_Type::POST_TYPE_SLUG, 'edit_posts' ) ) {
+			return new WP_Error( 'rest_forbidden', __( 'Sorry, you are not allowed to process links.', 'web-stories' ), [ 'status' => rest_authorization_required_code() ] );
```
swissspidy:
```suggestion
			return new WP_Error( 'rest_forbidden', __( 'Sorry, you are not allowed to insert external media.', 'web-stories' ), [ 'status' => rest_authorization_required_code() ] );
```

includes/REST_API/URL_Controller.php
```diff
@@ -0,0 +1,245 @@
...
+		set_transient( $cache_key, wp_json_encode( $data ), $cache_ttl );
+
+		return rest_ensure_response( $response );
+	}
+
+
+	/**
+	 * Prepares a single lock output for response.
+	 *
+	 * @since 1.10.0
```
swissspidy: This needs to be corrected here and also in Link_Controller

includes/REST_API/URL_Controller.php
```diff
@@ -0,0 +1,245 @@
...
+							'format'      => 'uri',
+						],
+					],
+				],
+			]
+		);
+	}
+
+	/**
+	 * Parses a URL to return some metadata for inserting links.
```
swissspidy:
```suggestion
	 * Parses a URL to return some metadata for inserting external media.
```

includes/REST_API/URL_Controller.php
```diff
@@ -0,0 +1,245 @@
...
+					'type'        => 'string',
+					'context'     => [ 'view', 'edit', 'embed' ],
+				],
+				'mime_type' => [
+					'description' => __( 'Mime Type', 'web-stories' ),
+					'type'        => 'string',
+					'context'     => [ 'view', 'edit', 'embed' ],
+				],
+				'ext'       => [
+					'description' => __( 'ext', 'web-stories' ),
```
swissspidy:
```suggestion
					'description' => __( 'File extension', 'web-stories' ),
```

includes/REST_API/URL_Controller.php
```diff
@@ -0,0 +1,245 @@
...
+		}
+
+		$headers = wp_remote_retrieve_headers( $response );
+
+		$mime_type          = $headers['content-type'];
+		$allowed_mime_types = $this->get_allowed_mime_types();
+		$allowed_mime_types = array_merge( ...array_values( $allowed_mime_types ) );
+
+		if ( ! in_array( $mime_type, $allowed_mime_types, true ) ) {
+			return new WP_Error( 'rest_invalid_file_type', __( 'File type', 'web-stories' ), [ 'status' => 404 ] );
```
swissspidy:
```suggestion
			return new WP_Error( 'rest_invalid_file_type', __( 'Invalid file type', 'web-stories' ), [ 'status' => 404 ] );
```

swissspidy (commented):
Great progress! Once #8638 is merged we will know better if this PR does the trick for it.

includes/REST_API/Media_Hotlinking_Controller.php
```diff
@@ -0,0 +1,290 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Web_Stories\REST_API;
+
+use Google\Web_Stories\Story_Post_Type;
+use Google\Web_Stories\Traits\Post_Type;
+use Google\Web_Stories\Traits\Types;
+use phpDocumentor\Reflection\Types\Integer;
```
swissspidy: Added by mistake?

includes/REST_API/Media_Hotlinking_Controller.php
```diff
@@ -0,0 +1,290 @@
...
+
+		$parts = wp_parse_url( $url );
+
+		if ( ! isset( $parts['host'] ) ) {
+			return new WP_Error( 'rest_invalid_url_host', __( 'Invalid URL Host', 'web-stories' ), [ 'status' => 400 ] );
+		}
+
+		if ( ! isset( $parts['path'] ) || empty( $parts['path'] ) ) {
+			return new WP_Error( 'rest_invalid_url_path', __( 'Invalid URL Path', 'web-stories' ), [ 'status' => 400 ] );
+		}
```
swissspidy: Can you elaborate on why this was added? It seems overly defensive. If I host a random cat image at `https://pascalscatimage.com`, this would be blocked by this check, no?

includes/REST_API/Media_Hotlinking_Controller.php
```diff
@@ -0,0 +1,290 @@
...
+
+		if ( empty( $url ) ) {
+			return new WP_Error( 'rest_invalid_url', __( 'Invalid URL', 'web-stories' ), [ 'status' => 400 ] );
+		}
+
+		$parts = wp_parse_url( $url );
+
+		if ( ! isset( $parts['host'] ) ) {
+			return new WP_Error( 'rest_invalid_url_host', __( 'Invalid URL Host', 'web-stories' ), [ 'status' => 400 ] );
+		}
```
swissspidy: Is there ever a case where this happens? 🤔  The REST API bails very early on invalid URLs already, so I don't see how/when this would be triggered.

includes/Plugin.php
```diff
@@ -117,6 +117,7 @@ class Plugin extends ServiceBasedPlugin {
 		'user.capabilities'            => User\Capabilities::class,
 		'rest.embed_controller'        => REST_API\Embed_Controller::class,
 		'rest.link_controller'         => REST_API\Link_Controller::class,
+		'rest.url_controller'          => REST_API\Media_Hotlinking_Controller::class,
```
swissspidy:
```suggestion
		'rest.media_hotlinking_controller' => REST_API\Media_Hotlinking_Controller::class,
```

includes/REST_API/Media_Hotlinking_Controller.php
spacedmonkey: The user can submit whatever they want to this url. 

So first case. 

#### Random string - example wibble. 
This would be rejected, as it would fail to be parsed as a url

#### Relative path /image/test.png
User could reasonably think that they can use a relative path. I think we should avoid this, as it is messy. So would not have a host, which is why we should regret it. 

#### Top level domain - https://pascalscatimage.com
Why anyone do this, it is insane. Either way, the likelihood is low that this is valid. It also generates other problems. Our code requires a file, to say generate a poster image etc. Not having a file name would break a lot of stuff. I think it is an edge can we should not allow.

includes/REST_API/Media_Hotlinking_Controller.php
spacedmonkey: > The REST API bails very early on invalid URLs already, so I don't see how/when this would be triggered.

See. https://github.com/google/web-stories-wp/pull/8731#discussion_r692816954

Relative paths, are a valid url. In my testing, it doesn't trigger an error when I passed, as `/image/test.png`. 

It will become clear, once I have unit test.

includes/REST_API/Media_Hotlinking_Controller.php
```diff
@@ -0,0 +1,301 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Web_Stories\REST_API;
+
+use Google\Web_Stories\Story_Post_Type;
+use Google\Web_Stories\Traits\Post_Type;
+use Google\Web_Stories\Traits\Types;
+use phpDocumentor\Reflection\Types\Integer;
```
spacedmonkey:
```suggestion
```

includes/REST_API/Media_Hotlinking_Controller.php
spacedmonkey: Friendly reminder, this is a draft, I don't want a code review. I requested a functional test to check it is fit for requirements. I will tidy up later.

swissspidy:
Functionality wise the API endpoint seems fine. With #8638 now merged this PR can now update the JS code to make use of it. Only then can we know whether it works as expected and every bit of data is in there that we need.

includes/REST_API/Media_Hotlinking_Controller.php
swissspidy: I understand that. Just trying to be helpful. I spotted it and might have forgotten later on, so I wrote it down. Better to point this out instead of having it accidentally slip through.

includes/REST_API/Media_Hotlinking_Controller.php
```diff
@@ -0,0 +1,300 @@
...
+ * Class Media_Hotlinking_Controller
+ */
+class Media_Hotlinking_Controller extends REST_Controller {
+	use Post_Type, Types;
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		$this->namespace = 'web-stories/v1';
+		$this->rest_base = 'url';
```
swissspidy: Update to `media-hotlinking` as well due to the confusion with `link`?

includes/REST_API/Media_Hotlinking_Controller.php
```diff
@@ -0,0 +1,300 @@
...
+					'type'        => 'string',
+					'context'     => [ 'view', 'edit', 'embed' ],
+				],
+				'file_size' => [
+					'description' => __( 'File size', 'web-stories' ),
+					'type'        => 'integer',
+					'context'     => [ 'view', 'edit', 'embed' ],
+				],
+				'mime_type' => [
+					'description' => __( 'Mime Type', 'web-stories' ),
```
swissspidy: Nit: 

```suggestion
					'description' => __( 'Mime type', 'web-stories' ),
```

Since the other descriptions are not capitalized either

spacedmonkey:
Should I add a check for if the file is embeddable? As in having the following headers?
```
access-control-allow-origin: *
cross-origin-resource-policy: cross-origin
```` 

I can add some data to response. `is_embeddable: true`. Thoughts @swissspidy @miina ?

**Comment by swissspidy** [comment]
> Should I add a check for if the file is embeddable? As in having the following headers?
> 
> ```
> access-control-allow-origin: *
> cross-origin-resource-policy: cross-origin
> ```
> 
> I can add some data to response. `is_embeddable: true`. Thoughts @swissspidy @miina ?

This shouldn't be needed and is also error prone, so no.

1. Sites could send different response headers depending the request headers.  
  So on the server you could get these headers, but not when requesting on the frontend - or vice-versa.
2. This information is not needed/redundant. On the frontend we can just try to load the media with some try-catch and show the placeholder if it fails to load.

File: .phpstorm.meta.php [hunk 17]
```diff
@@ -7,67 +7,68 @@
 
 		// TODO: I'd like to use Plugin::SERVICES directly here but it doesn't seem to work.
 		map( [
```
swissspidy: Isn't the arrow alignment incorrect now?

swissspidy:
> Functionality wise the API endpoint seems fine. With #8638 now merged this PR can now update the JS code to make use of it. Only then can we know whether it works as expected and every bit of data is in there that we need.

Are we aligned on this? Are there thoughts on whether this should be done in a separate PR instead? Not seeing any response to this so far.

spacedmonkey:
> Are we aligned on this? Are there thoughts on whether this should be done in a separate PR instead? Not seeing any response to this so far.

I was in the understanding that @miina would be implementing this in the editor. I have been speaking with her private with this in mind. There was no mention on the original ticket about needing an implementation.

.phpstorm.meta.php
spacedmonkey: Fixed

swissspidy:
> > Are we aligned on this? Are there thoughts on whether this should be done in a separate PR instead? Not seeing any response to this so far.
> 
> I was in the understanding that @miina would be implementing this in the editor. I have been speaking with her private with this in mind. There was no mention on the original ticket about needing an implementation.

I was mentioning this because I want to avoid merging something half-baked and then having lots of back and forth to fix the API endpoint and the implementation.

@miina Could you please review this PR and verify this fits the needs for streamlining the hotlinking on the client-side? Perhaps you could even whip up a quick POC for it. I'd definitely not want to merge this without your approval on it for sure.

There's no separate ticket for this implementation, but if y'all think one is needed, please create one.

miina:
@swissspidy The PR looks good to me, I tested with a few URLs as well and the data received is sufficient.

I think we could either:
1. Merge this PR, follow up with a separate PR using the new endpoint, and if any changes are still needed to the REST API, make those in a (or the same) follow-up PR.
2. Leave this PR waiting while I start with the client-side updates based on this PR -- once the new PR is in a POC state and can confirm the endpoint doesn't need changes, we can merge this -- I could probably do that tomorrow.

Either option is ok by me, let me know if you have a preference.

miina (approved):

miina:
Added a new issue as well: https://github.com/google/web-stories-wp/issues/8773

swissspidy:
Thanks @miina! Let's merge this then for now to make working on #8773 a tad easier.

PR #8727, opened 2021-08-18
Static Site: Add Template Poster Assets 

BrittanyIRL:
## Context

We're updating the dashboard's explore templates view to use screenshots instead of load every template as a story. 

## Summary

- Follow up from #8693  
- Adds screenshots for every page of every template 

## Relevant Technical Choices

screenshots are stored in each template subdirectory of images by page number. 

## User-facing changes

None

### QA

- [x] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1.

### UAT

- [ ] UAT should use the same steps as above.

This PR can be tested by following these steps:

1.

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [ ] This PR contains automated tests (unit, integration, and/or e2e) to verify the code works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [x] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Partially addresses #8373

Issue #8373, opened 2021-07-19
Dashboard: Replace Template Stories with Images


BrittanyIRL:
TBD on how we do this. 

Based on the conversation here: https://github.com/google/web-stories-wp/pull/8274/files#r670841347 

Templates in the dashboard are actual stories that are loading. we recently updated the my stories view to use the poster image of a story (#8274). Templates currently don't have poster image to use for this. 

This ticket is to work through how we want to create image assets for templates and then implement that for existing templates and update the template creation process so that future templates have this built in. 

Other questions: 
- are we ok leaving template details views as is so that animations can play?

swissspidy:
The discussion at https://github.com/google/web-stories-wp/pull/8274/files#r670841347 mentioned that the template preview images should be like poster images, without author, title, etc. in it.

But is that correct? IMO the preview should be === the cover page, ≠ the poster image. So it _should_ _include_ all those elements. (cc @choumx)

That would make creation of these preview images much easier as it basically just means taking a screenshot.

I am not sure we can just grab the images from Figma though as the aspect ratio might be different.

But perhaps there's another easy way to generate these?

I'll talk to the templates team to hear their thoughts.

> are we ok leaving template details views as is so that animations can play?

I think it would be OK to keep this part as-is for now

One disadvantage of the animated previews is that they are not super accurate. Text is usually displayed way bigger. (Can we fix that?)

If the animated previews don't provide much value, we could consider getting rid of them and only displaying static images.

**Alternative 1:**

We could look into showing preview images and loading the animated previews with a small delay. So that when I am only quickly browsing through templates it's much faster.

**Alternative 2:**

Use small videos for the previews instead.

----

Note: Also worth thinking about doing the same for page templates in the editor. #7985 is related.

BrittanyIRL:
Thanks, Pascal. I didn't mean to spread confusion by using the word `poster image` here 😅  - just that we would need some sort of likeminded substitute for templates since they do not have a poster image nor shall they.  The aspect ratio in figma isn't 3:4 so we can't just pull a screenshot. 

I know Marcin did some exploration to generate PNGs (re text sets that could also be used here probably) (#8117) but I'm wondering if we need a substitute in the meantime or just let it be. 

If we do want to just let it be for now, perhaps we just load the first page of the story and remove the hover/focus so it never "plays".  

Aesthetically, it's a little weird to me to have the my stories grid and templates be different ratios and I'd like to bring consistency sooner rather than later. 

[...]

PR #8737, opened 2021-08-19
Static Site: Add Template Poster WebP Assets

BrittanyIRL:
## Context

Follow up from #8727 to use WebP format of template posters when possible (browser support isn't 100% yet so we need png as a fallback). 

## Summary

- Every template poster image converted to WebP
https://github.com/google/web-stories-wp/pull/8727#issuecomment-901449899

## Testing Instructions

1.

## Checklist

---

Partially addresses #8373


spacedmonkey:
>  (browser support isn't 100% yet so we need png as a fallback).

Support is 95%, and unsupported versions, are like ie11, which WordPress doesn't support anyway.

swissspidy:
> > (browser support isn't 100% yet so we need png as a fallback).
> 
> Support is 95%, and unsupported versions, are like ie11, which WordPress doesn't support anyway.

The question is whether we support it, not WordPress. But yes, we don't support IE 11 either.

WebP is supported by all browsers we officially support, for Safari it just means on newer OS versions, but I'd say that's negligible.

Adding WebP versions in addition to PNG is of course the safest route. But if we can go all in on WebP that would make our lives easier.

swissspidy (approved):

PR #8742, opened 2021-08-20
Color Picker accessibility improvements

### LINKED ISSUE #8657: Add keyboard interaction model for new basic color picker
## Feature Description

The new basic color picker should be easier to use when using a keyboard-only. Ensure full a11y compliance including

- Keyboard navigation 
- Focus styling
- Labels

---

miina:
## Summary
- Adds label to the swatches
- Adds keyboard navigation

## Testing Instructions
- Open the color picker
- Tab through all the fields and verify the focus outline is displayed correctly
- Verify the swatches have aria-label
- Verify arrow keys can be used to navigate in a color group (e.g. within the Default colors), and tab can be used to move between the color groups

1.
## Checklist

---

Fixes #8657


packages/story-editor/src/components/colorPicker/basicColorList.js
```diff
@@ -85,6 +95,8 @@ function BasicColorList({
             pattern={pattern}
             isSelected={isSelected}
             isDisabled={isDisabled}
+            tabIndex={0 === i ? 0 : -1}
+            aria-label={__('Apply color', 'web-stories')}
```
miina: @barklund In the issue description it was mentioned that the labels need to be added to the new Color picker -- the only missing label I currently saw was for the Color Swatches. 

Were there any other labels you had in mind? 🤔 

It also seems to me that the focus styling is correct everywhere, but it is mentioned in the origin issue description. Let me know if I'm missing something.

barklund: I was thinking it might make sense to turn the parent swatch list in to a `role="listbox"` with a fitting label and the buttons into `role="option"`, but not sure about the semantics here. Alternatively, the swatch list should probably just be a `role="list"`, but probably still have a label, and the buttons should be (maybe wrapped in) `role="listitem"`.

Extending this a bit, visual users see a grouping of swatches into related colors in a row. This could be implemented in the semantics too using either `role="optgroup"` or `role="group"` depending on the above choice and these should be labelled too.

Also, the color swatches should actually use named colors rather than plain hex descriptions as their labels (so, "pink", "salmon", etc), but that's a whole other can of worms!

barklund: Oh, and when you do get to do this color naming, you'll find that the first two browns are identical in Figma, and in the code only differ by a single hex value (as identical values would cause duplicate `key`s).

Feel free to resolve this with UX 😂 .

barklund: Oh, I stand corrected. The proper role for an optgroup is still just [`role="group"`](https://w3c.github.io/aria/#group).

barklund (commented):
(sorry, wrote this review last week, but forgot to submit my comments - doing it now!)

packages/story-editor/src/components/colorPicker/basicColorList.js
miina: > Also, the color swatches should actually use named colors rather than plain hex descriptions as their labels (so, "pink", "salmon", etc), but that's a whole other can of worms!

I think we don't have to perhaps go that deep into this, considering there are also gradients available, probably not worth the effort as long as it's understandable generally. We could create a separate issue to look into that, including the Saved Color Panels then.

Reading about `listbox`:
> Elements with the role listbox have an implicit aria-orientation value of vertical.

Also, it looks like arrow up/down should move to the next option, which would not be our case. Generally, it looks like `listbox` would be more for an actual list (assuming one option per row).

Perhaps `group` with `listitem` would be the best fit then? If we'd use `listitem` role, we'd ideally actually use the `ul` and `li` tags as well, right?

Thoughts?

packages/story-editor/src/components/colorPicker/basicColorList.js
```diff
@@ -85,6 +99,8 @@ function BasicColorList({
             pattern={pattern}
             isSelected={isSelected}
             isDisabled={isDisabled}
+            tabIndex={0 === i ? 0 : -1}
```
barklund: If you open it for the background, the tabIndex should default to 0 on the color that's already highlighted for the background (`isSelected`) and not just the first item. However, this is actually doubly broken now, because it sets tabIndex=0 on the first item, which is transparent, which is disabled for the background, so you can't tab to the list of default colors at all.

Furthermore, I have a semitransparent color as my second saved color, so if tabbing to those, I can't use arrow keys, as the second item is not focusable. `useRovingTabIndex` should probably be updated to ignore/skip disabled buttons.

packages/story-editor/src/components/colorPicker/basicColorList.js
```diff
@@ -68,10 +79,13 @@ function BasicColorList({
...
 }) {
   const colorAsBackground = getPatternAsString(color);
+  const listRef = useRef(null);
+
+  useRovingTabIndex({ ref: listRef });
 
   return (
-    <SwatchList>
-      {colors.map((pattern) => {
+    <SwatchList ref={listRef}>
```
barklund: In my PR I added `{...rest}` here to allow the parent component to add `aria-labelledby` with a reference to the title for the list, which each list has. A `listbox` should have a label.

barklund (dismissed):
This has some bugs - disabled swatches should be skipped and the initial selection should be highlighted correctly.

packages/story-editor/src/components/colorPicker/basicColorList.js
miina: Hm, I assumed it did skip disabled items, I think we had this problem before already somewhere 🤔

packages/story-editor/src/components/colorPicker/basicColorList.js
```diff
@@ -68,23 +79,44 @@ function BasicColorList({
...
+  const selectedSwatchList = colors
+    .map((pattern, i) => {
+      const patternAsBackground = getPatternAsString(pattern);
+      const isSelected = colorAsBackground === patternAsBackground;
+      if (isSelected) {
+        return i;
+      }
+      return false;
+    })
+    .filter((c) => c !== false);
```
barklund: Well, I guess there's only one that can be selected, though?

So `findIndex` is probably a reasonable version:

```suggestion
  const selectedSwatchIndex = colors
    .map(getPatternAsString)
    .findIndex((c) => colorAsBackground === c);
```

packages/story-editor/src/components/colorPicker/basicColorList.js
```diff
@@ -68,23 +79,44 @@ function BasicColorList({
...
+    <SwatchList ref={listRef}>
+      {colors.map((pattern, i) => {
         const isTransparentAndInvalid = !allowsOpacity && hasOpacity(pattern);
         const isGradientAndInvalid = !allowsGradient && hasGradient(pattern);
         const isDisabled = isTransparentAndInvalid || isGradientAndInvalid;
 
         const patternAsBackground = getPatternAsString(pattern);
         const isSelected = colorAsBackground === patternAsBackground;
+        // By default, the first swatch can be tabbed into, unless there's a selected one.
+        let tabIndex = i === 0 ? 0 : -1;
```
barklund: Does it work if you have a custom background color that isn't in the list and open the color picker? Wouldn't it then set `tabIndex=0` on the first item in the list, which is the transparent one, that is disabled?

packages/story-editor/src/components/colorPicker/basicColorList.js
```diff
@@ -68,23 +79,44 @@ function BasicColorList({
...
-    <SwatchList>
-      {colors.map((pattern) => {
+    <SwatchList ref={listRef}>
+      {colors.map((pattern, i) => {
         const isTransparentAndInvalid = !allowsOpacity && hasOpacity(pattern);
         const isGradientAndInvalid = !allowsGradient && hasGradient(pattern);
         const isDisabled = isTransparentAndInvalid || isGradientAndInvalid;
 
         const patternAsBackground = getPatternAsString(pattern);
         const isSelected = colorAsBackground === patternAsBackground;
```
barklund, line 100: This has already been resolved above now.

packages/story-editor/src/utils/useRovingTabIndex.js
```diff
@@ -82,7 +82,11 @@ export function getSiblingDirection(isRTL, key) {
  * @return {Element} The sibling.
  */
 function getNextSibling(e, siblingDirection) {
```
barklund: Could we rename to `getNextEnabledSibling` just to clear up some confusion?

packages/story-editor/src/components/colorPicker/basicColorList.js
miina: Ah, right 💡

barklund: Note that similar could happen for the saved colors where the first item could be disallowed because of either transparency or gradient.

miina: I think it should work as expected now.

barklund (approved):

barklund:
I merged #8799, so now there are some conflicts, but should be trivial.

PR #8751, opened 2021-08-23
Use file-loader for importing images

### LINKED ISSUE #8744: Move image files to their packages
## Feature Description

Currently there are two images files in the [assets/images](https://github.com/google/web-stories-wp/tree/main/assets/images) directory. These files should be move package they are connected to, to keep asset with their package.

swissspidy:
## Summary

This PR updates the webpack configuration to use [`file-loader`](https://v4.webpack.js.org/loaders/file-loader/) for loading images in the source code instead of having to provide actual URLs, which is a bit brittle.

Side effect: images will be stored in `assets/js/images` instead of `assets/images` due to this change, since that is the configured public path.

## Relevant Technical Choices

* Removes any usage of `assetsUrl` in the JS codebase as now file-loader is being used
* Removes the `WEBSTORIES_ASSETS_URL` constant and the `assetsUrl` config value as those are no longer needed in the plugin
* Bonus point: updates storybook config to also use file-loader so the images are actually working there as well

Removes

## User-facing changes

N/A

## Testing Instructions

1.

## Checklist

---

Fixes #8744


.storybook/main.cjs
```diff
@@ -102,6 +102,17 @@ module.exports = {
...
         ],
         include: [/images\/.*\.svg$/],
+      },
+      {
+        test: /\.(png|jpe?g|gif|webp)$/i,
+        use: [
+          {
+            loader: 'file-loader',
+            options: {
+              outputPath: 'images',
```
spacedmonkey, line 112:
```suggestion
              outputPath: '../images',
```

Currently, images are saved in the following path. `assets/js/images`. Changing the value to this will save the images in `assets/images`

webpack.config.cjs
```diff
@@ -150,6 +150,17 @@ const sharedConfig = {
         use: [MiniCssExtractPlugin.loader, 'css-loader'],
         sideEffects: true,
       },
+      {
+        test: /\.(png|jpe?g|gif|webp)$/i,
+        use: [
+          {
+            loader: 'file-loader',
+            options: {
+              outputPath: 'images',
```
spacedmonkey:
```suggestion
              outputPath: '../images',
```
Currently, images are saved in the following path. assets/js/images. Changing the value to this will save the images in assets/images

spacedmonkey (changes requested):
Requested one small change, other than that, we are looking good.

webpack.config.cjs
swissspidy: I thought about doing that but I don't think it actually adds any value.

swissspidy: NB: this change would cause URLs to look like `https://web-stories.local/wp-content/plugins/web-stories-wp/assets/js/../images/631db33b339209fd6ab0bfa022326552.png`, so `publicPath` would need to be updated to remedy that. But I'd prefer to not override the default `__webpack_public_path__` value.

spacedmonkey: > I thought about doing that but I don't think it actually adds any value.

Images are currently stored in `assets/images`. Similarly, css is stored in it's own directory. Why would we change the path of images now?

swissspidy: Now it's just a directory that's completely dynamic, can be completely excluded and never looked at The structure doesn't really matter.

spacedmonkey (approved):

PR #8752, opened 2021-08-23
Fix flaky cloning test

### LINKED ISSUE #8677: Fix flaky clone integration test
## Bug Description

The test `should correctly clone 1 element` in `karma/cloneSelection.js` fails as often as not. The test has been disabled for now, but should be fixed.

---

miina:
## Context
Fixes the testing of cloning an element (`should correctly clone 1 element`)

## Relevant Technical Choices
- This PR also switches to using shapes instead of images in case of some tests. This is for avoiding the going over the safezone. We can change that back though if preferred since it doesn't seem to really make a difference. However, it might be good to test with more than 1 element type.

## User-facing changes
N/A

## Checklist

---

Fixes #8677

### DIFF
--- a/packages/story-editor/src/components/canvas/karma/cloneSelection.karma.js
+++ b/packages/story-editor/src/components/canvas/karma/cloneSelection.karma.js
@@ -47,7 +47,6 @@ fdescribe('Clone element integration', () => {
 
     // Save final objects
     img1 = await getElementByIndex(2);
-    img2 = await getElementByIndex(3);
   });
 
   afterEach(() => {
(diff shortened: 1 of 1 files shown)


packages/story-editor/src/components/canvas/karma/cloneSelection.karma.js
```diff
@@ -22,103 +22,114 @@ import { useStory } from '../../../app/story';
 
 describe('Clone element integration', () => {
   let fixture;
-  let bg, img1, img2;
+  let bg, bgRect;
```
miina, line 25: @barklund See the Relevant Technical Choices section in the PR description that explain the extensive changes here.

barklund (approved):
Looks like it's working perfectly now - good effort!

PR #8770, opened 2021-08-24
Media: Fix FFmpeg memory leaks

### LINKED ISSUE #8768: High memory consumption when using ffmpeg.wasm
## Bug Description

User report:

https://wordpress.org/support/topic/aw-snap-when-optimizing-videos/

Related upstream issues:

https://github.com/ffmpegwasm/ffmpeg.wasm/issues/111
https://github.com/ffmpegwasm/ffmpeg.wasm/issues/136
https://github.com/ffmpegwasm/ffmpeg.wasm/issues/200

There is an issue with memory consumption when invoking ffmpeg.wasm, where the worker threads are not properly closed by ffmpeg.wasm itself and accumulating with each use. This massively increases memory usage over time, which can only be "fixed" by reloading the tab (or closing it and opening it again).

Related to that: when uploading multiple files, multiple FFmpeg instances are created instead of re-using an existing one, which also increases memory usage. But instances can't really be shared because ffmpeg.wasm can only run one command at a time. So while not super great, this is negligible if we properly terminate instances after use.

## Expected Behaviour

FFMpeg threads should be properly cleaned up after usage.

`FS.unlink()` doesn't seem to bring the expected effect

Luckily, the new `exit()` method introduced [in the latest release](https://github.com/ffmpegwasm/ffmpeg.wasm/releases/tag/v0.10.0) (which we thought would only be helpful for _cancelling_ in-progress threads) helps with this, as this will properly close the threads.

So ideally we call `exit()` right after processing each file. In my testing that successfully closed all the worker threads.

## Steps to Reproduce

1. Open Chrome's Task Manager and the "Sources" tab in the dev tools.
2. Upload a video to the canvas so it gets transcoded
3. See a bunch of worker threads being created but not closed, even after upload is finished.
4. Upload another video and see how memory usage is steadily increasing

## Screenshots

Screenshot 2021-08-24 at 11 28 13

## Additional Context

- Plugin Version: 1.10.0
- WordPress Version: 5.8
- Operating System: macOS
- Browser: Chrome

swissspidy:
## Context

There is an issue with memory consumption when invoking ffmpeg.wasm, where the worker threads are not properly closed by ffmpeg.wasm itself and accumulating with each use.

## Summary

This PR fixes ffmpeg.wasm by completely exiting after file processing is done.

## Relevant Technical Choices

* Refactor `useFFmpeg` a little bit to better reuse FFmpeg config options
* Add `console.error()` statements so that errors using ffmpeg.wasm are not secretly discarded, making debugging easier.
* Moves the `exit()` call to the `finally` block so that it runs no matter what.
* Wraps the `exit()` call in a try catch to prevent the errors from bubbling up

## User-facing changes

N/A

## Testing Instructions

1. Open Chrome's Task Manager and the "Sources" tab in the dev tools.
2. Upload a video to the canvas so it gets transcoded
3. See a bunch of worker threads being created and memory usage to increase significantly
4. Wait for upload to finish
5. See worker threads being removed and memory decreasing again to previous levels

No

No

No

## Checklist

---

Fixes #8768


spacedmonkey (approved):
Done lots of testing on this one! This is great and will fix lots of problems for users and memory. Love it!

PR #7494, opened 2021-05-11
Add hooks for create and delete sites.

spacedmonkey:
## Summary

Tweak, the base plugin template to hook in on new site creation and deletion in multisite context. The pattern of having a method run on new site / delete follows the pattern of activation / deactivation. 

### QA

- [ ] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1.

### UAT

- [ ] UAT should use the same steps as above.

This PR can be tested by following these steps:

1.

## Checklist

- [ ] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [ ] I have tested this code to the best of my abilities
- [ ] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [ ] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [ ] This PR contains automated tests (unit, integration, and/or e2e) to verify the code works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [ ] I have added documentation where necessary
- [ ] I have added a matching `Type: XYZ` label to the PR

---

Fixes #

Issue #5473, opened 2020-11-27
PHP Tests: Separating unit and integration tests


swissspidy:
## Task Description

Might be worth looking into splitting up our current PHP integration tests into proper unit tests (with all of WP mocked) and separate integration tests.

Listing pros and cons, etc.

Related links:
https://github.com/Yoast/wordpress-seo
https://github.com/Yoast/wp-test-utils
https://giuseppe-mazzapica.gitbook.io/brain-monkey/

Some possible pros:

* Easier to extract/port non-WordPress specific code in the future if needed (decoupling)
* Better test performance
* Easier maintenance

PR #8792, opened 2021-08-25
PHP Tests: More robustness, `PluginFactory`, improved Multisite handling

swissspidy:
## Context

* https://github.com/ampproject/amp-toolbox-php/pull/328
* https://github.com/google/web-stories-wp/pull/8792#discussion_r696104419
* #7494

## Summary

This PR adds a single PHPUnit run that is randomly ordered.

This can detect scenarios where the test makes assumptions that are only fulfilled by global state manipulated by a previous test, but are not true in the actual logic to be tested.

In case such a randomly ordered test run fails, the random seed can be found in the GHA log so that this particular ordering can be replicated locally.

## Relevant Technical Choices

**Add a randomly ordered test run to the GHA workflow**

Adds a single PHPUnit run that is randomly ordered.

This can detect scenarios where the test makes assumptions that are only fulfilled by global state manipulated by a previous test, but are not true in the actual logic to be tested.

In case such a randomly ordered test run fails, the random seed can be found in the GHA log so that this particular ordering can be replicated locally.

**Note:** This PR also fixes **some tests** that failed during these randomly ordered test runs, but not **all**. There are simply too many issues with the tests, which requires another big overhaul.

**Adds new `HasSiteSetup ` and `HasSiteRemoval` interfaces**

Eliminates the procedural `setup_new_site()` code by moving logic attached to site setup/teardown in Multisite to the individual services. Basically implementing #7494

Added bonus: rewrite rules are no longer flushed twice upon plugin activation (once by the service container and once by `setup_new_site()`.

Renames existing similar interfaces accordingly:

* `PluginActivationAware`
* `PluginDeactivationAware`

**`PluginFactory`**

Replaces `get_plugin_instance()` with a dedicated `PluginFactory` and a simple `bootstrap_plugin()` function which is hooked to `plugins_loaded`.

Before that, the plugin was executed right away and not attached to any particular action.

**`Link_Controller` / `Hotlink_Controller`**

Fixes a bug in `Link_Controller` which unexpectedly caused an external HTTP request.

Slightly refactored both classes to be more aligned and use `wp_http_validate_url()` for URL validation.

Updates tests accordingly.

## To-do

* Split up integration tests and unit tests (#5473)
* Make use of `DependencyInjectedTestCase`
* Fix all the issues with the tests ⚠️ 

## User-facing changes

N/A

## Testing Instructions

## Checklist

---

Fixes #


tests/phpunit/tests/Admin/ImgAreaSelect_Patch.php
```diff
@@ -28,7 +28,13 @@
  *
  * @coversDefaultClass \Google\Web_Stories\Admin\ImgAreaSelect_Patch
  */
-class ImgAreaSelect_Patch extends Test_Case {
+class ImgAreaSelect_Patch extends TestCase {
+	public function tearDown() {
+		unset( $GLOBALS['current_screen'] );
```
swissspidy: `WP_UnitTestCase` should be doing this already 🤔

tests/phpunit/tests/Integrations/NextGen_Gallery.php
```diff
@@ -18,12 +18,17 @@
...
-use Google\Web_Stories\Tests\Test_Case;
+use Google\Web_Stories\Tests\TestCase;
 
 /**
  * @coversDefaultClass \Google\Web_Stories\Integrations\NextGen_Gallery
  */
-class NextGen_Gallery extends Test_Case {
+class NextGen_Gallery extends TestCase {
+	public function tearDown() {
+		unset( $_SERVER['REQUEST_URI'], $_GET[ Story_Post_Type::POST_TYPE_SLUG ] );
```
swissspidy: Or just `unset( $_SERVER['REQUEST_URI'], $_GET );`?

includes/namespace.php
```diff
@@ -87,11 +87,15 @@ function rewrite_flush() {
  */
 function activate( $network_wide = false ) {
 	$network_wide = (bool) $network_wide;
+
+	// Bootstrap the plugin upon activation.
+	PluginFactory::create()->register();
```
schlessera: The `ServiceBasedPlugin::activate()` should already contain a call to `ServiceBasedPlugin::registerServices()`, so this should not be needed.

Is the problem with the `setup_new_site()` ... ?

tests/phpunit/tests/Admin/Cross_Origin_Isolation.php
```diff
@@ -59,6 +59,12 @@ public function setUp() {
 		$user_preferences->register();
 	}
 
+	public function tearDown() {
+		unset( $GLOBALS['current_screen'] );
+
+		parent::tearDown();
+	}
```
schlessera: As this is a general oversight of WP Core, it might make sense to put it directly in your base `TestCase` (it already is in `DependencyInjectedTestCase`).

Also, see https://core.trac.wordpress.org/ticket/53431

tests/phpunit/tests/Admin/ImgAreaSelect_Patch.php
schlessera: Yes, see https://core.trac.wordpress.org/ticket/53431

tests/phpunit/tests/Block/Web_Stories_Block.php
```diff
@@ -18,15 +18,30 @@
...
+class Web_Stories_Block extends TestCase {
+	/** @var Block */
+	private $instance;
+
+	public function setUp() {
+		parent::setUp();
+
+		if ( WP_Block_Type_Registry::get_instance()->is_registered( 'web-stories/embed' ) ) {
+			unregister_block_type( 'web-stories/embed' );
+		}
```
schlessera: Why is this needed in `setUp()` when it is already in `tearDown()`?

includes/namespace.php
swissspidy: Yeah there was an error when activating the plugin via WP-CLI, see the backtracke here: https://github.com/google/web-stories-wp/runs/3422113347#step:11:715

I think an alternative solution would be to move the `setup_new_site()` call after the `PluginFactory::create()->activate()` call

tests/phpunit/tests/Admin/Cross_Origin_Isolation.php
swissspidy: Aha! _That's why_ I thought that core was already doing this.

I was checking the `WP_UnitTestCase` code locally (from trunk) and saw the code from https://core.trac.wordpress.org/changeset/51419 there, wondering why I still had to do it manually.

tests/phpunit/tests/Block/Web_Stories_Block.php
swissspidy: As you've said yourself, all tests currently simply assume the entire plugin has been loaded and executed

So here I was trying to update the `test_registers_block_type()` test  so that it doesn't simply assume the block is already registered, but instead calls `$embed_block->register()` itself do register it.

I did this because the assertion otherwise failed when tests were executed in random order.

includes/namespace.php
schlessera: Why would activating the plugin "setup a new site"? I don't fully grasp what is happening here and assume this could be refactored into something that "just works"...

I just looked into the `setup_new_site`, and noticed a few things:
- the rewrite rules are being flushed, but this also happens as part of `ServiceBasedPlugin::activate()`, so you seem to be flushing twice on activation
- the rest of the code can be either removed (registration is part of activation), or added as activation events.

So, I think that if you add `Activatable` to `User\Capabilities` and `Database_Upgrader` and call the needed methods, you can remove `setup_new_site()` entirely.

tests/phpunit/tests/Block/Web_Stories_Block.php
swissspidy: IIRC I also tried to make this test use `DependencyInjectedTestCase` when I tried looking into this. So perhaps just some leftover.

includes/namespace.php
schlessera: For the `Database_Upgrader` service, it is just a matter of adapting the `'admin_init'` registration action to something more permissive and then adding more detailed checks as `Conditional`.

schlessera: I also see thet `setup_new_site()` is used 'wp_initialize_site', which explains the function name.

Again, I think most of this can be simplified away via a better use of the services flow.

tests/phpunit/tests/Admin/Dashboard.php
```diff
@@ -38,6 +38,13 @@ public static function wpTearDownAfterClass() {
 		self::delete_user( self::$user_id );
 	}
 
+	public function tearDown() {
+		wp_dequeue_script( \Google\Web_Stories\Admin\Dashboard::SCRIPT_HANDLE );
+		wp_dequeue_style( \Google\Web_Stories\Admin\Dashboard::SCRIPT_HANDLE );
```
schlessera, line 53: Is the `WP_UnitTestCase` not already resetting all enqueues?

includes/namespace.php
swissspidy: Totally makes sense. I don't know why we aren't even using the `Activateable` and `Deactivateable` interfaces at all right now... I'll give it a try.

tests/phpunit/tests/Admin/Dashboard.php
swissspidy: Glancing over it, it doesn't seem so.

includes/namespace.php
swissspidy: Do you think it would make sense to create a new interface for hooking into `wp_initialize_site`?

schlessera: Yes, and I think we already discussed this before somewhere, let me see if I can find that...

schlessera: See https://github.com/google/web-stories-wp/pull/7494

schlessera: Not sure why that was just closed, might that have been by accident? Looks like it was fully on the right track... Maybe it was too difficult to get the existing tests to pass with that change...?

tests/phpunit/tests/Admin/Dashboard.php
swissspidy: But good call, I can add this to `TestCase`

tests/phpunit/tests/Block/Web_Stories_Block.php
swissspidy: I think what isn't clear to me yet is when using `DependencyInjectedTestCase` makes sense and when it doesn't.

includes/namespace.php
swissspidy: Refactored this now and introduced new `Site_Initializable` and `Site_Removable` interfaces (suggestions for better names welcome). Removing `setup_new_site()` entirely.

This seems to have uncovered some broken multisite tests, which is great.

tests/phpunit/tests/Block/Web_Stories_Block.php
schlessera: It depends on _how_ you want to test something.

A proper unit test tests the object under test in isolation. So, for a service, you'd instantiate the service directly, you'd mock/stub/fake its dependencies if any, and you'd BrainMonkey its calls to WP APIs.

A full integration test would make use of `DependencyInjectedTestCase`, where you basically load the full plugin engine and run tests against it. It is the easiest one to set up, but also the slowest to run. And your tests for one service can break because of changes in another service.

In-between, you'd have something like component testing, where you test a single service but with its dependencies (including WP). You'd use `WP_UnitTestCase` for that one. The main advantage of that is speed and stability VS the integration tests.

I'm afraid I can't help more with advice here, we're already starting off with the wrong foot when it comes to testing if we're talking about a WordPress plugin. For proper testing, you'd have all the core business logic in a PHP library, and the WordPress plugin would only be the code that is responsible for the integration between the business logic and the WordPress framework. This is a bit what we have with the `amp-toolbox-php` library. The more we port the pure logic over from the plugin to the library, the more the plugin becomes a pure integration layer, mapping from library inputs/outputs to WordPress hooks. Probably not something you want to evolve to with the web stories plugin...

schlessera: I suggest you use `DependencyInjectedTestCase` mostly for tests that test a certain behavior and don't care too much about which service(s) the logic was executed from.

includes/namespace.php
schlessera: > suggestions for better names welcome

`Multisiteable` and `Multiunsiteable` of course! ;)

Trying to come up with clean names that fit the already existing ones...

How about `HasSiteSetup` and `HasSiteTeardown` ?

tests/phpunit/tests/Block/Web_Stories_Block.php
schlessera: Also note that, whenever you can, you should not use `WP_UnitTestCase` at all and just have a pure PHP `TestCase`, it will be orders of magnitude faster.

Your regular `Test_Case` is now already an extension of `WP_UnitTestCase`, so _all_ tests are slow by default.

includes/namespace.php
swissspidy: I like it!

schlessera:
You've renamed the interfaces, but you've left the old method names. I'd rename these too to better fit the interface names and to be less confusing. Having a method named `remove_site()` sounds like the service is supposed to do the site deletion...

How about `setUpSite()` and `tearDownSite()` to make them fit the interfaces (in whatever casing you prefer, I'll adapt to PSR-12 as needed).

schlessera:
After some more thought (and a shower), I think the naming should be focused on the fact that the service is supposed to _react to an event_. Right now, we have bad naming, like `$service->activate()`, which does not activate the service at all, but rather lets the service act _on_ the plugin activation.

Therefore, I suggest the following interface/method renames:
- `PluginActivationAware::onPluginActivation()`
- `PluginDeactivationAware::onPluginDeactivation()`
- `SiteInitializationAware::onSiteInitialization()`
- `SiteRemovalAware::onSiteRemoval()`

So, `$service->onPluginActivation()` is much clearer. And, for example, turning `$service->removeSite()` into `$service->onSiteRemoval()` also makes it very clear that the service is not responsible for removing a site, but rather for what it itself needs to then _when_ a site is being removed.

The interface names `PluginActivationAware` and similar also read much better and immediately make it clear what is happening.

schlessera:
Btw, sorry for the back and forth, but it so happens that **naming is hard**... 😕

includes/REST_API/Hotlinking_Controller.php
```diff
@@ -86,6 +96,10 @@ public function register_routes() {
 	public function parse_url( $request ) {
 		$url = untrailingslashit( $request['url'] );
 
+		if ( empty( $url ) || ! wp_http_validate_url( $url ) ) {
+			return new WP_Error( 'rest_invalid_url', __( 'Invalid URL', 'web-stories' ), [ 'status' => 404 ] );
+		}
+
```
schlessera: Ist das hier nicht schon von `'validate_callback' => [ $this, 'validate_url' ]` abgedeckt?

includes/AMP_Story_Player_Assets.php
```diff
@@ -37,17 +37,12 @@
...
-	 * @since 1.8.0
+	 * Script handle.
 	 *
-	 * @return string
+	 * @var string
 	 */
-	public function get_handle(): string {
-		return 'standalone-amp-story-player';
-	}
+	const SCRIPT_HANDLE = 'standalone-amp-story-player';
```
spacedmonkey, line 45: Why is this change needed?

includes/Admin/Activation_Notice.php
```diff
@@ -114,8 +114,7 @@ public function enqueue_assets( string $hook_suffix ) {
 		 */
 		unset( $_GET['activate'] ); // phpcs:ignore WordPress.Security.NonceVerification, WordPress.VIP.SuperGlobalInputUsage
 
-		$font_handle = $this->google_fonts->get_handle();
-		$this->assets->enqueue_style( $font_handle );
+		$this->assets->enqueue_style( $this->google_fonts::SCRIPT_HANDLE );
```
spacedmonkey: Didn't we have a problem with this before? Why is this change needed?

includes/Admin/TinyMCE.php
```diff
@@ -153,17 +153,21 @@ public function register_assets() {
...
+	 * @param string|mixed $tag    The `<script>` tag for the enqueued script.
+	 * @param string       $handle The script's registered handle.
+	 * @param string       $src    The script's source URL.
 	 *
-	 * @return string $tag The `<script>` tag for the enqueued script.
+	 * @return string|mixed The filtered script tag.
 	 */
-	public function script_loader_tag( $tag, $handle, $src ): string {
+	public function script_loader_tag( $tag, $handle, $src ) {
+		if ( ! is_string( $tag ) ) {
```
spacedmonkey, line 167: 👍

includes/Database_Upgrader.php
```diff
@@ -91,39 +104,63 @@ class Database_Upgrader extends Service_Base {
 	 * @return void
 	 */
 	public function register() {
-		$version = get_option( self::OPTION, '0.0.0' );
-
-		if ( version_compare( WEBSTORIES_DB_VERSION, $version, '=' ) ) {
-			return;
+		if ( is_admin() ) {
```
spacedmonkey: Why change this from admin_init?

includes/Infrastructure/PluginActivationAware.php
```diff
@@ -29,15 +29,15 @@
...
-	 * Activate the service.
+	 * Act on plugin activation.
 	 *
 	 * @since 1.6.0
 	 *
 	 * @param bool $network_wide Whether the activation was done network-wide.
 	 * @return void
 	 */
-	public function activate( $network_wide );
+	public function on_plugin_activation( $network_wide );
```
spacedmonkey, line 42: Why change the name of this method?

includes/Infrastructure/PluginDeactivationAware.php
```diff
@@ -29,15 +29,15 @@
...
-	 * Deactivate the service.
+	 * Act on plugin deactivation.
 	 *
 	 * @since 1.6.0
 	 *
 	 * @param bool $network_wide Whether the deactivation was done network-wide.
 	 * @return void
 	 */
-	public function deactivate( $network_wide );
+	public function on_plugin_deactivation( $network_wide );
```
spacedmonkey, line 42: Why change the name of this method?

includes/namespace.php
```diff
@@ -30,48 +30,6 @@
...
-}
-
-/**
- * Flush rewrites.
- *
- * @since 1.7.0
- *
- * @return void
- */
-function rewrite_flush() {
```
spacedmonkey, line 69: Why remove this function, it stop lots of repeated code with `phpcs:ignore`

includes/namespace.php
```diff
@@ -297,19 +248,14 @@ function rest_preload_api_request( $memo, $path ): array {
...
 /**
- * Web stories Plugin Instance
+ * Bootstrap the plugin.
  *
- * @return Plugin
+ * @since 1.11.0
+ *
+ * @return void
  */
-function get_plugin_instance() {
```
spacedmonkey, line 304: Having a plugin instance in global was by design. Some other plugins could use this class in their own plugins and extend functionality from it. This maybe a breaking change for some.

includes/namespace.php
```diff
@@ -251,16 +198,20 @@ function rest_preload_api_request( $memo, $path ): array {
 	}
 
 	$method = 'GET';
-	if ( is_array( $path ) && 2 === count( $path ) ) {
+	if ( is_array( $path ) ) {
+		if ( 2 !== count( $path ) ) {
+			return $memo;
+		}
+
```
spacedmonkey, line 205: Why is this change needed?

tests/phpunit/includes/Test_REST_TestCase.php
```diff
@@ -49,7 +52,7 @@ abstract class Test_REST_TestCase extends Test_Case {
 	 */
 	protected function assertErrorResponse( $code, $response, $status = null ) {
 
-		if ( is_a( $response, 'WP_REST_Response' ) ) {
+		if ( $response instanceof WP_REST_Response ) {
```
spacedmonkey, line 55: This line is copied from core.

tests/phpunit/includes/TestCase.php
```diff
@@ -0,0 +1,57 @@
...
+
+use Google\Web_Stories\Services;
+use Yoast\WPTestUtils\WPIntegration\TestCase as PolyfilledTestCase;
+
+/**
+ * Class TestCase
+ *
+ * @package Google\Web_Stories\Tests
+ */
+abstract class TestCase extends PolyfilledTestCase {
```
spacedmonkey, line 23: Why change the name of the class? The naming convention, is underscores in WP / PHP.

tests/phpunit/tests/Admin/Editor.php
```diff
@@ -86,7 +86,6 @@ public function setUp() {
 
 	public function tearDown() {
 		$this->set_permalink_structure( '' );
-		$_SERVER['REQUEST_URI'] = '';
```
spacedmonkey, line 89: ?

tests/phpunit/tests/REST_API/Hotlinking_Controller.php
```diff
@@ -158,18 +165,6 @@ public function test_register_routes() {
 		$this->assertCount( 1, $routes[ self::REST_URL ] );
 	}
 
-	/**
```
spacedmonkey, line 161: Why remove this test?

includes/Infrastructure/PluginActivationAware.php
swissspidy: See https://github.com/google/web-stories-wp/pull/8792#issuecomment-906202092 and the PR description for explanation of why the interfaces and their method names have been renamed

includes/Infrastructure/PluginDeactivationAware.php
swissspidy: See https://github.com/google/web-stories-wp/pull/8792#issuecomment-906202092 and the PR description for explanation of why the interfaces and their method names have been renamed

includes/namespace.php
swissspidy: IMO this repetition does not warrant its own function just to save a couple of lines.

includes/namespace.php
swissspidy: PHPStan and IntelliJ complained about cases where `$path` was an array with only 1 element or more than 2 elements, in which case an array could potentially be passed to `wp_parse_url()` below.

tests/phpunit/includes/Test_REST_TestCase.php
swissspidy: And the core code is incorrect there :-)

tests/phpunit/includes/TestCase.php
swissspidy: When I copied over `DependencyInjectedTestCase` I figured this would make the names for test case classes more consistent. Happy to change back, I'm really indifferent on that one.

FWIW, we're not really consistent with the naming, but I also don't want us to necessarily follow WordPress naming conventions. On that part, we should probably adopt a PSR standard if anything.

tests/phpunit/tests/Admin/Editor.php
swissspidy: This cleanup is moved to cleanup in `TestCase` to avoid repetition.

includes/Database_Upgrader.php
spacedmonkey: Or make it is conditional service

tests/phpunit/tests/REST_API/Hotlinking_Controller.php
swissspidy: Because I removed the method. Schema information is now directly in the `register_routes()` call like we do in `Link_Controller` as well.

includes/namespace.php
swissspidy: Yeah you're right. I'm gonna add it back as a wrapper around the`PluginDirectory`.

includes/Database_Upgrader.php
swissspidy: `Database_Upgrader` is now not a `Delayed` service anymore, which means it will get registered right away and we can use it to run upgrade routines during site creation, removal and plugin activation/deactivation without being in admin context. (But of course it also runs in the admin if you visit it).

This was needed to make the removal of `setup_new_site` / the addition of `HasSiteSetup`/`HasSiteRemoval` work.

includes/AMP_Story_Player_Assets.php
swissspidy: Just some cleanup to make the code more consistent with the rest. We don't need a method for this handle as we don't need to pass instances of this class around. This simplifies this quite a bit.

includes/Admin/Activation_Notice.php
swissspidy: Refactored this because I was removing the `get_handle()` methods (see https://github.com/google/web-stories-wp/pull/8792#discussion_r696720274).

Actually, I can refactor this class more so we don't have to pass a `Google_Fonts` instance. Making this even simpler.

includes/AMP_Story_Player_Assets.php
spacedmonkey: The dependency between these classes is by design. Meaning, we know that service has been registered, so we know that script has been registered.

includes/Admin/Activation_Notice.php
spacedmonkey: I have never liked using class consts in that way. It doesn't show when code depends on another class.

includes/Database_Upgrader.php
spacedmonkey: It extends `Service_Base` which is delayed, on init.

includes/namespace.php
spacedmonkey: I don't understand why remove it? Just saves lots of `phpcs:ignore`

includes/Infrastructure/ServiceBasedPlugin.php
```diff
@@ -112,43 +112,101 @@ public function __construct(
...
+
+	/**
+	 * Act on site initialization on Multisite.
+	 *
+	 * @since 1.11.0
+	 *
+	 * @param WP_Site $site The site being initialized.
+	 * @return void
+	 */
+	public function on_site_initialization( $site ) {
```
spacedmonkey:
```suggestion
	public function on_site_initialization( WP_Site $site ) {
```

includes/Infrastructure/ServiceBasedPlugin.php
```diff
@@ -112,43 +112,101 @@ public function __construct(
...
+
+	/**
+	 * Act on site removal on Multisite.
+	 *
+	 * @since 1.11.0
+	 *
+	 * @param WP_Site $site The site being removed.
+	 * @return void
+	 */
+	public function on_site_removal( $site ) {
```
spacedmonkey:
```suggestion
	public function on_site_removal( WP_Site $site ) {
```

includes/Plugin.php
```diff
@@ -145,26 +145,6 @@ protected function get_service_classes(): array {
...
-	 * construct, it can be a `class` as well.
-	 *
-	 * Whenever you ask the injector to "make()" an <interface>, it will resolve
-	 * these mappings and return an instance of the final <class> it found.
-	 *
-	 * @since 1.6.0
-	 *
-	 * @return array<string> Associative array of fully qualified class names.
-	 */
-	protected function get_bindings(): array {
```
spacedmonkey, line 164: Why remove this?

includes/Plugin.php
```diff
@@ -214,33 +193,4 @@ protected function get_delegations(): array {
 			},
 		];
 	}
-
-	/**
```
spacedmonkey, line 218: This is for backward compatibility, why was this removed?

includes/REST_API/Hotlinking_Controller.php
```diff
@@ -294,24 +303,4 @@ public function validate_url( $value ) {
...
 	}
-
-	/**
-	 * Retrieves the query params for the posts collection.
-	 *
-	 * @since 1.11.0
-	 *
-	 * @return array Collection parameters.
-	 */
-	public function get_collection_params(): array {
```
spacedmonkey, line 311: Don't understand why this was removed? I overroad this method, by design. See the original method. 

https://github.com/WordPress/wordpress-develop/blob/793bde1af74ad24c407d3645c788ac49ea352fa6/src/wp-includes/rest-api/endpoints/class-wp-rest-controller.php#L333-L367

We don't support page, per_page and search, which is why I overroad it.

spacedmonkey: We should do this in all our customizer controller.

includes/PluginFactory.php
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * It can decide whether to return a shared or a fresh instance as needed.
+ *
+ * To read more about why this is preferable to a Singleton,
+ *
+ * @see https://www.alainschlesser.com/singletons-shared-instances/
+ *
+ * @since 1.11.0
+ */
+class PluginFactory extends ServiceBasedPlugin {
```
spacedmonkey, line 45: I have never understood the point of this class. A static method in a class seems a little pointless.

includes/Plugin.php
swissspidy: Because it's identical to the parent `\Google\Web_Stories\Infrastructure\ServiceBasedPlugin::get_bindings()` method and thus can be safely removed.

includes/Plugin.php
swissspidy: Don't need to keep BC stuff forever.

Plus, I couldn't find a single plugin or theme using any of these.

includes/REST_API/Hotlinking_Controller.php
swissspidy: `get_collection_params()` was only used in the `register_rest_route()` call further above and nowhere else. I am inlining the code instead of having this extra method. Functionality-wise nothing changes.

includes/namespace.php
swissspidy: I actually updated the PHPCS config because those PHPCS warnings are really false positives, so all those `phpcs:ignore` comments are not needed anymore.

includes/Database_Upgrader.php
swissspidy: Ah, you're right. But `init` is fine, as all our code is running there. But `admin_init` was too late. In short, this change allows us to have all the nice things.

includes/AMP_Story_Player_Assets.php
swissspidy: Even when passing instances around one doesn't need `get_handle()` method and can instead just call `$instance::SCRIPT_HANDLE`.

includes/PluginFactory.php
swissspidy: One could look at it as the OOO version of the `get_plugin_instance()` function we initially had.

includes/Infrastructure/ServiceBasedPlugin.php
```diff
@@ -112,43 +115,101 @@ public function __construct(
...
+
+	/**
+	 * Act on site initialization on Multisite.
+	 *
+	 * @since 1.11.0
+	 *
+	 * @param WP_Site $site The site being initialized.
+	 * @return void
+	 */
+	public function on_site_initialization( WP_Site $site ) {
```
swissspidy, line 169: Hmm since I've put the `WP_Site` type declaration here, the PHPUnit tests are failing:

```
Cannot mock Google\Web_Stories\Infrastructure\ServiceBasedPlugin::on_site_initialization() because a class or interface used in the signature is not loaded
```

This makes sense since it's a Multisite-only class that is not available otherwise.

The question now is how I could fix this.

Declare the class?

Mock the class?

Remove the type declaration again?

swissspidy: @schlessera maybe you have an idea here?

schlessera: You can mock it as a dependency, or you can create a file with conditional stubs that is pulled in via PHPUnit bootstrap.

The file would contain code like this:
```php
if ( ! class_exists( 'WP_Site' ) ) {
	class WP_Site {}
}
```

schlessera: I think the mocking would still require a stub , so you should probably go with the stubs here.

swissspidy: Hmm I thought as much. Works like a charm this way! Thanks!

includes/PluginFactory.php
schlessera: The main benefit is autoloading. PHP is heavily optimized towards objects by now, and anything that is not an object is penalized both in functionality and performance.

includes/Database_Upgrader.php
spacedmonkey: An improvement idea - https://github.com/google/web-stories-wp/pull/8949

includes/REST_API/Hotlinking_Controller.php
```diff
@@ -284,7 +293,7 @@ public function parse_url_permissions_check() {
 	public function validate_url( $value ) {
 		$url = untrailingslashit( $value );
 
-		if ( empty( $url ) ) {
+		if ( empty( $url ) || ! wp_http_validate_url( $url ) ) {
```
spacedmonkey, line 296: Are checks for host / path still needed?

swissspidy: The host one probably not anymore. But I'll check.

spacedmonkey:
This is the last thing - https://github.com/google/web-stories-wp/pull/8792#discussion_r701941575 then I will approve.

includes/REST_API/Hotlinking_Controller.php
swissspidy: Updated now

spacedmonkey (approved):
👍

PR #8804, opened 2021-08-26
Bump PHP requirement to 7.0.21

swissspidy:
Specifying a version higher than 7.0.0 enables us to use more up-to-date versions of some sub-dependencies, since 7.0.0 is otherwise often not supported due to security issues.

The majority of our PHP 7.0 users are on 7.0.22+ anyway, so this is a safe bump.


spacedmonkey:
Why 7.0.21 and not 7.0.9?

swissspidy:
The higher the better? :-)

Alain mentioned that 7.0.8+ is what makes the most difference, afterwards dependencies don't really change. But according to telemetry all users are on 7.0.22+ anyway, so we might just as well bump to 7.0.21+

spacedmonkey (approved):

PR #8806, opened 2021-08-26
Media hotlinking: use the hotlink endpoint

### LINKED ISSUE #8773: Media Hotlinking: use the `hotlink` endpoint for server-side validation
## Feature Description
In #8731, a new endpoint is added for validating a media URL and getting useful data about it (mime type, extension, etc.).

We should now update the client-side logic to actually use this endpoint as well.

Example request: `await wp.apiFetch({path: 'web-stories/v1/hotlink?url=https://www.learningcontainer.com/wp-content/uploads/2020/05/sample-webm-file.webm'})`

miina:
## Summary
Uses the new `hotlink` endpoint to get server-side validation for the inserted URL.

## Testing Instructions
1. Open the Insert by Link dialogue.
2. Insert an Unsplash image link.
3. Verify the image is inserted correctly.
4. Insert Coverr video link
5. Verify the video is inserted correctly as a new element.
6. Try also with incorrect URLs (including .svg as that's not allowed) and verify it shows an appropriate message.

## Checklist

---

Fixes #8773


packages/story-editor/src/components/library/panes/media/local/hotlink/useInsert.js
```diff
@@ -36,83 +35,106 @@ import {
...
-  gif: 'image/gif',
-  jpe: 'image/jpe',
-  jpeg: 'image/jpg',
-  m4v: 'video/mp4',
-  mp4: 'video/mp4',
-  png: 'image/png',
-  webm: 'video/webm',
-  webp: 'image/webp',
-};
+function getErrorMessage(code, description) {
```
miina, line 37: @swissspidy @spacedmonkey FYI This is the PR implementing the `hotlink` endpoint on the client side.

Everything seems to be covered and working as expected as far as I'm testing. 👍

packages/story-editor/src/components/library/panes/media/local/hotlink/useInsert.js
```diff
@@ -36,83 +35,106 @@ import {
...
-      onClose();
-    } catch (e) {
-      setErrorMsg(insertionError);
+  const insertMedia = useCallback(async () => {
+    const { type, mime_type: mimeType } = hotlinkData;
+    const resource = await getResourceFromUrl(link, type);
+    resource.mimeType = mimeType;
+
+    if ('video' === type && hasUploadMediaAction) {
+      const originalFileName = getFileNameFromUrl(link);
```
spacedmonkey: Instead of doing this, let's use the filename that come from `hotlinkData`.

packages/story-editor/src/components/library/panes/media/local/hotlink/hotlinkModal.js
```diff
@@ -57,7 +56,6 @@ function HotlinkModal({ isOpen, onClose }) {
     return () => clearTimeout(timeout);
   }, [isOpen, inputRef]);
 
-  // @todo We're not really uploading anything here, so should we have a fixed list instead?
   let description = __('No file types are currently supported.', 'web-stories');
   if (allowedFileTypes.length) {
```
swissspidy, line 60: Don't need `allowedFileTypes` here anymore at all I suppose?

miina: This is for the initial description to inform the user which types can be used:
Screenshot 2021-08-26 at 15 38 51

Are you thinking we should replace this with something else?

packages/story-editor/src/components/library/panes/media/local/hotlink/useInsert.js
spacedmonkey: You could even do this. 

```
 const originalFileName = fileName.replace('."+ext, '');
```

packages/story-editor/src/components/library/panes/media/local/hotlink/hotlinkModal.js
swissspidy: Ah, missed this. nvm then. Should be fine.

packages/story-editor/src/components/library/panes/media/local/hotlink/hotlinkModal.js
```diff
@@ -20,20 +20,19 @@ import { __, sprintf, translateToExclusiveList } from '@web-stories-wp/i18n';
 import { Input } from '@web-stories-wp/design-system';
 import {
   useState,
-  useCallback,
   useRef,
   useLayoutEffect,
+  useCallback,
```
barklund, line 25: Much better order 😂.

packages/story-editor/src/components/library/panes/media/local/hotlink/useInsert.js
```diff
@@ -98,21 +93,54 @@ function useInsert({
...
     onClose,
     setErrorMsg,
     setLink,
-    hasUploadMediaAction,
     uploadVideoPoster,
   ]);
+
+  useEffect(() => {
+    if (hotlinkData) {
+      insertMedia();
```
barklund: Hm, couldn't this be directly invoked rather than going through `setHotlinkData` and a `useEffect`? Or is there a new for a new render cycle?

barklund (approved):
Works for me, but would like 👍  from infra as well.

packages/story-editor/src/components/library/panes/media/local/hotlink/useInsert.js
```diff
@@ -98,21 +93,54 @@ function useInsert({
...
+
+  const onInsert = useCallback(() => {
+    if (!isValidUrl(link)) {
+      setErrorMsg(__('Invalid link.', 'web-stories'));
+      return;
+    }
+    getHotlinkInfo(link)
+      .then((hotlinkInfo) => {
+        setHotlinkData(hotlinkInfo);
+      })
```
spacedmonkey:
```suggestion
      .then(setHotlinkData)
```

packages/story-editor/src/components/library/panes/media/local/hotlink/useInsert.js
miina: Hm, I am now actually getting an unexpected answer from the REST API when testing with Coverr videos, for example:
`await wp.apiFetch({path: 'web-stories/v1/hotlink?url=https://storage.coverr.co/videos/ba21g02uIih01OZfcJBEaRoee9Pr231LnK/preview?token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhcHBJZCI6IjEyNjFDMkM5QUYxNEY5NkJDRTc2IiwiaWF0IjoxNjMwMDUxMjA4fQ.skZ_fN2Q2zXUPKVOSAnv7Z5cRDPOInWdV0qoWQ0aDe4'})`

So it looks like it might need some debugging from the endpoint.

Screenshot 2021-08-27 at 11 13 47

swissspidy: Two issues here:

First, you'll have to encode the Coverr URL with `encodeURIComponent()`, otherwise that `?` after `preview` can be misinterpreted to be a query param for the `path` you're providing.

```js
await wp.apiFetch({path: 'web-stories/v1/hotlink?url=' + encodeURIComponent('https://storage.coverr.co/videos/ba21g02uIih01OZfcJBEaRoee9Pr231LnK/preview?token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhcHBJZCI6IjEyNjFDMkM5QUYxNEY5NkJDRTc2IiwiaWF0IjoxNjMwMDUxMjA4fQ.skZ_fN2Q2zXUPKVOSAnv7Z5cRDPOInWdV0qoWQ0aDe4')})
```

Then the URL can be properly parsed.

Second, we're only performing `HEAD` requests and WordPress apparently does not follow redirects for `HEAD` requests by default.

Fo fix this, change the following line:

https://github.com/google/web-stories-wp/blob/c1e8e860f7563ff15fa5bd5bc4825c2eb3720227/includes/REST_API/Hotlinking_Controller.php#L107

To this:

```php
$response = wp_safe_remote_head(
	$url,
	[
		/** This filter is documented in wp-includes/class-http.php */
		'redirection' => apply_filters( 'http_request_redirection_count', 5, $url ),
	]
);
```

miina: Ah, I actually tried the request with `encodeURIComponent()` as well, forgot to mention. Will add this to always use encoding.

Thanks for looking into this -> testing the changes now.

miina: Works now with the change to `wp_safe_remote_head` 👍

swissspidy (approved):

Issue #8793, opened 2021-08-25
E2E Tests: flakey screenshots due to carousel page previews


spacedmonkey:
## Bug Description

Sometimes the screenshot is taken, before the pages of the story is preview, generating a different screenshot sometimes, see screenshots. We should wait until these pages load before taking screenshot. 

[Example](https://percy.io/google/web-stories-wp-e2e/builds/12110659/changed/682131647?browser=chrome&browser_ids=14%2C16&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=original&width=1600&widths=1600). 

## Screenshots

![885fd2583a3763c06e5cea0b80e96324a51fc06e8492f2595d4714251e80c24d](https://user-images.githubusercontent.com/237508/130794884-e97814c8-cdd1-444c-86cc-2c0cf01ecdfc.png)
![cf74e46f9edd2b3ab0368f6ee517c4cf348dce9a0b142e7ebef2e79232a8fc70](https://user-images.githubusercontent.com/237508/130794891-45d6f677-4849-4882-b2b6-7c3f800f0ecf.png)

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

PR #8813, opened 2021-08-26
Carousel: Cache previews as static images

### LINKED ISSUE #7987: Image rendering for thumbnails
## Feature Description
Render images for thumbnails instead of rendering all the elements one by one every time.

See the [discovery PR](https://github.com/google/web-stories-wp/pull/7887) that uses `html-to-image` for creating images from the thumbnails on initial render and uses those.

The current page preview should still be rendered normally since the changes on the page should be reflected right away.
This also means that after a page update, its image should be updated, too.

## Alternatives Considered
SVG rendering, see https://github.com/google/web-stories-wp/issues/7468

SVG rendering didn't seem to bring a significant performance improvement, see https://github.com/google/web-stories-wp/pull/7615#issuecomment-853368302

## Additional Context

---

barklund:
## Context

This PR caches thumbnails as images when not active. The difference is very subtle but seems quite effective for performance reasons.

## User-facing changes

| | Screenshot |
|-|-|
| Before | Screen Shot 2021-08-26 at 10 27 35 |
| After | Screen Shot 2021-08-26 at 10 27 06 |

### QA

1. Have a multi-page story
2. Observe that the non-current page thumbnail is an image rather than a HTML fragment

## Checklist

---

Fixes #7987


packages/story-editor/src/components/carousel/pagepreview/index.js
```diff
@@ -78,23 +85,56 @@ const PreviewWrapper = styled.div`
...
+            {hasImage ? (
+              <Image src={imageBlob} alt={label} />
+            ) : (
+              page.elements.map(({ id, ...rest }) => (
+                <DisplayElement
+                  key={id}
+                  previewMode
+                  element={{ id, ...rest }}
+                  page={page}
+                />
```
swissspidy: IIRC destructuring and assembling the props like that is a pattern we're trying to avoid.

```suggestion
              page.elements.map((element) => (
                <DisplayElement
                  key={element.id}
                  previewMode
                  element={element}
                  page={page}
                />
```

packages/story-editor/src/components/carousel/pagepreview/index.js
```diff
@@ -78,23 +85,56 @@ const PreviewWrapper = styled.div`
...
+  const setPageRef = useCallback((node) => node && setPageNode(node), []);
+  const hasImage = !isActive && imageBlob;
+  const pageAtGenerationTime = useRef();
+
+  useEffect(() => {
+    // If this is not the active page, there is a page node and we haven't
+    // snapshotted the current page
+    if (!isActive && pageNode && pageAtGenerationTime.current !== page) {
+      // Schedule an idle callback to actually generate the image
+      requestIdleCallback(
```
swissspidy: Import from `utils/idleCallback`?

packages/story-editor/src/components/carousel/pagepreview/index.js
```diff
@@ -78,23 +85,56 @@ const PreviewWrapper = styled.div`
...
-            {page.elements.map(({ id, ...rest }) => (
-              <DisplayElement
-                key={id}
-                previewMode
-                element={{ id, ...rest }}
-                page={page}
-              />
-            ))}
+            {hasImage ? (
+              <Image src={imageBlob} alt={label} />
```
swissspidy: Can we add `width` and `height`?

spacedmonkey:
Wondering if this will affect - https://github.com/google/web-stories-wp/issues/8793

miina:
@barklund 
From testing:
It looks like the thumbnails aren't currently updated when the page changes and displays the original Page thumbnail for any non-current page -- see the blank thumbnail on the right which is not actually a blank page:
thumbs

To repro:
1. Add a few pages with some elements and save changes.
2. Reload the page.
3. Select a page and make some changes.
4. Now select another page and see the thumbnail of the previously changed page -- it still shows the original thumbnail (and continues to do so)

merapi:
For me, the first thumbnail is broken:

https://user-images.githubusercontent.com/21036224/131327845-5c09e445-2cbb-4043-a15a-35b5a6fea1f4.mov

barklund:
@merapi, @miina, thanks for testing - you both found the same issue!

> For me, the first thumbnail is broken:

> From testing:
> It looks like the thumbnails aren't currently updated when the page changes and displays the original Page thumbnail for any non-current page -- see the blank thumbnail on the right which is not actually a blank page:

Both stem from the fact, that once an image had been generated, then when the page was later updated, the old image was still displayed, so when the new snapshot was taken, it was a snapshot of the old image rather than the actual page below, which is of course very wrong. So, I just had to hide the image snapshot once the page had updated (by unsetting the blob) and everything seems to be working correctly for both new and old stories.

Let me know if you still have issues?

packages/story-editor/src/components/carousel/pagepreview/index.js
barklund: Fixed in 65262304cd

packages/story-editor/src/components/carousel/pagepreview/index.js
barklund: Fixed in 65262304cd

packages/story-editor/src/components/carousel/pagepreview/index.js
barklund: Fixed in 65262304cd

miina (approved):
🕺

PR #8816, opened 2021-08-26
Canvas: fix disabling nav buttons if there's no page

swissspidy:
## Context

The next/previous navigation buttons next to the canvas ought to be disabled if there's no page to navigate to.

However, this was somehow forgotten during some refactoring, passing some non-existent `isDisabled` and `isHidden` props.

## Summary

This PR addresses this by disabling the buttons using `aria-disabled`, which means they can still receive focus and can get tabbed to, to ensure accessibility.

We probably need to do this in many other places too, but this is a start.

## Relevant Technical Choices

Updates the button component from the design system so it can 

## User-facing changes

Before:

Screenshot 2021-08-26 at 16 21 46

After:

Screenshot 2021-08-26 at 17 41 18
Screenshot 2021-08-26 at 17 41 14
Screenshot 2021-08-26 at 17 41 10

## Testing Instructions

1. See that next/prev buttons are disabled on the canvas when there's no page to navigate to

No

No

No

## Checklist

---

Fixes #

### DIFF
--- a/packages/design-system/src/components/button/button.js
+++ b/packages/design-system/src/components/button/button.js
@@ -69,7 +69,8 @@ const Base = styled.button(
       }
     `}
 
-    &:disabled {
+    &:disabled,
+    &[aria-disabled="true"] {
       pointer-events: none;
       background-color: ${theme.colors.interactiveBg.disable};
       color: ${theme.colors.fg.disable};
@@ -120,7 +121,8 @@ const tertiaryColors = ({ theme }) => css`
     background-color: ${theme.colors.interactiveBg.tertiaryHover};
   }
 
-  &:disabled {
+  &:disabled,
+  &[aria-disabled="true"] {
     background-color: ${theme.colors.interactiveBg.tertiaryNormal};
     &:hover,
     &:focus {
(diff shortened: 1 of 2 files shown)


packages/design-system/src/components/button/button.js
```diff
@@ -120,7 +121,8 @@ const tertiaryColors = ({ theme }) => css`
     background-color: ${theme.colors.interactiveBg.tertiaryHover};
   }
 
-  &:disabled {
+  &:disabled,
+  &[aria-disabled="true"] {
```
swissspidy:
```suggestion
  &[aria-disabled='true'] {
```

packages/design-system/src/components/button/button.js
```diff
@@ -225,7 +228,8 @@ const ButtonLink = styled(Base)`
       color: ${theme.colors.fg.linkHover};
     }
     &:active,
-    &:disabled {
+    &:disabled,
+    &[aria-disabled="true"] {
```
swissspidy:
```suggestion
    &[aria-disabled='true'] {
```

packages/design-system/src/components/button/button.js
```diff
@@ -155,7 +157,8 @@ const quaternaryColors = ({ theme }) => css`
       border-color: ${theme.colors.border.defaultPress};
     `}
 
-  &:disabled {
+  &:disabled,
+  &[aria-disabled="true"] {
```
swissspidy:
```suggestion
  &[aria-disabled='true'] {
```

BrittanyIRL (approved):
thank you!

csossi:
Verified in QA

PR #8824, opened 2021-08-27
Right Click Menu: Make shortcuts readable by assistive technologies

### LINKED ISSUE #8552: Context Menu: shortcuts not being read by voiceover
## Bug Description

Context menu shortcuts are not being read when using voiceover. Wondering if other assistive technologies will have the same problem.

## Expected Behaviour

Voiceover should read the context menu button with the text and the shortcut text.

## Steps to Reproduce

1. If testing on mac, use`Cmd` + `F5` to open voice over.
2. Right click an image or the canvas
3. Try to move around the menu. Should not hear any shortcuts announced.

## Screenshots

short-cutties

## Additional Context

- Plugin Version: 1.10.0-alpha.0+8ac805e
- WordPress Version: 5.8
- Operating System: MacOSX Big Sur 11.5.5
- Browser: Chrome

---

samwhale:
## Context

Right click menu shortcuts were not being read by assistive tech.

## Summary

Add visually hidden text to the context menu that can be read by screen readers.

## Relevant Technical Choices

The [`<kbd />`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd) element represents a span of inline text denoting textual user input from a keyboard, voice input, or any other text entry device.

However, assistive technologies like JAWS and NVDA only read the text content between the kbd tags and ignore punctuation.

[This github comment](https://github.com/WordPress/gutenberg/issues/15655#issuecomment-680819949) talks about the same issue and they suggest adding visually hidden text.

Thank you to @BrookeGraham for finding this!

## To-do

n/a

## User-facing changes

No visual changes.

voiceover-yay

## Testing Instructions

1. Use a screen reader if you have one! Or use [chromevox](https://chrome.google.com/webstore/detail/screen-reader/kgejglhpjiefppelpmljglcjbhoiplfn) or voiceover (if on mac). 
2. Open the right click menu by right clicking on an element in the canvas in the editor
3. Read through the options. Options with shortcuts.

no

no

no

## Checklist

---

Fixes #8552


swissspidy:
How is the keyboard shortcuts menu built in the Gutenberg block editor? Does it also use this approach? Curious to see how well that works with assistive technologies.

maxyinger (approved):
Makes sense. Great find and investigative work here!

maxyinger:
> How is the keyboard shortcuts menu built in the Gutenberg block editor? Does it also use this approach? Curious to see how well that works with assistive technologies.

Looks like it also uses the `aria-label` approach

Screen Shot 2021-08-27 at 1 37 42 PM

packages/design-system/src/components/contextMenu/menuItem.js
```diff
@@ -104,13 +105,19 @@ export const MenuItem = ({
...
+              disabled={disabled}
+              size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.X_SMALL}
+              forwardedAs="kbd"
+              aria-hidden
+            >
+              {shortcut.display}
+            </Shortcut>
+            <VisuallyHidden>
+              {shortcut?.title || shortcut.display}
+            </VisuallyHidden>
```
BrittanyIRL: 🔥

packages/story-editor/src/app/rightClickMenu/constants.js
```diff
@@ -65,63 +62,94 @@ const StyledKbd = styled(Text).attrs({
...
-    <>
-      <StyledKbd aria-label={optionOrAlt.title}>{optionOrAlt.symbol}</StyledKbd>{' '}
-      <StyledKbd aria-label={cmdOrCtrl.title}>{cmdOrCtrl.symbol}</StyledKbd>{' '}
-      <StyledKbd>{'P'}</StyledKbd>
-    </>
-  ),
+  SEND_BACKWARD: {
+    display: <StyledKbd>{prettifyShortcut('mod down')}</StyledKbd>,
+    title: sprintf(
+      /* Translators: 1: Command/Carat keyboard key.*/
```
BrittanyIRL:
```suggestion
      /* Translators: 1: Command/Caret keyboard key.*/
```

packages/story-editor/src/app/rightClickMenu/constants.js
```diff
@@ -65,63 +62,94 @@ const StyledKbd = styled(Text).attrs({
...
+      _x(
+        '%s down',
+        'The keyboard keys "Command" or "Carat" and "down"',
```
BrittanyIRL:
```suggestion
        'The keyboard keys "Command" or "Caret" and "down"',
```

packages/story-editor/src/app/rightClickMenu/constants.js
```diff
@@ -65,63 +62,94 @@ const StyledKbd = styled(Text).attrs({
...
+      getKeyForOS('mod')
+    ),
+  },
+  SEND_TO_BACK: {
+    display: <StyledKbd>{prettifyShortcut('mod shift down')}</StyledKbd>,
+    title: sprintf(
+      /* Translators: 1: Command/Carat keyboard key.*/
+      _x(
+        '%s Shift down',
+        'The keyboard keys "Command" or "Carat", "Shift", and "down"',
```
BrittanyIRL:
```suggestion
        'The keyboard keys "Command" or "Caret", "Shift", and "down"',
```

😎  `Carat` is the proportion of pure gold, like 24k.

BrittanyIRL (approved):
Great find on this, so glad this is all cleaned up now. 

One small nit - it's `caret` not `carat`

swissspidy:
> > How is the keyboard shortcuts menu built in the Gutenberg block editor? Does it also use this approach? Curious to see how well that works with assistive technologies.
> 
> Looks like it also uses the `aria-label` approach
> 
> Screen Shot 2021-08-27 at 1 37 42 PM

One key difference I see is that Gutenberg doesn't have translatable strings like `%s Shift up` or `%1$s %2$s O` like this PR has, but instead assembles the `aria-label` dynamically:

https://github.com/WordPress/gutenberg/blob/3da717b8d0ac7d7821fc6d0475695ccf3ae2829f/packages/edit-post/src/components/keyboard-shortcut-help-modal/shortcut.js

https://github.com/WordPress/gutenberg/blob/3da717b8d0ac7d7821fc6d0475695ccf3ae2829f/packages/keycodes/src/index.js#L240-L277

Can we do that too? I find strings like `%1$s %2$s O` very confusing to translate and assembling the list dynamically more fitting in this case.

Two things I'd like to point out regarding that `shortcutAriaLabel` example:

1. For non-Apple devices the shortcut list is concatenated with a `+` in-between
2. Strings like `Command` and `Shift` must be translatable

samwhale:
> Can we do that too? I find strings like `%1$s %2$s O` very confusing to translate and assembling the list dynamically more fitting in this case.
> 
> Two things I'd like to point out regarding that `shortcutAriaLabel` example:
> 
> 1. For non-Apple devices the shortcut list is concatenated with a `+` in-between
> 2. Strings like `Command` and `Shift` must be translatable

I'll take a look at this today 👍

samwhale:
Updated so that the shortcuts are created dynamically. There is an `aria-label` on each `kbd` but the `VisuallyHidden` title is still being read instead:
update-shortcuts

packages/design-system/src/components/contextMenu/menuItem.js
```diff
@@ -143,7 +139,6 @@ export const MenuItem = ({
     return (
       <Button
         ref={itemRef}
-        aria-label={itemLabel}
```
BrittanyIRL: Given the set up of context menus that don't have keyboard shortcuts,  could we keep the aria-labels for links and buttons that have no shortcuts?

samwhale: yup, I'll add that back in 👍 good catch

packages/design-system/src/components/contextMenu/menuItem.js
```diff
@@ -95,26 +86,35 @@ export const MenuItem = ({
         </Tooltip>
       );
     }
+
+    /* Shortcut title to be read by screen reader. */
+    const visuallyHiddenContent = shortcut?.title ? (
+      <VisuallyHidden>{shortcut?.title}</VisuallyHidden>
```
BrookeGraham: Just a nit, isn't this null check redundant since we have already checked that `shortcut.title`  exists on line 91?

PR #8848, opened 2021-08-30
Keep overlay after removing media as background

### LINKED ISSUE #7893: Filter applied to background media is not preserved when the media is removed from the background
## Bug Description
Originally reported by @embarks in Bug Bash 1.8

After editing a filter’s color and opacity on background media and then removing that image from the background, the filter is removed from the media and does not carry over to the now foreground image.

Does not occur in the other direction: can edit a filter and make the media a background image, the filter is not removed

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

---

miina:
## Summary
Keeps overlay (Filter) after removing media as background.

## Testing Instructions
1. Add media as background.
2. Add Filter to the media
3. Remove media from background
4. Verify the filter is still applied.

1.
## Checklist

---

Fixes #7893


barklund (approved):
🦘

PR #8853, opened 2021-08-30
Telemetry: Fix `trackError` usage

swissspidy:
`trackError` can be used to send [exception events](https://developers.google.com/analytics/devguides/collection/ga4/exceptions) to measure the number and type of crashes or errors that occur on a web page.

The first parameter is the error description, the second one indicates whether the error was fatal or not.

```js
trackError('Division by zero', false);
```

In almost all instances in the code base, this was used incorrectly. This PR fixes that.


packages/design-system/src/utils/localStore.js
```diff
@@ -37,17 +37,17 @@ function getItemByKey(key) {
   try {
     const stored = localStorage.getItem(key);
     parsed = JSON.parse(stored);
-  } catch (e) {
-    trackError('local_storage_read', e.message);
+  } catch (err) {
+    trackError(err.message);
```
dreamofabear:
```suggestion
    trackError('local_storage_read: ' + err.message);
```
Without the stack trace, the custom name may be helpful to identify the source of errors and categorize them. E.g. multiple `trackError` might have the same error message.

dreamofabear (approved):

packages/design-system/src/utils/localStore.js
swissspidy: Yeah that's a good point. That's the one thing bugging me about exception tracking in GA really.

WDYT about changing the function signature to something like `trackError(prefix, description, fatal = false)` to make that easier? Would make usage a bit more intuitive and require less file changes in this PR.

The `prefix + message` concatenation would then happen inside `trackError`

dreamofabear: SGTM

Issue #8778, opened 2021-08-24
Packages: Move `output` to its own package


swissspidy:
## Task Description

The output package is responsible for creating AMP output for a story.

This would help a lot for things like #8692 where story output needs to be  generated outside of the actual editor.

**To-do:**

* [ ] The unit tests operate with certain element types which are not really known here. Move these to new Jest-based integration tests (as they involve multiple packages)

Issue #7196, opened 2021-04-13
Templates: Load from external website


swissspidy:
## Feature Description

**tl;dr**: the current way of bundling templates with the plugin doesn't scale. This ticket proposes offloading the templates to the wp.stories.google site.

From the beginning we'e imagined the external templates in the editor to be loaded from an external API, i.e. a template directory. Theoretically, other users could contribute to this directory as well.

However, for simplicity's sake we didn't implement such an API initially since there were only 8 such templates, instead deferring this enhancement until it's needed, and instead bundle the templates directly with the final editor bundle.

A template consists of a JSON file (**~160KB** each) plus some metadata (name, description, tags, etc.), with images being served from the wp.stories.google site. That works well.

But now there are **55** additional templates in the making, which would cause the total file size for all templates to reach **10MB**. That's a lot! That not only makes the bundling **slower**, but also makes the bundle significantly **bigger**. A bigger bundle means users' servers could struggle with unzipping larger files.

I am proposing to move the template files to the wp.stories.google site (our `static-site` branch), in the same way that we already host the template images for the exact same reasons.

### Requirements

* Just like the template images, the template JSON files need to be versioned.  
  This way, a user on an older plugin version doesn't get templates from a newer, incompatible version.
* Template metadata (name, description, tags, etc.) needs to be translatable, in a format that is easy to share with our localization team.
* Templates need to be constantly migrated to the newest data version.  
  We currently do this in CI in the plugin, but that's in a different branch than `static-site`

## Alternatives Considered

Max had the interesting suggestion of using a WP instance somewhere for hosting the templates. Where designers could simply create a template in the editor and save it, without any development work needed to get the JSON data.

This poses a few questions though about versioning, internationalization, security, caching, and probably more.

Until we have good answers for those questions I don't think it's a viable path. Hence my suggestion to start with moving the static files to wp.stories.google first.

I also see potential for a simpler version of this: once decoupling of the editor from WP progresses, it might be simpler to have a standalone version of the editor read & write template files from disk, avoiding the need for a WP instance.

## Additional Context

Firebase's built-in support for [i18n rewrites](https://firebase.google.com/docs/hosting/i18n-rewrites) might come in handy here.

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Implementation Brief

**This section is still WIP**

**Prerequisites**

* [ ] Figure out localization process

**Actual Implementation**

CDN:

[...]

PR #8861, opened 2021-08-30
Dashboard: Remove Saved Templates Related Code

### LINKED ISSUE #8746: Dashboard: Clean up or Delete  "Saved Templates"
With grid updates for performance improvements to My Stories and Explore Templates, the saved templates view is increasingly obsolete. It doesn't make sense to take time to keep this up to date if we're not going to use it any time soon.

As it is not coming up soon on the roadmap, I think we should just clear out the code for it in the dashboard. When we are ready to implement, very little of the current code is relevant anymore so we'd really not be missing anything.

If this is wrong and we are wanting to implement this feature soon then instead we should take the time to catch the components up.

BrittanyIRL:
## Context

Last summer we had scaffolded out the 'saved templates' view in the dashboard ahead of getting data flowing. There's still questions out about what all a saved template is and how to handle them. The dashboard has changed since then and maintaining this code is not worth it since it's out of date and getting saved templates working is not a current priority. At this point, it'll be easier to just add in what we need when we need it rather than try to update all of this.

## Summary

Remove all references to saved templates (bookmark, create template from story, in progress story actions) in dashboard as well as 3 defunct experiments (`enableInProgressStoryActions` and `enableBookmarkActions`, `enableInProgressViews`). 

Now the only reason we need `@web-stories-wp/story-editor` is to create a story from a template (`getStoryPropsToSave`). 

## Relevant Technical Choices

Just removing relevant code. 

## To-do

None.

## User-facing changes

Just removing experiments.

## Testing Instructions

Just some regression testing here. Make sure that all actions expected are present in the dashboard for stories and templates, and when you go to `wp-admin/edit.php?post_type=web-story&page=stories-dashboard#/saved-templates` (saved templates route) you just get the same empty page you would if you went anywhere without a defined route in the dashboard.

1.

## Checklist

---

Fixes #8746


swissspidy:
> Now the only reason we need `@web-stories-wp/story-editor` is to create a story from a template (`getStoryPropsToSave`).

We're so close! #8778 and/or #7196 will help us eliminate this one eventually.

includes/Experiments.php
```diff
@@ -274,17 +252,6 @@ public function get_experiments(): array {
 				'description' => __( 'Enable in-progress template actions', 'web-stories' ),
 				'group'       => 'dashboard',
 			],
-			/**
-			 * Author: @brittanyirl
-			 * Issue: 2292
-			 * Creation date: 2020-06-11
-			 */
-			[
-				'name'        => 'enableBookmarkActions',
```
swissspidy, line 284: Looks like we used that experiment name for a PHP unit test which now fails because it was removed 😅 

https://github.com/google/web-stories-wp/blob/64bc19d15741ec175ca7a8121ce2db8ad1153499/tests/phpunit/tests/Experiments.php#L237-L244

Can you update that test to use a name that still exists?

swissspidy:
Some editor Karma tests seem to be failing? If they're flakey and completely unrelated, feel free to skip & file a ticket as usual.

BrittanyIRL:
> Some editor Karma tests seem to be failing? If they're flakey and completely unrelated, feel free to skip & file a ticket as usual.

I hadn't actually seen this test fail until today. I'll check in with prometheus and see if something's up.

packages/dashboard/src/app/font/fontProvider.js
```diff
@@ -1,96 +0,0 @@
...
- * External dependencies
- */
-import PropTypes from 'prop-types';
-import { useCallback, useEffect, useState } from '@web-stories-wp/react';
-import {
-  FontContext as Context,
-  useLoadFontFiles,
-} from '@web-stories-wp/story-editor';
-
-function FontProvider({ children }) {
```
samwhale, line 27: 🥳

packages/dashboard/src/types.js
```diff
@@ -51,7 +49,7 @@ export const TemplatePropType = PropTypes.shape({
   id: PropTypes.number.isRequired,
   status: DashboardStatusesPropType,
   title: PropTypes.string.isRequired,
-  pages: PropTypes.arrayOf(StoryPropTypes.page),
+  pages: PropTypes.array,
```
samwhale, line 52: Did this type change?

samwhale (approved):
Awesome!

packages/dashboard/src/types.js
BrittanyIRL: `pages` is just used to get the length of a template now.  I wanted to limit the imports of `@web-stories-wp/story-editor` so i opted to just swap it to an array since the content of that array is ditched in the serializer as soon as data's retrieved.

samwhale: 🤔 Probably not worth the trouble but maybe we should update that to `numPages`?

spacedmonkey:
Wow the `chunk-getStoryPropsToSave` is massive. Can't wait for https://github.com/google/web-stories-wp/issues/8778

spacedmonkey (approved):

packages/dashboard/src/types.js
BrittanyIRL: well we still need the pages so that we can generate stories from templates at this point.  I can add back in the dependency if this is confusing. it feels like maybe it's confusing given this thread. it just seemed redundant since it's not getting used in any component it's just getting passed.

PR #8801, opened 2021-08-25
Dashboard: Replace Template Details with Images 

BrittanyIRL:
## Context

Part of efforts to remove dependencies on the editor in the dashboard. 

## Summary

Instead of rendering each template as a story in the detail view, use screenshots. 

Perf metrics (macbook pro, high speed internet, chrome v 92.0.4515.159):
First load of a template details page (refresh and load dashboard with detail view): ~1493ms now vs. ~7478ms
Next template (arrow from detail view): ~100ms to load now vs ~1572ms

Screen Shot 2021-08-26 at 3 25 02 PM
demo detail page

## Relevant Technical Choices

- Cleaned up the `cardGallery` (not convinced this belongs in `/components`, think it should be in the `detailTemplate` view since it's specific - left it in place for easier diff reading). There's no need for recalculating the size anymore because we can respect the ratio of the image and just set up a grid instead. 
-  Added the poster images to templates when we grab them in `useTemplates`, updated explore templates and karma data accordingly. 

## To-do

- [x] Grab screenshots 
- [x] Grab perf metrics 
- [x] Double check for any lingering `FontProviders`, `TransformProviders`, etc.  🔥 
- [x] Testing directions

This clears the path for #8746 (that'll remove the final dependencies on the story-editor package) as well as a deeper perf dive for the dashboard now that we can see what else is shifting the page layout or rendering maybe a few more times than should be necessary. 

## User-facing changes

- Template detail grid now uses screenshots instead of rendering template as a story (no animations) 
- There should be a purple border around the selected corresponding small grid item to the larger (selected) page. This matches #8783 update for links on my stories/dashboard. 
- Mostly it should just load faster. 

## Testing Instructions

- Load the dashboard, go to explore templates. Verify explore templates is still loading the first page of the template as a screenshot (did some behind the scenes work, so regression testing on that is good). 
- Click on 'see details' of any template, page should load as usual. Verify that template details view has screenshots now. 
- Verify that gallery of template details pages is navigable by keyboard and with a cursor. Clicking any of the pages will load it in the larger view. 
- You can create a story from a template in the template detail view.

- [ ] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1.

### Does this PR have a security-related impact?

No

### Does this PR change what data or activity we track or use?

Cleaning up the detail template view, realized we're tracking template usage from the detail view but not the gallery - updated so we're tracking the same info from the gallery as we are the detail view. 

### Does this PR have a legal-related impact?

No

## Checklist

[...]

PR #8783, opened 2021-08-24
Dashboard: Update Grid Item UX for links

BrittanyIRL:
## Context

Came up recently when updating capabilities in the dashboard that it was inconsistent when a contributor saw a link to 'edit story' when they really can't. This fixes that. 

## Summary

Updates what's clickable to navigate from the dashboard to editing a story. It used to be the story title, now it's just clicking the story itself.

link vs no link

## Relevant Technical Choices

- When a user has edit capabilities (`story.capabilities.hasEditAction`)  an anchor element gets added to the entire grid item, positioned absolutely so that it doesn't interfere with the ability to open and interact with the context menu or renaming a story. 
- While here, I found a regression on the focus in the grid that exists in staging as well, thought I'd clean it up. I wanted to make sure I was fixing it in a way that wouldn't resurrect with the update @samwhale had in #8719 so I rebased on that while it was still in review (that's why the CLA is mad). Fixed the issue and cleaned up the unit tests to make sure of it. Now when the context menu is closed, check to see if it was with tab or shift + tab and refocus accordingly.
- I tried to clean up areas that I was already working in. We've got a clean up ticket (#8746 ) that will take care of a good chunk of this. Biggest thing I addressed here was forwarding the ref needed to control the focus of the grid (using arrows to navigate so ya don't get trapped tabbing through endlessly) to the gridItem. 

## To-do

- It'd be nice to separate out the karma tests on the dashboard - the 'my stories' file is so big and daunting, it's easy for things to get lost in there. Separating it out would make it easier to have more robust test sets for different functionality (like maintaining proper focus). 

## User-facing changes

If you have the ability to edit a story, you will get a purple border around grid items in the grid view of 'my stories' when you hover or focus on that item. If you do not have edit capabilities of a story then you will not see this nor will the story be clickable.

## Testing Instructions

Have a few stories in draft and published by at least an admin and contributor. Log in as both, hover and focus on stories. See that when you're a contributor you can only edit contributor stories and the published admin stories are not editable for you. 

- [ ] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1.

### Does this PR have a security-related impact?

No

### Does this PR change what data or activity we track or use?

No

### Does this PR have a legal-related impact?

No

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
[...]

PR #8869, opened 2021-08-30
Design System: Fix forwardRef in context menu link

BrittanyIRL:
## Context

Screen Shot 2021-08-30 at 1 41 16 PM

## Summary

Forwards a ref to `Link` option in the design system so that the `ContextMenu` `MenuItem` can use it and this warning goes away. The dashboard context menu's the only place where this is needed.  Originally this was done in #8783 but it got removed in #8801 on rebasing. 

## Testing Instructions

1.

## Checklist

---

Fixes #

### DIFF
--- a/packages/design-system/src/components/typography/link/index.js
+++ b/packages/design-system/src/components/typography/link/index.js
@@ -69,22 +69,7 @@ ConditionalSpanWrapper.propTypes = {
   isWrapped: PropTypes.bool,
 };
 
-function Link({ children, ...props }) {
-  const isExternalLink = props.target === '_blank';
-  return (
-    <StyledAnchor {...props}>
-      <ConditionalSpanWrapper isWrapped={isExternalLink}>
-        {children}
-        {isExternalLink && <StyledLaunch />}
-      </ConditionalSpanWrapper>
-    </StyledAnchor>
-  );
-}
-
-const LinkWithRef = forwardRef(function LinkWithRef(
-  { children, ...props },
-  ref
-) {
+const Link = forwardRef(function Link({ children, ...props }, ref) {
   const isExternalLink = props.target === '_blank';
   return (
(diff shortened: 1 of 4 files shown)


maxyinger (approved):

packages/design-system/src/components/typography/link/index.js
```diff
@@ -79,6 +80,21 @@ export function Link({ children, ...props }) {
...
+  const isExternalLink = props.target === '_blank';
+  return (
+    <StyledAnchor ref={ref} {...props}>
+      <ConditionalSpanWrapper isWrapped={isExternalLink}>
+        {children}
+        {isExternalLink && <StyledLaunch />}
+      </ConditionalSpanWrapper>
+    </StyledAnchor>
+  );
+});
```
samwhale: 💭 This is the same component as `Link`. What is the  reason for exporting one that is wrapped with `forwardRef` and one that isn't?

I feel like this should work for all cases:

```js
export function Link = forwardRef(function Link(
  { children, ...props },
  ref
) {
  const isExternalLink = props.target === '_blank';
  return (
    <StyledAnchor ref={ref} {...props}>
      <ConditionalSpanWrapper isWrapped={isExternalLink}>
        {children}
        {isExternalLink && <StyledLaunch />}
      </ConditionalSpanWrapper>
    </StyledAnchor>
  );
});
```

samwhale (changes requested):

packages/design-system/src/components/typography/link/index.js
BrittanyIRL: you're absolutely right. I had it this way originally because it seemed like a bunch of console warnings and test failures were associated with it, but it's fine now. Thank you for flagging this.

samwhale (approved):
Dope!

PR #8879, opened 2021-08-31
Enabled smart colors

### LINKED ISSUE #8238: Enable feature: Text presets: smart color
## Feature Description
The feature "Text presets: smart color" is currently behind a feature flag. For more context: https://github.com/google/web-stories-wp/pull/8118#discussion_r665745831

- [x] Start the image generation when hovering on the library header generally (or at least more widely on the Text tab) instead of directly on the `+` quick-action icon, to get the image ready sooner.
- [x] Do color calculation when hovering on the preset.
- [x] Skip image generation if presets are added one after another (since it's placed below the previous, no need for new image)
- [ ] Add opt-in toggle for using the feature
- [ ] Mark the feature flag `true` by default.

## Additional Context
image

---

## Implementation Brief
- Remove the feature flag (or better -- just mark `true` by default`).
- Do the color calculation already when hovering over the preset, we should already be able to get the information before the click is done.
- If possible, skip image generation if the preset is added below the previous since the area that's being used for calculation is not influenced by the previous text. This would improve the situation only for consecutively added text presets. We would need to move the previously added preset data ref higher up for it being accessible in the places needed.

- Generate the image when hovering anywhere on the text tab of the panel, not just the quick action item specifically.

barklund:
## Context

This enables smart colors by default.

## Testing Instructions

1. Add a text preset or text set to various backgrounds and see the elements added in proper contrasting color (potentially with a background/scrim).
2. Observe that this does not cause a "serious" lag before insertion.

## Checklist

---

Fixes #8238

### DIFF
--- a/includes/Experiments.php
+++ b/includes/Experiments.php
@@ -350,6 +350,7 @@ class Experiments extends Service_Base {
 				'label'       => __( 'Smart text color', 'web-stories' ),
 				'description' => __( 'Enable text insertion with smart color ensuring good contrast with the background', 'web-stories' ),
 				'group'       => 'editor',
+				'default'     => true,
 			],
 			/**
 			 * Author: @merapi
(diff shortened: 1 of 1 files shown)


barklund:
For now this will be delayed due to performance reasons until we have an opt-in UX for users.

PR #8880, opened 2021-08-31
Media Hotlinking: Improve the logic in case of empty value

### LINKED ISSUE #8847: Media hotlinking error handling
## Bug Description

Noticed while testing #8806

When clicking into the input field and then moving out of it, `https://` is inserted in the input field and subsequently flagged as invalid URL.

I am then unable to clear the input field, so it's impossible for me to see the list of allowed file types.

I can not even select the text below the input field due to this.

## Expected Behaviour

If the input field is empty, this should not be happening.

## Steps to Reproduce

See video below

## Screenshots

https://user-images.githubusercontent.com/841956/131357539-8c34bfdd-a1cc-4f7a-94db-ac4018dbfb13.mov

## Additional Context

- Plugin Version: 1.11.0-alpha
- WordPress Version: 5.8
- Operating System: macOS
- Browser: Chrome

miina:
## Summary
Improves how the messages are handled in case of empty link values.
- When link is empty, Insert button doesn't do anything
- When there's an error and the user changes the link, the informative message about file types to use is displayed again
- When the input is empty and the users tabs away from it, nothing happens.

## User-facing changes
empty-link

## Testing Instructions

1.

## Checklist

---

Fixes #8847

### DIFF
--- a/packages/story-editor/src/components/library/panes/media/local/hotlink/hotlinkModal.js
+++ b/packages/story-editor/src/components/library/panes/media/local/hotlink/hotlinkModal.js
@@ -85,12 +85,13 @@ function HotlinkModal({ isOpen, onClose }) {
 
   const onChange = useCallback(
     (value) => {
-      if (!value?.length) {
+      // Always set the error to null when changing.
+      if (errorMsg) {
         setErrorMsg(null);
       }
       setLink(value);
     },
-    [setLink]
+    [setLink, errorMsg]
   );
 
   return (
--- a/packages/story-editor/src/components/library/panes/media/local/hotlink/hotlinkModal.js
+++ b/packages/story-editor/src/components/library/panes/media/local/hotlink/hotlinkModal.js
@@ -75,11 +75,23 @@ function HotlinkModal({ isOpen, onClose }) {
   });
 
   const onBlur = useCallback(() => {
-    setLink(withProtocol(link));
-    if (!isValidUrl(link)) {
-      setErrorMsg(__('Invalid link.', 'web-stories'));
+    if (link?.length > 0) {
+      setLink(withProtocol(link));
+      if (!isValidUrl(link)) {
+        setErrorMsg(__('Invalid link.', 'web-stories'));
+      }
     }
   }, [link]);
+
+  const onChange = useCallback(
+    (value) => {
+      if (!value?.length) {
+        setErrorMsg(null);
+      }
+      setLink(value);
+    },
+    [setLink]
+  );
 
   return (
     <Dialog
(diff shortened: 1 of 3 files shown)


miina:
@swissspidy Let me know if you think this new logic makes sense -- I've added a gif and description of how it works now.

packages/story-editor/src/components/library/panes/media/local/hotlink/hotlinkModal.js
```diff
@@ -75,12 +75,24 @@ function HotlinkModal({ isOpen, onClose }) {
...
+        setErrorMsg(__('Invalid link.', 'web-stories'));
+      }
     }
   }, [link]);
 
+  const onChange = useCallback(
+    (value) => {
+      if (!value?.length) {
+        setErrorMsg(null);
+      }
```
barklund: Why only clear the message if empty? I think the message should always be cleared on change.

If you enter a bad URL, try to submit, see the error message, select everything and then paste a valid URL into the input, it still shows the message, which is misleading.

So, perhaps just always clear the error on change?

packages/story-editor/src/components/library/panes/media/local/hotlink/useInsert.js
```diff
@@ -111,6 +111,9 @@ function useInsert({ link, setLink, setErrorMsg, onClose }) {
   );
 
   const onInsert = useCallback(() => {
+    if (!link?.length) {
```
barklund: Isn't this just `if (!link)`?

packages/story-editor/src/components/library/panes/media/local/karma/hotlink.karma.js
```diff
@@ -54,12 +54,25 @@ describe('Embedding hotlinked media', () => {
...
+    await waitFor(() => expect(dialog.textContent).toContain('Invalid link'));
+
+    // Delete the value, verify now the informative message show instead again.
+    await fixture.events.click(input, { clickCount: 3 });
+    await fixture.events.keyboard.press('Del');
+    dialog = screen.getByRole('dialog');
+    await waitFor(() => expect(dialog.textContent).toContain('You can insert'));
+
     await fixture.events.click(input);
     await fixture.events.keyboard.type('https://example.jpg');
```
barklund, line 71: But that's not a valid link? 😱

miina: No, it's not a valid image link, that's why it will display an error (including `Media failed to load`).

packages/story-editor/src/components/library/panes/media/local/hotlink/hotlinkModal.js
miina: Hm, yep, makes sense.

barklund (approved):
🌐

PR #8888, opened 2021-08-31
Docs: Add FAQ item about compatibility PPC

dreamofabear:
Adds a new FAQ item about our compatibility (AMP validation) pre-publish check.

> ### How do I fix the error "Compatibility: Your story contains errors that can affect eligibility to appear on Google"?
> 
> This is typically caused by a conflict with another WordPress plugin or a hosting configuration issue. We recommend trying to identify the offending plugin by disabling other WordPress plugins and progressively re-enabling them until the error reappears. Please share the details of your issue on our support forum.

Also fixes a formatting bug in the sentence: "For an optimal Web Story creation experience, we recommend using the latest version of popular web browsers (Chrome, Firefox, Safari, Edge)."

## Screenshot

Screen Shot 2021-08-31 at 4 05 37 PM


swissspidy:
This should be fixed in the source file (`build/en/docs/index.njk`) and then ideally be properly built using eleventy. But should be OK to skip that build for now, since it's a bit buggy. But the source file should be updated for sure.

swissspidy (approved):

PR #8895, opened 2021-08-31
Right Click Menu: Fix the click problems

### LINKED ISSUE #8884: Right Click Menus: Clicking behaviors
This ticket is to improve the clicking behaviors of the right click menu stemming from feedback during bug bash. As these all kind of coexist, fixing them together is probably the most efficient way. Acceptance criteria is fixing everything in this task list.

See [bug bash doc ](https://docs.google.com/document/d/1OuZWwAWxPGDEw5Vm285b_z4Hxdoej3DGLtlYq7sqS-4/edit#)for gifs (can't pull those out of the file)

- [ ] When opening the right-click menu and left-clicking another element, the right-click menu for that other element appears while mouse is down, but disappears once mouse is lifted. Looks weird? (thanks @barklund ). Also reported by @choumx : in this video I'm opening the right-click menu and then left-clicking somewhere else, which shows another menu behind the first one.
- [ ] Disabled items should have pointer-events enabled(?) otherwise clicking on them will select the element underneath (@merapi). Clicking on disabled items shouldn't remove focus from the 'selected' element.  
- [ ] Clicking on the divider between selections should also capture and ignore pointer events. (@barklund)

samwhale:
## Context

Bug Bash for 1.11 uncovered [some click problems](https://docs.google.com/document/d/1OuZWwAWxPGDEw5Vm285b_z4Hxdoej3DGLtlYq7sqS-4/edit#heading=h.4swgle76797u) with the menu. This fixes those.

## Summary

1. Clicking anywhere in the menu that's _not_ an enabled button will keep the same element selected in the canvas.
2. Mouse down on the mask will not de-select the selected element.

## Relevant Technical Choices

n/a

## To-do

n/a

## User-facing changes

|Before|After|
|--|--|
|click-change|click-no-change|
|clicky-broke|clicky-fix|

## Testing Instructions

1. Right click any element other than the background
2. Click on a disabled item. Nothing should happen
3. Hold the mouse down away from the menu -> should close the menu

no

no

no

## Checklist

---

Fixes #8884

### DIFF
--- a/packages/design-system/src/components/contextMenu/mask.js
+++ b/packages/design-system/src/components/contextMenu/mask.js
@@ -64,6 +64,7 @@ export default function Mask({ onDismiss }) {
         data-testid="context-menu-mask"
         hasOnDismiss={Boolean(onDismiss)}
         onClick={onDismiss || noop}
+        onMouseDown={onDismiss || noop}
       />
     </>
   );
--- a/packages/design-system/src/components/contextMenu/mask.js
+++ b/packages/design-system/src/components/contextMenu/mask.js
@@ -58,12 +58,11 @@ export default function Mask({ onDismiss }) {
       {/*
         Disable Reason: Allow pointer events to pass through if there's no 'onDismiss' to preserve transition
         */}
-      {/* eslint-disable-next-line styled-components-a11y/click-events-have-key-events, styled-components-a11y/no-static-element-interactions */}
+      {/* eslint-disable-next-line styled-components-a11y/no-static-element-interactions */}
       <ScreenMask
         ref={maskRef}
         data-testid="context-menu-mask"
         hasOnDismiss={Boolean(onDismiss)}
-        onClick={onDismiss || noop}
         onMouseDown={onDismiss || noop}
       />
     </>
--- a/packages/design-system/src/components/contextMenu/mask.js
+++ b/packages/design-system/src/components/contextMenu/mask.js
@@ -22,7 +22,6 @@ import { useEffect, useRef } from '@web-stories-wp/react';
 /**
  * Internal dependencies
  */
-import { noop } from '../../utils';
 import { POPOVER_Z_INDEX } from './styled';
 
 const ScreenMask = styled.div`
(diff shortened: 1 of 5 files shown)


packages/design-system/src/components/contextMenu/menu.js
```diff
@@ -290,14 +290,18 @@ const Menu = ({
...
+      role="menu"
+      {...props}
+    >
       <MenuList
         data-testid="context-menu-list"
         ref={listRef}
         isIconMenu={isIconMenu}
         aria-label={groupLabel}
         role="group"
-        {...props}
```
samwhale, line 300: Moved props up one layer here so that props like `className` and `onMouseDown` apply to the outer container

packages/design-system/src/components/contextMenu/mask.js
```diff
@@ -64,6 +64,7 @@ export default function Mask({ onDismiss }) {
         data-testid="context-menu-mask"
         hasOnDismiss={Boolean(onDismiss)}
         onClick={onDismiss || noop}
+        onMouseDown={onDismiss || noop}
```
BrittanyIRL: 🤔 using both onClick and onMouseDown here duplicates the `onDismiss` getting fired. Since mouseDown occurs before the click action takes hold. Do we need both? Could onClick get removed in favor of the mouseDown? If not, then should mouseDown just stop bubbling and never trigger `onDismiss`?

packages/story-editor/src/components/canvas/rightClickMenu.js
```diff
@@ -56,6 +56,7 @@ const RightClickMenu = () => {
             'web-stories'
           )}
           maskRef={maskRef}
+          onMouseDown={(evt) => evt.stopPropagation()}
```
BrittanyIRL, line 59: 🎉

packages/design-system/src/components/contextMenu/mask.js
samwhale: Nice catch!

BrittanyIRL (approved):

BrittanyIRL:
(Looks like some of the unit tests need updating to use onMouseDown)

maxyinger (approved):
lgtm

PR #8925, opened 2021-09-01
Templates: Show colors as saved story colors 

### LINKED ISSUE #8828: e2e test: using a template loads saved colors
Let's add a new e2e test for using a template in the dashboard, loading it in the editor and making sure the same list of colors is present in the 'saved colors' panel as is in the detail view of the template.

### LINKED ISSUE #8829: Template colors added to story saved colors
## Description
- Template colors should be visible as "saved colors" in the editor under the "story" section (not all stories/global, specific to this one instance). 

## Additional information 

I did a bit of debugging here before pointing this and here's my starting point: 

- The colors' hex codes are in the metaData file of each respective template in the template package (packages/templates/src/raw/[template-name]/). They are stored as an array with a label for description and a hex code ({label: 'winter blue', color: '#a4deff'}).
- The saved colors isn't new, users can save their own colors already. Saved colors live in style_presets, accessible through the useStory context. When a story's loaded the saved colors get structured (see story-editor/src/app/story/effects/useLoadStory)
- Right now the colors of a template (template.colors) are not getting passed to getStoryPropsToSave when the 'use template' button is clicked and createStoryFromTemplate request is triggered.
- Pretty sure that the logic is all set up on the editor side to handle this, but that the saved colors of a template aren't quite making it into the story when the editor loads.

## Acceptance criteria
- When you use a template, once the editor loads you should see the same list of colors in the "saved colors" panel as you do in the template detail view of said template.

BrittanyIRL:
## Context

When a template is selected from the dashboard it should load the colors of that template as the the saved story colors. 

## Summary

Adding the template colors to the story data that's generating a story so that we get the template's colors as saved story colors. 

## Relevant Technical Choices
- The story structure wants colors as rgb values while template metadata assigns hex values as colors. Converting template colors when a template is selected to be used as a story and passed into the story as `currentStoryStyles.colors` so that they load as the saved story colors. 
- Update `Swatch` `pattern` propType to be more descriptive 
- Added 2 `data-testid`s to get counts of colors for an e2e test. the color swatches on a detail template view and color groups used as swatches in the saved colors panel. 
- Update the `Template` e2e test to track how colors present for a given template (the 'Fresh and Bright', id: 1 template) and then checking the same colors are shown when the template loads as a story in the saved colors. Left all other checks in play. Only real difference is that the 'use template' button that's getting clicked is the one on the detail view. 

## To-do

- [x] update to use patters util instead of polished 
- [x]  Check colors not just length of list in e2e test

## User-facing changes

Now when you use a template you will get the template's colors in the saved story colors panel. 

saved template colors

## Testing Instructions

- Go to dashboard templates, go to the detail view of a template. 
- Take note of the colors listed below the template description. 
- Now use that template as a story and go to the 'saved colors' panel. 
- The same colors that are in the template detail view should be listed as the saved story colors (not global). 

1.

No

No

No

## Checklist

---

Fixes #8828 
Fixes #8829


packages/dashboard/src/utils/getStoryColors.js
```diff
@@ -0,0 +1,44 @@
...
+      if (!color || color?.toString().charAt(0) !== '#') {
+        return null;
+      }
+      const parsedColor = parseToRgb(color);
+
+      const { red: r, green: g, blue: b } = parsedColor;
+      return { color: { r, g, b } };
+    })
+    .filter((color) => color);
+};
```
swissspidy: There's this handy `createSolidFromString` utility in the `@web-stories-wp/patterns` package that does most of this for you.

Should be able to simplify this to `return colors.map(createSolidFromString)`

On that note, curious to hear more about the safeguards you've added. Don't all templates have correct hex codes?

packages/design-system/src/components/swatch/swatch.js
```diff
@@ -168,7 +168,14 @@ function Swatch({
 
 Swatch.propTypes = {
   children: PropTypes.node,
-  pattern: PropTypes.object,
+  pattern: PropTypes.shape({
```
swissspidy: Use `HexPropType` from the `@web-stories-wp/patterns` package.

Maybe create a new `ColorPropType` there:

```js
export const ColorPropType = PropTypes.shape({
  color: HexPropType.isRequired,
});
```

packages/e2e-tests/src/specs/dashboard/templates/useTemplate.js
```diff
@@ -42,15 +42,42 @@ describe('Template', () => {
 
     await percySnapshot(page, 'Explore Templates');
 
-    await expect(firstTemplate).toClick('button', { text: 'Use template' });
+    await expect(firstTemplate).toClick('a', { text: 'See details' });
+    // Get count of template colors to compare to 'saved colors' in the editor.
```
swissspidy, line 46: Would it be trivially possible to not only compare the count but also the actual colors?

BrittanyIRL: Yesterday, it seemed like it wasn't worth the energy, but today - it feels easier to do and like it would be beneficial. 👍

packages/dashboard/src/utils/getStoryColors.js
BrittanyIRL: oh awesome, i poked around a little in the editor looking for a util like that thinking it had to exist already but i just couldn't find it!  I'll swap this out and simplify, thank you!

The safeguards are just in case something got really botched in a template's metaData because `parseToRgb` has no safeguards so it was just erroring out if you pass a value it doesn't expect.

maxyinger (approved):
Got a nice e2e test in there and everything 👏

swissspidy (approved):

Issue #680, opened 2020-03-20
Video Trimming


jauyong:
## Feature Description

For videos, I’d like to be able to choose the start and end time within any uploaded video to be able to cut the video displayed in my story.

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

diegovar:
Can you link to the designs or any PRD parts that mention this functionality?

jauyong:
@diegovar I don't think this has been spec'd at all. I originally filed this as a feature request a while back. 
@o-fernandez I see its a TBD P1 in the CUJ sheet in row 18. Are there further details for this feature right now?

o-fernandez:
I just looked through the spec and [this is all I could find](https://docs.google.com/document/d/1EQ12sD4a1plmMeBu7YMWOdUSg0FeegB9qi2_OlV3Tp0/edit#bookmark=id.3oa4doedhryq). Needless to say, no real spec for this yet (neither PM nor UX) so adding the necessary labels for Sam and I to take a look. 

@diegovar is this something you were planning to work on in the near future? If so, we may need to figure out how to prioritize to get the right specs to you.

diegovar:
I was trying to better understand which issues were available to grab and which ones need more info or are blocked. My objective is for the issues tagged as "media" to be in a good enough state so that it's clear whether an issue can be grabbed before having to open and read the discussion.

o-fernandez:
Ok got it. For now, if you find anything that is not spec'd, just add the "Needs More Info" or "UX Needed" labels and those are understood to need either a product spec or UX. I'll try to clean this up somewhat as well, so hopefully things are actionable for you!

aaskedall:
<img width="360" alt="description" src="https://user-images.githubusercontent.com/4030811/131363443-8b5a5edf-bd51-4fc1-9a45-a6232fb07959.png">
<img width="344" alt="video settings trim" src="https://user-images.githubusercontent.com/4030811/131363444-65487e01-6823-4fa4-ba16-ce162bb036dd.png">
<img width="1142" alt="If video is smaller than canvas" src="https://user-images.githubusercontent.com/4030811/131363447-b614446f-4ae6-48a5-8a11-43595f3e3e90.png">
<img width="1142" alt="full screen editor" src="https://user-images.githubusercontent.com/4030811/131363448-c368aeb0-1b91-4f88-b9a3-e49ada1f5f3f.png">
[...]

PR #8927, opened 2021-09-02
Added video trim dummy as a new edit mode

### LINKED ISSUE #8877: Create new video trim mode
## Feature Description

This starts the work on video trimming as per designs.

The purpose of this PR is to lay the foundation for the work needed to complete this feature - not develop all of it.

The main part of this is to create the video trim mode in the canvas layering system. It works similarly to the existing edit-mode, but is different from it as it has different UX. Either this new mode is an extension of the existing edit mode (using the same layering but with some "mode" flag or similar) or it is a duplicate (with proper DRY).

* [x] Add a new feature flag
* [x] Add a button to the video setting panel trigger video trim mode (for now, ignore UX weirdness with the "mute button" if both are present)
* [x] Add the new video trim mode to the canvas
* [x] For now, just show the video poster image on the canvas, not the actual video (as we do for the existing video edit component)
* [x] Show a placeholder video trimmer in place of the page menu (most likely on top of the existing page menu). Don't create the actual trim component, just a placeholder.
* [x] Clicking on the element or the video trimmer does nothing. Clicking outside either exits video trim mode. Pressing <kbd>esc</kbd> exits video trim mode.

### Design panel

video settings trim

### Canvas

| Background video | Foreground video |
|-|-|
| image (20) | image (21) |

NB: Ignore the crop marks/resize handles on the foreground video above. There is no resizing the video or mask in the video trim mode.

barklund:
## Context

This implements the first step of allowing video trimming in the editor.

## Relevant Technical Choices

* I decided to re-use the existing edit mode to do video trimming, as it made sense in the context. It is merely an edit element state flag. Edit element state was previously only used for text edit, but now also for video. This new edit mode can be triggered elsewhere by setting this state flag when entering edit mode.

## To-do

_Copied from #8877:_

* [x] Add a new feature flag
* [x] Add a button to the video setting panel trigger video trim mode (~for now, ignore UX weirdness with the "mute button" if both are present~ - this is actually solved properly)
* [x] Add the new video trim mode to the canvas
* [x] ~For now, just show the video poster image on the canvas, not the actual video (as we do for the existing video edit component)~ For now show the video muted, looping and auto-playing
* [x] Show a placeholder video trimmer in place of the page menu (most likely on top of the existing page menu). Don't create the actual trim component, just a placeholder.
* [x] Clicking on the element or the video trimmer does nothing. Clicking outside either exits video trim mode. Pressing <kbd>esc</kbd> exits video trim mode.

## User-facing changes

This gif demos that both the regular video edit mode and the new video trim mode co-exists:
trim-mode

## Checklist

---

Fixes #8877

### DIFF
--- a/packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
+++ b/packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
@@ -38,6 +38,7 @@ import { useCallback, useMemo, useEffect } from '@web-stories-wp/react';
 /**
  * Internal dependencies
  */
+import useCanvas from '../../../../app/canvas/useCanvas';
 import { Row as DefaultRow } from '../../../form';
 import { SimplePanel } from '../../panel';
 import { getCommonValue } from '../../shared';
@@ -50,6 +51,10 @@ const Row = styled(DefaultRow)`
 
 const StyledButton = styled(Button)`
   padding: 12px 8px;
+`;
+
+const TrimButton = styled(StyledButton)`
+  margin-left: 20px;
 `;
 
 const Label = styled.label`
@@ -72,6 +77,7 @@ const HelperText = styled(Text).attrs({
 
 function VideoOptionsPanel({ selectedElements, pushUpdate }) {
   const isMuteVideoEnabled = useFeature('enableMuteVideo');
+  const isVideoTrimEnabled = useFeature('enableVideoTrim');
   const { isTranscodingEnabled } = useFFmpeg();
   const { muteExistingVideo } = useLocalMedia((state) => ({
     muteExistingVideo: state.actions.muteExistingVideo,
@@ -105,11 +111,26 @@ function VideoOptionsPanel({ selectedElements, pushUpdate }) {
     isMuting,
   ]);
 
+  const shouldDisplayTrimButton = useMemo(
+    () => isSingleElement && isVideoTrimEnabled,
+    [isSingleElement, isVideoTrimEnabled]
+  );
+
   const buttonText = useMemo(() => {
     return isMuting
       ? __('Removing audio', 'web-stories')
       : __('Remove audio', 'web-stories');
   }, [isMuting]);
+
+  const { setEditingElementWithState } = useCanvas(
+    ({ actions: { setEditingElementWithState } }) => ({
+      setEditingElementWithState,
+    })
+  );
+
+  const handleTrim = useCallback(() => {
+    setEditingElementWithState(selectedElements[0].id, { isTrimming: true });
+  }, [setEditingElementWithState, selectedElements]);
 
   const speak = useLiveRegion();
 
@@ -137,6 +158,16 @@ function VideoOptionsPanel({ selectedElements, pushUpdate }) {
             {__('Loop', 'web-stories')}
           </Text>
         </Label>
+        {shouldDisplayTrimButton && (
+          <TrimButton
+            variant={BUTTON_VARIANTS.RECTANGLE}
+            type={BUTTON_TYPES.SECONDARY}
+            size={BUTTON_SIZES.SMALL}
+            onClick={handleTrim}
+          >
+            {__('Trim', 'web-stories')}
+          </TrimButton>
+        )}
       </Row>
       {shouldDisplayMuteButton && (
         <>
--- a/packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
+++ b/packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
@@ -122,15 +122,24 @@ function VideoOptionsPanel({ selectedElements, pushUpdate }) {
       : __('Remove audio', 'web-stories');
   }, [isMuting]);
 
-  const { setEditingElementWithState } = useCanvas(
-    ({ actions: { setEditingElementWithState } }) => ({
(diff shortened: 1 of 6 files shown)


packages/story-editor/src/elements/video/edit.js
```diff
@@ -17,12 +17,20 @@
 /**
  * Internal dependencies
  */
+import useCanvas from '../../app/canvas/useCanvas';
 import StoryPropTypes from '../../types';
 import MediaEdit from '../media/edit';
+import Trim from './trim';
 
-// Opacity is adjusted so that the double image opacity would equal
-// the opacity assigned to the video.
```
barklund, line 24: This message seemed out-of-date, so I just removed it.

packages/story-editor/src/elements/video/trim.js
```diff
@@ -0,0 +1,103 @@
...
+    >
+      <Video
+        poster={poster || resource.poster}
+        style={style}
+        {...videoProps}
+        preload="metadata"
+        loop
+        muted
+        autoPlay
+        tabIndex={0}
```
barklund, line 77: The first three are just so that you can see the video do something and the tab index is to allow focus, so you can press <kbd>esc</kbd> to exit trim mode.

packages/story-editor/src/elements/video/trim.js
```diff
@@ -0,0 +1,103 @@
...
+          tracks.map(({ srclang, label, kind, track: src, id: key }, i) => (
+            <track
+              srcLang={srclang}
+              label={label}
+              kind={kind}
+              src={src}
+              key={key}
+              default={i === 0}
+            />
+          ))}
```
barklund, line 91: This is just copied from the video display component. It might make sense to DRY this up, once we later refine this component and see how much it actually shares with the display version.

spacedmonkey:
https://user-images.githubusercontent.com/237508/131821063-1c3b0fa5-cea1-4528-b2e2-5231cbecf7e2.mov

When I click on the trim button, the dummy menu shows. But when the trim button is hit again, then it doesn't hide the dummy menu.

miina:
@barklund  do we have the specs of the feature written out somewhere? I checked the PR and have a question:

As far as I can see from the UX images and description then video trimming should be part of the existing edit mode. Is that correct? It looks like within this PR, the trimming mode is added as a separate form of edit mode, however, in the designs, it's said to be as an addition to the current edit mode.

If that's the case, then clicking `Trim` under the Video Settings should basically just go to Edit mode of the Video (which would have the trimming mode added, too).

Thoughts?

barklund:
@spacedmonkey 
> When I click on the trim button, the dummy menu shows. But when the trim button is hit again, then it doesn't hide the dummy menu.

That's a good point. I guess pressing the button again aborts the trim and exits the mode.

I'll add that.

@miina
> @barklund do we have the specs of the feature written out somewhere?

The feature is only described in these tickets, that I have created under the #680 epic.

> As far as I can see from the UX images and description then video trimming should be part of the existing edit mode. Is that correct?

That is old UX, the new UX might not be shared yet. The feature is to be implemented as described in these tickets.

miina:
This should be the new UX, I think: https://www.figma.com/file/1YM3jA9h4QGc4xLdyaXIJS/Stories---Sprint-1.7%2B?node-id=9735%3A384025

barklund:
> This should be the new UX, I think: https://www.figma.com/file/1YM3jA9h4QGc4xLdyaXIJS/Stories---Sprint-1.7%2B?node-id=9735%3A384025

Ah thanks, yes. With a few exceptions - the text paragraph is not correct and the non-background video shows crop handles, which is also not correct. In video trim mode, the only possible action is video trimming, not anything else.

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
```diff
@@ -105,12 +111,36 @@ function VideoOptionsPanel({ selectedElements, pushUpdate }) {
...
+      isEditing,
+      setEditingElementWithState,
+      clearEditing,
+    })
+  );
+
+  const handleTrim = useCallback(() => {
+    if (isEditing) {
+      clearEditing();
+    } else {
```
barklund, line 139: This does have the weird side-effect, that if you double-click a video to edit the frame/scale and then press the "trim" button, the trim button just exits edit mode without entering trim mode. But that's a minor concern we can always discuss later.

We can also discuss whether this button should appear in a pressed state when the trim mode is active, so it's clear that it's a toggle button?

miina: > We can also discuss whether this button should appear in a pressed state when the trim mode is active, so it's clear that it's a toggle button?

If it's a toggle then yes, it should have active mode :)

miina (approved):
LGTM pending the Trim toggle active state.

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
barklund: Hm, we don't actually have a toggle button design for a non-icon button - [all of the existing variants](https://google.github.io/web-stories-wp/storybook/?path=/story/designsystem-components-button--toggle-buttons) look misplaced in context if I insert them. So I will leave this as-is for now, and we can discuss it when fine-tuning the UX at the end.

PR #8938, opened 2021-09-02
Edit Story: Freeform Taxonomy Component

### LINKED ISSUE #8831: Freeform taxonomy component
## Description
To implement tags in the new taxonomy panel (#489) we need a new type of textarea input that presents a list that is comma based as individual entries (see the below screenshot of the "tags" section). The structure and behavior should be modeled off of the existing version of this used in WordPress if you go edit a post. 

(Demo of adding/removing tags in a post)
gutenberg demo

So we want to do this same thing but using our design system. 
This should be a new variant of the text area in the design system. 

Use the existing input structure and the dark variant of this sketch from @aaskedall - when questions arise, post them in the design channel in slack, I think we can get pretty far using the existing theming we have in place. 
image

## Acceptance criteria:
- Tags should be added with comma or enter, UI should change from type input to a little tag box when comma or enter are keyed.
- Each tag should have an 'x' button, clicking button should trigger a delete function
- Tooltip and cursor pointer when hovering over 'x'  
- RTL and LTR work
- Storybook demo
- accessible and aXe compliant input

maxyinger:
## Context
Part of story taxonomy support

## Summary
Adds a freeform taxonomy (tags) input component with a storybook implementation and unit tests.

## Relevant Technical Choices
just a storybook component for now, not introduced into the editor yet

## User-facing changes
NA

## Testing Instructions
- `npm run storybook`
- `edit story -> components -> form -> tags`
- mess with the tags input and see that it behaves like the WP one mentioned in the ticket.

1.

## Checklist

---

Fixes #8831

### DIFF
--- a/packages/story-editor/src/components/form/stories/tags.js
+++ b/packages/story-editor/src/components/form/stories/tags.js
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * External dependencies
+ */
+import styled from 'styled-components';
+
+/**
+ * Internal dependencies
+ */
+import Tags from '../tags';
+
+const Bg = styled.div`
+  height: calc(100vh - 32px);
+  background: ${({ theme }) => theme.colors.bg.primary};
+`;
+
+const Wrapper = styled.div`
+  position: relative;
+  width: 250px;
+  inset: 15vh 0 0 0;
+  margin: auto;
+`;
+
+export default {
+  title: 'Stories Editor/Components/Form/Tags',
+  component: Tags,
+};
+
+export const _default = () => {
+  return (
+    <Bg>
+      <Wrapper>
+        <Tags.Label>{'Add New Tag'}</Tags.Label>
+        <Tags.Input />
+        <Tags.Description>
+          {'Separate with commas or the Enter key.'}
+        </Tags.Description>
+      </Wrapper>
+    </Bg>
+  );
+};
--- a/packages/story-editor/src/components/form/stories/tags.js
+++ b/packages/story-editor/src/components/form/stories/tags.js
@@ -44,9 +44,9 @@ export const _default = () => {
   return (
     <Bg>
       <Wrapper>
-        <Tags.Label>{'Add New Tag'}</Tags.Label>
-        <Tags.Input />
-        <Tags.Description>
+        <Tags.Label htmlFor="tags-input">{'Add New Tag'}</Tags.Label>
+        <Tags.Input id="tags-input" aria-describedby="tags-description" />
+        <Tags.Description id="tags-description">
           {'Separate with commas or the Enter key.'}
         </Tags.Description>
       </Wrapper>
--- a/packages/story-editor/src/components/form/stories/tags.js
+++ b/packages/story-editor/src/components/form/stories/tags.js
@@ -45,7 +45,11 @@ export const _default = () => {
     <Bg>
       <Wrapper>
         <Tags.Label htmlFor="tags-input">{'Add New Tag'}</Tags.Label>
-        <Tags.Input id="tags-input" aria-describedby="tags-description" />
+        <Tags.Input
+          id="tags-input"
+          aria-describedby="tags-description"
+          name="story-tags"
+        />
         <Tags.Description id="tags-description">
           {'Separate with commas or the Enter key.'}
         </Tags.Description>
--- a/packages/story-editor/src/components/form/tags/input.js
+++ b/packages/story-editor/src/components/form/tags/input.js
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * External dependencies
+ */
+import { themeHelpers } from '@web-stories-wp/design-system';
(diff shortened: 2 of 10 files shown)


packages/story-editor/src/components/form/tags/input.js
```diff
@@ -0,0 +1,153 @@
...
+  margin: 3px 0;
+`;
+
+function Input({ onChange, ...props }) {
+  const [{ value, tags, offset }, dispatch] = useReducer(reducer, {
+    value: '',
+    tags: [],
+    offset: 0,
+  });
+  const [isInputFocused, setIsInputFocued] = useState(false);
```
samwhale, line 64:
```suggestion
  const [isInputFocused, setIsInputFocused] = useState(false);
```

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -0,0 +1,153 @@
...
+        removeTag: (tag) => () => {
+          dispatch({ type: ACTIONS.REMOVE_TAG, payload: tag });
+        },
+        handleFocus: () => {
+          setIsInputFocued(true);
+        },
+        handleBlur: () => {
+          dispatch({ type: ACTIONS.RESET_OFFSET });
+          setIsInputFocued(false);
+        },
```
samwhale, line 107: 💭 Would the `:focus-within:` css selector work here instead of having to set this state manually?

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -0,0 +1,153 @@
...
+          tag === INPUT_KEY ? (
+            <TextInput
+              {...props}
+              key={INPUT_KEY}
+              value={value}
+              onKeyDown={handleKeyDown}
+              onChange={handleChange}
+              onFocus={handleFocus}
+              onBlur={handleBlur}
+              size="4"
```
samwhale, line 132: 💭 Should we allow changing the size? Maybe we could default it to 4 if so

samwhale (commented):
This looks great dude. Really easy to use with both mouse and keyboard.

Question about accessibility: when focusing an option I feel like the option text should be read to me as well as `remove tag` 🤔. Seems like it'll be tricky to tell which tag is being removed
voiceover

samwhale:
Some focus questions:

1. When an option is highlighted with keyboard, should the container also be highlighted? I feel like that would be very nice



2. When removing a tag with the keyboard, should the next tag get focus? This may not matter if the first bullet changes.

packages/story-editor/src/components/form/stories/tags.js
```diff
@@ -0,0 +1,55 @@
...
+  title: 'Stories Editor/Components/Form/Tags',
+  component: Tags,
+};
+
+export const _default = () => {
+  return (
+    <Bg>
+      <Wrapper>
+        <Tags.Label htmlFor="tags-input">{'Add New Tag'}</Tags.Label>
+        <Tags.Input id="tags-input" aria-describedby="tags-description" />
```
BrittanyIRL: nit; add a `name`.

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -0,0 +1,153 @@
...
+  }, [tags]);
+
+  // Prepare and memoize event handlers to be as self
+  // contained and descriptive as possible
+  const { handleFocus, handleBlur, handleKeyDown, handleChange, removeTag } =
+    useMemo(
+      () => ({
+        // Key interactions are modeled after WordPress Tag input UX
+        // and are only available when focused on the text input
+        handleKeyDown: (e) => {
```
BrittanyIRL, line 87: tests for this would be great!

packages/story-editor/src/components/form/tags/tag.js
```diff
@@ -0,0 +1,103 @@
...
+    <Token>
+      <TokenText id={id}>{children}</TokenText>
+      <Tooltip
+        title={__('Remove Tag', 'web-stories')}
+        placement={TOOLTIP_PLACEMENT.BOTTOM}
+        hasTail
+      >
+        <Dismiss
+          onClick={onDismiss}
+          aria-label={__('Remove Tag', 'web-stories')}
```
BrittanyIRL, line 88: It'd be great if this specified what tag was getting removed since a screenreader doesn't read the spans.

BrittanyIRL (approved):
Dig it! 

couple little things but nothing I see as a blocker.  Thanks for all the reducer tests!

Probably not the average user but if you use the arrow key to go to the input placement before a tag that's added this happens which is a little jarring. The shift also happens in gutenberg, what I' curious about is not having that tooltip show up, seems a little weird to me. Thoughts?
shift

BrittanyIRL:
> Some focus questions:
> 
> 1. When an option is highlighted with keyboard, should the container also be highlighted? I feel like that would be very nice
> 
> 
> 
> 1. When removing a tag with the keyboard, should the next tag get focus? This may not matter if the first bullet changes.
> 
> 


These are both inline with gutenberg so i think it's ok
Kapture 2021-09-03 at 10 37 24

packages/story-editor/src/components/form/tags/input.js
maxyinger: We only want it focused when the input is focused. however there will be child tag elements within that can gain focus and we don't want it to look focused when they're focused

packages/story-editor/src/components/form/tags/tag.js
maxyinger: So gutenberg does it like this:
Screen Shot 2021-09-03 at 2 07 25 PM

Which is what we're doing here. I think it might be fine because we pass it a `aria-describedby` attribute.

BrittanyIRL: trueee, if it's good for Gutenberg it's fine for us. 🤝

maxyinger:
> what I' curious about is not having that tooltip show up, seems a little weird to me.

@BrittanyIRL I think it's just cus your mouse is hangin out there and the tag moves underneath your mouse when the content shift happens. shouldn't be an issue if the users navigating purely with keyboard

yeah... yup ok nevermind :)

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -0,0 +1,153 @@
...
+    border-radius: ${theme.borders.radius.small};
+    ${isInputFocused && themeHelpers.focusCSS};
+  `}
+  display: flex;
+  flex-wrap: wrap;
+  padding: 3px 6px;
+  margin-bottom: 6px;
+`;
+
+const TextInput = styled.input.attrs({ type: 'text' })`
```
samwhale: 💭 should we use the input from the design system here as a base?

samwhale:
💭 Should this component live in the design system or in the stories editor?

maxyinger:
> 💭 Should this component live in the design system or in the stories editor?

Me and Brittany kinda chatted about this earlier. I think for now we're just doing it in the editor cus it's a one off component, but if we start leveraging it anywhere else, it makes sense to move to the DS

packages/story-editor/src/components/form/tags/input.js
maxyinger: ah good call. will update

maxyinger: ehh double checking it now, I'm wondering if it has more stuff going on with it than we actually need here

maxyinger: maybe I'll just export a base input component from there that's just the text input with the styles instead of all the other stuff that comes with it. should help with not having to augment the DS input at all, but still share the relevant code

PR #8934, opened 2021-09-02
Add video trimming quick action

miina:
## Context
Adds quick action menu for the Video trimming feature (currently behind a feature flag).

## Testing Instructions
1. Enable Video trimming feature flag.
2. Add a video element
3. Verify it has the Trim quick action which enters the Trim mode.
4. Set the video as background.
5. Verify that now as well has the Trim quick action which enters the Trim mode.

- [ ] This is a non-user-facing change and requires no QA

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [ ] I have tested this code to the best of my abilities
- [ ] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [ ] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [ ] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [ ] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #8878

PR #8944, opened 2021-09-02
Added video context for trimming 

### LINKED ISSUE #8889: Create trimmable, playable video context
## Feature Description

The purpose of this ticket is to set the foundation for the actual video trimmer component, communication between the components as well as introduce a playable component in the layered frame.

* [x] Create a communication layer (new or existing context?) between the layered frame and the trimmer
* [x] Display the video in the frame with the regular play button as for a video display component (DRY-ly) but loop the video regardless of "loop" setting
* [x] Store _start-offset_, _current-position_, and _end-offset_ in the video as milliseconds in the comm layer
* [x] Update _current-position_ as the video plays in the comm layer
* [x] Make a "dummy" trimmer component with three text fields: _start-offset_ (editable), _current-position_ (non-editable), _end-offset_ (editable), that displays/updates the values from the comm layer
* [x] Make the video start at _start-offset_ and end at _end-offset_ as they are changed (note that the video should be looping based on those offsets)
* [x] If _current-position_ is not within _start-offset_ and _end-offset_, move the play-head so it is within the bounds (but maintain playing or paused state as it was before the ends were moved in)

barklund:
## Context

This adds the second step of the video trimming feature allowing communication between the video trim frame and the video trim component plus displays the video looping at the specified offsets.

## Relevant Technical Choices

* I reused the existing video control component, but the interface for the component is pretty ugly. I should clean that up and make another generalized component that's just the play/pause button and use that in the two existing use-cases more cleanly.
* I realized that this communication between edit-frame and outside component was very similar to the rich text context that we already have, so created another context, that lives in a place similar to that context.
* I went with ranges for the dummy inputs, as they actually work pretty well!
* I decided that a video cannot be trimmed to less than 1000ms. Is that a reasonable limit?
* I had to do the restart video functionality in a `setTimeout` as it otherwise did not work correctly with the default end offset being at the end of the video.

## To-do

_Copied from #8889:_
* [x] Create a communication layer between the layered frame and the trimmer
* [x] Display the video in the frame with the regular play button as for a video display component (DRY-ly) but loop the video regardless of the loop setting
* [x] Store _start-offset_, _current-position_, and _end-offset_ in the video as milliseconds in the comm layer
* [x] Update _current-position_ as the video plays in the comm layer
* [x] Make a "dummy" trimmer component with three ~text fields~ ranges: _start-offset_ (editable), _current-position_ (non-editable), _end-offset_ (editable), that displays/updates the values from the comm layer
* [x] Make the video start at _start-offset_ and end at _end-offset_ as they are changed (note that the video should be looping based on those offsets)
* [x] If _current-position_ is not within _start-offset_ and _end-offset_, move the play-head so it is within the bounds (but maintain playing or paused state as it was before the ends were moved in)

## User-facing changes

Dummy video-trimmer in action:
dummy-trimmer

A few things to note here:
* Offsets cannot get closer to each other than 1 second/1000 milliseconds.
* If an offset is dragged beyond the current time, the video skips to that offset (the start offset is dragged in the above gif where the current time suddenly skips).
* It works amazingly well!

## Checklist

---

Fixes #8889


packages/story-editor/src/components/videoTrim/useVideoNode.js
```diff
@@ -0,0 +1,91 @@
...
+    return () => {
+      videoNode.removeEventListener('timeupdate', onTimeUpdate);
+      videoNode.removeEventListener('loadedmetadata', onLoadedMetadata);
+    };
+  }, [startOffset, endOffset, videoNode]);
+
+  const setStartOffset = useCallback(
+    (offset) => {
+      // Start offset must be 1 sec smaller than end offset
+      offset = Math.min(endOffset - 1000, offset);
```
barklund: This should be moved to a constant, I guess?

packages/story-editor/src/components/videoTrim/useVideoNode.js
```diff
@@ -0,0 +1,91 @@
...
+      setCurrentTime(0);
+      rawSetEndOffset(duration);
+      setMaxOffset(duration);
+    }
+    function onTimeUpdate(evt) {
+      const currentOffset = Math.floor(evt.target.currentTime * 1000);
+      setCurrentTime(currentOffset);
+      if (currentOffset >= endOffset) {
+        setTimeout(() => {
+          videoNode.currentTime = startOffset / 1000;
```
barklund: This converting back and forth between seconds and milliseconds is a bit annoying, but I feel it's cleaner to have our API in millis. Is it, though?

packages/story-editor/src/components/videoTrim/useVideoNode.js
```diff
@@ -0,0 +1,91 @@
...
+          videoNode.play();
+        });
+      }
+    }
+    videoNode.addEventListener('timeupdate', onTimeUpdate);
+    videoNode.addEventListener('loadedmetadata', onLoadedMetadata);
```
barklund, line 55: What happens if this event already occured? Could that happen?

packages/story-editor/src/elements/video/trim.js
```diff
@@ -59,38 +69,73 @@ function VideoTrim({ box: { width, height }, element }) {
...
+              ))}
+          </Video>
+        </MediaDisplay>
+      </Wrapper>
+      <VideoControls
+        isEditing={false}
+        isTransforming={false}
+        isSingleElement
+        isSelected
+        box={boxAtOrigin}
```
barklund: This is super ugly and should be cleaned up. Why set all these variables to some magic values just to get the component to appear? Much better to create an extra component and use that in the two places where relevant.

packages/story-editor/src/components/videoTrim/useVideoNode.js
```diff
@@ -0,0 +1,91 @@
...
+      rawSetStartOffset(0);
+      setCurrentTime(0);
+      rawSetEndOffset(duration);
+      setMaxOffset(duration);
+    }
+    function onTimeUpdate(evt) {
+      const currentOffset = Math.floor(evt.target.currentTime * 1000);
+      setCurrentTime(currentOffset);
+      if (currentOffset >= endOffset) {
+        setTimeout(() => {
```
barklund: `setTimeout` is required if the end offset is set to end of the video - if the time is just reset inline here, the video goes to the first frame and stops there.

No gap is noticeable though by using `setTimeout`, so I feel it's okay.

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
```diff
@@ -0,0 +1,68 @@
...
+
+  const canEnterTrimMode = useMemo(
+    () => selectedElement?.type === 'video' && isVideoTrimEnabled,
+    [selectedElement, isVideoTrimEnabled]
+  );
+
+  return {
+    isTrimMode: isEditing && isTrimming,
+    canEnterTrimMode,
+    toggleTrimMode,
```
barklund: Once either is merged, #8934 should use this context and these properties as well, of course.

packages/story-editor/src/components/videoTrim/useVideoNode.js
miina: Yep, def to a constant. Also, would be good to confirm this value, usually video trimming tools allow trimming it to much shorter as well, not sure if there's a reason to add this kind of restriction.

packages/story-editor/src/components/videoTrim/useVideoNode.js
miina: As opposed to floats and seconds? Milliseconds makes sense to me, should be a common practice.

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
```diff
@@ -111,36 +110,19 @@ function VideoOptionsPanel({ selectedElements, pushUpdate }) {
...
-    ({
-      state: { isEditing },
-      actions: { setEditingElementWithState, clearEditing },
-    }) => ({
-      isEditing,
-      setEditingElementWithState,
-      clearEditing,
+  const { canEnterTrimMode, toggleTrimMode } = useVideoTrim(
+    ({ state: { canEnterTrimMode }, actions: { toggleTrimMode } }) => ({
+      canEnterTrimMode,
```
miina: `hasTrimMode` or something similar might be a bit more descriptive. `canEnterTrimMode` feels as if the trim mode exists but currently can't be entered. Which is true for a video in terms of the feature flag but once the feature flag gets removed it'll be inaccurate. Or will `canEnterTrimMode` be removed together with the feature flag? Thoughts?

miina (approved):
🍰  LGTM, looking forward to the final trimming interface.
Would be good to confirm the 1000 ms restriction and get it to a constant before merging.

Btw -- this will also influence the auto-advancement logic, didn't see it mentioned under the tasks of the following Trim-related issues, let's add it as a Todo to the relevant issue if it's missing currently.

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
miina: Btw, is it needed at all, to have a separate boolean for if it has trim mode or not -- if it's anyway used for video only?

barklund: `hasTrimMode` would work as well. It also checks whether it's a single selection of a video element, so I feel it's warranted.

barklund:
> Btw -- this will also influence the auto-advancement logic, didn't see it mentioned under the tasks of the following Trim-related issues, let's add it as a Todo to the relevant issue if it's missing currently.

How will that logic be impacted? Once trimmed, this video will be completely replaced by another shorter video in the same place, so won't the advancement logic just use the duration of the new video instead?

miina:
> How will that logic be impacted? Once trimmed, this video will be completely replaced by another shorter video in the same place, so won't the advancement logic just use the duration of the new video instead?

Ah, right, in this case it doesn't influence it indeed :)

packages/story-editor/src/elements/video/trim.js
barklund: Done in 04d0731ea6.

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
barklund: Done in 40c48cd768

packages/story-editor/src/components/videoTrim/useVideoNode.js
barklund: Moved in d54d034279

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
barklund: Renamed in a8f3d09038

merapi (approved):
Works very well 👌🏻 

> setTimeout is required if the end offset is set to end of the video - if the time is just reset inline here, the video goes to the first frame and stops there.

Looks like a conflict with https://github.com/google/web-stories-wp/blob/833cc68bd99067cc2288f7026a32f2f0060ff9b8/packages/story-editor/src/elements/video/controls.js#L170:L174

If we change the above to:
```js
    const onVideoEnd = () => {
      if (!isTrimMode) {
        videoNode.currentTime = 0;
        setIsPlaying(false);
        setShowControls(true);
      }
    };
```
we can remove the `setTimeout` - not a big deal, but it will also fix this weird play/pause button behavior:

https://user-images.githubusercontent.com/21036224/132416268-d9364923-3398-4f0c-8d1b-407236e72bf8.mov


https://user-images.githubusercontent.com/21036224/132416283-6fd37fdb-1261-4004-bb74-5516b01d38d6.mov

packages/story-editor/src/components/videoTrim/useVideoNode.js
barklund: Fixed it in 1bece9f174 thanks to @merapi's [comment](https://github.com/google/web-stories-wp/pull/8944#pullrequestreview-748467708).

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
```diff
@@ -0,0 +1,68 @@
...
+      clearEditing();
+    } else {
+      setEditingElementWithState(selectedElement.id, {
+        isTrimming: true,
+      });
+    }
+  }, [isEditing, clearEditing, setEditingElementWithState, selectedElement]);
+
+  const hasTrimMode = useMemo(
+    () => selectedElement?.type === 'video' && isVideoTrimEnabled,
```
spacedmonkey: We need to make sure that user do not trim video if the video is current
- Transcoding
- Muting 
- Not local. 

https://github.com/google/web-stories-wp/pull/8972/files#diff-d581131a68dd353c626d1714ea3b4d65115979020ed22b61c7e5d4a0a8fed0ddR116-R123

barklund: Yeah, that makes sense. A non-local video can never be trimmed (of course) and a transcoding or muting video can only be trimmed once the current process is done (which will turn those flags false). I'll add that, thanks!

spacedmonkey: I added in my PR #8972 . When you implement this, do feel free to remove my logic and load it from `hasTrimMode`

barklund: Implemented in a8459cbc7d - would that do it? I also added that you couldn't trim a video while it's already trimming, which seems trivial as well.

spacedmonkey: I missed it off my list but I implemented it 

```
 return (
      isSingleElement &&
      isVideoTrimEnabled &&
      !local &&
      !isMuting &&
      !isTranscoding &&
      !isTrimming
    );
```

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
```diff
@@ -0,0 +1,83 @@
...
+import { useFeature } from 'flagged';
+import { useCallback, useMemo } from '@web-stories-wp/react';
+
+/**
+ * Internal dependencies
+ */
+import { useCanvas, useStory } from '../../app';
+
+function useVideoTrimMode() {
+  const isVideoTrimEnabled = useFeature('enableVideoTrim');
```
spacedmonkey, line 30: We could also check to see if user can transcode video. See `isTranscodingEnabled`. 

https://github.com/google/web-stories-wp/blob/77e354b4ff9f927c0cc15b80f88fd29df16b78e3/packages/story-editor/src/app/media/utils/useFFmpeg.js#L419-L429

No point of showing this functionality if you can transcode.

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
```diff
@@ -0,0 +1,83 @@
...
+        isTrimMode,
+        setEditingElementWithState,
+        clearEditing,
+      })
+    );
+  const { selectedElement } = useStory(({ state: { selectedElements } }) => ({
+    selectedElement: selectedElements.length === 1 ? selectedElements[0] : null,
+  }));
+  const { isTranscoding, isMuting, isTrimming, local } =
+    selectedElement.resource;
```
spacedmonkey:
```suggestion
    selectedElement?.resource;
```

spacedmonkey: I am getting a fatal error.

Screenshot 2021-09-08 at 18 17 32

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
```diff
@@ -0,0 +1,83 @@
...
+  }, [isEditing, clearEditing, setEditingElementWithState, selectedElement]);
+
+  const hasTrimMode = useMemo(
+    () =>
+      selectedElement?.type === 'video' &&
+      isVideoTrimEnabled &&
+      !isTranscoding &&
+      !isMuting &&
+      !isTrimming &&
+      local,
```
spacedmonkey: Local means local to file system, not to the editor. Once the editor upload the file, is not local anymore

spacedmonkey: Should there also be a check for if the video is from coverr?

barklund: Oh, wait, we also need to check that it's not a hotlinked video from an external source, no? Or can you also transcode those - and do they then become WP assets? Are we "allowed" to do that?

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
barklund: Wait, now I don't have the button anymore. Why would transcoding not be available on my system? Is it because I'm serving over HTTP rather than HTTPS (for the cross-origin isolation stuff)?

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
barklund: @miina, is there a way to recognize that a video is hotlinked from an external source?

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
barklund: Fixed in 8162d8f69a.

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
barklund: Ah yes, Coverr videos too, I guess.

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
barklund: Fixed in 8162d8f69a.

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
spacedmonkey: Would need to make sure that all that cases match. So user can upload, cross origin isloation is enable and user has opt-in into transcoding? Are all these cases true?

barklund: Fixed the issue with `window.crossOriginIsolated`, thanks!

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
barklund: Filed this as #8982 for follow-up.

PR #8945, opened 2021-09-03
Change tooltip placement if there's not enough space

### LINKED ISSUE #8822: Change tooltip placement if there's not enough space
## Feature Description

While working on #8374, it was noted that tooltips can be cut off if they're near the edge of the viewport:

Screenshot 2021-08-24 at 15 25 44

For cases like this, we should look into automatically changing the tooltip position if there's not enough space at the desired position. Just like we do for popups. In the example above, the tooltip could be added to the left instead.

BrookeGraham:
## Context
When tooltips are near the bottom of the viewport they can get cutoff. When this occurs we want to force the placement elsewhere to show the entire tooltip content. 

## Relevant Technical Choices

Created a function to dynamically place tooltips at the `TOP` position when they would otherwise be cutoff while in the `BOTTOM` position. I avoided moving the placement from either side as to not conflict between RTL and LTR languages. Horizontal placement is already been addressed via offsets, so this fix was solely for the `bottom` cutoff. 

## To-do

N/A
### Before
image
### After
image

## Testing Instructions

1. In the story editor select `Document` from the inspector. 
2. Add an audio file, and hover over the trash can icon.
3. The tooltip should appear above. 

No

No

No

## Checklist

---

Fixes #8822

### DIFF
--- a/packages/design-system/src/components/popup/index.js
+++ b/packages/design-system/src/components/popup/index.js
@@ -75,14 +75,13 @@ function Popup({
   fillWidth = false,
   fillHeight = false,
   onPositionUpdate = noop,
-  onPlacementUpdate = noop,
 }) {
   const [popupState, setPopupState] = useState(null);
   const [mounted, setMounted] = useState(false);
   const popup = useRef(null);
 
   const positionPopup = useCallback(
-    (evt) => {
+    async (evt) => {
       if (!mounted) {
         return;
       }
@@ -91,7 +90,8 @@ function Popup({
       if (evt?.target?.nodeType && popup.current?.contains(evt.target)) {
         return;
       }
-      setPopupState({
+
+      await setPopupState({
         offset:
           anchor?.current && getOffset(placement, spacing, anchor, dock, popup),
       });
@@ -112,8 +112,10 @@ function Popup({
     };
   }, [isOpen, positionPopup]);
 
-  useLayoutEffect(onPositionUpdate, [popupState, onPositionUpdate]);
-  useLayoutEffect(onPlacementUpdate, [popupState, onPlacementUpdate]);
+  useLayoutEffect(
+    () => onPositionUpdate(popupState),
+    [popupState, onPositionUpdate]
+  );
 
   useResizeEffect({ current: document.body }, positionPopup, [positionPopup]);
 
@@ -149,7 +151,6 @@ Popup.propTypes = {
   fillWidth: PropTypes.oneOfType([PropTypes.number, PropTypes.bool]),
   fillHeight: PropTypes.oneOfType([PropTypes.number, PropTypes.bool]),
   onPositionUpdate: PropTypes.func,
-  onPlacementUpdate: PropTypes.func,
 };
 
 export { Popup, PLACEMENT };
--- a/packages/design-system/src/components/popup/index.js
+++ b/packages/design-system/src/components/popup/index.js
@@ -75,6 +75,7 @@ function Popup({
   fillWidth = false,
   fillHeight = false,
   onPositionUpdate = noop,
+  onPlacementUpdate = noop,
 }) {
   const [popupState, setPopupState] = useState(null);
   const [mounted, setMounted] = useState(false);
@@ -112,6 +113,7 @@ function Popup({
   }, [isOpen, positionPopup]);
 
   useLayoutEffect(onPositionUpdate, [popupState, onPositionUpdate]);
+  useLayoutEffect(onPlacementUpdate, [popupState, onPlacementUpdate]);
 
   useResizeEffect({ current: document.body }, positionPopup, [positionPopup]);
 
@@ -147,6 +149,7 @@ Popup.propTypes = {
   fillWidth: PropTypes.oneOfType([PropTypes.number, PropTypes.bool]),
   fillHeight: PropTypes.oneOfType([PropTypes.number, PropTypes.bool]),
   onPositionUpdate: PropTypes.func,
+  onPlacementUpdate: PropTypes.func,
 };
 
 export { Popup, PLACEMENT };
--- a/packages/design-system/src/components/tooltip/index.js
+++ b/packages/design-system/src/components/tooltip/index.js
@@ -106,6 +106,7 @@ function Tooltip({
   const [arrowDelta, setArrowDelta] = useState(null);
   const anchorRef = useRef(null);
   const tooltipRef = useRef(null);
+  const dynamicPlacement = useRef(null);
 
   const spacing = useMemo(
     () => ({
@@ -133,6 +134,17 @@ function Tooltip({
 
     setArrowDelta(delta);
   }, []);
+  // When near the bottom of the viewport and the tooltip is placed on the bottom we want to force the tooltip to the top as to not
+  // cutoff the contents of the tooltip.
+  const positionPlacement = useCallback(() => {
+    const tooltipElBoundingBox = tooltipRef.current?.getBoundingClientRect();
+    if (
+      placement.startsWith('bottom') &&
+      tooltipElBoundingBox?.height < window.visualViewport.height
+    ) {
+      dynamicPlacement.current = PLACEMENT.TOP;
+    }
+  }, [placement]);
 
   const delay = useRef();
   const onHover = useCallback(
@@ -191,15 +203,16 @@ function Tooltip({
 
       <Popup
         anchor={forceAnchorRef || anchorRef}
-        placement={placement}
+        placement={dynamicPlacement.current || placement}
         spacing={spacing}
         isOpen={Boolean(shown && (shortcut || title))}
         onPositionUpdate={positionArrow}
+        onPlacementUpdate={positionPlacement}
       >
         <TooltipContainer
           className={className}
           ref={tooltipRef}
-          placement={placement}
+          placement={dynamicPlacement.current || placement}
           shown={shown}
           {...tooltipProps}
         >
@@ -216,7 +229,10 @@ function Tooltip({
                   <path d="M1,1 L0.868,1 C0.792,1,0.72,0.853,0.676,0.606 L0.585,0.098 C0.562,-0.033,0.513,-0.033,0.489,0.098 L0.399,0.606 C0.355,0.853,0.283,1,0.207,1 L0,1 L1,1" />
                 </clipPath>
               </SvgForTail>
-              <Tail placement={placement} translateX={arrowDelta} />
+              <Tail
+                placement={dynamicPlacement.current || placement}
+                translateX={arrowDelta}
+              />
             </>
           )}
         </TooltipContainer>
--- a/packages/design-system/src/components/tooltip/index.js
+++ b/packages/design-system/src/components/tooltip/index.js
@@ -19,7 +19,13 @@
  */
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
-import { useState, useRef, useMemo, useCallback } from '@web-stories-wp/react';
+import {
+  useState,
+  useRef,
+  useMemo,
+  useCallback,
+  useDebouncedCallback,
+} from '@web-stories-wp/react';
 
 /**
  * Internal dependencies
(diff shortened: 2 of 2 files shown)


packages/design-system/src/components/tooltip/index.js
```diff
@@ -133,6 +134,17 @@ function Tooltip({
...
   }, []);
+  // When near the bottom of the viewport and the tooltip is placed on the bottom we want to force the tooltip to the top as to not
+  // cutoff the contents of the tooltip.
+  const positionPlacement = useCallback(() => {
+    const tooltipElBoundingBox = tooltipRef.current?.getBoundingClientRect();
+    if (
+      placement.startsWith('bottom') &&
+      tooltipElBoundingBox?.height < window.visualViewport.height
+    ) {
+      dynamicPlacement.current = PLACEMENT.TOP;
```
BrookeGraham: I purposefully moved the position from bottom to top as to avoid any conflicts with RTL vs LTR. What do y'all think about that?

packages/design-system/src/components/tooltip/index.js
```diff
@@ -133,6 +134,17 @@ function Tooltip({
 
     setArrowDelta(delta);
   }, []);
+  // When near the bottom of the viewport and the tooltip is placed on the bottom we want to force the tooltip to the top as to not
+  // cutoff the contents of the tooltip.
+  const positionPlacement = useCallback(() => {
+    const tooltipElBoundingBox = tooltipRef.current?.getBoundingClientRect();
```
BrookeGraham: Since this needs to be in a callback in order to get the size of the tooltip, I am assuming I needed the [`useEffectLayout` in `Popup`](https://github.com/google/web-stories-wp/pull/8945/files#diff-ceecde430ba66db8a84fe3b7d8d7be935229b867156e208d0589e29364ee6d0aR152) Let me know if you think there is a better way to handle this.

packages/design-system/src/components/tooltip/index.js
```diff
@@ -133,6 +134,17 @@ function Tooltip({
...
+    if (
+      placement.startsWith('bottom') &&
+      tooltipElBoundingBox?.height < window.visualViewport.height
```
BrookeGraham: Oops, that is not what I wanted. TODO: fix this "math"

packages/design-system/src/components/tooltip/index.js
```diff
@@ -191,15 +203,16 @@ function Tooltip({
 
       <Popup
         anchor={forceAnchorRef || anchorRef}
-        placement={placement}
+        placement={dynamicPlacement.current || placement}
```
samwhale: I think this makes a lot of sense

samwhale (approved):
Making sense to me! Works great too

packages/design-system/src/components/tooltip/index.js
samwhale: Chatted with Brooke about this but putting here for visibility:

Only thing I would do here is maybe put it inside a check like
```
if (placement.startsWith('bottom') {
...
}
```

That way we don't run that every time if we don't need to.

packages/design-system/src/components/tooltip/index.js
```diff
@@ -133,6 +134,17 @@ function Tooltip({
...
   }, []);
+  // When near the bottom of the viewport and the tooltip is placed on the bottom we want to force the tooltip to the top as to not
+  // cutoff the contents of the tooltip.
+  const positionPlacement = useCallback(() => {
+    const tooltipElBoundingBox = tooltipRef.current?.getBoundingClientRect();
+    if (
+      placement.startsWith('bottom') &&
+      tooltipElBoundingBox?.bottom > window.visualViewport.height
+    ) {
+      dynamicPlacement.current = PLACEMENT.TOP;
```
samwhale: In the future we may want to do this for placement 'top' tooltips too. For now I don't think that's necessary though.

BrittanyIRL (commented):
i like this alot! thanks for diving in on your first week into this bug!

packages/design-system/src/components/tooltip/index.js
```diff
@@ -133,6 +134,17 @@ function Tooltip({
 
     setArrowDelta(delta);
   }, []);
+  // When near the bottom of the viewport and the tooltip is placed on the bottom we want to force the tooltip to the top as to not
+  // cutoff the contents of the tooltip.
+  const positionPlacement = useCallback(() => {
+    const tooltipElBoundingBox = tooltipRef.current?.getBoundingClientRect();
+    if (
+      placement.startsWith('bottom') &&
+      tooltipElBoundingBox?.bottom > window.visualViewport.height
```
maxyinger: out of curiosity, why'd you use `window.visualViewport` here over `window.innerHeight` ? just never seen `visualViewport` before is why I ask

packages/design-system/src/components/tooltip/index.js
maxyinger: So I just learned this (thanks to Will), but I believe you can actually pass a callback ref to get this data:
https://reactjs.org/docs/refs-and-the-dom.html#callback-refs
https://medium.com/@teh_builder/ref-objects-inside-useeffect-hooks-eb7c15198780

I'm sure this works fine as is tho too

packages/design-system/src/components/tooltip/index.js
```diff
@@ -106,6 +106,7 @@ function Tooltip({
   const [arrowDelta, setArrowDelta] = useState(null);
   const anchorRef = useRef(null);
   const tooltipRef = useRef(null);
+  const dynamicPlacement = useRef(null);
```
maxyinger: What's the rationale for using a ref here instead of state? I guess I only worry that this can update and not cause a rerender causing `TooltipContainer` to take a stale `placement` prop

maxyinger (commented):
looks good!

packages/design-system/src/components/tooltip/index.js
BrookeGraham: Oh yes, I remember seeing this! That must be why I felt a little iffy on my implementation. 🙇🏻‍♀️

packages/design-system/src/components/tooltip/index.js
BrookeGraham: Honestly, I don't have a "good" reason. And after looking at the docs, `innerHeight` is probably best since explorer doesn't support `visualViewport`

packages/design-system/src/components/tooltip/index.js
BrookeGraham: I didn't want `Tooltip` to _have_ to rerender when the `dynamicPlacement` isn't necessary, and since `Popup` and `ToolTipContainer` will rerender when the `placement` prop changes anyway, it would always getting the updated / correct value for `placement`. At least I think.

packages/design-system/src/components/tooltip/index.js
BrookeGraham: I did find this info though. https://developers.google.com/web/updates/2017/09/visual-viewport-api

packages/design-system/src/components/tooltip/index.js
BrookeGraham: hmm you are correct, I am getting a stale state for placement. I update to use  `useState` 🙌🏽

packages/design-system/src/components/tooltip/index.js
swissspidy: We don't support IE, FWIW

swissspidy: Oh, jsdom doesn't mock `visualViewport`, so JS unit tests are currently failing because of that. Would need to use some optional chaining / fallback for that. Or just use `innerWidth`, which jsdom provides.

packages/design-system/src/components/tooltip/index.js
```diff
@@ -106,6 +112,8 @@ function Tooltip({
   const [arrowDelta, setArrowDelta] = useState(null);
   const anchorRef = useRef(null);
   const tooltipRef = useRef(null);
+  const placementRef = useRef(placement);
+  const [realPlacement, setRealPlacement] = useState(placement);
```
BrittanyIRL: Just noodlin' - this naming makes sense to you? I was throwing a bunch of stuff at the wall to see where changes occurred, thus the origin of 'realPlacement'. When you read through it, does this make sense or should it be something different?

packages/design-system/src/components/tooltip/index.js
```diff
@@ -121,18 +129,46 @@ function Tooltip({
...
+  // When near the bottom of the viewport and the tooltip is placed on the bottom we want to force the tooltip to the top as to not
+  // cutoff the contents of the tooltip.
+  const positionPlacement = useCallback(
+    ({ offset }) => {
+      if (offset) {
+        const neededVerticalSpace = offset.y + offset.height;
+        // if the tooltip was assigned as bottom we want to always check it
+        if (realPlacement.startsWith('bottom')) {
+          // check to see if there's an overlap with the window edge
+          if (neededVerticalSpace >= window.innerHeight) {
```
BrittanyIRL: This is my b, i didn't clean up my wip :) these 3 checks could be combined. 

I'd been looking to see if a reset needed to occur elsewhere in this step process, so i'd separated them out. Now that it's all solved it's unnecessary.

packages/design-system/src/components/tooltip/index.js
```diff
@@ -160,14 +196,15 @@ function Tooltip({
     (evt) => {
       setShown(false);
       onPointerLeave(evt);
+      resetPlacement();
```
BrittanyIRL, line 198: the other option for where to put this/structure is a use effect when shown becomes false. The reason I'd not done that before is so that it's not called on initial set up of allll the tooltips. Thoughts?

BrittanyIRL (approved):
This works great! Nicely done. 

Few little questions which are mostly my fault for work in process musing.

packages/design-system/src/components/tooltip/index.js
BrookeGraham: hahaha, I saw it and giggled to myself. I will update now. I think I liked `dynamicPlacement`

packages/design-system/src/components/tooltip/index.js
BrookeGraham: Totally.

packages/design-system/src/components/tooltip/index.js
BrookeGraham: I think this is fine. Saves us that initial reset call, and calling it here (and onBlur) seems more explicit.

packages/design-system/src/components/tooltip/index.js
```diff
@@ -113,7 +113,7 @@ function Tooltip({
   const anchorRef = useRef(null);
   const tooltipRef = useRef(null);
   const placementRef = useRef(placement);
-  const [realPlacement, setRealPlacement] = useState(placement);
+  const [dynamicPlacement, setDynamicPlacement] = useState(placement);
```
BrittanyIRL, line 116: 💃

packages/design-system/src/components/popup/index.js
```diff
@@ -90,7 +90,8 @@ function Popup({
       if (evt?.target?.nodeType && popup.current?.contains(evt.target)) {
         return;
       }
-      setPopupState({
+
+      await setPopupState({
         offset:
           anchor?.current && getOffset(placement, spacing, anchor, dock, popup),
       });
```
samwhale, line 97: 🤔 Does this need to be `await`ed? I don't think I've seen state setters be awaited before

packages/design-system/src/components/tooltip/index.js
```diff
@@ -121,18 +129,45 @@ function Tooltip({
...
-    const anchorElBoundingBox = anchorRef.current?.getBoundingClientRect();
-    const tooltipElBoundingBox = tooltipRef.current?.getBoundingClientRect();
-    if (!tooltipElBoundingBox || !anchorElBoundingBox) {
-      return;
-    }
-    const delta =
-      getBoundingBoxCenter(anchorElBoundingBox) -
-      getBoundingBoxCenter(tooltipElBoundingBox);
+  // When near the bottom of the viewport and the tooltip is placed on the bottom we want to force the tooltip to the top as to not
+  // cutoff the contents of the tooltip.
```
samwhale, line 133: Nice comments 👍

samwhale (approved):
Nice this works great 🥳 Left you a question

packages/design-system/src/components/popup/index.js
BrookeGraham: We were running into race condition issues with another `useLayoutEffect` but I think since we removed it, we probably don't need this anymore.

Issue #5108, opened 2020-11-02
Editor: Cannot use multiple publisher logos


swissspidy:
## Bug Description

It seems like the publisher logo section in the Document Panel does not work as expected. Maybe the spec is unclear or we never fully implemented this.

[The spec currently says](https://docs.google.com/document/d/1GJZljicdXcP7oRdywsF3EGUaCVpCHsDrK2gUMubnrbk/edit?pli=1#heading=h.ymf85pwospli):

> Allows the author to pick a publisher logo, as defined in AMP Stories’ metadata guidelines.
> 
> Inherited from global plugin settings, but can be overridden for the story at hand locally.

**[UX / Design specs](https://www.figma.com/file/Cd9iacVknHK62RO4pRwb68/Stories-Editor?node-id=5837%3A4)**

## Expected Behaviour

I should be able to use different publisher logos for different stories.

## Steps to Reproduce

1. Create story 1, select publisher logo A.
1. Publish story 1.
1. See how logo A is selected as the default in the dashboard settings (so far so good)
1. Create story 2, select publisher logo B.
1. Publish story 2.
1. See how logo B is selected as the default in the dashboard settings (so far so good)
1. Edit story 1
1. See logo B in the document panel instead of A (note: the saved story still contains logo A)
1. Save story 1, now the saved markup contains logo B

## Additional Context

- Plugin Version: 1.1.0
- Operating System: macOS
- Browser: Chrome 86;

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

swissspidy:
@o-fernandez I think this might need your clarification on how things should work

o-fernandez:
Yeah, Sam had a design for this that had a drop-down where you select the publisher logo, and it would show you all the configured publisher logos. It'd also had a link/CTA to the "settings" page for management of the logos. 

This is completely missing from the editor, so will move this to the UX needed column so we can attach the designs and add it to the workspace queue.

swissspidy:
Is there any update on this?

swissspidy:
Another user request: https://wordpress.org/support/topic/select-different-publisher-logo-for-each-story/

@o-fernandez Do you remember where that UX was left? I think even without this we need to somehow fix the issue where changing the publisher logo in the editor changes the default logo for _all_ stories.

o-fernandez:
Hm... I don't. @aaskedall did you get any UX from Sam on selecting a publisher logo from within the story editor? 
[...]

PR #8815, opened 2021-08-26
Remove featured_media_src field

spacedmonkey:
## Testing Instructions

- [ ] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1.

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [x] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [x] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #8790

PR #8946, opened 2021-09-03
Store publisher logo as post meta

### LINKED ISSUE #7846: Convert Publisher PHP trait to class
## Feature Description

The Publisher trait is used in 5 classes. Converting this to a class should be pretty simple. See https://github.com/google/web-stories-wp/issues/7844 for more detail. 

## Additional Context

---

swissspidy:
## Context

For #5108 we need to store the publisher logo as post meta of a story.

In addition to that, we need to stop updating the global active/default publisher logo whenever you update a story.

## Summary

This PR changes the whole publisher logo usage so that the publisher logo is stored as post meta for each story.

When a story doesn't have a publisher logo yet, it falls back to the global active/default publisher logo from the settings.

## User-facing changes

N/A

## Testing Instructions

1.

No

No

No

## Checklist

---

See #5108
closes #7846


includes/Traits/Publisher.php
```diff
@@ -43,74 +43,18 @@ trait Publisher {
...
-		 *
-		 * This should point to a square image.
-		 *
-		 * @since 1.0.0
-		 * @since 1.1.0 The $placeholder parameter was deprecated.
-		 *
-		 * @param string|null $logo_image_url URL to the publisher's logo if set.
-		 * @param string      $placeholder    Deprecated.
-		 */
-		return apply_filters( 'web_stories_publisher_logo', $logo_image_url, $placeholder );
```
miina, line 103: @swissspidy What about the cases where the filter was in use, wouldn't removing this filter be a breaking change? Should the filter be moved to `$sanitizers[ Story_Sanitizer::class ]` then perhaps as a wrapper around the sanitizer? Thoughts?

swissspidy: Yeah perhaps, though it would probably make more sense to have a new filter that passes the current story / post object as context.

includes/Story_Post_Type.php
```diff
@@ -132,6 +140,21 @@ public function register() {
 			]
 		);
 
+		$active_publisher_logo_id = absint( get_option( Settings::SETTING_NAME_ACTIVE_PUBLISHER_LOGO ) );
+
+		register_post_meta(
+			self::POST_TYPE_SLUG,
+			self::PUBLISHER_LOGO_META_KEY,
+			[
+				'sanitize_callback' => 'absint',
```
spacedmonkey, line 149: Maybe we could add a validation callback, to check if the id provided is a valid attachment id.

includes/Story_Post_Type.php
```diff
@@ -132,6 +140,21 @@ public function register() {
 			]
 		);
 
+		$active_publisher_logo_id = absint( get_option( Settings::SETTING_NAME_ACTIVE_PUBLISHER_LOGO ) );
+
+		register_post_meta(
```
spacedmonkey, line 145: I think this code could live in the media namespace. Similar to muting 

https://github.com/google/web-stories-wp/blob/main/includes/Media/Video/Muting.php

swissspidy: That namespace doesn't seem fitting at all. The `Media` namespace contains all sorts of code that for media items / attachments: video muting, image sizes, SVG upload, etc.

This code, on the other hand, simply registers some post meta for the story post type holding an integer which happens to be an attachment ID.

packages/story-editor/src/app/story/effects/useLoadStory.js
```diff
@@ -111,6 +109,22 @@ function useLoadStory({ storyId, shouldLoad, restore, isDemo }) {
...
+
+        if ('wp:publisherlogo' in embedded) {
+          publisherLogo = {
+            id: embedded['wp:publisherlogo'][0].id,
+            height: embedded['wp:publisherlogo'][0].media_details?.height,
+            width: embedded['wp:publisherlogo'][0].media_details?.width,
+            url: embedded['wp:publisherlogo'][0].source_url,
+          };
+        }
+
```
spacedmonkey, line 102: There is some repeated code here. How about a function like this?

```
function getEmbeddedAttachment( field, embedded ) {
        if (field in embedded) {
          return {
            id: embedded[field][0].id,
            height: embedded[field][0].media_details?.height,
            width: embedded[field][0].media_details?.width,
            url: embedded[field][0].source_url,
          };
        }
        return {
          id: 0,
          height: 0,
          width: 0,
          url: '',
        };
}
```

Could also be used like in this https://github.com/google/web-stories-wp/pull/8815

includes/Story_Post_Type.php
swissspidy: I thought about this too but it felt like very defensive programming. We don't do this on other places either (e.g. original video ID when optimizing videos)

packages/story-editor/src/app/story/effects/useLoadStory.js
swissspidy: With just 2 such fields I don't think it's worth extracting this into a function at this point. I'd consider it when there should be more fields, when the benefit would be bigger.

merapi (approved):
Tested it with my dropdown - in terms of what's needed for #5108 it's all good 👍🏻

includes/Traits/Publisher.php
miina: Ok, just want to make sure it wouldn't break for existing usages of the filter -- probably wouldn't be nice to just remove the existing filter unexpectedly.

packages/story-editor/src/app/story/effects/useLoadStory.js
spacedmonkey: It could be used [here](https://github.com/google/web-stories-wp/pull/8815/files#diff-6eac1ab8374e1da9ea0278131b7244d9d7e8e879c544d282a38c278453cdbe75R86-R99) and [here](https://github.com/google/web-stories-wp/pull/8815/files#diff-6eac1ab8374e1da9ea0278131b7244d9d7e8e879c544d282a38c278453cdbe75R137-R150). 

So the function is cleaner,as there is no need for a let.

includes/REST_API/Stories_Controller.php
```diff
@@ -484,6 +464,16 @@ protected function prepare_links( $post ): array {
 			}
 		}
 
+		// Publisher Logo.
+		$publisher_logo_id = get_post_meta( $post->ID, Story_Post_Type::PUBLISHER_LOGO_META_KEY, true );
+
+		if ( $publisher_logo_id ) {
+			$links['https://api.w.org/publisherlogo'] = [
+				'href'       => rest_url( sprintf( '%s/%s', $this->namespace, 'media/' ) . $publisher_logo_id ),
```
spacedmonkey:
```suggestion
				'href'       => rest_url( sprintf( '%s/%s/%s', $this->namespace, 'media', $publisher_logo_id ) ),
```

includes/Traits/Publisher.php
spacedmonkey: I agree, this filter dont not be removed, it should be maintained.

swissspidy: I'll take another look

packages/story-editor/src/app/story/effects/useLoadStory.js
spacedmonkey: Also could be used [here](https://github.com/google/web-stories-wp/blob/610bbea66f7e0d4361956cab79f279f70d0e3c1d/packages/dashboard/src/app/serializers/stories.js#L45)  and maybe even [here](https://github.com/google/web-stories-wp/blob/610bbea66f7e0d4361956cab79f279f70d0e3c1d/packages/stories-block/src/block/components/storiesPreview.js#L86)

swissspidy: I think just using optional chaining would do the trick just as well without hiding anything of this logic behind a function living in some other package.

```js
const publisherLogo = {
	  id: embedded?.['wp:publisherlogo']?.[0].id || 0,
	  height: embedded?.['wp:publisherlogo']?.[0]?.media_details?.height || 0,
	  width: embedded?.['wp:publisherlogo']?.[0]?.media_details?.width || 0,
	  url: embedded?.['wp:publisherlogo']?.[0]?.source_url || '',
};
```

includes/REST_API/Stories_Controller.php
```diff
@@ -459,6 +437,8 @@ protected function prepare_links( $post ): array {
 		$links = parent::prepare_links( $post );
 		add_post_type_support( Story_Post_Type::POST_TYPE_SLUG, 'revisions' );
 
+		// Post Locking.
+
```
spacedmonkey, line 459:
```suggestion
```

swissspidy: Hmm I think I'll actually just extract the whole post locking part here into its own method and remove this comment entirely.

includes/Story_Post_Type.php
```diff
@@ -132,6 +140,21 @@ public function register() {
...
+			self::POST_TYPE_SLUG,
+			self::PUBLISHER_LOGO_META_KEY,
+			[
+				'sanitize_callback' => 'absint',
+				'type'              => 'integer',
+				'description'       => __( 'Publisher logo ID.', 'web-stories' ),
+				'show_in_rest'      => true,
+				'default'           => $active_publisher_logo_id,
```
spacedmonkey, line 153: This worked great in my testing!

swissspidy:
There's a failing e2e test that I need to debug, but apart from that this should be finished.

spacedmonkey:
Lints are failing

swissspidy:
> Lints are failing

@spacedmonkey Thanks, appreciate the eyes. I am still working on this and will ping you once it's ready.

swissspidy:
@spacedmonkey this is now ready for review

There is one unrelated failing (flakey?) Karma test:

```
1) should not bubble up arrow keys from library tabs
     Canvas keys integration
     Error: Timed out in waitFor.
```

If it keeps failing it needs to be disabled in a separate PR & fixed in a dedicated ticket, as per our usual agreed upon process

includes/Traits/Publisher.php
```diff
@@ -43,74 +43,18 @@ trait Publisher {
...
-
-		if ( $active_publisher_logo ) {
-			$logo_image_url = wp_get_attachment_image_url( $active_publisher_logo, Image_Sizes::PUBLISHER_LOGO_IMAGE_SIZE );
-
-			if ( ! $logo_image_url ) {
-				$logo_image_url = null;
-			}
-		}
-
-		$placeholder = $this->get_publisher_logo_placeholder();
```
spacedmonkey, line 90: Let's remove get_publisher_logo_placeholder method as well.

includes/Integrations/AMP.php
```diff
@@ -162,8 +163,15 @@ public function add_amp_content_sanitizers( $sanitizers ) {
 
 		$story = new Story();
 		$story->load_from_post( $post );
+
+		$publisher_logo_id  = get_post_meta( $post->ID, Story_Post_Type::PUBLISHER_LOGO_META_KEY, true );
+		$publisher_logo_url = wp_get_attachment_image_url( $publisher_logo_id, Image_Sizes::PUBLISHER_LOGO_IMAGE_SIZE );
+
+		/* This filter is documented in includes/Discovery.php */
+		$publisher_logo_url = apply_filters( 'web_stories_publisher_logo', $publisher_logo_url, $post );
```
spacedmonkey: This logic is repeated a couple of times. Let's create a method in the publisher trait. Something like this. 

```
   public function get_story_publisher_logo( WP_Post $post ){
          $publisher_logo_id  = get_post_meta( $post->ID, Story_Post_Type::PUBLISHER_LOGO_META_KEY, true );
		  $publisher_logo_url = wp_get_attachment_image_url( $publisher_logo_id, Image_Sizes::PUBLISHER_LOGO_IMAGE_SIZE );
          /**
		  * Filters the publisher logo URL.
		  *
		  * @since 1.0.0
		  * @since 1.1.0 The second parameter was deprecated.
		  * @since 1.11.0 The second parameter was repurposed to provide the current post object.
		  *
		  * @param string|null $logo_image_url Publisher logo URL.
		  * @param WP_Post     $post           Current post object.
		 */
		return apply_filters( 'web_stories_publisher_logo', $url, $post );
  }
```

spacedmonkey: Would be specially usefully https://github.com/google/web-stories-wp/issues/7846

packages/story-editor/src/app/api/apiProvider.js
```diff
@@ -83,7 +82,7 @@ function APIProvider({ children }) {
     (storyId) => {
       const path = addQueryArgs(`${stories}${storyId}/`, {
         context: 'edit',
-        _embed: 'wp:featuredmedia,wp:lockuser,author',
+        _embed: 'wp:featuredmedia,wp:lockuser,author,wp:publisherlogo',
```
spacedmonkey: Should this be a const,as it is repeated.

includes/Traits/Publisher.php
swissspidy: That one is still used in `Story_Sanitizer`/`Sanitization_Utils`

includes/Integrations/AMP.php
swissspidy: It's not always the same usage. For example, in `Discovery.php` the publisher logo width & height are used as well. The URL alone is needed in 2 places outside of that.
IMO not a big deal if that's repeated like that.

spacedmonkey: But all the publisher replated code is meant to in that one place. Why remove it out, making the code less readable and discovable?

spacedmonkey: Also of note here, why `deprecated` the placeholder? Why not just add a new param like WordPress core does? There is a backward compatibility issue here.

spacedmonkey: Also think that code that is reused twice, should have a unit test.

swissspidy: The second param was never really used and removed a long time ago. We don't need to hold on to that forever like core does.

swissspidy: Due to different usages that would probably mean adding multiple different methods (`get_publisher_logo_id( $post )`, `get_publisher_logo_id( $post )`, `get_publisher_logo_url( $post )`, `get_publisher_logo_src( $post )`), making this a lot more complex and less comprehensible imo.

What about moving this code to the `Story` model instead? Since the logo is tied to a story anyway.

spacedmonkey: > What about moving this code to the Story model instead? Since the logo is tied to a story anyway.

Love this idea! Can't believe I didn't think of it is!

So two new properties? 
`publisher_logo_url` - string / url
`publisher_logo_size` - array, [width, height]

includes/Discovery.php
```diff
@@ -163,6 +158,31 @@ protected function get_schemaorg_metadata(): array {
 		$post = get_queried_object();
 
 		if ( $post instanceof WP_Post ) {
+			$publisher_logo_id = get_post_meta( $post->ID, Story_Post_Type::PUBLISHER_LOGO_META_KEY, true );
```
spacedmonkey: If we use a story model here, it means we can load other fields from the story model, like 
- title
- url
- poster ( maybe ).

swissspidy: tried this now

includes/Integrations/AMP.php
swissspidy: tried this now

spacedmonkey:
@googlebot I consent.

spacedmonkey (approved):
I made some small tweaks to tidy up, but I am happy with this PR now!

Love we removed the publisher trait!

PR #8947, opened 2021-09-03
Color Picker: use '000000' instead of 'transparent'.

### LINKED ISSUE #7896: Transparency filter shows the word "#transparent" in the filter control
## Bug Description

When you set apply a filter to an object, and open the color picker to to set the gradient curves, and colour picker is in the bottom left color #00000 and the opacity is set to 100% (ie transparent), it displays the word "transparent" instead of showing the color hex code

## Expected Behaviour

Should probably just still show the color hex code regardless

## Steps to Reproduce

1. Select an object in the canvas that is eligible to have a filter applied
2. Select the Linear filter
3. open the color picker to set the curves for the gradient
4. ensure that the color selected is bottom left corner ie: #000000
5. set the transparency slider to 0%

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

---

miina:
## Summary
Uses '000000' as the display value instead of `transparent` which is used by default by the color picker library.

## Testing Instructions
1. Add a shape.
2. Open the color picker and go to the Custom Color picker.
3. Set the color to black ('000000')
4. Now set the opacity to 0.
5. Verify that '000000' is displayed in the hex input of the Color Picker. Click on the value to enter edit mode of the input and verify it displays '000000' there, too.

## Checklist

---

Fixes #7896

### DIFF
--- a/packages/story-editor/src/components/colorPicker/editablePreview.js
+++ b/packages/story-editor/src/components/colorPicker/editablePreview.js
@@ -131,11 +131,14 @@ function EditablePreview({ label, value, width, format, onChange }) {
     );
   }
 
+  // The value is set to 'transparent' by the library if it's black with 0 opacity.
+  // We want to always display hex though.
+  const editValue = value === 'transparent' ? '000000' : value;
   return (
     <Wrapper ref={wrapperRef} tabIndex={-1} onBlur={handleOnBlur}>
       <Suspense fallback={null}>
         <EditableInput
-          value={value}
+          value={editValue}
           ref={editableRef}
           onChange={onChange}
           onChangeComplete={disableEditing}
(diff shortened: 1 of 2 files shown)


packages/story-editor/src/components/colorPicker/editablePreview.js
```diff
@@ -131,11 +131,14 @@ function EditablePreview({ label, value, width, format, onChange }) {
     );
   }
 
+  // The value is set to 'transparent' by the library if it's black with 0 opacity.
```
barklund, line 134: Is that the underlying `react-color` library, that does this? Is that a breaking change?

barklund (approved):

packages/story-editor/src/components/colorPicker/editablePreview.js
miina: Yep, that's the underlying `react-color`, it displays `transparent` by default if it's `000000` and opacity `0`. Shouldn't be a breaking change, all seems to work well. If that's what you mean :)

PR #8513, opened 2021-07-30
Implement dependency resolution for services

swissspidy:
As taken from https://github.com/ampproject/amp-wp/pull/6484

Are there any dependencies we'd need to add already?

PR #8949, opened 2021-09-03
Do not delay `Database_Upgrader` service

spacedmonkey:
## Summary

Improvement to https://github.com/google/web-stories-wp/pull/8792. Do not delay the service and hook into admin_init on register.

## Testing Instructions

1.

## Checklist

---

Fixes #

### DIFF
--- a/includes/Database_Upgrader.php
+++ b/includes/Database_Upgrader.php
@@ -29,7 +29,7 @@ namespace Google\Web_Stories;
 use Google\Web_Stories\Infrastructure\PluginActivationAware;
 use Google\Web_Stories\Infrastructure\Injector;
 use Google\Web_Stories\Infrastructure\Service;
-use Google\Web_Stories\Infrastructure\Conditional;
+use Google\Web_Stories\Infrastructure\Registerable;
 use Google\Web_Stories\Infrastructure\SiteInitializationAware;
 use WP_Site;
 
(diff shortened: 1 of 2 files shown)


includes/Database_Upgrader.php
```diff
@@ -104,20 +104,7 @@ public function __construct( Injector $injector ) {
...
-	/**
-	 * Get the action priority to use for registering the service.
-	 *
-	 * @since 1.6.0
-	 *
-	 * @return int Registration action priority to use.
-	 */
-	public static function get_registration_action_priority(): int {
-		return 5;
+		add_action( 'admin_init', [ $this, 'run_upgrades' ], 5 );
```
swissspidy, line 107: Is there still a good reason for priority 5?

includes/Database_Upgrader.php
```diff
@@ -104,20 +104,7 @@ public function __construct( Injector $injector ) {
 	 * @return void
 	 */
 	public function register() {
```
swissspidy, line 106: Note that tests need updating now and are failing.

includes/Database_Upgrader.php
spacedmonkey: It used to be set to 5.

swissspidy: I know, but is it worth reconsidering this?

With #8513 we have other ways to ensure that the database upgrade runs before anything else.

spacedmonkey: More than happy to have to default of 10.

includes/Database_Upgrader.php
spacedmonkey: Tests passing.

swissspidy (approved):

Issue #2918, opened 2020-07-01
Editor Decoupling


o-fernandez:
## Summary

Current proposal by @sayedtaqui for doing the decoupling in the following four steps.

### 1. Move all WordPress-related code from `story-editor` to `wp-story-editor`.

As [story-editor](https://github.com/google/web-stories-wp/tree/main/packages/story-editor) package is supposed to be the core editor, it should be free from any WordPress-related code so that it can easily be integrated with any third-party platform. In this step, we would only remove any WordPress-specific code from the `story-editor` package and this task can be broken down into the following sub-tasks.

- [x] Move all REST API requests logic to [wp-story-editor](https://github.com/google/web-stories-wp/tree/main/packages/wp-story-editor) and use more platform-agnostic names for WordPress-specific HTTP headers like `X-WP-*` used in REST requests #8966 , #9029
- [x] Move [`PostLock`](https://github.com/google/web-stories-wp/tree/2f4e6a3b76b4ad38bc722f99a09235e13434e5cb/packages/story-editor/src/components/postLock) and [`StatusCheck`](https://github.com/google/web-stories-wp/tree/2f4e6a3b76b4ad38bc722f99a09235e13434e5cb/packages/story-editor/src/components/statusCheck) components as they are WordPress specific. (#8990)
- [x] Move the “Story Published” dialog to `wp-story-editor`. (#8951)
- [x] Move WordPress media modal logic ( `wp.media` ) to `wp-story-editor`. (#9066)
- [x] Move [meta boxes](https://github.com/google/web-stories-wp/pull/5048) to `wp-story-editor`. (#9027)

### 2. Move all WordPress-related code from `dashboard` to `wp-dashboard`. (#9027)

Since the [dashboard](https://github.com/google/web-stories-wp/tree/main/packages/dashboard) package has a similar code pattern to some extent, our learnings from step 1 would be helpful to move WordPress-specific code to the [wp-dashboard](https://github.com/google/web-stories-wp/tree/main/packages/wp-dashboard) package. It may need the following work.

- [x] Refactor components to ease future decoupling (#9202 )
- [x] REST API: Refactor `ApiProvider` component to allow api callbacks via wp-dashboard configuration (#9237)
- [x] REST API: Make the expected API responses more platform-agnostic and move functions like `reshapeStoryObject` ( #9345 )
- [x] Move settings page ( analytics, video optimization, etc ) to `wp-dashboard` ( #9419 )
[...]

Issue #8388, opened 2021-07-20
Packages: Move `api-fetch` usage to new shared package


spacedmonkey:
## Feature Description

* [ ] Move some of the code from `/assets/src/edit-story/utils/base64Encode.js` to `/packages/api-fetch/src/`
* [ ] Move some of the code from `/packages/design-system/src/utils/addQueryArgs.js` to `/packages/api/src/` + tests
* [ ] `export * from '@wordpress/api-fetch';`
* [ ] Update imports where necessary.  
  Use `@web-stories-wp/...` imports for anything that is already a package, relative imports for everything else.
* [ ] Create `/packages/api-fetch/package.json` referencing its dependencies
* [ ] Create `/packages/api-fetch/README.md` with short documentation.

## Additional Context

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

spacedmonkey:
Adding this new package, adds a new number of benefits. 

1. `base64Encode.js` will not longer be loaded from editor in dashboard. 
2. `addQueryArgs` is moved to a more sensible place out of the design system. 
3. Neither the editor / dashboard will have direct dependencies on a WordPress package. 
4. Updating apiFetch package is done in one place. 

Thoughts @swissspidy ?

swissspidy:
With the [decoupling work](https://docs.google.com/document/d/1jTCydhOSl4HGK6PNA3io_Vmsm0wBCCUj0aW8RZXtzz8/edit) we'll end up with at least the following [packages](https://docs.google.com/document/d/1ImRe9wYKWGCAZiq3jro5KLpSJK_NMKXHC6FWqs3nwl0/edit):

* `dashboard`
* `wp-dashboard`
* `story-editor` (think `@wordpress/block-editor` in Gutenberg)
* `wp-story-editor` (think `@wordpress/edit-post` in Gutenberg)

There will probably be a fifth one with shared utilities used by both `wp-` packages, e.g. `wp-story-utils` or similar.

`addQueryArgs`, `@wordpress/api-fetch` usage, `base64Encode`, `getResourceFromAttachment`, etc. would all live in that package.

`story-editor` and `dashboard` would never use those directly. Only `wp-dashboard` and `wp-story-editor` would.

So I'd rather name this `wp-story-utils` or similar, not `api-fetch`.

spacedmonkey:
[...]

PR #8966, opened 2021-09-07
Decoupling: Move all REST API calls to `wp-story-editor`

mohdsayed:
## Context

Move REST API calls to `wp-story-editor` and remove `@wordpress/api-fetch` package dependency from the core story editor.

## To-do

- [x] Create callbacks for all REST API calls and move to wp-story-editor.
- [x] Pass API callbacks back to `wp-story-editor` via config and use in `APIProvider` actions
- [x] Move relevant test cases.
- [x] Move `flattenFormData` to `wp-story-editor`
- ~Handle cases when some API callbacks have not been provided via configuration.~

## Testing Instructions

1.

## Checklist

---


packages/wp-story-editor/src/api/media.js
```diff
@@ -0,0 +1,136 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * External dependencies
+ */
+import { addQueryArgs } from '@web-stories-wp/design-system';
+import { flattenFormData } from '@web-stories-wp/story-editor';
```
swissspidy: Just move this to `wp-story-editor`?

packages/wp-story-editor/src/api/test/api.js
```diff
@@ -0,0 +1,88 @@
...
+
+    apiFetch.mockReturnValue(
+      Promise.resolve({
+        body: GET_MEDIA_RESPONSE_BODY,
+        headers: GET_MEDIA_RESPONSE_HEADER,
+      })
+    );
+  });
+
+  it('getMedia with cacheBust:true should call api with &cache_bust=true', () => {
```
swissspidy, line 41: Since you moved this to `wp-story-editor`, I'd just test the `getMedia` callback directly instead of going through the `APIProvider`.

tests/phpunit/integration/data/story_post_content_filtered.json
```diff
@@ -1,153 +1,153 @@
 {
-  "version": 21,
-  "pages": [
-    {
-      "elements": [
+    "version": 21,
```
mohdsayed: Not sure how this change got pulled in the merge conflict. Will have to reset.

packages/wp-story-editor/src/index.js
```diff
@@ -35,8 +35,9 @@ import { initializeTracking } from '@web-stories-wp/tracking';
 /**
  * Internal dependencies
  */
-import './style.css'; // This way the general editor styles are loaded before all the component styles.
+import * as apiCallbacks from './api';
 import { PostPublishDialog } from './components';
+import './style.css'; // This way the general editor styles are loaded before all the component styles.
```
swissspidy: Why was the order of imports changed here? Now the comment is not accurate anymore.

Aside: it might not be true anyway since some other components are imported way earlier now, so to be actually load those styles earlier this import would probably need to be moved to line ~25, before `External dependencies`

tests/phpunit/integration/data/story_post_content_filtered.json
```diff
@@ -1,153 +1,153 @@
...
-          "b": 255
-        }
-      },
-      "type": "page",
-      "id": "a3f18c09-3f10-43c4-9446-794f0541ea1e"
-    }
-  ],
-  "autoAdvance": true,
-  "defaultPageDuration": 7
+	"version": 21,
```
swissspidy: You might want to tell your IDE to not accidentally change indentation of JSON files like this :-)

packages/story-editor/src/app/api/apiProvider.js
```diff
@@ -173,369 +112,141 @@ function APIProvider({ children }) {
...
+    [customPageTemplates, addPageTemplate]
+  );
+
+  actions.deletePageTemplate = useCallback(
+    (id) => deletePageTemplate(id, customPageTemplates),
+    [customPageTemplates, deletePageTemplate]
+  );
+
+  Object.entries(apiCallbacks).forEach(([name, callback]) => {
+    if (!callback) {
```
swissspidy: When would that be the case?

packages/story-editor/src/app/api/apiProvider.js
```diff
@@ -173,369 +112,141 @@ function APIProvider({ children }) {
...
+      actions[name] = undefined;
```
swissspidy, line 212: Alternative:

```js
actions[name] = () => {};
```

WDYT?

packages/wp-story-editor/src/api/index.js
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+export * from './authors';
+export * from './getStory';
+export * from './media';
+export * from './metaboxes';
+export * from './metadata';
+export * from './pageTemplate';
+export * from './saveStory';
+export * from './statusCheck';
```
swissspidy: Nit: this exposes `statusCheck` callbacks to `story-editor`, even though it's a `wp-story-editor` specific util.

swissspidy: Same for the `metaboxes` export above

packages/story-editor/src/app/api/apiProvider.js
mohdsayed: When some api callbacks are not passed from the configuration. One use case is Playground.

packages/story-editor/src/app/api/apiProvider.js
mohdsayed: It would be easier to check `undefined`, than to check an empty function.

packages/wp-story-editor/src/index.js
mohdsayed: I messed up in resolving a merge conflict. 😬

packages/story-editor/src/app/api/apiProvider.js
swissspidy: If it's an empty function we wouldn't have to check for it, no? It would simply be a no-op.

But I suppose that only works well for actions and not selectors.

packages/story-editor/src/app/api/apiProvider.js
swissspidy: Would the playground pass `apiCallbacks: { uploadMedia: null, ... }` or so in that case or simply not pass that at all? If it doesn't pass anything then it's not in `Object.entries` anyway

packages/story-editor/src/app/api/apiProvider.js
mohdsayed: We wouldn't have to check for it if we use the empty function however since they are supposed to return a promise, I was thinking of places where we would need to use `.then()` the code would probably look like this.
```js
    const request = getStoryById();
    if ( request ) {
      request.then( ( post ) => {

      } );
    }
```

and places where it's directly returning the API call like 
https://github.com/google/web-stories-wp/blob/b7412eaad26ad4cf43198806aa1d8dcca7c0588d/packages/story-editor/src/app/story/actions/useAutoSave.js#L45-L55

It would still be an inconsistent return and would need handling.

But by looking at the overall required changes, I think it would need fewer changes to use an empty function. Let me try.

packages/story-editor/src/app/api/apiProvider.js
mohdsayed: Good catch, the playground can simply not pass at all. Updated code.

packages/wp-story-editor/src/api/index.js
mohdsayed: Noted, and added `@todo`, will remove them as I work on moving meta boxes, StatusCheck, and PostLock components.

mohdsayed: Since those still exist in the `story-editor`, they will be required in `APIProivder` for now.

packages/story-editor/src/app/api/apiProvider.js
mohdsayed: When I look more closely, it's not working very well. For example, here how would I do an early return if I don't want to call these if `getMedia` was not provided.
```js
fetchMediaStart({ pageToken: p }); 
const trackTiming = getTimeTracker('load_media'); 
```

https://github.com/google/web-stories-wp/blob/b7412eaad26ad4cf43198806aa1d8dcca7c0588d/packages/story-editor/src/app/media/local/useContextValueProvider.js#L82-L95

packages/wp-story-editor/src/api/media.js
mohdsayed: Absolutely!! 
Moved in https://github.com/google/web-stories-wp/pull/8966/commits/d6d13ce7fcde9312809e3519592d67c2fa0b1f4c

packages/story-editor/src/app/api/apiProvider.js
mohdsayed: Since the condition is required to be added in a lot of files, I would do it later in a separate PR after this is merged.

packages/wp-story-editor/src/api/test/api.js
mohdsayed: Done in https://github.com/google/web-stories-wp/pull/8966/commits/81899b274c1cd6502ceede50146193c11cbf71b3

packages/story-editor/src/app/api/apiProvider.js
swissspidy: That's a good point, thanks for exploring & explaining.

Let's keep it as-is then for now 👍

spacedmonkey:
For context, I have created an issue for making a shared package for dashboard and editor for api interface. This as the other benefit of remove repeated code. https://github.com/google/web-stories-wp/issues/8388 . It was doing this work here, does not make sense to move it to it's own package and share between dashboard and editor?

packages/wp-story-editor/src/api/getStory.js
```diff
@@ -0,0 +1,65 @@
...
+ * External dependencies
+ */
+import { addQueryArgs } from '@web-stories-wp/design-system';
+/**
+ * WordPress dependencies
+ */
+import apiFetch from '@wordpress/api-fetch';
+
+// Important: Keep in sync with REST API preloading definition.
+const STORY_FIELDS = [
```
spacedmonkey: Let's move this file to constant file, it makes easier to share.

packages/wp-story-editor/src/api/getStory.js
```diff
@@ -0,0 +1,65 @@
...
+  'story_data',
+  'preview_link',
+  'edit_link',
+  'embed_post_link',
+  'publisher_logo_url',
+  'permalink_template',
+  'style_presets',
+  'password',
+].join(',');
+
```
spacedmonkey: See this line will need to added 

https://github.com/google/web-stories-wp/blob/9d18d5f52f1b8a88f170df6be099b0a9d69e40c2/packages/story-editor/src/app/api/apiProvider.js#L58 now that https://github.com/google/web-stories-wp/pull/8946 is merged.

packages/wp-story-editor/src/api/pageTemplate.js
```diff
@@ -0,0 +1,69 @@
...
+
+export function getCustomPageTemplates(page = 1, customPageTemplates) {
+  let apiPath = customPageTemplates;
+  const perPage = 100;
+  apiPath = addQueryArgs(apiPath, {
+    context: 'edit',
+    per_page: perPage,
+    page,
+    _web_stories_envelope: true,
+  });
```
spacedmonkey:
```suggestion
  const perPage = 100;
  const apiPath = addQueryArgs(customPageTemplates, {
    context: 'edit',
    per_page: perPage,
    page,
    _web_stories_envelope: true,
  });
```

packages/wp-story-editor/src/api/getStory.js
swissspidy: Share with whom? This is the only place where this constant is needed/used.

packages/wp-story-editor/src/api/storyLock.js
```diff
@@ -0,0 +1,46 @@
...
+export function getStoryLockById(storyId, stories) {
+  const path = addQueryArgs(`${stories}${storyId}/lock`, {
+    _embed: 'author',
+  });
+
+  return apiFetch({ path });
+}
+
+export function setStoryLockById(storyId, stories) {
+  const path = `${stories}${storyId}/lock`;
```
swissspidy: Realized that this trailing slash is missing here (even before). Let's add it. We've had some WAF compatibility issues in the past when the trailing slashes were omitted.

```suggestion
  const path = `${stories}${storyId}/lock/`;
```

packages/wp-story-editor/src/api/storyLock.js
```diff
@@ -0,0 +1,46 @@
...
+ */
+import { addQueryArgs } from '@web-stories-wp/design-system';
+
+/**
+ * WordPress dependencies
+ */
+import apiFetch from '@wordpress/api-fetch';
+
+export function getStoryLockById(storyId, stories) {
+  const path = addQueryArgs(`${stories}${storyId}/lock`, {
```
swissspidy:
```suggestion
  const path = addQueryArgs(`${stories}${storyId}/lock/`, {
```

swissspidy:
> For context, I have created an issue for making a shared package for dashboard and editor for api interface. This as the other benefit of remove repeated code. #8388 . It was doing this work here, does not make sense to move it to it's own package and share between dashboard and editor?

Yes, but it's too early to work on this though. Such a shared package will likely look quite different and contain more things as things evolve. 

Worth noting that this PR is just one step of a long journey. As we walk (see #2918 for roadmap), we will get a better understanding of what such a shared package will look like.

packages/wp-story-editor/src/api/pageTemplate.js
mohdsayed: Done in https://github.com/google/web-stories-wp/pull/8966/commits/a332ebafd61dfc4a7f9ceddb84412e739ce7b117

packages/wp-story-editor/src/api/storyLock.js
mohdsayed: Done in https://github.com/google/web-stories-wp/pull/8966/commits/a332ebafd61dfc4a7f9ceddb84412e739ce7b117

packages/wp-story-editor/src/api/storyLock.js
mohdsayed: Done in https://github.com/google/web-stories-wp/pull/8966/commits/a332ebafd61dfc4a7f9ceddb84412e739ce7b117

packages/wp-story-editor/src/api/getStory.js
mohdsayed: Done, thanks

swissspidy (approved):

packages/wp-story-editor/src/api/getStory.js
spacedmonkey: We have two other post type that follow the same structure, of story data etc. Might be useful in these other places.

packages/wp-story-editor/src/api/pageTemplate.js
```diff
@@ -0,0 +1,68 @@
...
+
+export function getCustomPageTemplates(page = 1, customPageTemplates) {
+  const perPage = 100;
+  const apiPath = addQueryArgs(customPageTemplates, {
+    context: 'edit',
+    per_page: perPage,
+    page,
+    _web_stories_envelope: true,
+  });
+  return apiFetch({ path: apiPath }).then(({ headers, body }) => {
```
spacedmonkey:
```suggestion
  const path = addQueryArgs(customPageTemplates, {
    context: 'edit',
    per_page: perPage,
    page,
    _web_stories_envelope: true,
  });
  return apiFetch({ path }).then(({ headers, body }) => {
```

packages/wp-story-editor/src/api/statusCheck.js
```diff
@@ -0,0 +1,40 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * External dependencies
+ */
+import { base64Encode } from '@web-stories-wp/story-editor';
```
spacedmonkey: This `base64Encode` function should be moved to this package. It is only used in the API provider and is linked to the WordPress implementation.

packages/wp-story-editor/src/api/story.js
```diff
@@ -0,0 +1,115 @@
...
+export function getDemoStoryById(storyId, stories) {
+  const path = addQueryArgs(`${stories}${storyId}/`, {
+    context: 'edit',
+    _embed: STORY_EMBED,
+    web_stories_demo: true,
+    _fields: STORY_FIELDS,
+  });
+
+  return apiFetch({ path });
+}
```
spacedmonkey:
```suggestion
export function getStoryById(storyId, stories, isDemo = false) {
  const path = addQueryArgs(`${stories}${storyId}/`, {
    context: 'edit',
    _embed: STORY_EMBED,
    web_stories_demo: isDemo,
    _fields: STORY_FIELDS,
  });

  return apiFetch({ path });
}

export function getDemoStoryById(storyId, stories) {
  return getStoryById(storyId, stories, true);
}
```

Is this a little cleaner?

packages/story-editor/src/app/api/apiProvider.js
```diff
@@ -71,75 +44,40 @@ function APIProvider({ children }) {
...
-        web_stories_demo: true,
-        _fields: STORY_FIELDS,
-      });
-
-      return apiFetch({ path });
-    },
-    [stories]
-  );
-
   const getStorySaveData = useCallback(
```
spacedmonkey: This function is very heavily linked to the WordPress implementation, like `base64Encode` and all the field mapping, why has this function remained here. This function feels like it should be moved.

packages/wp-story-editor/src/api/story.js
swissspidy: That would expose the `isDemo` param to `story-editor`

packages/wp-story-editor/src/api/getStory.js
mohdsayed: Done in https://github.com/google/web-stories-wp/pull/8966/commits/aa896b0221e33b7e4a96f5685a467aae8904e442

packages/wp-story-editor/src/api/pageTemplate.js
mohdsayed: Done in https://github.com/google/web-stories-wp/pull/8966/commits/951558e5067077554e973489c0745cfbff5ca482

packages/wp-story-editor/src/api/statusCheck.js
mohdsayed: Done in https://github.com/google/web-stories-wp/pull/8966/commits/57e6c7e5626e2065dee39537c42f60881807558f

packages/story-editor/src/app/api/apiProvider.js
mohdsayed: Done in https://github.com/google/web-stories-wp/pull/8966/commits/57e6c7e5626e2065dee39537c42f60881807558f

packages/wp-story-editor/src/api/story.js
mohdsayed: Done in https://github.com/google/web-stories-wp/pull/8966/commits/25cfc094e9734e4a3f558c0b972dd3ff58c53798

mohdsayed:
@spacedmonkey Just a friendly ping that my other PR/tasks are blocked by this PR to some extent 🙂

spacedmonkey (approved):

Issue #8255, opened 2021-07-08
Performance: Improve rendering time of the Page (Discovery)


miina:
## Feature Description
This needs some discovery for detecting ways to improve the page rendering speed.

From the initial render investigation:
Using `React.Profiler` it seemed like rendering the `DisplayLayer` took approximately the same time (or sometimes even less) to render as the `FramesLayer`. E.g. in one run, the display took 360ms and the frames layer took 480ms.

- `FramesLayer`, however, is not really used until the user interacts with the editor. Perhaps there's a way to optimise this so that we wouldn't need to render the frames all the time. E.g. maybe it would only make sense displaying for the active elements and when the user is hovering/focused on the page?

- Additionally, in case of a background image, it seems to always use the fullsize image in the editor, no matter the screen size. Is that purposefully done so?

- Related : #8254 -- perhaps we can also use a placeholder for the page when initially rendering to reduce the jumpiness / offer a better user experience.

It would be good to spend some time to investigate what else can be done to reduce the render time of a Page.

## Additional Context

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

Trapsta:
## Page Loading performance Discovery

For this discovery, I wanted to focus more on the story-editor in isolation and ignore any other bottlenecks that might be unrelated as can be seen in this Lighthouse audit report.

  

![](https://user-images.githubusercontent.com/276316/141190896-9873793e-2c4c-43e4-93e1-91fe8c730750.png)

[Lighthouse Report - Nov 11 2021.zip ](https://github.com/Trapsta/casino-app/files/7516152/localhost_8899-20211110T160636-first-run.html.zip)

  

### Test Lab Environment

1. Network Throttling - Fast 3G

2. CPU Throttling - None

3. Disable caching

4. Incognito session

5. Development build - ( would be better to use the prod build but some tools for profiling have been stripped to optimize the bundle 🥲)

  

### Test Lab Report

  

Saving the chrome dev tools performance profile report here to benchmark any future improvements. Profile is in json format and is easily portable and shareable.

  

[perf profile Nov 11 2021.json.zip](https://github.com/Trapsta/casino-app/files/7516585/Profile-20211110T175141-local-optimized-first-run.json.zip)

  

To optimize the test lab for the editor, I added a button that when clicked loads the editor instead of starting profiling with page reload to ignore any non-editor related bottlenecks.

  

[...]

PR #8971, opened 2021-09-07
Improve page navigation performance

### LINKED ISSUE #1916: Performance: slow navigation
## Bug Description

Please check performance on this issue. It’s not consistent for me. Occasionally, I’ll click on a page thumbnail and it hangs for a split sec before displaying the selected page

Edit:
This navigation performance issue should improve with #7987, however, it looks like thumbnails carousel is not the only part that makes it lag. Especially if there are many elements on the current page, the navigation can be quite slow. Let's figure out what else is making it lag and how can we improve it.

## Screenshots

See video: https://docs.google.com/presentation/d/1ZiZWrmLiX40rt3n_1LMSGJxYWI-xsWFsS9bkzdbRIV8/edit?ts=5ec3ef17#slide=id.g780d0a5251_0_103

## Additional Context

- Plugin Version:
- Operating System:
- Browser:

---

miina:
## Summary
This PR improves the page navigation speed by rendering an image for the active page when just navigating.
Previously, all the elements were displayed for an active page.

Further steps that would improve navigation performance as well (but are not directly related to navigation) could be determined in #8255

## Testing Instructions

1. Make sure the "enable thumbnail caching" experiment is activated.
1. Add a few pages with some elements.
2. Verify that if you navigate to a new pages, an image is displayed as a thumbnail for that page (check the thumbnail source to verify that).
3. Verify that once you change something on the page, it's also reflected on the thumbnail.

## Checklist

---

Fixes #1916


merapi (approved):
It does help a bit 👍🏻     (tested at 4x slowdown).

| before | after |
|--------|-------|
|   default |   optim    |

packages/story-editor/src/components/carousel/pagepreview/index.js
```diff
@@ -100,9 +100,19 @@ function PagePreview({ page, label, ...props }) {
   const [imageBlob, setImageBlob] = useState();
   const [pageNode, setPageNode] = useState();
   const setPageRef = useCallback((node) => node && setPageNode(node), []);
-  const hasImage = !isActive && imageBlob;
   const pageAtGenerationTime = useRef();
 
+  const activeStateTracker = useRef(isActive);
+  const isPageImageUnchanged = isActive && activeStateTracker.current === false;
+  const hasImage = (!isActive || isPageImageUnchanged) && imageBlob;
```
barklund: The `isActive &&` here is unnecessary, as `isPageImageUnchanged` will only ever be looked at if `isActive` is not false (so _truthy_).

Also, is the `=== false` necessary? Could it be other _falsey_ values?

These could probably just be collapsed to:

```suggestion
  const hasImage = (!isActive || !activeStateTracker.current) && imageBlob;
```

barklund (approved):
This works great. Note that the functionality is currently behind a feature flag (default false).

barklund:
@googlebot I consent.

packages/story-editor/src/components/carousel/pagepreview/index.js
```diff
@@ -101,10 +101,19 @@ function PagePreview({ page, label, ...props }) {
...
+  useEffect(() => {
+    // We're keeping track of the previous active state
+    // so that if the page was not active before but is now, an image would be displayed
+    // since this means that the page hasn't changed compared to its image.
+    activeStateTracker.current = isActive;
+  }, [isActive]);
+
   // Whenever the page is re-generated and this is not the active page
   // remove the old (and now stale) image blob
   useEffect(() => {
```
barklund, line 109: Actually, thinking a bit more about it, wouldn't just removing the `&& !isActive` check from the line below do the trick? Then the blob won't be cleared just because the page is made active - only once the page actually starts changing.

miina: You mean removing `&& !isActive` from below and displaying a thumbnail whenever `imageBlob` is truthy?

barklund: Yes, I think that would work - because the image blob is only removed when the page is changed and only regenerated once that page is inactive again.

PR #8972, opened 2021-09-07
Implement video trimming logic

### LINKED ISSUE #8891: Implement video trimming logic
## Feature Description

Implement the actual video trimming logic that is missing after #8890 has been created.

- [ ] Create new video asset in WordPress
- [ ] Make sure new video asset is hidden from media library
- [ ] Associate link to original video + offsets as metadata to new video asset

spacedmonkey:
## Summary

- Add a `web_stories_trim` post meta
- Create a new function in [useProcessMedia](https://github.com/google/web-stories-wp/blob/main/packages/story-editor/src/app/media/utils/useProcessMedia.js), similar to [muteExistingVideo](https://github.com/google/web-stories-wp/blob/6ed8435270cfa5470dbcc2df05241d739df034ea/packages/story-editor/src/app/media/utils/useProcessMedia.js#L171-L255). This would take the following params, resource, startTime, endTime. We will reuse logic that copies over data in the rest api. This will handle adding the new video to the media queue and updating existing elements in the pages. 
- Create a trimVideo function in [useFFmpeg](https://github.com/google/web-stories-wp/blob/main/packages/story-editor/src/app/media/utils/useFFmpeg.js). This will take file, startTime, endTime as params. 

## Testing Instructions

1.

## Checklist

---

Fixes #8891


packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
```diff
@@ -474,7 +474,12 @@ const useQuickActions = () => {
 
   const isVideoTrimEnabled = useFeature('enableVideoTrim');
   const videoCommonActions = useMemo(() => {
-    return isVideoTrimEnabled
+    if (!selectedElement?.resource) {
+      return [];
+    }
+    const { resource } = selectedElement;
+    const { isMuting, isTranscoding, isTrimming } = resource;
+    return isVideoTrimEnabled && !isMuting && !isTranscoding && !isTrimming
```
spacedmonkey: This conditional logic stops a user from transcoding, muting and trimming all at the same time.

packages/story-editor/src/app/media/utils/useFFmpeg.js
```diff
@@ -279,6 +279,66 @@ function useFFmpeg() {
     [getFFmpegInstance]
   );
 
+  /**
+   * Trim Video using FFmpeg.
+   *
+   * @param {File} file Original video file object.
+   * @param {string} start Time stamp of start time of new video. Example '00:01:00'.
```
spacedmonkey: @barklund Mark here, that ffmpeg expects this format. This is what we should store in post meta.

swissspidy: ffmpeg supports milliseconds, e.g. `00:01:02.345`. We want to support that too in the editor.

includes/Media/Video/Trim.php
```diff
@@ -0,0 +1,125 @@
...
+	 * @var string
+	 */
+	const TRIM_POST_META_KEY = 'web_stories_trim';
+
+	/**
+	 * Is trim.
+	 *
+	 * @var string
+	 */
+	const IS_TRIM_KEY = 'trim';
```
swissspidy: This needs to be prefixed

includes/Media/Video/Trim.php
```diff
@@ -0,0 +1,125 @@
...
+namespace Google\Web_Stories\Media\Video;
+
+use Google\Web_Stories\Service_Base;
+
+/**
+ * Class Optimization
+ *
+ * @package Google\Web_Stories\Media\Video
+ */
+class Trim extends Service_Base {
```
swissspidy: The `Muting` class is called `Muting`. Should we call this `Trimming` accordingly?

Also see the incorrect class name in the docblock here.

includes/Media/Video/Trim.php
spacedmonkey: Why does it need to be prefixed?

includes/Media/Video/Trim.php
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	/**
+	 * Register meta for attachment post type.
+	 *
+	 * @since 1.12.0
+	 *
+	 * @return void
+	 */
+	protected function register_meta() {
```
swissspidy: For video muting we also used `register_rest_field()`. Why not here? Seems like we should use the same logic for both as they're quite similar.

spacedmonkey: I used it to use register_rest_field, as we needed to support boolean and null.
https://github.com/google/web-stories-wp/blob/0f61a57fac4aa4dd4b56cb4575c55fdcfa5c657b/includes/Media/Video/Muting.php#L123

We don't need this here.

includes/Media/Video/Trim.php
spacedmonkey: This is a draft, so I will fix things like docblocks then. 

I changed the class name.

packages/story-editor/src/app/media/utils/useFFmpeg.js
spacedmonkey: Didn't say we didn't support milliseconds here. Just flagging the format is not a number, see original comment.
https://github.com/google/web-stories-wp/issues/8891#issuecomment-911737499

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -125,7 +125,11 @@ function useProcessMedia({
         ) {
           uploadVideoPoster(resource.id, resource.src);
         }
-        if ('video' === resource.type && !resource.local && !resource.isMuted) {
+        if (
+          'video' === resource.type &&
+          !resource.local &&
+          resource.isMuted === null
```
spacedmonkey, line 136: This was a mistake I fixed

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -168,6 +172,91 @@ function useProcessMedia({
...
+        let file = false;
+        try {
+          file = await fetchRemoteFile(url, mimeType);
+        } catch (e) {
+          // Ignore for now.
+          return;
+        }
+
+        await uploadMedia([file], {
+          onUploadSuccess,
```
spacedmonkey, line 251: Should we copy over the existing poster for the trimmed video or should force generate a new one @swissspidy ?

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -168,6 +172,91 @@ function useProcessMedia({
...
+          onUploadStart,
+          onUploadError,
+          onUploadProgress,
+          additionalData: {
+            is_muted: oldResource.isMuted,
```
spacedmonkey, line 256: Copy over existing is_muted value.

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
```diff
@@ -136,8 +152,16 @@ function VideoOptionsPanel({ selectedElements, pushUpdate }) {
       clearEditing();
     } else {
       setEditingElementWithState(selectedElements[0].id, { isTrimming: true });
+      trimExistingVideo({ resource, start: '00:00:01', end: '00:00:02' });
```
spacedmonkey: This is a test implementation for now. I will leave it to @barklund to do his thing to hook up these pieces.

packages/story-editor/src/app/media/utils/useProcessMedia.js
swissspidy: Good question! I think we need a new one, because the existing one would not necessarily be a correct representation of the video anymore.

Could also make this smarter by only generating a new one if the the video has been trimmed from the start (if start is 00:00:00 there's no need to do that)

includes/Media/Video/Trim.php
swissspidy: nvm, thought it's a post meta key at first glance.

packages/story-editor/src/app/media/utils/useFFmpeg.js
swissspidy: Let's update the JSDoc to make that clear.

```suggestion
   * @param {string} start Time stamp of start time of new video. Example '00:01:02.345'.
```

packages/story-editor/src/app/media/utils/useProcessMedia.js
spacedmonkey: I think it is safer to just generate a new one.

spacedmonkey: What is they already had a custom poster on it?

packages/media/src/createResource.js
```diff
@@ -121,6 +123,7 @@ function createResource({
   isPlaceholder = false,
   isOptimized = false,
   isMuted = false,
+  trim,
```
miina: Might be good to rename this to `trimData` to avoid confusion and express more clearly what this is. Just `trim` feels like a function without context.

spacedmonkey: Fixed https://github.com/google/web-stories-wp/pull/8972/commits/7527bbe5777d2946fb022410f05ebcc06f865e29

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -168,6 +173,92 @@ function useProcessMedia({
...
+
+      const onUploadError = () => {
+        updateExistingElements({
+          oldResource: { ...oldResource, isTrimming: false },
+        });
+      };
+
+      const onUploadSuccess = ({ resource }) => {
+        copyResourceData({ oldResource, resource });
+        updateOldMutedObject(oldResource.id, resource.id);
```
barklund: This function seems wrongly named now, that it's also used for trimmed videos. Maybe rename to `updateOldTranscodedObject`?

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -168,6 +173,92 @@ function useProcessMedia({
...
+        if ('video' === resource.type && !resource.local) {
+          if (!resource.posterId) {
+            uploadVideoPoster(resource.id, resource.src);
+          }
+          if (resource.isMuted === null) {
+            updateVideoIsMuted(resource.id, resource.src);
```
barklund, line 221: A similar renaming of this seems required to avoid confusion.

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -168,6 +173,92 @@ function useProcessMedia({
...
 
+  const trimExistingVideo = useCallback(
+    ({ resource: oldResource, start, end }) => {
+      const { src: url, mimeType } = oldResource;
+
+      const trimData = {
+        original: oldResource.id,
+        start,
+        end,
+      };
```
barklund, line 196: This is perfect - exactly what we need! 👏

packages/media/src/createResource.js
```diff
@@ -58,7 +58,8 @@ import normalizeResourceSizes from './normalizeResourceSizes';
  * @property {boolean} isOptimized Whether the resource has already been optimized.
  * @property {boolean|null} isMuted Whether the resource has already been muted.
  * @property {Object} sizes Object of image sizes.
- * @property {?Object} output An optional object of video sizes for rendering gifs as videos
+ * @property {?Object} output An optional object of video sizes for rendering gifs as videos.
+ * @property {?Object} trimData An optional object of video trim data.
```
barklund: Should we type this properly in JSDoc as well?

```
/**
 * @typedef {TrimData} Trim data object linking a trimmed video to its original
 * @property {number} original The ID of the original video.
 * @property {number} start This video's start offset relative to original video.
 * @property {number} end This video's end offset relative to original video.
 */
 ...
 * @property {?TrimData} trimData An optional object of video trim data.
```

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
barklund: This is fixed in #8944 - `isVideoTrimEnabled` will be false if the video is already transcoding for any reason.

barklund (commented):
Generally, this looks good to me and should have all the information we need for the UX. I can't fully approve before I try combining the two things, but for now, I'm very happy 🎉

packages/story-editor/src/app/media/utils/useProcessMedia.js
barklund: Oh wait, I'm probably mistaken about this one - it **does** have to do with muting.

includes/uninstall.php
```diff
@@ -124,6 +125,7 @@ function delete_stories_post_meta() {
 	delete_post_meta_by_key( Optimization::OPTIMIZED_ID_POST_META_KEY );
 	delete_post_meta_by_key( Muting::MUTED_ID_POST_META_KEY );
 	delete_post_meta_by_key( Muting::IS_MUTED_POST_META_KEY );
+	delete_post_meta_by_key( Trimming::TRIM_POST_META_KEY );
```
swissspidy, line 128: 👍

packages/media/src/createResource.js
swissspidy: Just need to ensure the types are consistent with what we store in the database etc. Right now I think start/end is a string.

includes/Media/Video/Trimming.php
```diff
@@ -0,0 +1,125 @@
...
+	 * @var string
+	 */
+	const TRIM_POST_META_KEY = 'web_stories_trim';
+
+	/**
+	 * Is trim.
+	 *
+	 * @var string
+	 */
+	const IS_TRIM_KEY = 'trim';
```
swissspidy:
```suggestion
	/**
	 * Is trim.
	 *
	 * @var string
	 */
	const TRIM_DATA_KEY = 'trim_data';
```

To match the naming with the JS side.

includes/Media/Video/Trimming.php
```diff
@@ -0,0 +1,125 @@
...
+	 *
+	 * @return void
+	 */
+	protected function register_meta() {
+		register_meta(
+			'post',
+			self::TRIM_POST_META_KEY,
+			[
+				'type'           => 'object',
+				'description'    => __( 'Trimmed data for video.', 'web-stories' ),
```
swissspidy:
```suggestion
				'description'    => __( 'Video trim data.', 'web-stories' ),
```

includes/Media/Video/Trimming.php
```diff
@@ -0,0 +1,125 @@
...
+ * @package Google\Web_Stories\Media\Video
+ */
+class Trimming extends Service_Base {
+
+	/**
+	 * The trim video post meta key.
+	 *
+	 * @var string
+	 */
+	const TRIM_POST_META_KEY = 'web_stories_trim';
```
swissspidy: Maybe also `web_stories_trim_data`?

packages/story-editor/src/app/media/useUploadMedia.js
```diff
@@ -195,6 +195,7 @@ function useUploadMedia({
      * @param {Function} args.onUploadSuccess Callback for when upload succeeds.
      * @param {Object} args.additionalData Object of additionalData.
      * @param {boolean} args.muteVideo Should the video being transcoded, should also be muted.
+     * @param {Object} args.trimData Trim data.
```
swissspidy: A proper JSDoc type would be good here too, see Morten's other comment.

packages/story-editor/src/app/media/utils/useFFmpeg.js
```diff
@@ -279,6 +279,66 @@ function useFFmpeg() {
...
+      const trackTiming = getTimeTracker('load_trim_video_transcoding');
+
+      let ffmpeg;
+
+      try {
+        ffmpeg = await getFFmpegInstance(file);
+
+        const tempFileName = uuidv4() + '.' + MEDIA_TRANSCODED_FILE_TYPE;
+        const outputFileName =
+          getFileName(file) + '-trimmed.' + MEDIA_TRANSCODED_FILE_TYPE;
```
swissspidy, line 303: Hmm shouldn't we use the actual file type of the input file here?

Would this convert a webm video to mp4 by accident?

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -168,6 +173,90 @@ function useProcessMedia({
     ]
   );
 
+  const trimExistingVideo = useCallback(
+    ({ resource: oldResource, start, end }) => {
```
swissspidy, line 189: Another function where JSDoc for proper format of `start` and `end` would be beneficial.

packages/story-editor/src/app/media/utils/useProcessMedia.js
swissspidy: Thinking about this more, let's just copy over the existing poster. It's the safer approach as the generation could fail. If users are not happy with the poster they can easily remove it to have a new one generated.

swissspidy (commented):
Overall approach looks good 👍 

As per discussion yesterday, it would be best for the actual UI to do sanity checks like ensuring start&end time are valid.

packages/story-editor/src/app/media/utils/useFFmpeg.js
spacedmonkey: This is how muting works....

packages/story-editor/src/app/media/utils/useProcessMedia.js
spacedmonkey: It now copies the existing poster.

packages/story-editor/src/app/media/utils/useFFmpeg.js
swissspidy: So is it indeed converting webm to mp4?

That means we'd need to fix that for muting too in that case.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
```diff
@@ -271,6 +274,62 @@ export function finishMuting(state, { payload: { id, file } }) {
...
+/**
+ * Starts trimming a file.
+ *
+ * @param {Object} state Current state.
+ * @param {Object} action Action object.
+ * @param {Object} action.payload Action payload.
+ * @param {string} action.payload.id Item ID.
+ * @return {Object} New state
+ */
+export function startTrimming(state, { payload: { id } }) {
```
swissspidy, line 286: These should have tests like all the other reducers

packages/story-editor/src/app/media/utils/useFFmpeg.js
spacedmonkey: Yeah, that seems to be the case. In testing, it doesnt seem that there is a easy way to fix this. As this is an existing issue, I will create a follow up issue and taken this there, as isn't a simple fix.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
spacedmonkey: Good catch. Added here - https://github.com/google/web-stories-wp/pull/8972/commits/ff4cd935b4dbd6459de2bcf7cd948a7834face71

packages/story-editor/src/app/media/utils/useFFmpeg.js
swissspidy: +1 to creating a dedicated ticket, perhaps even adding a TODO here in the code.

Curious to hear why it wouldn't be an easy fix. Seems like we should be able to get the existing resource's/file's type/extension and use that, or pass it down if needed.

swissspidy (approved):

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -173,9 +178,16 @@ function useProcessMedia({
     ]
   );
 
+  /**
+   * Trim existing video using FFmpeg.
+   *
+   * @param {import('@web-stories-wp/media').Resource} resource Resource object.
+   * @param {string} start Time stamp of start time of new video. Example '00:01:02.345'.
+   * @param {string} end Time stamp of end time of new video. Example '00:02:00'.
```
barklund, line 186: Why is it a timestamp here? Is it because that's what FFMPEG expects? I'd much prefer time in milliseconds as a number instead, but we can do the conversion in the UI of course.

barklund (approved):
Looks great to me with a single comment.

packages/story-editor/src/app/media/utils/useProcessMedia.js
spacedmonkey: I called this out early in the PR - https://github.com/google/web-stories-wp/pull/8972#discussion_r703621774

barklund: Thanks, hadn't seen that comment. I'm all good then!

PR #8975, opened 2021-09-07
Design System: Hierarchical Taxonomy Component

### LINKED ISSUE #8836: Hierarchical taxonomy component
The Taxonomy panel has 2 types of forms, freeform and hierarchical. Hierarchical is used for categories. But users can have custom taxonomies that are also hierarchical. This component will be used for both of those. 

There should be an ability to search the list of options and each option should be a checkbox. Each option should be able to be toggled true/false. 

category demo

We want to do this same thing in this gif but using our design system.

We should be able to reuse elements in the design system to build out this full component.

## Acceptance criteria 
- Storybook of this new component. 
- Unit tests
- Passing accessibility standards
- RTL and LTR support

samwhale:
## Context

Adding some taxonomy components to the design system. [Designs](https://www.figma.com/file/1YM3jA9h4QGc4xLdyaXIJS/Stories-Sprint-1.7?node-id=9721%3A383439)

## Summary

Adds a hierarchical input component to the design system.

## Relevant Technical Choices

The data structure and tree building was inspired from [gutenberg](https://github.com/WordPress/gutenberg/blob/3f9968e2815cfb56684c1acc9a2700d8e4a02726/packages/editor/src/utils/terms.js#L13-L40). Keeping the options flat makes it easy to find options when needing to update checked values.

## To-do

n/a

## User-facing changes

None - only in storybook

input

|||
|--|--|
|image|image|

## Testing Instructions

1. Load up storybook `$ npm run storybook`
2. Go to `Stories Editor/Components/Form/Hierarchical`
3. Play with the component. Should be keyboard accessible and work with voiceover/chromevox/screen readers similar to the gutenberg editor.
4. When testing accessibility: after typing the input should announce `n results found` 👍 

no

no

no

## Checklist

---

Fixes #8836


samwhale:
Now that I think about it - would be nice to get an aria live update of how many results there are after typing a search term in the input. If we debounce the notification then we won't get a notification on every keystroke too

package-lock.json
```diff
@@ -27070,6 +27071,11 @@
       "integrity": "sha1-8xwiIlqWMtK7+OSt2+8kCqdlph8=",
       "dev": true
     },
+    "node_modules/lodash.groupby": {
+      "version": "4.6.0",
+      "resolved": "https://registry.npmjs.org/lodash.groupby/-/lodash.groupby-4.6.0.tgz",
+      "integrity": "sha1-Cwih3PaDl8OXhVwyOXg4Mt90A9E="
+    },
```
BrittanyIRL: 🤔  we don't use lodash anywhere else except in the `/stories-block` package. I'm curious about maybe moving the util that is in the dashboard (or copying it for the time being) that gives us the same `groupBy` functionality without importing lodash.  it's in `dashboard/src/utils/groupBy`

BrittanyIRL (commented):
Nice work on this! It mirrors Gutenberg really well. 

Curious your opinion about ditching lodash in favor of a util, then i'm happy to approve!

packages/story-editor/src/components/form/hierarchical/utils/index.js
```diff
@@ -0,0 +1,106 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * External dependencies
+ */
+import groupBy from 'lodash.groupby';
```
swissspidy: Probably added by your IDE. This needs to be imported from the design system now :-)

packages/story-editor/package.json
```diff
@@ -38,6 +38,7 @@
         "flagged": "^2.0.1",
         "history": "^5.0.1",
         "html-to-image": "^1.7.0",
+        "lodash.groupby": "^4.6.0",
```
samwhale: @swissspidy @spacedmonkey What do y'all think about me adding this in? Should we write a util instead?

here's the sizing
image

packages/story-editor/src/components/form/hierarchical/utils/index.js
samwhale: Good catch! I ended up reverting that change though - turns out our `groupBy` is different than the `groupBy` from lodash:

https://github.com/google/web-stories-wp/blob/main/packages/dashboard/src/utils/groupBy.js#L17-L29

Our `groupBy` doesn't create an object of arrays like lodash's does: https://lodash.com/docs/4.17.15#groupBy

samwhale: I'm happy to write our own util here though

packages/story-editor/src/components/form/hierarchical/index.js
```diff
@@ -0,0 +1,204 @@
...
+        <Checkbox
+          {...checkboxProps}
+          id={optionId}
+          onChange={(evt) => onChange(evt, option)}
+        />
+        <Label htmlFor={optionId}>{optionLabel}</Label>
+      </CheckboxContainer>
+      <DirectionAware>
+        <StepContainer>
+          {options?.map((child) => (
```
BrookeGraham: Question, do we want the `StepContainer` to render if `options` is undefined? If not, could we move the `options` check above? Or is the extra empty div needed?

packages/story-editor/src/components/form/hierarchical/utils/index.js
```diff
@@ -0,0 +1,106 @@
...
+
+const fillTree = (groupedOptionsByParent, options = []) => {
+  return options?.map((option) => {
```
BrookeGraham: Will `options` ever be `null`? If not, the chaining operator check is redundant, won't options will always be `true` since we set a default.

packages/story-editor/package.json
swissspidy: So far we've been pretty strict/consistent with not using any lodash utils and I'd prefer to keep it this way.

samwhale: I'll add a custom util tomorrow. Thanks!

samwhale: Ended up adding one right now -> not using lodash anymore 🙌

spacedmonkey: I know that @barklund was really against using lodash.

packages/story-editor/src/components/form/hierarchical/utils/index.js
```diff
@@ -0,0 +1,128 @@
...
+export const buildOptionsTree = (flatOptions) => {
+  const formattedOptions = flatOptions.map((option) => ({
+    options: [],
+    parent: null,
+    ...option,
+  }));
+
+  const groupedOptionsByParent = groupBy(formattedOptions, 'parent');
+
+  return fillTree(groupedOptionsByParent, groupedOptionsByParent[null]);
```
swissspidy: LGTM flags this `null` usage here:

> This expression will be implicitly converted from null to string.

BrookeGraham:
This looks so good. Just a couple very small  questions.

packages/story-editor/src/components/form/hierarchical/utils/index.js
samwhale: Good call 👍 options won't ever be null, so we don't need to do this

packages/story-editor/src/components/form/hierarchical/index.js
samwhale: We should move this check!

package-lock.json
```diff
@@ -4,6 +4,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "web-stories-wp",
```
BrookeGraham: Is this needed? I didn't see a package.json change.

packages/story-editor/src/components/form/hierarchical/utils/index.js
```diff
@@ -0,0 +1,163 @@
...
+ * @param {Array.<Object>} option Array of options to count
+ * @return {number} the count of all options
+ */
+const countOptions = (option) => {
+  if (!option?.options?.length) {
+    return 1;
+  }
+
+  return (
+    1 + option.options.reduce((sum, child) => sum + countOptions(child), 0)
```
BrookeGraham, line 152: 🤯

package-lock.json
samwhale: Just for visibility:

The `groupBy` util in the dashboard doesn't do the same thing as lodash's but we also don't want to bring in lodash as a dep - so we opted to make our own.

packages/story-editor/src/components/form/hierarchical/index.js
```diff
@@ -0,0 +1,242 @@
...
+`;
+
+const Border = styled.div`
+  padding-top: 4px;
+  margin-top: -4px;
+  border-left: 1px solid ${({ theme }) => theme.colors.border.defaultNormal};
+  border-right: 1px solid ${({ theme }) => theme.colors.border.defaultNormal};
+  border-bottom: 1px solid ${({ theme }) => theme.colors.border.defaultNormal};
+  border-bottom-left-radius: ${({ theme }) => theme.borders.radius.small};
+  border-bottom-right-radius: ${({ theme }) => theme.borders.radius.small};
```
spacedmonkey: Could we use short hand here?

maxyinger (approved):
Awesome work here. Excellent documentation and tests. very well laid out 👏

packages/story-editor/src/components/form/hierarchical/utils/index.js
```diff
@@ -0,0 +1,163 @@
...
+  if (options) {
+    for (let i = 0; i < options.length; i++) {
+      newChildren = newChildren.concat(filterOption(options[i], labelText));
+    }
+  }
+
+  // Find if current option matches
+  const matchesCurrentOption = option.label
+    .toLowerCase()
+    .includes(labelText?.trim().toLowerCase());
```
BrookeGraham: Similar question here, do we expect `labelText` to ever be `null` ? Seems a little weird to use optional chaining when we default to a string.

BrookeGraham:
I love it.

BrookeGraham (approved):

packages/story-editor/src/components/form/hierarchical/utils/index.js
samwhale: good call!

Issue #7744, opened 2021-05-26
Editor: crop story poster image and publisher logo upon selection


o-fernandez:
## Feature Description

Currently, when I select an image from the WP media library to use as my story poster image or as the publisher logo, the image is used in the existing dimensions. This causes problems with images that are either (1) two small and/or (2) in the wrong dimensions. 

The proposal is to: 
1. Only allow the user to select an image that meets the minimum size requirements. For poster image a minimum of 640x853px is required and for publisher logo a minimum of 96x96. 
2. After the user selects the image, add a step to crop to the right size and aspect ratio. The aspect ratio should be fixed based on the requirements. For poster image 3:4 and for publisher logo 1:1. And the minimum size chosen should not be below the minimum requirements per step 1. 

This should make it easy and clear for the creator to set the right poster and publisher logo sizes. 

NOTE: we should consider both the workspace/editor panels and the settings page for publisher logo. 

Adding video from @swissspidy on existing WP solution we can leverage. 

https://user-images.githubusercontent.com/13389997/119665601-90979f80-be02-11eb-9a6e-a5ce1687834e.mov

## Additional Context

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

swissspidy:
Code used by WP for the cropping: https://github.com/WordPress/wordpress-develop/blob/2382765afa36e10bf3c74420024ad4e85763a47c/src/js/_enqueues/wp/customize/controls.js#L4711-L4735

swissspidy:
It should be possible to get a PoC for cropping poster images & publisher logos in the editor based on the precedence set by WP core on cropping within a media modal.

Difficulty of adding e2e tests for that has been mentioned, but that can be neglected when there's thorough manual QA.

Plus, a nice benefit of this is that we can probably remove the `web-stories-poster-portrait` image size in PHP, as the user will do manual cropping in the editor.

csossi:
Testing in PR 8067 in QA Environment

If I opt to Skip Cropping here:
![image.png](https://images.zenhubusercontent.com/5ea085e4704e72c0a99c6997/0da725a1-9629-4753-98f2-83ce5cbdab1d)

The editor crashes:
![image.png](https://images.zenhubusercontent.com/5ea085e4704e72c0a99c6997/5c7f3d12-c434-45de-b16e-d0311436f60d)

csossi:
Verified in QA

Issue #4148, opened 2020-08-20
Dashboard Settings: Ability to crop publisher logos


BrittanyIRL:
## Feature Description

When a user uploads a logo ensure it is cropped to be a perfect square, a user may crop the image during upload.

## Additional Context

- https://docs.google.com/document/d/1Xs4uOZ3KCNlTMdjDT6q7cvTUpA1xk-kzrnVLQALltQk/edit#

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance Criteria
- User has ability to crop an image to perfect square during upload

## Implementation Brief
Right now, the settings view in dashboard doesn't use the media uploader that wordpress has natively that's harnessed in the editor. I'm not super sure how we want to go about this.

swissspidy:
Related: #8067 

We could leverage the same for the Dashboard too. Publisher logo selection/upload would open the WordPress media modal for selection & cropping.

BrittanyIRL:
Since this is cropping media, @swissspidy should it be considered media related or dashboard related for pod assignment since you guys officially took over media recently?

swissspidy:
Good question! I'll assign to WP pod for now.

swissspidy:
Totally forgot that logos can be uploaded via drag & drop.

If we want to use the WordPress media modal, perhaps we should remove the drag & drop feature? Else, how would the cropping be done after dropping? (unless we build a super custom solution for the cropping)

cc @spacedmonkey

BrittanyIRL:
oh good catch!! 
I think removing the drag and drop for consistency in upload would make sense. As WordPress users, it's the customary pattern to uploading media. Just my 2 cents.

spacedmonkey:
Now the there is the ability to crop images built into the WordPress media library, https://github.com/google/web-stories-wp/issues/7337 becomes more useful

PR #8989, opened 2021-09-09
Add dropdown for selecting publisher logo

### LINKED ISSUE #5108: Editor: Cannot use multiple publisher logos

merapi:
## Summary

Instead of upload image dialog, this adds a dropdown list to select the publisher logo for a given story (each story has its own publisher logo).
According to https://xwp.slack.com/archives/C01ESBB8ZN1/p1617715788058500?thread_ts=1617197455.025200&cid=C01ESBB8ZN1

Before:
image

Now:
| Empty state, no logos uploaded | Default logo is selected when available | Dropdown expanded |
|---|---|---|
| Screenshot 2021-09-09 at 16 55 40 | Screenshot 2021-09-09 at 17 01 53 | Screenshot 2021-09-09 at 17 02 55 |
|---|---|---|

## Relevant Technical Choices

- I've added an API function for fetching the settings.

## To-do

- [ ] File a bug about dropdown: when search is disabled, navigation with keyboard (arrows) is broken
- [ ] File a bug about dropdown: dropdown cannot be highlighted (from the pre-publish checklist)

## User-facing changes

Well explained above.

## Testing Instructions

I've tested this code with:

1. No publisher logos (hard to do on QA env because we cannot remove the last logo?)
2. Saved the story with no publisher logos then added a logo in settings and checked the story - because it was saved without the logo it stays like that (is that ok?)
3. Tested that each story has its own logo and uploading a new one doesn't change any of them.

## Checklist

---

Fixes #5108


packages/story-editor/src/components/panels/document/publish/publish.js
```diff
@@ -250,22 +297,23 @@ function PublishPanel() {
...
+              <Row>
+                <Required />
+                <Link
+                  rel="noopener noreferrer"
+                  target="_blank"
+                  href={settingsLink}
+                  size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.X_SMALL}
+                >
+                  {__('Manage', 'web-stories')}
+                </Link>
```
merapi: Should it open in a new tab or do we want to rely on the `UnsavedChangesWarning`? also, is there a better way than hardcoding the URL like this?

swissspidy:
This PR seems to remove the ability to upload a custom publisher logo to the editor [as described in Figma](https://www.figma.com/file/Cd9iacVknHK62RO4pRwb68/Stories-Editor?node-id=5837%3A4).

Screenshot 2021-09-09 at 17 44 57

Removing the ability to upload a custom publisher logo is quite the step backward.

spacedmonkey:
> Removing the ability to upload a custom publisher logo is quite the step backward. 

Not to mention, removing all the work I did lately into cropping the image when it is picked.

merapi:
> Removing the ability to upload a custom publisher logo is quite the step backward.

You can still upload them on the Settings page.

We need a product decision because I feel like [Omar iterated over what we currently have](https://xwp.slack.com/archives/C01ESBB8ZN1/p1617197455025200) vs developing new components as per Figma (which are not ideal since as soon somebody will upload the first logo there is zero discoverability in that DND feature [other parts of the editor don't have it, like poster image next to it, so it's a bit confusing when you can do that]).

From the Slack conversation, [it was clear](https://xwp.slack.com/archives/C01ESBB8ZN1/p1617717591059000?thread_ts=1617197455.025200&cid=C01ESBB8ZN1) that upload will be only possible via Manage link, and no one argued against it.

merapi:
> Not to mention, removing all the work I did lately into cropping the image when it is picked.

BTW: shouldn't that be also possible on the Settings page?

swissspidy:
> > Removing the ability to upload a custom publisher logo is quite the step backward.
> 
> You can still upload them on the Settings page.
> 
> We need a product decision because I feel like [Omar iterated over what we currently have](https://xwp.slack.com/archives/C01ESBB8ZN1/p1617197455025200) vs developing new components as per Figma (which are not ideal since as soon somebody will upload the first logo there is zero discoverability in that DND feature [other parts of the editor don't have it, like poster image next to it, so it's a bit confusing when you can do that]).

@choumx @aaskedall Can we revisit this please and?

There should be a way to upload a new publisher logo directly from within the editor & crop it. Doesn't have to be via drag and drop, but at least there should be an upload button like we currently have.

> From the Slack conversation, [it was clear](https://xwp.slack.com/archives/C01ESBB8ZN1/p1617717591059000?thread_ts=1617197455.025200&cid=C01ESBB8ZN1) that upload will be only possible via Manage link, and no one argued against it.

At least to me that's not very clear at all and not how I interpret the discussion there.

> > Not to mention, removing all the work I did lately into cropping the image when it is picked.
> 
> BTW: shouldn't that be also possible on the Settings page?

We're tracking this in #4148.

merapi:
> At least to me that's not very clear at all and not how I interpret the discussion there.

_"click manage to do **anything** else (e.g., change the default, etc.)"_ - was not super clear to me either, but 
_"Yep, that’s how I understand Figma as well — the user can either choose between the existing directly or click “Manage”_ to **upload new** / set default." and Omar's +1 removed any doubt + it was not a big deal for me (how often do you upload a new publisher logo?), it's still just a few clicks away.

merapi:
Some options for us (both open media picker with crop + we need to make an additional API call to add the selected image to publisher logos, right?):

pencil icon somewhere:
image

as one of the options (should be first? so you don't have to scroll):
image

dreamofabear:
Good catch and discussion. That Slack thread was back in March (before we added the ability to crop publisher logo on upload in June/July #7744), so it's understandable there's some confusion.

I agree with Pascal that we should keep current functionality (upload custom publisher logo and crop it) while adding a dropdown to select from already-uploaded logos. 

I like Marcin's option 2:
> as one of the options (should be first? so you don't have to scroll):
> image

Can we add or just use an icon for the "Open media picker" button that's nested inside the dropdown? Maybe reuse the same upload ☁️ ⬆️   icon that we have for the media library. @aaskedall for input.

merapi:
Just the icon:
| image  | image  |
|---|---|

It looks like one of the logos 🤔 

Another option (BTW: it's not only upload, because it opens media picker):
image

spacedmonkey:
>dropdown to select from already-uploaded logos.

This is no longer possible as of https://github.com/google/web-stories-wp/pull/8946.  In this line we remove the updating of [option](https://github.com/google/web-stories-wp/pull/8946/files#diff-6b011397c22cc9aff84c1cf303869ae7444ff773dcbaaeef8b61e5d0520fc326L171). 

We should keep track of logos a user has uploaded before. So if a user creates 3 stories, all with different logos and goes to make a 4th, then the should be about to select any of the publisher logos they used before.

swissspidy:
> > dropdown to select from already-uploaded logos.
> 
> This is no longer possible as of #8946. In this line we remove the updating of [option](https://github.com/google/web-stories-wp/pull/8946/files#diff-6b011397c22cc9aff84c1cf303869ae7444ff773dcbaaeef8b61e5d0520fc326L171).
> 
> We should keep track of logos a user has uploaded before. So if a user creates 3 stories, all with different logos and goes to make a 4th, then the should be about to select any of the publisher logos they used before.

That's as simple as getting all existing logos via `GET /wp-json/web-stories/v1` and updating `web_stories_publisher_logos` via `POST /wp-json/web-stories/v1` after uploading accordingly.

merapi:
> This is no longer possible as of #8946. In this line we remove the updating of [option](https://github.com/google/web-stories-wp/pull/8946/files#diff-6b011397c22cc9aff84c1cf303869ae7444ff773dcbaaeef8b61e5d0520fc326L171).

You can still upload logos on the Settings page (even without what we will do next here).

> We should keep track of logos a user has uploaded before. So if a user creates 3 stories, all with different logos and goes to make a 4th, then the should be about to select any of the publisher logos they used before.

Yep, as I said:
> Some options for us (both open media picker with crop + we need to make an additional API call to add the selected image to publisher logos, right?):

spacedmonkey:
> That's as simple as getting all existing logos via GET /wp-json/web-stories/v1 and updating web_stories_publisher_logos via POST /wp-json/web-stories/v1 after uploading accordingly.

I guess you mean, updating settings endpoint. Sadly editors and authors can not do this, as they do not have the manage options permission.

swissspidy:
Good point. We should not offer these options for users without the necessary capabilities. Just like today these users can't upload publisher logos already if they cannot upload media.

merapi:
> Good point. We should not offer these options for users without the necessary capabilities. Just like today these users can't upload publisher logos already if they cannot upload media.

I'm already checking for `hasUploadMediaAction` when showing that extra button, `canManageSettings` will be used for the link, but, we need to solve one more issue: those roles can't even read the settings to get the publisher logos list - extra endpoint or get it when fetching the story data? otherwise, they can't change the logo, they must use the default one (if set).

spacedmonkey:
I would hook into the publisher logo id being saved in post meta, get the existing active publisher logos and if the id is not in the list, update the option. This should be done in the backend for the user, no rest api request needed. 

I can prototype something, if this is at all helpful

merapi:
🤔   so how do we do this (as per spec/Figma): _"drop-down where you select the publisher logo, and it would show you all the configured publisher logos"_

swissspidy:
Folks, let's create a ticket for this. We can discuss specifics there.

> I would hook into the publisher logo id being saved in post meta, get the existing active publisher logos and if the id is not in the list, update the option. This should be done in the backend for the user, no rest api request needed.

That does not address the part where users need to get a list of all publisher logos.

> I'm already checking for `hasUploadMediaAction` when showing that extra button, `canManageSettings` will be used for the link, but, we need to solve one more issue: those roles can't even read the settings to get the publisher logos list - extra endpoint or get it when fetching the story data? 

Yeah an extra endpoint sounds reasonable to me for this.

> otherwise, they can't change the logo, they must use the default one (if set).

Right. Let's go with this for now until we have something implemented for you.

merapi:
> Let's go with this for now until we have something implemented for you.

Should be ready for review now:
image

miina:
@merapi Some tests are failing FYI.

packages/story-editor/src/components/panels/document/publish/publish.js
```diff
@@ -137,51 +179,108 @@ function PublishPanel() {
...
+      <Text as="span" size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL}>
+        {__('Select logo', 'web-stories')}
+      </Text>
+    ) : (
+      <Text as="span" size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL}>
+        {__('No logo', 'web-stories')}
+      </Text>
+    );
+
+  const settingsLink = `${window.location.origin}/wp-admin/edit.php?post_type=web-story&page=stories-dashboard#/editor-settings`;
```
miina: There's `dashboardSettingsLink` in config that can be used here.

packages/story-editor/src/components/panels/document/publish/publish.js
```diff
@@ -137,51 +179,108 @@ function PublishPanel() {
...
+          <LogoImg src={src} alt="" />
+        </Option>
+      );
+    }
+  );
+  const activeItemRenderer = () =>
+    publisherLogo.id ? (
+      <LogoImg src={publisherLogo.url} alt="" />
+    ) : publisherLogos.length ? (
+      <Text as="span" size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL}>
```
miina: Maybe for readability we could do something like this:
```
const displayText = publisherLogos.length ? __( 'Select logo', ... ) : __( 'No logo', ... );
const activeItemRenderer = () => publisherLogo.id ? <LogoImg .... /> : <Text ...>{displayText}</Text>;
```

packages/story-editor/src/components/panels/document/publish/publish.js
```diff
@@ -137,51 +179,108 @@ function PublishPanel() {
...
-            height: image.sizes?.full?.height || image.height,
-          },
-        },
-      });
-    },
-    [updateStory]
-  );
-
-  const publisherLogoErrorMessage = useMemo(() => {
+  const posterErrorMessage = useMemo(() => {
```
miina: This looks like something that could be turned into `getErrorMessage`, accepting param to return the message either for the poster or publisher (unless we move the publisher logo logic to a separate file).

packages/story-editor/src/components/panels/document/publish/publish.js
```diff
@@ -232,15 +331,17 @@ function PublishPanel() {
...
+          <DropdownWrapper>
             <MediaWrapper>
-              <StyledMedia
-                width={72}
-                height={72}
-                cropParams={{
-                  width: 96,
-                  height: 96,
-                }}
+              <AdvancedDropDown
```
miina, line 323: Maybe the Publisher logo would deserve its own file now?

packages/story-editor/src/components/form/advancedDropDown/container.js
```diff
@@ -40,7 +40,7 @@ const Container = styled.div`
   flex-direction: row;
   flex-wrap: wrap;
   width: 100%;
-  min-width: 160px;
+  min-width: 120px;
```
miina, line 43: Q: Does this change influence any other use cases of the advanced dropdown?

packages/story-editor/src/app/story/effects/useLoadStory.js
```diff
@@ -95,9 +95,17 @@ function useLoadStory({ storyId, shouldLoad, restore, isDemo }) {
         const publisherLogo = {
           id: embedded?.['wp:publisherlogo']?.[0].id || 0,
```
miina: Maybe extract `embedded?.['wp:publisherlogo']?.[0]` into a separate constant for readability?

miina (commented):
Works great generally.

One issue from testing.
When choosing new Media, after cropping, the Media Library does not close (even though the image gets cropped and selected):
cropping

Let me know if that's a bug that existed before already, will approve then.

packages/story-editor/src/components/panels/document/publish/publish.js
```diff
@@ -137,51 +179,108 @@ function PublishPanel() {
...
+        {__('No logo', 'web-stories')}
+      </Text>
+    );
+
+  const settingsLink = `${window.location.origin}/wp-admin/edit.php?post_type=web-story&page=stories-dashboard#/editor-settings`;
+
+  const publisherLogosWithUploadOption = [...publisherLogos];
+  if (hasUploadMediaAction) {
+    publisherLogosWithUploadOption.unshift(
+      <Option key="upload" onClick={openMediaPicker}>
```
miina: Ah, okay, now I see where the Media Picker "not closing" bug comes from. `onClick={openMediaPicker}` is used twice here -- once on the option and once on `Text`. So it gets opened twice.

swissspidy:
@merapi @spacedmonkey Created #9007 to address the open questions from the above discussion. PTAL

packages/story-editor/src/components/panels/document/publish/publish.js
merapi: Yep, thanks for catching that.

packages/story-editor/src/components/form/advancedDropDown/container.js
merapi: We use it for font picker and authors, both are safe (they are 100% width, 276px most of the time).

packages/story-editor/src/app/story/effects/useLoadStory.js
merapi: I was thinking about that but we don't do that generally(?)
image

packages/story-editor/src/components/panels/document/publish/publish.js
merapi: Done in https://github.com/google/web-stories-wp/pull/8989/commits/4756907aa4ce2e383463fdc6e37848966af7a60b.
About moving to its own file - no strong opinion, it's just a bit longer than it was before.

miina: Yep, that would be the reason -- not to allow files to grow very large whenever possible. Let's keep it as is for now then and do it whenever it should grow more.

packages/story-editor/src/app/story/effects/useLoadStory.js
miina: Hm, okay. Not sure what's the reason 🤔

miina (approved):
LGTM Pending tests.

packages/e2e-tests/src/specs/editor/inspector.js
```diff
@@ -20,24 +20,24 @@ import { createNewStory, withUser } from '@web-stories-wp/e2e-test-utils';
...
       await expect(page).toClick('li[role="tab"]', { text: 'Document' });
-      await await expect(page).toMatchElement('[aria-label="Publisher Logo"]');
+      await expect(page).toMatchElement('[aria-label="Publisher Logo"]');
     });
     describe('Contributor User', () => {
       withUser('contributor', 'password');
 
-      it('should not display publisher logo', async () => {
+      // eslint-disable-next-line jest/no-disabled-tests
+      it.skip('should not display publisher logo upload', async () => {
```
swissspidy: Why is this now skipped?

merapi: I'm just testing what's going on here (_"Test fail reason"_).
Locally I'm getting this, any ideas?

image

swissspidy: The easiest way to find out is by opening the editor as a contributor user and then checking the browser console.
The more complicated way would be by running your tests in non-headless mode and observing the behavior.

Currently, they contributors get a 403 because they don't have access to the `web-stories/v1/settings` route.

You need to check for `canManageSettings` capabilities before trying to access it.

As per our lengthy discussion above, I opened #9007 to move publisher logos to a separate REST route that's open for contributors. I might get to that tomorrow if Jonny doesn't before me.

Until then, you'd need to work around it.

swissspidy: Happy to chat about it tomorrow if things are unclear.

swissspidy:
@merapi since this PR is still WIP, now it would be possible to leverage the new publisher logos REST API routes here.

merapi:
> @merapi since this PR is still WIP, now it would be possible to leverage the new publisher logos REST API routes here.

After checking what was added/changed I've decided that it will be better to work a little more on this and utilize the new stuff (just resolving conflicts is not an option).

packages/story-editor/src/components/panels/document/publish/publish.js
```diff
@@ -116,73 +158,115 @@ function PublishPanel() {
...
+    /* translators: %s: list of allowed file types. */
+    __('Please choose only %s as a poster.', 'web-stories')
+  );
 
-    if (allowedImageFileTypes.length) {
-      message = sprintf(
-        /* translators: %s: list of allowed file types. */
-        __('Please choose only %s as a poster.', 'web-stories'),
-        translateToExclusiveList(allowedImageFileTypes)
+  const openMediaPicker = useMediaPicker({
```
swissspidy: ⚠️ 

This needs to use the `<MediaUpload` component from the config, something like this:

```js
const { MediaUpload } = useConfig();

// ...

<MediaUpload
  onSelect={onNewPublisherLogoSelected}
  onSelectErrorMessage={publisherLogoErrorMessage}
  type={allowedImageMimeTypes}
  render={renderUploadButtonIcon}
title={__('Select as publisher logo', 'web-stories')}
/>
```

packages/story-editor/package.json
```diff
@@ -28,6 +28,7 @@
         "@web-stories-wp/text-sets": "*",
         "@web-stories-wp/tracking": "*",
         "@web-stories-wp/units": "*",
+        "@web-stories-wp/wp-story-editor": "*",
```
swissspidy: ⚠️  `wp-story-editor` requires `story-editor`, not the other way around.

See my other comment about how not to use `useMediaPicker`

packages/story-editor/src/components/panels/document/publish/publish.js
```diff
@@ -17,24 +17,33 @@
...
 import styled from 'styled-components';
 import { __, sprintf, translateToExclusiveList } from '@web-stories-wp/i18n';
 import {
-  MEDIA_VARIANTS,
+  Link,
   Text,
   THEME_CONSTANTS,
+  Icons,
 } from '@web-stories-wp/design-system';
+import { useMediaPicker } from '@web-stories-wp/wp-story-editor/src/components/mediaUpload/mediaPicker';
```
swissspidy: This needs to be removes as per my other comment

packages/wp-story-editor/src/api/media.js
```diff
@@ -59,6 +59,10 @@ export function getMedia(
     apiPath = addQueryArgs(apiPath, { cache_bust: true });
   }
 
+  if (include) {
+    apiPath = addQueryArgs(apiPath, { include });
```
swissspidy: Is this still needed? `getPublisherLogos` returns all publisher logos + their URLs

merapi: Nah, I just didn't revert that.

packages/story-editor/src/components/panels/document/publish/publish.js
```diff
@@ -116,73 +158,115 @@ function PublishPanel() {
...
+        {displayText}
+      </Text>
+    );
+  };
 
-    return message;
-  }, [allowedImageFileTypes]);
+  const renderUploadButton = (open) => (
+    <Option onClick={open} aria-label={__('Add new', 'web-stories')}>
+      <Icons.ArrowCloud height={32} width={32} />
```
miina, line 246: Isn't 32x32 the default for icons btw?

merapi: It will still grow depending on the context, without it: 
image

swissspidy (approved):
Seems to work well 👍 🚀

spacedmonkey (dismissed):
I am seeing an error when I try and set as an author. 

https://user-images.githubusercontent.com/237508/135874128-8f114b64-bc45-40e8-a6f6-d4e74d131bec.mov

CC @swissspidy

swissspidy:
@spacedmonkey Image cropping for authors also appears to be broken on `main` and is not related to this PR. So this warrants its own ticket.

swissspidy:
@merapi The `should show cards that require the hasUploadMediaAction permission` karma test appears to be failing in this PR.

swissspidy:
> @spacedmonkey Image cropping for authors also appears to be broken on `main` and is not related to this PR. So this warrants its own ticket.

Looks like it's because authors can only edit/crop attachments they have uploaded themselves, but not ones uploaded by other users.

spacedmonkey (approved):

miina:
> The should show cards that require the hasUploadMediaAction permission karma test appears to be failing in this PR.

@swissspidy Looks like choosing the poster image was broken -- not sure if it came from merge issues or something else but I changed it back to how it is in `main` and now it seems to work both in the editor and for tests (locally at least, waiting for the tests to run here, too): c5c1a79

PR #8990, opened 2021-09-09
Decoupling: Move `PostLock` and `StatusCheck` to `wp-story-editor`

mohdsayed:
## Context

Move `PostLock` and `StatusCheck` to `web-story-editor`

## Summary

- [x] Move `PostLock` component to `wp-story-editor` and export its dependencies.
- [x] Move `StatusCheck ` component to `wp-story-editor` and export its dependencies.
- [x] Move `wp:*` items to `getStoryById` and `getDemoStoryById` callbacks ( then to `wp-story-editor` )
- [x] Remove usage of `_embeded` from the core editor API response and use consistent names for fields.
- [x] Update tests
- [x] Remove `PostLock` and `StatusCheck` related API callbacks from core editor.

## Testing Instructions

1.

## Checklist

---


mohdsayed:
@swissspidy I am thinking about these WordPress specific `wp:*` items, any thoughts?

https://github.com/google/web-stories-wp/blob/3558944c06fe1c0303d35b823f3b8f4f8d7571d1/packages/story-editor/src/app/story/effects/useLoadStory.js#L81-L101

swissspidy:
@sayedtaqui I think this logic can be easily moved to `getDemoStoryById` / `getStoryById`. They could parse `embedded[...]` and the capabilities and jut provide the parsed objects for use in `useLoadStory()`. Does that make sense?

packages/story-editor/src/app/api/apiProvider.js
```diff
@@ -80,6 +80,49 @@ function APIProvider({ children }) {
...
+        links,
+        'wp:action-publish'
+      ),
+      hasAssignAuthorAction: Object.prototype.hasOwnProperty.call(
+        links,
+        'wp:action-assign-author'
+      ),
+    };
+
+    post.lock_user = {
```
swissspidy: I think we can easily use camelCase for all these vars. `lockUser`, `featuredMedia`, etc.

packages/story-editor/src/app/api/apiProvider.js
```diff
@@ -80,6 +80,49 @@ function APIProvider({ children }) {
     withoutImages: [],
   });
 
+  // Add custom fields and prepare story response.
+  const prepareGetStoryResponse = (post) => {
```
swissspidy: Nit: I'd use the verb `transform`

packages/story-editor/src/app/api/apiProvider.js
```diff
@@ -204,6 +247,14 @@ function APIProvider({ children }) {
         path,
         data: getStorySaveData(story),
         method: 'POST',
+      }).then((data) => {
+        const { _embedded: embedded = {} } = data;
+
+        if ('wp:featuredmedia' in embedded) {
+          data.featured_media_url = embedded['wp:featuredmedia'][0].source_url;
```
swissspidy: Same thing here about camel cases.

Also wondering why we don't just update `featuredMedia` and instead have this separate field 🤔 

Seems redundant.

packages/story-editor/src/app/api/apiProvider.js
mohdsayed: I am thinking about the time when we document the expected response from these API callbacks. The fields should use the same pattern

https://github.com/google/web-stories-wp/blob/c0909f4495f80a12edb97f0ad8c30e8059ddac45/packages/story-editor/src/app/story/effects/useLoadStory.js#L47-L60

That's the same reason why and to make the API response look more platform agnotic, I have removed `_embeded` and `_links` because they look more WordPress specific. Thoughts?

packages/story-editor/src/app/api/apiProvider.js
mohdsayed: https://github.com/google/web-stories-wp/pull/8990#discussion_r708094816

packages/story-editor/src/app/api/apiProvider.js
swissspidy: > I have removed `_embedded` and `_links` because they look more WordPress specific

Yeah and that makes total sense.

I was just suggesting using camelCase here so we don't do have to rename from `lock_user` to `lockUser` everywhere else. We could save this step by just using `lockUser` everywhere.

packages/story-editor/src/app/api/apiProvider.js
swissspidy: Right, that's for the camel case. But the `featured_media_url` vs. `featuredMedia` part seems like a bug (not in this PR, but also on `main`). It's a bit confusing, to say the least.

swissspidy: What I mean is, we use `featuredMedia` everywhere, for example in `useLoadStory`.

But in `useSaveStory` we update `featuredMediaUrl` instead of `featuredMedia`

packages/story-editor/src/app/api/apiProvider.js
mohdsayed: > I was just suggesting using camelCase here

I think I could not understand. So if I use a camelCase for `lock_user` here inside [.then( transformGetStoryResponse )](https://github.com/google/web-stories-wp/blob/f311d45e15e9faa6d3e08e77067e92ec8ba38057/packages/wp-story-editor/src/api/story.js#L83) that would mean the expected API response fields from the callbacks would look something like this

```
          title: { raw: title },
          status,
          slug,
          date,
          modified,
          excerpt: { raw: excerpt },
          link,
          story_data,
          permalink_template,
          style_presets,
          password,
          preview_link,
          edit_link,
          embed_post_link,
          lockUser
```

Where `lockUser` is using camel case while the rest of the params are using `_`.

🤔

sayedtaqui: It will surely save us one step but API we would document later would look something like this.

image

packages/story-editor/src/app/api/apiProvider.js
mohdsayed: Is this what you are referring to?
https://github.com/google/web-stories-wp/pull/8990/commits/f65324a6433be2fe281e797ad81be3b071f60efb

packages/story-editor/src/app/story/effects/useLoadStory.js
```diff
@@ -58,51 +56,33 @@ function useLoadStory({ storyId, shouldLoad, restore, isDemo }) {
...
-        const lockUser = {
-          id: embedded?.['wp:lockuser']?.[0].id || 0,
-          name: embedded?.['wp:lockuser']?.[0].name || '',
-          avatar: embedded?.['wp:lockuser']?.[0].avatar_urls?.['96'] || '',
+        const defaultMedia = {
+          id: 0,
+          height: 0,
+          width: 0,
+          url: '',
         };
```
spacedmonkey: This const could be moved out of the hook for better performance.

packages/wp-story-editor/src/api/story.js
```diff
@@ -29,6 +29,52 @@ import apiFetch from '@wordpress/api-fetch';
...
+    name: embedded?.['wp:lockuser']?.[0].name || '',
+    avatar: embedded?.['wp:lockuser']?.[0].avatar_urls?.['96'] || '',
+  };
+
+  post.featured_media = {
+    id: embedded?.['wp:featuredmedia']?.[0].id || 0,
+    height: embedded?.['wp:featuredmedia']?.[0]?.media_details?.height || 0,
+    width: embedded?.['wp:featuredmedia']?.[0]?.media_details?.width || 0,
+    url: embedded?.['wp:featuredmedia']?.[0]?.source_url || '',
+  };
```
spacedmonkey, line 63: This logic is repeated a lot. Worth it's own function. Something like 

```
function getEmbeddedMediaObject( embedded, field ) {
    return {
          id: embedded?.[field]?.[0].id || 0,
          height: embedded?.[field]?.[0]?.media_details?.height || 0,
          width: embedded?.[field]?.[0]?.media_details?.width || 0,
          url: embedded?.[field]?.[0]?.source_url || '',
  };
}
```

swissspidy: IIRC we discussed this very recently and my take was that using optional chaining like here is better for readability.

packages/wp-story-editor/src/api/story.js
```diff
@@ -111,6 +157,19 @@ export function saveStoryById(story, stories, encodeMarkup) {
...
+    const { _embedded: embedded = {} } = data;
+
+    if ('wp:featuredmedia' in embedded) {
+      data.featured_media = {
+        id: embedded?.['wp:featuredmedia']?.[0].id || 0,
+        height: embedded?.['wp:featuredmedia']?.[0]?.media_details?.height || 0,
+        width: embedded?.['wp:featuredmedia']?.[0]?.media_details?.width || 0,
+        url: embedded?.['wp:featuredmedia']?.[0]?.source_url || '',
+      };
+    }
```
swissspidy: I'd remove the `if()` here. Then we don't need the `featuredMedia = featured_media ?? defaultMedia` check in `useLoadStory`

```suggestion
    data.featured_media = {
      id: embedded?.['wp:featuredmedia']?.[0].id || 0,
      height: embedded?.['wp:featuredmedia']?.[0]?.media_details?.height || 0,
      width: embedded?.['wp:featuredmedia']?.[0]?.media_details?.width || 0,
      url: embedded?.['wp:featuredmedia']?.[0]?.source_url || '',
    };
```

packages/story-editor/src/app/story/effects/useLoadStory.js
```diff
@@ -58,51 +56,33 @@ function useLoadStory({ storyId, shouldLoad, restore, isDemo }) {
...
-            embedded?.['wp:publisherlogo']?.[0]?.media_details?.height || 0,
-          width: embedded?.['wp:publisherlogo']?.[0]?.media_details?.width || 0,
-          url: embedded?.['wp:publisherlogo']?.[0]?.source_url || '',
-        };
-
-        const taxonomies =
-          links?.['wp:term']?.map(({ taxonomy }) => taxonomy) || [];
-        const terms = embedded?.['wp:term'] || [];
+        const featuredMedia = featured_media ?? defaultMedia;
+        const publisherLogo = publisher_logo ?? defaultMedia;
```
swissspidy: `transformGetStoryResponse` ensures that this object always exists, so no need for this check here

swissspidy (approved):
Can simplify the featuremedia/publisherlogo handling, but otherwise LGTM

PR #8991, opened 2021-09-09
Always focus on the element when clicking on it

### LINKED ISSUE #8981: Selecting a media element on a page does not cause (re)focus
## Bug Description

Selecting a non-background media element on a page does not cause a blur on the old focused element and a focus on the newly selected element. It *does* happen if you select a background media element or another type of element (text or shape).

This is particularly a problem if you click on a thumbnail in the carousel, then click on an image on the newly shown page and press <kbd>backspace</kbd> to delete the element. Because focus never left the thumbnail, the thumbnail still has keyboard focus and the entire page is deleted.

## Expected Behaviour

* A selected element should always receive keyboard focus
* In this particular set of events, the thumbnail should lose focus (the selection rectangle should turn gray) and only the element should be deleted, not the whole page.

## Steps to Reproduce

1. Create a story with multiple pages and a non-background image each
2. Click a non-active page in the carousel
3. Observe a blue outline around the newly selected thumbnail page in the carousel indicating it has keyboard focus
4. Click the image element on the new page
5. Observe that the blue outline remains on the selected thumbnail in the carousel
6. Press <kbd>backspace</kbd> in an attempt to delete the selected element
7. Observe that the whole page is deleted rather than the element

## Screenshots

image-select-no-focus

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

miina:
## Summary
Updates the logic so that elements gets focus on even if it wasn't previously selected.

## Testing Instructions

I
1. Add 2 pages, on the second page, add some media.
2. While being on the second page, unselect all elements.
3. Focus on the thumbnail of the page (by clicking on it)
4. Now click on one of the images and hit "Delete".
5. Verify that the Image got deleted (and not the Page, as previously)

II
1. Add 2 pages, on the second page, add one media element and one shape.
2. Select the shape.
3. Now click on the Media element and press "Delete"
4. Verify that the Image got deleted (and not the Shape, as previously).

## Checklist

---

Fixes #8981

### DIFF
--- a/packages/story-editor/src/components/canvas/frameElement.js
+++ b/packages/story-editor/src/components/canvas/frameElement.js
@@ -145,11 +145,10 @@ function FrameElement({ element }) {
     !mask?.type || (isBackground && mask.type !== MaskTypes.RECTANGLE);
   const eventHandlers = {
     onMouseDown: (evt) => {
-      if (isSelected) {
-        elementRef.current.focus({ preventScroll: true });
-      } else {
+      if (!isSelected) {
         handleSelectElement(id, evt);
       }
+      elementRef.current.focus({ preventScroll: true });
       if (!isBackground) {
         evt.stopPropagation();
       }


barklund (approved):
Not completely sure exactly why this fixes it, but it _does_ work!

PR #6546, opened 2021-02-26
Performance: Split editor/dashboard bundles into chunks with webpack

dreamofabear:
## Context

Split editor and dashboard bundles into chunks for shared code/modules. Related to #868.

## Summary

Reduces total load of both editor and dashboard by 26% from 1.45MB to 1.07MB (gzip).

- Dedupes common `node_modules` between editor/dashboard into `vendors~edit-story~stories-dashboard` (2.28MB / 258KB gzip)
- Dedupes common JSON (fonts, templates) into shared chunks (2.4MB / 208KB gzip)

### Before

Screen Shot 2021-02-26 at 3 41 33 PM
Screen Shot 2021-02-26 at 3 41 53 PM

### After

Screen Shot 2021-02-26 at 3 46 17 PM

## Relevant Technical Choices

1. Combine editor/dashboard webpack configs and use [`splitChunks`](https://webpack.js.org/guides/code-splitting/#splitchunksplugin) to analyze shared code and split into chunks
2. Use [`html-webpack-plugin`](https://github.com/jantimon/html-webpack-plugin) to output chunk manifest of each bundle as a PHP snippet (example below)
3. Include the PHP snippet in `Dashboard.php` and `Story_Post_Type.php` to register/enqueue each chunk script into WP

#### assets/js/edit-story.chunks.php
```php
<?php return array(
    'css' => ["vendors~edit-story"],
    'js' => ["vendors~edit-story~stories-dashboard-374bdd1db54db5dfc664","vendors~edit-story-981a678a2f0719a0921b"]);
```

## To-do

- [ ] Consider moving `assets/js/*.[asset|chunks].php` files into a new `assets/php/` directory

## User-facing changes

None, except faster loading going from dashboard -> editor.

### QA

This PR can be tested by following these steps:

1. Sanity check that this doesn't accidentally break any dashboard or editor functionality.

### UAT

- [x] UAT should use the same steps as above.

### Does this PR have a security-related impact?

No.

### Does this PR change what data or activity we track or use?

No.

### Does this PR have a legal-related impact?

No.

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testiing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
[...]

PR #8995, opened 2021-09-10
I18N: Fix loading of JavaScript translations

### LINKED ISSUE #8994: JavaScript translations not being loaded
## Bug Description

I happened to do some testing with localizations (using `pt_BR` and `de_DE`) and realized that basically none of the UI strings in the dashboard and editor are localized, despite the plugin currently being ~50% translated into these two locales.

The main reason for that is code splitting, which we began to make heavy use of recently.

To implement JavaScript localization, WordPress extracts translatable strings from all JavaScript files in a plugin (which we have a ton of in `assets/js`). For each JavaScript file, there is 1 corresponding translation file.

When registering scripts in WordPress, one has to use `wp_set_script_translations()` so that WordPress knows to look for translation files for that script.

(Learn more at https://pascalbirchler.com/internationalization-in-wordpress-5-0/)

There are two issues with how we utilise this function in our plugin:

First, the chunks we register for each main script are not localized due to a missing `wp_set_script_translations()` call.

https://github.com/google/web-stories-wp/blob/9d18d5f52f1b8a88f170df6be099b0a9d69e40c2/includes/Assets.php#L124-L134

Second, we use dynamic imports for many of the chunks, e.g. all the templates.
Since they are all imported dynamically, WordPress doesn't know that these scripts exist and that they have localizations to be loaded. WordPress also doesn't offer a way to import localizations dynamically together with the actual script import.

For this reason, we need to somehow load all translations for these dynamic imports upfront.

## Expected Behaviour

Translations for the dashboard and editor should be loaded correctly.

## Steps to Reproduce

1. Switch your user locale to `pt_BR` or `de_DE` in WordPress
2. Ensure translations are up-to-date under Dashboard -> Updates
3. Open Web Stories dashboard and editor
4. See that all texts are in English

## Additional Context

- Plugin Version: 1.11.0
- WordPress Version: 5.8
- Operating System: macOS
- Browser: Chrome

swissspidy:
## Context

As per #8994, JavaScript translations are currently not being loaded.

## Summary

This PR ensures that chunks from dynamic imports are registered as scripts in WordPress, without actually enqueueing them. The chunks' translations will be printed whenever the main script is enqueued.

## Relevant Technical Choices

* Add missing `wp_set_script_translations()` call for chunks
* Register chunks from dynamic imports as scripts and print their translations whenever the main script is enqueued.

## To-do

- As per discussion below, some more predictable file names might be valuable.

## User-facing changes

After manually renaming some translation files as per below instructions, the UI is localized again:

Screenshot 2021-09-10 at 11 44 28

## Testing Instructions

While it's technically a user-facing change (UI should be translated again), it's not currently easily possible to test this change.

Why?

To implement JavaScript localization, WordPress extracts translatable strings from all JavaScript files in a plugin. For each JavaScript file, there is 1 corresponding translation file, which contains the MD5 hash of the JS file path.

The problem is that our JS files have hashes in their file names.

So when you download translations from WordPress.org, the translation file name will be `web-stories-de_DE-<hash of the JS file from v1.11.0>.json`, but the hash will be different when building the plugin from `main`. `web-stories-de_DE-<hash of the JS file from main branch>.json` does not exist of course.

The only way to test this is by manually renaming some of the JSON translation file names in `wp-content/languages/plugins` with updated MD5 hashes.

## Checklist

---

Fixes #8994


includes/Assets.php
```diff
@@ -144,6 +152,38 @@ public function register_script_asset( string $script_handle, array $script_depe
...
+
+			$output = <<<JS
+( function( domain, translations ) {
+	var localeData = translations.locale_data[ domain ] || translations.locale_data.messages;
+	localeData[""].domain = domain;
+	wp.i18n.setLocaleData( localeData, domain );
+} )( "web-stories", {$json_translations} );
+JS;
+
+			wp_add_inline_script( $script_handle, $output );
```
spacedmonkey: Could / should we break this crunch into its own method?

includes/Assets.php
```diff
@@ -94,11 +94,13 @@ public function get_asset_metadata( string $handle ): array {
...
-		$asset['dependencies'] = $asset['dependencies'] ?? [];
 		// A hash calculated based on the file content of the entry point bundle at <$handle>.js.
 		$asset['version'] = $asset['version'] ?? WEBSTORIES_VERSION;
-		$asset['js']      = $chunks['js'] ?? [];
-		$asset['css']     = $chunks['css'] ?? [];
+
+		$asset['dependencies'] = $asset['dependencies'] ?? [];
+		$asset['js']           = $chunks['js'] ?? [];
+		$asset['css']          = $chunks['css'] ?? [];
+		$asset['chunks']       = $chunks['chunks'] ?? [];
```
spacedmonkey, line 104: These lines are moved just to tidy up the code right?

includes/Assets.php
spacedmonkey: Can't we just call [print_translations](https://github.com/WordPress/wordpress-develop/blob/4205214eba6a467b2dd97e866327e3d17a5c6a0a/src/wp-includes/class.wp-scripts.php#L591) here?

spacedmonkey:
> there is 1 corresponding translation file, which contains the MD5 hash of the JS file path.

Couldn't we just change so that crunks are name predictably and use the file hash for the script version. So you would end up with something like this. 

`file-chunk-1.js?v=<md5>`. I mentioned this idea in the original PR - https://github.com/google/web-stories-wp/pull/6546#issuecomment-787834581

includes/Assets.php
swissspidy: Yes

includes/Assets.php
swissspidy: > Can't we just call [print_translations](https://github.com/WordPress/wordpress-develop/blob/4205214eba6a467b2dd97e866327e3d17a5c6a0a/src/wp-includes/class.wp-scripts.php#L591) here?

No. `print_translations()` returns a `<script>` tag. `wp_add_inline_script()` does not allow `<script>` tags. It requires the plain JS.

swissspidy:
> > there is 1 corresponding translation file, which contains the MD5 hash of the JS file path.
> 
> Couldn't we just change so that chunks are name predictably and use the file hash for the script version. So you would end up with something like this.
> 
> `file-chunk-1.js?v=<md5>`. I mentioned this idea in the original PR - [#6546 (comment)](https://github.com/google/web-stories-wp/pull/6546#issuecomment-787834581)

Yeah that would help with the path matching, but only to some degree I think. Our JS file names also change often for other reasons, for example when we move code around and then webpack decides to split up chunks differently.

More predictable file names would help to avoid this scenario:

1. Install plugin with
2. Install language pack
3. Have translated UI
4. Update plugin
5. No new language pack available yet, so UI won't be translated.

That said, this PR is still needed, regardless of file naming.

includes/Assets.php
spacedmonkey: > No. print_translations() returns a <script> tag. wp_add_inline_script() does not allow <script> tags. 

If you pass $echo = false, it doesn't print / wrap in a script tag.

includes/Assets.php
```diff
@@ -144,6 +152,38 @@ public function register_script_asset( string $script_handle, array $script_depe
...
+				$json_translations = '{ "locale_data": { "messages": { "": {} } } }';
+			}
+
+			$output = <<<JS
+( function( domain, translations ) {
+	var localeData = translations.locale_data[ domain ] || translations.locale_data.messages;
+	localeData[""].domain = domain;
+	wp.i18n.setLocaleData( localeData, domain );
+} )( "web-stories", {$json_translations} );
+JS;
```
spacedmonkey: This worked in my testing. 

```suggestion
			$output = wp_scripts()->print_translations( $dynamic_chunk, false );
```

includes/Assets.php
swissspidy: How could I miss this 🤦 I wrote this stuff.

swissspidy: Fixed now

spacedmonkey (approved):
Tested with spanish pack installed

webpack.config.cjs
```diff
@@ -232,11 +232,35 @@ const templateContent = ({ htmlWebpackPlugin }) => {
...
 
-  return `<?php return array(
-    'css' => ${JSON.stringify(css)},
-    'js' => ${JSON.stringify(js)});`;
+  // We're only interested in chunks from dynamic imports;
+  // ones that are not already in `js` and not primaries.
+  const chunks = chunkNames.filter(
+    (chunk) =>
+      !js.includes(chunk) &&
+      !['wp-dashboard', 'wp-story-editor'].includes(chunk)
```
dreamofabear: Nit: would be nice to dedupe the `wp-dashboard` and `wp-story-editor` strings now that they're used outside of the single object literal below.

```js
const EDITOR_CHUNK = 'wp-story-editor';
const DASHBOARD_CHUNK = 'wp-dashboard';
```

includes/Assets.php
```diff
@@ -121,17 +123,23 @@ public function register_script_asset( string $script_handle, array $script_depe
...
-				$script_chunk,
-				$base_script_path . $script_chunk . '.js',
+				$chunk,
+				$base_script_path . $chunk . '.js',
 				[],
 				WEBSTORIES_VERSION,
 				$in_footer
 			);
+
+			wp_set_script_translations( $chunk, 'web-stories' );
```
dreamofabear: Should we just do this in `register_script()`?

dreamofabear (approved):
Sorry for missing this originally, thanks!

PR #8996, opened 2021-09-10
Color Picker: Propagate dimensions when switching mode

### LINKED ISSUE #8673: Color picker can overflow viewport when switching state
## Bug Description

When switching from custom color picker to basic color picker, the popup window extends and this can cause the popup to move slightly off-screen.

This can also happen if you add a new custom color in the custom view as the picker switches to the basic state, but the list of saved colors now extends by one more row.

## Expected Behaviour

Popup should always be inside the visible viewport area.

## Steps to Reproduce

1. Make sure to have 7+ saved global colors.
2. Open the new basic color picker
3. Switch to custom color picker mode
4. Scroll the design panel so the bottom of the color picker aligns with the bottom of the viewport
5. Switch to basic color picker mode

## Screenshots

propagation

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

---

miina:
## Summary
Ensures part of the color picker doesn't get hidden when switching views.

## User-facing changes
cpicker

## Testing Instructions

1. Make sure to have 7+ saved global colors.
2. Add a shape and assign radial color from the Custom picker mode.
3. Close the Color Picker
4. Scroll so that the Color Input in the design panel would be in the bottom part of the screen.
5. Now open the color picker for the radial color.
6. Switch to Basic Color Picker view 
7. Verify that all the lower parts of the Color Picker are visible (and not out of the screen), including the "Custom" button.

## Checklist

---

Fixes #8673

### DIFF
--- a/packages/story-editor/src/components/colorPicker/colorPicker.js
+++ b/packages/story-editor/src/components/colorPicker/colorPicker.js
@@ -72,13 +72,20 @@ function ColorPicker({
   allowsSavedColors = false,
   onClose = () => {},
   changedStyle = 'background',
+  onDimensionChange = () => {},
 }) {
   // If initial color is a gradient, start by showing a custom color picker.
   // Note that no such switch happens if the color later changes to a gradient,
   // only if it was a gradient at the moment the color picker mounted.
   const [isCustomPicker, setCustomPicker] = useState(hasGradient(color));
-  const showCustomPicker = useCallback(() => setCustomPicker(true), []);
-  const hideCustomPicker = useCallback(() => setCustomPicker(false), []);
+  const showCustomPicker = useCallback(() => {
+    setCustomPicker(true);
+    onDimensionChange();
+  }, [onDimensionChange]);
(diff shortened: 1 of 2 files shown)


packages/story-editor/src/components/colorPicker/colorPicker.js
```diff
@@ -72,13 +72,20 @@ function ColorPicker({
   allowsSavedColors = false,
   onClose = () => {},
   changedStyle = 'background',
+  onDimensionChange = () => {},
```
barklund, line 75: Does it need a default? Isn't it always set (except maybe for some tests)?

barklund (approved):
🌕

packages/story-editor/src/components/colorPicker/colorPicker.js
miina: Yep, it's not set in some tests. Probably doesn't harm, having a default.

PR #5792, opened 2020-12-22
Infrastructure: Update build tooling to use webpack v5

swissspidy:
## Summary

Updates webpack from v4 to v5 which contains a ton of _**major**_ changes, which means there are major changes to our build tooling as well.

On the plus side, there are significant **bundle size and build performance improvements**.

Many webpack plugins haven't kept up yet, so this PR is a draft for the time being.

**webpack 5 migration guide:** https://webpack.js.org/migrate/5/

**storybook + webpack 5 migration guide:** https://storybook.js.org/blog/storybook-for-webpack-5/

## Relevant Technical Choices

* **Prevent stylis plugin warnings**  
  Stylis has been throwing warnings because the stylis RTL plugin doesn't have a "name" (anymore?). Fixed this by adding `Object.defineProperty(stylisRTLPlugin, 'name', { value: 'stylisRTLPlugin' });` everywhere, which seems to be the recommended approach.

* **~Replace usage of `global.*` with `window.*`~**  
  This was split out into #10487 

* **~Replace usage of `process.env` with custom variables~**  
  This was split out into #10487 

* **Replaces `optimize-css-assets-webpack-plugin` with `css-minimizer-webpack-plugin`**
  The former does not support webpack 5, and recommends the latter going forward.

* **Replaces `url-loader` and `file-loader` with [asset modules](https://webpack.js.org/guides/asset-modules/)**  
  This changes the way SVG imports are written.

* **Replaces `expose-loader` with the `mangleExports` option**

* **Updates webpack config to remove deprecated fields and maintain existing behavior**
  The webpack config schema has been updated significantly. This PR removes things like `jsonpFunction` (deprecated) and adds things like `resolve.fullySpecified` (allows importing ES modules without file extensions)

* **Changes how Karma tests are run** ⚠️ 
  `karma-webpack` needed to be updated to its next major version (`5.0.0`) in order to support webpack 5. This meant changing the way we run tests, because the "alternative usage" option we've been following is **no longer supported**. This essentially reverts #2887. However, I don't see a negative performance impact. **On the plus side:** this should allow running individual tests again.

* **Update storybook to use webpack 5**  
  Since there are still upstream issues like https://github.com/storybookjs/storybook/issues/14877, this contains a couple of workarounds that we can hopefully remove soon.

* Addresses various deprecation warnings and other minor code quality issues

## To-do

* Port some of the webpack changes to the Rollup config

## User-facing changes

N/A

## Testing Instructions

Verify that all tests still pass, and editor still works.

[...]

PR #9001, opened 2021-09-10
I18N: Replace `@wordpress/i18n` with Tannin

### LINKED ISSUE #8577: I18N: Use Tannin instead of `@wordpress/i18n`
## Feature Description

The story editor and dashboard use `@web-stories-wp/i18n`, which currently is a simple wrapper around `@wordpress/i18n` plus some useful utility functions.

We did this in #6183 in order to remove a direct dependency on a GPL-licensed (but externalized!) dependency across tons of files, so that we could eventually swap out `@wordpress/i18n` for something else.

It's something we _have to_ to before being able to distribute the editor & dashboard package without this dependency being externalized.

And this "something else" is [Tannin](https://github.com/aduth/tannin), the library that also powers `@wordpress/i18n` but is MIT-licensed.

Since Tannin essentially offers the same API as `@wordpress/i18n`, no changes to `story-editor` & `dashboard` need to be made.

## Alternatives Considered

* Just wait for the Gutenberg package to be dual-licensed.

## Additional Context

---

## Implementation Brief

In `@web-stories-wp/i18n`:

Instantiate a `new Tannin` instance. Export `__`, `_x`, etc. helper functions acting as wrappers around `dcnpgettext`, so that the API stays the same for consumers. Also export `sprintf` from `@tannin/sprintf`. Perhaps also export a `setLocaleData` method to set `tannin.data` from outside.

In `wp-story-editor` and `wp-dashboard`:

Option A:

Get locale data from `@wordpress/i18n` using `getLocaleData` and pass this to our Tannin instance.

Unfortunately `getLocaleData` is only available as of very recently and does not exist on WP 5.5.

Option B:

Don't use `wp_set_script_translations` but use something custom [like Site Kit does](https://github.com/google/site-kit-wp/blob/d6ea05554e20353df9e770abd193c98b006cdf79/includes/Core/Assets/Script.php) to pass locale data to our JavaScript via some inline script.

swissspidy:
## Context

See #8577 for context

## Summary

Replace `@wordpress/i18n` with Tannin, the underlying library used by it.

## To-do

- [x] Ensure string extraction still works (i18n lint should pass)
- [x] Reduce bundle size if possible
- [x] Ensure  that `wp-i18n` script is not unnecessarily enqueued
- [x] Fix inline script calls for all scripts _and_ chunks

## User-facing changes

N/A

## Testing Instructions

1. Switch your site/user language to something like Portuguese (Brazil)
2. Go to Dashboard -> Updates to ensure all translations are installed
3. Go to the story dashboard and the story editor and verify that the majority of strings are translated.

## Checklist

---

Fixes #8577


includes/Admin/Editor.php
```diff
@@ -340,6 +340,7 @@ public function get_editor_settings(): array {
 				'encodeMarkup'                 => $this->decoder->supports_decoding(),
 				'metaBoxes'                    => $this->meta_boxes->get_meta_boxes_per_location(),
 				'ffmpegCoreUrl'                => trailingslashit( WEBSTORIES_CDN_URL ) . 'js/@ffmpeg/core@0.10.0/dist/ffmpeg-core.js',
+				'localeData'                   => json_decode( load_script_textdomain( self::SCRIPT_HANDLE, 'web-stories' ), true ),
```
spacedmonkey: Should this be moved to a trait or another class or something?

swissspidy: Probably not worth it for this 1 line of code.

swissspidy:
@sayedtaqui How familiar are you with webpack? The replacement is basically done, but I am having troubles getting the JS output into [a form supported by the WP-CLI string extraction](https://github.com/wp-cli/i18n-command/blob/01d6bd9c7933d67b87cc041620a4a5918914c633/src/JsFunctionsScanner.php#L257-L349).

For example, with `@wordpress/i18n`, code looked like `Object(u.__)( "translation" )`, which was recognizable due to the `__` being kept, but with this PR I'm only getting things like `Object(u.b)( "translation" )`

mohdsayed:
@swissspidy Just have a decent knowledge of Webpack, but will surely look at it next week.

mohdsayed:
> or example, with @wordpress/i18n, code looked like Object(u.__)( "translation" ), which was recognizable due to the __ being kept, but with this PR I'm only getting things like Object(u.b)( "translation" )

@swissspidy Just had a quick look. If it triggers any thought, it gets resolves if we pass `usedExports: false` to `optimization` in `webpack.config.cjs`. Will try to debug this week.

swissspidy:
Hmm I _think_ I've tried that option in my debugging (among dozens of other attempts).

[The description](https://v4.webpack.js.org/configuration/optimization/#optimizationusedexports) would fit perfectly:

> export names are mangled to single char identifiers when all usages are compatible

I'm gonna try it again.

My only worry there is that it bloats up bundle size, but we'll see.

swissspidy:
Yay, i18n lints are passing with this 🎉 

Now let's see the bundle size change...

mohdsayed:
It's temporary trade for ~ +53.2 kB until we find a better solution, but the good thing is that at least we aren't blocked by it.
I am wondering how Gutenberg didn't have this issue.

swissspidy:
> +102 kB (+3%)

That's really not ideal 😕 

> I am wondering how Gutenberg didn't have this issue.

I think for Gutenberg it's different because they expose the whole package under the `wp.i18n` global, so webpack can't just mangle export names there.

mohdsayed:
> I think for Gutenberg it's different because they expose the whole package under the wp.i18n global, so webpack can't just mangle export names there.

@swissspidy 
Will it make sense to expose `i18n` functions in global objects too. That should solve the problem though. Like this https://github.com/google/web-stories-wp/pull/9132

mohdsayed:
Though I know it would be weird for only i18n, but just throwing out ideas.

webpack.config.cjs
```diff
@@ -204,6 +204,9 @@ const sharedConfig = {
           output: {
             comments: /translators:/i,
           },
+          mangle: {
+            reserved: ['__', '_x', '_n', '_nx']
```
mohdsayed: I also tried this today, does it have any effect?

swissspidy: Not really, no. Or at least not the desired effect. Because `usedExports` does tree shaking _and_ mangling of exports, so we'll get output like `Object(i18n["a" /* __ */])('Direction', 'web-stories')` that is then passed to Terser.

That would actually be fine, since the i18n tooling recognizes this format.

But then Terser removes the `/* __ */` comments and does all sorts of mangling and we'll end up with something more like `Object(c.a)("Direction","web-stories")`

swissspidy:
After some more digging, I _think_ what we really need here is the [`mangleExports`](https://webpack.js.org/configuration/optimization/#optimizationmangleexports) option. It seems to do exactly what we need. Sadly, it's webpack v5+ only, so blocked by #5792.

`usedExports` is important for tree shaking, so I definitely want to keep that.

Exposing a global variable should be OK for now.

webpack.config.cjs
```diff
@@ -197,7 +227,7 @@ const sharedConfig = {
           // they're _likely_ component names, and these are useful to have
           // in tracebacks and error messages.
           keep_fnames: /__|_x|_n|_nx|sprintf|^[A-Z].+$/,
-          output: {
+          format: {
```
swissspidy: Whoops, need to undo this.

swissspidy:
I think I finally got it working! 🎉

mohdsayed (approved):

packages/i18n/package.json
```diff
@@ -10,9 +10,11 @@
   },
   "main": "./src/index.js",
   "type": "module",
+  "sideEffects": false,
```
spacedmonkey, line 13: Why is this needed?

packages/i18n/src/i18n.js
```diff
@@ -0,0 +1,100 @@
...
+
+const TEXT_DOMAIN = 'web-stories';
+
+const tannin = new Tannin({
+  [TEXT_DOMAIN]: {
+    '': {
+      plural_forms(n) {
+        return n === 1 ? 0 : 1;
+      },
+    },
```
spacedmonkey, line 30: Can you explain this?

spacedmonkey (changes requested):
I couldn't see this working locally. 
#### In main with Spanish
Screenshot 2021-11-01 at 11 50 49

#### This branch with Spanish
Screenshot 2021-11-01 at 11 50 00

packages/i18n/package.json
swissspidy: This helps webpack to optimize imports, see https://webpack.js.org/configuration/optimization/#optimizationsideeffects

packages/i18n/src/i18n.js
swissspidy: This is just the default configuration needed for Jed / Tannin

swissspidy:
> I couldn't see this working locally.
> 
> #### In main with Spanish
> Screenshot 2021-11-01 at 11 50 49
> 
> #### This branch with Spanish
> Screenshot 2021-11-01 at 11 50 00

Have you updated your translations under Dashboard -> Updates?

swissspidy:
@spacedmonkey I really cannot reproduce this

Have you ensured your language in the user profile is set to Spanish (or Site Default)?

Are translations up to date?

Have you built the JS & PHP?


Everything's properly translated here:

Screenshot 2021-11-01 at 17 47 04

spacedmonkey (approved):
Updating the translations, fixed my issue. Approved!

PR #9023, opened 2021-09-13
Disable video transcoding for non-WP media 

### LINKED ISSUE #8982: Disable video transcoding for non-WP media
## Feature Description

Video trimming and muting should only be allowed for videos that reside on the WP installation itself, and not for Coverr videos nor hotlinked videos.

Muting is currently not an option for Coverr videos, as they're always without sound by default, but hotlinked media could have audio and should not be mutable.

Video trimming should always be disabled for any non-WP videos similarly.

### Implementation

One simple way to check this is to see if `!isNaN(element.resource.id)`, as Coverr elements have `id="media/coverr:XXX"` and hotlinked videos don't have an id at all.

A more future-proof way would be to add a new flag to the resource object when adding external content (both from the third-party library and via hotlinking) and add a migration for this (setting the flag on old stories based on the above id check).

spacedmonkey:
## Summary

Add `isExternal` attribute to resource not hosted by WordPress. This includes, hotlinks assets, coverr, unsplash and tenor. 
Add a migration for existing image/gif/video elements, to add this new attribute. 

#### Before 
Screenshot 2021-09-13 at 23 19 38

#### After 
Screenshot 2021-09-13 at 23 21 08

## Testing Instructions

1. Open editor.
2. Add hotlink video into a page - https://vjs.zencdn.net/v/oceans.mp4
3. Ensure that mute / trim options are not visual. 

## Checklist

---

Fixes #8982


packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
```diff
@@ -214,17 +214,8 @@ const useQuickActions = () => {
...
-    const is3PGif =
-      (!idOrigin || idOrigin?.toLowerCase() === 'media/tenor') &&
-      selectedElements?.[0]?.resource?.type?.toLowerCase() === 'gif';
-    const is3PVideo = idOrigin?.toLowerCase() === 'media/coverr';
-    const is3PImage = idOrigin?.toLowerCase() === 'media/unsplash';
-
-    const panelToFocus =
-      is3PImage || is3PVideo || is3PGif ? states.MEDIA3P : states.MEDIA;
+    const isExternal = selectedElements?.[0]?.resource?.isExternal;
+    const panelToFocus = isExternal ? states.MEDIA3P : states.MEDIA;
```
swissspidy: Interesting, didn't know about this code.

Looks like it's used to highlight the Media3p tab when using quick actions on a media3p image/video.

The thing is, with this change it will also the highlight the Media3p tab for hotlinked media, which is incorrect.

So this change should be reverted or perhaps replaced by a new `isMedia3p` flag on resources (in addition to `isExternal`)

packages/media/src/createResource.js
```diff
@@ -57,6 +57,7 @@ import normalizeResourceSizes from './normalizeResourceSizes';
  * @property {boolean} local Whether the resource has been already uploaded to the server.
  * @property {boolean} isOptimized Whether the resource has already been optimized.
  * @property {boolean|null} isMuted Whether the resource has already been muted.
+ * @property {boolean|null} isExternal Whether the resource is externally hosted.
```
swissspidy, line 69: also need to update the `Resource` JSDoc a bit below

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
swissspidy: An alternative would be to simplify this check like this:

```js
const is3pMedia = resourceId.startsWith('media/');
const panelToFocus = is3pMedia ? states.MEDIA3P : states.MEDIA;
```

packages/migration/src/migrations/test/v0034_markVideoAsExternal.js
```diff
@@ -0,0 +1,837 @@
...
+              {
+                _test: 'element2',
+                type: 'image',
+                resource: {
+                  type: 'image',
+                  mimeType: 'image/png',
+                  src: 'https://example.com/image.png',
+                  width: 265,
+                  height: 527,
+                  isExternal: true,
```
swissspidy:
```suggestion
```

Need to remove this so we can actually test that `isExternal: true` is added by the migration

swissspidy (approved):
QA instructions needed, otherwise looks good

spacedmonkey:
> QA instructions needed, otherwise looks good

QA instructions added

packages/media/src/createResource.js
```diff
@@ -99,6 +101,10 @@ import normalizeResourceSizes from './normalizeResourceSizes';
  * @property {string|null} lengthFormatted The formatted length for the "video" type.
  * @property {string|null} alt The user-readable accessibility label for the resource.
  * @property {boolean} local Whether the resource has been already uploaded to the server.
+ * @property {boolean} isOptimized Whether the resource has already been optimized.
+ * @property {boolean|null} isMuted Whether the resource has already been muted.
+ * @property {boolean|null} isExternal Whether the resource is externally hosted.
+ * @property {boolean|null} isPlaceholder Whether the resource is a placeholder.
```
swissspidy: I don't think these are ever `null` because they have default values in the function signature below.

```suggestion
 * @property {boolean} isMuted Whether the resource has already been muted.
 * @property {boolean} isExternal Whether the resource is externally hosted.
 * @property {boolean} isPlaceholder Whether the resource is a placeholder.
```

swissspidy:
@spacedmonkey FYI there's a merge conflict now.

spacedmonkey:
> FYI there's a merge conflict now. 

Fixed

swissspidy:
aaand a lint error too apparently

swissspidy:
@googlebot I consent

Issue #8952, opened 2021-09-06
Decoupling: Force `isSaving`/`isEnabled` state when saving meta boxes


swissspidy:
## Background

The WordPress story editor supports a WordPress-specific feature called "custom meta boxes" that allows other WordPress plugins to integrate with the editor by showing their custom UI for customizing some story settings. This is particularly useful for SEO plugins like All in One SEO or Yoast SEO, as users can configure some SEO metadata from within the editor this way.
See #5048 for a demo.

Due to the way custom meta boxes work in WordPress, we need hook into the saving process of the editor and save any custom meta boxes in a separate HTTP request.

So basically, whenever the story has just been saved (but it wasn't an autosave), we save the meta boxes as well.

The interesting part:

Whether we are saving the story itself _or_ the meta boxes, we show a loading spinner next to the publish buttons. This happens here:

https://github.com/google/web-stories-wp/blob/9ab1bd62ab46ed47362ffe4db30a2b78beb8c416/packages/story-editor/src/app/story/storyProvider.js#L167

Similarly, we always allow updating the story when there are any custom meta boxes, even if the story itself hasn't changed:

https://github.com/google/web-stories-wp/blob/9ab1bd62ab46ed47362ffe4db30a2b78beb8c416/packages/story-editor/src/components/header/buttons/update.js#L70-L74

## Proof-of-concept

Recently, @sayedtaqui has been working on extracting all custom meta boxes logic from `packages/story-editor` into `packages/wp-story-editor`.

See https://github.com/rtCamp/web-stories-wp/pull/763 for a POC.

Now, solving the `isSaving` and `isEnabled` parts is really the last remaining piece for finalizing this.

With all custom meta boxes functionality moved to `wp-story-editor`, how can we force the `isSaving` state while meta boxes are saving, without having any meta boxes logic within `story-editor` itself? And how can we enforce `isEnabled` if there are any meta boxes present?

## Context

Gutenberg has faced the same question and has basically solved this by moving the whole header area with the publish/update button to the `edit-post` package (which is basically `wp-story-editor`), but keep the individual components to assemble the header in the `editor` package (aka `story-editor`), with `forceIsSaving` props to override the state.

Source:

https://github.com/WordPress/gutenberg/blob/408170a9b4007857166b41041aceedb1cef012c8/packages/edit-post/src/components/header/index.js#L66-L82

[...]

PR #9027, opened 2021-09-14
Decoupling: Move meta boxes to `wp-story-editor`

mohdsayed:
## Context

Move meta boxes to `wp-story-editor` and refactor code for decoupling the story-editor

## Relevant Technical Choices

- It would probably be better to have a central store in a different package like Gutenberg for maintaining some global states like `isSaving`, so they can be accessed and modified by both `story-editor` and`wp-story-editor`. Though the current approach of assembling the header components and overriding their `isSaving` logic would work for us for now. 
Currently `isSaving` is being used in the following areas/components. We have got the first 5 items covered in this PR, however we would either need to revisit and think of the last two items or we think about better solution as mentioned above.

1. Loader 
2. Update Button
3. Preview Button
4. Publish Button
5. SaveDraftButton
6. Author selection dropdown.
7. StoryAMPValidationErrors

- Use Slot/Fill for appending metabox and its menu items instead of querying the DOM before using React Portal.

## To-do

- [x] Export `Layout` as `InterfaceSkeleton` and use in `wp-story-editor`
- [x] Export all header items.
- [x] Assembel header pieces in `wp-story-editor`
- [x] Drill down `header` prop from `InterfaceSkeleton` to `NavLayer` component from the integration.
- [x] Update `Buttons` component unit test and integration test.
- [x] Remove unused code [titleMissingDialog](https://github.com/google/web-stories-wp/blob/main/packages/story-editor/src/components/header/titleMissingDialog.js)
- [x] Pass metaboxes component from `<InterfaceSkeleton>` children.
- [x] Remove metabox menu item from MenuItem component and use Slot/Fill or an ID for it.
- [x] Insert Menu Item from the `wp-story-editor`.
- [x] Move `useSaveMetaBoxes` to wp-story-editor.
- [x] Handle `isSaving` and `isEnabled` in `wp-story-editor`.
- [x] Move metaboxes related API callbacks to `wp-story-editor`.
- [x] Manual testing the functionality

## Testing Instructions

Test if
- Meta boxes fields saving - Yes
- When saving in progress, is the preview button disabled: Yes
- Is a loader visible in the meta box when saving in progress: Yes

<!DOCTYPE html><figure class="md-table-fig" cid="n186" mdtype="table" style="box-sizing: border-box; margin: 1.2em 0px; overflow-x: auto; max-width: calc(100% + 16px); padding: 0px; cursor: default; caret-color: rgb(51, 51, 51); color: rgb(51, 51, 51); font-family: &quot;Open Sans&quot;, &quot;Clear Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, sans-serif; font-size: 14px; font-style: normal; font-variant-caps: normal; font-weight: normal; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-size-adjust: auto; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration: none;">

Action | Loading State  ( main/current branch ) | Enabled State  ( main/current branch )
-- | -- | --
Click on Publish Button | True/True | True/True
Click on Update Button | True/True | True/True
Click on Switch To Draft Button | True/True | True/True
Click on Draft Button | True/True | True/True

</figure>

## Checklist

---

See https://github.com/google/web-stories-wp/issues/8952


packages/wp-story-editor/src/components/header/buttons/index.js
```diff
@@ -0,0 +1,103 @@
...
+    ({
+      state: {
+        story: { status, embedPostLink, link },
+        meta: { isFreshlyPublished },
+      },
+    }) => ({
+      status,
+      embedPostLink,
+      link,
+      isFreshlyPublished,
```
swissspidy: Since these are not actually used, we can just as well remove them.

packages/story-editor/src/index.js
```diff
@@ -19,11 +19,13 @@
  */
 import StoryEditor from './storyEditor';
 import Dialog from './components/dialog';
+import { default as InterfaceSkeleton } from './components/layout';
```
swissspidy:
```suggestion
export { default as InterfaceSkeleton } from './components/layout';
```

?

packages/wp-story-editor/src/components/header/buttons/index.js
mohdsayed: I think I found some more code not being used like https://github.com/google/web-stories-wp/blob/main/packages/story-editor/src/components/header/titleMissingDialog.js

swissspidy: Oh wow, yeah that can be totally removed! 😅

packages/wp-story-editor/src/components/metaBoxes/metaBoxesProvider.js
```diff
@@ -44,21 +48,27 @@ function MetaBoxesProvider({ children }) {
   const state = useMemo(
     () => ({
       state: {
+        isSavingMetaBoxes,
```
swissspidy, line 48: Where are `isSavingMetaBoxes` and `setIsSavingMetaBoxes` actually used? `useSaveMetBoxes` has its own state for that?

packages/wp-story-editor/src/components/metaBoxes/metaBoxesProvider.js
```diff
@@ -44,21 +48,27 @@ function MetaBoxesProvider({ children }) {
...
         metaBoxesVisible,
         metaBoxes,
         locations,
         hasMetaBoxes,
+        apiUrl,
```
swissspidy: Where is that being used? `useSaveMetaboxes` gets it straight from `useConfig`

packages/wp-story-editor/src/components/metaBoxes/metaBoxes.js
```diff
@@ -13,29 +13,68 @@
...
     metaBoxesVisible: state.metaBoxesVisible,
   }));
-
+  const [showMenuButton, updateMenuButtonState] = useState(false);
+  const menuButtonContainer = useRef(null);
   const { postType, metaBoxes = {} } = useConfig();
+  const { pages } = useStory(({ state: { pages } }) => ({ pages }));
+
+  useEffect(() => {
+    let timeout = null;
```
swissspidy: Nit:

```suggestion
    let timeout;
```

packages/wp-story-editor/src/components/metaBoxes/metaBoxes.js
```diff
@@ -13,29 +13,68 @@
...
+
+    if (hasMetaBoxes && pages.length >= 0) {
```
swissspidy: Did you perhaps mean `pages.length > 0`?

If length is `0` it means the story hasn't fully loaded in the editor yet.

packages/wp-story-editor/src/components/metaBoxes/metaBoxesProvider.js
mohdsayed: Not being used anywhere, thanks.

packages/wp-story-editor/src/components/metaBoxes/metaBoxesProvider.js
mohdsayed: `useSaveMetBoxes` state definition was left in doing back and forth of code. Removed from `useSaveMetBoxes` in https://github.com/google/web-stories-wp/pull/9027/commits/cde6cce47015e688b6f722aac5b4533016a5e84e 
Thank you

mohdsayed:
@swissspidy  

( To be done later )
What if we export all panels from Document and assemble them in `wp-story-editor` like below

```js
import {
  PublishPanel,
  ExcerptPanel,
  SlugPanel,
  StatusPanel,
  PageAdvancementPanel,
  BackgroundAudioPanel,
  TaxonomiesPanel,
} from '@web-stories-wp/story-editor';

const Document = () => {
  function DocumentInspector() {
    const enabledTaxonomies = useFeature('enableTaxonomiesSupport');
    return (
      <>
        <StatusPanel />
        <PublishPanel />
        <ExcerptPanel />
        <SlugPanel />
        <PageAdvancementPanel />
        <BackgroundAudioPanel />
        {enabledTaxonomies ? <TaxonomiesPanel /> : null}
      </>
    );
  }
}

function Layout() {
  return (
    <MetaBoxesProvider>
      <InterfaceSkeleton
        header={<Header />}
        inspector={ {
            document: {
              component: <Document />,
            },
          }
        }
      >
        <MetaBoxes />
      </InterfaceSkeleton>
    </MetaBoxesProvider>
  );
}

```

That will help us in a couple of ways

1. We would have more control over individual components like `PublishPanel`, so we can easily pass more configuration options ( props ) to handle `isSaving` or passing labels, etc.
2. It would be possible to remove/add panels from a third-party integration.

mohdsayed:
Just a thought that may be later the entire inspector panel may be configured from the third party integration using the following. 

```js
      <InterfaceSkeleton
        header={<Header />}
        inspector={ {
          design: {
            title: __( 'Design', 'web-stories' ),
            component: <Design />
          },
          document: {
            title: __( 'Document', 'web-stories' ),
            component: <Document />,
          },
          ...
        } }
      >
```

packages/wp-story-editor/src/components/metaBoxes/useSaveMetaBoxes.js
```diff
@@ -17,45 +17,47 @@
...
+ * @param {boolean} props.isAutoSavingStory Whether autosaving is in progress.
  * @return {{isSavingMetaBoxes: boolean}} Metaboxes status.
  */
-function useSaveMetaBoxes({ story, isSaving, isAutoSaving }) {
-  const { hasMetaBoxes, locations } = useMetaBoxes(({ state }) => ({
-    hasMetaBoxes: state.hasMetaBoxes,
-    locations: state.locations,
-  }));
+function useSaveMetaBoxes({ story, isSavingStory, isAutoSavingStory }) {
+  const { hasMetaBoxes, locations, isSavingMetaBoxes, setIsSavingMetaBoxes } =
```
swissspidy, line 39: Is it weird that this hook re-exports `isSavingMetaBoxes`?

swissspidy:
@sayedtaqui Good idea! It's [similar to what Gutenberg does](https://github.com/WordPress/gutenberg/blob/164bef81386e96330128c6218a66f29c9ed8e865/packages/edit-post/src/components/sidebar/settings-sidebar/index.js), but of course there the `block-editor` package is more low-level than our `story-editor`. 

I opened #9070 to track & discuss this future enhancement

swissspidy (approved):
One question but otherwise LGTM

packages/wp-story-editor/src/components/metaBoxes/useSaveMetaBoxes.js
mohdsayed: Ah, that totally slipped my mind, thank you! That also led to some code simplification. Done in https://github.com/google/web-stories-wp/pull/9027/commits/ea11f5f953ffb7de53e29c6bdeaa374926d775c6

swissspidy:
Ah, there are now quite a few merge conflicts after #8990 got in.

packages/story-editor/src/app/api/apiProvider.js
```diff
@@ -158,6 +160,12 @@ function APIProvider({ children }) {
     [currentUser, updateCurrentUser]
   );
 
+  // @todo Move to wp-story-editor along with StatusCheck component.
+  actions.getStatusCheck = useCallback(
```
swissspidy: Accidentally added back?

mohdsayed: 😬

mohdsayed:
@swissspidy Fixed the test, the PR is ready!

packages/wp-story-editor/src/components/metaBoxes/menuItem/index.js
```diff
@@ -0,0 +1,97 @@
...
+  align-items: center;
+`;
+
+const Space = styled.span`
+  width: 8px;
+`;
+
+const Item = styled.div`
+  order: -1;
+`;
```
swissspidy: Hmm this really messes with the tab order when navigating with the keyboard. The DOM structure is now slightly different too, causing a slight layout:

Before:

Screenshot 2021-09-22 at 10 14 20

After:

Screenshot 2021-09-22 at 10 14 30


See regression on Percy: https://percy.io/google/web-stories-wp-e2e/builds/12843630/changed/725855111

See how order of elements changed in the DOM:

Before:

Screenshot 2021-09-22 at 10 18 06

After:

Screenshot 2021-09-22 at 10 18 37

mohdsayed: In that case, without Slot/Fill, we will have to keep a `<AdditionalMenuItems />` area inside `<MenuItems>` so the menu items can be inserted in the same order.

Now

single-item-menu-order


If required In the future more menu items can also be inserted inside `<AdditionalMenuItems />` and the tab should still work

order

mohdsayed: Done in https://github.com/google/web-stories-wp/pull/9027/commits/784b712c99125b55847152bb2412953b8011000f 

Does it make sense?

sayedtaqui: Before

before



After:

after

mohdsayed:
Don't know why percy is complaining, I used https://www.diffchecker.com/image-diff/ and there seems to be no difference in the menu items.

Before & After Difference

Before-After

swissspidy:
@googlebot I consent

swissspidy:
> Don't know why percy is complaining, I used [diffchecker.com/image-diff](https://www.diffchecker.com/image-diff/) and there seems to be no difference in the menu items.
> 
> Before & After Difference
> 
> Before-After

There was some small difference because that spacer span was not inside `MenuItems` anymore but nested inside `MenuItem` within `additional-primary-menu-items` where it has no effect
The zoom dropdown was also a bit smaller because of this.

Pushing some changes now to simplify the menu items alignment a bit.

swissspidy (approved):

PR #9031, opened 2021-09-14
Media: Ensure that muted / trimmed videos keep their file type

### LINKED ISSUE #8998: Media: Ensure that muted / trimmed videos keep their file type
## Feature Description

Ensure that muted / trimmed videos return same file type as give. There is no need to convert a webm to an mp4. 

Context https://github.com/google/web-stories-wp/pull/8972#discussion_r705253445

spacedmonkey:
## Summary
Ensure that muted / trimmed videos return same file type as give. There is no need to convert a webm to an mp4.

## Testing Instructions

1.

## Checklist

---

Fixes #8998


packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -264,10 +265,17 @@ function useMediaUploadQueue() {
           // TODO: Only transcode & optimize video if needed (criteria TBD).
           // Probably need to use FFmpeg first to get more information (dimensions, fps, etc.)
           if (isTranscodingEnabled && canTranscodeFile(file)) {
+            const ext = getFileExtFromUrl(resource.src);
```
swissspidy: Why do we need `getFileExtFromUrl` when we can just derive the file extension from the mimeType? 🤔

packages/story-editor/src/app/media/utils/useFFmpeg.js
```diff
@@ -285,10 +285,12 @@ function useFFmpeg() {
    * @param {File} file Original video file object.
    * @param {string} start Time stamp of start time of new video. Example '00:01:02.345'.
    * @param {string} end Time stamp of end time of new video. Example '00:02:00'.
+   * @param {string} type Mime type of output file.
+   * @param {string} ext File extension of output file.
    * @return {Promise<File>} Transcoded video file object.
    */
   const trimVideo = useCallback(
-    async (file, start, end) => {
+    async (file, start, end, mineType, fileExt) => {
```
swissspidy:
```suggestion
    async (file, start, end, mimeType, fileExt) => {
```

packages/story-editor/src/app/media/utils/useFFmpeg.js
```diff
@@ -285,10 +285,12 @@ function useFFmpeg() {
    * @param {File} file Original video file object.
    * @param {string} start Time stamp of start time of new video. Example '00:01:02.345'.
    * @param {string} end Time stamp of end time of new video. Example '00:02:00'.
+   * @param {string} type Mime type of output file.
```
swissspidy:
```suggestion
   * @param {string} mimeType Mime type of output file.
```

packages/story-editor/src/app/media/utils/useFFmpeg.js
swissspidy: Isn't the mime type part of `file` already?

swissspidy: See https://developer.mozilla.org/en-US/docs/Web/API/File/type

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
spacedmonkey: Nope, an m4v and mp4 both have the mime type of video/mp4. 

This is also about just keeping the file name, if you named it wrong, as say .mov, this way we keep that.

swissspidy:
Looks like this needs a rebase now

packages/story-editor/src/app/media/utils/useFFmpeg.js
```diff
@@ -297,10 +298,10 @@ function useFFmpeg() {
...
         ffmpeg = await getFFmpegInstance(file);
 
-        // @todo: ensure that this function return the same file type submitted. https://github.com/google/web-stories-wp/issues/8998.
-        const tempFileName = uuidv4() + '.' + MEDIA_TRANSCODED_FILE_TYPE;
-        const outputFileName =
-          getFileName(file) + '-trimmed.' + MEDIA_TRANSCODED_FILE_TYPE;
+        const ext = fileExt || MEDIA_TRANSCODED_FILE_TYPE;
+        const type = file?.type || MEDIA_TRANSCODED_MIME_TYPE;
+        const tempFileName = uuidv4() + '.' + ext;
+        const outputFileName = getFileName(file) + '-trimmed' + ext;
```
spacedmonkey:
```suggestion
        const outputFileName = getFileName(file) + '-trimmed.' + ext;
```

packages/story-editor/src/app/media/utils/useFFmpeg.js
```diff
@@ -285,10 +285,11 @@ function useFFmpeg() {
    * @param {File} file Original video file object.
    * @param {string} start Time stamp of start time of new video. Example '00:01:02.345'.
    * @param {string} end Time stamp of end time of new video. Example '00:02:00'.
+   * @param {string} ext File extension of output file.
    * @return {Promise<File>} Transcoded video file object.
    */
   const trimVideo = useCallback(
-    async (file, start, end) => {
+    async (file, start, end, fileExt) => {
```
swissspidy: The parameter name should match the JSDoc name.

packages/story-editor/src/app/media/utils/useFFmpeg.js
```diff
@@ -285,10 +285,11 @@ function useFFmpeg() {
    * @param {File} file Original video file object.
    * @param {string} start Time stamp of start time of new video. Example '00:01:02.345'.
    * @param {string} end Time stamp of end time of new video. Example '00:02:00'.
+   * @param {string} ext File extension of output file.
```
spacedmonkey:
```suggestion
   * @param {string} fileExt File extension of output file.
```

packages/media/src/test/getFileExtFromUrl.js
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * Internal dependencies
+ */
+import getFileExtFromUrl from '../getFileExtFromUrl';
+
+describe('getFileExtFromUrl', () => {
+  it('should file ext', () => {
```
swissspidy:
```suggestion
  it('should find ext', () => {
```

packages/story-editor/src/app/media/utils/useFFmpeg.js
```diff
@@ -347,19 +344,18 @@ function useFFmpeg() {
...
       let ffmpeg;
 
       try {
         ffmpeg = await getFFmpegInstance(file);
-
-        const tempFileName = uuidv4() + '.' + MEDIA_TRANSCODED_FILE_TYPE;
-        const outputFileName =
-          getFileName(file) + '-muted.' + MEDIA_TRANSCODED_FILE_TYPE;
-
+        const ext = fileExt || MEDIA_TRANSCODED_FILE_TYPE;
```
swissspidy: Shouldn't `ext` fall back to the extension derived from `file?.type`? 🤔

spacedmonkey: See https://github.com/google/web-stories-wp/pull/9031#discussion_r708341160

swissspidy: No, what I mean is basically `const ext = fileExt || getTypeFromMime(file.type) || MEDIA_TRANSCODED_FILE_TYPE`.

`MEDIA_TRANSCODED_FILE_TYPE` should really only be the last resort when we don't have any file type information whatsoever. To prevent transcoding.

So basically

```js
const mimeType = file?.type || MEDIA_TRANSCODED_MIME_TYPE;
const ext = fileExt || getTypeFromMime(mimeType);
```

packages/media/src/test/getFileExtFromUrl.js
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * Internal dependencies
+ */
+import getFileExtFromUrl from '../getFileExtFromUrl';
+
+describe('getFileExtFromUrl', () => {
+  it('should find ext', () => {
```
swissspidy: Nit: could probably use `it.each` with a few more inputs & expected outputs

packages/story-editor/src/app/media/utils/useFFmpeg.js
spacedmonkey: Fixed in https://github.com/google/web-stories-wp/pull/9031/commits/20608c4e25d312014341e489bfd680c23ae80015

I went with this logic. 

```
const ext = fileExt || getTypeFromMime(file?.type) || MEDIA_TRANSCODED_FILE_TYPE;
```

So, if we can get file ext from url, use that, else, use the file?.type conditional to get type by mime. If the file is null or type doesn't exist, it can then fallback to `MEDIA_TRANSCODED_FILE_TYPE`.

packages/media/src/test/getFileExtFromUrl.js
spacedmonkey: Improved test https://github.com/google/web-stories-wp/pull/9031/commits/8e93ef0b84587d64ad46cdfc03f711ee48943a43

packages/media/src/test/getFileExtFromUrl.js
```diff
@@ -0,0 +1,42 @@
...
+    '#',
+    '-1',
+    '',
+  ])('should %s not find ext', (url) => {
+    expect(getFileExtFromUrl(url)).toStrictEqual('');
+  });
+
+  it('should null not find ext', () => {
+    expect(getFileExtFromUrl(null)).toBeNull();
+  });
```
swissspidy: Sorry for not being clear here. `it.each` could be used for _all_ assertions, passing input & output.

Example:

```suggestion
  it.each([
    ['https://www.example.com/test.jpg', 'jpg'],
    ['https://www.example.com/path/test.jpg', 'jpg'],
    ['test.jpg', 'jpg'],
    ['https://www.example.com/test', ''],
    ['https://www.example.com/', ''],
    ['#', ''],
    ['-1', ''],
    ['', ''],
    [null, null],
  ])('should get file extension from URL %s', (url, expectedOutput) => {
    expect(getFileExtFromUrl(url)).toStrictEqual(expectedOutput);
  });
```

packages/story-editor/src/app/media/utils/useFFmpeg.js
```diff
@@ -347,19 +345,19 @@ function useFFmpeg() {
...
       try {
         ffmpeg = await getFFmpegInstance(file);
-
-        const tempFileName = uuidv4() + '.' + MEDIA_TRANSCODED_FILE_TYPE;
-        const outputFileName =
-          getFileName(file) + '-muted.' + MEDIA_TRANSCODED_FILE_TYPE;
-
+        const ext =
+          fileExt || getTypeFromMime(file?.type) || MEDIA_TRANSCODED_FILE_TYPE;
+        const type = file?.type || MEDIA_TRANSCODED_MIME_TYPE;
```
swissspidy: If you reserve this it becomes a bit shorter:

```suggestion
        const type = file?.type || MEDIA_TRANSCODED_MIME_TYPE;
        const ext = fileExt || getTypeFromMime(type);

```

spacedmonkey: I explained why I did this why here. https://github.com/google/web-stories-wp/pull/9031#discussion_r709221541

TDLR, if type is null for some reason, it will fallback to `MEDIA_TRANSCODED_FILE_TYPE`.

packages/media/src/test/getFileExtFromUrl.js
spacedmonkey: I know, but the title of the test isn't super clear. Breaking the tests up makes for cleaner test titles.

packages/story-editor/src/app/media/utils/useFFmpeg.js
swissspidy: Unless I am missing something, my code and yours do exactly the same thing:

1. Get extension from URL
2. Fall back to extension derived from `file.type`
3. Fall back to extension derived from `MEDIA_TRANSCODED_MIME_TYPE` (mp4)

My code is just shorter and easier to read IMO.

spacedmonkey: I don't think we should be using `getTypeFromMime`. Mime !== file ext. 

Example 
- application/svg+xml - svg
- image/svg+xml - svg
- video/ogg - ogv
- video/3gpp- 3gp
- video/3gpp2 - 3g2

I know we don't support these file formats, but this just to prove that mime is not file extension. 

I think using it `getTypeFromMime` is extremely britel , if anything I would want to remove it's usage here. I like the usage of `MEDIA_TRANSCODED_FILE_TYPE` here, it is something we have more control. If we don't use `MEDIA_TRANSCODED_FILE_TYPE` let's use a function with a switch statement.

packages/media/src/getFileExtFromUrl.js
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+const getFileExtFromUrl = (url) =>
+  url && url.substring(url.lastIndexOf('/')).split('.').slice(1, 2).join('.');
+
+export default getFileExtFromUrl;
```
swissspidy:
```suggestion
/**
 * Returns the file extension from a given URL.
 *
 * @param {?string} url File URL.
 * @return {?string} File extension.
 */
const getFileExtFromUrl = (url) =>
  url?.substring(url?.lastIndexOf('/')).split('.').slice(1, 2).join('.');
```

packages/media/src/test/getFileExtFromUrl.js
swissspidy: The current titles aren't super clear either.

The spec title can easily be reworded, that should be the least problem. There are even many formatting options for that: https://jestjs.io/docs/api#testeachtablename-fn-timeout

A data provider like this makes it way easier to add more test cases.

I don't think it can get any more readable than this:

```suggestion
  it.each`
    url                                        | expected
    ${'https://www.example.com/test.jpg'}      | ${'jpg'}
    ${'https://www.example.com/path/test.jpg'} | ${'jpg'}
    ${'test.jpg'}                              | ${'jpg'}
    ${'https://www.example.com/test'}          | ${''}
    ${'https://www.example.com/'}              | ${''}
    ${'#'}                                     | ${''}
    ${'-1'}                                    | ${''}
    ${''}                                      | ${''}
    ${null}                                    | ${undefined}
  `('should return "$expected" for url "$url"', ({ url, expected }) => {
    expect(getFileExtFromUrl(url)).toStrictEqual(expected);
  });
```

packages/story-editor/src/app/media/utils/useFFmpeg.js
swissspidy: Just had to take another look at `getTypeFromMime` to realize that it returns the _element_ type from a mime type, not the _file_ type. 🤦 

So for `video/mp4` it returns `video`. So yeah we definitely shouldn't use it, albeit for a different reason.

But my point still stands: switching the two lines avoids some repetition.

Anyhow, resolving this as it's not a big deal.

packages/story-editor/src/app/media/utils/useFFmpeg.js
```diff
@@ -297,10 +298,11 @@ function useFFmpeg() {
       try {
         ffmpeg = await getFFmpegInstance(file);
 
-        // @todo: ensure that this function return the same file type submitted. https://github.com/google/web-stories-wp/issues/8998.
-        const tempFileName = uuidv4() + '.' + MEDIA_TRANSCODED_FILE_TYPE;
-        const outputFileName =
-          getFileName(file) + '-trimmed.' + MEDIA_TRANSCODED_FILE_TYPE;
+        const ext =
+          fileExt || getTypeFromMime(file?.type) || MEDIA_TRANSCODED_FILE_TYPE;
```
swissspidy: So since `getTypeFromMime` returns an _element_ type instead of a _file_ type, we need a different solution.

Instead of a hardcoded list of mime types or a switch statement, could you explore using [`mime/lite`](https://www.npmjs.com/package/mime)? That seems like the most robust yet lightweight solution.

When we can simply do:

```js
import mime from 'mime/lite';

const fileExt = mime. getExtension(file.type)
```

packages/media/src/getFileExtFromUrl.js
spacedmonkey: Does this work?

swissspidy: Yes. I tested this locally before sharing. Feel free to test it as well.

spacedmonkey:
I decided to go a very different direction, by passing object into the config. This is what I plan to use as for mime / exts going forward.

swissspidy:
> I decided to go a very different direction, by passing object into the config. This is what I plan to use as for mime / exts going forward.

@spacedmonkey Let's please discuss such things first. I don't agree with your decision and I would have appreciated a response to [my suggestion](https://github.com/google/web-stories-wp/pull/9031#discussion_r710017505) first and then eventually find an agreement together (whatever that may look like).

spacedmonkey:
> Let's please discuss such things first. I don't agree with your decision and I would have appreciated a response to my suggestion first and then eventually find an agreement together (whatever that may look like).

Let's jump on a call tomorrow for this. This ticket has got out of hand and the backwards and forwards is not helpful.

swissspidy (approved):

PR #9040, opened 2021-09-15
Fetch page attachment link icon using API

### LINKED ISSUE #9028: Page Attachment: Automatically fetch link icon
## Feature Description

In #8578 we updated the page attachment CTAs. As per #8120, automatically fetching the link icon was a P1 so we omitted it initially.

But there was no follow-up ticket created for it, so I am doing that now :-)

It should be super trivial to add because we have all the components for this already.

miina:
## Summary
Fetches the icon of the Page Attachment via API and populates the field if icon found.

## Testing Instructions

1. Type in `stories.google` to the Page Attachment field.
2. Verify that icon is populated automatically (it will take a bit of time to fetch it)
3. Verify that when you save the Story and see the preview, the icon displays for the Page Attachment link.
4. Now type in something incorrect
5. Verify that "Invalid link" is displayed as a warning
6. Type in `google.com`
7. Verify that the previously populated icon is not there anymore.

## Checklist

---

Fixes #9028


packages/story-editor/src/components/panels/design/pageAttachment/pageAttachment.js
```diff
@@ -125,10 +146,6 @@ function PageAttachmentPanel() {
     [updateCurrentPageProperties, pageAttachment]
   );
 
-  const debouncedUpdate = useDebouncedCallback((props) => {
-    updatePageAttachment(props);
-  }, 300);
-
   const [isInvalidUrl, setIsInvalidUrl] = useState(
     !isValidUrl(withProtocol(url || '').trim())
   );
```
merapi:
```suggestion
  const [isInvalidUrl, setIsInvalidUrl] = useState(
    url && !isValidUrl(withProtocol(url || '').trim())
  );
```
Something like this, otherwise I have an error from the beginning (empty string is not considered valid by this library).
The same can be fixed in `link.js`

image

packages/story-editor/src/components/panels/design/pageAttachment/pageAttachment.js
```diff
@@ -106,14 +110,31 @@ function PageAttachmentPanel() {
     }
   }, [url, displayWarning]);
 
+  const debouncedUpdate = useDebouncedCallback((props) => {
+    updatePageAttachment(props);
+  }, 300);
```
merapi:
```suggestion
  const debouncedUpdate = useDebouncedCallback(updatePageAttachment, 300);
```
Saving a few bytes ;) (+declaration needs to be moved up)

merapi (commented):
The UI is a bit inconsistent, are we okay with that?

- we hide fields after the error is detected vs all fields visible when you re-open the form with invalid link
- we don't display the invalid link message, even if we know it's invalid (fresh form)

https://user-images.githubusercontent.com/21036224/133604553-0c94bf57-61fc-4294-9d18-eef28c590735.mov

miina:
> The UI is a bit inconsistent, are we okay with that?

That's the same as Link panel logic currently. I think there's something wrong with `isValidUrl` and it's not returning expected results, making it possible to know if the link is valid only when the API request has been made.

packages/story-editor/src/components/panels/design/pageAttachment/pageAttachment.js
miina: Ah, I thought I already fixed this :)

Thanks! Will adjust.

packages/story-editor/src/components/panels/design/pageAttachment/pageAttachment.js
```diff
@@ -125,12 +123,29 @@ function PageAttachmentPanel() {
...
+        setIsInvalidUrl(true);
+      })
+      .finally(() => {
+        setFetchingMetadata(false);
+      });
+  }, 1200);
 
   const [isInvalidUrl, setIsInvalidUrl] = useState(
-    !isValidUrl(withProtocol(url || '').trim())
+    url && !isValidUrl(withProtocol(url || '').trim())
```
barklund: Doesn't the `|| ''` take care of this? If not, we should just remove that part.

packages/story-editor/src/components/panels/design/pageAttachment/pageAttachment.js
```diff
@@ -139,9 +154,14 @@ function PageAttachmentPanel() {
   const handleChangeUrl = useCallback(
     (value) => {
       _setUrl(value);
-      debouncedUpdate({ url: value });
+      setIsInvalidUrl(false);
+      const urlWithProtocol = withProtocol(value.trim());
+      const valid = isValidUrl(urlWithProtocol);
```
barklund, line 159: Shouldn't the invalid URL flag be set if not valid here?

```suggestion
      const urlWithProtocol = withProtocol(value.trim());
      const valid = isValidUrl(urlWithProtocol);
      setIsInvalidUrl(!valid);
```

packages/story-editor/src/components/panels/design/pageAttachment/pageAttachment.js
```diff
@@ -180,6 +200,16 @@ function PageAttachmentPanel() {
   );
 
   const checkboxId = `cb-${uuidv4()}`;
+
+  let hint;
+  if (displayWarning || isInvalidUrl) {
```
barklund: This expression is used twice - move to a `const`?

barklund (approved):
Looks good, just a few minor nits.

packages/story-editor/src/components/panels/design/pageAttachment/pageAttachment.js
miina: `isValidUrl` doesn't seem to be working reliably so it doesn't work fully as expected when doing that. Tried that initially but it didn't have consistent behaviour like this.

PR #9041, opened 2021-09-15
Fix some prop type warnings

swissspidy:
Fixes some prop types warnings I've seen during development. Caused by incorrect prop type declarations and a change in how the publisher logo is stored (#8946)


spacedmonkey:
Unit test failing here seems valid

Issue #7739, opened 2021-05-26
Migrate to use  `<amp-story-auto-analytics>`


swissspidy:
## Feature Description

Going forward we might want to exclusively use [`<amp-story-auto-analytics>`](https://amp.dev/documentation/components/amp-story-auto-analytics/).

Perhaps we want to use it in conjunction with `amp-analytics` during a migration period.

**Considerations**

What about Site Kit? Site Kit currently prints an `<amp-analytics>` tag on stories. Should they also switch? Needs coordination.

## Additional Context

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

swissspidy:
Pinging @felixarntz for awareness. We might want to update Site Kit to print `<amp-story-auto-analytics>` instead of `<amp-story-analytics>`.

felixarntz:
@swissspidy Please keep me in the loop on whether this should be done or not. I've created a tracking issue on Site Kit side. What I'm mostly curious about is whether that new component has any drawbacks in using - for example the `amp-analytics` component allows for more flexibility in the events setup, even though it is missing the useful defaults that the new one provides.

swissspidy:
> @swissspidy Please keep me in the loop on whether this should be done or not. I've created a tracking issue on Site Kit side. What I'm mostly curious about is whether that new component has any drawbacks in using - for example the `amp-analytics` component allows for more flexibility in the events setup, even though it is missing the useful defaults that the new one provides.

One would not be able to customize the events anymore, but with auto analytics there's a guaranteed set of events configured out of the box. Which makes it easier to develop stories-specific analytics dashboards and the like, as we 100% know what data is being collected.

felixarntz:
> One would not be able to customize the events anymore, but with auto analytics there's a guaranteed set of events configured out of the box. Which makes it easier to develop stories-specific analytics dashboards and the like, as we 100% what data is being collected.

Hmm that sounds like a trade-off I'm not sure about what's better for Site Kit. I wonder whether it would be possible for the `amp-story-auto-analytics` component to still support an (optional) JSON configuration for _additional_ configuration and events. As it currently stands, I don't see enough of a benefit (since Site Kit & Web Stories take care of the whole event setup anyway) from what I know so far.

swissspidy:
Suggestion by @choumx in sprint planning meeting just now:

[...]

PR #9050, opened 2021-09-16
Add `<amp-story-auto-analytics>` migration option

### LINKED ISSUE #9045: Add initial `<amp-story-auto-analytics>` migration implementation
## Feature Description

While UX on #7739 is being finalized, we can already implement the actual migration under the hood.

### Implementation Brief

* Add new feature flag for `enableAutoAnalyticsMigration` (or something like that)
* Add a new database migration that sets a flag in the database for everyone with a non-empty tracking ID setting (`web_stories_ga_tracking_id`). We can call the flag `web_stories_using_legacy_analytics` or so. If no tracking ID is set, `web_stories_using_legacy_analytics` will be `false`.
* This new flag should be available via the RES API
* In the editor, in addition to all other settings, retrieve the value of `web_stories_using_legacy_analytics`
* If `web_stories_using_legacy_analytics` is `true`, display explanatory text (exact copy TBD, see #7739 for example) and a "Migrate" button below.
* Clicking on "Migrate" should set `web_stories_using_legacy_analytics` to `false` in the database and ideally show some success snackbar message
* In `Analytics.php`, check value of `web_stories_using_legacy_analytics` and either print a `amp-story-auto-analytics` tag or the regular on.  
  ```html
  <amp-story-auto-analytics gtag-id="UA-123456789-0"></amp-story-auto-analytics>
  ```

swissspidy:
## Context

We want to migrate users to the new `<amp-story-auto-analytics>` component, giving them an opt-in if they used the old configuration so far.

## Summary

This PR adds the initial implementation for the `<amp-story-auto-analytics>` migration, behind a feature flag.

See #7739

## User-facing changes

**New UI option:**

Screenshot 2021-09-16 at 22 54 10

**Demo:**

https://user-images.githubusercontent.com/841956/133685535-e4de6438-c2c8-4dc0-8b70-d50c17afd225.mov

## Testing Instructions

**Note:** before testing, ensure you have a tracking ID set.

**Note** To reset the state between tests, visit `wp-admin/options.php` and set `web_stories_using_legacy_analytics` to `1`. This will make the migration button appear again.

1. On a published story, verify that you see `<amp-analytics>` in the source code
1. Enable `Auto Analytics` experiment
2. Go to dashboard settings
3. See "Migrate" button under Analytics section
1. On a published story, verify that you now see `<amp-story-auto-analytics>` in the source code

## Checklist

---

Fixes #9045
See #7739


packages/dashboard/src/app/views/editorSettings/videoCache/index.js
```diff
@@ -74,10 +74,6 @@ export default function VideoCacheSettings({
...
 
 VideoCacheSettings.propTypes = {
-  isEnabled: PropTypes.bool.isRequired,
+  isEnabled: PropTypes.bool,
   updateSettings: PropTypes.func.isRequired,
 };
-
-VideoCacheSettings.defaultProps = {
-  isEnabled: false,
-};
```
spacedmonkey, line 83: Is this change related?

packages/dashboard/src/app/views/editorSettings/googleAnalytics/index.js
```diff
@@ -157,10 +185,50 @@ function GoogleAnalyticsSettings({
...
+        {usingLegacyAnalytics && isAutoAnalyticsMigrationEnabled && (
+          <>
+            <SettingSubheading
+              size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL}
+            >
+              <TranslateWithMarkup
+                mapping={{
+                  a: (
+                    <InlineLink
+                      href="https://example.com"
```
spacedmonkey, line 205: Let's add a todo here, so we don't forget to update this.

packages/dashboard/src/app/views/editorSettings/googleAnalytics/index.js
```diff
@@ -157,10 +185,50 @@ function GoogleAnalyticsSettings({
...
         <SettingHeading htmlFor="gaTrackingID" as="h3">
           {TEXT.SECTION_HEADING}
+          {usingLegacyAnalytics && isAutoAnalyticsMigrationEnabled && (
+            <StyledNotificationBubble notificationCount={1} isSmall />
+          )}
         </SettingHeading>
         <SettingSubheading size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL}>
           {siteKitDisplayText}
         </SettingSubheading>
+        {usingLegacyAnalytics && isAutoAnalyticsMigrationEnabled && (
```
spacedmonkey, line 195: This migration is only a WordPress thing right? This migration is only need for WordPress. Should we call this out and leave a todo for when break up the packages.

packages/dashboard/src/app/views/editorSettings/videoCache/index.js
swissspidy: Not related; I was seeing these prop types warning during development and updated it here while I was at it.

`isEnabled` has a default value of `false` in the function signature, so it doesn't neeed `isRequired` nor `defaultProps`.

includes/Settings.php
```diff
@@ -141,6 +148,18 @@ public function register() {
 			]
 		);
 
+		register_setting(
```
spacedmonkey, line 151: We can update the settings test to check this setting is registered.

packages/dashboard/src/app/views/editorSettings/googleAnalytics/index.js
```diff
@@ -223,8 +291,10 @@ function GoogleAnalyticsSettings({
 }
 
 GoogleAnalyticsSettings.propTypes = {
-  handleUpdate: PropTypes.func,
   googleAnalyticsId: PropTypes.string,
+  handleUpdateAnalyticsId: PropTypes.func,
+  usingLegacyAnalytics: PropTypes.bool,
```
spacedmonkey, line 297: Should we set default for `usingLegacyAnalytics` to false?

swissspidy: The value is coming from the settings provider and the default is already `false` in `defaultSettingsState`. So this is already covered.

packages/dashboard/src/app/views/editorSettings/googleAnalytics/index.js
swissspidy: Yes, it's WordPress-specific. Like a lot of other things in this component, e.g. the Site Kit information. No need for a todo in the code for this though, there will be an in-depth analysis of the dashboard as part of #2918 anyway.

packages/dashboard/src/app/views/editorSettings/googleAnalytics/index.js
spacedmonkey: If you had default to false, you wouldn't have had to update lots of the tests.

packages/dashboard/src/app/views/editorSettings/googleAnalytics/index.js
spacedmonkey: I think we should be adding todo to new functionality we know is WordPress. Why make more work for ourselves later to find it again?

swissspidy: Such a todo doesn't need to be in the code, it can be noted externally in a doc.

spacedmonkey (approved):
One nitpic, but otherwise,we are looking good!

spacedmonkey (approved):

PR #8958, opened 2021-09-06
Settings: Allow disabling archive page

spacedmonkey:
## Context

See #4862 for history / context on this matter.

## Summary

We want to allow users to disable the default /web-stories archive page.

## Relevant Technical Choices

As there are only two options for now, enabled and disabled, made this PR simpler and just made the dashboard setting a checkbox. If we have more options in the future, at that point change to a dropdown.

#### New setting in dashboard
Screenshot 2021-09-09 at 13 43 25
Screenshot 2021-09-09 at 13 43 35

#### Archive link option does not show for blocks
Screenshot 2021-09-06 at 17 04 35
Screenshot 2021-09-06 at 17 04 47
Screenshot 2021-09-06 at 17 04 55

#### When archive is disabled, link should not rendered on the front end
Screenshot 2021-09-06 at 17 05 07

#### Options should render in widget form to show archive links
Screenshot 2021-09-06 at 17 08 36

#### Archive url, should 404.
Screenshot 2021-09-06 at 17 05 28
 

## Testing Instructions

- [ ] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

**Note:** Ensure the "Disable Archive" experiment is enabled first. 

#### Test dashboard setting
1. Go to dashboard > settings. 
2. Ensure Stories Archive option is present. 
3. Disable / Enable archives, reload page, ensure setting is saved. 

#### Test archive is really disabled
1. Diable Stories Archive ( see above ). 
2. Go to /web-stories on the front end. 
3. Ensure page is a 404. 

#### Test widget presents settings are not rendered
1. Diable Stories Archive ( see above ). 
2. Enable classic widget plugin. 
3. Create a web stories widget. 
4. Ensure no options for archives are present. 
5. Save widget. 
6. Ensure no link archive is rendered on front end.

#### Test block settings are not rendered
1. Diable Stories Archive ( see above ). 
2. Create a new post
3. Create a web stories block
4. Select Latest stories.
5. Select any layout style
6. Ensure no options for archives are present. 
7. Save post. 
8. Ensure no link archive is rendered on front end.

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
[...]

Issue #4862, opened 2020-10-09
Allow customization of archives page


swissspidy:
## Feature Description

Right now we simply use the built-in WordPress archives for the `/web-stories/` site. Depending on the theme, this archive page can look very bleak as they don't usually accommodate for a visual format like stories. But because it's a built-in feature that's steered by the theme, the user has no control over its appearance.

To fix this, we could do it [like WooCommerce](https://docs.woocommerce.com/document/woocommerce-pages/) and give the user an option to define a custom archive page.

<img width="692" alt="woocommerce-page-setup" src="https://user-images.githubusercontent.com/841956/95609932-ce0c1000-0a5f-11eb-866c-1f3512968061.png">

If set, we disable the built-in archive and instead refer to that page as the archive.

They can then edit this custom archive page using Gutenberg and insert whatever content they'd like. For example a Latest Stories block.

## Additional Context

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

spacedmonkey:
In #4863 the rtCamp adds a [theme supports](https://github.com/rtCamp/web-stories-wp/blob/c852f9892a4d45b708ffaa3f632c265afc6733ba/includes/Traits/ThemeSupport.php#L41). 

Maybe we have some conditional logic, that allows developers to support improved archives, otherwise, fallback to the default   behavior we have now. Sadly this would make this feature opt-in but may fix some issues around styling etc.

swissspidy:
Yeah this might be something to look into, but that's another topic I think.

We'd still want to give the user full power for customizing the archive page to their liking.

Imagine editing the archive page in Gutenberg, adding some paragraph blocks and a "Latest Stories" block below.

So the user could basically choose between two options

```
Web Stories Archive Page:

[ ] Use default page generated by theme / WordPress / us

[ ] Use custom page
```

Aside: If theme developers want to have improved archives, it's probably easiest to just create a `archive-web-story.php` template file and improve the layout using the blocks provided by us.

spacedmonkey:
WordPress core also as a method of marking pages as special pages. The example is the privacy policy page. 

<img width="333" alt="Screenshot 2021-01-26 at 13 17 48" src="https://user-images.githubusercontent.com/237508/105850094-2ea46600-5fd9-11eb-9add-9e44cb546e1f.png">

[...]

Issue #4459, opened 2020-09-11
Make post type permalink base customizable


swissspidy:
## Feature Description

Sometimes users are already using `/stories/` for a page or through some other post type. To avoid conflicts with existing content, we can make our permalink base customizable.

For example, [WooCommerce](https://docs.woocommerce.com/document/permalinks/) does that too. Screenshots:

![2016-06-13-at-11-23](https://user-images.githubusercontent.com/841956/92895645-f5010300-f41b-11ea-8a5f-f6658bcb91a3.png)
![product-permalinks-settings](https://user-images.githubusercontent.com/841956/92895650-f5999980-f41b-11ea-8e77-45035cc29c27.png)

## Additional Context

Making this customizable means we need to flush rewrite rules every time this is changed, and make sure we never hardcode the permalink base anywhere.

This is done automatically if we also add this to the "Permalinks" settings screen in WordPress, like WooCommerce does.

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

westonruter:
Perhaps `/web-stories/` would be a safer default base?

swissspidy:
With #4501 this becomes less urgent I think

swissspidy:
Closing for now as it doesn't seem needed.

PR #9051, opened 2021-09-16
Add option for choosing custom archive page

### LINKED ISSUE #8812: Settings: Allow choosing custom archive page
## Feature Description

See #4862 for history / context on this matter.

We want to allow users to use a custom static page for the `/web-stories` archive page.

The idea is to add a new `Custom` option to the dropdown added in #8811.

**Custom**

Screenshot 2021-08-12 at 15 43 36

For convenience, a link to the current archive page can be displayed.

There's an autocomplete dropdown to select the page.

The custom page can be annotated as such in the posts list table:

Screenshot 2021-08-12 at 15 40 06

"Custom" would mean selecting a custom page for the archives and that users would simply edit that page in the block editor, using the Web Stories block if they want to.

If users create a custom page like this, there would be no pagination etc.

### Dashboard

* Add a third "Custom" option to the "Stories Archive" dropdown.
* Once that option is selected, display a dropdown with autocomplete to select an existing page in WordPress.   
  * There are existing autocomplete/dropdown components for this.
  * The pages can be searched via the REST API
* Once a page is selected, the page ID is stored in `wp_options`
* Ensure rewrite rules are flushed when this has been changed  
  Note: ideally we don't flush rewrite rules twice.

### Page list table

* Filter `display_post_states` to display a "Web Stories Page" label in the page list table in WP admin

### `Story_Post_Type`

* Set `has_archive` depending on whether the archive has been enabled/disabled/customized
  Pseudo code:
  ```php
  $archive_page_option    = get_option(...);
  $custom_archive_page_id = get_option(...);
  $has_archive            = true;
  if ( 'disabled' === $archive_page_option ) {
    $has_archive = false;
  } elseif ( 'custom' === $archive_page_option && get_post( $custom_archive_page_id ) ) {
    $has_archive = urldecode( get_page_uri( $custom_archive_page_id ) );
  }
  ```
  (WordPress supports this out of the box)

### Out of scope

Out of scope for this ticket:

* Changing URL structure for individual stories
* Any custom redirect from `/web-stories/` to custom archive page

swissspidy:
## Context

We want to allow users to use a custom static page for the /web-stories archive page. See #8812 and #4862 for full context.

## Summary

**Note:** Depends on #8958 being merged first.

This PR adds a new option to the Archive Page dropdown in the dashboard settings to allow users to choose a custom archive page.

## To-do

- Tests
- Code cleanup

In the future, it would be great to handle this scenario:

1. User has currently chosen page XYZ as the archive page
2. User goes to settings, sees link to `example.com/xyz` there
3. User changes setting to "Default"
4. Link is not updated to `example.com/web-stories/`

To fix this, the URL would need to be retrieved from the REST API instead of being passed via config.

## User-facing changes

Screenshot 2021-09-17 at 00 58 02

Screenshot 2021-09-21 at 19 20 45

## Testing Instructions

1. Enable archive page customizaton experiment
2. Go to story settings and choose between three archive page options (default, disabled, custom)
3. Verify that archive page URL is now the page URL, showing page contents instead of default archive
4. Visit page list table in wp-admin, see "Web Stories Archive Page" label next to the chosen custom page

No

No

No

## Checklist

---

Fixes #8812


includes/Story_Post_Type.php
```diff
@@ -121,7 +123,7 @@ public function on_plugin_deactivation( $network_wide ) {
 	 * @return \WP_Post_Type|\WP_Error
 	 */
 	public function register_post_type() {
-		$has_archive = 'default' === get_option( Settings::SETTING_NAME_ARCHIVE, 'default' );
+		$has_archive = $this->get_has_archive();
```
spacedmonkey: You could do this 
```
'has_archive'           => $this->get_has_archive(),
```

packages/dashboard/src/karma/fixture.js
```diff
@@ -87,6 +87,8 @@ const defaultConfig = {
     currentUser: '/wp/v2/users/me',
     fonts: '/web-stories/v1/fonts',
     settings: '/wp/v2/settings',
+    search: '/wp/v2/search',
+    pages: '/wp/v2/pages',
```
spacedmonkey, line 90: These fixtures are massively out of date, we haven't had a fonts api in years!

spacedmonkey:
To confirm, this doesn't do the clever part, that highjacking the WP_Query to force page to load?

packages/dashboard/src/karma/fixture.js
swissspidy: I totally forgot that we even had a fonts API!

swissspidy:
> To confirm, this doesn't do the clever part, that highjacking the WP_Query to force page to load?

Correct. If you select a custom page in the settings and then visit the post type archive (= the page's permalink), WordPress will actually automatically just display that page's content. That's perfect for our use case as that's exactly what we need.

Edit: weird, not seeing this behavior right now. Let me double check this.

swissspidy:
> To confirm, this doesn't do the clever part, that highjacking the WP_Query to force page to load?

Correct, yeah. While I have it working locally, I'll put this in a separate PR to make review easier.

spacedmonkey:
Question, if I reset back to none and then back to custom, it seems to keep the state, is this to be expected?

https://user-images.githubusercontent.com/237508/134023596-16c374f6-8371-4c27-b2ad-82aa30ade71d.mov

includes/Admin/Dashboard.php
```diff
@@ -393,6 +393,8 @@ public function get_dashboard_settings(): array {
 					'users'       => '/web-stories/v1/users/',
 					'templates'   => sprintf( '/web-stories/v1/%s/', $templates_rest_base ),
 					'settings'    => '/web-stories/v1/settings/',
+					'search'      => '/wp/v2/search/',
```
spacedmonkey: Why use search endpoint? What benefit does this give us over use just pages endpoints, like this?  
```http://localhost:8899/wp-json/wp/v2/pages?search=term```

**Comment by swissspidy** [comment]
> Question, if I reset back to none and then back to custom, it seems to keep the state, is this to be expected?

Yes, see the todo section in the PR description and the corresponding todo in the code.

File: packages/dashboard/src/app/api/test/useCategoriesApi.js [hunk 4]
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * External dependencies
+ */
+import { act, renderHook } from '@testing-library/react-hooks';
+
+/**
+ * Internal dependencies
+ */
+import useCategoriesApi from '../useCategoriesApi';
```
spacedmonkey: What is happening here?

packages/dashboard/src/app/views/editorSettings/archive/index.js
```diff
@@ -83,20 +162,52 @@ export default function ArchiveSettings({
...
+        {ARCHIVE_TYPE.CUSTOM === archive && (
+          <TextInputHelperText
+            size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL}
+          >
+            <Search
+              placeholder={__('Select page', 'web-stories')}
+              selectedValue={selectedPage}
+              options={searchOptions}
+              handleSearchValueChange={handleSearchChange}
+              emptyText={__('No options available', 'web-stories')}
```
spacedmonkey: Can we do better here? Like, no pages or something?

packages/dashboard/src/app/api/test/useCategoriesApi.js
swissspidy: Just some merge conflict because this file was actually removed. Fixed now.

includes/Admin/Dashboard.php
swissspidy: The search endpoint is a bit more lightweight, but I suppose I could achieve the same with the pages endpoint + using `?_fields`. I'll take a look.

spacedmonkey: Honest question, just trying to understand the benefits here. We are already pass fields [here](https://github.com/google/web-stories-wp/pull/9051/files#diff-f2fd97f362a91bb56a0137a84ae805d30529fa2624f88d764540044248c525e5R29) already.

swissspidy: Fixed now

swissspidy:
> > To confirm, this doesn't do the clever part, that highjacking the WP_Query to force page to load?
> 
> Correct, yeah. While I have it working locally, I'll put this in a separate PR to make review easier.

@spacedmonkey Decided to put it in this PR, since it turned out to be only 1 added method

includes/Story_Post_Type.php
```diff
@@ -340,4 +348,55 @@ public function update_archive_setting() {
...
+	 *
+	 * @param WP_Query $query Current query instance, passed by reference.
+	 * @return void
+	 */
+	public function pre_get_posts( WP_Query $query ) {
+		if ( ! is_string( $this->get_has_archive() ) ) {
+			return;
+		}
+
+		if ( $query->is_admin || ! $query->is_main_query() ) {
```
spacedmonkey, line 394: Don't you need to check is `is_post_type_archive( self:POST_TYPE_SLUG );` here?

swissspidy: Whoops, accidentally removed this 😅 🤦 

Fixed now.

BrittanyIRL (approved):
👍 🚢 🚀 🕺

includes/Story_Post_Type.php
```diff
@@ -340,4 +348,59 @@ public function update_archive_setting() {
...
+			return;
+		}
+
+		if ( ! $query->is_post_type_archive( self::POST_TYPE_SLUG ) ) {
+			return;
+		}
+
+		$custom_archive_page_id = get_option( Settings::SETTING_NAME_ARCHIVE_PAGE_ID );
```
spacedmonkey: What happens here if you delete the page?

packages/dashboard/src/karma/fixture.js
```diff
@@ -85,8 +85,9 @@ const defaultConfig = {
     stories: '/web-stories/v1/web-story',
     users: '/wp/v2/users',
     currentUser: '/wp/v2/users/me',
-    fonts: '/web-stories/v1/fonts',
     settings: '/wp/v2/settings',
+    search: '/wp/v2/search',
```
spacedmonkey:
```suggestion
```

includes/Story_Post_Type.php
swissspidy: This function bails early because `$this->get_has_archive()` won't be a string, since that method checks whether the page exists.

So if the page has been deleted, the default archive behavior will be restored.

includes/Story_Post_Type.php
spacedmonkey: In my testing, this didn't work. 
```
! $query->is_post_type_archive( self::POST_TYPE_SLUG )
``` 
always returned false for me.

spacedmonkey: Screenshot 2021-09-21 at 17 13 49

includes/Story_Post_Type.php
spacedmonkey: What if the post type is trashed or draft?

spacedmonkey: [get_page_uri](https://developer.wordpress.org/reference/functions/get_page_uri/) doesn't seem to have a check for post status?

includes/Story_Post_Type.php
```diff
@@ -174,9 +181,10 @@ public function register_post_type() {
 				'rewrite'               => [
 					'slug'       => self::REWRITE_SLUG,
```
spacedmonkey, line 183:
```suggestion
					'slug'       => is_string( $this->get_has_archive() ) ? $this->get_has_archive() : self::REWRITE_SLUG,
```

I think we need to this, so the url structure of stories changed as well.

spacedmonkey:
What should happen when you go to `/web-stories`? Should it 301 to the new archive link? Should the url structure of the story change to match? `/custom-page/story-uri`

swissspidy:
> What should happen when you go to `/web-stories`? Should it 301 to the new archive link?

Redirects are out of scope for this ticket and can be looked at in a new ticket.

(The answer is probably yes.)

> Should the url structure of the story change to match? `/custom-page/story-uri`

The URL structure for single stories will not be changed and remains unaffected by this ticket.

Related: #4459

includes/Story_Post_Type.php
swissspidy: This slug is completely ignored for archives if `has_archive` is a string. See https://github.com/WordPress/wordpress-develop/blob/0f0089391b0f92d4508b8a4df567df2052161340/src/wp-includes/class-wp-post-type.php#L605

Changing the URL structure for individual stories is out of scope.

includes/Story_Post_Type.php
swissspidy: Good point. Added a check for post status now.

swissspidy:
> > What should happen when you go to `/web-stories`? Should it 301 to the new archive link?
> 
> Redirects are out of scope for this ticket and can be looked at in a new ticket.
> 
> (The answer is probably yes.)

#9100 now covers the redirect part.

PR #9060, opened 2021-09-17
Dynamic Dropdown placement when not enough space

BrookeGraham:
## Context

Dropdown can be hidden if it's located too low on the screen.  If bottom of the dropdown `popup` would be below the window, we force placement to be at the top.

## Relevant Technical Choices

Copied callback used in tooltip to dynamically set placement. Added `popupRect.bottom` to `getOffset` util func. 

## To-do
n/a

## User-facing changes

When a dropdown is close to the bottom of the screen and the list would be cutoff, we change the placement to TOP

## Testing Instructions

1.  In storybook -> Dropdown make window short than the height of where the dropdown would end. 
2. Open dropdown and it should render on top
3. Making the window taller and reopening the dropdown should place it back on the bottom if there is enough space to render the popup container. 

no

no

no

## Checklist

---

Fixes #
9049

### DIFF
--- a/packages/design-system/src/components/dropDown/index.js
+++ b/packages/design-system/src/components/dropDown/index.js
@@ -96,9 +96,8 @@ export const DropDown = forwardRef(
     const positionPlacement = useCallback(
       (popupRef) => {
         // check to see if there's an overlap with the window edge
-        const dimensions = popupRef.current?.getBoundingClientRect();
-        const bottom = dimensions?.bottom;
-        const top = dimensions?.top;
+        const { bottom, top } = popupRef.current?.getBoundingClientRect() || {};
+
         // if the popup was assigned as bottom we want to always check it
         if (
           dynamicPlacement.startsWith('bottom') &&
--- a/packages/design-system/src/components/dropDown/index.js
+++ b/packages/design-system/src/components/dropDown/index.js
@@ -22,6 +22,8 @@ import {
   useCallback,
   useMemo,
   useRef,
+  useState,
+  useDebouncedCallback,
 } from '@web-stories-wp/react';
 import PropTypes from 'prop-types';
 import { v4 as uuidv4 } from 'uuid';
@@ -85,11 +87,33 @@ export const DropDown = forwardRef(
     ref
   ) => {
     const internalRef = useRef();
+    const placementRef = useRef(placement);
+    const [dynamicPlacement, setDynamicPlacement] = useState(placement);
 
     const { activeOption, isOpen, normalizedOptions } = useDropDown({
       options,
       selectedValue,
     });
+
+    const positionPlacement = useCallback(
+      (popupDimensions) => {
+        // check to see if there's an overlap with the window edge
+        const neededVerticalSpace = popupDimensions?.offset.bottom;
+        // if the popup was assigned as bottom we want to always check it
+        if (
+          popupDimensions &&
+          dynamicPlacement.startsWith('bottom') &&
+          neededVerticalSpace >= window.innerHeight
+        ) {
+          setDynamicPlacement(PLACEMENT.TOP);
+        }
+      },
+      [dynamicPlacement]
+    );
+
+    const resetPlacement = useDebouncedCallback(() => {
+      setDynamicPlacement(placementRef.current);
+    }, 100);
 
     const handleSelectClick = useCallback(
       (event) => {
@@ -102,7 +126,8 @@ export const DropDown = forwardRef(
     const handleDismissMenu = useCallback(() => {
       isOpen.set(false);
       internalRef.current.focus();
-    }, [isOpen, internalRef]);
+      resetPlacement();
+    }, [isOpen, resetPlacement]);
 
     const handleMenuItemClick = useCallback(
       (event, menuItem) => {
@@ -168,7 +193,8 @@ export const DropDown = forwardRef(
           <Popup
             anchor={internalRef}
             isOpen={isOpen.value}
-            placement={placement}
+            placement={dynamicPlacement}
+            onPositionUpdate={positionPlacement}
             fillWidth={popupFillWidth}
             zIndex={popupZIndex}
           >
--- a/packages/design-system/src/components/dropDown/index.js
+++ b/packages/design-system/src/components/dropDown/index.js
@@ -23,7 +23,6 @@ import {
   useMemo,
   useRef,
   useState,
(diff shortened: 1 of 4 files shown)


packages/design-system/src/components/dropDown/index.js
```diff
@@ -85,12 +87,34 @@ export const DropDown = forwardRef(
...
+        if (
+          popupDimensions &&
+          dynamicPlacement.startsWith('bottom') &&
+          neededVerticalSpace >= window.innerHeight
+        ) {
+          setDynamicPlacement(PLACEMENT.TOP);
+        }
+      },
+      [dynamicPlacement]
+    );
```
BrookeGraham: This is essentially the exact same callback as used in tooltip, should I move into a shared util file?

packages/design-system/src/components/tooltip/index.js
```diff
@@ -134,7 +134,7 @@ function Tooltip({
   const positionPlacement = useCallback(
     ({ offset }) => {
       // check to see if there's an overlap with the window edge
-      const neededVerticalSpace = offset.y + offset.height;
+      const neededVerticalSpace = offset.bottom;
```
BrookeGraham, line 137: I realized while fixing the dropdown, that the way I was calculating height for tooltip was incorrect. It didn't account for the content of the tooltip being more than one line.  The `offset.height` that is returned is for the `anchorRect`

packages/design-system/src/components/popup/utils/getOffset.js
```diff
@@ -117,5 +117,7 @@ export function getOffset(placement, spacing, anchor, dock, popup) {
     y: offsetY,
     width: anchorRect.width,
     height: anchorRect.height,
+    // We want to know where the popRect container stops before scrolling begins
+    bottom: popupRect?.bottom,
```
BrookeGraham, line 121: Before (in the tooltip dynamic placement) the `offset.height` that was returned is for the `anchorRect` this meant that the offset was being calculated based off the anchor and not the popup contents.

packages/design-system/src/components/dropDown/index.js
```diff
@@ -85,12 +87,34 @@ export const DropDown = forwardRef(
...
+        // check to see if there's an overlap with the window edge
+        const neededVerticalSpace = popupDimensions?.offset.bottom;
+        // if the popup was assigned as bottom we want to always check it
+        if (
+          popupDimensions &&
+          dynamicPlacement.startsWith('bottom') &&
+          neededVerticalSpace >= window.innerHeight
+        ) {
+          setDynamicPlacement(PLACEMENT.TOP);
+        }
```
maxyinger: I think this is a pretty far out edge case, but if someone were to scroll up to cause this to happen, then scroll back down would we want this to reset back to the original position? this might fix it:
```
        if (
          popupDimensions &&
          dynamicPlacement.startsWith('bottom') &&
          neededVerticalSpace >= window.innerHeight
        ) {
          setDynamicPlacement(PLACEMENT.TOP);
        } else {
          setDynamicPlacement(null);
        }
        
        ....
        
         placement={dynamicPlacement || placement}
```

example of lil edge case:
https://user-images.githubusercontent.com/35983235/133857060-0ff1c29c-5958-4570-982f-1a5eed1e0625.mp4

packages/design-system/src/components/dropDown/index.js
```diff
@@ -102,7 +126,8 @@ export const DropDown = forwardRef(
     const handleDismissMenu = useCallback(() => {
       isOpen.set(false);
       internalRef.current.focus();
-    }, [isOpen, internalRef]);
+      resetPlacement();
+    }, [isOpen, resetPlacement]);
```
maxyinger: I don't think we'll need this `recetPlacement()` logic if we just do `placement={dynamicPlacement || placement}` and make sure to set `dynamicPlacement` to `null` when it doesn't meet the particular conditions

maxyinger (approved):
Just a small suggestion that should fix some edge cases. Other than that looks great!

spacedmonkey:
The PR description could do with a screenshot

packages/design-system/src/components/dropDown/index.js
samwhale: I think it's okay keeping it out of a util for now. If it happens again then let's do it!

samwhale (approved):
This is rad - I agree with Max on his feedback too! We shouldn't need `resetPlacement` with those updates

packages/design-system/src/components/dropDown/index.js
BrookeGraham: Ok sorry this took a minute. I couldn't pass `null` as that would lead to an infinite loop. I realized also that by grabbing `popupDimensions` from the `onPositionUpdate` callback, I was getting stale values. So instead I am passing in the `popupRef` itself in order to calculate the popup dimension here.

BrookeGraham:
https://user-images.githubusercontent.com/1820266/134071319-9b82899e-e0a9-4302-a207-0ed07387f3e3.mov

packages/design-system/src/components/dropDown/index.js
```diff
@@ -85,12 +86,34 @@ export const DropDown = forwardRef(
...
       options,
       selectedValue,
     });
 
+    const positionPlacement = useCallback(
+      (popupRef) => {
+        // check to see if there's an overlap with the window edge
+        const dimensions = popupRef.current?.getBoundingClientRect();
+        const bottom = dimensions?.bottom;
+        const top = dimensions?.top;
```
BrittanyIRL: 🤔  nit: 

```suggestion
        const { bottom, top } = popupRef.current?.getBoundingClientRect();
```

BrittanyIRL (approved):
this is awesome!

packages/design-system/src/components/dropDown/index.js
BrookeGraham: Since this is a callback, `popupRef` won't be defined on the first pass and we can't destructure `bottom` and `top` from `undefined`. I could add an if statement and destructure when it is defined. 
```
const dimensions = popupRef.current?.getBoundingClientRect();
if (dimensions) {
  const { bottom, top } = dimensions;
  ...
}
```

BrookeGraham: Or I guess I could give it a fallback 🤦🏻‍♀️

PR #9062, opened 2021-09-17
Pre-Publish Checklist: Use WASM AMP Validator

### LINKED ISSUE #8477: Pre-Publish Checklist: Use WASM AMP validator
While setting up AMP validation checks in the checklist, Pascal mentioned that now the validator is also available in Web Assembly and we could pull it in without needing to update PHP at all. 

For the time being, the PHP is working just fine. This method could be more suitable. This ticket is to explore and potentially implement that and retire the php implementation that was added in #8443. 

Original thread: https://github.com/google/web-stories-wp/pull/6185/files#r675183318

BrittanyIRL:
## Context

https://github.com/ampproject/amphtml/issues/36067#issuecomment-919532568

## Summary

The current way we're validating stories in the checklist of the editor is getting deprecated, so we need to update to use `https://cdn.ampproject.org/v0/validator_wasm.js`

## Relevant Technical Choices

Using the `_wasm` version of the validator script requires us to use `WebAssembly` to instantiate the validator and then validate the string. 

The problem I'm running into is with the script getting loaded ahead of time I'm unable to figure out how to make the validator from `window.` into an ArrayBuffer and get something readable for WebAssembly. 

The new script gives us `window.amp.validator` like the previous one, but it won't work without web assembly and there's no promise as far as I can tell to access there. the new script also gives us `window.loadValidatorWasm` which seems like what we want because `wasm` - that's a function, waiting on it will give you all the same options that the `amp.validator` does but it has a promise there with it. 
Screen Shot 2021-09-17 at 10 57 15 AM

Pretty much, It's catching on `WebAssembly` and throwing an error in the try/catch. 
I've kept console logs here to try and make debugging easier. I'm _sure_ i've been staring at this too long and that someone with more knowledge about WebAssembly and the validator can probably figure this out faster. 

I also have it as a codepen with the same console errors if you don't want to run the branch: https://codepen.io/brittanyIRL/pen/MWoQOrO?editors=1011 

_______
### Update 
I was just doing this wrong. Thought we needed to use WebAseembly to gain access to validating, but really we just needed to import and initialize. Thanks (again) @swissspidy. 

## To-do

- [x] One validator script works need to swap out what we're looking for: https://github.com/ampproject/amphtml/blob/cfb102aafc1543788b77f3ef9bd138c753e13097/validator/validator.proto#L853 

## Testing Instructions

1.

## Checklist

---

Fixes #8477


swissspidy:
@BrittanyIRL I think all you were missing is a call to `amp.validator.init()`.

With that, the validator runs just fine.

The weird thing is I'm getting the following `STYLESHEET_AND_INLINE_STYLE_TOO_LONG` error for a blank story when I should be getting `MISSING_URL` because of a missing story poster image:

```json
{
    "severity": "ERROR",
    "code": "STYLESHEET_AND_INLINE_STYLE_TOO_LONG",
    "line": 6,
    "col": 2864,
    "specUrl": "https://amp.dev/documentation/components/amp-story",
    "paramsList": [
        "poster-portrait-src",
        "amp-story"
    ],
    "params": [
        "poster-portrait-src",
        "amp-story"
    ]
}
```

No idea why that happens yet. Looking into it a bit.

swissspidy:
@googlebot I consent

BrittanyIRL:
> I think all you were missing is a call to `amp.validator.init()`.
> 
> With that, the validator runs just fine.

😭  but also 🎉  

> The weird thing is I'm getting the following `STYLESHEET_AND_INLINE_STYLE_TOO_LONG` error for a blank story when I should be getting `MISSING_URL` because of a missing story poster image

That's odd, i wonder why! i'll poke around a bit too later today now that the validator's actually working.

BrittanyIRL:
Looking at the [list of errors](https://github.com/ampproject/amphtml/blob/cfb102aafc1543788b77f3ef9bd138c753e13097/validator/validator.proto#L853) there's no longer any note about the publisher logo and the `missing_url` - but in the error returned for the inline style too long it specifies the poster portrait src which makes me think it's addressing the same thing. 

Adding in a publisher logo on a blank story resolves the inline style too long.

packages/story-editor/src/components/checklist/checks/storyAmpValidationErrors.js
```diff
@@ -35,38 +40,22 @@ export async function getStoryAmpValidationErrors({ link, status }) {
...
-        ) {
-          return false;
-        }
-        // Missing video posters
-        if ('INVALID_URL_PROTOCOL' === code && params?.[0].startsWith('src')) {
-          return false;
-        }
-
-        return true;
-      });
```
swissspidy: Why was this contextual handling of error codes removed?

This sounds like any `MISSING_URL` errors are filtered, while we only need to filter them when the error is for specific elements (like story poster or publisher logo).

But if `MISSING_URL` happens on some other element, we still want to surface that error.

So I think this needs to be reverted.

swissspidy:
> Looking at the [list of errors](https://github.com/ampproject/amphtml/blob/cfb102aafc1543788b77f3ef9bd138c753e13097/validator/validator.proto#L853) there's no longer any note about the publisher logo and the `missing_url` - but in the error returned for the inline style too long it specifies the poster portrait src which makes me think it's addressing the same thing.
> 

Yes it  does reference the poster attribute, but the error code is clearly wrong and doesn't match what you get on validator.amp.dev, which is why I wanted to look into this.

This needs further clarification before merging ⚠️ 

> Adding in a publisher logo on a blank story resolves the inline style too long.

I assume you meant a poster image?

swissspidy:
Opened https://github.com/ampproject/amphtml/issues/36112 for looking into the error code issue

packages/story-editor/src/components/checklist/checks/storyAmpValidationErrors.js
BrittanyIRL: Think it's as simple as I misunderstood. I'll pop it back in and swap this PR back to a draft pending what you find out about the wrong error code.

swissspidy:
> Opened [ampproject/amphtml#36112](https://github.com/ampproject/amphtml/issues/36112) for looking into the error code issue

Bug has been confirmed; fix is being worked on

swissspidy:
The validator error code bug has now been fixed!

swissspidy (approved):

Issue #9056, opened 2021-09-17
Fetch all hierarchical taxonomies, not just first 100


samwhale:
#9048 Adds the hierarchical input to the editor. Hierarchical taxonomies are loaded when the editor is mounted - but only the first 100 are loaded.

Gutenberg grabs all the pages, 100 items at time when there's more than 1 page. We actually have a little util for this from many moons ago:

https://github.com/google/web-stories-wp/blob/main/packages/dashboard/src/app/api/fetchAllFromPages.js

_Originally posted by @BrittanyIRL in https://github.com/google/web-stories-wp/pull/9048#discussion_r711174473_

swissspidy:
🧐 I don‘t think you need this util.

Pass `-1` as the number to fetch all pages, one by one, like Gutenberg does.

This util could probably be deleted

BrittanyIRL:
Well, today i learned! 

cool, this util isn't actually in use anywhere anymore, it was left over from something in the dashboard and kept it around incase we need it. 

let's keep this ticket to implement that (and gut the util) just to track it. dropping points though.

BrookeGraham:
I was gonna remove the util, but it is used `useCategoriesApi` and `useTagsApi`

BrittanyIRL:
> I was gonna remove the util, but it is used `useCategoriesApi` and `useTagsApi`

good note, we can clean that out separately, those are actually not in use and probably irrelevant now that these are separated so if we were to ever bring them back these would need updating anyways.

swissspidy:
useTagsApi is not used anywhere. useCategoriesApi probably neither

BrittanyIRL:
> useTagsApi is not used anywhere. useCategoriesApi probably neither

It's not. Here's the PR to clean it out: https://github.com/google/web-stories-wp/pull/9065 - I was planning on addressing it this coming week but had 30 minutes this AM so went ahead with it. The reason for the separation is that Brooke's PR is getting added to Samuel's for categories in the editor, which is already a big PR. It seemed unnecessary to add unrelated changes to that PR in addition. (Plus, Brooke's new on the project and still learning the codebase, I don't blame her for wanting to make sure it wasn't in play somewhere that's hard to find.) Anyways, that's just context. Happy to remove that now that it's not even a valuable reference material for anything else.

swissspidy:
As mentioned on the other PR, use this to quickly generate lots of test data to tinker with:

```bash
wp term generate <taxonomyslug> --count=1000
```

PR #9064, opened 2021-09-17
Enhancement/9056 fetch all taxonomies

BrookeGraham:
## Context

fetch all pages, one by one, like Gutenberg does.

## Testing Instructions

1. add over 100 categories
2. load page and look at network for more then one request for categories

no

no

no

## Checklist

---

Fixes #
9056


packages/story-editor/src/app/taxonomy/taxonomyProvider.js
```diff
@@ -191,11 +204,29 @@ function TaxonomyProvider(props) {
...
 
+  // Fetch hierarchical taxonomies on mount
+  useEffect(() => {
+    // only fetch when `shouldRefetchCategories` is true
+    if (shouldRefetchCategories && taxonomies?.length) {
+      const hierarchicalTaxonomies = taxonomies.filter(
+        (taxonomy) => taxonomy.hierarchical
+      );
+      hierarchicalTaxonomies.forEach((taxonomy) =>
+        addSearchResultsToCache(taxonomy, { perPage: -1 })
```
maxyinger, line 223: Does `per_page: -1` fetch all items? I'm not seeing this being a feature of the wp rest api in the docs:
https://developer.wordpress.org/rest-api/using-the-rest-api/pagination/

I think we may have to fetch all pages like:
```
promiseAll([
  fetchPage1(),
  fetchPage2(),
  ...etc
])
```
we should be able to get the total number of pages off of the `X-WP-TotalPages` header attribute of the response.

Bummer. it looks like the old code we had in the dashboard that did something similar got deleted in the past week or so. let me see if I can dig up an old commit where we can reference it

maxyinger (changes requested):

packages/story-editor/src/app/taxonomy/taxonomyProvider.js
maxyinger: Here's that old function for reference that fetched all pages. will have to adapt it a bit, but should be a good reference point:
https://github.com/google/web-stories-wp/pull/9065/files#r712270861

samwhale: Pascal added this to the ticket: https://github.com/google/web-stories-wp/issues/9056#issuecomment-922013624

>Pass -1 as the number to fetch all pages, one by one, like Gutenberg does.

So I guess we'd only need to make one call?

samwhale:
@googlebot I consent.

packages/story-editor/src/app/taxonomy/taxonomyProvider.js
BrittanyIRL: yup! passing `-1` should return us all of the pages - you can see this in action if you have a bunch of categories on a post, you'll see multiple calls get made to grab all the pages. I haven't tested this one yet but I'm assuming it's the same way. 

This is also why I went ahead and deleted that old util. :)

maxyinger: oh that's awesome

PR #9087, opened 2021-09-20
Pre-Publish Checklist: Update Thumbnails for PagePreview Cache

### LINKED ISSUE #8883: Checklist Thumbnails: Caching catch up
Post https://github.com/google/web-stories-wp/issues/7987 we need to catch the checklist thumbnails up to changes, they are rendering as just solid black now.

BrittanyIRL:
## Context

Last release, prometheus introduced a cool new performance win with the `PagePreview` used in the carousel not realizing that the checklist is using that same component to render pages in the checklist. The initial viewing of the checklist if there's page previews happening caches as a black rectangle and doesn't update until you collapse the checklist once. 

## Summary

Conditionally rendering the checklist's page preview thumbnails so that what they only render when needed and don't cache early before they're accurate.

## Relevant Technical Choices

Moving the `openPanel` state out of the base `checklist` and into the provider that's used to track the checklist's current status so that we know what panel is open in the checklist and can pass `isVisible` where needed from the different sets of checks.
Still overriding the `PagePreview` button to be a div so that it's not double nesting buttons

## User-facing changes

- If the thumbnail caching experiment is on and you open the checklist with one or more of the 4 checks that will render a thumbnail visible you will no longer get a black rectangle the first time you open the checklist (see videos). 

**Before (Main)**
https://user-images.githubusercontent.com/10720454/134074641-eb5393a4-20b8-4fbd-8124-03e8e8c3b1ae.mp4

**After (This branch)**
https://user-images.githubusercontent.com/10720454/134074645-35ac9650-ed49-48bb-9f41-e07ff8bb70f2.mp4

## Testing Instructions

The four checks you want to look at are: 
- Too little text
- Too much text
- Too many links 
- Bad color contrast between bg and fg 

Just make sure that the page on the thumbnail's actually visible and when you update the page the thumbnail also updates.

1.

No

No

No

## Checklist

---

Fixes #8883

### DIFF
--- a/packages/story-editor/src/components/checklist/checklistContent/designChecks.js
+++ b/packages/story-editor/src/components/checklist/checklistContent/designChecks.js
@@ -91,9 +91,9 @@ export function DesignChecks(props) {
     <ChecklistCategoryProvider category={ISSUE_TYPES.DESIGN}>
       <DesignPanel {...props}>
         <StoryPagesCount />
-        <PageTooMuchText />
+        <PageTooMuchText isVisible={isVisible} />
         <PageTooLittleText isVisible={isVisible} />
-        <PageTooManyLinks />
+        <PageTooManyLinks isVisible={isVisible} />
         {hasUploadMediaAction && <VideoElementResolution />}
         {hasUploadMediaAction && <ImageElementResolution />}
       </DesignPanel>
--- a/packages/story-editor/src/components/checklist/checklistContent/designChecks.js
+++ b/packages/story-editor/src/components/checklist/checklistContent/designChecks.js
@@ -17,6 +17,7 @@
  * External dependencies
  */
 import { __ } from '@web-stories-wp/i18n';
+import { useEffect, useState } from '@web-stories-wp/react';
 import PropTypes from 'prop-types';
 /**
  * Internal dependencies
@@ -32,6 +33,7 @@ import { ChecklistCategoryProvider } from '../countContext';
 import { PanelText, StyledTablistPanel } from '../styles';
 import { useIsChecklistMounted } from '../popupMountedContext';
 import { useConfig } from '../../../app';
+import { useChecklist } from '..';
 
 function DesignPanel({
   badgeCount = 0,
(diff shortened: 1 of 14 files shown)


packages/story-editor/src/components/checklist/checklistContent/designChecks.js
```diff
@@ -69,17 +71,29 @@ DesignPanel.propTypes = {
...
     hasUploadMediaAction: capabilities.hasUploadMediaAction,
   }));
 
+  const { openPanel } = useChecklist(({ state: { openPanel } }) => ({
+    openPanel,
+  }));
+
+  useEffect(() => {
+    if (openPanel === ISSUE_TYPES.DESIGN) {
+      setIsVisible(true);
```
BrittanyIRL: We only need to care the first time it opens so that the cache populates right. After that I didn't see a need to hide the thumbnail again and need to repaint.

maxyinger (approved):
Looks good. nice catch

BrookeGraham (approved):
💅

packages/story-editor/src/components/checklist/checklistContent/designChecks.js
BrookeGraham: I like it.

BrittanyIRL:
Was having a lot of trouble getting thumbnails to render for the currently getting edited page in the _collapsed_ checklist panel (so, if i've got design open and i'm editing something that also has a contrast issue), when I expand the accessibility section that same page is now a black rectangle (sad!). Troubleshooting this for a while, trying various solutions where you wait for the initial paint or only render when the panel's active and no dice. Did however, realize that in the carousel, the active page isn't a thumbnail cache (which is why there's that distinction in the PagePreview - I just wasn't familiar with the component and it took me some time). Anyways, I updated this PR to just have the thumbnail previews set as active to keep them in sync and current. Kept the updates for which panel's open in context because that feels useful and like it's not harming anything, also kept my update to unify the thumbnail page previews.

PR #9088, opened 2021-09-20
Taxonomies: Update shallow level key names to be camel case

### LINKED ISSUE #9057: Taxonomies: Update property names for taxonomy tags when they are fetched from the API
## Feature Description

After fetching the taxonomies we should update property names to make them easier to work with.

Currently the response looks like this:

```js
  capabilities: PropTypes.object,
  description: PropTypes.string,
  hierarchical: PropTypes.bool,
  labels: PropTypes.object,
  name: PropTypes.string.isRequired,
  rest_base: PropTypes.string.isRequired,
  show_cloud: PropTypes.bool,
  slug: PropTypes.string.isRequired,
  types: PropTypes.arrayOf(PropTypes.string),
  visibility: PropTypes.object,
  _links: PropTypes.object,
```

Let's use camel case for `rest_base` and `show_cloud`

## Additional Context

From this comment: https://github.com/google/web-stories-wp/pull/9048#discussion_r711151986

samwhale:
## Context

https://github.com/google/web-stories-wp/pull/9048#discussion_r711151986

## Summary

Taxonomies have snake case property names. After being fetched, the property names are updated to be in camel case.

This only happens at the highest level. Nested properties are still in snake case. 

## Relevant Technical Choices

We can change the nested properties too but it seemed like overkill and it wasn't specifically asked for.

## To-do

n/a

## User-facing changes

n/a

## Testing Instructions

- tests should still pass
- all taxonomy stuff should still work the same

no

no

no

## Checklist

---

Fixes #9057


BrittanyIRL (approved):
😻

packages/story-editor/src/app/taxonomy/taxonomyProvider.js
```diff
@@ -67,7 +68,19 @@ function TaxonomyProvider(props) {
...
+        const formattedTaxonomies = Object.values(result).map((taxonomy) => {
+          const entries = Object.entries(taxonomy);
+
+          const formattedEntries = entries.map((entry) => [
+            snakeCaseToCamelCase(entry[0]),
+            entry[1],
+          ]);
+
+          return Object.fromEntries(formattedEntries);
+        });
```
swissspidy: This should be moved to `packages/wp-story-editor/src/api/taxonomy.js` instead, as some sort of `transformResponse` option.

`packages/story-editor` should not have to deal with this.

packages/story-editor/src/app/taxonomy/taxonomyProvider.js
```diff
@@ -67,7 +68,19 @@ function TaxonomyProvider(props) {
...
+
+        const formattedTaxonomies = Object.values(result).map((taxonomy) => {
+          const entries = Object.entries(taxonomy);
+
+          const formattedEntries = entries.map((entry) => [
+            snakeCaseToCamelCase(entry[0]),
+            entry[1],
+          ]);
+
+          return Object.fromEntries(formattedEntries);
```
maxyinger: I have a little util in `app/taxonomy/utils` that's called `mapObjectKeys()`. should be able to just:
```
const entries = mapObjectKeys(taxonomy, (key) => snakeCaseToCamelCase(key))
```

maxyinger (approved):
lgtm

packages/story-editor/src/app/taxonomy/taxonomyProvider.js
samwhale: Chatted with Max about this - this util is in a different package, so for now we're holding off on using it. In the future if this happens more we could move these utils to a shared package like the design system.

packages/wp-story-editor/src/api/taxonomy.js
```diff
@@ -23,21 +23,46 @@ import { addQueryArgs } from '@web-stories-wp/design-system';
...
+ * @param {Object} options Options to transform the data.
+ * @param {Object} options.asCamelCase Response properties will be in camel case.
  * @return {Promise} Taxonomies promise.
  */
-export function getTaxonomies(postType, path) {
-  return apiFetch({
+export async function getTaxonomies(
+  postType,
+  path,
+  { asCamelCase = false } = {}
```
swissspidy: What's the reason for adding this param? 🤔 

I'd remove this param and just always return in camel case, really.

BrittanyIRL:
@googlebot I consent.

packages/wp-story-editor/src/api/taxonomy.js
```diff
@@ -31,13 +35,24 @@ import apiFetch from '@wordpress/api-fetch';
  * @param {string} path API path.
  * @return {Promise} Taxonomies promise.
  */
-export function getTaxonomies(postType, path) {
-  return apiFetch({
+export async function getTaxonomies(postType, path) {
+  const result = apiFetch({
```
swissspidy:
```suggestion
  const result = await apiFetch({
```

Seems more readable to have the `await` here instead of doing `await result` below

BrittanyIRL: sounds good, fixed it.

swissspidy (approved):

PR #9095, opened 2021-09-21
Add Moveable for Text Presets

### LINKED ISSUE #8359: Allow inserting text presets by drag and drop (Moveable)
## Feature Description
Currently, other elements can all be added to the page by dragging and dropping in addition to just clicking on the element.

Text presets, however, can only be added by clicking. For consistency, we should also allow dragging presets onto the page.

## Additional Context
Note that this will influence the smart text color feature since the color calculation needs to be done once dropped.

---

miina:
## Context
Allows inserting text presets by drag and drop.

## To-do

- [x] Tests

## Testing Instructions

1. Open the Text elements panel
2. Try dragging the presets (Title 1, Paragraph, etc.) onto the Page
3. Verify the texts are inserted as expected
4. Verify it's still possible to insert texts via clicking as well
5. Verify it's still possible to insert Text Presets via keyboard as well

## Checklist

---

Fixes #8359


merapi (approved):
LG, I just noticed two things:
- in the text pane the font is Arial (and when you drag) but when you drop it's Roboto (this is a bug, right?)
- should we create a follow-up ticket for smart text on DND or deal with it here?

miina:
> should we create a follow-up ticket for smart text on DND or deal with it here?

Smart text is not expected to happen on DND currently for any of the DND of texts, this might get resolved with the "magic wand or similar" feature for 1.13.

> in the text pane the font is Arial (and when you drag) but when you drop it's Roboto (this is a bug, right?)

Hm, yep, would be good to file a bug for adjusting the font in the panel.

Issue #9034, opened 2021-09-14
Taxonomy: Predictive suggestions for tags


BrittanyIRL:
## Feature Description

When typing in new tags, if there are tags available already that fit what has currently been typed so far, offer the options that fit below the input. 

Using the keyboard arrows should allow you to navigate the list of matching strings but the input should stay active so you can choose to keep typing instead. 

![Kapture 2021-09-14 at 12 32 12](https://user-images.githubusercontent.com/10720454/133322088-10aed04a-e043-4067-affb-6d566d995bbb.gif)

swissspidy:
This is super important for contributors who cannot create new tags but only assign existing ones. Without an autocomplete you have no idea if the tag assignment works or not.

BrittanyIRL:
> This is super important for contributors who cannot create new tags but only assign existing ones. Without an autocomplete you have no idea if the tag assignment works or not.

Yup, realized that after finding out that tags were only blocked from create not also updating to use existing tags with the capabilities from the story. Part of reason for pushing back the release.

PR #9047, opened 2021-09-16
Taxonomy: Register new taxonomies

spacedmonkey:
## Summary

Register two new taxonomies for stories. These taxonomies are public for now, but can be updated via the REST API. The REST API has a custom controller, in the web stories namespace. This means, a number of existing endpoints needed to updated to ensure that they all point to the rest apis in the correct namespace. 

## Relevant Technical Choices

- For now, capabilities are mapped to existing web stories capabilities. 
- Taxonomy base, is designed to make taxonomy classes super generic, with defining properties and the base class handles all the register etc. 

## Testing Instructions

- [ ] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1.

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [ ] I have tested this code to the best of my abilities
- [ ] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [ ] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [ ] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [ ] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #9038

PR #1468, opened 2020-05-04
Move capabilities out of history

spacedmonkey:
## Summary

Fixes #630

## To-do

---

Fixes #

Issue #630, opened 2020-03-18
Move capabilities out of history


miina:
## Feature Description

Follow-up discovery issue for potentially reworking how capabilities are stored.

See discussion thread starting from [here](https://github.com/ampproject/amp-wp/pull/4122#discussion_r367949420) with the conclusion to follow up in a separate issue.

Edit after testing:
Capabilities seem to create extra history entries which can "mess up" the user's history, especially if the offset of the current history state is something else than 0.

To replicate:
- Open a new story without history
- Go to Document tab
- See new history entry (Undo becoming available).

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

spacedmonkey:
@miina This is a duplicate #196

miina:
@spacedmonkey Thanks, saw that issue but it wasn't clear from the description it was a duplicate. 
Will close this.

miina:
Reopening as #196 seems to cover the capabilities wider, including leveraging more capabilities in the editor, whereas this issue is specifically about the architecture.

miina:
Although it's an old issue and not very visibly user-facing, it's still odd that capabilities are tracked in history, it would be good to rework how capabilities are stored and get this fixed.

Might not be a priority for beta but tagging you here just in case if you think otherwise:
cc @dvoytenko @pbakaus 

Note after testing: it looks like capabilities are actually causing extra history entries so this issue is really a bug issue as well.

swissspidy:
Ha, I was just looking for this! Definitely needed, especially since we want to decouple the editor from WP.

dvoytenko:
IMHO, we need to prioritize it relatively high.

spacedmonkey:
I think that #196 will fix this issue. How we deal with capabilities in general is very messy, I still think this needs fixing before we go wider with this plugin.

barklund:
We could move it to the config provider. I guess we can even load it in statically on initial page load as a parameter for the editor. Or load it initially - but still store it in the config provider. It's not clear-cut what config it provides and what it doesn't, but this could fit the bill.

jauyong:
Closing in favour of #196

PR #9099, opened 2021-09-21
Taxonomy: Check capabilities for managing and assigning terms

### LINKED ISSUE #9011: Taxonomy: Respect User Role Capabilities
Users have permissions on whether they can create or assign a taxonomy depending on their user role.

This ticket is to read these permissions dynamically from the REST API and alter user taxonomy interaction abilities accordingly.

see:
https://github.com/google/web-stories-wp/pull/9004/files#r707369583
https://github.com/google/web-stories-wp/pull/8591

I think the separated out create vs assign permissions are pretty intuitive UX wise for the hierarchical input, but not sure how this would manifest in the freeform input where the create and assign actions are pretty intertwined in the UI.

@spacedmonkey @swissspidy Could we possibly get a couple steps on how we'd see the freeform input on the Gutenberg editor with limited action permissions? that or just a screen capture for UX reference?

BrittanyIRL:
## Context

Not all users have the capability to add new categories in taxonomies. This limits that. 

## Summary

(Revised after Pascal's [comment](https://github.com/google/web-stories-wp/pull/9099#issuecomment-924398219)/[contribution](https://github.com/google/web-stories-wp/pull/9099/commits/974ac11a7b76c34c66d4c4cc6a9e8241d49029d2))

Taxonomy capabilities for managing and assigning terms are listed in the story response under `capabilities`, they are given based on the taxonomy `slug` and prefixed by either `create-` or `assign-` (respectively). 
We can use these values to decide what all to show in the panel. 

## Relevant Technical Choices

- Did a nested ternary 🥴 for the rendering of the `add category` toggle. It _feels_ easy to read still and seemed like the most straight forward option. Memoizing it would've taken in a lot of props, could make its own component, it just doesn't feel helpful - mostly just passing a buncha props and abstracting something to make it harder to parse through - plus we've got other PRs open or issues getting worked on that would conflict with that update.
- Opted for an e2e test to check on contributor vs admin capabilities because adjusting the story response in the fixture seems like over kill given we're not doing that for any other karma tests, feels like the wrong time to start. 
- Cleaned out the `sort` for ordering the UI, it doesn't seem necessary. 
- Filtering out taxonomies that users don't have the ability to see (`visibility.show_ui`) and can't assign to, if that filtering leaves the user with 0 taxonomies to see then the panel doesn't render. 
- Added some ids to buttons on hierarchical so that e2e selections are easier
- Added axe test to karma 

## To-do

- [x] - Testing this, seems like my default admin user doesn't have permissions for the other taxonomy options for colors, verticals, etc so those aren't rendering anymore. Wondering if some slugs are off or the wrong casing is occurring somewhere? Had a kinda weird rebase. Want to confirm this isn't a regression on my part. 
- [x] - Follow up with #9034 to let users that can't create terms but can still manage add tags that exist. 
- [ ] - Sadly, the e2e tests have been a time suck so I isolated them in their own ticket: #9149 to unblock this.

## User-facing changes

(Blocked by a feature flag, assuming that flag is on if you're logged in as an admin or an editor you can add new categories, if you're logged in as an author or a contributor you won't see that button anymore)

| Admin, Editor | Author, Contributor |
| --- | --- |
|Screen Shot 2021-09-21 at 1 51 52 PM|  Screen Shot 2021-09-21 at 1 52 15 PM
|Screen Shot 2021-09-21 at 1 51 45 PM |  Screen Shot 2021-09-21 at 1 52 10 PM |

## Testing Instructions

- Enable Taxonomies experiment
- Log into the editor as different roles (our local environments come set up for `admin`, `editor`, `author`, `contributor`, `subscriber` (passwords all default to `password`) and go to taxonomies. Admin and editor roles should be able to add in new categories while authors and contributors should not see that option at all.

1.

## Checklist

---

Fixes #9011


maxyinger (dismissed):
Looks awesome!

swissspidy:
Heya 👋 

I'm afraid `manage_categories` won't help much here.

Apologies for not communicating this better, but we've just added the `web_story_category` and the `web_story_tag` taxonomy directly to this plugin the other day (see #9047). (Note the underscores instead of the dashes)

-> That means you no longer need the custom example plugin I've built the other day for these to work.

As part of this change, we've added custom capabilities for to these taxonomies, similar to `manage_categories`.

For reference, they're called:

* `manage_terms_web-stories`
* `edit_terms_web-stories`
* `delete_terms_web-stories`
* `assign_terms_web-stories`

**However**, we should not simply add those to the `ConfigProvider`, however. We rather want to get rid of this old way of passing capabilities.

Instead, we should get the user's capabilities for each taxonomy from the REST API response. Specifically, the `links` section will contain `wp:action-...` links if the user has the necessary permission. (needs `?context=edit`)

Here's a screenshot:

Screenshot 2021-09-21 at 23 23 25

Here's where we currently get the capabilities for the story being edited and add them to the state:

https://github.com/google/web-stories-wp/blob/8c758a6740fd335e98cc7ae6be3c17d5afce46e8/packages/wp-story-editor/src/api/story.js#L41-L50

We now need to simply extend this for all the taxonomies.

`wp:action-create-web_story_category` means the user is allowed to create new categories, delete them, etc.

`wp:action-assign-web_story_category` means the user is allowed to assign an existing category to a story.

Ditto for tags.

Happy to provide some POC or chat about it on Slack.

Sorry again for the troubles.

BrittanyIRL:
> We now need to simply extend this for all the taxonomies.
> 
> `wp:action-create-web_story_category` means the user is allowed to create new categories, delete them, etc.
> 
> `wp:action-assign-web_story_category` means the user is allowed to assign an existing category to a story.
> 
> Ditto for tags.
> 
> Happy to provide some POC or chat about it on Slack.
> 
> Sorry again for the troubles.

It's ok :) thank you for catching it right after I posted and for the helpful response of the updates. I'll start in on this today and see if I can post anything before heading out for the day.

Are the roles to capabilities for these new custom actions documented anywhere for future reference? Looking at Gutenberg, the only thing that's limited are the ability to create new hierarchical terms - I'll make sure to check for them all as you've outlined, but for the sake of testing, it would be helpful to know if we're expecting anything different outta the gate.

swissspidy:
> Are the roles to capabilities for these new custom actions documented anywhere for future reference? Looking at Gutenberg, the only thing that's limited are the ability to create new hierarchical terms - I'll make sure to check for them all as you've outlined, but for the sake of testing, it would be helpful to know if we're expecting anything different outta the gate.

Role-wise the setup should be the same as for the built-in taxonomies in WordPress / Gutenberg.

Editors & admins can create & assign tags + categories
Contributors & authors can only assign existing tags + categories, but not create new ones

BrittanyIRL:
OK @swissspidy, I'm either doing something wrong or something is afoot - or both, who knows 😝  

I've got a role of admin and full permissions, I can go to various tag and category pages within the editor (nice work btw) and have full CRUD. I look at the `response` in `wp-story-editor/src/api/taxonomy`, where `?context=edit` is part of the path and I only get back `_links.wp:items` (used to create). 
Screen Shot 2021-09-21 at 4 48 56 PM

I feel like it's at this point that `_links` should be populating like so. I can see all the `_links` in `story.js` as `links` but they don't seem to be making it into post.taxonomies.
```
wp:action-create-story-categories: [{…}]
wp:action-create-story-colors: [{…}]
wp:action-create-story-tags: [{…}]
wp:action-create-story-verticals: [{…}]
wp:action-create-web_story_category: [{…}]
```

Screen Shot 2021-09-21 at 4 31 14 PM

Do you mean that to get capabilities for each given action there's an additional call to be made to get `_links` back? They feel like they should be with `wp:items` in `_links`. 

I'm playing catch up on the work ya'll have been doing in taxonomies, so definitely appreciate the help here. I'm heading out for the night but will check back in tomorrow am (my am!) and we can chat at some point.

swissspidy:
The action links are in the `web-stories/v1/web-story/<story_id>` response, not the one for taxonomies.

So this really just means extending this list I've mentioned:

https://github.com/google/web-stories-wp/blob/8c758a6740fd335e98cc7ae6be3c17d5afce46e8/packages/wp-story-editor/src/api/story.js#L41-L50

swissspidy:
@googlebot I consent

packages/story-editor/src/components/form/stories/tags.js
```diff
@@ -48,7 +48,7 @@ export const _default = () => {
         <Tags.Input
           id="tags-input"
           aria-describedby="tags-description"
-          name="story-tags"
+          name="web_story_tag"
         />
         <Tags.Description id="tags-description">
           {'Separate with commas or the Enter key.'}
```
spacedmonkey, line 54: Should this line be translated?

swissspidy: I was like "yes absolutely 💯 " but then realized that this is just a storybook story component. So no need.

BrittanyIRL:
> The action links are in the `web-stories/v1/web-story/<story_id>` response, not the one for taxonomies.

🤦‍♀️  ugh, thank you

packages/story-editor/src/components/panels/document/taxonomies/test/taxonomies.js
```diff
@@ -25,9 +25,20 @@ import { screen } from '@testing-library/react';
...
+import { StoryContext } from '../../../../../app/story';
 
 function arrange({ taxonomies }) {
   const storyContextValue = {
+    state: {
+      capabilities: taxonomies.reduce((acc, curr) => {
+        acc[`assign-${curr.slug}`] = true;
+        acc[`create-${curr.slug}`] = true;
+        return acc;
+      }, {}),
```
BrittanyIRL: @swissspidy  thank you for all of this, i see where I veered off-course.

swissspidy: happy to help!

packages/e2e-tests/src/specs/editor/taxonomy.js
```diff
@@ -0,0 +1,51 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * External dependencies
+ */
+import { createNewStory, withUser } from '@web-stories-wp/e2e-test-utils';
+
+describe('Taxonomy', () => {
```
swissspidy, line 25: 😻

packages/story-editor/src/components/header/buttons/publish.js
```diff
@@ -125,10 +125,7 @@ function PublishButton({ forceIsSaving }) {
       <ButtonWithChecklistWarning
         onClick={handlePublish}
         disabled={
-          !capabilities?.hasPublishAction ||
-          isSaving ||
-          forceIsSaving ||
-          isUploading
+          !capabilities?.publish || isSaving || forceIsSaving || isUploading
```
BrittanyIRL, line 128: somewhere here the `forceIsSaving` value got lost in a different PR? Since its propType and prop still existed I am restoring it to disabling the publish button when true.

packages/story-editor/src/components/panels/document/taxonomies/HierarchicalTermSelector.js
```diff
@@ -124,7 +128,7 @@ function HierarchicalTermSelector({ noParentId = NO_PARENT_VALUE, taxonomy }) {
     );
 
   const categories = useMemo(() => {
-    if (termCache[taxonomy.restBase]) {
+    if (termCache?.[taxonomy.restBase]) {
```
BrittanyIRL, line 127: using optional chaining here so that unit tests don't need all of these props to run

packages/story-editor/src/karma/fixture/db/getTaxonomiesResponse.js
```diff
@@ -15,9 +15,79 @@
  */
 
 export default [
+  {
```
BrittanyIRL, line 18: move categories to top of array because this is the only reason that that sort of the taxonomies array was needed

packages/story-editor/src/karma/fixture/db/getTaxonomiesResponse.js
```diff
@@ -285,7 +285,7 @@ export default [
       ],
       'wp:items': [
         {
-          href: 'http://localhost:8899/wp-json/wp/v2/story-verticals',
+          href: 'http://localhost:8899/wp-json/web-stories/v1/web_story_category',
```
BrittanyIRL, line 288: there were a lot of rebases here, @littlemilkstudio is this the right href here?

swissspidy: `web-stories/v1/web_story_category` is correct

includes/Experiments.php
```diff
@@ -363,6 +363,7 @@ public function get_experiments(): array {
 				'label'       => __( 'Taxonomies', 'web-stories' ),
 				'description' => __( 'Enable support of tags and categories for stories', 'web-stories' ),
 				'group'       => 'editor',
+				'default'     => true,
```
BrittanyIRL: temp! will remove this. just trying to get to the bottom of why this new e2e test if failing remotely

packages/e2e-tests/src/specs/editor/taxonomy.js
```diff
@@ -0,0 +1,82 @@
...
+    await expect(page).toClick('li[role="tab"]', { text: 'Document' });
+    await expect(page).toMatch('Categories and Tags');
+
+    // is panel already open?
+    await expect(page).toMatch('Add New Category');
+    // Toggle the panel which is collapsed by default.
+    // await expect(page).toClick('[aria-label="Categories and Tags"]');
+    // await page.waitForSelector('button#expand_add_new_hierarchical_term');
+    await expect(page).toClick('button#expand_add_new_hierarchical_term');
+    await page.waitForSelector('input[name="New Category Name"]');
```
BrittanyIRL: There's wip in here - this test was completely fine running it locally but remote it's failing - just trying to isolate and figure out what's up.

packages/e2e-tests/src/specs/editor/inspector.js
```diff
@@ -24,7 +24,7 @@ describe('Inspector', () => {
       await createNewStory();
 
       await expect(page).toClick('li[role="tab"]', { text: 'Document' });
-      await await expect(page).toMatchElement('[aria-label="Publisher Logo"]');
```
maxyinger, line 27: Ah I think this test was failing on mine too. Thanks for looking into it!

packages/e2e-tests/src/specs/editor/taxonomy.js
maxyinger: 🔥

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
```diff
@@ -53,13 +54,18 @@ function FlatTermSelector({ taxonomy }) {
 
   const handleFreeformTermsChange = useCallback(
     (termNames) => {
+      // TODO(#9034): Allow assigning existing terms from autocomplete, but don't create terms if missing capability.
+      if (!canCreateTerms) {
+        return;
+      }
+
```
maxyinger: I think we might get some conflicts on here with the update to make this work with history. should be able to re-add this in tho.

Just for my UX understanding here, the idea is that a user that can't create terms will only choose them through the autocomplete in the future?

packages/story-editor/src/components/panels/document/taxonomies/taxonomies.js
```diff
@@ -35,34 +40,37 @@ function TaxonomiesPanel(props) {
...
-        return -1;
-      }
-      return 0;
-    }
-  );
+  const availableTaxonomies = taxonomies.filter((taxonomy) => {
+    const isVisible = taxonomy?.visibility?.show_ui;
+    const canAssignTerms = Boolean(capabilities[`assign-${taxonomy?.slug}`]);
+
+    return isVisible && canAssignTerms && taxonomy;
```
maxyinger: what's the need for the `&& taxonomy` here? isn't that always gonna be an object?

packages/story-editor/src/karma/fixture/containers/documentPanel/categoriesAndTags.js
```diff
@@ -35,7 +35,7 @@ export class CategoriesAndTags extends Container {
   }
 
   get addNewCategoryButton() {
-    return this.getByRole('button', { name: /Add New Category/ });
+    return this.queryByRole('button', { name: /Add New Category/ });
```
maxyinger: what's the difference between these two?

maxyinger (approved):
looks great. easy to read and reason about. nice work!

packages/e2e-tests/src/specs/editor/inspector.js
BrittanyIRL: not sure it's fixed but now it doesn't await twice 😂

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
BrittanyIRL: yup, that's exactly it. just can't add anything new - but if you can _manage_ terms then you can add any existing terms. 

I suspect some conflicts, totally fine :)

packages/story-editor/src/components/panels/document/taxonomies/taxonomies.js
BrittanyIRL: ah, yup!

packages/story-editor/src/karma/fixture/containers/documentPanel/categoriesAndTags.js
BrittanyIRL: queryBy let's the button not be found and doesn't error. I had added this in for when I thought that capabilities were a check coming from `config` rather than the stories response. i don't think it's necessary now.

packages/story-editor/src/components/panels/document/taxonomies/taxonomies.js
```diff
@@ -35,34 +40,37 @@ function TaxonomiesPanel(props) {
...
-        return -1;
-      }
-      return 0;
-    }
-  );
+  const availableTaxonomies = taxonomies.filter((taxonomy) => {
+    const isVisible = taxonomy?.visibility?.show_ui;
+    const canAssignTerms = Boolean(capabilities[`assign-${taxonomy?.slug}`]);
+
+    return isVisible && canAssignTerms;
```
BrookeGraham, line 47: 🙌🏽

BrookeGraham (approved):

packages/story-editor/src/components/panels/document/taxonomies/HierarchicalTermSelector.js
```diff
@@ -254,49 +258,55 @@ function HierarchicalTermSelector({ noParentId = NO_PARENT_VALUE, taxonomy }) {
...
-          </ButtonContainer>
-        </AddNewCategoryForm>
+      {canCreateTerms ? (
+        <>
+          <LinkButton
+            ref={toggleRef}
+            aria-expanded={false}
+            onClick={handleToggleNewCategory}
+            $isVisible={showAddNewCategory}
+            id="expand_add_new_hierarchical_term"
```
swissspidy: Is there a way to target this element in the e2e tests without a specific ID?

If not, why not use `data-testid`?

BrittanyIRL: good call, data ids will work just fine. Targeting without the id has been a little iffy, especially when trying to check it's not present for the contributor.

swissspidy:
When I assign a category and hit the Undo button, the whole Taxonomies panel disappears. Is that a known bug?

https://user-images.githubusercontent.com/841956/134770119-17068089-5178-460b-b874-99861af9e6dc.mov

BrittanyIRL:
> When I assign a category and hit the Undo button, the whole Taxonomies panel disappears. Is that a known bug?

Ah, looks like some wires got crossed between the history PR (where this wasn't the case) and this one with capabilities. The capabilities getting looked for here: `Boolean(capabilities[`assign-${taxonomy?.slug}`]);` are getting turned into `assign_`. I'll fix this later today - I'm sure it's just something small to find.

packages/story-editor/src/app/story/effects/useHistoryEntry.js
```diff
@@ -25,7 +25,7 @@ import { useEffect, useRef } from '@web-stories-wp/react';
 import { useHistory } from '../../history';
 
 // Record any change to core variables in history (history will know if it's a replay)
-function useHistoryEntry({ story, current, pages, selection }) {
+function useHistoryEntry({ story, current, pages, selection, capabilities }) {
```
BrittanyIRL, line 28: @swissspidy  This was why the taxonomy panel was disappearing after undoing something. 

So in the the storyProvider where `useHistoryEntry` effect is used, `capabilities` are being passed in but they weren't being accessed here so they'd just get lost to the story on undo. I suppose until now this hasn't been an issue. Should be all fixed though.

BrittanyIRL:
@swissspidy - do you want to hold this until the e2e tests are passing on #9150 ?

swissspidy:
> @swissspidy - do you want to hold this until the e2e tests are passing on #9150 ?

no should be fine

packages/story-editor/src/app/story/effects/useHistoryEntry.js
spacedmonkey: Context https://github.com/google/web-stories-wp/issues/630 https://github.com/google/web-stories-wp/pull/1468

spacedmonkey: CC @miina for extra context on this matter.

packages/wp-story-editor/src/api/story.js
```diff
@@ -38,16 +38,17 @@ const transformGetStoryResponse = (post) => {
...
-  };
+  post.capabilities = {};
+
+  for (const link of Object.keys(links)) {
+    if (!link.startsWith('wp:action-')) {
+      continue;
+    }
+
+    // Turn 'wp:action-assign-author' into 'assign-author'
+    const capability = link.replace('wp:action-', '');
```
spacedmonkey, line 49: Could we use [snakeCaseToCamelCase](https://github.com/google/web-stories-wp/blob/8d72b8d75d431fbd4635bc6cfc59e6d7dda69922/packages/wp-story-editor/src/api/utils/snakeCaseToCamelCase.js#L23). 

It would be cleaner to have an object like this 

```
{
     assignAuthor: true, 
     assignCategory: true
}
```

packages/story-editor/src/app/story/effects/useHistoryEntry.js
miina: @swissspidy @BrittanyIRL 

We had quite a long discussion on this about some point, resulting in a conclusion that capabilities changing is not something that should be recorded as a history entry.

When the capabilities were included in the history, switching to the Document panel caused a new history entry (since capabilities were updated). This is actually not a real change in the user editing history though.

Removal issue: #630  

I haven't checked the code in the PR but just wanted to confirm there's no regression for capabilities creating a new history entry.

Thoughts?

packages/wp-story-editor/src/api/story.js
swissspidy: This would break capability names for taxonomies containing hyphens.

So there's `wp:action-assign-wp_story_tag` but there could also be `wp:action-assign-super-awesome-tax`

Using `snakeCaseToCamelCase`  would turn this into something like `assignSuperAwesomeTax` which seems very unexpected and fragile. Because it means you need to use `snakeCaseToCamelCase` everywhere you'd want to check for that cap.

packages/story-editor/src/app/story/effects/useHistoryEntry.js
BrittanyIRL: That's good to know. We're having to handle something similar when updating taxonomy to avoid redundant history updates. That is in PR #9147  ([here](https://github.com/google/web-stories-wp/pull/9147/files#diff-b0d9645393a6258bd68def9cad9560e65e01a48f7631bc87c6a1cb75ce344872R116-R123) and [here](https://github.com/google/web-stories-wp/pull/9147/files#diff-c4aa265ebf112ac8d5f1b42b1af4004298fd441f520c662e4176d32ce0b6c925R30-R36))


It's critical to taxonomies working that capabilities stay put in the stories object. It feels weird that they'd be part of the story object and then after any undo disappear.

Looking at https://github.com/google/web-stories-wp/issues/630 and how to replicate the issue it's still working as expected here while keeping capabilities synced with the story object. 

https://user-images.githubusercontent.com/10720454/134946802-f607a085-b827-4901-bc9e-c6abd9c14a80.mp4

BrittanyIRL:
> > @swissspidy - do you want to hold this until the e2e tests are passing on #9150 ?
> 
> no should be fine

cool, thanks. i want to do other changes more isolated since this has PR has grown.

packages/story-editor/src/app/story/effects/useHistoryEntry.js
BrittanyIRL: @miina let me know if there's any other worries here. I'll wait to merge it until I hear from you.

miina: @BrittanyIRL Thanks for the answer. If the history is working as expected then there are no worries :)

packages/e2e-tests/src/specs/editor/taxonomy.js
```diff
@@ -0,0 +1,102 @@
...
+  });
+  it.todo('should be able to add new tags and existing tags');
+
+  describe('Contributor User', () => {
+    withUser('contributor', 'password');
+    // right now enabling the taxonomy experiment is needed to make this test at all worthwhile
+    // TODO https://github.com/google/web-stories-wp/issues/8833
+    // eslint-disable-next-line jest/no-disabled-tests
+    it.skip('should be able to manage categories but not add new ones', async () => {
+      await goToAndExpandTaxonomyPanel();
```
samwhale, line 89: I think you should be able to enable this experiment using this util: https://github.com/google/web-stories-wp/blob/main/packages/e2e-test-utils/src/experimentalFeatures.js#L29

But we can always do this later in the issue that you created :)

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -68,7 +68,7 @@ function Input({
     tagBuffer: null,
     offset: 0,
   });
-  const [isInputFocused, setIsInputFocued] = useState(false);
+  const [isInputFocused, setIsInputFocused] = useState(false);
```
samwhale, line 71: nice

packages/story-editor/src/components/header/buttons/publish.js
samwhale: `forceIsSaving` is still there right? I don't think it was removed, it was just on a different line 🤔

packages/story-editor/src/components/panels/document/taxonomies/HierarchicalTermSelector.js
```diff
@@ -254,49 +258,55 @@ function HierarchicalTermSelector({ noParentId = NO_PARENT_VALUE, taxonomy }) {
...
-            </AddNewCategoryButton>
-            <AddNewCategoryButton
-              aria-expanded
-              onClick={handleToggleNewCategory}
-            >
-              {__('Cancel', 'web-stories')}
-            </AddNewCategoryButton>
-          </ButtonContainer>
-        </AddNewCategoryForm>
+      {canCreateTerms ? (
```
samwhale, line 263: 👍

packages/story-editor/src/components/panels/document/taxonomies/karma/taxonomies.karma.js
```diff
@@ -59,6 +59,12 @@ describe('Categories & Tags Panel', () => {
     return terms;
   }
 
+  it('should have no aXe accessibility violations', async () => {
+    await openCategoriesAndTagsPanel();
+    const { categoriesAndTags } = fixture.editor.inspector.documentPanel;
+    await expectAsync(categoriesAndTags.node).toHaveNoViolations();
+  });
```
samwhale, line 66: 👍

packages/story-editor/src/components/panels/document/taxonomies/karma/taxonomies.karma.js
```diff
@@ -103,12 +109,9 @@ describe('Categories & Tags Panel', () => {
 
         // open the new category section
         await fixture.events.click(categoriesAndTags.addNewCategoryButton);
-        // Input should be focused
-        expect(document.activeElement).toBe(
-          categoriesAndTags.newCategoryNameInput
-        );
```
samwhale, line 109: should this not be focused anymore? I think this should still work the same 🤔

samwhale: Oh I see - this is cursor interactions so it doesn't really matter 👍

packages/story-editor/src/components/panels/document/taxonomies/karma/taxonomies.karma.js
```diff
@@ -557,16 +539,14 @@ describe('Categories & Tags Panel', () => {
...
       await waitFor(async () => {
         currentStoryTerms = await getStoryTerms();
-        expect(currentStoryTerms['story-tags'].length).toEqual(
+        expect(currentStoryTerms['web_story_tag'].length).toEqual(
           initialTagsLength - 1
         );
       });
-
-      // see that thee correct token was removed
+      // see that the correct token was removed
```
samwhale, line 549: 👍

packages/story-editor/src/karma/fixture/db/getTaxonomiesResponse.js
samwhale: 💭 I initially added the sorting because the api response while running it locally had categories underneath tags. If that's changed then we don't need this

samwhale (approved):
Nice this works great. Left some small things but good work with this

samwhale:
Should users without 'add' permissions also not be allowed to create `tags`? or is this just for hierarchical change

packages/e2e-tests/src/specs/editor/taxonomy.js
swissspidy: Handled in #9150

packages/story-editor/src/karma/fixture/db/getTaxonomiesResponse.js
swissspidy: We should not rely on the amount or order of taxonomies in the API response, nor should we care.

There might be the taxonomies `categories, tags`, or there might be the taxonomies `colors, tags, categories, foo, bar`.

The taxonomy panel needs to be built in an agnostic way and not artificially sort or hide taxonomies or anything.

swissspidy:
> Should users without 'add' permissions also not be allowed to create `tags`? or is this just for hierarchical change 

They should be able to assign existing tags. There's a separate 'assign' permission for that.

To make that useful, however, requires the autocomplete/suggestions feature which is in a separate ticket.

packages/story-editor/src/components/header/buttons/publish.js
BrittanyIRL: This PR's been up for a while, it was missing at one point but now it is back, you're right. I'll update the order so that this diff goes away.

BrittanyIRL: oh i get why this is confusing. So `forceIsSaving` _was_ removed here, adding it back in to avoid a regression - but someone else did the same so that's moot now. the capabilities did get updated here however so it's looking for `publish` not `hasPublishAction`

Issue #9096, opened 2021-09-21
Dashboard: Remove incorrect warning dialog when removing publisher logos


swissspidy:
## Feature Description

When removing a publisher logo from the Settings screen in the dashboard, you are presented with the following warning:

![Dialog saying "The logo will be removed from any stories that currently use it as their publisher logo."](https://user-images.githubusercontent.com/841956/134181265-7ef3f3f5-f888-41ee-a68a-7add07c81b91.png)

This dialog is totally incorrect though.

Removing a publisher logo only removes it from the list of publisher logos, but does not remove the actual attachment in the media library. So it does not change any existing stories at all.

For this reason, I suggest completely removing this dialog.

BrittanyIRL:
Has it always been wrong 🥴 ? I feel like when this first started it was an accurate warning. Regardless of my memory, yeah let's remove it.

swissspidy:
I'm not 100% sure how things were originally, but I think the underlying attachment has never been deleted when removing the publisher logo.

And when creating a story we currently store the publisher logo (attachment) ID, so as long as that one is not deleted by the user in the media library, there is no issue.

The publisher logo section is basically just a selection of attachments, removing an item from the selection has no impact on existing stories or new ones, other than the selection being different.

BrittanyIRL:
@BrookeGraham if you have time today or tomorrow after [9101](https://github.com/google/web-stories-wp/pull/9101) is merged can you knock this out?

dreamofabear:
IMO we should keep the dialog and update the copy to inform the user that stories using deleted logos *won't* have those logos removed. I can see some users expecting "deletion propagation" to be the default behavior.

Aside: I'm curious if there's a use case to changing publisher logos across multiple stories. Probably niche but we can keep an eye out.

BrittanyIRL:
> IMO we should keep the dialog and update the copy to inform the user that stories using deleted logos won't have those logos removed.

this feels like something that could just as easily get added to the copy so that users don't have to acknowledge that it won't remove the logo from stories every time they perform the action. 

![Screen Shot 2021-09-24 at 12 11 39 PM](https://user-images.githubusercontent.com/10720454/134728074-2c6656bf-94e1-49f7-bebc-588495662923.png)

csossi:
Verified in QA

Note: the dialog is no longer appearing, but the new content mentioned above ("Removing a publisher logo..." isn't there. Since it's note in QA instructions for 9145, assuming it's not in scope)

[...]

PR #9101, opened 2021-09-21
Publisher Logos: Add and leverage custom REST endpoint

### LINKED ISSUE #9007: Publisher Logos: Handling for non-admins
## Feature Description

Splitting this out from #8989.

**Context:**

1. Not all users can upload media / new publisher logos
2. The list of existing publisher logos can be retrieved from the settings REST API endpoint
3. Not all users have the capability to retrieve settings

**What we need:**

1. When uploading a new publisher logo, it needs to be added to the list of existing publisher logos
1. Even non-admins should be able to choose from the list of existing publisher logos

**Proposal to address this:**

_Amending the list of existing publisher logos:_

Whenever `web_stories_publisher_logo` post meta is updated for a story, add the logo ID to the list of existing publisher logos if not already in there.

_Retrieving the list of existing publisher logos:_

Add a new `web-stories/v1/publisher-logos` REST API endpoint to `GET` the list of existing publisher logos (from the `web_stories_publisher_logos` option).

* Everyone with `edit_web-story` capability should have access
* Only `GET` requests needed for now, don't need anything else

Proposed REST API response format:

```json
[
    {
        "id": 123,
        "_links": {
            "wp:attachment": [
                {
                    "embeddable": true,
                    "href": "https://web-stories.local/wp-json/wp/v2/media/123"
                }
            ]
        }
    },
    {
        "id": 456,
        "_links": {
            "wp:attachment": [
                {
                    "embeddable": true,
                    "href": "https://web-stories.local/wp-json/wp/v2/media/456"
                }
            ]
        }
    }
]
```

Then the list of all publisher logos could be retrieved via `GET web-stories/v1/publisher-logos?_embed=wp:attachment`

**Non-Goals**

* We don't need to update the Dashboard to use this new REST API endpoint for now

swissspidy:
## Context

We want to:

* Enable all users who can create stories to retrieve all publisher logos in the editor.
  This is not possible with the `web-stories/v1/settings` endpoint, which only works for admins
* Allow adding new publisher logos to the list in the editor, regardless of user role.

This PR unblocks further work on #8989, as originally outlined in #9007.

## Summary

This PR creates a new `/web-stories/v1/publisher-logos` REST endpoint for managing publisher logos.

A publisher logo has the following fields:

* `id`
* `url`
* `title` (for alt text)
* `active` (whether it's the current default publisher logo)

The following methods are possible:

* `GET /web-stories/v1/publisher-logos` - retrieve all publisher logo. available for everyone who can create stories.
* `POST /web-stories/v1/publisher-logos` - takes an attachment ID to add a new publisher logo to the list. available for everyone who can create stories.
* `DELETE /web-stories/v1/publisher-logos/<id>` - remove a publisher logo from the list. available for everyone who can create stories.
* `POST /web-stories/v1/publisher-logos/<id>` - allows setting an existing logo as the new default. available for admins only.

This PR updates the dashboard to leverage this new dedicated endpoint as some sort of proof-of-concept, while the main goal is actually unblocking #8989.

Some benefits:

* We can potentially open up publisher logo management in the dashboard to non-admins in the future
* Avoids multiple HTTP requests (one to retrieve list of publisher logo IDs, one to retrieve the actual attachments to get the URLs)
* Makes implementation more platform-agnostic, maintainable, future-proof
* Adopting this endpoint in the dashboard validates that it works as expected and meets requirements

## Relevant Technical Choices

* Custom REST API endpoint to have better control
* Moves some of the client-side logic to the server to ensure the list of publisher logos is always valid
* Adds a new `usePublisherLogosApi` hook with a custom reducer to the dashboard to facilitate working with publisher logos.

## To-do

N/A

## User-facing changes

N/A

## Testing Instructions

1. As an _admin_, verify that managing publisher logos on the dashboard still works as expected (adding, removing, setting as default)

No

No

No

## Checklist

---

Fixes #9007


packages/dashboard/src/karma/fixture.js
```diff
@@ -81,12 +81,14 @@ const defaultConfig = {
   wpListURL: 'http://localhost:8899/wp-admin/edit.php?post_type=web-story',
   cdnURL: 'https://cdn.example.com/',
   version: '1.0.0-alpha.9',
+  archiveURL: 'https://example.com/',
```
swissspidy, line 84: Fixes some unrelated but quite annoying prop types warnings when running tests

packages/dashboard/src/app/views/editorSettings/index.js
```diff
@@ -425,6 +378,7 @@ function EditorSettings() {
         </Layout.Scrollable>
       </Wrapper>
 
+      {/* TODO: Remove this dialog, as it's not technically correct. */}
```
swissspidy, line 364: See #9096

includes/REST_API/Publisher_Logos_Controller.php
```diff
@@ -0,0 +1,435 @@
...
+	 */
+	protected function filter_publisher_logos( $publisher_logos ): array {
+		return array_filter(
+			$publisher_logos,
+			static function( $logo ) {
+				$post = get_post( $logo );
+
+				return ! ( ! $post || 'attachment' !== $post->post_type );
+			}
+		);
```
spacedmonkey:
```suggestion
		return array_filter( $publisher_logos, 'wp_attachment_is_image' );
```

Would this be a little cleaner?

includes/REST_API/Publisher_Logos_Controller.php
```diff
@@ -0,0 +1,435 @@
...
+		if ( rest_is_field_included( 'title', $fields ) ) {
+			$data['title'] = get_the_title( $post->ID );
+		}
+
+		if ( rest_is_field_included( 'active', $fields ) ) {
+			$active_publisher_logo_id = absint( get_option( Settings::SETTING_NAME_ACTIVE_PUBLISHER_LOGO ) );
+			$data['active']           = $post->ID === $active_publisher_logo_id;
+		}
+
+		if ( in_array( 'url', $fields, true ) ) {
```
spacedmonkey:
```suggestion
		if ( rest_is_field_included( 'url', $fields ) ) {
```

includes/REST_API/Publisher_Logos_Controller.php
```diff
@@ -0,0 +1,435 @@
...
+		$active_publisher_logo_id = absint( get_option( Settings::SETTING_NAME_ACTIVE_PUBLISHER_LOGO ) );
+
+		if ( $post->ID === $active_publisher_logo_id || ! in_array( $active_publisher_logo_id, $publisher_logos, true ) ) {
+			// Mark the first available publisher logo as the new default.
+			update_option( Settings::SETTING_NAME_ACTIVE_PUBLISHER_LOGO, ! empty( $publisher_logos[0] ) ? $publisher_logos[0] : 0 );
+		}
+
+		update_option( Settings::SETTING_NAME_PUBLISHER_LOGOS, $publisher_logos );
+
+		return $this->prepare_item_for_response( $post, $request );
```
spacedmonkey: Delete endpoints response normally like [this](https://github.com/WordPress/wordpress-develop/blob/cb5e2611a618e8de608469c5d56a6569cb6a1a01/src/wp-includes/rest-api/endpoints/class-wp-rest-widgets-controller.php#L387-L392).

includes/REST_API/Publisher_Logos_Controller.php
swissspidy: It would!

includes/REST_API/Publisher_Logos_Controller.php
```diff
@@ -0,0 +1,428 @@
...
+		$base = sprintf( '%s/%s', $this->namespace, $this->rest_base );
+
+		// Entity meta.
+		$links = [
+			'self'       => [
+				'href' => rest_url( trailingslashit( $base ) . $post->ID ),
+			],
+			'collection' => [
+				'href' => rest_url( $base ),
+			],
```
spacedmonkey, line 378:
```suggestion
			],
			'https://api.w.org/publisherlogo' => [
				'href'       => rest_url( sprintf( '%s/%s/%s', $this->namespace, 'media', $post->ID ) ),
				'embeddable' => true,
			],
```

Does it make sense to have the attachment rest embedded here?

includes/REST_API/Publisher_Logos_Controller.php
```diff
@@ -0,0 +1,428 @@
...
+		$fields = $this->get_fields_for_response( $request );
+
+		// Base fields for every post.
+		$data = [];
+
+		if ( rest_is_field_included( 'id', $fields ) ) {
+			$data['id'] = $post->ID;
+		}
+
+		if ( rest_is_field_included( 'title', $fields ) ) {
```
spacedmonkey, line 324: Instead of embedding title and url, does make sense to embed the attachment object, see https://github.com/google/web-stories-wp/pull/9101/files#r714074117

includes/REST_API/Publisher_Logos_Controller.php
```diff
@@ -0,0 +1,428 @@
...
+	 * Get an existing publisher logo's post object, if valid.
+	 *
+	 * @since 1.12.0
+	 *
+	 * @param int $id Supplied ID.
+	 * @return WP_Post|WP_Error Post object if ID is valid, WP_Error otherwise.
+	 */
+	protected function get_publisher_logo( $id ) {
+		$publisher_logos = $this->filter_publisher_logos( (array) get_option( Settings::SETTING_NAME_PUBLISHER_LOGOS, [] ) );
+
```
spacedmonkey, line 291: Should we have a check to see if `$publisher_logos` is an empty array?

includes/REST_API/Publisher_Logos_Controller.php
swissspidy: I thought about it, but I didn't see any benefits in doing so.

The current solution is super lean, with just like 4 fields that we need.

It would be the more RESTful way to do it, but in reality we would just end up using `?_fields` to request as few fields as needed and adding some custom JS function to get the data into the same shape as we have today. So just a lot of extra code to get to the same result.

includes/REST_API/Publisher_Logos_Controller.php
spacedmonkey, line 324: Should this follow [this](https://github.com/WordPress/wordpress-develop/blob/cb5e2611a618e8de608469c5d56a6569cb6a1a01/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php#L1769-L1781) core pattern.

includes/REST_API/Publisher_Logos_Controller.php
swissspidy: If it's an empty array the `in_array()` check below will simply fail. So don't see a need for an extra check here, seems redundant.

includes/REST_API/Publisher_Logos_Controller.php
swissspidy: Let's discuss in the other thread

swissspidy: We don't really need the raw title, as we only need the rendered title to put into the `alt` attribute of the ``. So kept it simple here to only return that.

includes/REST_API/Publisher_Logos_Controller.php
spacedmonkey: Is it worth adding the link so we could use later down the line?

spacedmonkey:
Couple of nicpic / improvements to PHP, javascript is looking fine. 

Will defer to @BrittanyIRL for javascript review here.

includes/REST_API/Publisher_Logos_Controller.php
swissspidy: If we should ever need this down the line, it can be easily added then

includes/REST_API/Publisher_Logos_Controller.php
```diff
@@ -0,0 +1,435 @@
...
+	 */
+	protected $namespace = 'web-stories/v1';
+
+	/**
+	 * The base of this controller's route.
+	 *
+	 * @since 1.12.0
+	 * @var string
+	 */
+	protected $rest_base = 'publisher-logos';
```
spacedmonkey:
```suggestion
	/**
	 * Constructor.
	 */
	public function __construct() {
		$this->namespace = 'web-stories/v1';
		$this->rest_base = 'publisher-logos';
	}
```

Nitpic, but other controllers do this. I don't like the current implementation, even through it does the same thing, because it implies you are defining those properties and not setting them.

spacedmonkey (approved):
One nicpic, but looking good from a PHP stand point.

includes/REST_API/Publisher_Logos_Controller.php
swissspidy: I'm not sure I understand the explanation, but either approach works for me. It's indeed exactly the same thing.

packages/dashboard/src/app/views/editorSettings/publisherLogo/gridItem.js
```diff
@@ -39,68 +39,61 @@ export const GridItem = forwardRef(
...
           onFocus={() => {
-            setActivePublisherLogoId(publisherLogo.id);
+            setFocusedPublisherLogoId(publisherLogo.id);
           }}
           data-testid={`uploaded-publisher-logo-${index}`}
-          isSelected={isActive}
-          tabIndex={isActive ? 0 : -1}
+          isSelected={isFocused}
+          tabIndex={isFocused ? 0 : -1}
           onClick={(e) => {
```
BrittanyIRL, line 84: this is fine because that dialog is about to be removed, but the reason for this was to return focus to the current logo if a user hits 'cancel' instead of confirming the delete, just sharing the background on it.

BrittanyIRL (approved):
What a lovely update!

Is it intentional that the default logo is now not moved to the front of the grouping? Personally, I think it's still just as clear. 

| new | old |
| --- | --- |
| Screen Shot 2021-09-23 at 9 18 18 AM | Screen Shot 2021-09-23 at 9 21 26 AM|

packages/dashboard/src/app/views/editorSettings/publisherLogo/gridItem.js
swissspidy: Not changing that logic here, just renaming the variable :)

swissspidy:
> Is it intentional that the default logo is now not moved to the front of the grouping? Personally, I think it's still just as clear.

Good point. I'll update the PHP code to sort them on the server side.

packages/dashboard/src/app/views/editorSettings/publisherLogo/gridItem.js
swissspidy: Fixed now

PR #9142, opened 2021-09-24
Make trim button functional

### LINKED ISSUE #9105: Implement actual trim
## Feature Description

Implement the connection between the trim button in the trim mode UI implemented in #8890 with the underlying trim functionality created by Infra in #8891.

If necessary, show a loader or relevant placeholder icon while the transformation occurs.

miina:
## Summary
Adds the missing part of making the trim work.

## Relevant Technical Choices
- Exits edit mode and displays a spinner for the time of Trimming.

## User-facing changes
| State | Image |
|----|-----|
| During trimming - design panel | Screenshot 2021-09-24 at 15 41 30 |
| During trimming - quick action menu (hidden Trim action) | Screenshot 2021-09-24 at 15 41 38 |
| Before trimming - origin video in Media Panel | Screenshot 2021-09-24 at 15 41 45 |
| After trimming - new video in Media Panel | Screenshot 2021-09-24 at 15 41 56 |

## Testing Instructions

1. Ensure that video trimming experiment is enabled.
2. Add a video (better to take a video that's a bit longer, e.g. 30 seconds, it will give more time to verify the things happening during trimming).
3. Click on the "Trim" button in the video options to ender Trim mode.
4. Trim the video from both ends using the handles.
5. Click "Trim" button that appeared on the right side of the trim bar.
6. Verify that Trim edit mode is exited.
7. Verify that Trim button in the design panel shows now "Trimming...", a spinner, and is disabled.
8. Verify that Mute button in the design panel below the Trim button is also disabled while trimming.
9. Verify that the "Trim" icon in the element menu (quick actions menu) is not visible while trimming.
10. Once trimming has ended, verify that a new video was added to the Media Library panel with correct length.

## Checklist

---

Fixes #9105


packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
```diff
@@ -159,14 +177,18 @@ function VideoOptionsPanel({ selectedElements, pushUpdate }) {
...
-            variant={BUTTON_VARIANTS.RECTANGLE}
-            type={BUTTON_TYPES.SECONDARY}
-            size={BUTTON_SIZES.SMALL}
-            onClick={toggleTrimMode}
-          >
-            {__('Trim', 'web-stories')}
-          </TrimButton>
+          <TrimWrapper>
+            <TrimButton
+              disabled={isTrimming}
```
miina: @spacedmonkey I saw that the Mute button is disabled when muting but it's hidden when trimming or transcoding. I made it similar for consistency, however, it would make more sense to me if both buttons would still be visible when transcoding/trimming/muting but just disabled. Thoughts?

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
```diff
@@ -159,14 +177,18 @@ function VideoOptionsPanel({ selectedElements, pushUpdate }) {
...
+              variant={BUTTON_VARIANTS.RECTANGLE}
+              type={BUTTON_TYPES.SECONDARY}
+              size={BUTTON_SIZES.SMALL}
+              onClick={toggleTrimMode}
+            >
+              {trimButtonText}
+            </TrimButton>
+            {isTrimming && <CircularProgress size={32} />}
```
miina: Added the circular progress here for making it more visible when trimming is actually happening. Let me know if this makes sense to you, then we could add it for Mute as well (otherwise will remove)

Screenshot 2021-09-24 at 13 36 45

packages/story-editor/src/components/videoTrim/provider.js
```diff
@@ -40,6 +48,40 @@ function VideoTrimProvider({ children }) {
...
+    const { resource } = selectedElements[0];
+    if (!resource) {
+      return;
+    }
+    trimExistingVideo({
+      resource,
+      start: msToHMS(startOffset),
+      end: msToHMS(endOffset),
+    });
+    toggleTrimMode();
```
miina, line 67: It seems to me that exiting the trim mode when starting trimming makes the most sense. Otherwise, the trim bar should be disabled anyway in trim mode and then once finished updated to the new values of the trimmed video. When existing the edit mode, the user can continue doing other stuff and the Trim button is disabled with the Loading circle to indicate the processing happening.

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
swissspidy: Looks a bit odd, but I like the thought.

WDYT about just changing the text to "Trimming…" with ellipsis?

packages/story-editor/src/components/videoTrim/provider.js
```diff
@@ -40,6 +48,40 @@ function VideoTrimProvider({ children }) {
     resetOffsets,
   } = useVideoNode();
 
+  /**
+   * Converts milliseconds to HH:MM:SS.SSS format.
+   *
+   * @type {function(*): string}
+   */
+  const msToHMS = useCallback((ms) => {
```
swissspidy: We have some similar function already in the media package. See `getVideoLength` and `formatDuration`

Haven't looked closely, but perhaps it's possible to reuse something there by extracting into a shared function or something.

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
miina: You mean instead of using the spinner?

Could be, it's just that it's not very easily noticeable maybe. E.g. I realized the text of the Mute button was changing for the time of muting only when I saw it in code.

What if it's a bit smaller, e.g.


https://user-images.githubusercontent.com/3294597/134668538-6983557d-9772-42a8-b817-7fa72952100c.mp4

If you still think it looks odd, let's use the `Trimming...` and we could bring it up with UX to follow up later for both buttons.

swissspidy: > What if it's a bit smaller, e.g.

Yeah that helps 👍 

> If you still think it looks odd, let's use the Trimming...

Let's do both actually, while at it

packages/story-editor/src/components/videoTrim/provider.js
miina: Looks like the logic in `getVideoLength` is a bit different and is using full seconds. Added a new util using `formatDuration` and added it to the media package.

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
```diff
@@ -465,7 +465,12 @@ const useQuickActions = () => {
   );
 
   const videoCommonActions = useMemo(() => {
-    return hasTrimMode
+    const resource = selectedElements?.[0]?.resource;
+    if (!resource) {
+      return [];
+    }
+    const { isTrimming, isMuting, isTranscoding } = resource;
+    return !isTrimming && !isMuting && !isTranscoding && hasTrimMode
```
spacedmonkey: This is going to change in https://github.com/google/web-stories-wp/pull/9098

Is this check for isTrimming etc part of hasTrimMode?
https://github.com/google/web-stories-wp/blob/36f34ea5d9995fb4c5f38d68cb605ab6e6c87a2c/packages/story-editor/src/components/videoTrim/useVideoTrimMode.js#L59-L74

packages/story-editor/src/components/videoTrim/provider.js
```diff
@@ -40,6 +49,25 @@ function VideoTrimProvider({ children }) {
     resetOffsets,
   } = useVideoNode();
 
+  const performTrim = useCallback(() => {
+    const { resource } = selectedElements[0];
+    if (!resource) {
+      return;
+    }
+    trimExistingVideo({
+      resource,
```
spacedmonkey:
```suggestion
    trimExistingVideo({
      resource: {
          ...resource, 
          length: (endOffset * 1000) - (startOffset * 1000)
          lengthFormatted: formatMsToHMS(endOffset - startOffset)
      },
```

We could do this, so preview time stamp works workly.

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
miina: > Is this check for isTrimming etc part of hasTrimMode?

It was but I removed it since otherwise the button gets hidden when `isTrimming` but it should just be disabled.

packages/story-editor/src/components/videoTrim/provider.js
miina: Ah, good point, thanks!

spacedmonkey: I worked on a PR - https://github.com/google/web-stories-wp/pull/9143

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
spacedmonkey: If that is the case, can you code review #9098

miina: @spacedmonkey This is a good place to use `canTranscodeResource` that you added in #9098.

packages/media/src/formatMsToHMS.js
```diff
@@ -0,0 +1,53 @@
...
+import formatDuration from './formatDuration';
+
+/**
+ * Converts length in milliseconds to HH:MM:SS.SSS format.
+ *
+ * @param {number} ms Original length in milliseconds.
+ * @param {boolean} isForDisplay If the output is for displaying to the user.
+ * @return {string} Formatted length.
+ */
+function formatMsToHMS(ms, isForDisplay = false) {
```
miina: @spacedmonkey @swissspidy I've added a `isForDisplay` flag here now for formatting ms to H:M:S format either for displaying for the user or not. Let me know if you think it's better to split it out to two separate utils as Jonny had [in his PR](https://github.com/google/web-stories-wp/pull/9143/files#diff-6e2ce5102e38dfd909c13da6ea6cdd896c0a7b5e8cce6a094606a7f73c54cdf8R28).

swissspidy: Separate functions seem cleaner, boolean flag params are often considered an anti-pattern.

The shared part (getting an array of hours, minutes, seconds from ms) could be extracted into a shared function

miina: Yep, I also felt like it got a bit confusing. Will split it up.

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
```diff
@@ -105,19 +114,23 @@ function VideoOptionsPanel({ selectedElements, pushUpdate }) {
...
+      ? __('Removing audio…', 'web-stories')
       : __('Remove audio', 'web-stories');
   }, [isMuting]);
 
+  const trimButtonText = useMemo(() => {
+    return isTrimming
+      ? __('Trimming…', 'web-stories')
+      : __('Trim', 'web-stories');
+  }, [isTrimming]);
+
```
spacedmonkey: The useMemos here were removed this in this PR https://github.com/google/web-stories-wp/pull/9098. Just noted

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
```diff
@@ -159,28 +184,33 @@ function VideoOptionsPanel({ selectedElements, pushUpdate }) {
...
-            variant={BUTTON_VARIANTS.RECTANGLE}
-            type={BUTTON_TYPES.SECONDARY}
-            size={BUTTON_SIZES.SMALL}
-            onClick={toggleTrimMode}
-          >
-            {__('Trim', 'web-stories')}
-          </TrimButton>
+          <TrimWrapper>
+            <TrimButton
+              disabled={isTrimming || isMuting || isTranscoding}
```
barklund: Won't `isTranscoding` be true if either of the two other flags is true? So we can just check that flag only? Or am I misunderstanding something?

spacedmonkey:
@miina This will need to be rebased now that #9098 is merged

barklund (approved):
This works really great. I feel we should be better at informing the user what's going on (a progress bar with percentage), but it's a pretty awesome feature.

I even tried adding three instances of the same video, trimming two of them and muting the last – at the same time. And it just worked! Great work here!

Once @spacedmonkey is happy, I'm happy.

spacedmonkey:
> I feel we should be better at informing the user what's going on (a progress bar with percentage), 

@barklund I had the same thought, which is why I created https://github.com/google/web-stories-wp/issues/9158 https://github.com/google/web-stories-wp/issues/9159

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
swissspidy: `isTranscoding` is only `true` when compressing/optimizing videos. Not when any general ffmpeg operation is in progress.

spacedmonkey (approved):
Looks good to me

PR #9143, opened 2021-09-24
Fix the length of the video while processing.

spacedmonkey:
## Summary

Break out from #9142. 

While the video is processed and uplaoded, ensure that the length / timestamp is correct. 

## Testing Instructions

1.

## Checklist

---

Fixes #

### DIFF
--- a/packages/media/src/formatTimeFromLength.js
+++ b/packages/media/src/formatTimeFromLength.js
@@ -17,23 +17,24 @@
 /**
  * Internal dependencies
  */
-import preloadVideoMetadata from './preloadVideoMetadata';
-import formatTimeFromLength from './formatTimeFromLength';
+import formatDuration from './formatDuration';
 
 /**
- * Get video length from a video.
+ * Format time to be human readable
  *
- * @param {string} src Video source.
- * @return {Promise} Video length object.
+ * @param {number} length Length in seconds.
+ * @return string
  */
-const getVideoLength = async (src) => {
-  const video = await preloadVideoMetadata(src);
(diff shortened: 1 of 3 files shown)


packages/media/src/formatTimeFromLength.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+import formatDuration from './formatDuration';
+
+/**
+ * Format time to be human readable
+ *
+ * @param {number} length Length in seconds.
+ * @return string
+ */
+const formatTimeFromLength = ( length ) => {
```
swissspidy, line 28: Can we add tests?

Also, does this make `formatMsToHMS` obsolete? Or how do they relate?

swissspidy (changes requested):
Some open questions and lint errors

packages/media/src/formatTimeFromLength.js
miina: Looks like the difference is that this one takes in the original value in seconds. These could still probably be unified, it could always take milliseconds in.

miina: So perhaps we could just adjust `formatMsToHMS` to skip the hours when irrelevant and just use it with `length * 1000` in `getVideoLength` and then we wouldn't need a second formatter?

miina: @spacedmonkey I'll make that change directly in the origin PR, hope you don't mind.

miina:
Thanks, Jonny! Closing in favour of the origin PR (#9105) to make changes directly there.

Issue #9149, opened 2021-09-25
Taxonomy: e2e tests enabled 


BrittanyIRL:
Originally included e2e tests on #9011 to verify that user capabilities are being expected are having trouble passing remotely. They work fine locally. Isolating this work because of deadlines. 

## Acceptance criteria 
- Taxonomy e2e tests passing and more robust

PR #9150, opened 2021-09-25
Improve taxonomy e2e tests

swissspidy:
See #9099
See #9149

### DIFF
--- a/packages/e2e-tests/src/config/bootstrap.js
+++ b/packages/e2e-tests/src/config/bootstrap.js
@@ -25,6 +25,7 @@ import {
   setCurrentUser,
   trashAllPosts,
   deleteAllMedia,
+  trashAllTerms,
 } from '@web-stories-wp/e2e-test-utils';
 
 // Extend Jest matchers.
@@ -242,6 +243,8 @@ beforeAll(async () => {
   await setCurrentUser('admin', 'password');
   await trashAllPosts();
   await trashAllPosts('web-story');
+  await trashAllTerms('web_story_category');
+  await trashAllTerms('web_story_tag');
   await deleteAllMedia();
 });
 
--- a/packages/e2e-tests/src/config/bootstrap.js
+++ b/packages/e2e-tests/src/config/bootstrap.js
@@ -167,6 +167,17 @@ function observeConsoleLogging() {
       return;
     }
 
+    // Special case: ignore 403 errors on logout page.
+    // See https://github.com/google/web-stories-wp/pull/7889
+    if (
+      text.includes(
+        'Failed to load resource: the server responded with a status of 403 (Forbidden)'
+      ) &&
+      message.stackTrace()?.[0]?.url?.endsWith('wp-login.php?action=logout')
+    ) {
+      return;
+    }
+
     //eslint-disable-next-line security/detect-object-injection
     const logFunction = OBSERVED_CONSOLE_MESSAGE_TYPES[type];
 
--- a/packages/e2e-tests/src/specs/editor/taxonomy.js
+++ b/packages/e2e-tests/src/specs/editor/taxonomy.js
@@ -13,88 +13,117 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 /**
  * External dependencies
  */
 import {
   createNewStory,
-  publishStory,
+  withExperimentalFeatures,
   withUser,
+  publishStory,
 } from '@web-stories-wp/e2e-test-utils';
 
-describe('Taxonomy', () => {
-  const goToAndExpandTaxonomyPanel = async () => {
-    await expect(page).toClick('li[role="tab"]', { text: 'Document' });
-    await expect(page).toMatch('Categories and Tags');
+async function goToAndExpandTaxonomyPanel() {
+  await expect(page).toClick('li[role="tab"]', { text: 'Document' });
+  await expect(page).toMatch('Categories and Tags');
 
-    const taxonomyPanel = await page.$(
-      'button[aria-label="Categories and Tags"]'
-    );
+  const taxonomyPanel = await page.$(
+    'button[aria-label="Categories and Tags"]'
+  );
 
-    const isCollapsed = await page.evaluate(
-      (button) => button.getAttribute('aria-expanded') == 'false',
-      taxonomyPanel
-    );
-    if (isCollapsed) {
-      await taxonomyPanel.click();
-    }
-  };
+  const isCollapsed = await page.evaluate(
+    (button) => button.getAttribute('aria-expanded') === 'false',
+    taxonomyPanel
+  );
 
-  const addChildCategory = async ({ parent, child }) => {
-    await expect(page).toClick(
-      'button[data-testid="expand_add_new_hierarchical_term"]'
-    );
-    await page.type('input[name="New Category Name"]', child);
-    await expect(page).toClick('button[aria-label="Parent Category"]');
+  if (isCollapsed) {
+    await taxonomyPanel.click();
+  }
+}
+
+async function addCategory(name, parent) {
+  await expect(page).toClick('button', { text: 'Add New Category' });
+  await page.type('input[name="New Category Name"]', name);
+
+  if (parent) {
+    await expect(page).toClick('button', { text: 'Parent Category' });
     await expect(page).toClick('li[role="option"]', { text: parent });
-    await expect(page).toClick(
-      'button[data-testid="submit_add_new_hierarchical_term"]'
-    );
-  };
+  }
 
-  // TODO  https://github.com/google/web-stories-wp/issues/9149
-  // eslint-disable-next-line jest/no-disabled-tests
-  it.skip('should be able to add new categories', async () => {
+  await page.focus('button[data-testid="submit_add_new_hierarchical_term"]');
+
+  await expect(page).toClick(
+    'button[data-testid="submit_add_new_hierarchical_term"]',
+    { text: 'Add New Category' }
+  );
+
+  // Todo: Investigate why this sometimes takes longer to appear.
+  await expect(page).toMatch(name);
+}
+
+describe('Taxonomy', () => {
+  withExperimentalFeatures(['enableTaxonomiesSupport']);
+
+  // Create some categories before running all tests so that they are available there.
+  beforeAll(async () => {
     await createNewStory();
     await goToAndExpandTaxonomyPanel();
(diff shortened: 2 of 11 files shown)


packages/e2e-tests/src/specs/editor/taxonomy.js
```diff
@@ -13,88 +13,117 @@
...
+    (button) => button.getAttribute('aria-expanded') === 'false',
+    taxonomyPanel
+  );
+
+  if (isCollapsed) {
+    await taxonomyPanel.click();
+  }
+}
+
+async function addCategory(name, parent) {
```
BrittanyIRL, line 50: nice update here

packages/e2e-tests/src/config/bootstrap.js
```diff
@@ -242,6 +243,8 @@ beforeAll(async () => {
   await setCurrentUser('admin', 'password');
   await trashAllPosts();
   await trashAllPosts('web-story');
+  await trashAllTerms('web_story_category');
+  await trashAllTerms('web_story_tag');
```
BrittanyIRL, line 258: ❤️

BrittanyIRL:
Thanks for updating utils surrounding this and for the clean up. 

I'm calling it for the weekend. I tweaked a few things if you look at commits. Summarized - 

Karma tests now pass:
- Fixed tests which were trying to match checkboxes by name, name got updated in component - needed to update in karma test too
- Fixed tests that were defaulting to `parentId: undefined` when they should have `0` 

e2e tweaks: 
- removed taxonomy experiment getting toggled in `trashAllTerms` because we've got the `useExperimentalFeature` hook getting called before that so it's redundant. 
- adding `await publishStory` to the `beforeAll` on taxonomies so that categories get saved and can be found by other stories, now that history's working we need that. 
- The contributor test is still a little funky - it seems like half the time the `beforeAll` is running and trying to add in categories and fails because it's a contributor so it can't? I tried to simplify it just to limit potential things going wrong.

swissspidy:
Thanks for chiming, despite the weekend! Don't make it the norm though, this is not urgent :-)

> * removed taxonomy experiment getting toggled in `trashAllTerms` because we've got the `useExperimentalFeature` hook getting called before that so it's redundant.

This call in `trashAllTerms` is necessary though. `trashAllTerms` should also be called when excluding the taxonomy specific test just so we have a clean sheet everywhere.

Also, it does not hurt to keep it in. If you want to enable the experiment but it's already enabled, it's simply a no-op.

> * adding `await publishStory` to the `beforeAll` on taxonomies so that categories get saved and can be found by other stories, now that history's working we need that.

Is this a new requirement? 🤔 

In Gutenberg, when I add a new category, it gets automatically created via REST API and then assigned to the post. Pressing Undo does undo the assignment, but not the creation.

So requiring publishing/saving here sounds like a bug to me.

> * The contributor test is still a little funky - it seems like half the time the `beforeAll` is running and trying to add in categories and fails because it's a contributor so it can't? I tried to simplify it just to limit potential things going wrong.

Hmm in my testing it wasn't an issue with the user being the wrong one, but with the creation simply taking a bit too long and the dropdown not closing etc.

packages/e2e-tests/src/specs/editor/taxonomy.js
```diff
@@ -13,88 +13,119 @@
...
       );
-      expect(addNewCategoryButton).toBeNull();
+      await expect(page).toClick('label', { text: 'rock' });
+      await expect(page).toMatchElement(
+        'input[name="hierarchical_term_rock"][checked]'
+      );
+      // Todo: this is giving us what we expect but it's timing out because nothing is found.
+      // await expect(page).not.toMatchElement('button', {
+      //   text: 'Add New Category',
+      // });
```
swissspidy, line 126: Looks like the issue is that the `Checkbox` component somehow doesn't add the `checked` attribute in the DOM when you toggle it. You'll see the checkmark though.

When saving the story and refreshing the page, the `checked` attribute _is_ there.

packages/e2e-tests/src/specs/editor/taxonomy.js
```diff
@@ -13,88 +13,119 @@
...
-      const addNewCategoryButton = await page.$(
-        'button[aria-label="Add New Category"]'
+      await expect(page).not.toMatchElement(
+        'input[name="hierarchical_term_rock"][checked]'
       );
-      expect(addNewCategoryButton).toBeNull();
+      await expect(page).toClick('label', { text: 'rock' });
+      await expect(page).toMatchElement(
+        'input[name="hierarchical_term_rock"][checked]'
+      );
```
swissspidy: Looks like the issue is that the `Checkbox` component somehow doesn't add the `checked` attribute in the DOM when you toggle it. You'll see the checkmark though.

When saving the story and refreshing the page, the `checked` attribute _is_ there.

BrittanyIRL: huh 🤔  i wonder if this has something to do with saving categories as state before adding them to the actual story on a story save.

BrittanyIRL:
> Is this a new requirement? 🤔
> 
> In Gutenberg, when I add a new category, it gets automatically created via REST API and then assigned to the post. Pressing Undo does undo the assignment, but not the creation.

Nah we are good, this isn't an issue - i thought it was because categories were disappearing when running puppeteer slowly and watching but it was probably just timing. 


Makes sense about trashing terms! thanks for adding that back in.

packages/e2e-tests/src/specs/editor/taxonomy.js
BrittanyIRL: Suppose for now we could just check that the term exists, not if it's checked or not.

BrittanyIRL: (@BrookeGraham  is gonna look into the checkbox issue today/tomorrow)

BrookeGraham: Opened a pr for the checkbox fix. https://github.com/google/web-stories-wp/pull/9199

swissspidy:
@BrittanyIRL e2e tests are now passing 🥳 

There's an unrelated `Duplicate > pastes copied element in a new location and selects it` karma test failure that just seems to be flakey and can be ignored.

packages/wp-story-editor/src/components/header/buttons/index.js
```diff
@@ -76,7 +76,7 @@ function Buttons() {
     })
   );
 
-  const isDraft = 'draft' === status;
+  const isDraft = 'draft' === status || !status;
```
swissspidy, line 79: Fixes an unrelated bug I noticed during testing.

While the story is being loaded in the editor, the `status` is `undefined`.

This caused the `SwitchToDraftButton` button to briefly render before changing to the `UpdateButton`.

This fix prevents that flickering.

packages/e2e-tests/src/specs/editor/taxonomy.js
```diff
@@ -13,88 +13,156 @@
...
+    (button) => button.getAttribute('aria-expanded') === 'false',
+    taxonomyPanel
+  );
+
+  if (isCollapsed) {
+    await taxonomyPanel.click();
+  }
+
+  // Small trick to ensure we scroll to this panel.
+  await taxonomyPanel.focus();
```
BrittanyIRL, line 47: SMART

packages/e2e-tests/src/specs/editor/taxonomy.js
```diff
@@ -13,88 +13,156 @@
...
+}
+
+async function addCategory(name, parent) {
+  await expect(page).toClick('button', { text: 'Add New Category' });
+  await expect(page).toFill('input[name="New Category Name"]', name);
```
BrittanyIRL, line 52: today I learned `toFill`

packages/e2e-tests/src/specs/editor/taxonomy.js
```diff
@@ -13,88 +13,156 @@
...
-  // eslint-disable-next-line jest/no-disabled-tests
-  it.skip('should be able to add new categories', async () => {
+    await expect(page).toMatchElement('button[aria-label="Parent Category"]', {
+      text: parent,
+    });
+
+    // TODO: Add assertion here to verify the parent was chosen and the dropdown closed.
+  }
+
+  await page.keyboard.press('Enter');
```
BrittanyIRL, line 70: nice work around to make sure it's always submitted.

packages/e2e-tests/src/specs/editor/taxonomy.js
```diff
@@ -13,88 +13,156 @@
...
-    // Save terms so that they are available to contributor
-    await publishStory();
+    // No need to save/publish the story, as the new categories will have been
+    // created in the background via the REST API already.
   });
+
+  describe('Administrator', () => {
+    it('should be able to add new categories', async () => {
+      await createNewStory();
+      await insertStoryTitle('Taxonomies - Categories - Admin');
```
BrittanyIRL, line 95: It's interesting to me that the story title is necessary to get the test to work. Was this because it gives the rest of the page more time to be 'ready' and avoid a timeout or is the title needed for something else here?

BrittanyIRL (approved):
WOW THANK YOU. This took so many tweaks and tries and thank you so much. I am learning _alot_ about e2e tests here. The awaits you added are clever to avoid timeouts and the previous reasons they were failing when I'd make attempts at fixing them make more sense with your solutions for making some helpers async.

packages/e2e-tests/src/specs/editor/taxonomy.js
swissspidy: It's not a default puppeteer method, but a helper util in [`expect-puppeteer`](https://github.com/smooth-code/jest-puppeteer/tree/master/packages/expect-puppeteer/) that we use. Just a tiny wrapper around `page.type()`/`element.type()` basically.

packages/e2e-tests/src/specs/editor/taxonomy.js
swissspidy: It's not necessary at all :-)

But it helps with debugging & the percy snapshots to know which step you're currently on

packages/e2e-tests/src/specs/editor/taxonomy.js
swissspidy: This also helped with the dropdown/popup not disappearing, which is the part that was never functioning properly when simply clicking on the `li`.

PR #9172, opened 2021-09-27
Page Templates: add screenshot urls to template

### LINKED ISSUE #9091: Add template screenshot asset src urls on export
## Description

Because both the editor and dashboard will now be relying on the same static images we can set up the webp and png for each page in templates/src/index by default and avoid having to set that up in two places with the dashboard and now the editor.  (Preemptive cleanup to move forward)

## Acceptance Criteria

In the Dashboard's explore templates and template details, screenshots are being rendered fine but don't need to get added to the data via serializer anymore.

BrookeGraham:
## Context
In story editor, we will be using the cli generated screenshots as the images for all the layouts. 

## Summary
Instead of recreating the image urls in both the dashboard templates and the editor templates, we are placing the src urls within the templates export. 

## Relevant Technical Choices
Removed `posterByPage` from Dashboard -> Serializer -> reshapeTemplateObject and instead exported as part of the template. 

## To-do

n/a

## Testing Instructions

1. 

no

no

no

## Checklist

---

Fixes #9091

### DIFF
--- a/packages/dashboard/src/app/serializers/test/templates.js
+++ b/packages/dashboard/src/app/serializers/test/templates.js
@@ -32,27 +32,14 @@ describe('reshapeTemplateObject', () => {
       pages: [{ id: 1 }, { id: 2 }, { id: 3 }, { id: 4 }],
       creationDate: '2020-03-26T20:57:24',
       modified: '2020-03-26T20:57:24',
+      status: 'template',
     };
 
-    const reshapedObj = reshapeTemplateObject(responseObj, 'example.com/');
+    const reshapedObj = reshapeTemplateObject(responseObj, false);
 
     expect(reshapedObj).toMatchObject({
       id: 1,
       slug: 'template-1-slug',
-      postersByPage: {
-        0: {
-          webp: 'example.com/images/templates/template-1-slug/posters/1.webp',
-          png: 'example.com/images/templates/template-1-slug/posters/1.png',
-        },
-        1: {
-          webp: 'example.com/images/templates/template-1-slug/posters/2.webp',
-          png: 'example.com/images/templates/template-1-slug/posters/2.png',
-        },
-        2: {
-          webp: 'example.com/images/templates/template-1-slug/posters/3.webp',
-          png: 'example.com/images/templates/template-1-slug/posters/3.png',
-        },
-      },
       centerTargetAction: `${APP_ROUTES.TEMPLATE_DETAIL}?id=1&isLocal=false`,
       creationDate: toDate('2020-03-26T20:57:24.000Z'),
       status: 'template',
@@ -66,7 +53,7 @@ describe('reshapeTemplateObject', () => {
         pages: [{ id: 1 }, { id: 2 }, { id: 3 }, { id: 4 }],
         creationDate: '2020-03-26T20:57:24',
       },
-      'example.com/'
+      true
     );
 
     expect(reshapedObj).toBeNull();
--- a/packages/templates/src/index.js
+++ b/packages/templates/src/index.js
@@ -31,24 +31,10 @@ export default async function ({ cdnURL }) {
   };
 
   return templates.map((template, index) => {
-    const postersByPage = template.pages.reduce((memo, _, i) => {
-      const srcPath = `${cdnURL}images/templates/${template.slug}/posters/${
-        i + 1
-      }`;
-      return {
-        ...memo,
-        [i]: {
-          webp: `${srcPath}.webp`,
-          png: `${srcPath}.png`,
-        },
-      };
-    }, {});
-
     return {
       id: index + 1,
       ...globalConfig,
       ...template,
-      postersByPage,
     };
   });
 }
--- a/packages/templates/src/index.js
+++ b/packages/templates/src/index.js
@@ -31,10 +31,24 @@ export default async function ({ cdnURL }) {
   };
 
   return templates.map((template, index) => {
+    const postersByPage = template.pages.reduce((memo, _, i) => {
+      const srcPath = `${cdnURL}images/templates/${template.slug}/posters/${
+        i + 1
+      }`;
+      return {
+        ...memo,
+        [i]: {
+          webp: `${srcPath}.webp`,
+          png: `${srcPath}.png`,
+        },
+      };
+    }, {});
+
     return {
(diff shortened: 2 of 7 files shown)


packages/templates/src/index.js
```diff
@@ -31,10 +31,24 @@ export default async function ({ cdnURL }) {
...
+      }`;
+      return {
+        ...memo,
+        [i]: {
+          webp: `${srcPath}.webp`,
+          png: `${srcPath}.png`,
+        },
+      };
+    }, {});
+
```
BrookeGraham: Gonna move this to the `getTemplate` so that it can be memoized as well.

packages/dashboard/src/app/serializers/test/templates.js
```diff
@@ -32,27 +32,14 @@ describe('reshapeTemplateObject', () => {
...
-        },
-        1: {
-          webp: 'example.com/images/templates/template-1-slug/posters/2.webp',
-          png: 'example.com/images/templates/template-1-slug/posters/2.png',
-        },
-        2: {
-          webp: 'example.com/images/templates/template-1-slug/posters/3.webp',
-          png: 'example.com/images/templates/template-1-slug/posters/3.png',
-        },
-      },
```
BrookeGraham, line 55: Since this no longer happens in this serializer, I removed the tests here. Should I create new test for `getTemplates`? @BrittanyIRL

BrookeGraham: There currently aren't any tests in templates for `getTemplates`.

BrittanyIRL: 🤔  we've only got tests for the `raw` templates right now, i think adding in a test to make sure they are formatting accurately is a great idea.

packages/templates/src/test/getTemplates.js
```diff
@@ -0,0 +1,66 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * External dependencies
+ */
+
```
maxyinger: remove?

packages/templates/src/test/getTemplates.js
```diff
@@ -0,0 +1,66 @@
...
+              `example.com/images/templates/${template.slug}/posters/${
+                i + 1
+              }.png`
+            ),
+            webp: expect.stringContaining(
+              `example.com/images/templates/${template.slug}/posters/${
+                i + 1
+              }.webp`
+            ),
+          })
```
maxyinger: I think `toStrictEqual()` tests to see if the objects serialize to the same value. So you should be able to remove those `expect.Containing()` s.

this should work I think:
```
        expect(page).toStrictEqual({
            png: `example.com/images/templates/${template.slug}/posters/${
                i + 1
              }.png`,
            webp: `example.com/images/templates/${template.slug}/posters/${
                i + 1
              }.webp`,
        })
```

packages/templates/src/test/getTemplates.js
```diff
@@ -0,0 +1,66 @@
...
+
+    templates.forEach((template) => {
+      template.pages.forEach((page) => {
+        page.elements.forEach((element) => {
+          expect(element.resource?.src).not.toBe(
+            '__WEB_STORIES_TEMPLATE_BASE_URL__/'
+          );
+          expect(element.resource?.poster).not.toBe(
+            '__WEB_STORIES_TEMPLATE_BASE_URL__/'
+          );
```
maxyinger: Maybe better to use this:
https://jestjs.io/docs/expect#expectnotstringcontainingstring

Just cus I think this would pass for an element with a `poster` or `src` value of:
```
'__WEB_STORIES_TEMPLATE_BASE_URL__/some/path/after'
```

maxyinger (approved):
Looks great!

packages/templates/src/test/getTemplates.js
maxyinger: Ah just saw your other PR where you add a `type` property to this object. Ignore this comment

packages/templates/src/test/getTemplates.js
```diff
@@ -0,0 +1,61 @@
...
+
+    templates.forEach((template) => {
+      template.pages.forEach((page) => {
+        page.elements.forEach((element) => {
+          expect(element.resource?.src).toStrictEqual(
+            expect.not.stringContaining('__WEB_STORIES_TEMPLATE_BASE_URL__/')
+          );
+          expect(element.resource?.poster).toStrictEqual(
+            expect.not.stringContaining('__WEB_STORIES_TEMPLATE_BASE_URL__/')
+          );
```
samwhale, line 67: It'd be good to add a quick check here to make sure that the image url is also in the string
```suggestion
          expect(element.resource?.src).toStrictEqual(
            expect.not.stringContaining('__WEB_STORIES_TEMPLATE_BASE_URL__/')
          );
          expect(element.resource?.src).toStrictEqual(
            expect.stringContaining('example.com/')
          );
          expect(element.resource?.poster).toStrictEqual(
            expect.not.stringContaining('__WEB_STORIES_TEMPLATE_BASE_URL__/')
          );
          expect(element.resource?.poster).toStrictEqual(
            expect.stringContaining('example.com/')
          );
```

samwhale (approved):

BrittanyIRL (approved):
👍 

one quick small thing, can you update the PropType in dashboard/src/types TemplatePropType like so to remove the console warning about it being wrong? 
Screen Shot 2021-10-01 at 2 12 43 PM

packages/templates/src/test/getTemplates.js
BrookeGraham: I didn't add the `strictEqual` check because  sometimes `resource` is `undefined`, so the test will fail in those cases.

swissspidy:
What's missing here to merge this?

BrookeGraham:
@swissspidy I was rerunning tests since they were failing (timed out). I am happy to merge if you think those failed tests are good to go.

swissspidy:
Seems to be a flakey test in the editor. I'll keep an eye on it.

PR #9200, opened 2021-09-29
Docs: Explain amp-story-auto-analytics migration

dreamofabear:
## Context

User documentation for #7739.

## Summary

Adds an explanation and a table comparing the old vs. new analytics config.

## Relevant Technical Choices

FYI there seems to be some broken images (in prod too) which are flagged as errors during build. 

For example in https://wp.stories.google/docs/#Adding-borders this image is missing: https://wp.stories.google/docs/img/docs/corner-radius.png

I ended up manually editing `public/docs/index.html` instead of committing my local build artifacts.

## User-facing changes

Just documentation updates.

---

Partial for #7739.


swissspidy (approved):

swissspidy:
> FYI there seems to be some broken images (in prod too) which are flagged as errors during build.

Ugh, crap. No idea how that happened.

I'll try to see if I can resurrect them somehow.

But large parts of the content & images are quite outdated anyway and need an overhaul.

PR #9201, opened 2021-09-29
Taxonomies: support custom taxonomies

samwhale:
## Summary

Adds some custom taxonomies that existed before (color, verticals). Playing with the labeling and the set up of these custom taxonomies.

The new taxonomies are prefixed with `web_story_`.

## Testing Instructions

1. run `composer install` to set up the new taxonomies
2. Run the editor. Should see the new taxonomies on a network request where the story is being fetched under `_links -> wp:term`:

image

## Checklist

---

Fixes #


includes/Taxonomy/Color_Taxonomy.php
```diff
@@ -0,0 +1,95 @@
...
+			'most_used'             => _x( 'Most Used', 'Colors', 'web-stories' ),
+			'back_to_items'         => __( '&larr; Go to Colors', 'web-stories' ),
+			'item_link'             => _x( 'Color Link', 'navigation link block title', 'web-stories' ),
+			'item_link_description' => _x( 'A link to a color.', 'navigation link block description', 'web-stories' ),
+		];
+		$args   = [
+			'labels'                => $labels,
+			'hierarchical'          => true,
+			'public'                => false,
+			'show_ui'               => false,
```
samwhale, line 85: Currently these will not show because the `show_ui` field is `false`

samwhale:
When enabled, the UI is built out for `hierarchical` and `non-hierarchical` (flat term selection):

BrittanyIRL:
Hey @swissspidy / @spacedmonkey  - can you take a look at this? Yesterday it got brought up that non category and tag taxonomies aren't showing up yet. Is this the path forward, are you good with it?

swissspidy:
> Hey @swissspidy / @spacedmonkey - can you take a look at this? Yesterday it got brought up that non category and tag taxonomies aren't showing up yet. Is this the path forward, are you good with it?

@BrittanyIRL Not sure what exactly we would want to move forward with here?

This PR just registers some dummy taxonomies. I originally created those in my Gist for development. We don't need those in the actual plugin.

I think Jonny's concern yesterday was that the taxonomy panel **used to be ** hardcoded to categories & tags only. As this PR proves, that is no longer the case.

So I think we're all good & this can be closed.

spacedmonkey:
I think this PR should be closed in favour of #9214.

BrittanyIRL:
> > Hey @swissspidy / @spacedmonkey - can you take a look at this? Yesterday it got brought up that non category and tag taxonomies aren't showing up yet. Is this the path forward, are you good with it?
> 
> @BrittanyIRL Not sure what exactly we would want to move forward with here?
> 
> This PR just registers some dummy taxonomies. I originally created those in my Gist for development. We don't need those in the actual plugin.
> 
> I think Jonny's concern yesterday was that the taxonomy panel **used to be ** hardcoded to categories & tags only. As this PR proves, that is no longer the case.
> 
> So I think we're all good & this can be closed.

Great, mostly just wanted to level set with what we needed moving forward. #9214  does that. 👍

PR #9203, opened 2021-09-29
Add re-trimming from original video

### LINKED ISSUE #8892: Re-trimming a video
## Feature Description

Once #8891 has been implemented, make sure that re-trimming a video, that has already been trimmed, will work off of the original video and the offsets relative to that video rather than based on the current (shorter) video. This information will exist in the metadata of the trimmed video element.

However, if this link is somehow broken (e.g. if the original video has been deleted), use the current video as a new blank starting point.

miina:
## Summary
Adds re-trimming for videos so that the video is always trimmed from the original video instead of the current video.

## To-do

- [x] Make sure the original video is actually used for trimming as well

## Testing Instructions

1. Add a video element, longer (e.g. 30 seconds) is better for testing.
2. Trim it very short (e.g. 2 seconds)
3. Now enter trim mode again. Verify that the video trim menu still shows the full length of the video and the "active" part is accurate length trimmed in the previous step.
4. Now trim the video longer again (e.g. 15 sec)
5. Verify it works as expected, a new video with accurate length is created, and when you open the trim mode again, the video trim menu shows the full length again.

## Checklist

---

Fixes #8892


packages/story-editor/src/elements/video/trim.js
```diff
@@ -89,17 +126,54 @@ function VideoTrim({ box, element }) {
...
+  }, [
+    original,
+    getMediaById,
+    setOriginalResource,
+    start,
+    end,
+    setOriginalEndOffset,
+    setOriginalStartOffset,
+    getMsFromHMS,
+  ]);
```
barklund: I know this is only WIP, but I'd really like to see this moved to the provider instead if possible. Maybe as a third custom hook there or alternatively built into the `useVideoNode` hook.

miina: @barklund Thanks for checking. 

Some refactoring has been done now -- if you have time, please review the PR to see if you notice any other general places for improvement. There are still a few outstanding things but generally it should be getting ready-ish.

packages/story-editor/src/elements/video/editMenu.js
```diff
@@ -19,17 +19,28 @@
...
+    })
+  );
 
   if (!isTrimMode) {
     return false;
   }
-
+  // Don't display trimmer if the original resource isn't loaded.
+  if (shouldFetchResource(resource?.trimData || {})) {
+    return false;
```
miina, line 37: Since the `editMenu` is using the general values from the provider and not values per the selected video element then unless we hide it for the time of fetching the original resource, it displays the previous value.

One thing that's still a todo here is to reset the values in case there is an existing value but the new video doesn't have origin resource. So entering trim mode for a video that sets the trim values and then entering trim mode for a video that doesn't have an origin resource will currently display the offsets of the previous video instead of "clean" offsets.

I'm thinking that perhaps it would make sense to reset the start and end offsets and the `originalResource` whenever we enter trim mode.

packages/wp-story-editor/src/api/media.js
```diff
@@ -69,6 +69,25 @@ export function getMedia(
...
+ * Get media by ID.
+ *
+ * @param {number} mediaId Media ID.
+ * @param {Object} media   Media object.
+ * @return {Promise} Media object promise.
+ */
+export function getMediaById(mediaId, media) {
+  const path = addQueryArgs(`${media}${mediaId}/`, {
+    context: 'edit',
+    _web_stories_envelope: true,
```
spacedmonkey: Why do you need envelope here?

miina: Don't need, thanks!

packages/story-editor/src/components/videoTrim/useVideoNode.js
```diff
@@ -102,6 +127,48 @@ function useVideoNode() {
...
   );
 
+  const fetchOriginalResource = useCallback(
+    ({ original, endTime, startTime }) => {
+      // If already fetching, cancel.
+      if (isFetchingResource) {
+        return;
+      }
+      setIsFetchingResource(true);
+      getMediaById(original)
```
spacedmonkey, line 138: Should you also preload the video file as well?

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -189,8 +189,11 @@ function useProcessMedia({
     ({ resource: oldResource, start, end }) => {
       const { src: url, mimeType, poster } = oldResource;
 
+      // If we already have an original ID, use that instead.
       const trimData = {
-        original: oldResource.id,
+        original: oldResource.trimData?.original
+          ? oldResource.trimData.original
+          : oldResource.id,
```
barklund, line 196: Wait, can the original have trim data as well? Would the trim data for this video point directly to that original and not point to an intermittent one? I think this can only be the case right now because we don't have this PR in effect.

In reality (once this work is merged), this (i.e. _"the original has an original"_) can never be the case.

packages/story-editor/src/elements/video/trim.js
```diff
@@ -89,17 +117,35 @@ function VideoTrim({ box, element }) {
...
+    const startTime = getMsFromHMS(start);
+    fetchOriginalResource({ original, endTime, startTime });
+    return (
+      <Wrapper>
+        <Spinner>
+          <CircularProgress />
+        </Spinner>
+      </Wrapper>
+    );
+  }
```
barklund, line 132: Calling a function with side-effects (`fetchOriginalResource` in particular) directly in the render function is bad karma (can have unforeseen consequences) and even throws a warning:

```
Warning: Cannot update a component (`VideoTrimProvider`) while rendering a different component (`VideoTrim`).
```

The loader return should probably just be based on `isFetchingResource` being true (or maybe a combination of `isFetchingResource || shouldFetchResource`).

The call to `fetchOriginalResource` should be moved inside a (properly guarded) `useEffect`.

But I'd still like to see this call moved to the provider. The provider knows when the trim mode is triggered and can already then determine if a(nother) video is to be loaded. This component actually doesn't need to know anything about original videos or not - it'll just display whatever video that the provider tells it to. The provider should abstract out whether it's the "real" video or an "original" video. The same goes for the video trimmer bar.

barklund (changes requested):
It works well, but I'd like to see it abstracted better.

I'd prefer that all mentions of original video versus actual video only takes place in the provider and associated hooks. The underlying components including the video display and the trimmer bar shouldn't care at all whether this is the real video or an original - they'll just display whatever the provider tells them to display.

Let me know if you want me to detail it further or maybe even sketch out a solution. Or just take over.

packages/story-editor/src/app/media/utils/useProcessMedia.js
miina: No, the very original can't have trim data but the `resource` is modified as needed when calling `trimExistingVideo` so that if an `original` exists for the video that's being trimmed, the very original ID (`oldResource.trimData.original`) would be available. 

To make things more clear, perhaps we should change `trimExistingVideo` instead to accept the needed value as a separate param instead.

PR #9211, opened 2021-09-29
Taxonomy: Add "Most Used" Tags section

### LINKED ISSUE #9035: Taxonomy: Add "Most Used" Tags section
## Feature Description

**Not in scope from v1 of taxonomy** 

Gutenberg has a word cloud/list of the most used tags listed in posts. To mirror what users expect we should implement this in the editor. 

Designs: https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor-Sprint-Feature-Level-Designs_?node-id=9721%3A383439 

Screen Shot 2021-09-14 at 12 41 15 PM
Screen Shot 2021-09-14 at 12 41 11 PM

## Additional Context

Clicking on the blue text will add that string as a tag for the Story. 
List should be keyboard navigable.

maxyinger:
## Context
Part of Freeform taxonomy enhancements

## Summary
Adds "Most Used" section to freeform taxonomy input.
Screen Shot 2021-09-29 at 4 10 20 PM

## Relevant Technical Choices
- updated `addSearchResultsToCache` to be a little more flexible with regards to what args you can pass it
- Freeform selector fetches all terms for taxonomy on mount

## To-do
- [x] add a lil karma test.

## User-facing changes
NA

## Testing Instructions
- make sure taxonomies are enabled
- go to editor
- go to document
- play with the word cloud and history and see that it adds new terms to the freeform input correctly

(Make sure you have stories with associated tags that are published or else there's no Most Used terms and this component wont show up)

1.

## Checklist

---

Fixes #9035

### DIFF
--- a/packages/story-editor/src/app/taxonomy/taxonomyProvider.js
+++ b/packages/story-editor/src/app/taxonomy/taxonomyProvider.js
@@ -168,8 +168,8 @@ function TaxonomyProvider(props) {
       };
       setTermCache((cache) => mergeNestedDictionaries(cache, termResults));
 
-      if (addNameToSelection) {
-        const selectedTermSlug = cleanForSlug(name);
+      if (addNameToSelection && args.search) {
+        const selectedTermSlug = cleanForSlug(args.search);
         const selectedTerm = response.find(
           (term) => term.slug === selectedTermSlug
         );
@@ -225,7 +225,11 @@ function TaxonomyProvider(props) {
         // We could pull down only the exact term, but
         // we're modeling after Gutenberg.
         if (e.code === 'term_exists') {
-          addSearchResultsToCache(taxonomy, { name: termName }, addToSelection);
+          addSearchResultsToCache(
+            taxonomy,
+            { search: termName },
+            addToSelection
+          );
         }
       }
     },
@@ -258,6 +262,7 @@ function TaxonomyProvider(props) {
         createTerm,
         addSearchResultsToCache,
         setTerms,
+        addTermToSelection,
       },
     }),
     [
@@ -267,6 +272,7 @@ function TaxonomyProvider(props) {
       createTerm,
       addSearchResultsToCache,
       setTerms,
+      addTermToSelection,
     ]
   );
 
--- a/packages/story-editor/src/app/taxonomy/taxonomyProvider.js
+++ b/packages/story-editor/src/app/taxonomy/taxonomyProvider.js
@@ -149,7 +149,7 @@ function TaxonomyProvider(props) {
       let response = [];
       const termsEndpoint = taxonomy['_links']?.['wp:items']?.[0]?.href;
       if (!termsEndpoint) {
-        throw new Error("no href present in taxonomy['_links']['wp:items']");
+        return [];
       }
       try {
         response = await getTaxonomyTerm(termsEndpoint, args);
--- a/packages/story-editor/src/app/taxonomy/taxonomyProvider.js
+++ b/packages/story-editor/src/app/taxonomy/taxonomyProvider.js
@@ -145,32 +145,21 @@ function TaxonomyProvider(props) {
   );
 
   const addSearchResultsToCache = useCallback(
-    async (
-      taxonomy,
-      {
-        name,
-        // This is the per_page value Gutenberg is using
-        perPage = 20,
-      },
-      addNameToSelection = false
-    ) => {
+    async (taxonomy, args, addNameToSelection = false) => {
       let response = [];
       const termsEndpoint = taxonomy['_links']?.['wp:items']?.[0]?.href;
       if (!termsEndpoint) {
-        return;
+        throw new Error("no href present in taxonomy['_links']['wp:items']");
       }
       try {
-        response = await getTaxonomyTerm(termsEndpoint, {
-          search: name,
-          per_page: perPage,
-        });
+        response = await getTaxonomyTerm(termsEndpoint, args);
       } catch (e) {
         // Do we wanna do anything here?
       }
 
       // Avoid update if we're not actually adding any terms here
       if (response.length < 1) {
-        return;
+        return response;
       }
 
       // Format results to fit in our { [taxonomy]: { [slug]: term } } map
@@ -189,6 +178,8 @@ function TaxonomyProvider(props) {
           addTermToSelection(taxonomy, selectedTerm);
         }
       }
+
+      return response;
     },
     [getTaxonomyTerm, addTermToSelection]
   );
@@ -249,7 +240,7 @@ function TaxonomyProvider(props) {
         (taxonomy) => taxonomy.hierarchical
       );
       hierarchicalTaxonomies.forEach((taxonomy) =>
-        addSearchResultsToCache(taxonomy, { perPage: -1 })
+        addSearchResultsToCache(taxonomy, { per_page: -1 })
       );
 
       setShouldRefetchCategories(false);
--- a/packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
+++ b/packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
@@ -24,6 +24,7 @@ import {
   useState,
   useEffect,
 } from '@web-stories-wp/react';
+import { __ } from '@web-stories-wp/i18n';
 import PropTypes from 'prop-types';
 /**
  * Internal dependencies
@@ -32,7 +33,7 @@ import Tags, { deepEquals } from '../../../form/tags';
 import cleanForSlug from '../../../../utils/cleanForSlug';
 import { useTaxonomy } from '../../../../app/taxonomy';
 import { useHistory } from '../../../../app';
-import { ContentHeading, TaxonomyPropType } from './shared';
+import { ContentHeading, TaxonomyPropType, WordCloud } from './shared';
 
 function FlatTermSelector({ taxonomy, canCreateTerms }) {
   const [mostUsed, setMostUsed] = useState([]);
@@ -42,16 +43,23 @@ function FlatTermSelector({ taxonomy, canCreateTerms }) {
     addSearchResultsToCache,
     terms = [],
     setTerms,
+    addTermToSelection,
   } = useTaxonomy(
     ({
       state: { termCache, terms },
-      actions: { createTerm, addSearchResultsToCache, setTerms },
+      actions: {
+        createTerm,
+        addSearchResultsToCache,
+        setTerms,
+        addTermToSelection,
+      },
     }) => ({
       termCache,
       createTerm,
       addSearchResultsToCache,
       terms,
       setTerms,
+      addTermToSelection,
     })
   );
 
@@ -97,7 +105,7 @@ function FlatTermSelector({ taxonomy, canCreateTerms }) {
       return;
     }
     addSearchResultsToCache(taxonomy, {
-      name: value,
+      search: value,
       // This is the per_page value Gutenberg is using
       per_page: 20,
     });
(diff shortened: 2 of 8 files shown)


packages/story-editor/src/app/taxonomy/taxonomyProvider.js
```diff
@@ -234,7 +225,11 @@ function TaxonomyProvider(props) {
         // We could pull down only the exact term, but
         // we're modeling after Gutenberg.
         if (e.code === 'term_exists') {
-          addSearchResultsToCache(taxonomy, { name: termName }, addToSelection);
+          addSearchResultsToCache(
```
BrittanyIRL, line 228: cool, this will work nicely with what I had mapped out here: https://github.com/google/web-stories-wp/pull/9212/files#diff-b0d9645393a6258bd68def9cad9560e65e01a48f7631bc87c6a1cb75ce344872 

I do think it's actually `search_items` not `search`

BrittanyIRL (commented):
I dig this but I can't get it to work for me. When was the last time you ran `composer install`? When I was setting up autocomplete I noticed that the new api is actually expecting `search_items`  in `includes/Taxonomy/Tag_Taxonomy.php` which was getting me what I need on my branch. I'm curious if maybe something out of sync here?

packages/story-editor/src/app/taxonomy/taxonomyProvider.js
swissspidy: `search` is correct

packages/story-editor/src/app/taxonomy/taxonomyProvider.js
```diff
@@ -145,32 +145,21 @@ function TaxonomyProvider(props) {
...
-        perPage = 20,
-      },
-      addNameToSelection = false
-    ) => {
+    async (taxonomy, args, addNameToSelection = false) => {
       let response = [];
       const termsEndpoint = taxonomy['_links']?.['wp:items']?.[0]?.href;
       if (!termsEndpoint) {
-        return;
+        throw new Error("no href present in taxonomy['_links']['wp:items']");
```
samwhale: 💭 Should this crash the app or just log and return?

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
```diff
@@ -110,6 +125,19 @@ function FlatTermSelector({ taxonomy, canCreateTerms }) {
...
 
+  useEffect(() => {
+    (async function () {
+      const results = await addSearchResultsToCache(taxonomy, {
+        context: 'edit',
+        orderby: 'count',
+        order: 'desc',
+        hide_empty: true,
+      });
+      // console.log('RESULT ', results);
```
samwhale: Could remove this?

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
```diff
@@ -110,6 +125,19 @@ function FlatTermSelector({ taxonomy, canCreateTerms }) {
     [taxonomy, termCache]
   );
 
+  useEffect(() => {
+    (async function () {
+      const results = await addSearchResultsToCache(taxonomy, {
```
samwhale, line 133: just for my own edumacation: what's the benefit of using an IIFE vs not?

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
```diff
@@ -129,6 +157,37 @@ function FlatTermSelector({ taxonomy, canCreateTerms }) {
...
+            <WordCloud.List>
+              {mostUsed.map((term, i) => (
+                <WordCloud.ListItem key={term.id}>
+                  <WordCloud.Word
+                    onClick={() => {
+                      if (terms[taxonomy.restBase]?.includes(term.id)) {
+                        return;
+                      }
+                      addTermToSelection(taxonomy, term);
+                    }}
```
samwhale, line 176: 💭 should this be memoized?

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
```diff
@@ -129,6 +157,37 @@ function FlatTermSelector({ taxonomy, canCreateTerms }) {
...
+                    {term.name}
+
+                    {i < mostUsed.length - 1 &&
+                      /* translators: delimiter used in a list */
+                      __(',', 'web-stories')}
+                  </WordCloud.Word>
+                  {
+                    /* Browser only respects the white space in the li, not the button */
+                    i < mostUsed.length - 1 && ' '
+                  }
```
samwhale, line 187: 💭 Would the il8n utils for lists be helpful here?
https://github.com/google/web-stories-wp/blob/main/packages/i18n/src/translateToInclusiveList.js
https://github.com/google/web-stories-wp/blob/main/packages/i18n/src/translateToExclusiveList.js

packages/story-editor/src/app/taxonomy/taxonomyProvider.js
maxyinger: good point

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
maxyinger: just to do async operations in the scope of an effect with async/await instead of traditional promises. I'm only doing it cus I thought it was preferred to use async/await vs promises stylistically amongst the teams.

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
maxyinger: I reached for those originally, but don't they include `and` or `or` in the list? the list here is just supposed to be a comma separated list where each item is a button.

Is there a way to use those functions that wouldn't be more trouble than it's worth to augment them for this use case?

packages/story-editor/src/app/taxonomy/taxonomyProvider.js
BrittanyIRL: great, i'll update it in my pr - it's weird that that was returning me terms when 'search' wasn't. i'll go figure that out this am.

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
maxyinger: Yea I was thinking about that yesterday. Short answer is I'm not sure. 

You're either creating a big object that you're recreating every time `terms` updates, or re-rendering these small components slightly more often than usual.

I think this portion makes the memoization possibly more work than not:
```
if (terms[taxonomy.restBase]?.includes(term.id)) {
    return;
}
```

And the only reason we have that in there is to be sensitive about creating empty history entries.

I guess these components are small (and there's only a max of 10 words in the cloud ever) so limiting re-renders is probably negligible and I would err on the side of legibility/complexity/less code.

What are your thoughts?

maxyinger:
> expecting `search_items` in `includes/Taxonomy/Tag_Taxonomy.php` which was getting me what I need on my branch. I'm curious if maybe something out of sync here?

Hmph. just composer installed and started and stopped the env. still not running into any issues on this branch. maybe I'll rebase from main, and try again and see if I run into anything.

BrittanyIRL:
> > expecting `search_items` in `includes/Taxonomy/Tag_Taxonomy.php` which was getting me what I need on my branch. I'm curious if maybe something out of sync here?
> 
> Hmph. just composer installed and started and stopped the env. still not running into any issues on this branch. maybe I'll rebase from main, and try again and see if I run into anything.

If anything it was out of sync on my branch. i just pushed changes back to just `search` and everything's fine? I will pull this one again today and see if tags start showing up as most used.

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
samwhale: It's a small component so I don't think I'm too worried about extra re-renders

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
samwhale: Ah then yeah they wouldn't make a lot of sense. Makes sense to just combine using a separator. We could always make a util for that too that doesn't use any conjunctions 🤔

maxyinger: yea I guess my only concern is that it's not a list of words, but instead a list of components each with different props. Only worry there is the abstraction being more complexity when we're only running into this use case here.

Do your utils above handle react nodes like this? or just join words and return a string?

samwhale: I think they just join words, so it would be more complex to render react nodes. Seems like a util is a problem for another day

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
BrittanyIRL, line 176: can you move this to use a callback?

packages/story-editor/src/components/panels/document/taxonomies/shared.js
```diff
@@ -42,3 +48,32 @@ export const ContentHeading = styled(Headline).attrs({
...
+export const WordCloud = {
+  Heading: styled(ContentHeading)`
+    margin-bottom: 4px;
+  `,
+  Wrapper: styled.div`
+    padding: 18px 0px;
+  `,
+  List: styled.ul`
+    all: unset;
+    /* Only adding this here to prevent cursor flash between text/pointer when hovering*/
```
BrittanyIRL, line 69: 👍

BrittanyIRL:
Well, turns out last night i just did a bad job reading your testing directions and failed to process that stories need to be _published_ for the 'most used' to register. Sorry about that. I still think that's a bit weird 🤔  but this works great given that condition. Just needs tests passing ✅ ✅ ✅

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
maxyinger: reference this earlier thread here:
https://github.com/google/web-stories-wp/pull/9211#discussion_r719463109

Let me know what you think after you read through it

samwhale (approved):

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
```diff
@@ -113,6 +128,19 @@ function FlatTermSelector({ taxonomy, canCreateTerms }) {
...
+    (async function () {
+      const results = await addSearchResultsToCache(taxonomy, {
+        context: 'edit',
+        orderby: 'count',
+        order: 'desc',
+        hide_empty: true,
+      });
+      setMostUsed(results);
+    })();
+    // setMostUsed([]);
```
samwhale: remove?

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
```diff
@@ -113,6 +128,19 @@ function FlatTermSelector({ taxonomy, canCreateTerms }) {
...
+  useEffect(() => {
+    (async function () {
+      const results = await addSearchResultsToCache(taxonomy, {
+        context: 'edit',
+        orderby: 'count',
+        order: 'desc',
+        hide_empty: true,
+      });
+      setMostUsed(results);
+    })();
```
samwhale: Was seeing these before but now not seeing any. Seems like my terms are coming back as an empty array. When is the `mosUsed` section populated?

I feel like it should populate on mount because the effect runs when the flat term selector is rendered and the `taxonomy` changes 🤔

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
maxyinger: dangit. I thought I removed this

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
maxyinger: Yea, I think this is basically running on mount of the individual component because 
`taxonomy, addSearchResultsToCache` essentially never update in the lifetime of this component.

The only reason I didn't hoist this state to be held up at the provider level is because it would make the state keeping more complex to hold it for every flat term selector vs holding it in the components themselves. This set up equates to an extra network call every time the FlatTermSelector mounts but I think that's negligible

BrittanyIRL (approved):
Nice test updates

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
samwhale: Figured it out with max! Need to publish at least one story with tags for this to show up.

samwhale (approved):

PR #9212, opened 2021-09-29
Taxonomy: Autocomplete Options for Flat Terms

### LINKED ISSUE #9034: Taxonomy: Predictive suggestions for tags

BrittanyIRL:
## Context

Adds a dropdown menu ala Gutenberg's of available terms that start with the value typed. 

## Relevant Technical Choices

- autocomplete suggestions stored as `flatSearchResults` based on `addSearchResultsToCache` - just the response isolated so it's only things that match the current value. Implementation taken from #9211 
- dropdown placement was ripped from the dropdown component (thanks @BrookeGraham ) - this might be a cool thing to extract eventually. 
- Using state to decide when the autocomplete menu's focused. So when arrow up/down (depending on direction menu opened) it gets set to focused, when focus returns to parent from menu it's set to false again and also set to false when a value is submitted. 
- Added `,` to the keys since that's what the key is for `Comma` - left in comma because i'm curious if this was for a browser edge case - haven't looked into it yet. 
- Update `SUBMIT_VALUE` action to take a value optionally so that options from dropdown menu can be treated the same as values added via typing. 

## To-do

- [ ] Fix the tests that are failing
- [x] Discuss with Max what's up with the offset reset and update accordingly.
- [x] Test RTL 
- [x] Test roles 
- [x] PR description 

## User-facing changes

(Behind feature flag) 
Screen Shot 2021-09-30 at 3 28 55 PM

## Testing Instructions

Go to the tags section after enabling the taxonomy experiment and add some tags. You should be able to still add new tags (if you're admin) and anytime there's tags that match what your typed value starts with and you've typed more than 2 characters you'll get a menu with those options. You can select options from there with your keyboard or your cursor. It should mimic Gutenberg's.

No

No

No

## Checklist

---

Fixes #9034

### DIFF
--- a/packages/e2e-tests/src/specs/editor/taxonomy.js
+++ b/packages/e2e-tests/src/specs/editor/taxonomy.js
@@ -72,13 +72,15 @@ async function addTag(name) {
   await page.type('input#web_story_tag-input', name);
   await page.keyboard.press('Enter');
 
-  // Todo find a way to validate these spans or the delete buttons
-  // await expect(page).toMatchElement(
-  //   'span[data-testid="flat-term-token"] span',
-  //   {
-  //     text: name,
-  //   }
-  // );
+  await page.focus('input#web_story_tag-input');
+
+  const tokens = await page.evaluate(() =>
+    Array.from(
+      document.querySelectorAll('[data-testid="flat-term-token"] > span'),
+      (element) => element.textContent
+    )
+  );
+  await expect(tokens).toContainValue(name);
 }
 
 describe('Taxonomy', () => {
@@ -89,12 +91,14 @@ describe('Taxonomy', () => {
     await createNewStory();
     await goToAndExpandTaxonomyPanel();
 
-    await addCategory('music genres');
-    await addCategory('rock', 'music genres');
+    await Promise.all([
+      await addCategory('music genres'),
+      await addCategory('rock', 'music genres'),
 
-    await addTag('adventure');
-    await addTag('sci-fi');
-    await addTag('comedy');
+      await addTag('adventure'),
+      await addTag('sci-fi'),
+      await addTag('comedy'),
+    ]);
 
     // No need to save/publish the story, as the new categories will have been
     // created in the background via the REST API already.
@@ -148,7 +152,25 @@ describe('Taxonomy', () => {
       // Find an existing tag and select it.
       await page.focus('input#web_story_tag-input');
       await page.type('input#web_story_tag-input', 'adven');
-      await expect(page).toClick('adventure');
+      await expect(page).toMatchElement(
+        'ul[data-testid="suggested_terms_list"]'
+      );
+      await expect(page).toMatchElement('li[role="option"]', {
+        text: 'adventure',
+      });
+      await expect(page).toClick('li[role="option"]', { text: 'adventure' });
+      await page.focus('input#web_story_tag-input');
+
+      const tokens = await page.evaluate(() =>
+        Array.from(
+          document.querySelectorAll('[data-testid="flat-term-token"] > span'),
+          (element) => element.textContent
+        )
+      );
+
+      await expect(tokens).toStrictEqual(
+        expect.arrayContaining(['noir', 'action', 'adventure'])
+      );
 
       await publishStory();
 
@@ -159,9 +181,17 @@ describe('Taxonomy', () => {
       await goToAndExpandTaxonomyPanel();
 
       // See that added tags persist.
-      await expect(page).toMatchElement('span[data-testid="flat-term-token"]', {
-        text: 'noir',
-      });
+      const tokens2 = await page.evaluate(() =>
+        Array.from(
+          document.querySelectorAll('[data-testid="flat-term-token"] > span'),
+          (element) => element.textContent
+        )
+      );
+
+      await expect(tokens2).toContainValue('adventure');
+      await expect(tokens2).toContainValue('noir');
+
+      await percySnapshot(page, 'Taxonomies - Tags - Admin');
     });
   });
 
@@ -201,8 +231,58 @@ describe('Taxonomy', () => {
       await percySnapshot(page, 'Taxonomies - Categories - Contributor');
     });
 
-    it.todo(
-      'should be able to add tags that already exist but not create new tags'
-    );
+    it('should be able to add tags that already exist but not create new tags', async () => {
+      await createNewStory();
+      await insertStoryTitle('Taxonomies - Tags - Contributor');
+
+      await goToAndExpandTaxonomyPanel();
+
+      // Add some new tags that won't stick on refresh
+      await addTag('rom-com');
+      await addTag('creature feature');
+
+      // Find an existing tag and select it.
+      await page.focus('input#web_story_tag-input');
+      await page.type('input#web_story_tag-input', 'adven');
+      await expect(page).toMatchElement(
+        'ul[data-testid="suggested_terms_list"]'
+      );
+      await expect(page).toMatchElement('li[role="option"]', {
+        text: 'adventure',
+      });
+      await expect(page).toClick('li[role="option"]', { text: 'adventure' });
+      await page.focus('input#web_story_tag-input');
+
+      const tokens = await page.evaluate(() =>
+        Array.from(
+          document.querySelectorAll('[data-testid="flat-term-token"] > span'),
+          (element) => element.textContent
+        )
+      );
+
+      await expect(tokens).toContainValue('adventure');
+
+      await publishStory();
+
+      // Refresh page to verify that the assignments persisted.
+      await page.reload();
+      await expect(page).toMatchElement('input[placeholder="Add title"]');
+
+      await goToAndExpandTaxonomyPanel();
+
+      // See that added tags persist.
+      const tokens2 = await page.evaluate(() =>
+        Array.from(
+          document.querySelectorAll('[data-testid="flat-term-token"] > span'),
+          (element) => element.textContent
+        )
+      );
+
+      await expect(tokens2).toContainValue('adventure');
+      await expect(tokens2).not.toContainValue('rom-com');
+      await expect(tokens2).not.toContainValue('creature feature');
+
+      await percySnapshot(page, 'Taxonomies - Tags - Contributor');
+    });
   });
 });
--- a/packages/e2e-tests/src/specs/editor/taxonomy.js
+++ b/packages/e2e-tests/src/specs/editor/taxonomy.js
@@ -231,7 +231,10 @@ describe('Taxonomy', () => {
       await percySnapshot(page, 'Taxonomies - Categories - Contributor');
     });
 
-    it('should be able to add tags that already exist but not create new tags', async () => {
+    // Disable reason: capabilities not met, right now contributor users are granted the ability to `create-web_story_tag` which makes this test fail.
+    // https://github.com/google/web-stories-wp/issues/9236
+    // eslint-disable-next-line jest/no-disabled-tests
+    it.skip('should be able to add tags that already exist but not create new tags', async () => {
       await createNewStory();
       await insertStoryTitle('Taxonomies - Tags - Contributor');
 
--- a/packages/e2e-tests/src/specs/editor/taxonomy.js
+++ b/packages/e2e-tests/src/specs/editor/taxonomy.js
@@ -78,12 +78,14 @@ async function addTag(name) {
   await page.focus('input#web_story_tag-input');
   await page.type('input#web_story_tag-input', name);
   await page.keyboard.press('Enter');
-  await expect(page).toMatchElement(
-    'span[data-testid="flat-term-token"] span',
-    {
-      text: name,
-    }
-  );
+
+  // Todo find a way to validate these spans or the delete buttons
+  // await expect(page).toMatchElement(
+  //   'span[data-testid="flat-term-token"] span',
+  //   {
+  //     text: name,
+  //   }
+  // );
 }
 
 describe('Taxonomy', () => {
@@ -164,12 +166,9 @@ describe('Taxonomy', () => {
       await goToAndExpandTaxonomyPanel();
 
       // See that added tags persist.
-      await expect(page).toMatchElement(
-        'span[data-testid="flat-term-token"] span',
-        {
-          text: 'noir',
-        }
-      );
+      await expect(page).toMatchElement('span[data-testid="flat-term-token"]', {
+        text: 'noir',
+      });
     });
   });
 
--- a/packages/e2e-tests/src/specs/editor/taxonomy.js
+++ b/packages/e2e-tests/src/specs/editor/taxonomy.js
@@ -74,16 +74,32 @@ async function addCategory(name, parent) {
   });
 }
 
+async function addTag(name) {
+  await page.focus('input#web_story_tag-input');
+  await page.type('input#web_story_tag-input', name);
+  await page.keyboard.press('Enter');
+  await expect(page).toMatchElement(
+    'span[data-testid="flat-term-token"] span',
+    {
+      text: name,
+    }
+  );
+}
+
 describe('Taxonomy', () => {
   withExperimentalFeatures(['enableTaxonomiesSupport']);
 
-  // Create some categories before running all tests so that they are available there.
+  // Create some categories and tags before running all tests so that they are available there.
   beforeAll(async () => {
     await createNewStory();
     await goToAndExpandTaxonomyPanel();
 
     await addCategory('music genres');
     await addCategory('rock', 'music genres');
+
+    await addTag('adventure');
+    await addTag('sci-fi');
+    await addTag('comedy');
 
     // No need to save/publish the story, as the new categories will have been
     // created in the background via the REST API already.
@@ -123,9 +139,39 @@ describe('Taxonomy', () => {
 
       await percySnapshot(page, 'Taxonomies - Categories - Admin');
     });
-  });
 
-  it.todo('should be able to add new tags and existing tags');
+    it('should be able to add new tags and existing tags', async () => {
+      await createNewStory();
+      await insertStoryTitle('Taxonomies - Tags - Admin');
+
+      await goToAndExpandTaxonomyPanel();
+
+      // Add some new tags.
+      await addTag('noir');
+      await addTag('action');
+
+      // Find an existing tag and select it.
+      await page.focus('input#web_story_tag-input');
+      await page.type('input#web_story_tag-input', 'adven');
+      await expect(page).toClick('adventure');
+
+      await publishStory();
+
+      // Refresh page to verify that the assignments persisted.
+      await page.reload();
+      await expect(page).toMatchElement('input[placeholder="Add title"]');
+
+      await goToAndExpandTaxonomyPanel();
+
+      // See that added tags persist.
+      await expect(page).toMatchElement(
+        'span[data-testid="flat-term-token"] span',
+        {
+          text: 'noir',
+        }
+      );
+    });
+  });
 
   describe('Contributor', () => {
     withUser('contributor', 'password');
--- a/packages/e2e-tests/src/specs/editor/taxonomy.js
+++ b/packages/e2e-tests/src/specs/editor/taxonomy.js
@@ -52,6 +52,7 @@ async function addCategory(name, parent) {
 
   if (parent) {
     await expect(page).toClick('button[aria-label="Parent Category"]');
+    await page.waitForSelector('li[role="option"]');
     await expect(page).toMatchElement('li[role="option"]', { text: parent });
 
     await expect(page).toClick('li[role="option"]', { text: parent });
@@ -68,22 +69,36 @@ async function addCategory(name, parent) {
 }
 
 async function addTag(name) {
+  // get number of input children before we add the token
+  // so we have a metric to see when the new token renders
+  const numChildren = await page.$eval(
+    '#web_story_tag-input',
+    (el) => el.parentNode.children.length
+  );
+
+  // Add the new token
   await page.focus('input#web_story_tag-input');
   await page.type('input#web_story_tag-input', name);
   await page.keyboard.press('Enter');
 
-  await page.focus('input#web_story_tag-input');
-
-  const tokens = await page.evaluate(() =>
-    Array.from(
-      document.querySelectorAll('[data-testid="flat-term-token"] > span'),
-      (element) => element.textContent
-    )
+  // wait for token to render
+  await page.waitForFunction(
+    (originalNumChildren) =>
+      document.getElementById('web_story_tag-input').parentNode.children
+        .length > originalNumChildren,
+    {},
+    numChildren
   );
-  await expect(tokens).toContainValue(name);
+
+  const tokenNames = await page.$$eval(
+    '[data-testid="flat-term-token"]',
+    (nodes) => Array.from(nodes, (node) => node.innerText)
+  );
+
+  await expect(tokenNames).toContainValue(name);
 }
 
-describe('Taxonomy', () => {
+describe('taxonomy', () => {
   withExperimentalFeatures(['enableTaxonomiesSupport']);
 
   // Create some categories and tags before running all tests so that they are available there.
@@ -152,6 +167,7 @@ describe('Taxonomy', () => {
       // Find an existing tag and select it.
       await page.focus('input#web_story_tag-input');
       await page.type('input#web_story_tag-input', 'adven');
+      await page.waitForSelector('ul[data-testid="suggested_terms_list"]');
       await expect(page).toMatchElement(
         'ul[data-testid="suggested_terms_list"]'
       );
@@ -163,8 +179,8 @@ describe('Taxonomy', () => {
 
       const tokens = await page.evaluate(() =>
         Array.from(
-          document.querySelectorAll('[data-testid="flat-term-token"] > span'),
-          (element) => element.textContent
+          document.querySelectorAll('[data-testid="flat-term-token"]'),
+          (element) => element.innerText
         )
       );
 
--- a/packages/e2e-tests/src/specs/editor/taxonomy.js
+++ b/packages/e2e-tests/src/specs/editor/taxonomy.js
@@ -49,26 +49,19 @@ async function goToAndExpandTaxonomyPanel() {
 
 async function addCategory(name, parent) {
   await expect(page).toClick('button', { text: 'Add New Category' });
-  await expect(page).toFill('input[name="New Category Name"]', name);
 
   if (parent) {
     await expect(page).toClick('button[aria-label="Parent Category"]');
     await expect(page).toMatchElement('li[role="option"]', { text: parent });
-    await expect(page).toClick('li[role="option"]', { text: parent });
-    await expect(page).toMatchElement(
-      'li[role="option"][aria-selected="true"]',
-      { text: parent }
-    );
 
+    await expect(page).toClick('li[role="option"]', { text: parent });
     await expect(page).toMatchElement('button[aria-label="Parent Category"]', {
       text: parent,
     });
-
-    // TODO: Add assertion here to verify the parent was chosen and the dropdown closed.
   }
 
+  await expect(page).toFill('input[name="New Category Name"]', name);
   await page.keyboard.press('Enter');
-
   await expect(page).toMatchElement('label', {
     text: name,
   });
--- a/packages/story-editor/src/app/taxonomy/taxonomyProvider.js
+++ b/packages/story-editor/src/app/taxonomy/taxonomyProvider.js
@@ -42,7 +42,6 @@ import {
 function TaxonomyProvider(props) {
   const [taxonomies, setTaxonomies] = useState([]);
   const [termCache, setTermCache] = useState({});
-  const [flatSearchResults, setFlatSearchResults] = useState([]);
   // Should grab categories on mount
   const [shouldRefetchCategories, setShouldRefetchCategories] = useState(true);
   const {
@@ -180,8 +179,6 @@ function TaxonomyProvider(props) {
         if (selectedTerm) {
           addTermToSelection(taxonomy, selectedTerm);
         }
-      } else if (args.search) {
-        setFlatSearchResults(response);
       }
 
       // eslint-disable-next-line consistent-return
@@ -260,7 +257,6 @@ function TaxonomyProvider(props) {
   const value = useMemo(
     () => ({
       state: {
-        flatSearchResults,
         taxonomies,
         termCache,
         terms: Array.isArray(terms) ? {} : terms,
@@ -268,20 +264,17 @@ function TaxonomyProvider(props) {
       actions: {
         createTerm,
         addSearchResultsToCache,
-        setFlatSearchResults,
         setTerms,
         addTermToSelection,
       },
     }),
     [
-      flatSearchResults,
       termCache,
       terms,
       taxonomies,
       createTerm,
       addSearchResultsToCache,
       addTermToSelection,
-      setFlatSearchResults,
       setTerms,
     ]
   );
--- a/packages/story-editor/src/app/taxonomy/taxonomyProvider.js
+++ b/packages/story-editor/src/app/taxonomy/taxonomyProvider.js
@@ -268,6 +268,7 @@ function TaxonomyProvider(props) {
       actions: {
         createTerm,
         addSearchResultsToCache,
+        setFlatSearchResults,
         setTerms,
         addTermToSelection,
       },
@@ -280,6 +281,7 @@ function TaxonomyProvider(props) {
       createTerm,
       addSearchResultsToCache,
       addTermToSelection,
+      setFlatSearchResults,
       setTerms,
     ]
   );
--- a/packages/story-editor/src/app/taxonomy/taxonomyProvider.js
+++ b/packages/story-editor/src/app/taxonomy/taxonomyProvider.js
@@ -161,11 +161,11 @@ function TaxonomyProvider(props) {
       }
       try {
         response = await getTaxonomyTerm(termsEndpoint, {
-          search: name,
+          search_items: name,
           per_page: perPage,
         });
       } catch (e) {
-        // Do we wanna do anything here?
+        // TODO #9033
       }
 
       // Avoid update if we're not actually adding any terms here
--- a/packages/story-editor/src/app/taxonomy/taxonomyProvider.js
+++ b/packages/story-editor/src/app/taxonomy/taxonomyProvider.js
@@ -161,7 +161,7 @@ function TaxonomyProvider(props) {
       }
       try {
         response = await getTaxonomyTerm(termsEndpoint, {
-          search_items: name,
+          search: name,
           per_page: perPage,
         });
       } catch (e) {
--- a/packages/story-editor/src/app/taxonomy/taxonomyProvider.js
+++ b/packages/story-editor/src/app/taxonomy/taxonomyProvider.js
@@ -42,6 +42,7 @@ import {
 function TaxonomyProvider(props) {
   const [taxonomies, setTaxonomies] = useState([]);
   const [termCache, setTermCache] = useState({});
+  const [flatSearchResults, setFlatSearchResults] = useState([]);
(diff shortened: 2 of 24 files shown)


packages/story-editor/src/app/taxonomy/taxonomyProvider.js
```diff
@@ -161,11 +161,11 @@ function TaxonomyProvider(props) {
       }
       try {
         response = await getTaxonomyTerm(termsEndpoint, {
-          search: name,
+          search_items: name,
```
BrittanyIRL: https://github.com/google/web-stories-wp/blob/main/includes/Taxonomy/Tag_Taxonomy.php#L61

packages/e2e-tests/src/specs/editor/taxonomy.js
```diff
@@ -123,9 +141,36 @@ describe('Taxonomy', () => {
...
+      // Refresh page to verify that the assignments persisted.
+      await page.reload();
+      await expect(page).toMatchElement('input[placeholder="Add title"]');
+
+      await goToAndExpandTaxonomyPanel();
+
+      // See that added tags persist.
+      await expect(page).toMatchElement('span[data-testid="flat-term-token"]', {
+        text: 'noir',
+      });
```
BrittanyIRL: todo tomorrow, wanted to get the rest of the PR up for review

packages/story-editor/src/app/taxonomy/taxonomyProvider.js
swissspidy: Attention. You mixed up two completely different things here.

search_items is just the name of the label for when you want to display „Search Tags“ in the UI. See how it‘s in the „labels“ array in the code you linked to.

It‘s not related to the REST API in any way.

search ought to be correct for REST API querying, but consult the WP REST API reference documentation or ask us to confirm.

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -144,16 +244,50 @@ function Input({
...
+                    listId={listId}
+                    hasMenuRole
+                    handleReturnToParent={handleReturnFocusToInput}
+                    isRTL={isRTL}
+                    options={[{ group: filteredAutocompleteSuggestions }]}
+                    onMenuItemClick={handleTagSelectedFromSuggestions}
+                    onDismissMenu={noop} // No need to dismiss, it's either open with options or hidden
+                    menuAriaLabel={sprintf(
+                      /* translators: %s: title of freeform terms */
+                      __('Available terms for %s', 'web-stories'),
```
swissspidy: Can we use the taxonomy's existing `items_list` label instead? It will be "Tags list" or "Categories list" (or "Colors list", or ...) depending on the taxonomy. Allows us to avoid the generic "term" term.

packages/story-editor/src/app/taxonomy/taxonomyProvider.js
BrittanyIRL: yeah, i realize that now. Sorry about that - i must have gotten some wires crossed and thought this was the path forward. thought it was a bit odd but i was getting results i wanted it seemed. It's all fixed now. Appreciate the call out.

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -56,24 +67,76 @@ const TextInput = styled(BaseInput).attrs({ type: 'text' })`
...
+      return [
+        ...accum,
+        {
+          label: suggestion,
+          value: suggestion,
+          id: uuidv4(),
+        },
+      ];
+    }, []);
+  }, [suggestedTerms, value]);
```
maxyinger: what's the need for filtering these at all? What made you go with this approach vs just passing in the results we get from the primary search query in `FlatTermSelector` ?

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -110,8 +181,12 @@ function Input({
...
             dispatch({ type: ACTIONS.REMOVE_TAG });
           }
-          if (['Comma', 'Enter', 'Tab'].includes(e.key)) {
-            dispatch({ type: ACTIONS.SUBMIT_VALUE });
+          if (['Comma', ',', 'Enter', 'Tab'].includes(e.key)) {
+            e.preventDefault();
+            e.stopPropagation();
+
+            await dispatch({ type: ACTIONS.SUBMIT_VALUE });
+            await inputRef?.current.focus();
```
maxyinger: `dispatch(...)` and `.focus()` methods return a promise?

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -124,16 +199,41 @@ function Input({
...
+    setSuggestedTermsFocus(false);
+  }, []);
+
+  const handleTagSelectedFromSuggestions = useCallback(
+    async (e, selectedValue) => {
+      e.preventDefault();
+      // It's important these run in order so that focus remains intact
+      await setSuggestedTermsFocus(false);
+      await dispatch({ type: ACTIONS.SUBMIT_VALUE, payload: selectedValue });
+      await inputRef?.current.focus();
```
maxyinger: I don't think any of these are async operations. we probably don't need to make this function async

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -124,16 +199,41 @@ function Input({
...
+
+  const handleSubmitOnFocusOut = useCallback(() => {
+    if (filteredSuggestedTerms.length <= 0 && value.length > 0) {
+      setSuggestedTermsFocus(false);
+      dispatch({ type: ACTIONS.SUBMIT_VALUE, payload: value });
+    }
+  }, [filteredSuggestedTerms, value]);
+
+  useFocusOut(containerRef, handleSubmitOnFocusOut, [handleSubmitOnFocusOut]);
+
```
maxyinger: This doesn't seem like behavior exhibited in Guttenberg. Unless it was specifically requested, I wouldn't worry about adding the complexity of needing to submit on focus out:

Screen Shot 2021-09-30 at 12 41 19 PM

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -144,16 +244,45 @@ function Input({
...
+                    handleReturnToParent={handleReturnFocusToInput}
+                    isRTL={isRTL}
+                    options={[{ group: filteredSuggestedTerms }]}
+                    onMenuItemClick={handleTagSelectedFromSuggestions}
+                    onDismissMenu={noop} // No need to dismiss, it's either open with options or hidden
+                    menuAriaLabel={suggestedTermsLabel}
+                    isMenuFocused={suggestedTermsFocus}
+                    isPositionedOnTop={dynamicPlacement === PLACEMENT.TOP}
+                  />
+                </Popup>
```
maxyinger: Could we just render this in the `Border` element vs within the `Fragment` containing the `TextInput` ?

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
```diff
@@ -125,6 +133,8 @@ function FlatTermSelector({ taxonomy, canCreateTerms }) {
           onInputChange={handleFreeformInputChange}
           tagDisplayTransformer={termDisplayTransformer}
           tokens={tokens}
+          suggestedTerms={termCacheAsArray}
```
maxyinger: I think we should be pulling these off of the results we get back from the query called in `addSearchResultsToCache` .

The BE has a much more robust searching mechanism for tags, and we shouldn't be doing any of that work on the frontend

packages/story-editor/src/components/form/tags/reducer.js
```diff
@@ -137,10 +138,10 @@ function reducer(state, action) {
       };
     }
 
-    case ACTIONS.RESET_OFFSET: {
+    case ACTIONS.RESET_VALUE: {
```
maxyinger, line 149: I don't see you calling this anymore with the current setup.

The only bug I see resulting from this is that we do want to reset the offset whenever we blur the input and we're not submitting a value

maxyinger (changes requested):
There may be some bugs around focus when the suggestions aren't shown and you try to tab out of it while the input's focused.

Also noticed this:
Screen Shot 2021-09-30 at 1 01 29 PM

Screen Shot 2021-09-30 at 1 08 53 PM
>

I think we want to use the response from `addSearchResultsToCache()` and filter out any terms that are currently selected instead of searching though our `termCache` for terms that start with what's in the input

packages/story-editor/src/components/form/tags/input.js
BrittanyIRL: no good reason, i'll fix this 🙃

packages/story-editor/src/components/form/tags/input.js
BrittanyIRL: I'll retest this, because clearly i'm off this week, but this was to make sure the value was submitted and THEN the input refocused to restart the reducer state with 'onFocus' without a value present as input.

packages/story-editor/src/components/form/tags/input.js
BrittanyIRL: it just felt weird, seemed like an improvement to clean up on focus out.

packages/story-editor/src/components/form/tags/reducer.js
BrittanyIRL: so `RESET_OFFSET` isn't part of the `ACTIONS` outlined that are exported and available, so it wasn't ever actually triggering this case when it's "called" in the tag input.

packages/story-editor/src/components/form/tags/input.js
BrittanyIRL: 🤕 yeah yup, fixing all of these, went a little overboard doing too much at once.

packages/story-editor/src/components/form/tags/reducer.js
maxyinger: ooo. I think it was getting triggered, it was just mapping to `undefined`. Cus I see the functionality working on main, and not on this branch. Good catch tho

maxyinger: I think we still need that behavior in there regardless tho

packages/story-editor/src/components/form/tags/input.js
BrittanyIRL: so i remember why i did this - i was just trying to use the cache not add more state. following the same pattern now that you're establishing with the most used terms.

packages/story-editor/src/components/form/tags/input.js
maxyinger: I guess I would close the suggestion list, but I think submitting the value seems a bit weird. Maybe that's just me tho 🤷 . Maybe worth chatting with Amy about if we have differing expectations? I'm not opposed to the behavior you have here, just saying it's not what I'd expect off of first glance

packages/story-editor/src/components/form/tags/input.js
maxyinger: ah yea. tough race case. I don't think the `await` s there actually change any behavior, but I think the `dispatch({ type: ACTIONS.SUBMIT_VALUE })` _should_ always get called and update state before any dispatches called by `onFocus`

My rationale for that is that it will have fired off the `dispatch({ type: ACTIONS.SUBMIT_VALUE });` before the `onFocus` event gets put in the event queue, and react wont batch the `dispatch` calls unless they're in the same event handler callback. That's at least my best guess. so it shouldn't change anything removing those `await` s and it should still function how you're envisioning. but that being said, I'm not 100% sure

maxyinger:
Something I just noticed here, might be simpler to not use a popup and just render a component that causes a layout shift inside the `Border` element. Only do this if it makes your life easier, but just wanted to pitch it out as an idea that might minimize complexity here:

https://user-images.githubusercontent.com/35983235/135522539-ce670942-728e-48e2-8c38-c86ecbdd1f0b.mp4

packages/story-editor/src/components/form/tags/reducer.js
BrittanyIRL: weird that it was working in main since there's no diff here removing it? I'll restore it!

packages/story-editor/src/app/taxonomy/taxonomyProvider.js
```diff
@@ -145,50 +146,46 @@ function TaxonomyProvider(props) {
   );
 
   const addSearchResultsToCache = useCallback(
-    async (
```
BrittanyIRL: pulled this change in from https://github.com/google/web-stories-wp/pull/9211 to sync up on callback updates

packages/story-editor/src/components/form/tags/input.js
BrittanyIRL: nah that's a great point, i'll take it out and if after review we want to change the behavior of anything we can set tickets for it and do it intentionally.

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
BrittanyIRL: added in `flatSearchResults` to context and things make way more sense now.

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
```diff
@@ -110,6 +128,15 @@ function FlatTermSelector({ taxonomy, canCreateTerms }) {
     [taxonomy, termCache]
   );
 
+  const suggestedTerms = useMemo(
```
BrittanyIRL, line 131: added this to clarify and simplify what's expected for the list in `Input` - could pass in `flatSearchResults` as is if that's the preference.

BrittanyIRL:
> Something I just noticed here, might be simpler to not use a popup and just render a component that causes a layout shift inside the `Border` element. Only do this if it makes your life easier, but just wanted to pitch it out as an idea that might minimize complexity here:

yeah this is way easier 🤕  totally my bad. thank you for the many suggestions here. This should be cleaned up - I need to address tests still - didn't quite get to those yet.  Tagging for rereview just to make sure this is tracking.

packages/story-editor/src/app/taxonomy/taxonomyProvider.js
```diff
@@ -145,50 +146,46 @@ function TaxonomyProvider(props) {
...
+        const selectedTermSlug = cleanForSlug(args.search);
         const selectedTerm = response.find(
           (term) => term.slug === selectedTermSlug
         );
 
         if (selectedTerm) {
           addTermToSelection(taxonomy, selectedTerm);
         }
+      } else if (args.search) {
+        setFlatSearchResults(response);
```
maxyinger: So I think this approach won't be taxonomy agnostic. I would just grab the results from this function in the `FlatTermSelector` and store them in state there, then they will be taxonomy agnostic. Basically just do all this state keeping in that component vs in the provider here. shouldn't be too bad of an update since it shouldn't really change how you're using `flatSearchResults` will just be stored in a different place

packages/story-editor/src/app/taxonomy/taxonomyProvider.js
```diff
@@ -145,50 +146,46 @@ function TaxonomyProvider(props) {
...
       }
+
+      // eslint-disable-next-line consistent-return
+      return response;
```
maxyinger: I wouldn't disable this, I still need to update my PR, but I'm thinking we can just do something like this on the areas where we're not returning anything:
```
console.error('some helpful issue message');
return [];
```

I guess I'll be adding this in my PR too tho, so I guess disabling it temporarily is fine so we're not doing the same work 🤷

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -55,9 +56,23 @@ const TextInput = styled(BaseInput).attrs({ type: 'text' })`
...
+  list-style-type: none;
+  width: 100%;
+  margin-top: 4px;
+  li {
+    padding: 4px;
+    width: 100%;
+    ${themeHelpers.focusableOutlineCSS()}
+    border-radius: ${({ theme }) => theme.borders.radius.small};
+  }
+`;
```
maxyinger: Nice. I'm glad this simplified things. I think any time we can use less popups in this app, it's a win since there's so many in every interaction rn. 

My only design feedback would be to add a `border-top` with the same border as the `Border` component just so this component doesn't feel like it's floating in there

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -161,14 +220,39 @@ function Input({
           )
         )
       }
+      {suggestedTerms.length > 0 && value.length >= 3 && (
```
maxyinger: you should be able to remove the `value.length >= 3` since that's an implementation detail covered in when we fetch results vs how we display results

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
```diff
@@ -87,10 +96,19 @@ function FlatTermSelector({ taxonomy, canCreateTerms }) {
...
+
     if (value.length < 3) {
       return;
     }
-    addSearchResultsToCache(taxonomy, { name: value });
+    addSearchResultsToCache(taxonomy, {
+      search: value,
+      // This is the per_page value Gutenberg is using
+      per_page: 20,
+    });
```
maxyinger: after you move the `searchResults` to be encapsulated in this component vs the provider, you should be able to just do this here:
```
const results = await  addSearchResultsToCache(taxonomy, {
      search: value,
      // This is the per_page value Gutenberg is using
      per_page: 20,
});
setSearchResults(results);
```

maxyinger (commented):
Everything is looking great so far, I would just move the `flatTermSearchResults` state to be stored in the `FlatTermSelector` component vs the `TaxonomyProvider` , then it should inherently be taxonomy agnostic

packages/story-editor/src/components/form/tags/input.js
BrittanyIRL: Def, good note. Should be all fixed up.

Screen Shot 2021-10-01 at 8 57 58 AM
Screen Shot 2021-10-01 at 8 57 36 AM

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -116,8 +149,10 @@ function Input({
               onUndo(e);
             }
           }
-
-          if (['Comma', 'Enter', 'Tab'].includes(e.key)) {
+          if (e.key === 'ArrowDown' && suggestedTerms.length > 0) {
+            menuRef?.current?.firstChild?.focus();
+          }
+          if (['Comma', ',', 'Enter', 'Tab'].includes(e.key)) {
```
maxyinger: So having this `Tab` here messes up the flow a little bit because it automatically submits whatever's in the input field when you try and tab out. Feels a little off if say you have `some tag` typed in, then you press `Shift + Tab` and then the tag is created and you're focused on its closed button.

Could we just remove `Tab` out of the keys here? what scenario did you add that functionality in here for?

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -55,9 +66,26 @@ const TextInput = styled(BaseInput).attrs({ type: 'text' })`
...
 `;
 
+const SuggestionList = styled(List)`
+  border-top: ${({ theme }) =>
+    `1px solid ${theme.colors.border.defaultNormal}`};
+  display: block;
+  list-style-type: none;
+  width: 100%;
+  padding: 6px 4px 4px;
+  margin-top: 6px;
```
maxyinger: can we add a max height in here with overflow scrollable to prevent this scenario?
Screen Shot 2021-10-01 at 2 34 48 PM

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -55,9 +66,26 @@ const TextInput = styled(BaseInput).attrs({ type: 'text' })`
...
+  li {
+    padding: 4px;
+    width: 100%;
+    ${themeHelpers.focusableOutlineCSS()}
+    border-radius: ${({ theme }) => theme.borders.radius.small};
+  }
```
maxyinger: add:
```
cursor: pointer;
```

Not sure if we want to add other hover styles, but that's for another day

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -135,48 +171,110 @@ function Input({
...
+          role="listbox"
+          ref={menuRef}
+          data-testid="suggested_terms_list"
+        >
+          {suggestedTerms.map(({ name, id }, index) => (
+            <li
+              key={id}
+              aria-selected={value === name}
+              role="option"
+              tabIndex={0}
```
maxyinger, line 275: not sure if it makes sense to make these `tabIndex={-1}` in gutenberg you can only access them through the up/down arrow from being focused on the input

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -135,48 +171,110 @@ function Input({
...
+              onClick={(e) => handleTagSelectedFromSuggestions(e, name)}
+              onKeyDown={(e) => handleSuggestionKeyDown(e, index, name)}
+            >
+              {name}
+            </li>
+          ))}
+        </SuggestionList>
+      )}
```
samwhale: 💭 For accessibility: when this appears I think it would be helpful to broadcast a message about how many 'results' there are. We can use the `useLiveRegion` hook for this

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -135,48 +171,110 @@ function Input({
...
+      [onUndo, suggestedTerms]
     );
 
   const renderedTags = tagBuffer || tags;
+
+  const handleTagSelectedFromSuggestions = useCallback((e, selectedValue) => {
+    e.preventDefault();
+    dispatch({ type: ACTIONS.SUBMIT_VALUE, payload: selectedValue });
+    inputRef?.current.focus();
+  }, []);
```
maxyinger: Might be nice to store some state for if the suggestions list is open so it closes immediately on selection. something like:
```js
  const [isSuggestionsOpen, setIsSuggestionsOpen] = useState(false);
...
  const handleTagSelectedFromSuggestions = useCallback((e, selectedValue) => {
    e.preventDefault();
    dispatch({ type: ACTIONS.SUBMIT_VALUE, payload: selectedValue });
    inputRef?.current.focus();
    setIsSuggestionsOpen(false)
  }, []);
...
  const suggestionLength = suggestedTerms.length;
  useEffect(() => {
    setIsSuggestionsOpen(suggestionLength > 0);
  }, [suggestionLength]);
```

maxyinger (commented):
I'm sorry this is the PR from hell. I got some more feedback. If you just want to create any tickets for the requests I've made, it's fine to just make tickets and merge this. Let me know if that's the route you wanna go and I can slap an approval on here

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -135,48 +171,110 @@ function Input({
...
+                key={INPUT_KEY}
+                value={value}
+                onKeyDown={handleKeyDown}
+                onChange={handleChange}
+                onFocus={handleFocus}
+                onBlur={handleBlur}
+                size="4"
+                ref={inputRef}
+                autoComplete="off"
+              />
```
samwhale: 💭 I think we should change the [`autocomplete`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) property to "on" when the terms are being auto-completed. This should help with accessibility

💭 We should also add an `aria-expanded` and an `aria-describedby` to associate this to the autocomplete 

```suggestion
              <TextInput
                {...props}
                key={INPUT_KEY}
                value={value}
                onKeyDown={handleKeyDown}
                onChange={handleChange}
                onFocus={handleFocus}
                onBlur={handleBlur}
                size="4"
                ref={inputRef}
                autoComplete={value.length > 3 ? "on" : "off"}
                aria-describedby={/* id of suggestion list */}
              />
```

packages/story-editor/src/components/form/tags/input.js
samwhale: 💭 I think we should remove `Tab` from being used for submission. Currently it will submit the value and then focus an option in the autocomplete list.



Guttenberg doesn't submit on `Tab` and that way we can avoid this awkward jump:

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
```diff
@@ -89,11 +91,21 @@ function FlatTermSelector({ taxonomy, canCreateTerms }) {
...
       return;
     }
-    addSearchResultsToCache(taxonomy, { name: value });
+    const results = await addSearchResultsToCache(taxonomy, {
+      search: value,
+      // This is the per_page value Gutenberg is using
+      per_page: 20,
+    });
+    setSearchResults(results);
   }, 1000);
```
samwhale: 🤔 Do we want to wait a whole second here? Feels like there's a lot of lag. maybe 300 ms would feel a little more natural and still not shoot a request off each keystroke.

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -126,6 +161,7 @@ function Input({
         },
         removeTag: (tag) => () => {
           dispatch({ type: ACTIONS.REMOVE_TAG, payload: tag });
+          inputRef?.current.focus();
```
samwhale, line 180: 🤔 may be able to just use `setIsInputFocused` here instead of adding a new ref? Same with a few other places.

If not then I think we should remove that extra state and only focus using the ref.

packages/story-editor/src/components/form/tags/input.js
BrittanyIRL: easy! think this was to go with the focus out and save that we already removed.

samwhale: Oh just commented about this - yeah guttenberg doesn't use `Tab` for submission either so I think we would be fine to remove this.

packages/story-editor/src/components/form/tags/input.js
samwhale: Would that still let screen readers read the options?

packages/story-editor/src/components/form/tags/input.js
samwhale: I like this idea - there's some time between selection and closing due to the debounce and the network request

packages/story-editor/src/components/form/tags/input.js
BrittanyIRL: Screen Shot 2021-10-01 at 1 25 03 PM

packages/story-editor/src/components/form/tags/input.js
BrittanyIRL: i'll plop it in, easy enough!

packages/story-editor/src/components/form/tags/input.js
BrittanyIRL: Let's table this for its own ticket? I feel like our arrow nav is a big vague in the app and it'd be great to unify it.

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
BrittanyIRL: does feel like a lot, this was existing so i had just left it. i'll reduce it to 300

packages/story-editor/src/components/form/tags/input.js
BrittanyIRL: hrmmm this gets a little weird because the canvas takes focus away without forcing the input back into focus on these events. I was seeing the ref and the state separately since the state controls what the input looks like with the border.

packages/story-editor/src/components/form/tags/input.js
BrittanyIRL: nice catch, updated it to sync with our new bff the gutenberg taxonomy component 
Screen Shot 2021-10-01 at 1 51 28 PM

packages/story-editor/src/components/form/tags/input.js
BrittanyIRL: let's hold on adding a live region here? It feels like it might be disruptive?

samwhale: hmm I feel like it's okay if it's disruptive, but let's at least move this to another ticket 😄 It may not be needed with the `autocomplete="on"` 🤔

packages/story-editor/src/components/form/tags/input.js
samwhale: maybe we remove the other state then and just use the ref?

BrittanyIRL:
> I'm sorry this is the PR from hell. I got some more feedback. If you just want to create any tickets for the requests I've made, it's fine to just make tickets and merge this. Let me know if that's the route you wanna go and I can slap an approval on here

Totally fine! I think I got all the changes in here. Only things I think would be better left isolated are: figuring out keyup/down for menus w/ focus maintaining (thinking about the font picker dropdown - this list works a little more like that one than gutenberg), holding on an aria live region, and keeping both inputRef and input focus state to maintain style state and force focus to avoid jumps to the canvas.

packages/story-editor/src/components/form/tags/input.js
maxyinger: Eh honestly no need to create a ticket imo. I'm fine with it as is

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
maxyinger: yea I just put it in there arbitrarily.  reducing to 300 works for me

maxyinger (approved):
Lookin Good :shipit:

packages/story-editor/src/components/form/tags/input.js
BrittanyIRL: then the blue border would disappear 😕

packages/e2e-tests/src/specs/editor/taxonomy.js
```diff
@@ -49,41 +49,56 @@ async function goToAndExpandTaxonomyPanel() {
...
     await expect(page).toMatchElement('li[role="option"]', { text: parent });
-    await expect(page).toClick('li[role="option"]', { text: parent });
-    await expect(page).toMatchElement(
-      'li[role="option"][aria-selected="true"]',
-      { text: parent }
-    );
 
+    await expect(page).toClick('li[role="option"]', { text: parent });
     await expect(page).toMatchElement('button[aria-label="Parent Category"]', {
       text: parent,
```
BrittanyIRL: So this was having trouble determining that the dropdown was closed and submit button available, what I did was move it so that if it's a nested category do the dropdown category first and then add your input value and hit enter - eliminates the need to get puppeteer to see the button.

packages/e2e-tests/src/specs/editor/taxonomy.js
```diff
@@ -49,41 +49,56 @@ async function goToAndExpandTaxonomyPanel() {
...
+async function addTag(name) {
+  await page.focus('input#web_story_tag-input');
+  await page.type('input#web_story_tag-input', name);
+  await page.keyboard.press('Enter');
+
+  await page.focus('input#web_story_tag-input');
+
+  const tokens = await page.evaluate(() =>
+    Array.from(
+      document.querySelectorAll('[data-testid="flat-term-token"] > span'),
```
BrittanyIRL: Since the tags are in spans there wasn't really a great way to zero in on them besides this? Down for alternatives, but this lets us check that the term asked for has made it to the list.

packages/story-editor/src/components/form/tags/input.js
samwhale: oh that's right. okay makes sense! Maybe we just add a comment to make sure people know why we're using both in the future?

BrittanyIRL: sure!

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -135,48 +187,112 @@ function Input({
...
+                {...props}
+                key={INPUT_KEY}
+                value={value}
+                onKeyDown={handleKeyDown}
+                onChange={handleChange}
+                onFocus={handleFocus}
+                onBlur={handleBlur}
+                size="4"
+                ref={inputRef}
+                autoComplete={isSuggestionsOpen ? 'off' : 'on'}
```
samwhale, line 248: I think this is backwards
```suggestion
                autoComplete={isSuggestionsOpen ? 'on' : 'off'}
```
backwards

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -135,48 +187,112 @@ function Input({
...
+                aria-autocomplete="list"
```
samwhale, line 250: 🤔 Should this be `"none"` until the autocomplete is showing?

packages/story-editor/src/components/form/tags/input.js
samwhale, line 248: 💭 After playing around with this I think we also need the `aria-expanded` flag to get the screen reader to read it as open/close
```suggestion
                autoComplete={isSuggestionsOpen ? 'off' : 'on'}
                aria-expanded={isSuggestionsOpen}
```

samwhale (approved):
Thanks for all of your hard work on this. Only a couple more things and I'm good to go 🚢

packages/story-editor/src/components/form/tags/input.js
BrittanyIRL: I think having it on when the list of suggestions isn't rendered is beneficial because then if they've previously entered something and the list hasn't rendered yet their browser autocomplete could pitch in. `off` in this usage just means that when the list of suggestions starts showing the user won't see both [browser autocomplete ](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete)and the suggested list of tags at once, chaos ensuing :)

packages/story-editor/src/components/form/tags/input.js
BrittanyIRL: this is just the browser's autocomplete attribute - it's what helps in forms when you're entering common information for the browser to help you by suggesting a value to input.  If reversed, this happens: 
Screen Shot 2021-10-03 at 2 06 48 PM

BrittanyIRL: you're right. So i'd done this and then I ran aXe and got yelled at for it. Dug into it briefly today since tomorrow I'm stuck at jury duty and it's because the input role needed to be updated to a combobox. Thanks for keeping this in mind. should be fixed!

samwhale: oh I see!

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -246,8 +246,10 @@ function Input({
                 size="4"
                 ref={inputRef}
                 autoComplete={isSuggestionsOpen ? 'off' : 'on'}
+                aria-expanded={isSuggestionsOpen}
```
samwhale, line 249: 😄

packages/story-editor/src/components/form/tags/input.js
maxyinger: what's the need to turn it on at all vs leaving it off the whole time?

BrittanyIRL: just thought it was nice that if you had some previous input you'd already put in that you wanted to make a tag that your browser'd pick up on it and offer it before your value's long enough for the dropdown to be visible. happy to just leave it off the whole time.

packages/e2e-tests/src/specs/editor/taxonomy.js
```diff
@@ -52,6 +52,7 @@ async function addCategory(name, parent) {
 
   if (parent) {
     await expect(page).toClick('button[aria-label="Parent Category"]');
+    await page.waitForSelector('li[role="option"]');
```
BrittanyIRL, line 55: 👍

PR #9213, opened 2021-09-30
Added focus management to trim component

### LINKED ISSUE #9180: Video Trimming: Add proper focus management in trim mode
## Bug Description

Dragging a handle (or moving it with the keyboard) and then hitting cancel will let the user focus the buttons underneath the slider

Also, Tabbing between the handles and the buttons seems to be reversed in RTL. The tabbing order goes from Cancel (on right) -> far left handle -> far right handle -> Trim (on left)

Also, While the trim bar is visible, I was able to tab to undo that is under the bar. I was not able to tab to the handles for the trim. 
image16

## Expected Behaviour

Both initial focus to first handle as well as focus trapping within the trim mode

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

barklund:
## Context

This adds proper focus management to trim component including RTL support, which actually doesn't require any work, as the direction of a playing tape is universally left-to-right regardless of reading direction.

## To-do

* [x] Focus trapping
* [x] Auto-focus cancel button on mount
* ~Display sliders and scrims correctly in RTL mode~
* ~Handle moving sliders correctly when using keyboard in RTL mode~
* ~Handle moving sliders correctly when using mouse in RTL mode~

## User-facing changes

LTR:
trim-focus-ltr

RTL:
trim-focus-rtl

## Testing Instructions

1. Add a video and enter trim mode
2. Observe how keyboard focus is immediately placed on the cancel button
3. Observe how tabbing keeps focus within the trim component
4. Observe how tab order is logical (cancel ➡️  start-handle ➡️  end-handle  ➡️  trim button ➡️  cancel) rather than physical (from one side of the screen to the other)
5. Observe how arrow keys move the handles correctly
6. Observe how the video play direction in the trim rail is always left-to-right as it follows the direction of a tape being played and not the passing of time ([as per Material Design guidelines](https://material.io/design/usability/bidirectionality.html#:~:text=Local%20cafe%20icon-,the%20passage%20of%20time,-Anything%20depicting%20the))
7. 6. Do the above in both RTL and LTR mode

## Checklist

---

Fixes #9180


packages/story-editor/src/components/videoTrim/trimmerComponents.js
```diff
@@ -88,11 +88,11 @@ export const ButtonWrapper = styled.div`
   justify-content: ${({ isStart }) => (isStart ? 'flex-end' : 'flex-start')};
 `;
 
-export function Scrim({ width, atStart = false }) {
-  const floatStyle = atStart ? { left: '-1px' } : { right: '-1px' };
+export function Scrim({ width, isLeftAligned = false }) {
```
barklund, line 91: I renamed this as part of changing the direction of tape play for RTL, but then reversed all that again. I kept the renaming as it was clearer though.

miina (approved):

PR #9220, opened 2021-09-30
Taxonomies: Check `rest-base` and `slug` for permissions

### LINKED ISSUE #8849: Taxonomy: Support custom set ups
## Feature Description

Some users will have custom taxonomies set up. These follow the same structure and can be handled the same way. We should look at supporting these in our taxonomy panel. 

## Additional Context

As work is done on the initial taxonomy set up for categories and tags I'm sure this will make more sense, the only things I'm initially hesitant about adding this to scope at the moment are UX related - does this expand the panel? Scroll? And associated titling. So let's set up the categories and tags first so we can iterate more informed.

samwhale:
## Context

Custom taxonomy support

## Summary

Check the `rest-base` as well as the `slug` on a taxonomy when checking if a user has the correct permission for that taxonomy.

## Relevant Technical Choices

n/a

## To-do

n/a

## User-facing changes

none

no

no

no

## Checklist

---

Fixes #8849

### DIFF
--- a/packages/story-editor/src/components/panels/document/taxonomies/taxonomies.js
+++ b/packages/story-editor/src/components/panels/document/taxonomies/taxonomies.js
@@ -42,9 +42,10 @@ function TaxonomiesPanel(props) {
 
   const availableTaxonomies = taxonomies.filter((taxonomy) => {
     const isVisible = taxonomy?.visibility?.show_ui;
-    const canAssignTerms =
-      Boolean(capabilities[`assign-${taxonomy?.slug}`]) ||
-      Boolean(capabilities[`assign-${taxonomy?.restBase}`]);
+    const canAssignTerms = Boolean(
+      capabilities[`assign-${taxonomy?.restBase}`] ||
+        capabilities[`assign-${taxonomy?.slug}`]
+    );
 
     return isVisible && canAssignTerms;
   });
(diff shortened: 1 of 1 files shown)


packages/story-editor/src/components/panels/document/taxonomies/taxonomies.js
```diff
@@ -42,7 +42,9 @@ function TaxonomiesPanel(props) {
 
   const availableTaxonomies = taxonomies.filter((taxonomy) => {
     const isVisible = taxonomy?.visibility?.show_ui;
-    const canAssignTerms = Boolean(capabilities[`assign-${taxonomy?.slug}`]);
+    const canAssignTerms =
+      Boolean(capabilities[`assign-${taxonomy?.slug}`]) ||
+      Boolean(capabilities[`assign-${taxonomy?.restBase}`]);
```
swissspidy: I'd check for the `restBase` variant first.

Also you can avoid some repetition by doing

```js
const canAssignTerms =
      Boolean(
        capabilities[`assign-${taxonomy?.slug}`] ||
        capabilities[`assign-${taxonomy?.restBase}`]
      );
```

or even

```js
const canAssignTerms = Boolean(capabilities[`assign-${taxonomy?.restBase || taxonomy?.slug}`])
```

as originally suggested

samwhale: Do we want to check each individually or just default to restBase first and then `slug`? so

```js
Boolean(capabilities[`create-${taxonomy.slug}`]) || Boolean(capabilities[`create-${taxonomy?.restBase}`])
```
vs

```js
Boolean(capabilities[`assign-${taxonomy?.restBase || taxonomy?.slug}`])
```

In my mind this is a little sturdier in case the permission matches the slug rather than the rest base but either way works for me 🤔

maxyinger (approved):

swissspidy (approved):

PR #9221, opened 2021-09-30
Archive: Redirect default archive page to custom one

### LINKED ISSUE #9100: Archive: Redirect default archive page to custom one
## Feature Description

When users choose a custom page as the Web Stories archive page, we probably want to add a redirect from `/web-stories/` to the custom archive page for UX reasons.
Probably also from `/web-stories/` to `/new/custom/path/feed/`

## Alternatives Considered

Just adding an FYI note to the settings screen that users should set up redirects themselves

spacedmonkey:
## Context

When users choose a custom page as the Web Stories archive page, we probably want to add a redirect from `/web-stories/` to the custom archive page for UX reasons.
Probably also from `/web-stories/feed/` to `/new/custom/path/feed/`

## Summary

Adds a redirect from the default archive page to the custom one if one is set up.

## Testing Instructions

1. Enable, archive page experiment. 
2. Go to Dashboard > Setting. 
3. Change Stories Archives to `custom`
4. Select a page from dropdown, ( you will need at least one page ). 
5. Go to `<site_url>/web-stories`. Browser should redirect to path of selected page. 
6. Go to a single stories url, see it not redirect. 
7. Go to Dashboard > Setting. 
8. Set archive back to default. 
9. Go to `<site_url>/web-stories`. Browser shouldn't redirect to path of selected page. 

## Checklist

---

Fixes #9100


includes/Story_Post_Type.php
```diff
@@ -315,6 +316,37 @@ public function change_default_title( $data ) {
...
+	public function redirect_post_type_archive_urls( $bypass, $query ) {
+		global $wp_rewrite;
+
+		if ( $bypass || ! is_string( $this->get_has_archive() ) || ( ! $wp_rewrite instanceof \WP_Rewrite || ! $wp_rewrite->using_permalinks() ) ) {
+			return $bypass;
+		}
+
+		// 'pagename' is for most permalink types, name is for when the %postname% is used as a top-level field.
+		if ( self::REWRITE_SLUG === $query->get( 'pagename' ) || self::REWRITE_SLUG === $query->get( 'name' ) ) {
+			$redirect_url = get_post_type_archive_link( self::POST_TYPE_SLUG );
```
swissspidy, line 371: Could there be an infinite loop if the page slug is `/web-stories/` and thus `get_post_type_archive_link( self::POST_TYPE_SLUG )` returns `https://example.com/web-stories/`?

spacedmonkey: I don't think so. The check ```! is_string( $this->get_has_archive() )``` means that check to see if archive is a page. If the page url is `/web-stories/`, then it will not 404, meaning this could will never run.

swissspidy:
While reviewing the code I noticed that most parts of the redirect function were not actually tested, i.e. there was no assertion that the redirect actually worked. I also realized there was a missing check for the experiment being enabled.

Pushed some quick updates to add such a check and improve code coverage.

spacedmonkey:
@swissspidy The tests are failing now. Let me take over this PR.

swissspidy:
Ah right, the `get_has_archive()` tests of course would need to be updated as well now.

tests/phpunit/integration/tests/Story_Post_Type.php
```diff
@@ -93,7 +96,12 @@ public static function wpSetUpBeforeClass( $factory ) {
 	public function set_up() {
 		parent::set_up();
 
-		$this->instance = $this->injector->make( \Google\Web_Stories\Story_Post_Type::class );
+		$experiments = $this->createMock( \Google\Web_Stories\Experiments::class );
+		$experiments->method( 'is_experiment_enabled' )
+					->willReturn( true );
```
swissspidy, line 101: Hmm not sure I like basically enabling the experiment by default with this and then only having `test_get_has_archive_disabled_experiments` to test the _disabled_ scenario 🤔 

Need to review a bit more.

spacedmonkey: There is only one place where we test for is the experiment is disabled and that is here. If we don't do this, then have to lots of repeated code to enabled it in the tests that need it. 

Once we have removed the feature flag only one test will have to be changed.

includes/Story_Post_Type.php
```diff
@@ -335,6 +346,44 @@ public function change_default_title( $data ) {
 		return $data;
 	}
 
+	/**
+	 * Handles redirects to the post type archive.
+	 *
+	 * @since 1.12.0
```
swissspidy:
```suggestion
	 * @since 1.13.0
```

swissspidy (approved):
Just needs QA instructions

spacedmonkey:
QA instructions added.

PR #9199, opened 2021-09-28
Taxonomies: Fix updating the `checked` attribute

BrookeGraham:
## Context

[See pr 9150](https://github.com/google/web-stories-wp/pull/9150#discussion_r716187510) 

When the checkbox input is checked/unchecked the attribute is not updated on the DOM's input element. 
## Relevant Technical Choices
Force a rerender by using checked status as part of the component's `key` value. 

## To-do

n/a

## Testing Instructions

- [x] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1. Check / uncheck a checkbox within the taxonomy hierarchal category selections. 
2. Look at the html element in dev tools to see if the `checked` attribute is reflected correctly. 

https://user-images.githubusercontent.com/1820266/135165316-fd731b7b-d771-4f1a-bd39-3fdbbf372b00.mov

### Does this PR have a security-related impact?
no

### Does this PR change what data or activity we track or use?
no

### Does this PR have a legal-related impact?
no

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [x] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [x] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Partially Fixes #9149

PR #9223, opened 2021-09-30
Taxonomies: Undo term submission for free form input

### LINKED ISSUE #9148: Taxonomy: Inconsistent "undo" of added freeform terms
In Gutenberg if you type in a new string and hit `cmd+z` it does nothing.
In Gutenberg if you type in a new string and then hit space or enter and it becomes a term and _then_ hit `cmd+z` it removes that term. 

In the editor if you type in a new string and hit `cmd+z` it will remove that last character you typed, makes sense. 
In the editor if you type in a new string and then hit space or enter and it becomes a term and _then_ hit `cmd+z` it does nothing. 

Found in #9147 which fixes taxonomy to work with history, does't feel like it's important to hold up - especially given that you can just as easily hit `shift+tab` and remove the term that way. It feels like an issue with the native input control and when to override it.

samwhale:
## Context

Taxonomies

## Summary

Pressing `cmd/ctrl + z` while focused on the freeform input will trigger the `undo` function from history.

## Relevant Technical Choices

After some investigation I saw that _all_ inputs in the editor eat up the `undo` event, not just this input. **If an input is focused and a user presses `cmd/ctrl + z` then the `undo` function in history won't be triggered. The global event listener never receives the event**

Most inputs it makes sense that `cmd/ctrl + z` would only undo/redo the characters pressed, but for the freeform input it would be expected that `cmd/ctrl + z` would undo the submission of a tag (or other taxonomy).

Passing in `undo` from the history provider fixes this issue.

## To-do

none

## Testing Instructions

1. Add a tag in the freeform input
2. Press `cmd/ctrl + z` to undo the submission
3. Should see the tag un-submitted
4. Pressing `cmd/ctrl + shift + z` won't re-submit it, but will re-type the characters that were in the input

no

no

no

## Checklist

---

Fixes #9148

### DIFF
--- a/packages/story-editor/src/components/form/hierarchical/index.js
+++ b/packages/story-editor/src/components/form/hierarchical/index.js
@@ -116,7 +116,7 @@ const Option = (option) => {
         <Label htmlFor={optionId}>{label}</Label>
       </CheckboxContainer>
       {options?.map((child) => (
-        <StepContainer key={`${child.id}-${child.checked}`}>
+        <StepContainer key={child.id}>
           <Option onChange={onChange} {...child} />
         </StepContainer>
       ))}
@@ -211,7 +211,7 @@ const HierarchicalInput = ({
               {filteredOptions.length ? (
                 filteredOptions.map((option) => (
                   <Option
-                    key={`${option.id}-${option.checked}`}
+                    key={option.id}
                     {...option}
                     onChange={handleCheckboxChange}
                   />
--- a/packages/story-editor/src/components/panels/document/taxonomies/test/taxonomies.js
+++ b/packages/story-editor/src/components/panels/document/taxonomies/test/taxonomies.js
@@ -92,7 +92,7 @@ describe('TaxonomiesPanel', () => {
           slug: 'web_story_category',
           restBase: 'web_story_category',
           name: 'Categories',
-          labels: {},
+          labels: { not_found: '' },
           hierarchical: false,
         },
       ],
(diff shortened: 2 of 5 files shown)


packages/story-editor/src/components/panels/document/taxonomies/test/taxonomies.js
```diff
@@ -92,7 +92,7 @@ describe('TaxonomiesPanel', () => {
           slug: 'web_story_category',
           restBase: 'web_story_category',
           name: 'Categories',
-          labels: {},
+          labels: { not_found: '' },
```
samwhale, line 95: Removes a prop type warning from these tests

maxyinger (approved):
dope. nice solution 🙌

packages/story-editor/src/components/form/hierarchical/index.js
```diff
@@ -116,7 +116,7 @@ const Option = (option) => {
         <Label htmlFor={optionId}>{label}</Label>
       </CheckboxContainer>
       {options?.map((child) => (
-        <StepContainer key={`${child.id}-${child.checked}`}>
```
samwhale, line 119: unfortunately this breaks a karma test: https://github.com/google/web-stories-wp/pull/9199. The checkboxes lose focused when pressed because they get re-created entirely.

BrittanyIRL (approved):
awesome! thanks for also fixing that karma test!

PR #9233, opened 2021-10-01
Bundling: Remove chunk hashes from file names

### LINKED ISSUE #9046: Bundling: Remove chunk hashes from file names
## Task Description

As mentioned at https://github.com/google/web-stories-wp/pull/8995#issuecomment-916812850, if we'd name our files `file-chunk-1.js?v=<hash>` instead of `file-chunk-1-<hash>.js`, that might save us some trouble with translation loading.

spacedmonkey:
## Summary

As mentioned at #8995 (comment), if we'd name our files file-chunk-1.js?v=<hash> instead of file-chunk-1-<hash>.js, that might save us some trouble with translation loading.

## Testing Instructions

1.

## Checklist

---

Fixes #9046


swissspidy:
Today we discussed potentially using the chunk's own hash for its version parameter instead of the entry point's hash.

While I looked into it and it seemed very straightforward to implement, it's actually not needed at all.

All chunks are dynamically imported by webpack and not enqueued through WordPress, so the version their registered with does not matter at all.

So what's needed there is actually a way to have the version as part of the URL for that dynamic import. Using `chunkFilename: '[name].js?v=[chunkhash]'` achieves that.

swissspidy (approved):

PR #9234, opened 2021-10-01
Fix dashboard invalid dom nesting warning

mohdsayed:
## Context

`div` is wrapped inside a `p` which is invalid

Screenshot 2021-10-01 at 8 36 19 PM

From [w3.org](https://www.w3.org/TR/html401/struct/text.html#h-9.3.1)
> The P element represents a paragraph. It cannot contain block-level elements 

and therefore it throws the following error in the console.

Screenshot 2021-10-01 at 8 35 40 PM

## Relevant Technical Choices

The `<Search>` component should not have been wrapped inside a `<TextInputHelperText>` component in the first place. But it would need some refactoring.

## Testing Instructions

1.

## Checklist

---

Fixes #


swissspidy:
Does the component need to be wrapped in `TextInputHelperText` in the first place? Can it just be wrapped in a simple `div`?

Not sure why I added it like this...

BrittanyIRL (approved):
lgtm, thanks for the fix!

PR #9251, opened 2021-10-04
Video Trimming: Enable by default

### LINKED ISSUE #8893: Enable video trimming for all users
## Feature Description

Once the feature is completed and approved, enable video trimming feature flag for all users.

Make sure the feature is properly covered by both unit and integration testing.

barklund:
## Context

As this feature has been both bug-bashed and thoroughly QA'ed through many iterations, this removes the feature flag in time for release.

A full test suite is still remaining, will be implemented in a separate PR.

## User-facing changes

_Video trim is enabled_

## Testing Instructions

1. Observe that the video trim experiment is enabled by default on the experiments page
2. Create a new story and add a video
3. Observe the video trim mode still exists and still works

## Checklist

---

See #8893


swissspidy:
Nit: We usually just enable the feature by default so we could easily roll it back if there are issues. We then remove the flag code entirely in the next release.

swissspidy:
> A full test suite is still remaining, will be implemented in a separate PR.

Is there a technical reason why no tests have been added so far? 🤔

miina:
Hm, I wonder if we should add a `trackEvent` for video trimming? Thoughts?

swissspidy:
> Hm, I wonder if we should add a `trackEvent` for video trimming? Thoughts?

Yeah absolutely

packages/story-editor/src/components/videoTrim/provider.js
```diff
@@ -64,6 +65,9 @@ function VideoTrimProvider({ children }) {
       start: formatMsToHMS(startOffset),
       end: formatMsToHMS(endOffset),
     });
+    trackEvent('video_trim', {
```
miina, line 68: @swissspidy I've added some track events, not sure if we have some specific naming we follow here, let me know if that looks ok to you.

includes/Experiments.php
```diff
@@ -353,6 +353,7 @@ public function get_experiments(): array {
 				'label'       => __( 'Video trimming', 'web-stories' ),
 				'description' => __( 'Enable video trimming', 'web-stories' ),
 				'group'       => 'editor',
+				'default'     => true,
```
miina, line 377: @barklund I changed this to use `true` for the flag right now instead of removing it fully per Pascal's comment. Perhaps we can remove it fully together with the tests PR.

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
```diff
@@ -47,12 +48,18 @@ function useVideoTrimMode() {
...
+      });
     } else {
       setEditingElementWithState(selectedElement.id, {
         isTrimMode: true,
         hasEditMenu: true,
         showOverflow: false,
       });
+      trackEvent('video_trim', {
+        name: 'entered_video_trim_mode',
+      });
```
swissspidy: Here I'd use something that works for both:

```js
trackEvent('video_trim_dialog_toggled', {
  status: isEditing ? 'closed' : 'open',
});
```

packages/story-editor/src/components/videoTrim/provider.js
swissspidy: Thanks.

We use extra params like `name` only to pass additional information for an event, e.g. `search_term` for a `search` event.

Here I'd do something more like this:

```js
trackEvent('video_trim', {
  original_length: resource.length,
  new_length: lengthInSeconds,
  start_offset: startOffset,
  end_offset: endOffset,
})
```

includes/Experiments.php
swissspidy: I still don't understand why we add tests **after** enabling the feature by default.

miina: I believe the feature was split up to separate small issues, none of which alone was fully functional and didn't make sense to be tested yet (except for the last issue before completing all tasks), so the idea was instead of bloating the last piece with adding all the tests, to split it out to add it separately.

I know that Morten has been working on adding the tests and it involves quite a lot of work. Perhaps if only part of the tests are ready, we could already add some with enabling the feature (if it's possible). WDYT, @barklund?

swissspidy: > I believe the feature was split up to separate small issues, none of which alone was fully functional and didn't make sense to be tested yet (except for the last issue before completing all tasks), so the idea was instead of bloating the last piece with adding all the tests, to split it out to add it separately.

That part I understand. But you still don't need this PR here in order to write tests. You don't need to enable this feature by default in order to add tests.

My worry is that we merge this, ship it (the release is planned for _today_!) and then add tests _tomorrow_ and find lots of bugs.

miina: > That part I understand. But you still don't need this PR here in order to write tests. You don't need to enable this feature by default in order to add tests.

Ah, okay, now I understand. The tests were originally intended to be part of the feature enabling PR as well, however, since they didn't get ready on time for release (today) but the feature has already been bug-bushed and QA-ed then the idea was to still get the feature into the release. This is not the ideal case so alternatively we could also just postpone enabling the feature to 1.14 to wait for the tests. In this case, enabling can be done together/after the tests are ready. Thoughts?

barklund: Yeah, I have a separate local PR with tests, but it's a bit messy currently. I could strip out a few of the tests currently ready though. I'm not finding any bugs in the tests (just a few minor semantic changes), so I thought the whole feature safe to ship and wanted the feature flag change merged before release today.

swissspidy: I am probably missing some context about this, so if there was a decision about this being OK, then it's fine with me.

I just don't understand the mindset of waiting until the last minute to add tests and potentially release a feature without any proper test coverage, hence my asking. Sounds like a bad habit and is not how we typically ship features.

The epic also still has plenty of open tickets too. Not just bugs, but also enhancements like #9265 for example. Was there a decision that these are out of scope for this release?

Again, probably missing some context here and just trying to better understand this.

swissspidy (approved):
code wise it's fine

includes/Experiments.php
barklund: The items missing from the epic have been ok'd for a delay, as they were not crucial features to be able to ship, but have been categorized as follow-ups post-release.

The missing/late tests are definitely a bad habit. It's what tends to happen when we use the feature flag, I think. We (I) tend to "move quickly and not test things" to quickly get started on the feature and to be able to allow many people to work on it at the same time and thus delay the testing efforts until "later". Which ends up biting us in the behind as the effort at that later point is suddenly growing huge and a bit overwhelming. Also as I've been working on these tests, we've been adding more minor features and bug fixes, that even changed the result of these tests, so they had to be re-made.

We definitely need a better process for testing things even while behind a feature flag.

barklund: For this particular feature though, the fact that it has been thoroughly bug bashed and QA'ed (the whole feature was QA'ed before bug bash and all the bug fixes and enhancements since bug bash have been QA'ed individually) gives me a good feeling for this one, so I felt it safe to include in the release.

swissspidy: Thanks for sharing your thoughts on this!

While I agree that it can be difficult to add tests when something is constantly changing, I don't think this is related to the feature flag at all.

In my experience other features have had reasonable amounts of test coverage in the past before being enabled by default.

Taking this particular feature as an example, it seems to me that **time** is the actual issue. If we had 1-2 more weeks, there would have been plenty of time to add tests and even fix some of the low hanging fruit, and overall have more assurance that the feature works the way it should. Instead, we're rushing things just so we can make it into this release.

Anyway, let's take this to Slack for further discussion with Tls.

alishanvr:
Hi,

Trimming is not enabled for non admin users. For example: Trimming is not enabled for the authors. May you please let me know how I can enable the trimming for all user roles on WordPress?

Or which capability should I have to set for the user to use the Trimming Videos option?

Thank You,


trimming-button-not-visible

swissspidy:
@alishanvr Hi there, we don't provide support via GitHub. Please open a new support topic [on the WordPress.org support forums](https://wordpress.org/support/plugin/web-stories/) for your question.

In short: you need to enable `Video optimization` in the settings and use the latest version of Chrome/Firefox/Safari. If that doesn't work, please provide your Site Health information on the forums so we can help you.

PR #9259, opened 2021-10-04
E2E Tests: Add test for pre-publish checklist for contributors

### LINKED ISSUE #9232: Checklist: Empty state not showing with contributor role
## Bug Description

The empty state of the checklist is not being shown to a user with the `contributor` role. When the checklist items are complete, the publish button tooltip still says the user should review the checklist.

## Expected Behaviour

Empty state of the checklist should be displayed when all checklist issues that the user can fix are fixed.

## Steps to Reproduce

1. Sign in as a contributor
2. Create a new story
3. Keep editing until all checklist items are completed
4. See that there is no empty state (checkmark icon with some nice text)

## Screenshots

image
image

## Additional Context

- Plugin Version: 1.12.0
- WordPress Version: 5.8.1
- Operating System: MacOSX Big Sur
- Browser: Chrome

samwhale:
## Context

Found a little while back where contributors were seeing the wrong UI: #9232. Couldn't reproduce, so we added a quick e2e test to make sure there is no regression.

## Summary

Adds a quick publishing flow test for a user with the `Contributor` role. 

## Relevant Technical Choices

n/a

## To-do

n/a

## User-facing changes

n/a

## Testing Instructions

Tests should pass

no

no

no

## Checklist

---

Fixes #9232

### DIFF
--- a/packages/e2e-tests/src/specs/editor/prepublishChecklist.js
+++ b/packages/e2e-tests/src/specs/editor/prepublishChecklist.js
@@ -73,7 +73,7 @@ describe('Pre-Publish Checklist', () => {
       });
 
       // verify that publish button is disabled
-      await page.$('button:disabled', { text: 'Publish' });
+      await expect(page).toMatchElement('button:disabled', { text: 'Publish' });
     });
   });
 });
--- a/packages/e2e-tests/src/specs/editor/prepublishChecklist.js
+++ b/packages/e2e-tests/src/specs/editor/prepublishChecklist.js
@@ -19,8 +19,10 @@
  */
 import {
   createNewStory,
+  insertStoryTitle,
   publishStory,
   triggerHighPriorityChecklistSection,
+  withUser,
 } from '@web-stories-wp/e2e-test-utils';
 import percySnapshot from '@percy/puppeteer';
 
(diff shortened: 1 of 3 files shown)


packages/e2e-tests/src/specs/editor/publishingFlow.js
```diff
@@ -25,8 +25,55 @@ import {
   withPlugin,
   publishStory,
   getEditedPostContent,
+  withUser,
 } from '@web-stories-wp/e2e-test-utils';
 
+describe('Contributor User', () => {
+  let stopRequestInterception;
+
+  beforeAll(async () => {
```
BrittanyIRL: 🤔  since this is to verify proper messaging for contributors, should it live in the `prepublishChecklist` file instead since this is to verify the publishing flow?

packages/e2e-tests/src/specs/editor/publishingFlow.js
```diff
@@ -25,8 +25,55 @@ import {
...
+        });
+      } else {
+        request.continue();
+      }
+    });
+  });
+
+  afterAll(async () => {
+    await page.setRequestInterception(false);
+    stopRequestInterception();
```
swissspidy: Can remove beforeAll & afterAll here. Not relevant for this test

packages/e2e-tests/src/specs/editor/prepublishChecklist.js
```diff
@@ -51,4 +53,29 @@ describe('Pre-Publish Checklist', () => {
...
+
+      await insertStoryTitle('Publishing Flow: Contributor');
+
+      const checklistToggle = await page.$(
+        'button[aria-owns="checklist_companion"]',
+        { text: 'Checklist' }
+      );
+
+      // verify no issues are present
+      await checklistToggle.click();
```
swissspidy: Could also just do it like the test above:

```js
await expect(page).toClick('button[aria-label^="Checklist: "]');
await expect(page).toMatchElement(
  '#pre-publish-checklist[data-isexpanded="true"]'
);
```

Any reason why this test would need to do it differently?

samwhale: Nope that's a good catch. Just moved this over, so I can update that here

packages/e2e-tests/src/specs/editor/prepublishChecklist.js
```diff
@@ -51,4 +53,27 @@ describe('Pre-Publish Checklist', () => {
...
+        '#pre-publish-checklist[data-isexpanded="true"]'
+      );
+
+      // verify no issues are present
+      await expect(page).toMatchElement('p', {
+        text: 'You are all set for now. Return to this checklist as you build your Web Story for tips on how to improve it.',
+      });
+
+      // verify that publish button is disabled
+      await page.$('button:disabled', { text: 'Publish' });
```
swissspidy:
```suggestion
      await expect(page).toMatchElement('button:disabled', { text: 'Publish' });
```

This needs an `expect` so that the test would fail if the element isn't found

samwhale: oh oops 🤦 good catch

swissspidy (approved):

Issue #9173, opened 2021-09-28
Safari page previews: some shapes display incorrectly


miina:
## Bug Description
Some shapes in Safari seem to display incorrectly, for example Pentagon, Heart. 

## Steps to Reproduce
In Safari:
1. Add Pentagon to the Page
2. See the Page preview

## Screenshots
<img width="162" alt="Screenshot 2021-09-28 at 11 28 56" src="https://user-images.githubusercontent.com/3294597/135052485-b43989ae-0832-4049-8bb1-05442a5e4b17.png">

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

swissspidy:
Closing as per discussion on #9933

PR #9168, opened 2021-09-27
Canvas to image improvements / Safari images fix

merapi:
## Summary

This fixes Safari missing images issue due to unfinished image decoding (the library was waiting an arbitrary 100ms to take the "screenshot", after an exhaustive investigation it turns out that adding more delay (and doing various other things), only partially fixes the issue [it happened less often]).
Thanks to that discovery we can also remove the 100ms delay, speeding up all `html-to-image` usages (Presets/Text Sets smart color + Eyedropper + Thumbnails caching) 🎉 

## To-do

- [ ] Upstream PRs

## Testing Instructions

- [ ] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

Can be also tested on other browsers but the Safari is the main goal.
1. Enable thumbnails caching in the Experiments
2. Create a story and insert a few pages with images
3. Check if thumbnails are generated properly (images not missing)

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [ ] I have tested this code to the best of my abilities
- [ ] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [ ] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [ ] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [ ] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #8923

PR #9264, opened 2021-10-05
Carousel: Enable page thumbnail caching by default

### LINKED ISSUE #9263: Enable feature by default: Page carousel - Cache with image thumbnails
## Feature Description
Now that issues with Safari have been fixed in #9168, we can enable the feature and set the feature flag to `true` by default.

miina:
## Summary
Sets the feature flag for caching page thumbnails to `true` now that issues for Safari have been fixed within https://github.com/google/web-stories-wp/pull/9168.

## Testing Instructions

1. Open the editor in Safari
2. Add a few pages with different images etc.
3. Verify that when navigating between the pages, the page thumbnails display the page content correctly and not just empty / gray as previously.
Note: some shapes display incorrectly for Safari, please ignore that, see #9173 

## Checklist

---

Fixes #9263


barklund (approved):

swissspidy (approved):
This should only be merged _after_ the 1.12.0 release

swissspidy:
@miina there are some odd e2e test failures now on this branch 🤔 

```
Error loading remote stylesheet SecurityError: Failed to read the 'cssRules' property from 'CSSStyleSheet': Cannot access rules
Error while reading CSS rules from fonts.googleapis.com/css?family=Google+Sans%7CGoogle+Sans%3Ab%7CGoogle+Sans%3A500&display=swap&ver=1.13.0-alpha.0 SecurityError: Failed to read the 'cssRules' property from 'CSSStyleSheet': Cannot access rules
```

It appears that `html-to-image` tries to read CSS rules:

https://github.com/bubkoo/html-to-image/blob/8dd487abfa9c12ee9c3553043cd5f21c87a78f91/src/embedWebFonts.ts#L123-L179

Are these errors negligible and we can ignore them in `bootstrap.js`? Or does that mean something is broken?

miina:
> there are some odd e2e test failures now on this branch 🤔

Hm, the error can be seen in console for the Contributor role outside of tests as well but doesn't look like it actually breaks anything. Seeing this in console with the Contributor role:
Screenshot 2021-10-07 at 11 46 42

A quick search says that supposedly adding `crossorigin="anonymous"` to the links could help with this but not sure if this is applicable here.

Would be ideal to fix these issues even if it doesn't cause issues using the editor.

One option could be keeping the feature disabled for contributors and creating another follow-up issue to fix it. Alternatively we could ignore the console errors until following up.
Any thoughts on why these errors happen for the Contributor role but not for other roles?

miina:
@choumx @gajasinski This might need a product decision.

With the thumbnail caching feature enabled, there seem to be console errors with the Contributor role, and also some e2e tests are failing due to the errors.
We could either ignore the console messages for now (since the functionality doesn't seem to be affected); or we could disable the feature for Contributor role (it's working for all the other roles), and in either case follow up with an issue to fix it.

Thoughts?

cc @bmattb for visibility

swissspidy:
Before we escalate this and get everyone involved here, I would really, really like to understand why this seemingly only happens for contributors. Could you please do some digging on that?

Are different fonts stylesheets loaded for these users? Do text elements with custom fonts still appear fine in the thumbnails?

Whatever solution we come up with would be really quick to implement. But we need to first understand what exactly is happening.

swissspidy:
@miina So the issue is not the contributor role in any way. The reason for what we're seeing here is that we just so happen to have video optimization enabled (thus using cross-origin isolation and `crossorigin="anonymous"`). Contributors simply cannot upload media, so they don't have any of that.

This is used for font embedding in html-to-image, but I am not sure about the difference this makes.

We could disable font embedding by [providing our own CSS](https://github.com/bubkoo/html-to-image#fontembedcss) or adding `crossorigin="anonymous"`.

Now, the former seems to be very beneficial for performance according to https://github.com/bubkoo/html-to-image/pull/109

> When calling the library multiple times for different elements, the embed code is run each time. This is where most of the execution time is spent when snapshotting elements on pages with webfonts. This change allows you to do the heavy lifting up front a single time so that the snapshotting process is much faster. In our testing, this further reduced time spent snapshotting 13 elements on the page from over 30 seconds to 12-13 seconds on average.

We could look into this.

Alternatively, we can also just pass an empty string to disable the font embedding. Again, I am not sure what difference this makes.

In any case, this could be done in a separate ticket and this message simple be ignored in tests for now.

miina:
@swissspidy The reason why I asked for product decision was that this is currently scheduled for 1.13 and I don't think we'll have time to fix this ahead of it. So I was thinking of creating a follow-up issue for it to do some digging there but meanwhile have a course of action for as it is now not to block the feature getting into 1.13.

> In any case, this could be done in a separate ticket and this message simple be ignored in tests for now.

That's essentially what I was looking to get a confirmation for -- what to do with the feature while we're looking into this in a separate issue.

miina:
Follow-up issue: https://github.com/google/web-stories-wp/issues/9327

swissspidy:
Thanks @miina!

PR #9268, opened 2021-10-05
Story Locking: only show if there's an actual lock user

swissspidy:
## Summary

Fixes showing the lock dialog if there wasn't actually a lock user set.

## Relevant Technical Choices

Remove redundant `showDialog` state variable

## User-facing changes

Fixes this bug:

Screenshot 2021-10-05 at 10 22 57

## Testing Instructions

1. Open editor
2. Don't see lock message

No

No

No

## Checklist

---

Fixes #

### DIFF
--- a/packages/wp-story-editor/src/components/postLock/postLock.js
+++ b/packages/wp-story-editor/src/components/postLock/postLock.js
@@ -67,14 +67,13 @@ function PostLock() {
   );
 
   const { enablePostLocking } = useFeatures();
-  const [showDialog, setShowDialog] = useState(false);
   const [isFirstTime, setIsFirstTime] = useState(true);
   const [user, setUser] = useState({});
   const [nonce, setNonce] = useState(firstNonce);
 
   // When dialog is closed, then set current user to lock owner.
   const closeDialog = useCallback(() => {
-    setShowDialog(false);
+    setUser({});
     setStoryLockById(storyId, stories);
   }, [storyId, stories]);
 
@@ -94,7 +93,6 @@ function PostLock() {
             avatar: _embedded?.author?.[0]?.avatar_urls?.['96'] || '',
           };
           if (locked && lockAuthor?.id && lockAuthor?.id !== currentUser.id) {
-            setShowDialog(true);
             setUser(lockAuthor);
           } else {
             setStoryLockById(storyId, stories);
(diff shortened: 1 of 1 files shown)


spacedmonkey:
@swissspidy Why does this fix the problem? What was the original problem?

packages/wp-story-editor/src/components/postLock/postLock.js
```diff
@@ -124,8 +122,7 @@ function PostLock() {
 
   useEffect(() => {
     if (enablePostLocking && showLockedDialog && currentUserLoaded) {
-      if (lockUser?.id !== currentUser.id) {
```
swissspidy, line 127: > Why does this fix the problem? What was the original problem?

@spacedmonkey This part was the underlying problem.

`lockUser` is always an object, with `lockUser.id` being `0` if there is no user.

So `lockUser?.id !== currentUser.id` is always `true`

dreamofabear: > lockUser is always an object

This is a bit confusing. Is making `lockUser` nullable possible?

> lockUser.id being 0 if there is no user.

Could we change the fallback `id` to `NaN`?

packages/wp-story-editor/src/components/postLock/postLock.js
```diff
@@ -140,7 +137,7 @@ function PostLock() {
   // Register an event on user navigating away from current tab to release / delete lock.
   useEffect(() => {
     function releasePostLock() {
-      if (enablePostLocking && showLockedDialog && !showDialog && nonce) {
+      if (enablePostLocking && showLockedDialog && user?.id && nonce) {
```
dreamofabear, line 140: Similar situation? `user` is always non-null but with falsy properties?

dreamofabear (approved):
Looks ok for a quick fix but I'd suggest some refactoring for better readability long-term:

1. Use null to represent "no user" or "no locked user" if possible
2. Use `NaN` or a random string for "invalid" property semantics to avoid unintentional truthy comparisons

swissspidy:
> Use null to represent "no user" or "no locked user" if possible

Yeah that makes sense.

We have similar situations in the editor for things like story poster, where we use an "empty" object instead of null. Best to tackle that coherently in a dedicated ticket 👍

packages/wp-story-editor/src/components/postLock/postLock.js
spacedmonkey, line 127: Would not have been cleaner to do this? 
```
if (lockUser && lockUser.id !== currentUser.id) {
```

PR #9289, opened 2021-10-05
Templates: swap templates to screenshots

### LINKED ISSUE #6678: Page Templates: Clicking on a template after applying a different one causes the thumbnails to jump
## Bug Description

If I scroll through the Page Layouts panel and apply one page, confirm and then scroll through it again and click on a different page layout, the panel automatically jumps to a different position, without applying the layout I clicked on.

## Expected Behaviour

Panel should not automatically jump, and user should be able to click any page layout visible on the panel at any moment to apply it.

## Steps to Reproduce

Apply a page layout, confirm, scroll through panel and click on a different page layout

## Screenshots

https://images.zenhubusercontent.com/235435637/72a65a24-703d-4e77-9ae5-b37c1b261477/screen_recording_2021_03_09_at_11_06_41_am.mp4

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System: MacOS
- Browser: Chrome Version 88.0.4324.192 (Official Build) (x86_64)

---

### LINKED ISSUE #9093: Default Page Templates: Swap for screenshots
## Description

This will include: removing the virtualized list, making sure page templates are memoized so we don't load excessively, and making sure that filtering and toggling images are still working - essentially, ensuring no regression. Right now the custom and default page templates share some components, it would probably make sense at this point to make a new page templates list component for static page templates, so that this work doesn't impact custom page templates at all but could be used for those as well once looking into that. 

## To Do 
- Do we want to keep the placeholder grid around? (@choumx following up on this) 
- Do we want to swap the toggle so that images are there by default? (@choumx following up on this)

Reference: #8826

## Acceptance Criteria

- Using screenshots for default page templates instead of previewing stories for each page. 
- Filtering by page template type still works. 
- Toggling between templates with images and with the placeholder has been removed
- Performance metrics to see improvements. 
- Passing tests
- Added aXe integration test for accessibility -- will be done in #8736

BrookeGraham:
## Summary

Using screenshots for default page templates instead of previewing stories for each page.
Filtering by page template type still works.
Toggling between templates with images and with the placeholder grid removed

Old metrics for initial load: 

staging load 1
staging load 2
staging load 3

New metrics for initial load:

new load 1
new load 2
new load 3

---
Old metrics for scrolling to the end of templates list:
staging scrolling 1
staging scrolling 2
staging scrolling 3

New metrics for scrolling to the end of templates list:
new scrolling 1
new scrolling 2
new scrolling 3

## To-do

- [x] Use arrow for keyboard nav. 
- [x] Clicking to add page bug, right now you have to click twice on an image before `handlePageClick` fires 🤷🏻‍♀️
- [x] Tests - aXe integration test for accessibility. - using useGridViewKeys, which have tests for keyboard accessibility.
- [x] Performance metrics to see improvements. 

## Testing Instructions

1. Make sure default and saved templates work same as before. 
2. Should be keyboard navigable
3. Default templates should now appear as images

no

no

no

## Checklist

---

Fixes #9093 
Fixes #6678

### DIFF
--- a/packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
+++ b/packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
@@ -21,11 +21,7 @@ import PropTypes from 'prop-types';
 import { useState, forwardRef, useFocusOut } from '@web-stories-wp/react';
 import styled from 'styled-components';
 import { _x, sprintf } from '@web-stories-wp/i18n';
-import {
-  Button,
-  BUTTON_TYPES,
-  themeHelpers,
-} from '@web-stories-wp/design-system';
+import { Button, BUTTON_TYPES } from '@web-stories-wp/design-system';
 /**
  * Internal dependencies
  */
@@ -43,7 +39,6 @@ const PageTemplateWrapper = styled(Button).attrs({ type: BUTTON_TYPES.PLAIN })`
   border-radius: ${({ theme }) => theme.borders.radius.small};
   cursor: pointer;
 
-  ${({ isHighlighted }) => isHighlighted && themeHelpers.focusCSS};
   ${focusStyle};
 `;
 
@@ -80,26 +75,21 @@ PageTemplateTitle.propTypes = {
 };
 
 const DefaultPageTemplate = forwardRef(
-  ({ page, pageSize, columnWidth, ...rest }, ref) => {
-    const [isActive, setIsActive] = useState(false);
+  ({ page, pageSize, columnWidth, isActive, ...rest }, ref) => {
     const [isHover, setIsHover] = useState(false);
     const isActivePage = isHover || isActive;
 
     useFocusOut(ref, () => setIsHover(false), []);
 
-    const { highlightedTemplate } = rest;
-
     const handleSetHover = () => setIsHover(!isHover);
+
     return (
       <PageTemplateWrapper
         columnWidth={columnWidth}
         ref={ref}
         onMouseEnter={handleSetHover}
         onMouseLeave={handleSetHover}
-        onFocus={() => setIsActive(true)}
-        onBlur={() => setIsActive(false)}
         aria-label={page.title}
-        isHighlighted={page.id === highlightedTemplate}
         {...rest}
       >
         <PosterWrapper>
--- a/packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
+++ b/packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
@@ -1,0 +1,159 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * External dependencies
+ */
+import PropTypes from 'prop-types';
+import {
+  useState,
+  useCallback,
+  forwardRef,
+  useFocusOut,
+  useEffect,
+  useRef,
+} from '@web-stories-wp/react';
+import styled from 'styled-components';
+import { _x, sprintf } from '@web-stories-wp/i18n';
+import {
+  Button,
+  BUTTON_TYPES,
+  themeHelpers,
+} from '@web-stories-wp/design-system';
+/**
+ * Internal dependencies
+ */
+import { PageSizePropType } from '../../../../types';
+import { focusStyle } from '../../../panels/shared';
+import { PAGE_TEMPLATE_TYPES } from './constants';
+
+const PageTemplateWrapper = styled(Button).attrs({ type: BUTTON_TYPES.PLAIN })`
+  position: relative;
+  z-index: 1;
+  display: flex;
+  flex-direction: column;
+  height: auto;
+  width: ${({ columnWidth }) => columnWidth}px;
+  padding: 0;
+  border-radius: ${({ theme }) => theme.borders.radius.small};
+  cursor: pointer;
+
+  ${({ isHighlighted }) => isHighlighted && themeHelpers.focusCSS};
+  ${focusStyle};
+`;
+
+const PosterWrapper = styled.div`
+  width: 100%;
+  height: 100%;
+  z-index: -1;
+`;
+
+const PosterImg = styled.img`
+  display: block;
+  width: 100%;
+  object-fit: cover;
+  border-radius: ${({ theme }) => theme.borders.radius.small};
+`;
+
+const PageTemplateTitle = styled.div`
+  position: absolute;
+  bottom: 0;
+  background-color: ${({ theme }) => theme.colors.opacity.overlayDark};
+  border-radius: ${({ theme }) => theme.borders.radius.small};
+  border-top-right-radius: 0;
+  border-top-left-radius: 0;
+  opacity: ${({ isActive }) => (isActive ? 1 : 0)};
+
+  padding: 8px;
+  font-size: 12px;
+  line-height: 22px;
+  width: 100%;
+  align-self: flex-end;
+`;
+
+PageTemplateTitle.propTypes = {
+  isActive: PropTypes.bool.isRequired,
+};
+
+function DefaultPageTemplate({
+  page,
+  isActive,
+  pageSize,
+  columnWidth,
+  ...rest
+}) {
+  const [isHover, setIsHover] = useState(false);
+  const isActivePage = isHover || isActive;
+  const ref = useRef();
+
+  useFocusOut(ref, () => setIsHover(false), []);
+
+  const { highlightedTemplate } = rest;
+
+  const handleSetHoverActive = useCallback(() => setIsHover(true), []);
+
+  const handleSetHoverFalse = useCallback(() => {
+    setIsHover(false);
+  }, []);
+
+  useEffect(() => {
+    if (isActive && ref.current) {
+      ref.current.focus();
+    }
+  });
+
+  return (
+    <PageTemplateWrapper
+      columnWidth={columnWidth}
+      role="listitem"
+      ref={ref}
+      onMouseEnter={handleSetHoverActive}
+      onMouseLeave={handleSetHoverFalse}
+      aria-label={page.title}
+      isHighlighted={page.id === highlightedTemplate}
+      {...rest}
+    >
+      <PosterWrapper>
+        {page.webp && (
+          <PosterImg src={page.png} alt={page.title} crossOrigin="anonymous" />
+        )}
+        {page.title && (
+          <PageTemplateTitle isActive={isActivePage}>
+            {sprintf(
+              /* translators: 1: template name. 2: page template
+            name. */ _x('%1$s %2$s', 'page template title', 'web-stories'),
+              page.title,
+              PAGE_TEMPLATE_TYPES[page.type].name
+            )}
+          </PageTemplateTitle>
+        )}
+      </PosterWrapper>
+    </PageTemplateWrapper>
+  );
+}
+
+const DefaultPageTemplateWithRef = forwardRef(DefaultPageTemplate);
+
+DefaultPageTemplate.propTypes = {
+  isActive: PropTypes.bool,
+  page: PropTypes.object.isRequired,
+  pageSize: PageSizePropType.isRequired,
+  columnWidth: PropTypes.number.isRequired,
+};
+
+DefaultPageTemplate.displayName = 'DefaultPageTemplate';
+
+export default DefaultPageTemplateWithRef;
--- a/packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
+++ b/packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
@@ -18,7 +18,7 @@
  * External dependencies
  */
 import PropTypes from 'prop-types';
-import { useState, forwardRef, useFocusOut } from '@web-stories-wp/react';
+import { forwardRef } from '@web-stories-wp/react';
 import styled from 'styled-components';
 import { _x, sprintf } from '@web-stories-wp/i18n';
 import { Button, BUTTON_TYPES } from '@web-stories-wp/design-system';
@@ -26,10 +26,11 @@ import { Button, BUTTON_TYPES } from '@web-stories-wp/design-system';
  * Internal dependencies
  */
 import { PageSizePropType } from '../../../../types';
-import { focusStyle } from '../../../panels/shared';
 import { PAGE_TEMPLATE_TYPES } from './constants';
 
-const PageTemplateWrapper = styled(Button).attrs({ type: BUTTON_TYPES.PLAIN })`
+const PageTemplateWrapper = styled.div``;
+
+const PageTemplateButton = styled(Button).attrs({ type: BUTTON_TYPES.PLAIN })`
   position: relative;
   display: flex;
   flex-direction: column;
@@ -38,8 +39,26 @@ const PageTemplateWrapper = styled(Button).attrs({ type: BUTTON_TYPES.PLAIN })`
   padding: 0;
   border-radius: ${({ theme }) => theme.borders.radius.small};
   cursor: pointer;
+`;
 
-  ${focusStyle};
+const PageTemplateTitle = styled.div`
+  position: absolute;
+  bottom: 0;
+  background-color: ${({ theme }) => theme.colors.opacity.overlayDark};
+  border-radius: ${({ theme }) => theme.borders.radius.small};
+  border-top-right-radius: 0;
+  border-top-left-radius: 0;
+  padding: 8px;
+  font-size: 12px;
+  line-height: 22px;
+  width: 100%;
+  align-self: flex-end;
+  opacity: 0;
+
+  ${PageTemplateButton}:hover &,
+  ${PageTemplateButton}:focus & {
+    opacity: 1;
+  }
 `;
 
 const PosterWrapper = styled.div`
@@ -54,63 +73,39 @@ const PosterImg = styled.img`
   border-radius: ${({ theme }) => theme.borders.radius.small};
 `;
 
-const PageTemplateTitle = styled.div`
-  position: absolute;
-  bottom: 0;
-  background-color: ${({ theme }) => theme.colors.opacity.overlayDark};
-  border-radius: ${({ theme }) => theme.borders.radius.small};
-  border-top-right-radius: 0;
-  border-top-left-radius: 0;
-  opacity: ${({ isActive }) => (isActive ? 1 : 0)};
-
-  padding: 8px;
-  font-size: 12px;
-  line-height: 22px;
-  width: 100%;
-  align-self: flex-end;
-`;
-
 PageTemplateTitle.propTypes = {
   isActive: PropTypes.bool.isRequired,
 };
 
 const DefaultPageTemplate = forwardRef(
-  ({ page, pageSize, columnWidth, isActive, ...rest }, ref) => {
-    const [isHover, setIsHover] = useState(false);
-    const isActivePage = isHover || isActive;
-
-    useFocusOut(ref, () => setIsHover(false), []);
-
-    const handleSetHover = () => setIsHover(!isHover);
-
+  ({ page, columnWidth, isActive }, ref) => {
     return (
-      <PageTemplateWrapper
-        columnWidth={columnWidth}
-        ref={ref}
-        onMouseEnter={handleSetHover}
-        onMouseLeave={handleSetHover}
-        aria-label={page.title}
-        {...rest}
-      >
-        <PosterWrapper>
-          {page.webp && (
-            <PosterImg
-              src={page.png}
-              alt={page.title}
-              crossOrigin="anonymous"
-            />
-          )}
+      <PageTemplateWrapper ref={ref}>
+        <PageTemplateButton
+          columnWidth={columnWidth}
+          aria-label={page.title}
+          tabIndex={isActive ? 0 : -1}
+        >
+          <PosterWrapper>
+            {page.webp && (
+              <PosterImg
+                src={page.png}
+                alt={page.title}
+                crossOrigin="anonymous"
+              />
+            )}
+          </PosterWrapper>
           {page.title && (
-            <PageTemplateTitle isActive={isActivePage}>
+            <PageTemplateTitle>
               {sprintf(
-                /* translators: 1: template name. 2: page template
-            name. */ _x('%1$s %2$s', 'page template title', 'web-stories'),
+                /* translators: 1: template name. 2: page template name. */
+                _x('%1$s %2$s', 'page template title', 'web-stories'),
                 page.title,
                 PAGE_TEMPLATE_TYPES[page.type].name
               )}
             </PageTemplateTitle>
           )}
-        </PosterWrapper>
+        </PageTemplateButton>
       </PageTemplateWrapper>
     );
   }
--- a/packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
+++ b/packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
@@ -25,7 +25,6 @@ import { Button, BUTTON_TYPES } from '@web-stories-wp/design-system';
 /**
  * Internal dependencies
  */
-import { PageSizePropType } from '../../../../types';
 import { PAGE_TEMPLATE_TYPES } from './constants';
 
 const PageTemplateWrapper = styled.div``;
@@ -74,17 +73,18 @@ const PosterImg = styled.img`
 `;
 
 PageTemplateTitle.propTypes = {
-  isActive: PropTypes.bool.isRequired,
+  isActive: PropTypes.bool,
 };
 
 const DefaultPageTemplate = forwardRef(
-  ({ page, columnWidth, isActive }, ref) => {
+  ({ page, columnWidth, isActive, ...rest }, ref) => {
     return (
       <PageTemplateWrapper ref={ref}>
         <PageTemplateButton
           columnWidth={columnWidth}
           aria-label={page.title}
           tabIndex={isActive ? 0 : -1}
+          {...rest}
         >
           <PosterWrapper>
             {page.webp && (
@@ -114,7 +114,6 @@ const DefaultPageTemplate = forwardRef(
 DefaultPageTemplate.propTypes = {
   isActive: PropTypes.bool,
   page: PropTypes.object.isRequired,
-  pageSize: PageSizePropType.isRequired,
   columnWidth: PropTypes.number.isRequired,
 };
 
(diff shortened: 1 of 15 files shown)


packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplateList.js
```diff
@@ -0,0 +1,168 @@
...
+      setTabIndex('-1');
+      setIsGridFocused(false);
+    }
+  };
+  // reset tabIndex onBlur so that we can use the keyboard to get back into templates
+  const handleBlur = () => {
+    if (tabIndex === '-1') {
+      setTabIndex('0');
+    }
+  };
```
BrookeGraham: This is a little hacky, totally open to thoughts on how to allow users to exit the focus trap.

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
```diff
@@ -0,0 +1,128 @@
...
+  position: relative;
+  display: flex;
+  flex-direction: column;
+  height: auto;
+  width: ${({ columnWidth }) => columnWidth}px;
+  padding: 0;
+  border-radius: ${({ theme }) => theme.borders.radius.small};
+  cursor: pointer;
+
+  ${focusStyle};
```
BrittanyIRL: Now that these are getting the default DS button styles focus styles are taken care of already so `focusStyle` is not necessary. 
```suggestion
```

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
```diff
@@ -0,0 +1,128 @@
...
+  display: block;
+  width: 100%;
+  object-fit: cover;
+  border-radius: ${({ theme }) => theme.borders.radius.small};
+`;
+
+const PageTemplateTitle = styled.div`
+  position: absolute;
+  bottom: 0;
+  background-color: ${({ theme }) => theme.colors.opacity.overlayDark};
```
BrittanyIRL: Not your fault but we should look at fixing[ this contrast ratio](https://contrast-ratio.com/#%23E4E5E6-on-rgba%280%2C0%2C0%2C0.47%29), it wasn't noticeable before to me but now it is:

Screen Shot 2021-10-07 at 4 35 41 PM

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
```diff
@@ -0,0 +1,128 @@
...
+`;
+
+const PosterImg = styled.img`
+  display: block;
+  width: 100%;
+  object-fit: cover;
+  border-radius: ${({ theme }) => theme.borders.radius.small};
+`;
+
+const PageTemplateTitle = styled.div`
```
BrittanyIRL: 🤔 Perhaps this would be better splitting the container and the text now that the DOM is so much cleaner? 

That would make it really straight forward to use `Text` from the design system, like so: 
`<Text as="span" size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL} isBold />`

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
```diff
@@ -0,0 +1,128 @@
...
+`;
+
+PageTemplateTitle.propTypes = {
+  isActive: PropTypes.bool.isRequired,
+};
+
+const DefaultPageTemplate = forwardRef(
+  ({ page, pageSize, columnWidth, isActive, ...rest }, ref) => {
+    const [isHover, setIsHover] = useState(false);
+    const isActivePage = isHover || isActive;
```
BrittanyIRL: Now that nothing is animating, can we rely on browser level hover focus?

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
```diff
@@ -0,0 +1,128 @@
...
+              src={page.png}
+              alt={page.title}
+              crossOrigin="anonymous"
+            />
+          )}
+          {page.title && (
+            <PageTemplateTitle isActive={isActivePage}>
+              {sprintf(
+                /* translators: 1: template name. 2: page template
+            name. */ _x('%1$s %2$s', 'page template title', 'web-stories'),
```
BrittanyIRL: nit: weird line break? 

```suggestion
                /* translators: 1: template name. 2: page template name. */
                _x('%1$s %2$s', 'page template title', 'web-stories'),
```

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
```diff
@@ -0,0 +1,128 @@
...
+        <PosterWrapper>
+          {page.webp && (
+            <PosterImg
+              src={page.png}
+              alt={page.title}
+              crossOrigin="anonymous"
+            />
+          )}
+          {page.title && (
+            <PageTemplateTitle isActive={isActivePage}>
```
BrittanyIRL: Let's use css pseudo classes here instead of `isActive`

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplateList.js
```diff
@@ -0,0 +1,168 @@
...
+            pageSize={pageSize}
+            onFocus={() => {
+              setCurrentPageId(page.id);
+            }}
+            onBlur={handleBlur}
+            isActive={currentPageId === page.id}
+            onClick={() => handlePageClick(page.story)}
+            columnWidth={pageSize.width}
+            tabIndex={tabIndex}
+            {...rest}
```
BrittanyIRL: You're passing rest in here as well as in the wrapper, is it needed in both (or either)?

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplateList.js
BrittanyIRL: Took this for a little spin and think we can simplify it a bit :) 

`useGridViewKeys` should be taking care of all of this now that it's just a group of buttons that we want to traverse with arrow keys! I'm leaving comments through here to show what I'm thinking.

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplateList.js
```diff
@@ -0,0 +1,168 @@
...
+    if (!node) {
+      return undefined;
+    }
+
+    node.addEventListener('keydown', handleFocusOut);
+
+    return () => {
+      node.removeEventListener('keydown', handleFocusOut);
+    };
+  }, [containerRef]);
```
BrittanyIRL: This can be deleted, focus should be handled by the `useGridViewKeys` util.

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplateList.js
BrittanyIRL: This can also be deleted, as focus should be handled by that `useGridViewKeys` util as well.

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplateList.js
```diff
@@ -0,0 +1,168 @@
...
+    if (!isGridFocused) {
+      const newGridItemId = pageRefs.current?.[currentPageId]
+        ? currentPageId
+        : pages[0].id;
+
+      !currentPageId && setCurrentPageId(newGridItemId);
+      setIsGridFocused(true);
+      pageRefs.current?.[newGridItemId]?.focus();
+    }
+  }, [currentPageId, isGridFocused, pages]);
```
BrittanyIRL: We don't need any of this either now that we're just dealing with buttons! A lot of the code that was in here was the way it was because of how gross the HTML was for rendering actual story previews.

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplateList.js
```diff
@@ -0,0 +1,168 @@
...
+
+function DefaultTemplateList({ pages, parentRef, pageSize, ...rest }) {
+  const { addPage } = useStory(({ actions }) => ({
+    addPage: actions.addPage,
+  }));
+  const { isRTL } = useConfig();
+  const { showSnackbar } = useSnackbar();
+  const [currentPageId, setCurrentPageId] = useState();
+  const [isGridFocused, setIsGridFocused] = useState(false);
+  const [tabIndex, setTabIndex] = useState('0');
```
BrittanyIRL: These can be removed. Let's use the `currentPageId` in tandem with the `page.id` when mapping through the list to find out which button should have tabIndex 0.

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplateList.js
```diff
@@ -0,0 +1,168 @@
...
+    currentItemId: currentPageId,
+    isRTL,
+  });
+
+  return (
+    <WrapperGrid
+      ref={containerRef}
+      columnWidth={pageSize.width}
+      rowHeight={pageSize.containerHeight}
+      onFocus={handleGridFocus}
```
BrittanyIRL: Let's remove this onFocus and pass in `onFocus` to the `DefaultPageTemplate` since buttons are organically focusable.

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplateList.js
```diff
@@ -0,0 +1,168 @@
...
+      {pages.map((page) => {
+        return (
+          <DefaultPageTemplate
+            ref={(el) => (pageRefs.current[page.id] = el)}
+            key={page.id}
+            data-testid={`page_template_${page.id}`}
+            page={page}
+            pageSize={pageSize}
+            onFocus={() => {
+              setCurrentPageId(page.id);
```
BrittanyIRL: in additoin to setting the currentPageId here, let's add a `useEffect` above doing something like this: 
```
if (pages.length > 0 && !currentPageId) {
      setCurrentPageId(pages[0].id);
    }
```
so that if (first time) the user is tabbing to this grid, the first button is set to have focus available since `currentPageId` is getting used to decide `tabIndex`

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplateList.js
```diff
@@ -0,0 +1,168 @@
...
+            }}
+            onBlur={handleBlur}
+            isActive={currentPageId === page.id}
+            onClick={() => handlePageClick(page.story)}
+            columnWidth={pageSize.width}
+            tabIndex={tabIndex}
```
BrittanyIRL: remove :)

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
```diff
@@ -0,0 +1,128 @@
...
+};
+
+const DefaultPageTemplate = forwardRef(
+  ({ page, pageSize, columnWidth, isActive, ...rest }, ref) => {
+    const [isHover, setIsHover] = useState(false);
+    const isActivePage = isHover || isActive;
+
+    useFocusOut(ref, () => setIsHover(false), []);
+
+    const handleSetHover = () => setIsHover(!isHover);
```
BrittanyIRL: These can hopefully be removed

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
```diff
@@ -0,0 +1,128 @@
...
+
+    return (
+      <PageTemplateWrapper
```
BrittanyIRL: So part of the issue while trying to use the `useGridViewKeys` util that you were running up against was the button being the "wrapper" here, take a look at other implementations of this hook and see that there's a div around the focusable item. If you wrap `PageTemplateWrapper` (maybe renaming it `PageTemplateButton`?) in a `<div>` and put the `ref` on the `div` and everything else on the button (ask it is) and also add in: `tabIndex={isActive ? 0 : -1}` focusing should work.

BrittanyIRL (changes requested):
😅  this was a _lot_ to untangle and refactor, great work. 

Left some comments about further clean up and simplification. Happy to jump on a call next week about anything!!

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
BrookeGraham: I left this as a div since we have so much custom styling that comes with it. ie. focus and hover state, absolute positioning, etc.

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
BrookeGraham: Should I check with Amy or Andrew on this?

BrookeGraham: Asked in design channel. Will update once I receive confirmation.

BrookeGraham: Any confirmed that she's okay with using the darker theme color.

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
maxyinger: I would just break up the styling per element a little bit to use some of our text styling. just make this `PageTemplateTitleContainer` and do something like:
```
            <PageTemplateTitleContainer>
              <Text as="span" size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL} isBold >
                {sprintf(
                  /* translators: 1: template name. 2: page template name. */
                  _x('%1$s %2$s', 'page template title', 'web-stories'),
                  page.title,
                  PAGE_TEMPLATE_TYPES[page.type].name
                )}
              <Text />
            </PageTemplateTitleContainer>
```

either that or use one of the `themeHelpers` like `themeHelpers.expandPresetStyles({ preset, theme })` to get the majority of the font styles in here, then tweak as needed with custom styles

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
```diff
@@ -0,0 +1,122 @@
...
+          tabIndex={isActive ? 0 : -1}
+          {...rest}
+        >
+          <PosterWrapper>
+            {page.webp && (
+              <PosterImg
+                src={page.png}
+                alt={page.title}
+                crossOrigin="anonymous"
+              />
```
maxyinger: Adding draggable false here should prevent the user from being able to drag the image around and possibly making them thing that this component is draggable
```jsx
              <PosterImg
                src={page.png}
                alt={page.title}
                crossOrigin="anonymous"
                draggable={false}
              />
```

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
```diff
@@ -0,0 +1,122 @@
...
+const PageTemplateButton = styled(Button).attrs({ type: BUTTON_TYPES.PLAIN })`
+  position: relative;
+  display: flex;
+  flex-direction: column;
+  height: auto;
+  width: ${({ columnWidth }) => columnWidth}px;
+  padding: 0;
+  border-radius: ${({ theme }) => theme.borders.radius.small};
+  cursor: pointer;
+`;
```
maxyinger: Okay, I think we can clean these styles up a bit. 

I think the mental model we should go by here is that the parent container sets the grid column width, so we should just respect that width all the way down to the image and make the image take up that full width. The image should then layout according to its own aspect ratio which should set the height for everything. I'll walk you through the css with some comments sprinkled in this file.

maxyinger: Here are the properties we'll remove
```css
  /* the contents of this element are 1 element taking up the full size and another element 
      laid over top of it, so we don't need these properties*/
  display: flex;
  flex-direction: column;
  
  /* we're gonna let the image dictate it's own height later, so we don't need to worry about it here */
  height: auto;
  
  /* we're just taking up the full width of the parent component, so we don't need an explicit width here */
  width: ${({ columnWidth }) => columnWidth}px;
  
  /* this should already be inherited by the base button class, so we don't need to redeclare it here */
  cursor: pointer;
```

```suggestion
const PageTemplateButton = styled(Button).attrs({ type: BUTTON_TYPES.PLAIN })`
 /* will take up full width of parent */
  display: block;
  
  /* this will make all relative & absolute positioned elements relative to this */
  position: relative;
  
  padding: 0;
  border-radius: ${({ theme }) => theme.borders.radius.small};
  
   /* since the overflow is hidden, we won't need to redeclare border-radiuses on children cus those bits will just get cut off with this */
  overflow: hidden;
`;
```

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
```diff
@@ -0,0 +1,122 @@
...
+  font-size: 12px;
+  line-height: 22px;
+  width: 100%;
+  align-self: flex-end;
+  opacity: 0;
+
+  ${PageTemplateButton}:hover &,
+  ${PageTemplateButton}:focus & {
+    opacity: 1;
+  }
```
maxyinger:
```suggestion
  /* since the button above is relative, this will lay this content over all the other elements in here and position it at the bottom. no need to touch anything regarding flex here*/
  position: absolute;
  bottom: 0;
  
  width: 100%;
  padding: 8px;
  background-color: ${({ theme }) => theme.colors.opacity.black64};
  opacity: 0;

  ${PageTemplateButton}:hover &,
  ${PageTemplateButton}:focus & {
    opacity: 1;
  }
```

maxyinger: since the parent has `overflow: hidden` we don't have to worry about rounding any edges here because all the edges are already visually cut off by the parent

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
```diff
@@ -0,0 +1,122 @@
...
+`;
+
+const PosterWrapper = styled.div`
+  width: 100%;
+  height: 100%;
+`;
```
maxyinger: We can just remove this container entirely and position the image directly inside the button

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
```diff
@@ -0,0 +1,122 @@
...
+
+const PosterImg = styled.img`
+  display: block;
+  width: 100%;
+  object-fit: cover;
+  border-radius: ${({ theme }) => theme.borders.radius.small};
+`;
```
maxyinger:
```suggestion
const PosterImg = styled.img`
  display: block;
   /* by default the image lays out to it's true width, we want to manually set it to take up the full width of it's parent */
   width: 100%; 
`;
```

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplates.js
```diff
@@ -73,22 +57,17 @@ function DefaultTemplates({ pageSize }) {
...
-  const toggleId = useMemo(() => `toggle_page_templates_${uuidv4()}`, []);
 
   // load and process pageTemplates
   useEffect(() => {
     async function loadPageTemplates() {
       const trackTiming = getTimeTracker('load_page_templates');
-      setPageTemplates(
-        await getPageTemplates({ showImages: showTemplateImages })
-      );
+      setPageTemplates(await getPageTemplates());
```
maxyinger, line 65: ahh so if we do want the template images can just pass a simple showImages here:
```
      setPageTemplates(
        await getPageTemplates({ showImages: true })
      );
```

packages/story-editor/src/components/library/panes/pageTemplates/templateList.js
```diff
@@ -56,7 +56,7 @@ function TemplateList({
   const containerRef = useRef();
   const pageRefs = useRef({});
 
-  const pageIds = pages?.map((page) => page.id) || [];
+  const pageIds = useMemo(() => pages.map((page) => page.id), [pages]);
```
maxyinger: should we still be cautionary here?
```
 const pageIds = useMemo(() => pages?.map((page) => page.id) || [], [pages]);
```

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplates.js
```diff
@@ -168,24 +140,9 @@ function DefaultTemplates({ pageSize }) {
...
-            >
-              {__('Show Images', 'web-stories')}
-            </Text>
-            <Toggle
-              id={toggleId}
-              name={toggleId}
-              checked={showTemplateImages}
-              onChange={handleToggleClick}
-            />
-          </TemplatesToggle>
```
samwhale, line 185: Just to double check: we're removing these because we're now rendering these as static images. So we can't get screenshots of what they look like without the images?

maxyinger (approved):
Looks great! Just left some minor implementation details that don't really change anything here. Awesome work!! so much cleaner and works so much better 👏

samwhale (approved):
Nice, max's review covers css stuff in depth. I just have one question about a toggle that was removed, but other than that it works great

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplates.js
BrookeGraham: Correct.

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplates.js
BrookeGraham: For now, we want to add templates without images.  [See thread](https://formidablelabs.slack.com/archives/C01ESBB8ZN1/p1634236997169600?thread_ts=1634234530.168300&cid=C01ESBB8ZN1)

BrookeGraham:
> Looks great! Just left some minor implementation details that don't really change anything here. Awesome work!! so much cleaner and works so much better 👏

Thanks, so much better now!

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplates.js
BrittanyIRL: this still feels weird. i get that this came from design but it feels odd to me as a user to select a template and then have it show up and look different from what i expected.

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
```diff
@@ -0,0 +1,115 @@
...
+  ({ page, columnWidth, isActive, ...rest }, ref) => {
+    return (
+      <PageTemplateWrapper ref={ref}>
+        <PageTemplateButton
+          columnWidth={columnWidth}
+          aria-label={page.title}
+          tabIndex={isActive ? 0 : -1}
+          {...rest}
+        >
+          {page.webp && (
```
BrittanyIRL: nit: you're looking for `page.webp` here and then using `page.png` as the img src within, should probably check for `.png` if using that or swap to `webp`

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
```diff
@@ -0,0 +1,115 @@
...
+    opacity: 1;
+  }
+`;
+
+const PosterImg = styled.img`
+  display: block;
+  width: 100%;
+`;
+
+PageTemplateTitleContainer.propTypes = {
```
BrittanyIRL: `PageTemplateTitleContainer` is getting `isActive` but that prop's not getting passed through at all. If it's still needed it should be a transient prop and passed through like `<PageTemplateTitleContainer $isActive={isActive}` since it's only getting passed for styles and serves no other purpose at that level.  

That said, pretty sure with the simplifications you were able to make this is obsolete! 😅

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplateList.js
```diff
@@ -0,0 +1,129 @@
...
+      {pages.map((page) => {
+        return (
+          <DefaultPageTemplate
+            ref={(el) => (pageRefs.current[page.id] = el)}
+            key={page.id}
+            data-testid={`page_template_${page.id}`}
+            page={page}
+            pageSize={pageSize}
+            onFocus={() => {
+              setCurrentPageId(page.id);
```
BrittanyIRL: 🤔  Callback?
`onFocus={() => handlePageFocus(page.id)}`

`const handlePageFocus = ({id}) => setCurrentPageId(id), []);`

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplates.js
```diff
@@ -108,30 +87,27 @@ function DefaultTemplates({ pageSize }) {
   const filteredPages = useMemo(
     () =>
       pageTemplates.reduce((pages, template) => {
-        const templatePages = template.pages.reduce((acc, page) => {
+        const templatePosters = Object.values(template.postersByPage).map(
+          (poster, index) => ({
+            id: `${template.slug}_${index}`,
+            title: template.title,
+            story: template.pages[index],
+            type: poster.type,
```
BrittanyIRL: Was there a specific purpose in setting this here separately from the spread `poster` below it?

BrittanyIRL (changes requested):
Thank you for the updates! 

Few smaller things I noticed on re-review and then these two things that are more generic: 
Screen Shot 2021-10-18 at 11 55 05 AM

Looks like on scroll the templates are overlapping with this button. Ideally, there's no overlap.
I know this isn't necessarily new, but we should leave it better off with this update since the styles are getting a facelift in this PR. 

The focusing is _much_ improved, i did notice one thing though that i'm curious if you can reproduce. If you use the keyboard and filter templates to anything besides all and then go into the grid and select one, then tab back to the filters and select a new filter that isn't all and tab back to the grid - the focus disappears.

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplates.js
BrookeGraham: nope. Removed

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplateList.js
BrookeGraham: ☑️

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
BrookeGraham: Nice catch!

packages/story-editor/src/components/library/panes/pageTemplates/defaultPageTemplate.js
BrookeGraham: Totes.

BrookeGraham:
> Looks like on scroll the templates are overlapping with this button. Ideally, there's no overlap.
I know this isn't necessarily new, but we should leave it better off with this update since the styles are getting a facelift in this PR.

Fixed spacing

> The focusing is much improved, i did notice one thing though that i'm curious if you can reproduce. If you use the keyboard and filter templates to anything besides all and then go into the grid and select one, then tab back to the filters and select a new filter that isn't all and tab back to the grid - the focus disappears.
 
Found it. The `currentPageId` was being set to the first item on initial load, which was a template type `cover` so when we filtered to a type that wasn't `cover`, the template was no longer `active`. Added a check in the useEffect to make sure the `currentPageId` is indeed in the pages array after filtering.

BrittanyIRL (approved):
This works great and is truly a massive improvement, thank you for all of your diligence here.

BrittanyIRL:
(I think the page template karma tests are going to need some updates - i know there's a follow up for that but we can't merge this (after qa) with the failing tests)

BrookeGraham:
> (I think the page template karma tests are going to need some updates - i know there's a follow up for that but we can't merge this (after qa) with the failing tests)

Yeah totally, they are failing right now. I will update the tests and link the ticket to this pr.

swissspidy:
Is this e2e test really flakey? 🤨

It seems to run fine on main, but is broken in this PR, indicating a regression that should get fixed, not skipped.

PR #9293, opened 2021-10-05
`ServiceBasedPlugin`: Don't resolve service IDs as callables

### LINKED ISSUE #9292: `ServiceBasedPlugin` tries to resolve service ID as callable
## Bug Description

Our services have IDs like `svg`.

In `ServiceBasedPlugin` not only the class names are tried to resolve as callables, but also the IDs.

So if a function like `svg` actually exists as a callable, there's gonna be an error because it tries to actually call that function.

## Expected Behaviour

Service IDs should not be used as callables.

## Steps to Reproduce

1. Try to install [Call Now Button](https://wordpress.org/plugins/call-now-button/) plugin, which declares a global `svg` function
2. See error

## Additional Context

- Plugin Version: 1.12.0
- WordPress Version: 5.8
- Operating System: macOS
- Browser: Chrome

swissspidy:
## Context

If the service ID is the same as a public function in the global namespace, `ServiceBasedPlugin` tried to call it.

## Summary

Don't resolve service IDs as callables.

## Relevant Technical Choices

--

## To-do

--

## User-facing changes

None

## Testing Instructions

1. Activate Web Stories plugin
2. Activate [Call Now Button](https://wordpress.org/plugins/call-now-button/) plugin
3. Don't see error

No

No

No

## Checklist

---

Fixes #9292

### DIFF
--- a/includes/Infrastructure/ServiceBasedPlugin.php
+++ b/includes/Infrastructure/ServiceBasedPlugin.php
@@ -275,7 +275,7 @@ abstract class ServiceBasedPlugin implements Plugin {
 		}
 
 		while ( null !== key( $services ) ) {
-			$id    = $this->maybe_resolve( key( $services ) );
+			$id    = key( $services );
 			$class = $this->maybe_resolve( current( $services ) );
 
 			// Delay registering the service until all requirements are met.
--- a/includes/Infrastructure/ServiceBasedPlugin.php
+++ b/includes/Infrastructure/ServiceBasedPlugin.php
@@ -275,7 +275,7 @@ abstract class ServiceBasedPlugin implements Plugin {
 		}
 
 		while ( null !== key( $services ) ) {
-			$id    = key( $services );
+			$id    = $this->maybe_resolve( key( $services ) );
 			$class = $this->maybe_resolve( current( $services ) );
 
 			// Delay registering the service until all requirements are met.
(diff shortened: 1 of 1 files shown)


swissspidy:
FYI @schlessera @westonruter since the AMP plugin contains the same code.

schlessera:
Another approach would be to fix the `maybe_resolve()` method to not accept function strings.

Suggestion based on the experiment here: https://3v4l.org/e1ja6#v8.1rc3:
```php
	protected function maybe_resolve( $value ) {
		if ( is_callable( $value ) && ! ( is_string( $value ) && function_exists( $value ) ) {
			$value = $value( $this->injector, $this->service_container );
		}

		return $value;
	}
```

This would ensure this resolution cannot break in other places where it is being used either.

swissspidy:
I suppose both could work for extra hardening?

schlessera:
Yes, but not resolving the IDs means you can never have dynamically computed IDs, of course.

spacedmonkey:
@swissspidy Can you provide more details on this PR. I can not sure I understand the issue here.

swissspidy:
@spacedmonkey it's explained in the ticket description #9292

Our services have IDs like `svg`.

In `ServiceBasedPlugin` not only the class names are tried to resolve as callables, but also the IDs.

So if a function like `svg` actually exists as a callable, there's gonna be an error because it tries to actually call that function.

spacedmonkey (approved):

PR #9298, opened 2021-10-06
Output: Fix malformed `<a>` tags in the `<head>`

### LINKED ISSUE #9297: Mangled `<script>` / `<a>` tags on WordPress.com VIP
## Bug Description

We've seen stories in the wild and got direct feedback from partners on WordPress.com VIP for stories with broken markup when not publishing as a super admin, indicating issues with KSES.

Our KSES implementation is supposed to allow `<script>` tags, but somehow on VIP there are some higher powers that still prevent `<script>` tags.

Publishing as a super admin (which has `unfiltered_html` caps on Multisite) is of course only a temporary workaround.

Ideally we'd fix the root cause of this at the KSES step before saving. But we could also fix this before serving the page.

## Expected Behaviour

- No broken output
- Passing AMP validation

## Steps to Reproduce

TBD

## Screenshots

Expected markup in the database:
Screenshot 2021-10-06 at 15 43 33

Actual markup in the database:

Screenshot 2021-10-06 at 15 43 22

What it will look like on the frontend:

Screenshot 2021-10-06 at 15 42 59

## Additional Context

- Plugin Version: 1.12.0
- WordPress Version: 5.8
- Operating System: macOS
- Browser: Chrome
- Environment: WordPress.com VIP

Related: #2413

swissspidy:
## Context

There can be some mangled `<script>` tags turned into `<a>` on WP.com VIP.

## Summary

This PR detects incorrect `<a>` tags in the HTML output and turns them into actual `<script>` tags as expected.

## Relevant Technical Choices

Since it is **not** possible to fix this with the sanitizers, and since the root cause with the KSES logic is unclear, this workaround lives in the `HTML` class on purpose.

## To-do

- Testing

## User-facing changes

N/A

## Testing Instructions

1. TBD

No

No

No

## Checklist

---

Fixes #9297


includes/Renderer/Story/HTML.php
```diff
@@ -67,6 +68,7 @@ public function __construct( Story $story ) {
 	 */
 	public function render(): string {
 		$markup = $this->story->get_markup();
+		$markup = $this->fix_malformed_script_link_tags( $markup );
```
spacedmonkey, line 71: Could we maybe only conditionally run this logic. Like checking for `VIP_GO_ENV` const. [See](https://github.com/Automattic/vip-go-mu-plugins/blob/291e8d413545ccc6a537e9415bc7e9ebce543e4e/000-vip-init.php#L83-L94).

swissspidy: While I have only encountered this on VIP sites so far, it doesn't mean that it can't happen on other environments as well.

So I don't want to run this conditionally. There's no harm in always running this, performance-wise the impact should be minimal, the regex is pretty lightweight.

swissspidy:
I've sent this to the partner for review and will provide an update once I know whether it works as expected.

swissspidy:
Got confirmation that it works as expected! 🎉

includes/Renderer/Story/HTML.php
swissspidy: Just got confirmation that it was reproducible outside of the WPCOMVIP environment. So definitely not going to run this conditionally.

spacedmonkey (changes requested):
There are two PHPCS lint warnings I would like to see fixed before approving this.

swissspidy:
> There are two PHPCS lint warnings I would like to see fixed before approving this.

Those are just false positives that I now ignored.

Also weird that these are just warnings and not errors causing the check to fail 🤔

spacedmonkey (approved):

PR #9302, opened 2021-10-06
Taxonomy: Allow categories with duplicate names

### LINKED ISSUE #9279: Taxonomy: Duplicating Categories
## Bug Description

Reported by @barklund 
>  I can’t add a duplicate category name somewhere else in the category tree. Is that WP standard? I want to add a category for the clothing type “Cardigan”, but nothing happens when I try to do so - I suspect it’s because we already have the corgi sub-category “Cardigan”, though that’s a completely different thing.

If you add categories in the categories list table (the thing that looks like a categories dashboard in the stories plugin, see screenshot below) you can add duplicative categories. This makes sense, given examples (as pointed out by @spacedmonkey ) like `Countries > Canada > Maple Syrup` or `Syrups > Maple Syrup` where both cases of maple syrup are independent and should be valid and separate. 

## Expected Behaviour

The editor should allow duplicative categories when their hierarchy is different.

Also, let's add a new e2e test case here.

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

maxyinger:
## Context
Part of Taxonomy follow up from bug bash

## Summary
This PR allows for users to have hierarchical terms with the same name see terms `Rock` & `Roll` both existing, but under different parent categories:
Screen Shot 2021-10-06 at 12 24 10 PM

## Relevant Technical Choices
Just prepended the slug with the parent slug for storage. Luckily this works for multiple tiers cus every slug created will have already been prepended with its parents slug under this update.

## To-do
NA

## User-facing changes
Users can now have Nested Categories with the same name and differing parents.

## Testing Instructions

1. Go to wp-admin > stories > categories and bulk action delete all your categories
2. Open up the Editor -> Document Tab -> Categories & Tags Panel
3. Add 2 new categories with the same name, but different parents. See the term `Butt` here for instance
```
[ ] Pork
    [ ] Loin
    [ ] Butt
[ ] Ultimate Appendicular Combat Moves
    [ ] Head
        [ ] Butt 
    [ ] Arm
        [ ] Ponch
        [ ] Elbow 
```

## Checklist

---

Fixes #9279

### DIFF
--- a/packages/story-editor/src/app/taxonomy/taxonomyProvider.js
+++ b/packages/story-editor/src/app/taxonomy/taxonomyProvider.js
@@ -190,9 +190,16 @@ function TaxonomyProvider(props) {
   );
 
   const createTerm = useCallback(
-    async (taxonomy, termName, parentId, addToSelection = false) => {
+    async (taxonomy, termName, parent, addToSelection = false) => {
+      const data = { name: termName };
+      if (parent?.id) {
+        data.parent = parent.id;
+        data.slug = `${parent.slug}-${cleanForSlug(data.name)}`;
+      }
+
       // make sure the term doesn't already exist locally
-      const cachedTerm = termCache[taxonomy.restBase]?.[cleanForSlug(termName)];
+      const cachedTerm =
+        termCache[taxonomy.restBase]?.[data?.slug || cleanForSlug(termName)];
       if (cachedTerm) {
         if (addToSelection) {
           addTermToSelection(taxonomy, cachedTerm);
@@ -208,11 +215,6 @@ function TaxonomyProvider(props) {
 
       // create term and add to cache
       try {
-        const data = { name: termName };
-        if (parentId) {
-          data.parent = parentId;
-        }
-
         const newTerm = await createTaxonomyTerm(termsEndpoint, data);
         const incomingCache = {
           [taxonomy.restBase]: { [newTerm.slug]: newTerm },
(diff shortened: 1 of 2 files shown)


packages/story-editor/src/app/taxonomy/taxonomyProvider.js
```diff
@@ -190,9 +190,16 @@ function TaxonomyProvider(props) {
...
+    async (taxonomy, termName, parent, addToSelection = false) => {
+      const data = { name: termName };
+      if (parent?.id) {
+        data.parent = parent.id;
+        data.slug = `${parent.slug}-${cleanForSlug(data.name)}`;
+      }
+
       // make sure the term doesn't already exist locally
-      const cachedTerm = termCache[taxonomy.restBase]?.[cleanForSlug(termName)];
+      const preEmptiveSlug = data?.slug || cleanForSlug(termName);
```
BrittanyIRL, line 201: nit: preemptive is one word, so probably just `preemptiveSlug`

BrittanyIRL (changes requested):
🤔  Testing this, I can make duplicative child categories when parents are unique, however I'm unable to add a new parent category with the same name as a category that is a child. 

For instance: 
geology > rock 
music > rock 
☹️  rock ☹️

maxyinger:
@BrittanyIRL 
> 🤔 Testing this, I can make duplicative child categories when parents are unique, however I'm unable to add a new parent category with the same name as a category that is a child.
> 
> For instance: geology > rock music > rock ☹️ rock ☹️


Um make sure to clear all your old categories before trying this out. If you had a category like `geology -> rock` before the update, that would have the `rock` slug and block this from properly working.

Just added a lil note to the testing instructions. Thanks for pointing out! Hopefully this was the issue 🤞

BrittanyIRL (approved):
Thanks for the clarification, didn't even think about the existing slugs! That was absolutely the issue!

samwhale (approved):
Nice this works great 🔥

PR #9303, opened 2021-10-06
Taxonomies: Enable e2e tests for contributor role

### LINKED ISSUE #9236: Taxonomy: Contributors have capability to create web story tags
Came up in #9212 and am isolating here since that ticket has expanded. 

Right now, contributors are given the ability to create new web story tags: 

Is this something we want? If yes, we can update the e2e test and call it a day, otherwise this ticket should be to address that config discrepancy. 

In posts, contributors cannot create new tags, just assign existing ones.

BrittanyIRL:
## Context

Enables the contributor tags e2e test now that we've cleared up that it's intentional that contributors are able to create tags. 

## Relevant Technical Choices

Also cleaned up some `expect` discrepancies from earlier tweaks so tag tests are set up the same.

## Testing Instructions

1.

## Checklist

---

Fixes #9236

### DIFF
--- a/packages/e2e-tests/src/specs/editor/taxonomy.js
+++ b/packages/e2e-tests/src/specs/editor/taxonomy.js
@@ -200,13 +200,14 @@ describe('taxonomy', () => {
       // See that added tags persist.
       const tokens2 = await page.evaluate(() =>
         Array.from(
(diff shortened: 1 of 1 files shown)


swissspidy (approved):
Awesome, thanks!

PR #9313, opened 2021-10-06
Fix Page Templates usage

### LINKED ISSUE #9294: Choosing Saved Templates in the Page Templates panel crashes the editor
## Expected Behaviour
Saved Templates should open instead of the editor crashing.

## Steps to Reproduce
1. Open the Page Templates tab
2. Choose "Saved Templates" from the dropdown
3. See the editor crashing

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

swissspidy:
## Context

It was reported by us and also by users that the "Saved Templates" function is broken in the last release due to [multiple bugs](https://github.com/google/web-stories-wp/issues/9294#issuecomment-936609703).

## Summary

Since this is actually three bugs in one, this PR contains some JavaScript changes but also some PHP changes.

In essence, it makes saving custom page templates work again.

See below for details.

## Relevant Technical Choices

**Note:** see also individual commits for a better understanding.

**Don't fetch page templates if `nextTemplatesToFetch` is `false`**

This is self-explanatory. Basically, this would lead to REST API requests being made with `?page=false`, which will be rejected by the API because `false` clearly is not an `integer`.

**Ensure `TemplateList` doesn't crash when `pages` is `null`**

This is self-explanatory. `pages` is always `null` at the beginning. And when REST API fetching fails, it will stay `null`. However, the component didn't support this and crashed when `pages` wasn't an array. So this adds some extra hardening.

**Always show `TemplateSave` button**

Again, self-explanatory. Always show the button - even if templates haven't been loaded yet.

**Improve service registration to ensure correct order**

This change allows us to clearly define dependencies between services using `HasRequirements`, without having to use different action priorities at the same time.

**Example:** `Settings`, `Story_Post_Type` and `Page_Template_Post_Type` are all delayed to be registered on `init`, with priority 10.

However, because `Story_Post_Type` depends on `Settings`, it will have priority 11 to ensure it runs after.

And because `Page_Template_Post_Type` depends on `Story_Post_Type`, it will have new priority 12 to ensure the `Story_Post_Type` service is properly registered first.

This, together with extended usage of `HasRequirements` throughout the code base makes dependency resolution more robust.

**Remove unnecessary abstraction for REST controllers**

`Autosaves_Controller` and `Lock_Controller` were both abstract classes, theoretically making it easier to implement them for multiple post types. However, there are no plans for ever using those for any post types other than web-story, so having abstract classes is overkill.

The code has been merged into `Stories_Autosaves_Controller` and `Stories_Lock_Controller`, respectively.

**E2E tests**

To prevent regressions, this PR adds an e2e test verifying that custom page templates can be loaded & saved.

## To-do

N/A

## User-facing changes

* The "Save page as template" button now doesn't have a hover effect when disabled
* The "Saved Templates" panel is no longer broken

## Testing Instructions

1. Create a new story
1. Go to page templates tab
2. Verify that the "Save page as template" button is disabled
3. Add an element to the page
4. Verify that the button is no longer disabled and allows you to create a new page template

No

No

No

## Checklist

---

Fixes #9294


BrittanyIRL (commented):
Running this locally after a fresh install and I think something is missing for the `POST` route when trying to save a current page as a template. I'm getting a 404 response. Post route looks like this: `http://localhost:8899/wp-json/web-stories/v1/web-story-page//?_locale=user` so that 404 makes sense :) 

Also tagging Miina since Prometheus implemented this originally and she should have context for this update as well.

swissspidy:
> Running this locally after a fresh install and I think something is missing for the `POST` route when trying to save a current page as a template. I'm getting a 404 response. Post route looks like this: `http://localhost:8899/wp-json/web-stories/v1/web-story-page//?_locale=user` so that 404 makes sense :)
> 
> Also tagging Miina since Prometheus implemented this originally and she should have context for this update as well.

Yeah one of my later refactorings broke it again, re-fixing it right now.

packages/story-editor/src/components/library/panes/pageTemplates/templateSave.js
```diff
@@ -143,21 +144,17 @@ function TemplateSave({ setShowDefaultTemplates, updateList }) {
     ]
   );
 
-  const textId = useMemo(() => `template_save_btn_${uuidv4()}`, []);
   return (
     <SaveButton
       aria-disabled={isDisabled}
       onClick={handleSaveTemplate}
-      aria-labelledby={textId}
```
miina, line 151: Curious why switch from using labelled by to `aria-label`?

packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
```diff
@@ -151,14 +162,12 @@ function PageTemplatesPane(props) {
...
         <>
-          {savedTemplates && (
-            <ButtonWrapper>
-              <TemplateSave
-                setShowDefaultTemplates={setShowDefaultTemplates}
-                updateList={updateTemplatesList}
-              />
-            </ButtonWrapper>
-          )}
+          <ButtonWrapper>
```
miina, line 165: I think that clicking on the button before the `savedTemplates` has been loaded will result in a failure, that's why it was hidden originally. Perhaps we should just disable it then? Since adding a new template relies on the array existing.

E.g. if I quickly click on the button after switching to the tab, I get this error:
Screenshot 2021-10-07 at 12 21 26

Trying again works though and it's unlikely that someone would click on it very quickly, but it might happen in case of slow connection. Thoughts?

miina (approved):
Had 2 questions, generally LGTM and works well, too (didn't review PHP).

miina:
@swissspidy FYI I'm seeing an unrelated console error when actually adding a new page template with images:

Screenshot 2021-10-07 at 12 31 20
Screenshot 2021-10-07 at 12 30 50

Is this something you remember seeing before, or needs a new issue?

packages/story-editor/src/components/library/panes/pageTemplates/templateSave.js
swissspidy: This is a `button` element, for which its value already is the label.

There is simply no need to use `aria-labelledby` for a button.

Probably don't even need `aria-label` and just can use the label below, but then we'd probably need `aria-hidden` on the `svg` since its decorative.

packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
swissspidy: Ah, that's a good point. Yeah disabling when it's still null and not an array makes sense.

includes/REST_API/Autosaves_Controller.php
```diff
@@ -1,221 +0,0 @@
-<?php
```
spacedmonkey, line 1: Why is this class being removed?

includes/REST_API/Publisher_Logos_Controller.php
```diff
@@ -50,18 +51,40 @@ class Publisher_Logos_Controller extends REST_Controller {
...
+	 * Get the list of service IDs required for this service to be registered.
+	 *
+	 * Needed because the story post type needs to be registered first.
+	 *
+	 * @since 1.13.0
+	 *
+	 * @return string[] List of required services.
+	 */
+	public static function get_requirements(): array {
+		return [ 'story_post_type' ];
```
spacedmonkey:
```suggestion
		return [ 'settings', 'story_post_type' ];
```

includes/REST_API/Stories_Lock_Controller.php
```diff
@@ -24,22 +24,380 @@
...
+
+	/**
+	 * Delete post lock
+	 *
+	 * @since 1.6.0
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 * @return WP_REST_Response Response object on success.
+	 */
+	public function delete_item( $request ): WP_REST_Response {
```
spacedmonkey, line 177:
```suggestion
	public function delete_item( WP_REST_Request  $request ): WP_REST_Response {
```

includes/REST_API/Stories_Lock_Controller.php
```diff
@@ -24,22 +24,380 @@
...
+		return $response;
+	}
+
+	/**
+	 * Get the lock, if the ID is valid.
+	 *
+	 * @param int $post_id Supplied ID.
+	 * @return array|false Lock as string or default to false.
+	 */
+	protected function get_lock( $post_id ) {
```
spacedmonkey:
```suggestion
	protected function get_lock( int $post_id ) {
```

includes/REST_API/Autosaves_Controller.php
swissspidy: See PR description.

includes/REST_API/Stories_Lock_Controller.php
swissspidy: FWIW this is not newly introduced code, just moved code. But yeah I suppose we can add this.

includes/Infrastructure/ServiceBasedPlugin.php
```diff
@@ -475,16 +541,18 @@ protected function get_identifier_from_fqcn( $fqcn ): string {
...
 				$this->maybe_register_service( $id, $class );
 			} else {
 				add_action(
-					$registration_action,
+					$action,
 					function () use ( $id, $class ) {
 						$this->maybe_register_service( $id, $class );
-					}
+					},
+					$priority
```
spacedmonkey, line 555: I wonder why this was missing?

includes/REST_API/Stories_Lock_Controller.php
swissspidy: This is not possible ⚠️ 

This is incompatible with the method declaration in the parent class.

spacedmonkey:
Created a follow up issue from this PR, as it opens up some great options https://github.com/google/web-stories-wp/issues/9316

includes/REST_API/Stories_Lock_Controller.php
spacedmonkey: 😢

includes/Infrastructure/ServiceBasedPlugin.php
```diff
@@ -284,7 +284,6 @@ public function register_services() {
 				&&
 				! $this->requirements_are_met( $id, $class, $services )
 			) {
-				next( $services );
```
spacedmonkey, line 287: Why was this line removed?

tests/phpunit/integration/tests/REST_API/Publisher_Logos_Controller.php
```diff
@@ -90,7 +91,10 @@ public static function wpSetUpBeforeClass( $factory ) {
 	public function set_up() {
 		parent::set_up();
 
-		$this->controller = new \Google\Web_Stories\REST_API\Publisher_Logos_Controller( new Settings() );
+		$this->controller = new \Google\Web_Stories\REST_API\Publisher_Logos_Controller(
+			new Settings(),
+			new \Google\Web_Stories\Story_Post_Type( new Settings() )
+		);
```
spacedmonkey:
```suggestion
		$settings         = new Settings();
		$this->controller = new \Google\Web_Stories\REST_API\Publisher_Logos_Controller(
			$settings,
			new \Google\Web_Stories\Story_Post_Type( $settings )
		);
```

includes/Infrastructure/ServiceBasedPlugin.php
swissspidy: See https://github.com/ampproject/amp-wp/pull/6529 and also https://github.com/ampproject/amp-wp/pull/6529

spacedmonkey:
Question: Would remove the delayed interface on some of the services help here? For example, not delay the settings service at all.

swissspidy:
> Question: Would remove the delayed interface on some of the services help here? For example, not delay the settings service at all.

I don't think it would help. Especially not with this current PR.

If you remove `Delayed`, `Settings::register()` would need to be updated to only register options on `init` (with some priority, e.g. 10).

That does mean the `Story_Post_Type` class for instance gets a valid `Settings` instance which has been registered, but the _options_ might not have necessarily been registered yet (because `Story_Post_Type` could be hooked to a higher, prio, e.g. 5.

So that would actually be quite dangerous.

swissspidy:
> @swissspidy FYI I'm seeing an unrelated console error when actually adding a new page template with images:
> 
> Screenshot 2021-10-07 at 12 31 20
> 
> Screenshot 2021-10-07 at 12 30 50
> 
> Is this something you remember seeing before, or needs a new issue?

This needs a separate ticket issue for sure. I do remember talking about adding proper `catch` to `preloadImage` in the past.

packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
swissspidy: Found a better solution, just fixing `updateTemplatesList`

BrittanyIRL (approved):
working for me - js looks good, you don't want me to review the PHP. 😅

miina:
> This needs a separate ticket issue for sure. I do remember talking about adding proper catch to preloadImage in the past.

Added https://github.com/google/web-stories-wp/issues/9326

includes/Infrastructure/ServiceBasedPlugin.php
swissspidy: The whole priority thing seemed like not fully implemented when I worked on this PR. But alas 🤷

spacedmonkey (approved):

PR #9301, opened 2021-10-06
Prefix custom REST API fields

spacedmonkey:
## Summary

To prevent this incompatibility, all our REST API fields that we register should ideally be prefixed with web_stories.

## Testing Instructions

- [ ] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

To reproduce the issue I was debugging:

1. Activate the official Unsplash plugin
2. Activate the Web Stories plugin
3. Try to upload media in the editor

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [x] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [x] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #9290

PR #9315, opened 2021-10-07
Video: Allow re-trimming of already trimmed video

### LINKED ISSUE #8892: Re-trimming a video

barklund:
## Context

This implements re-trimming an already trimmed video by abstracting out which video is being trimmed completely from most components.

## To-do

Some things that could be added, but don't have to be:

* [ ] Caching the loaded media and checking if we might even already have it in the media library.
* [ ] Agree what the custom poster should be (should it be taken from the video being re-trimmed or the original video?)

## User-facing changes

Trimming a video, that has already been trimmed, now shows the full original video with the handles placed correctly:
retrim

## Testing Instructions

1. Add an original (not trimmed) video to a page
2. Click to enter trim mode and observe, that the handles start at the far end for the trim rail
3. Move both handles in and press to trim the video
4. When trimming is fully completed press to trim the new video again
5. Observe that the trim handles are now placed exactly as before with the same indents made before pressing the trim button
6. Move the trim handles to completely new positions and press trim to re-trim the video
7. Once trimming has completed, observe that this third video is correctly trimmed to the desired interval (optionally by entering trim mode to see the handle positions and play the loop)
8. Insert another copy of the newly trimmed video on the canvas
9. Press to enter trim mode and observe the same trim position is still the case

## Checklist

---

Fixes #8892

### DIFF
--- a/packages/story-editor/src/app/api/apiProvider.js
+++ b/packages/story-editor/src/app/api/apiProvider.js
@@ -53,6 +53,7 @@ function APIProvider({ children }) {
     saveStoryById,
     autoSaveById,
     getMedia,
+    getMediaById,
     uploadMedia,
     updateMedia,
     deleteMedia,
@@ -117,6 +118,11 @@ function APIProvider({ children }) {
     ({ mediaType, searchTerm, pagingNum, cacheBust }) =>
       getMedia({ mediaType, searchTerm, pagingNum, cacheBust }, media),
     [media, getMedia]
+  );
+
+  actions.getMediaById = useCallback(
+    (mediaId) => getMediaById(mediaId, media),
+    [getMediaById, media]
   );
 
   actions.uploadMedia = useCallback(
--- a/packages/story-editor/src/app/media/utils/useProcessMedia.js
+++ b/packages/story-editor/src/app/media/utils/useProcessMedia.js
@@ -59,17 +59,11 @@ function useProcessMedia({
   );
 
   const updateExistingElements = useCallback(
-    ({ oldResource }) => {
-      const { id } = oldResource;
+    ({ oldResource: resource }) => {
+      const { id } = resource;
       updateElementsByResourceId({
         id,
-        properties: () => {
-          return {
-            resource: {
-              ...oldResource,
-            },
-          };
-        },
+        properties: () => ({ resource }),
       });
     },
     [updateElementsByResourceId]
@@ -186,11 +180,11 @@ function useProcessMedia({
    * @param {string} end Time stamp of end time of new video. Example '00:02:00'.
    */
   const trimExistingVideo = useCallback(
-    ({ resource: oldResource, start, end }) => {
+    ({ resource: oldResource, trimSourceId, start, end }) => {
       const { src: url, mimeType, poster } = oldResource;
 
       const trimData = {
-        original: oldResource.id,
+        original: trimSourceId,
         start,
         end,
       };
--- a/packages/story-editor/src/app/media/utils/useProcessMedia.js
+++ b/packages/story-editor/src/app/media/utils/useProcessMedia.js
@@ -180,11 +180,13 @@ function useProcessMedia({
    * @param {string} end Time stamp of end time of new video. Example '00:02:00'.
    */
   const trimExistingVideo = useCallback(
-    ({ resource: oldResource, trimSourceId, start, end }) => {
-      const { src: url, mimeType, poster } = oldResource;
+    ({ resource: oldResource, canvasResourceId, start, end }) => {
+      const { id, src: url, mimeType, poster } = oldResource;
+
+      const canvasResource = { ...oldResource, id: canvasResourceId };
 
       const trimData = {
-        original: trimSourceId,
+        original: id,
         start,
         end,
       };
@@ -192,7 +194,7 @@ function useProcessMedia({
       const onUploadStart = () => {
         updateExistingElements({
           oldResource: {
-            ...oldResource,
+            ...canvasResource,
             trimData,
             isTrimming: true,
           },
@@ -201,7 +203,10 @@ function useProcessMedia({
 
       const onUploadError = () => {
         updateExistingElements({
-          oldResource: { ...oldResource, isTrimming: false },
+          oldResource: {
+            ...canvasResource,
+            isTrimming: false,
+          },
         });
       };
 
@@ -218,9 +223,9 @@ function useProcessMedia({
       };
 
       const onUploadProgress = ({ resource }) => {
-        const oldResourceWithId = { ...resource, id: oldResource.id };
+        const newResourceWithCanvasId = { ...resource, id: canvasResourceId };
         updateExistingElements({
-          oldResource: oldResourceWithId,
+          oldResource: newResourceWithCanvasId,
         });
       };
 
--- a/packages/story-editor/src/components/videoTrim/provider.js
+++ b/packages/story-editor/src/components/videoTrim/provider.js
@@ -50,7 +50,7 @@ function VideoTrimProvider({ children }) {
   } = useVideoNode(videoData);
 
   const performTrim = useCallback(() => {
-    const { resource } = videoData;
+    const { resource, element } = videoData;
     if (!resource) {
       return;
     }
@@ -58,9 +58,13 @@ function VideoTrimProvider({ children }) {
     trimExistingVideo({
       resource: {
         ...resource,
+        // Preserve the id of the source resource even though we might be trimming from a third resource
+        id: element.resource.id,
         length: lengthInSeconds,
         lengthFormatted: getVideoLengthDisplay(lengthInSeconds),
       },
+      // However, also keep the trim source id available (it might be identical)
+      trimSourceId: resource.id,
       start: formatMsToHMS(startOffset),
       end: formatMsToHMS(endOffset),
     });
--- a/packages/story-editor/src/components/videoTrim/provider.js
+++ b/packages/story-editor/src/components/videoTrim/provider.js
@@ -25,19 +25,17 @@ import { trackEvent } from '@web-stories-wp/tracking';
 /**
  * Internal dependencies
  */
-import { useLocalMedia, useStory } from '../../app';
+import { useLocalMedia } from '../../app';
 import VideoTrimContext from './videoTrimContext';
 import useVideoTrimMode from './useVideoTrimMode';
 import useVideoNode from './useVideoNode';
 
 function VideoTrimProvider({ children }) {
-  const { selectedElements } = useStory(({ state: { selectedElements } }) => ({
-    selectedElements,
-  }));
   const { trimExistingVideo } = useLocalMedia((state) => ({
     trimExistingVideo: state.actions.trimExistingVideo,
   }));
-  const { isTrimMode, hasTrimMode, toggleTrimMode } = useVideoTrimMode();
+  const { isTrimMode, hasTrimMode, toggleTrimMode, videoData } =
+    useVideoTrimMode();
   const {
     hasChanged,
     currentTime,
@@ -48,10 +46,10 @@ function VideoTrimProvider({ children }) {
     setEndOffset,
     setVideoNode,
     resetOffsets,
-  } = useVideoNode();
+  } = useVideoNode(videoData);
 
   const performTrim = useCallback(() => {
-    const { resource } = selectedElements[0];
+    const { resource } = videoData;
     if (!resource) {
       return;
     }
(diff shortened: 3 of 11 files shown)


miina:
@barklund I'm getting this error currently:
Screenshot 2021-10-07 at 11 28 44

With the following REST response (might be since there were some API changes in #8513, I think, but not sure) (cc @spacedmonkey since you implemented the API part of trimming):
```
{"code":"rest_invalid_param","message":"Invalid parameter(s): media_source","data":{"status":400,"params":{"media_source":"media_source[0] is not of type integer."},"details":{"media_source":{"code":"rest_invalid_type","message":"media_source[0] is not of type integer.","data":{"param":"media_source[0]"}}}}}
```

Also, I wasn't able to test it yet due to the previous error but just wanted to confirm that the same video can be trimmed more than once and the initial video is always used. Part of the logic in the original PR you had questions about (in terms of necessity) was for ensuring that the original video is used for the second and third etc. trim as well. Would be good to add that case to the testing instructions as well (trimming the video more than once and ensuring the same origin video is always used).

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
```diff
@@ -54,11 +59,39 @@ function useVideoTrimMode() {
...
+        // Load correct video resource and set original offsets
+        getMediaById(selectedElement.resource.trimData.original).then(
+          (originalResource) =>
+            setVideoData({
+              element: selectedElement,
+              resource: originalResource,
+              start: getMsFromHMS(selectedElement.resource.trimData.start),
+              end: getMsFromHMS(selectedElement.resource.trimData.end),
+            })
+        );
```
swissspidy: Perhaps add a `catch()` in case loading fails?

miina:
>  I'm getting this error currently:

Looks like the PR works in the QA environment so it seems like the error is just happening locally.
However, re-trimming the same video more than one time doesn't seem to work as expected when testing in QA. Steps:
1. Add a video
2. Re-trim the video to very short.
3. Now re-trim the video again to longer
4. Click "Trim" again. See that the handles are still showing trimmed short as trimmed in step 2.

spacedmonkey:
> With the following REST response (might be since there were some API changes in #8513, I think, but not sure) (cc @spacedmonkey since you implemented the API part of trimming):

I think this issue will be fixed in https://github.com/google/web-stories-wp/pull/9301.  Do you have the unsplash plugin installed out of interest?

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
```diff
@@ -54,11 +59,39 @@ function useVideoTrimMode() {
         hasEditMenu: true,
         showOverflow: false,
       });
+
+      if (selectedElement.resource?.trimData?.original) {
```
spacedmonkey: Might worth extracting trimData make this a little easier to read. 

```
const { resource: { trimData } } = selectedElement;
```

packages/wp-story-editor/src/api/media.js
```diff
@@ -69,6 +69,24 @@ export function getMedia(
...
+ */
+export function getMediaById(mediaId, media) {
+  const path = addQueryArgs(`${media}${mediaId}/`, {
+    context: 'edit',
+    _fields: MEDIA_FIELDS,
+  });
+
+  return apiFetch({ path }).then((attachment) =>
+    getResourceFromAttachment(attachment)
+  );
```
spacedmonkey:
```suggestion
  return apiFetch({ path }).then(getResourceFromAttachment);
```

miina:
> I think this issue will be fixed in #9301. Do you have the unsplash plugin installed out of interest?

Hm, I did have it installed and enabled, didn't even remember. After disabling it there's no error locally anymore.

packages/wp-story-editor/src/api/media.js
swissspidy: Should be OK in this case, but this is a good reminder to be careful about this: https://jakearchibald.com/2021/function-callback-risks/

spacedmonkey: It is also done here 

https://github.com/google/web-stories-wp/blob/81f69b13ac8573f493a92269e12f0571cee6f8d2/packages/wp-story-editor/src/api/media.js#L93

spacedmonkey:
Seeing something weird. I trimmed a video to 1 second. Then retrimmed to 5 seconds. After the upload is complete, I am still seeing the 1 second video when I play it. 

https://user-images.githubusercontent.com/237508/136385476-1c1d38a1-91b5-457e-9dc2-3898670f3903.mov

Am I missing something here?

miina:
> Seeing something weird. I trimmed a video to 1 second. Then retrimmed to 5 seconds. After the upload is complete, I am still seeing the 1 second video when I play it.

@spacedmonkey This is the same bug as reported here: https://github.com/google/web-stories-wp/pull/9315#issuecomment-937633569

barklund:
> > Seeing something weird. I trimmed a video to 1 second. Then retrimmed to 5 seconds. After the upload is complete, I am still seeing the 1 second video when I play it.
> 
> @spacedmonkey This is the same bug as reported here: [#9315 (comment)](https://github.com/google/web-stories-wp/pull/9315#issuecomment-937633569)

I'm looking into this now. There's definitely a bug here somewhere, probably something trivial.

packages/wp-story-editor/src/api/media.js
barklund: Fixes in a76495ec8c.

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
barklund: Fixed in a76495ec8c.

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
barklund: Fixed in a76495ec8c.

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -186,11 +180,11 @@ function useProcessMedia({
    * @param {string} end Time stamp of end time of new video. Example '00:02:00'.
    */
   const trimExistingVideo = useCallback(
-    ({ resource: oldResource, start, end }) => {
+    ({ resource: oldResource, trimSourceId, start, end }) => {
       const { src: url, mimeType, poster } = oldResource;
 
       const trimData = {
-        original: oldResource.id,
+        original: trimSourceId,
```
barklund: This fixed the bug mentioned about actually retrimming a video. It's not super elegant, but it works.

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -59,17 +59,11 @@ function useProcessMedia({
...
       updateElementsByResourceId({
         id,
-        properties: () => {
-          return {
-            resource: {
-              ...oldResource,
-            },
-          };
-        },
+        properties: () => ({ resource }),
```
barklund, line 66: This is just a minor optimization of this piece of code. Should be completely identical to the original (except for spreading the resource, but I don't see how that should have an impact, as the resource is inline created every time this function is called anyway).

packages/story-editor/src/components/videoTrim/provider.js
```diff
@@ -49,20 +47,24 @@ function VideoTrimProvider({ children }) {
...
     trimExistingVideo({
       resource: {
         ...resource,
+        // Preserve the id of the source resource even though we might be trimming from a third resource
+        id: element.resource.id,
         length: lengthInSeconds,
         lengthFormatted: getVideoLengthDisplay(lengthInSeconds),
       },
+      // However, also keep the trim source id available (it might be identical)
+      trimSourceId: resource.id,
```
barklund: Again, not super pretty, but fixes the bug. Does it need more comments? 😄

packages/story-editor/src/components/videoTrim/provider.js
```diff
@@ -50,17 +50,21 @@ function VideoTrimProvider({ children }) {
...
+    const { resource, element } = videoData;
     if (!resource) {
       return;
     }
     const lengthInSeconds = Math.round(endOffset / 1000 - startOffset / 1000);
     trimExistingVideo({
       resource: {
         ...resource,
+        // Preserve the id of the source resource even though we might be trimming from a third resource
+        id: element.resource.id,
```
spacedmonkey: Can you explain this a little more.

barklund: Yeah, it's a bit messy, but we need both IDs in the media processing function.

This ID here is the resource from which to copy the information over to the new element. This is considered the source for all data except the actual raw video. It is used in `updateExistingElements` many times. It is also the element that is temporarily marked as being "trimmed", though in effect it is not necessarily this element's video that is being trimmed.

The other ID listed a few lines below, `trimSourceId`, is to be considered the ID of the resource of the raw video, that is to be processed into a new video. It is used as the `original` key in the trim data for the newly create resource.

I would be good to clean this up a bit more, but I'm not really sure how to do that...

spacedmonkey: Different naming might be useful.

packages/story-editor/src/app/api/apiProvider.js
```diff
@@ -119,6 +120,11 @@ function APIProvider({ children }) {
     [media, getMedia]
   );
 
+  actions.getMediaById = useCallback(
+    (mediaId) => getMediaById(mediaId, media),
+    [getMediaById, media]
+  );
```
spacedmonkey, line 126: @sayedtaqui  Flagging will need to taken in account for https://github.com/google/web-stories-wp/pull/9318

packages/story-editor/src/app/media/utils/useProcessMedia.js
spacedmonkey: To be clear, this saves the orignal source video id.

packages/story-editor/src/components/videoTrim/provider.js
barklund: I reorganized it a bit in 9afe07f158 - does that make it clearer? I separate it between the resource and the canvas resource. The former is the actual resource where working on, where the canvas resource is the element on the canvas that should be cloned into a new element with the resulting video.

packages/story-editor/src/app/media/utils/useProcessMedia.js
barklund: This has been changed now.

spacedmonkey:
What if you do this?

- Trim 1 minute to 30 seconds. 
- Mute new 30 second video. 
- Retrim video to 45 seconds. 

Would the 45 second video be muted? Should it be trimmed?

swissspidy:
> What if you do this?
> 
> * Trim 1 minute to 30 seconds.
> * Mute new 30 second video.
> * Retrim video to 45 seconds.
> 
> Would the 45 second video be muted? Should it be trimmed?

Good question!

Sounds like a product question that is best resolved in a follow-up ticket.

As a user I would expect it to be still muted. And if it was optimized before, it should be still be optimized. Everything else would be super confusing.

The question is how we can achieve this? Sounds like we'd need to refactor the media upload queue and perhaps even useFFmpeg quite a bit to basically do trimming & muting in one step... Hence best done in a follow-up.

barklund:
> What if you do this?
> 
> * Trim 1 minute to 30 seconds.
> * Mute new 30 second video.
> * Retrim video to 45 seconds.
> 
> Would the 45 second video be muted? Should it be trimmed?

Yeah, I was thinking about that too somewhere along the way.

All of this would be trivial if we moved all video transcoding to a "rendering" step just before publishing/updating, but that's not where we are right now. In our current setup, this is a bit tricky.

But let's file that as a follow-up ticket.

miina (approved):
LGTM!

One potential improvement:
If the user is trying to re-trim the video to the same length that the original video is, perhaps there's no need to create a new video and just the old video can be used? E.g. the user re-trims a 30s video to 5s and then back to 30s, currently it creates a new 30s video.

barklund:
> LGTM!
> 
> One potential improvement: If the user is trying to re-trim the video to the same length that the original video is, perhaps there's no need to create a new video and just the old video can be used? E.g. the user re-trims a 30s video to 5s and then back to 30s, currently it creates a new 30s video.

Yes, in general if you create a trimmed video, of which we already have an identical copy in the library, it would make sense to re-use that rather than make a new one.

But your example is of course even more simple, that if you want to return to the original, we rather than reuse said original instead just make a duplicate, which is pretty stupid.

I'll file as a follow-up.

barklund:
@swissspidy, what do you say about either caching the API response for the get media request here or making a very lightweight API request for just the video source URL, as I think that's the only thing we need here?

Currently, the loading time on my setup is about 0.5-1.0 seconds, which is pretty long to wait every time I press to re-trim a video. Or is it just my local virtual environment, that's slow?

I'd file it as a follow-up enhancement, but not sure how to best file it, as I don't know what would be the recommended path forward?

swissspidy:
@barklund If you only need the source URL and nothing else, you can use the `?_fields` parameter to only fetch the fields you'd like. e.g. `/wp-json/wp/v2/media/123456?_fields=source_url`

PR #9318, opened 2021-10-07
Decoupling: Refactor story editor callback handling

mohdsayed:
## Context

Further improvements of https://github.com/google/web-stories-wp/pull/8966

## Summary

Do not pass API paths from `wp-story-editor` to `story-editor` and then again to the API callbacks. Instead, bind `config` to all API callbacks exposed to the core editor.

## Testing Instructions

1.

## Checklist

---


packages/story-editor/src/app/api/apiProvider.js
```diff
@@ -29,54 +29,15 @@ import Context from './context';
...
+  const { apiCallbacks, cdnURL } = useConfig();
   const pageTemplates = useRef({
     base: [],
     withoutImages: [],
   });
 
-  const actions = { getTaxonomyTerm, createTaxonomyTerm };
+  const actions = {
+    ...apiCallbacks,
+  };
```
spacedmonkey: Would not just be easier to do this.
```
const { apiCallbacks: actions, cdnURL } = useConfig();
```

mohdsayed: Maybe later but currently `apiCallbacks` is being used on line#66 in the loop.

mohdsayed: Removed the loop for now in https://github.com/google/web-stories-wp/pull/9318/commits/b4d904b34c6da46aed0c78159152e11db70a4062 as it was anyways incomplete.

packages/story-editor/src/app/api/apiProvider.js
```diff
@@ -29,55 +29,12 @@ import Context from './context';
...
-    getCustomPageTemplates,
-    addPageTemplate,
-    deletePageTemplate,
-    getHotlinkInfo,
-    getTaxonomies,
-    getTaxonomyTerm,
-    createTaxonomyTerm,
-  } = apiCallbacks;
-
+  const { apiCallbacks: actions, cdnURL } = useConfig();
```
swissspidy, line 32: 😻 💯

packages/wp-story-editor/src/api/utils/getApiCallbacks.js
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Bind configuration arguments and get all api callbacks.
+ *
+ * @param {Object} config Editor configuration.
+ * @param {Object} config.api API paths.
+ * @param {boolean} config.encodeMarkup Encode markup or not.
+ * @param {string} config.postType Post type.
+ * @return {Object} api callbacks.
+ */
+const getApiCallbacks = ({ api, encodeMarkup, postType }) => {
```
swissspidy: The downside of this is that every time we add a new API method, we need to update `getApiCallbacks` as well. Can we avoid that?

Maybe something like this (untested):

```js
import * as taxonomiesApi from './api/taxonomies';
import * as mediaApi from './api/media';

Object.entries(taxonomiesApi).reduce((callbacks, [name, callback]) => {
  callbacks[name] = callback.bind(null, api.taxonomies)
}, {});

Object.entries(mediaApi).reduce((callbacks, [name, callback]) => {
  callbacks[name] = callback.bind(null, api.media)
}, {});
```

mohdsayed: Correct. We also need to pass `encodeMarkup`, `postType` in some callbacks and may have to pass other params in the future, so we would end up with multiple conditions with callback names.
Also, we are currently not passing anything to `getTaxonomyTerm` and `createTaxonomyTerm` because we still need to think of what needs to be done with taxonomies.
https://github.com/google/web-stories-wp/blob/efb741ddd4e916324115df08cb3bc2e3c3dd8a91/packages/story-editor/src/app/taxonomy/taxonomyProvider.js#L204-L216

Not very clean but the other way may be that we pass `config.api` object instead of individual API keys like this.

https://github.com/google/web-stories-wp/blob/47690102441dd7f5df651e1953c9ece0921adc92/packages/wp-story-editor/src/api/utils/getApiCallbacks.js#L30-L54

and then use it like this

https://github.com/google/web-stories-wp/blob/47690102441dd7f5df651e1953c9ece0921adc92/packages/wp-story-editor/src/api/taxonomy.js#L38-L40

thoughts?

swissspidy: Hmm good points! That would make `getApiCallbacks` quite complex and still require manual changes whenever a new callback is added.

At that point it's probably easier if we just pass the whole `config` object as the first parameter

.github/workflows/tests-karma-editor.yml
```diff
@@ -98,4 +98,3 @@ jobs:
           github.event.pull_request.draft == false &&
           github.event.pull_request.head.repo.fork == false &&
           github.event.pull_request.user.login != 'dependabot[bot]'
-
```
mohdsayed, line 101: Automatically done in merging `main` branch, probably by the auto fixer 🤔  https://github.com/google/web-stories-wp/pull/9318/commits/ba4c781d827aeb3d8fb90df86ebde8b07f512aa6

packages/wp-story-editor/src/api/utils/getApiCallbacks.js
mohdsayed: Done in https://github.com/google/web-stories-wp/pull/9318/commits/aef509c6c61c3b9a2205afde5c2c55dfbf529eb7

.github/workflows/tests-karma-editor.yml
swissspidy: Yeah looks like there was an errant newline there.

mohdsayed:
@swissspidy Need some help here with e2e, cannot figure out why it's failing. As before, e2e is failing for me even on `main` branch locally, so I get no clue of what to debug.

mohdsayed:
Fixed the issue in https://github.com/google/web-stories-wp/pull/9318/commits/8767e48480f2b6f0913c6a0d5b57444a5357a812 , e2e is passing now.

swissspidy (approved):

PR #9321, opened 2021-10-07
Taxonomies: Add Live Regions

### LINKED ISSUE #9283: Taxonomies: Show/Confirm to user when a new Category was created
## Feature Description

Keyboard Accessibility Feature request. When submitting a new category with the keyboard it would be nice to show the user that the category is created. Maybe we could use a snackbar or somehow add it to the top of the stack?

BrittanyIRL:
## Context

Missed some live regions found during bug bash as nice to haves for taxonomy feature. 

## Summary

Adds in a live region to adding new hierarchical (category) terms. 
Adds in a live region to freeform (tag) autocomplete list results. 

## Relevant Technical Choices

List results mimics that which already existed in hierarchical.
Set the debounced callback to `leading: true` so that it would be [triggered immediately the first time around](https://github.com/xnimorz/use-debounce#leadingtrailing-calls). 
Removed a redundant state for autocomplete results showing.

## User-facing changes

If a screen reader is in use and the user creates a new category they will hear "Category added." announced by the SR. 
If a screen reader is in use and the user types in a value on the tag input that expands the autocomplete list they will hear the result count of options announced by the SR. 

## Testing Instructions

Turn on a SR (I use this one: https://chrome.google.com/webstore/detail/screen-reader/kgejglhpjiefppelpmljglcjbhoiplfn) and try the above.

1.

## Checklist

---

Fixes #9283

### DIFF
--- a/packages/story-editor/src/components/form/tags/input.js
+++ b/packages/story-editor/src/components/form/tags/input.js
@@ -108,6 +108,7 @@ function Input({
 
   // isInputFocused is used to update the styled state of the input area.
   const [isInputFocused, setIsInputFocused] = useState(false);
+  const [isSuggestionsOpen, setIsSuggestionsOpen] = useState(false);
 
   // inputRef is used to return focus to input after keydown actions to avoid focused state being hijacked.
   const inputRef = useRef();
@@ -120,7 +121,11 @@ function Input({
 
   const suggestionListId = uuidv4();
   const totalSuggestions = suggestedTerms.length;
-  const isSuggestionsOpen = totalSuggestions > 0;
+  // suggestedTerms is returned from a debounced callback,
+  // we want to update isSuggestionsOpen via a useEffect to prevent sluggish closing.
+  useEffect(() => {
+    setIsSuggestionsOpen(totalSuggestions > 0);
+  }, [totalSuggestions]);
 
   // Allow parents to pass onTagsChange callback
   // that updates as tags does.
@@ -192,6 +197,7 @@ function Input({
 
   const handleTagSelectedFromSuggestions = useCallback((e, selectedValue) => {
     e.preventDefault();
+    setIsSuggestionsOpen(false);
     dispatch({ type: ACTIONS.SUBMIT_VALUE, payload: selectedValue });
     inputRef?.current.focus();
   }, []);
--- a/packages/story-editor/src/components/form/tags/input.js
+++ b/packages/story-editor/src/components/form/tags/input.js
@@ -108,7 +108,6 @@ function Input({
 
   // isInputFocused is used to update the styled state of the input area.
   const [isInputFocused, setIsInputFocused] = useState(false);
-  const [isSuggestionsOpen, setIsSuggestionsOpen] = useState(false);
 
   // inputRef is used to return focus to input after keydown actions to avoid focused state being hijacked.
   const inputRef = useRef();
@@ -121,9 +120,7 @@ function Input({
 
   const suggestionListId = uuidv4();
   const totalSuggestions = suggestedTerms.length;
-  useEffect(() => {
-    setIsSuggestionsOpen(totalSuggestions > 0);
-  }, [totalSuggestions]);
+  const isSuggestionsOpen = totalSuggestions > 0;
 
   // Allow parents to pass onTagsChange callback
   // that updates as tags does.
@@ -195,7 +192,6 @@ function Input({
 
   const handleTagSelectedFromSuggestions = useCallback((e, selectedValue) => {
     e.preventDefault();
-    setIsSuggestionsOpen(false);
     dispatch({ type: ACTIONS.SUBMIT_VALUE, payload: selectedValue });
     inputRef?.current.focus();
   }, []);
@@ -222,7 +218,6 @@ function Input({
     },
     [handleTagSelectedFromSuggestions, totalSuggestions]
   );
-
   return (
     <Border ref={containerRef} isInputFocused={isInputFocused}>
       <InputWrapper>
--- a/packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
+++ b/packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
@@ -24,8 +24,9 @@ import {
   useState,
   useEffect,
 } from '@web-stories-wp/react';
-import { __ } from '@web-stories-wp/i18n';
+import { __, sprintf, _n } from '@web-stories-wp/i18n';
 import PropTypes from 'prop-types';
+import { useLiveRegion } from '@web-stories-wp/design-system';
 /**
  * Internal dependencies
  */
@@ -65,6 +66,8 @@ function FlatTermSelector({ taxonomy, canCreateTerms }) {
 
   const { undo } = useHistory(({ actions: { undo } }) => ({ undo }));
   const [searchResults, setSearchResults] = useState([]);
+  const speak = useLiveRegion('assertive');
+  const debouncedSpeak = useDebouncedCallback(speak, 500, { leading: true });
 
   const handleFreeformTermsChange = useCallback(
     (termNames) => {
@@ -116,6 +119,14 @@ function FlatTermSelector({ taxonomy, canCreateTerms }) {
       per_page: 20,
     });
     setSearchResults(results);
+
+    const count = results.length;
+    const message = sprintf(
+      /* Translators: %d: Number of results. */
+      _n('%d result found.', '%d results found.', 'web-stories', 'web-stories'),
+      count
+    );
+    debouncedSpeak(message);
   }, 300);
 
   const tokens = useMemo(() => {
--- a/packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
+++ b/packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
@@ -67,7 +67,6 @@ function FlatTermSelector({ taxonomy, canCreateTerms }) {
   const { undo } = useHistory(({ actions: { undo } }) => ({ undo }));
   const [searchResults, setSearchResults] = useState([]);
   const speak = useLiveRegion('assertive');
-  const debouncedSpeak = useDebouncedCallback(speak, 500, { leading: true });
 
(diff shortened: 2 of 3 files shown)


packages/story-editor/src/components/panels/document/taxonomies/HierarchicalTermSelector.js
```diff
@@ -189,9 +193,17 @@ function HierarchicalTermSelector({
       setShowAddNewCategory(false);
       resetInputs();
       setToggleFocus(showAddNewCategory);
+      debouncedSpeak(
+        sprintf(
+          /* Translators: %s: Taxonomy label name. */
+          __('%s added.', 'web-stories'),
+          taxonomy.labels.singular_name
+        )
+      );
```
samwhale: 🤔 Does this need to be debounced?

I feel like it doesn't need to be debounced, but rather it should just wait until the category is added, right?

---

If this is debounced too, it closes and reads the button again - then the screen reader says that the category has been added. Feels like it should be announced right away.

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
```diff
@@ -116,6 +119,14 @@ function FlatTermSelector({ taxonomy, canCreateTerms }) {
...
     });
     setSearchResults(results);
+
+    const count = results.length;
+    const message = sprintf(
+      /* Translators: %d: Number of results. */
+      _n('%d result found.', '%d results found.', 'web-stories', 'web-stories'),
+      count
+    );
+    debouncedSpeak(message);
```
samwhale: this one makes sense to debounce, but isn't the request that's fired off already debounced? if we announce this after the search is complete then I don't think it would need to be debounced

samwhale:
I think that we don't need the debounce for either 🤔

packages/story-editor/src/components/form/tags/input.js
```diff
@@ -121,9 +120,7 @@ function Input({
 
   const suggestionListId = uuidv4();
   const totalSuggestions = suggestedTerms.length;
-  useEffect(() => {
-    setIsSuggestionsOpen(totalSuggestions > 0);
-  }, [totalSuggestions]);
+  const isSuggestionsOpen = totalSuggestions > 0;
```
samwhale: Oh yeah! This is derived state, nice catch

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
```diff
@@ -116,6 +118,14 @@ function FlatTermSelector({ taxonomy, canCreateTerms }) {
...
     });
     setSearchResults(results);
+
+    const count = results.length;
+    const message = sprintf(
+      /* Translators: %d: Number of results. */
+      _n('%d result found.', '%d results found.', 'web-stories', 'web-stories'),
+      count
+    );
+    speak(message);
```
samwhale, line 128: 👍

packages/story-editor/src/components/panels/document/taxonomies/HierarchicalTermSelector.js
```diff
@@ -203,12 +205,20 @@ function HierarchicalTermSelector({
         slug: selectedParentSlug,
       };
       createTerm(taxonomy, newCategoryName, parentValue, true);
```
samwhale, line 191: 💭 can we await this so that we don't speak until the term is created? `createTerm` is an async fn so it should be gucci
```suggestion
      await createTerm(taxonomy, newCategoryName, parentValue, true);
```

packages/story-editor/src/components/panels/document/taxonomies/HierarchicalTermSelector.js
```diff
@@ -203,12 +205,20 @@ function HierarchicalTermSelector({
...
+      speak(
+        sprintf(
+          /* Translators: %s: Taxonomy label name. */
+          __('%s added.', 'web-stories'),
+          taxonomy.labels.singular_name
+        )
+      );
```
samwhale, line 198: With chromevox I'm hearing that the category was added after it re-focuses the `Add New Category` button. Wonder if we should automatically focus the checkbox that gets added?

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
samwhale, line 128: Do we want this for when we add a term too?

packages/story-editor/src/components/panels/document/taxonomies/HierarchicalTermSelector.js
BrittanyIRL: so i played with this a bit last week in terms of having it announce ahead of announcing the new focus on submit and nothing I tried got the order to change. I think it's because the live region is waiting for the next available moment to announce, given the refocus occurring AND that the new category is getting added and then announced, i think this actually makes sense. The outcome is the same regardless of awaiting `createTerm` or not, as is putting `speak` in `.then` of the `createTerm` or above the call to `createTerm`.

packages/story-editor/src/components/panels/document/taxonomies/HierarchicalTermSelector.js
BrittanyIRL: 🤔  so we chose to refocus the add new category button because it's more inline with Gutenberg. In post categories when you add a new one the 'add new' section doesn't toggle like ours does, and the input stays focused. If we want to change that then that feels like a design choice.  Having the toggle button focus again feels like the less jarring option for having your focus hijacked. Like, this way at least you still have your bearings navigationally. 

Happy to discuss more, but I do feel like updating the focus is a different issue than adding live regions.

packages/story-editor/src/components/panels/document/taxonomies/FlatTermSelector.js
BrittanyIRL: nope!  mimicking Gutenberg.

packages/story-editor/src/components/panels/document/taxonomies/HierarchicalTermSelector.js
samwhale: > the live region is waiting for the next available moment to announce, given the refocus occurring AND that the new category is getting added and then announced

So it will always wait, even with a super slow connection? Maybe it's an edge case, but the new category could get added much later on a very slow connection.

Feels like it's better to just make sure it's queued after the `createTerm` function finishes rather than after it's called and the request gets sent, no?

samwhale (approved):

packages/story-editor/src/components/panels/document/taxonomies/HierarchicalTermSelector.js
BrittanyIRL: yeah i think it's tied to the toggle happening and the refocus, so those two things getting announced are more important because it's affecting the available options in the panel whereas the added category announcement isn't impacting the immediate navigability. So if we wanna keep the `add new` toggle expanded until a user toggles it manually this would probably no longer be an issue but i feel like that toggle is nice?

packages/story-editor/src/components/form/tags/input.js
maxyinger: Ah I know this appear to be derived, but we have this as state because `totalSuggestions` is debounced. This helps the dropdown from being sluggish on close. It's why we have the `setIsSuggestionsOpen(false);` call in `handleTagSelectedFromSuggestions` so it immediately closes instead of waiting for the debounced timeout to update `totalSuggestions`.

I'm open to other approaches here, or adding some explanatory comments, but that's the context behind holding it as state.

packages/story-editor/src/components/panels/document/taxonomies/HierarchicalTermSelector.js
maxyinger: I'm not sure if I follow the flow completely here, but the `await` before `createTerm` is optional for if you wanna perform an action after we get a response. just because it is a promise doesn't necessarily mean we have to do anything with the resolved value.

I would think of the promise here more like an optional callback that you can hook into if you want, but it's not necessary

maxyinger (approved):
Looks good, but I would revert that `isSuggestionsOpen` back to state so that there's no delay when closing the dropdown and the app doesn't feel sluggish to users

packages/story-editor/src/components/form/tags/input.js
samwhale: Oops that's right! Let's add a comment there instead of refactoring this 👍

packages/story-editor/src/components/panels/document/taxonomies/HierarchicalTermSelector.js
samwhale: >So if we wanna keep the add new toggle expanded until a user toggles it manually this would probably no longer be an issue but i feel like that toggle is nice

I just noticed this morning with gutenberg: it keeps this UI open until the user closes it.



> I'm not sure if I follow the flow completely here, but the `await` before `createTerm` is optional for if you wanna perform an action after we get a response. just because it is a promise doesn't necessarily mean we have to do anything with the resolved value.
> 
> I would think of the promise here more like an optional callback that you can hook into if you want, but it's not necessary

Yeah, we don't need the resolved value - but rather to wait to announce "category added" until after the category has come back from the backend

packages/story-editor/src/components/form/tags/input.js
BrittanyIRL: great catch, updating and adding a note to prevent that getting lost again

packages/story-editor/src/components/panels/document/taxonomies/HierarchicalTermSelector.js
BrittanyIRL: 🤔  so the reason the live region feels delayed is because the DOM is updating and that + refocusing the toggle to avoid the focus getting hijacked by the canvas is taking priority because it tells the user their immediate context and the live region is just waiting until there's a silent moment to chime in. If this becomes a concern the solution is to leave the new category form visible until the user toggles it back closed, which feels like a new issue (should that be deemed desirable), so for now I think we're good.

Issue #4432, opened 2020-09-09
A11Y: Design > Layers


dreamofabear:
## Solution to be implemented

1. Hide entire layer panel for screen readers by using `aria-hidden="true"`
1. Display layer manipulation shortcuts in the keyboard shortcut dialog:
  a. <kbd>⌘</kbd><kbd>↑</kbd> - move element forward
  b. <kbd>⌘</kbd><kbd>↓</kbd> - move element backward
  c. <kbd>⌘</kbd><kbd>⇧</kbd><kbd>↑</kbd> - move element to front
  d. <kbd>⌘</kbd><kbd>⇧</kbd><kbd>↓</kbd> - move element to back
#### 1. Navigating into Layers section blocked after expansion.

- User enables VoiceOver while Layers is collapsed. 
- User expands Layers via the arrow toggle button. 
- Expected: User selects "next item" in screen reader and the layers list group is selected.
- Actual: User is unable to navigate beyond the toggle button into the list group of layers via "next item" VoiceOver navigation. 
  - User's VoiceOver navigation actually becomes trapped between the "Layers" label and the toggle button.

#### 2. Layer items in list group are announced twice.

- User enables VoiceOver while Layers is expanded.
- User navigates into the Layers list group.
- Expected: Each layer is announced once with its layer contents (e.g. text, image, video).
- Actual: Each layer is preceded by a blank group (e.g. "group 1 of 16" with 8 layers).

## Additional Context

- Plugin Version: 1.0.0-beta.2+fd68af7
- Operating System: macOS 10.15.6
- Browser: Chrome 85

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

bmattb:
@o-fernandez can you confirm that this should be prioritized by Prometheus for Stable in Sprint 37 right now

dreamofabear:
Re: (1), "keyboard focus trapped" can be worked around with the new keyboard shortcuts (#4454). Command+Shift+1/2/3 (macOS) switches focus between element/workspace/design panels.

dreamofabear:
We've received some pushback on the workaround for (1). 

@barklund WDYT about triggering a keyboard focus change to the Layer list-group after expand button is clicked?

barklund:
> > Re: (1), "keyboard focus trapped" can be worked around with the new keyboard shortcuts (#4454). Command+Shift+1/2/3 (macOS) switches focus between element/workspace/design panels.
[...]

PR #9322, opened 2021-10-07
Layer panel updates

### LINKED ISSUE #8931: Layers Panel: Confusing size changes
## Bug Description

Let’s say I have three pages:
Page 1: 10 elements
Page 2: 5 elements
Page 3: 10 elements

I go to page 1, manually expand the layers panel to accommodate all elements
I go to page 2, the layers panel shrinks to very small size
I go to page 3, the layers panel stays that small

There’s no way for me to say “Expand the layers panel to max possible height whenever I change pages”

## Additional Context

- Plugin Version: 1.11.0
- WordPress Version:
- Operating System:
- Browser:

_____________

# Issue Update Based on Design Feedback Sept/Oct 2021 

_Kept the above for historical context since this panel's been a bit, well, like the title says, confusing_ 

## Description 
Update the layers panel height and the behaviors of controlling the height to make it (🤞 ) less confusing to users and more of a useful tool. 

- Default layers panel height to 25% of the total screen height 
  - @agingoldseco can you confirm the percentage? ✅ 
- Unless the user adjusts the height manually, the panel will stay this default height regardless of the layer count (any layers that aren’t visible are still known to the user because of the layers number badge at the top of the panel (#9136) )
- If the user changes the height of the palette, it stays their height as they go through their workflow
- If the user toggles the layers panel to expand/collapse using the arrow at the top of the panel this is also respected as the height through their workflow. 
  - this is retooling the current saved height set up.
- If they close the palette using the arrow toggle, the number of layers will visibly change on the badge as they add but it will just stay closed throughout their session unless they open it again

This issue's had quite a lengthy discussion in slack and I suspect there will need to be more here. Whoever picks this up to work on it, please open a dialog in the design slack channel for any questions and to get confirmation on things that feel ambiguous.  

## Acceptance criteria 
- Layers panel height is updated in the ways listed above 
- Tests updated and added accordingly
- Layers panel is still usable for keyboards and SR users, if there isn't a test yet - add an aXe validation test in karma too for this.

samwhale:
## Context

Layer panel is getting a shiny new coat of paint. Small updates ahead of larger panel ux updates. 

## Summary

Layer panel height updates:
- Default height is now 25%
- Changing the height of the panel is persisted across pages
- There is no maximum height
- Collapsing the panel will keep the panel closed across pages
- Any size change is persisted in local storage

## Relevant Technical Choices

n/a

## To-do

- [x] Layer panel was hidden from screen readers in #4432. Waiting for [this discussion](https://xwp.slack.com/archives/C01ESBB8ZN1/p1633620351120500) to decide whether to update accessibility. Decision: don't do anything for now

## Testing Instructions

1. Load editor
2. Play with layers panel. The following should all happen:
    - Default layers panel height to 25% of the total screen height
    - Unless the user adjusts the height manually, the panel will stay this default height regardless of the layer count
    - If the user changes the height of the palette, it stays their height as they go through their workflow
    - If the user toggles the layers panel to expand/collapse using the arrow at the top of the panel this is also respected as the height through their workflow.
    - The height and collapsed state will persist when the page is refreshed

no

no

no

## Checklist

---

Fixes #8931


packages/story-editor/src/components/panels/design/layer/layerPanel.js
```diff
@@ -50,24 +49,16 @@ const Divider = styled.div`
...
   return (
     <Container>
       <Panel
-        name="layers"
-        initialHeight={Math.min(
-          numLayersVisible * LAYER_HEIGHT,
-          window.innerHeight / 3
-        )}
-        maxHeight={layers?.length * LAYER_HEIGHT}
+        name="layer"
```
samwhale: ‼️ This panel name was updated. This sets new values in local storage and ignores the old ones, so every user will start with the layer panel at 25% when this code is updated. This will only happen once.

packages/story-editor/src/components/panels/design/preset/resize.js
```diff
@@ -38,8 +38,9 @@ function Resize({ position }) {
     state: { inspectorContentHeight },
   } = useInspector();
 
-  // Max panel height is set to 70% of full available height.
-  const maxHeight = Math.round(inspectorContentHeight * 0.7);
+  // Default max panel height is set to 70% of full available height.
+  const maxHeight =
+    maxHeightOverride || Math.round(inspectorContentHeight * 0.7);
```
samwhale, line 43: 🤔 tried to keep the previous max height here for other panels. Now the `maxHeightOverride` allows us to set any max height rather than always being constrained to 70% of the inspector content height

BrittanyIRL (approved):
I feel like these updates make a lot of sense. 

This is odd, but I checked it against main and it's new or rather, probably newly noticeable. I think it should be a new ticket. 
 
Screen Shot 2021-10-07 at 4 14 45 PM

packages/story-editor/src/components/panels/design/layer/layerPanel.js
```diff
@@ -50,24 +49,16 @@ const Divider = styled.div`
...
-        )}
-        maxHeight={layers?.length * LAYER_HEIGHT}
+        name="layer-panel"
+        initialHeight={window.innerHeight / 4}
         resizeable
-        ariaHidden
         collapsedByDefault={false}
       >
-        <StyledPanelTitle isSecondary isResizable>
+        <StyledPanelTitle isSecondary isResizable maxHeight={9999999999999}>
```
swissspidy: That is already the default value in the `Panel` component. See `MAX_HEIGHT_DEFAULT`. Maybe you can reuse that const?

packages/story-editor/src/components/panels/design/layer/karma/layer.karma.js
```diff
@@ -66,17 +66,4 @@ describe('Layer Panel', () => {
     ]);
     expect(section.clientHeight).toBe(initialHeight - 30);
   });
-
-  it('should not allow increasing the height of the layer panel past the max height', async () => {
```
miina, line 110: Should this perhaps be replaced with a more suitable test, e.g. testing the layer panel height stays the same after switching pages or something similar?

packages/story-editor/src/components/panels/design/layer/layerPanel.js
```diff
@@ -50,24 +54,20 @@ const Divider = styled.div`
...
     <Container>
       <Panel
-        name="layers"
-        initialHeight={Math.min(
-          numLayersVisible * LAYER_HEIGHT,
-          window.innerHeight / 3
-        )}
-        maxHeight={layers?.length * LAYER_HEIGHT}
+        name="layer-panel"
+        initialHeight={window.innerHeight / 4}
```
maxyinger: not a biggie, but might be nice to memoize this just so it's not calcing innerHeight on every render

packages/story-editor/src/components/panels/design/preset/resize.js
maxyinger: yea I guess it's not a bad default to have. I don't see it breaking anything by not having it tho 🤷 . I'm fine either way

maxyinger (approved):
looks and works great!

packages/story-editor/src/components/panels/design/layer/layerPanel.js
swissspidy: Rounding the number might also be helpful here

BrookeGraham (approved):
Nice.

packages/story-editor/src/components/panels/panel/shared/title.js
```diff
@@ -245,6 +245,7 @@ function Title({
             data-testid="panel-badge"
             notificationCount={count}
             variant={BUBBLE_VARIANTS.PRIMARY}
+            aria-hidden
```
samwhale, line 248: layer panel is not visible to screen reader. Adding this to the `NotificationBubble` - so this is just so we can target the layer heading for tests

packages/story-editor/src/components/panels/design/layer/layerPanel.js
```diff
@@ -50,24 +55,23 @@ const Divider = styled.div`
...
-        name="layers"
-        initialHeight={Math.min(
-          numLayersVisible * LAYER_HEIGHT,
-          window.innerHeight / 3
-        )}
-        maxHeight={layers?.length * LAYER_HEIGHT}
+        name="layer-panel"
+        initialHeight={initialHeight}
         resizeable
-        ariaHidden
```
swissspidy, line 67: Is this intended? Now the layers panel is no longer hidden from screen readers

swissspidy:
@samwhale Looks like the layer panel is no longer hidden from screen readers, which causes these aXe failures in the e2e tests being surfaced:

It's related to `<VisibleImage>` which is used for the layer panel icons.

> Alternative text of images should not be repeated as text

https://dequeuniversity.com/rules/axe/4.3/image-redundant-alt

https://github.com/google/web-stories-wp/runs/3927433248

Using `alt=""` to mark them as decorative would fix the issue.

packages/story-editor/src/components/panels/design/layer/layerPanel.js
samwhale: Oh I didn't realize. Will update this now

samwhale:
> @samwhale Looks like the layer panel is no longer hidden from screen readers, which causes these aXe failures in the e2e tests being surfaced

Nice catch! The layer panel should remain hidden - I put the `aria-hidden` in the wrong place there.

I'll fix this so the tests don't fail anymore

samwhale:
Alright fixed it. @swissspidy I'll merge it now 👍

PR #9323, opened 2021-10-07
Pre-Publish Checklist: Add check for missing publisher logo

### LINKED ISSUE #9320: PPC: No Publisher Logos Check
@swissspidy pointed out that the issue that shows in the prepublish checklist when users have no publisher logos yet is incorrect - showing that they need to increase the size of the logo rather than just add a new one. 

Let's add a check for a publisher logo existing in a story before telling the user they need to make it larger. 
If there is no publisher logo let's display `Add a publisher logo to your story` that when clicked takes you to the publisher logo upload button. The helper text can stay the same as for increasing the size.

samwhale:
## Context

The "publisher logo too small" check was showing when there was no publisher logos

## Summary

Adds a publisher logo existence card to the checklist and prevents the 'publisher logo too small' card from showing.

## Relevant Technical Choices

n/a

## To-do

n/a

## User-facing changes

existence

## Testing Instructions

1. Delete all of the publisher logos in your app. This can be done by running `npm run env:reset-site` (which blows up everything) or deleting the media one by one from the media library

2. Create a new story from a template (easiest since it has a lot of pages)
3. Open the checklist. See the new card in the priority section
4. Click the card - should highlight the publisher logo section like in the gif

no

no

no

## Checklist

---

Fixes #9320

### DIFF
--- a/packages/story-editor/src/components/checklist/checklistContent/accessibilityChecks.js
+++ b/packages/story-editor/src/components/checklist/checklistContent/accessibilityChecks.js
@@ -46,7 +46,7 @@ function AccessibilityPanel({
   return isChecklistMounted ? (
     <StyledTablistPanel
       badgeCount={badgeCount}
-      isExpanded={badgeCount && isOpen}
+      isExpanded={Boolean(badgeCount) && isOpen}
       onClick={onClick}
       maxHeight={maxHeight}
       title={title}
--- a/packages/story-editor/src/components/checklist/checks/publisherLogoExistence.js
+++ b/packages/story-editor/src/components/checklist/checks/publisherLogoExistence.js
@@ -30,7 +30,7 @@ import { useRegisterCheck } from '../countContext';
 import { useIsChecklistMounted } from '../popupMountedContext';
 
 export function publisherLogoExistence(publisherLogo) {
-  return !publisherLogo?.url && !publisherLogo?.height && !publisherLogo?.width;
+  return !publisherLogo?.id;
 }
 
 const PublisherLogoExistence = () => {
--- a/packages/story-editor/src/components/checklist/checks/publisherLogoExistence.js
+++ b/packages/story-editor/src/components/checklist/checks/publisherLogoExistence.js
@@ -25,19 +25,15 @@ import { List, THEME_CONSTANTS } from '@web-stories-wp/design-system';
 import { useStory } from '../../../app';
 import { states, useHighlights } from '../../../app/highlights';
 import { ChecklistCard, ChecklistCardStyles } from '../../checklistCard';
-import { PRIORITY_COPY, PUBLISHER_LOGO_DIMENSION } from '../constants';
+import { PRIORITY_COPY } from '../constants';
 import { useRegisterCheck } from '../countContext';
 import { useIsChecklistMounted } from '../popupMountedContext';
 
-export function publisherLogoSize(publisherLogo) {
-  return (
-    Boolean(publisherLogo?.id) &&
-    (publisherLogo?.height < PUBLISHER_LOGO_DIMENSION ||
-      publisherLogo?.width < PUBLISHER_LOGO_DIMENSION)
-  );
+export function publisherLogoExistence(publisherLogo) {
+  return !(publisherLogo?.id || publisherLogo?.url);
 }
 
-const PublisherLogoSize = () => {
+const PublisherLogoExistence = () => {
   const isChecklistMounted = useIsChecklistMounted();
   const publisherLogo = useStory(({ state }) => state?.story?.publisherLogo);
   const setHighlights = useHighlights(({ setHighlights }) => setHighlights);
@@ -49,10 +45,10 @@ const PublisherLogoSize = () => {
     [setHighlights]
   );
 
-  const isRendered = publisherLogoSize(publisherLogo);
-  useRegisterCheck('PublisherLogoSize', isRendered);
+  const isRendered = publisherLogoExistence(publisherLogo);
+  useRegisterCheck('PublisherLogoExistence', isRendered);
 
-  const { footer, title } = PRIORITY_COPY.logoTooSmall;
+  const { footer, title } = PRIORITY_COPY.noPublisherLogo;
   return (
     isRendered &&
     isChecklistMounted && (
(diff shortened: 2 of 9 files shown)


packages/story-editor/src/components/checklist/checklistContent/accessibilityChecks.js
```diff
@@ -46,7 +46,7 @@ function AccessibilityPanel({
   return isChecklistMounted ? (
     <StyledTablistPanel
       badgeCount={badgeCount}
-      isExpanded={badgeCount && isOpen}
+      isExpanded={Boolean(badgeCount) && isOpen}
```
samwhale, line 49: Proptype was failing because `badgeCount` is a number. When it is `0` then isExpanded is not a boolean

packages/story-editor/src/components/checklist/checks/publisherLogoExistence.js
```diff
@@ -0,0 +1,72 @@
...
+ */
+import { useStory } from '../../../app';
+import { states, useHighlights } from '../../../app/highlights';
+import { ChecklistCard, ChecklistCardStyles } from '../../checklistCard';
+import { PRIORITY_COPY } from '../constants';
+import { useRegisterCheck } from '../countContext';
+import { useIsChecklistMounted } from '../popupMountedContext';
+
+export function publisherLogoExistence(publisherLogo) {
+  return !(publisherLogo?.id || publisherLogo?.url);
```
BrittanyIRL: Since a publisher logo can have an id of 0, let's make sure that's not returning true and showing this check. 
(reset site and uploaded 1 logo)
Screen Shot 2021-10-07 at 3 37 39 PM

(publisherLogo logged before return)
Screen Shot 2021-10-07 at 3 40 09 PM

BrittanyIRL (changes requested):
Thanks for whipping this up so fast! 

Needs a little extra check for when the logo id is 0, then i think it's good!

packages/story-editor/src/components/checklist/checks/publisherLogoExistence.js
BrittanyIRL:
```suggestion
  return Boolean(publisherLogo?.id || publisherLogo?.url);
```

This will fix it

spacedmonkey:
SGTM

packages/story-editor/src/components/checklist/checks/publisherLogoExistence.js
samwhale: I thought publisher logos couldn't have an id of zero 🤔 

So to make sure we don't have one, should we just check that there's no url, no height, and no width? Since `id === 0` doesn't guarantee that there's no publisher logo?

swissspidy: I think what Brittany is saying that right now when there is no publisher logo we return an "empty" object with `id === 0` instead of `null`. Something we're probably going to change.

Checking for something like `const hasPublisherLogo = Boolean(publisherLogo?.id)` should be enough

samwhale: >Checking for something like const hasPublisherLogo = Boolean(publisherLogo?.id) should be enough

Currently if there is no publisher logo then the publisher logo on in the story context is `{ id: 0, height: 0, width: 0, url: '' }`, so the `hasPublisherLogo` check would always be true.

Once the stored 'empty state' of the publisher logo changes to `null` that will work but for now I think it needs to check if the id is `0`, no? Unless I'm just having major Friday brain

swissspidy: > Once the stored 'empty state' of the publisher logo changes to null that will work but for now I think it needs to check if the id is 0, no?

If you coerce `0` to a `boolean` it will be `false`. Just what we want.

```js
const publisherLogo = { id: 0 }
const hasPublisherLogo = Boolean(publisherLogo?.id) // this will be false. exactly what we want.
```

packages/story-editor/src/components/checklist/checks/publisherLogoExistence.js
```diff
@@ -0,0 +1,72 @@
...
+import { useStory } from '../../../app';
+import { states, useHighlights } from '../../../app/highlights';
+import { ChecklistCard, ChecklistCardStyles } from '../../checklistCard';
+import { PRIORITY_COPY } from '../constants';
+import { useRegisterCheck } from '../countContext';
+import { useIsChecklistMounted } from '../popupMountedContext';
+
+export function publisherLogoExistence(publisherLogo) {
+  return !publisherLogo?.url && !publisherLogo?.height && !publisherLogo?.width;
+}
```
swissspidy:
```suggestion
export function publisherLogoExistence(publisherLogo) {
  return !(publisherLogo?.id);
}
```

swissspidy: Also, FWIW, this function checks whether the logo is **missing**, but the function is named publisherLogo_Existence_. That is super confusing.

Either rename this to `publisherLogoInexistence` (which is also confusing), or reverse the check:

```suggestion
/**
 * Determines whether a publisher logo exists.
 * @param {Object} publisherLogo Publisher logo object.
 * @return {boolean} Whether a publisher logo exists.
 */
export function publisherLogoExistence(publisherLogo) {
  return Boolean(publisherLogo?.id);
}
```

Of course update `PublisherLogoExistence` below accordingly

packages/story-editor/src/components/checklist/checks/publisherLogoExistence.js
swissspidy: No need for the `url` check really. Just checking for the ID is enough.

samwhale: Oh wait we're saying the same thing 😮‍💨 okay, it should be updated correctly now.

packages/story-editor/src/components/checklist/checks/publisherLogoExistence.js
BrittanyIRL: +1 for `PublisherLogoMissing`

maxyinger (approved):

packages/story-editor/src/components/checklist/checks/publisherLogoExistence.js
samwhale: Ah good shout

BrittanyIRL (approved):
👍

PR #9330, opened 2021-10-08
Karma: use dots reporter

### LINKED ISSUE #8807: Karma: use dots reporter
## Task Description

Karma test output can be quite long.

Let's use a dot reporter like in all our other test suites (unit, PHP, e2e) to reduce output.

With that, successful tests will be simple dots (`.`), whereas failures can still display detailed output.

spacedmonkey:
## Summary

Change karma reporter to dots from spec. 

## Relevant Technical Choices

Remove spec reporter all together. 

## Testing Instructions

1.

## Checklist

---

Fixes #8807

### DIFF
--- a/packages/story-editor/src/index.js
+++ b/packages/story-editor/src/index.js
@@ -35,7 +35,7 @@ export { default as FontContext } from './app/font/context';
 export { default as useLoadFontFiles } from './app/font/actions/useLoadFontFiles';
 export { default as StoryPropTypes } from './types';
 export { GlobalStyle } from './theme';
-export { default as theme } from './theme'; // @todo To be refactored at some point.
+export { default as theme } from './theme'; // @todo To be refactored.
 export { GlobalStyle as CropMoveableGlobalStyle } from './components/moveable/cropStyle';
 export { default as CircularProgress } from './components/circularProgress';
 export { default as InterfaceSkeleton } from './components/layout';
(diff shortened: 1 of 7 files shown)


packages/story-editor/src/index.js
```diff
@@ -35,7 +35,7 @@ export { default as FontContext } from './app/font/context';
 export { default as useLoadFontFiles } from './app/font/actions/useLoadFontFiles';
 export { default as StoryPropTypes } from './types';
 export { GlobalStyle } from './theme';
-export { default as theme } from './theme'; // @todo To be refactored.
+export { default as theme } from './theme'; // @todo To be refactored at some point.
```
swissspidy: 🤔

swissspidy:
Can you convert this back to a draft again since it's WIP? Then we don't unnecessarily do Percy snapshots and things like that.

FWIW, we should update the reporters config in the config files instead of passing `--reporters`. We can remove `spec` and `karma-spec-reporter` entirely in favor of `dots`

packages/story-editor/src/index.js
spacedmonkey: This was to force the karma test to run.

spacedmonkey: Removed now

swissspidy:
Odd unit test failure: https://github.com/google/web-stories-wp/runs/3838941852

Does that happen locally as well? Might be a flakey test that's worth looking into in a separate ticket.

swissspidy (approved):

spacedmonkey:
> Odd unit test failure:

Tests are now passing. Happy to merge?

PR #9331, opened 2021-10-08
Use `memo` for `ReorderableScroller`

### LINKED ISSUE #8587: Performance: Memoize direct descendants of context providers in the editor
## Feature Description

As [this article](https://dev.to/leighhalliday/that-react-component-right-under-your-context-provider-should-probably-use-react-memo-glm) argues it probably makes sense to memoize every direct child below a context provider to optimize the render tree.

This includes both the always-present global providers used in the global architecture of the editor, but also the smaller local providers used for panels, layers, etc.

This does not guarantee a performance increase for _all_ providers, but probably for _most_. Only apply it where it actually makes a difference (if the child component always re-renders anyway, it of course doesn't matter).

Please try to make some measurements of user experience before and after and add these to the PR.

---

miina:
## Summary
This goal of this PR was to `memo` the components that are direct children of context providers.

Testing showed that Max [was correct](https://github.com/google/web-stories-wp/issues/8587#issuecomment-892727011) and the way we use the providers already comes with the benefit. Adding `memo` to the children of providers used as we mostly do, did not reduce the count of renders.

There was one component (`ReorderableScroller`) where it came out to be beneficial though and as an example, the times of being re-rendered was reduced from 8 to 6 in case of switching between two empty pages.

To clarify: in case of `ReorderableScroller`, the component is used directly inside the provider and not as we usually do.

| Before | After |
|--------|------|
| Screenshot 2021-10-08 at 15 36 38 | Screenshot 2021-10-08 at 15 36 52 |

## User-facing changes
N/A

## Testing Instructions

1. Add one empty page (so you have 2 empty pages as a result).
2. Start the profiler.
3. Click on the non-selected page (ensure that's the only thing you do, no hovering over other elements etc.)
4. Stop the profiler and keep the window open
5. Do the same on `main` in a separate window
6. Compare the results for `ReorderableScoller`.

No

No

No

## Checklist

---

Fixes #8587

### DIFF
--- a/packages/story-editor/src/components/reorderable/reorderableScroller.js
+++ b/packages/story-editor/src/components/reorderable/reorderableScroller.js
@@ -24,6 +24,7 @@ import {
   useState,
   useCallback,
   useContext,
+  memo,
 } from '@web-stories-wp/react';
 
 /**
@@ -120,4 +121,4 @@ ReorderableScroller.propTypes = {
   size: PropTypes.number.isRequired,
 };
 
-export default ReorderableScroller;
+export default memo(ReorderableScroller);


barklund (approved):
Thanks for looking into this. If there were some cheap wins, awesome. But if not, well, we tried 🤷.

(though this tiny win is of course still a win)

PR #9337, opened 2021-10-08
Video Muting: Fix resource state updates

### LINKED ISSUE #9335: Video Muting: "Remove audio" button not disappearing
## Bug Description

We got a user report about issues with resource updates when muting a video. The "Remove audio" button was simply not going away, yet the new file was uploading successfully.

I looked into it a bit and found some areas for improvement here.

## Expected Behaviour

When muting a video, the "Remove audio" button should be properly removed after the process finishes.

## Steps to Reproduce

1. Click "Remove audio" on video
2. Wait for new video to be uploaded
3. See that "Remove audio" button is still there

See also original bug report: https://wordpress.org/support/topic/remove-audio-not-working-anymore/#post-14951400

## Additional Context

- Plugin Version: 1.12.0
- WordPress Version: 5.8
- Operating System: macOS
- Browser: Chrome

swissspidy:
## Context

We got a user report about issues with resource updates when muting a video. The "Remove audio" button was simply not going away, yet the new file was uploading successfully.

## Summary

The reason for this bug:

While the `isMuting`/`isMuted` flags were properly set during start and progress, they got reset again at the end of the upload process? `additionalData.is_muted` was set to `true` when uploading the file to WordPress, yet the REST API response contained `is_muted: false` because the meta value was not properly updated

This PR ensures that the flags are kept and that the "Remove audio" button is correctly disappearing after muting.

## Relevant Technical Choices

* See individual commits for details
* Updates `updateExistingElements` to actually update the current state and only update the resource properties that are needed, so no data is potentially lost
* Adding some todos for followup tickets.
* Code simplification
* Prefixes the REST API field

## To-do

N/A

## User-facing changes

https://user-images.githubusercontent.com/841956/136575412-938a7afa-826d-471b-adba-3e6687e57bae.mov

## Testing Instructions

1. Add a new video with sound to the page
2. Click on "Remove audio"
3. See that new video is uploading
4. After upload finishes, see that "Remove audio" button is removed

Also:

* Verify that video trimming and video optimization via the design panel still works.

No

No

No

## Checklist

---

Fixes #9335


includes/Media/Video/Muting.php
```diff
@@ -201,6 +200,10 @@ public function update_callback_is_muted( $value, $object ) {
 			);
 		}
 
-		return (bool) update_metadata( $meta_type, $object_id, $meta_key, (int) $value );
+		$value = rest_sanitize_boolean( $value );
+
+		update_post_meta( $object_id, $meta_key, $value );
+
+		return true;
```
spacedmonkey, line 205:
```suggestion
		return (bool) update_post_meta( $object_id, $meta_key, $value );;
```

I think this is cleaner.

swissspidy: Attention. This has different semantics.

I changed this on purpose. Specifically to fix the actual bug at play here.

For two reasons:

1. The return value doesn't matter and is not used if it's not a `WP_Error`
2. `update_post_meta` returns the meta ID if the key didn't exist, `true` on successful update, `false` on failure or if the value passed to the function is the same as the one that is already in the database.

So "returns false if the value is already the same as the one that is already in the database". That is really dangerous!

It means if `true` is in the database and we save `true` again, we get `false`. That's not what we want!

Thus, this does not work. Which is why I changed it.

spacedmonkey: Okay, thanks for the context, it is useful.

spacedmonkey:
Add @barklund to this PR, as it might effect his changes in https://github.com/google/web-stories-wp/pull/9315

spacedmonkey:
>  because the meta value was not properly updated

Can you explain what fixes the issues explainly. A lot of this PR, seems to be change the field name.

swissspidy:
> > because the meta value was not properly updated
> 
> Can you explain what fixes the issues explainly. A lot of this PR, seems to be change the field name.

Looking at the individual commits one by one makes reviewing a bit easier. Renaming is one of the technical changes in the PR. The rest of the PR is just additional hardening on the client side while upload is in progress.

8666ef873c8047de2d374667bfd32d85a9044b8a is the most important fix as saving the post meta was not working as expected.

As mentioned in the PR description, we sent `is_muted: true` in the REST API request, but got `is_muted: false` in return, because the post meta was not properly saved with this value.

I think it's because the value was cast to an **integer**, but the meta key is registered as a **boolean**. So WP rejected it.

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -59,11 +59,17 @@ function useProcessMedia({
...
 
   const updateExistingElements = useCallback(
-    ({ oldResource: resource }) => {
-      const { id } = resource;
+    (resourceId, resource) => {
       updateElementsByResourceId({
-        id,
-        properties: () => ({ resource }),
+        id: resourceId,
+        properties: (element) => {
```
swissspidy, line 65: With this change to `updateExistingElements` we can 

1. update only the fields of the `resource` that we want to
2. ensure that we update the element's current state

spacedmonkey:
> Looking at the individual commits one by one makes reviewing a bit easier. 

I did that, as I read this. 

> See individual commits for details

But this PR has a lot of any important changes in it. It makes it hard to understand that https://github.com/google/web-stories-wp/commit/8666ef873c8047de2d374667bfd32d85a9044b8a is the only bit that really fixes anything. 

> I think it's because the value was cast to an integer, but the meta key is registered as a boolean. So WP rejected it.

Where is this being casted to an integer, PHP or javascript?

swissspidy:
> Where is this being casted to an integer, PHP or javascript?

In PHP.

Here:

https://github.com/google/web-stories-wp/pull/9337/files#diff-db8df12f2eaa4f08f3404bad2383b1aa9dc167646ef53ae99a55063545acc322L204

This was the old code:

```php
return (bool) update_metadata( $meta_type, $object_id, $meta_key, (int) $value );
```

includes/Media/Video/Muting.php
```diff
@@ -201,6 +200,10 @@ public function update_callback_is_muted( $value, $object ) {
 			);
 		}
 
-		return (bool) update_metadata( $meta_type, $object_id, $meta_key, (int) $value );
+		$value = rest_sanitize_boolean( $value );
```
spacedmonkey: Isn't this covered by 
```php
'arg_options' => [
						'sanitize_callback' => 'rest_sanitize_boolean',
					],
```

swissspidy: Hmm yeah I might be able to remove this.

spacedmonkey (approved):
One question, but looks good

includes/Media/Video/Muting.php
swissspidy: Removed! 👍

PR #9338, opened 2021-10-08
Design Panel: Add badge to layer panel

### LINKED ISSUE #9136: Layers Panel: Add layer count badge to title area
## Description

Sometimes not all layers are visible in the layers panel and it's not super obvious that there are more if you scroll. Add in a badge that displays the count of all the layers in the title area of the layers panel. 

Figma: https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor-Sprint-Feature-Level-Designs_?node-id=9954%3A422177 

Design as of Oct 4, 2021:
Screen Shot 2021-10-04 at 10 50 38 AM

## To Do
- [x] - Final UX from design team

## Acceptance Criteria 

- badge displays with layers count, badge stays current even when panel is collapsed. 
- updated tests

samwhale:
## Context

Layer Panel Updates

## Summary

Adds a badge for the layer count on the layer panel

## Relevant Technical Choices

n/a

## To-do

n/a

## User-facing changes

badge

## Testing Instructions

1. Open the editor
2. Add things to your page
3. Check out the badge as it changes on the layer panel

no

no

no

## Checklist

---

Fixes #9136

### DIFF
--- a/packages/design-system/src/components/notificationBubble/index.js
+++ b/packages/design-system/src/components/notificationBubble/index.js
@@ -44,9 +44,7 @@ function getBubbleWidth(numDigits) {
 const Bubble = styled.div`
   ${({ theme, variant }) => css`
     color: ${theme.colors.fg.primary};
-    background-color: ${variant === BUBBLE_VARIANTS.ACCENT
-      ? theme.colors.accent.primary
-      : theme.colors.bg[variant]};
+    background-color: ${theme.colors.bg[variant]};
     border-radius: ${theme.borders.radius.round};
   `}
   position: relative;
--- a/packages/design-system/src/components/notificationBubble/index.js
+++ b/packages/design-system/src/components/notificationBubble/index.js
@@ -24,15 +24,29 @@ import PropTypes from 'prop-types';
  */
 import { themeHelpers } from '../../theme';
 
+export const BADGE_VARIANTS = {
+  PRIMARY: 'primary',
+  SECONDARY: 'secondary',
+  TERTIARY: 'tertiary',
+  QUATERNARY: 'quaternary',
+  POSITIVE: 'positive',
+  NEGATIVE: 'negative',
+  ACCENT: 'accent',
+};
+
+const BUBBLE_DIAMETER = 24;
+
 const Bubble = styled.div`
-  ${({ theme }) => css`
+  ${({ theme, variant }) => css`
     color: ${theme.colors.fg.primary};
-    background-color: ${theme.colors.accent.primary};
+    background-color: ${variant === 'accent'
+      ? theme.colors.accent.primary
+      : theme.colors.bg[variant]};
     border-radius: ${theme.borders.radius.round};
   `}
   position: relative;
   height: 24px;
-  width: 24px;
+  width: ${({ digitLen }) => BUBBLE_DIAMETER + 9 * (digitLen - 1)}px;
 
   ${({ isSmall }) =>
     isSmall &&
@@ -56,13 +70,20 @@ const Inner = styled.span`
 export const NotificationBubble = ({
   notificationCount,
   isSmall,
+  variant = BADGE_VARIANTS.ACCENT,
   ...props
 }) => (
-  <Bubble isSmall={isSmall} {...props}>
+  <Bubble
+    variant={variant}
+    isSmall={isSmall}
+    digitLen={notificationCount?.toString().length || 1}
+    {...props}
+  >
     <Inner isSmall={isSmall}>{notificationCount}</Inner>
   </Bubble>
 );
 NotificationBubble.propTypes = {
   notificationCount: PropTypes.number,
   isSmall: PropTypes.bool,
+  variant: PropTypes.oneOf(Object.values(BADGE_VARIANTS)),
 };
--- a/packages/design-system/src/components/notificationBubble/index.js
+++ b/packages/design-system/src/components/notificationBubble/index.js
@@ -24,7 +24,7 @@ import PropTypes from 'prop-types';
  */
 import { themeHelpers } from '../../theme';
 
-export const BADGE_VARIANTS = {
+export const BUBBLE_VARIANTS = {
   PRIMARY: 'primary',
   SECONDARY: 'secondary',
   TERTIARY: 'tertiary',
@@ -35,6 +35,11 @@ export const BADGE_VARIANTS = {
 };
 
 const BUBBLE_DIAMETER = 24;
+const SMALL_BUBBLE_DIAMETER = 20;
+
+function getBubbleWidth(numDigits) {
+  return 9 * (numDigits - 1);
+}
 
 const Bubble = styled.div`
   ${({ theme, variant }) => css`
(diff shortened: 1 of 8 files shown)


packages/story-editor/src/components/panels/panel/shared/title.js
```diff
@@ -74,6 +76,25 @@ const Heading = styled.span`
     })};
 `;
 
+const Badge = styled.span(
+  ({ digitLen, theme }) => css`
+    position: relative;
+    height: ${BADGE_DIAMETER}px;
+    width: ${BADGE_DIAMETER + 8 * (digitLen - 1)}px;
```
samwhale: Every time the layer count increases by a digit we add 8px to the width.

The badge's children are absolutely positioned to keep the badge circular. The width will not respect children's length, so as the number increases in width we need to add width to the badge as well.

Lmk if y'all have a better way to make a badge 😄

swissspidy: Can't we just use `NotificationBubble` from the design system...?

packages/story-editor/src/components/panels/panel/shared/title.js
```diff
@@ -229,6 +252,11 @@ function Title({
         >
           {children}
         </Heading>
+        {count !== undefined && (
```
swissspidy: Since you do `count={layers?.length}` it means `count` can also be `null`. And we don't really want to accidentally render `null` here.

I'd do

```suggestion
        {Boolean(count) && (
```

instead

samwhale: `0` is a possibility here too, no? I guess we always have a background layer so maybe we don't have to worry about the `0` case

This is also on the base `panel` so trying to make it agnostic to the layer panel.

Maybe this to account for the `0` case:

```js
(count === 0 || Boolean(count)) && ...
```

packages/story-editor/src/components/panels/panel/shared/title.js
maxyinger: I think notification `NotificationBubble` does the same thing too with the fixed width and absolute positioning to make the bg a circle. Should recycle if you can, but you'll probably need to account for the same problem there I think 🤔

maxyinger (approved):

packages/story-editor/src/components/panels/panel/shared/title.js
samwhale: Ah, thanks for the heads up. I didn't realize this existed because it's not in storybook 😭 

image

packages/design-system/src/components/notificationBubble/index.js
```diff
@@ -24,21 +24,40 @@ import PropTypes from 'prop-types';
...
+    background-color: ${variant === 'accent'
+      ? theme.colors.accent.primary
+      : theme.colors.bg[variant]};
     border-radius: ${theme.borders.radius.round};
   `}
   position: relative;
-  height: 24px;
-  width: 24px;
+  height: ${BUBBLE_DIAMETER}px;
+  width: ${({ digitLen }) => BUBBLE_DIAMETER + getBubbleWidth(digitLen)}px;
```
samwhale, line 52: Expand bubble depending on the number of digits of the bubble

packages/design-system/src/components/notificationBubble/index.js
```diff
@@ -49,20 +68,32 @@ const Inner = styled.span`
     themeHelpers.expandTextPreset(
       ({ paragraph }, sizes) => paragraph[isSmall ? sizes.X_SMALL : sizes.SMALL]
     )({ theme })};
-  color: ${({ theme }) => theme.colors.bg.primary};
+  color: ${({ invertColor, theme }) =>
+    invertColor ? theme.colors.inverted.fg.primary : theme.colors.fg.primary};
```
samwhale: Allow for choosing the text color

packages/story-editor/src/components/panels/panel/shared/title.js
samwhale: Should be fixed now!

packages/design-system/src/components/notificationBubble/index.js
```diff
@@ -24,21 +24,40 @@ import PropTypes from 'prop-types';
...
+  return 9 * (numDigits - 1);
+}
+
 const Bubble = styled.div`
-  ${({ theme }) => css`
+  ${({ theme, variant }) => css`
     color: ${theme.colors.fg.primary};
-    background-color: ${theme.colors.accent.primary};
+    background-color: ${variant === 'accent'
+      ? theme.colors.accent.primary
```
BrittanyIRL: Nit: So this is a slippery slope here to introduce a new background color in play that has otherwise only been an active or hover state for interactive bgs. If we're going to start having this violet color be a background i think it should be added to the background colors. Maybe `bg.accent`? 

(also i think this was here before your changes, so apologies for the scope creep)

packages/design-system/src/components/notificationBubble/index.js
```diff
@@ -49,20 +68,32 @@ const Inner = styled.span`
...
 export const NotificationBubble = ({
   notificationCount,
   isSmall,
+  variant = BUBBLE_VARIANTS.ACCENT,
+  invertTextColor,
   ...props
 }) => (
-  <Bubble isSmall={isSmall} {...props}>
-    <Inner isSmall={isSmall}>{notificationCount}</Inner>
+  <Bubble
```
BrittanyIRL, line 83: Thoughts on using some [transient prop](https://styled-components.com/docs/api#transient-props)s here?

packages/design-system/src/components/notificationBubble/stories/index.js
```diff
@@ -0,0 +1,91 @@
...
+  label {
+    display: flex;
+    align-items: center;
+  }
+`;
+
+export const _default = () => {
+  return (
+    <>
+      <Headline as="h1">{'Notification Bubble'}</Headline>
```
BrittanyIRL: nbd but just so you know this defaults to h1

packages/story-editor/src/components/panels/design/layer/karma/layer.karma.js
```diff
@@ -24,20 +24,54 @@ import { useInsertElement } from '../../../../canvas';
...
+        font: TEXT_ELEMENT_DEFAULT_FONT,
+        content: 'Doo doo doo',
+        x: 40,
+        y: 40,
+        width: 250,
+      })
+    );
+
+    expect(layerPanel.layers.length).toBe(3);
+    expect(layerPanel.panelBadge.textContent).toBe('3');
```
BrittanyIRL: thoughts on checking if the textContent matches `layerPanel.layers.length.toString()`?

BrittanyIRL (commented):
Thanks for reusing and refactoring the existing notification bubble, it'll be great to have that aligned going forward. 

It's working well, few nits. When comparing visually Figma's showing the number as weight: 700 font size 14px, i'm seeing 400 weight and 12px when I run this branch locally.

packages/design-system/src/components/notificationBubble/stories/index.js
samwhale: I can update the proptype here to not have the `as` prop being required 👍 https://github.com/google/web-stories-wp/blob/main/packages/design-system/src/components/typography/headline/index.js#L49

packages/design-system/src/components/notificationBubble/index.js
samwhale: yeah I think that's a good idea 👍

BrookeGraham (approved):
🗯

PR #9341, opened 2021-10-08
Improve embed block markup and AMP for WP compatibility

### LINKED ISSUE #9266: Improve block markup for accessibility & SEO
## Feature Description

The Web Stories block markup for things like the list embeds or grid embeds does not contain any link to the story itself. Instead, the URL is just in a `data-story-url` attribute.

Screenshot 2021-10-05 at 15 58 38

This is not good for accessibility and SEO, and also for when JavaScript is disabled on a page.

Instead, there should be a proper `<a>` tag around the story.

When JS is disabled, this still allows visiting the story by clicking on it.

When JS is enabled, we prevent the click event and open the lightbox as today.

https://github.com/google/web-stories-wp/blob/94aa5519ef7d037ff98f71dec919d07bbd93e9eb/includes/Renderer/Stories/Renderer.php#L521-L523

https://github.com/google/web-stories-wp/blob/94aa5519ef7d037ff98f71dec919d07bbd93e9eb/includes/Renderer/Stories/Renderer.php#L550-L552

https://github.com/google/web-stories-wp/blob/94aa5519ef7d037ff98f71dec919d07bbd93e9eb/includes/Renderer/Stories/Renderer.php#L560-L578

### LINKED ISSUE #9314: Web Stories block incompatibility with AMP for WP plugin
Hi, I posted the below in the wordpress forum but they are saying its a bug and there is no idea when it will be fixed.  I have now been trying for more than a year to get these web stories live, but had to wait for updates. Please can someone help me finally get them live as it is so frustrating to spend a whole year on something and it still not work.

I can not get the web stories to display, please see here https://www.webpagescreenshot.info/#v2=fYajXe3WC

Example web story below

https://nonfungibleownership.com/web-stories/sandbox-review/

Page not displaying on

https://nonfungibleownership.com/nft-art-tokens/

The page I need help with: https://nonfungibleownership.com/nft-art-tokens/

This was the reply:

@bitcoincasino666 Hello, you may be experiencing a known issue with amp-story-player as reported in the broader AMP service, ampproject/amphtml#35959.

Unfortunately, I don’t have a timeline on when this will be resolved, but we will update you here when the fix has been released.

swissspidy:
## Context

We got a user report about an incompatibility with the AMP for WP plugin, where our embed block is removed completely and causing AMP validation errors (see screenshots below).

Unfortunately the AMP for WP plugin is really inferior to the official AMP plugin and does not automatically add missing attributes to ensure validity, nor does it support extracting stylesheets  like the official AMP (Page Experience) plugin does. 😞 

## Summary

This PR improves embed block markup and improves compatibility with the AMP for WP plugin.

* Adds some required element attributes like `tabindex` which the official AMP plugin would normally add for us automatically.
* Hooks into `amp_post_template_css` provided by AMP for WP to load the necessary stylesheets.

## Relevant Technical Choices

* Adds embeds styles as inline stylesheet for the AMP for WP plugin
* Improve embeds markup to include proper `<a>` links for improved accessibility and SEO
* Fixes accessibility issue with the lightbox close button

## To-do

N/A

## User-facing changes

Before:

Screenshot 2021-10-08 at 17 25 41

After:

Screenshot 2021-10-08 at 17 25 30

## Testing Instructions

1. Activate the AMP for WP plugin (not the official AMP plugin)
2. Publish a new blog post and insert a Latest Stories Web Stories block
3. View the AMP version of the blog post on the site and verify that the block works as intended

No

No

No

## Checklist

---

Fixes #9266
Fixes #9314


packages/stories-block/src/css/lightbox.css
```diff
@@ -21,7 +21,7 @@
 /* amp-lightbox needs to have z-index to render on top of other elements in the page. */
 .web-stories-list__lightbox-wrapper amp-lightbox,
 .web-stories-list__lightbox.show {
-  z-index: 9999999;
+  z-index: 999999999;
```
westonruter, line 24: _MOAR!_

includes/Renderer/Stories/Renderer.php
```diff
@@ -681,15 +709,19 @@ protected function generate_amp_lightbox_html_amp( $story ) {
...
 			on="lightboxClose:AMP.setState({<?php echo esc_attr( $lightbox_state ); ?>: false})"
+			role="button"
+			tabindex="0"
 		>
 			<div class="web-stories-list__lightbox show">
-				<div
+				<button type="button"
 					class="story-lightbox__close-button"
 					on="tap:<?php echo esc_attr( $lightbox_id ); ?>.close"
+					tabindex="0"
```
westonruter: The `tabindex` is not needed if using a `button` element. It's only needed if using an element which is not natively focusable. 

```suggestion
```

includes/Renderer/Stories/Renderer.php
```diff
@@ -272,11 +272,33 @@ public function init() {
...
 		// Web Stories styles for AMP and non-AMP pages.
 		$this->assets->register_style_asset( self::STYLE_HANDLE );
 
 		// Web Stories lightbox script.
 		$this->assets->register_script_asset( self::LIGHTBOX_SCRIPT_HANDLE, [ AMP_Story_Player_Assets::SCRIPT_HANDLE ] );
+
+		$path = $this->assets->get_base_path( sprintf( 'assets/css/%s.css', self::STYLE_HANDLE ) );
+		if ( is_rtl() ) {
+			$path = $this->assets->get_base_path( sprintf( 'assets/css/%s-rtl.css', self::STYLE_HANDLE ) );
+		}
```
westonruter:
```suggestion
		$path = $this->assets->get_base_path( sprintf( 'assets/css/%s%s.css', self::STYLE_HANDLE, is_rtl() ? '-rtl' : '' ) );
```

westonruter (commented):
LGTM. Just a couple nits.

includes/Renderer/Stories/Renderer.php
swissspidy: Ah, good to know! I think I added this before I turned it into a button. Will happily remove it again.

packages/stories-block/src/css/lightbox.css
swissspidy: For documentation, I added this because the AMP for WP reader theme (or whatever is called) has some social buttons with this `z-index`. Increasing it so these buttons aren't visible when opening the stories lightbox.

includes/Embed_Base.php
```diff
@@ -81,8 +84,13 @@ public function register() {
...
 			$css = file_get_contents( $path ); // phpcs:ignore WordPressVIPMinimum.Performance.FetchingRemoteData.FileGetContentsUnknown
 
-			if ( $css ) {
-				wp_add_inline_style( self::SCRIPT_HANDLE, $css );
+			if ( defined( 'AMPFORWP_VERSION' ) ) {
+				add_action(
+					'amp_post_template_css',
+					static function() use ( $css ) {
+						echo $css; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+					}
```
spacedmonkey: I personally hate these static functions as callbacks to actions. They are very hard to unhook. 

How about this. 
```
$this->css = file_get_contents( $path );
```
```
if( defined( 'AMPFORWP_VERSION' ) ) {
     add_action('amp_post_template_css', [$this, 'output_css']);
}
```

```
public function output_css(){
   if ( $this->css ) {
       echo $this->css;
   }
}
```


This would too, we can also add a unit test.

includes/Renderer/Story/Embed.php
```diff
@@ -107,7 +101,7 @@ public function render( array $args = [] ): string {
 		);
 
 		// This CSS is used for AMP and non-AMP.
-		$this->assets->enqueue_style_asset( self::SCRIPT_HANDLE );
+		$this->assets->enqueue_style_asset( Embed_Base::SCRIPT_HANDLE );
```
spacedmonkey, line 104: Is there a worry that the embed base service will not be registered here?

includes/Embed_Base.php
```diff
@@ -68,6 +67,10 @@ public function __construct( Assets $assets ) {
 	 * @return void
 	 */
 	public function register() {
+		if ( wp_style_is( self::SCRIPT_HANDLE, 'registered' ) ) {
+			return;
+		}
```
spacedmonkey, line 72: How does this fix the problem?

includes/Renderer/Stories/Renderer.php
```diff
@@ -681,15 +709,18 @@ protected function generate_amp_lightbox_html_amp( $story ) {
...
 			on="lightboxClose:AMP.setState({<?php echo esc_attr( $lightbox_state ); ?>: false})"
+			role="button"
+			tabindex="0"
 		>
 			<div class="web-stories-list__lightbox show">
-				<div
+				<button type="button"
 					class="story-lightbox__close-button"
 					on="tap:<?php echo esc_attr( $lightbox_id ); ?>.close"
+					aria-label="<?php esc_attr_e( 'Close', 'web-stories' ); ?>"
```
spacedmonkey, line 721: 👍

includes/Embed_Base.php
swissspidy: This method can be called many times when there are many embed blocks on a page, but assets should only be registered once. And of course the inline CSS should only be printed once.

This should probably refactored at some point, but in a new ticket.

includes/Embed_Base.php
swissspidy: Uhm, let me just move the whole logic into such a method. Really don't like using class properties for this kinda stuff.

includes/Renderer/Story/Embed.php
swissspidy: Not to my knowledge, but if so, this is not introduced by this PR and would need its own ticket.

This change here just removes an unneeded constant from this class because the actual style with this name is registered by `Embed_Base`. This class here does not have its own styles.

includes/Embed_Base.php
spacedmonkey: There is already logic in the assets class to ensure it is not registered twice. See 
https://github.com/google/web-stories-wp/blob/1723f472d2bb966faef7b0ccad8c5342608ac0fa/includes/Assets.php#L193-L195

Which is basically the same logic. These lines should not be needed.

swissspidy: It's still needed because otherwise `add_action( 'amp_post_template_css', [ $this, 'add_amp_post_template_css' ] );` gets called dozens of times, causing the CSS to be printed dozens of times on AMP pages.

spacedmonkey (approved):

PR #9342, opened 2021-10-08
Taxonomies: Add indentation for children of hierarchical terms

### LINKED ISSUE #9280: Taxonomies: Make it obvious to users what the top level Categories are
## Feature Description
originally reported by @barklund 
As can be seen, the categories are not presented in a tree when choosing a parent category, so it’s not clear from the dropdown that “Corgis” is a child category of “Dogs”.
image9

maxyinger:
## Context
Discovered during bug bash. It's hard when selecting a category as a parent to know how it fits in the nesting.

## Summary
This PR orders the dropdown options such that all children options are below their parent and have an indent next to them to indicate their level.

## Relevant Technical Choices
Initially was working on passing in a custom `renderItem` that was just an extension of the base `DefaultListElement`, however I think I ran into a styled components bug or something because in react dev tools, it was definitely receiving the appended `$level` prop, however the output css would never receive that prop. to mitigate this, I just added a small style to the `DefaultListElement` that indents items if they have a `$level` prop.

## To-do
NA

## User-facing changes
In Editor -> Document -> Taxonomies -> Categories -> Add new Category, the dropdown to select the parent value now has visual indicators to respect the Category hierarchy. Now look like:
Screen Shot 2021-10-08 at 10 53 44 AM

## Testing Instructions

1. Open up the editor -> document tab -> categories panel
2. Make sure you have some nested Categories already (if not create some)
3. Add a new category and select a parent for it
4. See that the dropdown now has visual indicators for the hierarchy, like so:

Screen Shot 2021-10-08 at 10 53 44 AM

## Checklist

---

Fixes #9280

### DIFF
--- a/packages/design-system/src/components/menu/index.js
+++ b/packages/design-system/src/components/menu/index.js
@@ -15,7 +15,6 @@
  */
 
 export { Menu } from './menu';
-export { DefaultListItem } from './list';
 export {
   DROP_DOWN_ITEM,
   DROP_DOWN_ITEMS,
--- a/packages/design-system/src/components/menu/list/defaultListItem.js
+++ b/packages/design-system/src/components/menu/list/defaultListItem.js
@@ -40,6 +40,7 @@ const ActiveIcon = styled(CheckmarkSmall)`
 const DefaultListItem = ({ option, isSelected, ...rest }, ref) => (
   <ListItem
     {...rest}
+    {...option}
     ref={ref}
     isSelected={isSelected}
     disabled={option.disabled}
--- a/packages/design-system/src/components/menu/list/defaultListItem.js
+++ b/packages/design-system/src/components/menu/list/defaultListItem.js
@@ -40,7 +40,6 @@ const ActiveIcon = styled(CheckmarkSmall)`
 const DefaultListItem = ({ option, isSelected, ...rest }, ref) => (
   <ListItem
     {...rest}
-    {...option}
     ref={ref}
     isSelected={isSelected}
     disabled={option.disabled}
--- a/packages/design-system/src/components/menu/list/components.js
+++ b/packages/design-system/src/components/menu/list/components.js
@@ -41,17 +41,21 @@ export const ListItemLabel = styled.li`
   align-items: center;
 `;
 
-export const ListItem = styled.li(
-  ({ disabled, theme }) => css`
-    position: relative;
-    display: grid;
-    grid-template-columns: 32px 1fr;
-    padding: 6px 8px;
-    margin: 4px 8px;
+export const ListItem = styled.li`
+  position: relative;
+  display: grid;
+  grid-template-columns: 32px 1fr;
+  padding: 6px 8px;
+  margin: 4px 8px;
+  align-items: center;
+  & > span:before {
+    color: ${({ theme }) => theme.colors.fg.tertiary};
+    content: ${({ $level }) =>
+      `'${Array.from({ length: $level }, () => '— ').join('')}'`};
+  }
+  ${({ disabled, theme }) => css`
     border-radius: ${theme.borders.radius.small};
-    align-items: center;
     cursor: ${disabled ? 'default' : 'pointer'};
-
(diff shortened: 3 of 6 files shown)


packages/design-system/src/components/menu/list/defaultListItem.js
```diff
@@ -40,6 +40,7 @@ const ActiveIcon = styled(CheckmarkSmall)`
 const DefaultListItem = ({ option, isSelected, ...rest }, ref) => (
   <ListItem
     {...rest}
+    {...option}
```
swissspidy: Does this cause any props to accidentally end up as DOM attributes?

packages/design-system/src/components/menu/index.js
```diff
@@ -15,6 +15,7 @@
  */
 
 export { Menu } from './menu';
+export { DefaultListItem } from './list';
```
swissspidy: I don't see where this new export is used?

maxyinger: Ahh It was for a failed earlier approach I documented in the Technical Details section. thanks for pointing out. will cleanup

packages/design-system/src/components/menu/list/defaultListItem.js
BrittanyIRL: 🤔  so, I'm hesitant about spreading the option on the default list item used by dropDowns. 

I get why you did this - the only "custom" part of the list items for this case are the dashes. 

Just for the sake of spreading knowledge, overriding the rendered list item is pretty straightforward, there's a demo of it in the dropdown storybook (last component in that file) 

You'd just need to do add this to the `DropDown` 
```
renderItem={forwardRef(({ option, ...rest }, ref) => (
                  <CustomStyledListItem
                    ref={ref}
                    width={option.width}
                    $level={option.$level}
                    {...rest}
                  >
                    {option.label}
                  </CustomStyledListItem>
                ))}
```

Another option is `menuStylesOverride`but you need to know the `$level` so again, I get it. 

Lastly, and I think this might work given the parameters, you could just change the label rendered on the list items. 
```
 options={dropdownCategories.map(
                  ({ $level, label, ...category }) => ({
                    ...category,
                    label: `${Array.from({ length: $level }, () => '— ').join(
                      ''
                    )} ${label}`,
                  })
                )}
```

This has the added benefit of the selected option showing the depth of level just like in the categories list view 
Screen Shot 2021-10-08 at 10 41 36 AM


Thoughts?

maxyinger: >  overriding the rendered list item is pretty straightforward, there's a demo of it in the dropdown storybook

Yea this was my second approach, I basically did:
```
const StyledDefaultListItem = styled(DefaultListItem)`/* look for $level in here and apply styles */`
...
renderItem={StyledDefaultListItem}
```
However there was an issue with styled components where `${({ $level }) => $level * 2}px` some reason this `$level` was always read as undefined even tho the component was receiving it 🤷 

> Lastly, and I think this might work given the parameters, you could just change the label rendered on the list items.

Yea this was my first approach, but was trying to do it with padding and apply specific styles to the indent prefix. Happy to revert back to this. might be easiest with the design choice we settled on. (might just have to make the dashes white tho)

BrittanyIRL: yeah so the reason it was undefined is because `option` wasn't getting spread into the rendered item. 

hrmmmm ok, would a happy compromise be renaming `$level` to be something like `displayNestedDashes` or something like that so it's clear to future versions of us that are in the dropDown code and then not spreading `option` but instead passing that value in so that it's another variant of dropdown menu item that can be reused?

maxyinger: Just removed it so we don't have to worry about it

maxyinger: just went with that last recommended approach. looks good and removes that spreading issue

BrittanyIRL (approved):

PR #9349, opened 2021-10-11
Post Types: Create a new Post type base class

### LINKED ISSUE #9316: Post Types: Create a new Post type base class
## Feature Description
In #9047, we create a Taxonomy base class, designed to make creating taxonomies more generate and DRY. A base post type class could be created to perform a similar action. Using a similar format as [Taxonomy_Base](https://github.com/google/web-stories-wp/blob/main/includes/Taxonomy/Taxonomy_Base.php), repeated logic could moved moved to this base class. Examples. 

https://github.com/google/web-stories-wp/blob/26826bf3391645624c42d2498d9c239083ff590c/includes/Story_Post_Type.php#L122-L132

https://github.com/google/web-stories-wp/blob/26826bf3391645624c42d2498d9c239083ff590c/includes/Story_Post_Type.php#L134-L144

https://github.com/google/web-stories-wp/blob/26826bf3391645624c42d2498d9c239083ff590c/includes/Story_Post_Type.php#L146-L156

https://github.com/google/web-stories-wp/blob/26826bf3391645624c42d2498d9c239083ff590c/includes/Story_Post_Type.php#L219-L228

We could also take this opportunity, to get rid of the trait [Post_Type](https://github.com/google/web-stories-wp/blob/main/includes/Traits/Post_Type.php). We can change the signature of the method from this 
```
$this->get_post_type_cap( $this->story_post_type::POST_TYPE_SLUG, 'edit_posts' )
```
to
```
$this->story_post_type->cap( 'edit_posts' ) 
```

spacedmonkey:
## Summary

Create a new Post type base class. This allows for multiple traits to be remove and make dependencies between classes much clearer. 

## Relevant Technical Choices

Removed the Theme supports trait, as only used in one place. 
Remove post type trait as no longer needed. 
Register, unregister post types on activation, deactivation, site creation. 
Use get_slug where ever possible, as we have tighter typing control of this function. 
Keep class name, const and existing namespaces for backwards compatibility. 

## Testing Instructions

1.

## Checklist

---

Fixes #9316


includes/Traits/Theme_Support.php
```diff
@@ -26,15 +26,14 @@
...
 
-use Google\Web_Stories\Story_Post_Type;
+use Google\Web_Stories\Services;
 
 /**
  * Trait Theme_Support
  *
  * @package Google\Web_Stories\Traits
  */
 trait Theme_Support {
```
swissspidy: Just realized that this trait is only used in 1 place: the `Customizer` class.

There's no point in having this trait.

Can we just move all that code in this trait to that class so we can properly inject the `Story_Post_Type` instance there?

tests/phpunit/integration/tests/Admin/Editor.php
```diff
@@ -77,6 +77,16 @@ class Editor extends TestCase {
...
+	/**
+	 * @var \Google\Web_Stories\Story_Post_Type
+	 */
+	private $spt;
+
+	/**
+	 * @var \Google\Web_Stories\Page_Template_Post_Type
+	 */
+	private $ptpt;
+
```
swissspidy:
```suggestion
	/**
	 * @var \Google\Web_Stories\Story_Post_Type
	 */
	private $story_post_type;

	/**
	 * @var \Google\Web_Stories\Page_Template_Post_Type
	 */
	private $page_template_post_type;

```

includes/Admin/Dashboard.php
```diff
@@ -374,16 +374,25 @@ public function enqueue_assets( $hook_suffix ) {
...
+					'post_type' => $this->story_post_type->get_slug(),
 				],
 				'post-new.php'
 			)
 		);
 
+		$classic_wp_list_url = admin_url(
+			add_query_arg(
+				[
+					'post_type' => 'web-story',
```
spacedmonkey:
```suggestion
					'post_type' => $this->story_post_type->get_slug(),
```

includes/Admin/Dashboard.php
```diff
@@ -374,16 +374,25 @@ public function enqueue_assets( $hook_suffix ) {
...
 			add_query_arg(
 				[
-					'post_type' => $this->story_post_type::POST_TYPE_SLUG,
+					'post_type' => $this->story_post_type->get_slug(),
 				],
 				'post-new.php'
 			)
 		);
 
+		$classic_wp_list_url = admin_url(
```
swissspidy: We actually removed this functionality

includes/Embed_Base.php
```diff
@@ -48,15 +48,24 @@ abstract class Embed_Base extends Service_Base {
...
 	 *
-	 * @param Assets $assets            Assets instance.
+	 * @param Assets          $assets            Assets instance.
+	 * @param Story_Post_Type $story_post_type Story_Post_Type instance.
 	 */
-	public function __construct( Assets $assets ) {
-		$this->assets = $assets;
+	public function __construct( Assets $assets, Story_Post_Type $story_post_type ) {
+		$this->assets          = $assets;
+		$this->story_post_type = $story_post_type;
```
spacedmonkey: Needed downstream. See https://github.com/google/web-stories-wp/pull/9349/files#diff-1574251e9291dbc26794d35e1fc7ab7089f835389d95beced8ce3ebcaac68bbeR185

swissspidy: We should look into simplifying the whole embed logic. It's too confusing with the current setup.

includes/REST_API/Stories_Controller.php
```diff
@@ -404,18 +403,24 @@ protected function add_response_headers( WP_REST_Response $response, WP_REST_Req
...
+		}
+
+		if ( current_user_can( $post_type->cap->edit_posts ) ) {
 			$statuses['draft']   = 'draft';
 			$statuses['future']  = 'future';
 			$statuses['pending'] = 'pending';
 		}
 
-		if ( $this->get_post_type_cap( $this->post_type, 'publish_posts' ) ) {
+		if ( current_user_can( $post_type->cap->publish_posts ) ) {
```
swissspidy, line 418: Much more comprehensible with `current_user_can` 👍

swissspidy: Actually, I see there's the `get_cap()` method in the base class. Why not use that?

includes/Admin/Editor.php
```diff
@@ -138,6 +166,19 @@ public function register() {
 		add_filter( 'use_block_editor_for_post_type', [ $this, 'filter_use_block_editor_for_post_type' ], 10, 2 );
 	}
 
+	/**
+	 * Get the list of service IDs required for this service to be registered.
+	 *
+	 * Needed because the story post type needs to be registered first.
```
swissspidy:
```suggestion
	 * Needed because the story and page template post types need to be registered first.
```

includes/Post_Type_Base.php
```diff
@@ -59,15 +164,14 @@ protected function get_post_type_rest_base( string $slug ): string {
...
+	 * @since 1.14.0
 	 *
-	 * @param string $slug The post type slug.
 	 * @param string $cap Capability name.
 	 *
 	 * @return bool
 	 */
-	protected function get_post_type_cap( string $slug, string $cap ): bool {
-		$capability_name = $this->get_post_type_cap_name( $slug, $cap );
+	public function get_cap( string $cap ): bool {
```
swissspidy: This is to similar to `get_cap_name` below.

Can we rename this to `has_cap` instead?

includes/Post_Type_Base.php
```diff
@@ -59,15 +164,14 @@ protected function get_post_type_rest_base( string $slug ): string {
 	/**
 	 * Get post type capability on the post type slug and name.
```
swissspidy:
```suggestion
	 * Determines whether the current user has a specific capability for this post type.
```

includes/Post_Type_Base.php
```diff
@@ -59,15 +164,14 @@ protected function get_post_type_rest_base( string $slug ): string {
...
 	 *
-	 * @since 1.7.0
+	 * @since 1.14.0
 	 *
-	 * @param string $slug The post type slug.
 	 * @param string $cap Capability name.
 	 *
 	 * @return bool
```
swissspidy:
```suggestion
	 * @return bool Whether the user has the capability.
```

includes/Post_Type_Base.php
```diff
@@ -79,15 +183,14 @@ protected function get_post_type_cap( string $slug, string $cap ): bool {
 	/**
 	 * Get post type capability name on the post type slug and name.
 	 *
-	 * @since 1.7.0
+	 * @since 1.14.0
 	 *
-	 * @param string $slug The post type slug.
 	 * @param string $cap Capability name.
 	 *
 	 * @return string|false
```
swissspidy:
```suggestion
	 * @return string|false Capability name if found, false otherwise.
```

includes/Post_Type_Base.php
```diff
@@ -79,15 +183,14 @@ protected function get_post_type_cap( string $slug, string $cap ): bool {
 	/**
 	 * Get post type capability name on the post type slug and name.
```
swissspidy:
```suggestion
	 * Returns a specific post type capability name.
```

includes/Post_Type_Base.php
```diff
@@ -24,31 +26,134 @@
...
 
 /**
- * Trait Post_Type
+ * Class Post_Type_Base.
  *
- * @package Google\Web_Stories\Traits
+ * @package Google\Web_Stories
  */
-trait Post_Type {
+abstract class Post_Type_Base extends Service_Base implements PluginActivationAware, PluginDeactivationAware, SiteInitializationAware {
```
swissspidy, line 47: Should this class perhaps already live in the `Post_Type` namespace? Or should we do that move later?

swissspidy, line 47: What about `SiteRemovalAware`? `Taxonomy_Base` implements it to unregister the taxonomy in that case. So rewrites can be flushed after.

swissspidy:
I see some `use Google\Web_Stories\Traits\Post_Type;` statements in the code still

includes/REST_API/Stories_Controller.php
swissspidy: Ah, right, this is in `Stories_Controller` nvm. As discussed, this is fine.

includes/Embed_Base.php
spacedmonkey: In another ticket.

includes/Post_Type_Base.php
spacedmonkey: We spoke about this in our meetings. Happy to do this, but does it makes to move just this class and not the others?

spacedmonkey:
Would you be against converting `Stories_Script_Data` into a class and injectoring it?

includes/Embed_Base.php
swissspidy: Absolutely another ticket. Just something to think about.

includes/Post_Type_Base.php
swissspidy: Yeah that's the part I am not sure about.

But we can probably do everything together later on.

swissspidy:
> Would you be against converting `Stories_Script_Data` into a class and injectoring it?

Seems reasonable 👍

includes/Renderer/Stories/FieldStateFactory/Factory.php
```diff
@@ -45,7 +45,7 @@ class Factory implements FieldStateFactory {
 	/**
 	 * Injector instance.
 	 *
-	 * @var Injector|Service Locale instance.
+	 * @var Injector Locale instance.
```
swissspidy:
```suggestion
	 * @var Injector Injector instance.
```

includes/Renderer/Stories/FieldStateFactory/Factory.php
```diff
@@ -65,7 +65,7 @@ public function __construct( Injector $injector ) {
 	 *
 	 * @param string $view View Type.
 	 *
-	 * @return FieldState
+	 * @return FieldState|object
```
swissspidy: Interesting, when is this a simple `object`? 🤔

includes/Post_Type_Base.php
swissspidy: As per our discussion yesterday we can remove `SiteRemovalAware` from taxonomies. The site is getting removed at that point anyway, no need to unregister content types first.

includes/Embed_Base.php
```diff
@@ -48,6 +48,7 @@ abstract class Embed_Base extends Service_Base {
 	 */
 	protected $assets;
 
+
```
swissspidy: ?

includes/Admin/Customizer.php
```diff
@@ -591,4 +611,101 @@ public function render_stories(): string {
...
+	}
+
+	/**
+	 * Get theme support configuration.
+	 *
+	 * @since 1.14.0
+	 *
+	 * @return array
+	 */
+	public function get_stories_theme_support() : array {
```
swissspidy, line 649: 👍

includes/Stories_Script_Data.php
```diff
@@ -104,4 +124,20 @@ protected function fields_states() : array {
...
 	}
+
+	/**
+	 * Get supported layouts for web stories.
+	 *
+	 * @since 1.14.0
+	 *
+	 * @return mixed|void
+	 */
+	public function get_layouts() {
```
swissspidy, line 137: 👍

includes/Embed_Base.php
spacedmonkey:
```suggestion
```

includes/Renderer/Stories/FieldStateFactory/Factory.php
spacedmonkey: Because I am using the injector here, it will PHPstan think it might be a object and not a fieldstate object. Just phpstan being dumb here.

includes/Embed_Base.php
spacedmonkey:
```suggestion
```

includes/Admin/Editor.php
```diff
@@ -138,6 +165,19 @@ public function register() {
...
+	 * Get the list of service IDs required for this service to be registered.
+	 *
+	 * Needed because the story and page template post types need to be registered first.
+	 *
+	 * @since 1.14.0
+	 *
+	 * @return string[] List of required services.
+	 */
+	public static function get_requirements(): array {
+		return [ 'story_post_type' ];
```
swissspidy: Why is the page template post type service no longer listed here?

Now the PHPDoc above isn't accurate anymore

includes/Renderer/Stories/FieldStateFactory/Factory.php
swissspidy: Let's use PHPDoc annotations instead to teach PHPStan what the expected return types are:

```php
/**
 * Returns field state for the provided view type.
 *
 * @since 1.5.0
 *
 * @param string $view View Type.
 *
 * @return FieldState
 */
public function get_field( $view = 'grid' ) {
	switch ( $view ) {
		case 'grid':
			/**
			 * GridView instance.
			 *
			 * @var FieldState
			 */
			return $this->injector->make( GridView::class );
		case 'list':
			/**
			 * ListView instance.
			 *
			 * @var FieldState
			 */
			return $this->injector->make( ListView::class );
		case 'circles':
			/**
			 * CircleView instance.
			 *
			 * @var FieldState
			 */
			return $this->injector->make( CircleView::class );
		case 'carousel':
			/**
			 * CarouselView instance.
			 *
			 * @var FieldState
			 */
			return $this->injector->make( CarouselView::class );
		default:
			/**
			 * CircleView instance.
			 *
			 * @var FieldState $default_field_state
			 */
			$default_field_state = $this->injector->make( CircleView::class );

			/**
			 * Filters the field state object.
			 *
			 * This depicts
			 *
			 * @since 1.5.0
			 *
			 * @param FieldState $default_field_state Field state object.
			 */
			$field_state = apply_filters( 'web_stories_default_field_state', $default_field_state );

			return $field_state instanceof FieldState ? $field_state : $default_field_state;
	}
}
```

swissspidy (approved):
Just a few minor nits but otherwise looks good

includes/Admin/Editor.php
spacedmonkey: Because I got an error when adding  page template post type service.

includes/Page_Template_Post_Type.php
```diff
@@ -154,4 +161,15 @@ public function register() {
...
+	/**
+	 * Get the action priority to use for registering the service.
+	 *
+	 * @since 1.14.0
+	 *
+	 * @return int Registration action priority to use.
+	 */
+	public static function get_registration_action_priority(): int {
+		return 11;
+	}
```
spacedmonkey: Without this, the editor doesn't work. There is a some upstream bug here. I don't want to tackle this.

includes/Renderer/Stories/FieldStateFactory/Factory.php
spacedmonkey: Good catch!

includes/Admin/Editor.php
spacedmonkey: Fixed the issue with changing the property on the page template post type.

PR #9390, opened 2021-10-12
Allow re-triggering adaptive text style

### LINKED ISSUE #9206: Retriggering smart coloring
## Feature Description

Quick action + design panel action for retriggering smart logic (e.g. after position change)

Todo:
- Add a new item to the quick action menu of a single text element and using the text magic icon from [Figma](https://www.figma.com/file/Ha6kk9yXDN1rB80NzcDf2J/_Editor---Design-System-%22Style-Guide%22?node-id=3441%3A80829).
- When clicking on the icon, the auto-styling logic is triggered on the single text element.
- Note that this feature will be applied to single text element only, multi-selection will not support it!
- Add a button to the design panel for triggering this auto-styling, [see Figma](https://www.figma.com/file/3t6ZYIDqzRpyBCRsZ1G7wJ/Adaptive-Text-Colors-Button?node-id=0%3A1), use the tooltip `Adaptive text colors` for it.
- Update the Color input to use icon for opacity ([Icon](https://www.figma.com/file/Ha6kk9yXDN1rB80NzcDf2J/_Editor---Design-System-%22Style-Guide%22?node-id=3441%3A80829) )
Screenshot 2021-10-11 at 18 03 56

## Additional Context
Icon for the quick action menu:
Screenshot 2021-10-06 at 11 42 32

Design panel:

Screenshot 2021-10-08 at 12 50 30

miina:
## Summary
This PR implements allowing re-triggering adaptive text colors for a single text element.

## User-facing changes
adaptive

Screenshot 2021-10-13 at 13 19 12

## Testing Instructions

1. Assign some image or dark color background to the page that would have very low contrast with default text color (black)
2. Now add a text element
3. Click on the adaptive text colors button in the design panel, verify the text color is changed as expected (see the screenshot/gif below for detecting the button)
4. Verify the same using the text element's quick action menu
5. Verify that the button is disabled for multi-selection of text elements.

No

- Adds new tracking for using `auto_style_text` from the quick action menu
- Adds tracking `trackEvent('auto_style_text');` for using the button from the design panel.

No

## Checklist

---

Fixes #9206


packages/story-editor/src/components/panels/design/textStyle/karma/textStyle.other.karma.js
```diff
@@ -77,7 +79,55 @@ describe('Text Style Panel', () => {
...
+      await fixture.events.keyboard.type('000');
+      await fixture.events.keyboard.press('Tab');
+
+      // Add text element.
+      await fixture.events.click(fixture.editor.library.textAdd);
+      await fixture.events.click(
+        fixture.editor.inspector.designPanel.textStyle.adaptiveColor
+      );
+
+      // @todo This does not wait at all and just continues with the selector result being null.
```
miina: @barklund Do you have any thoughts on why this would not wait and just continue right away even though the result of the query selector is `null` (and later it won't)? It doesn't time out but just doesn't wait at all.

I tried a few different stuff, e.g. just waiting without `expect`, having different `expect` conditions etc.

Perhaps I'm missing something obvious. Thoughts?

miina: Ok, looks like when an error is thrown inside `waitFor`, it will try again until it doesn't error, otherwise it doesn't wait for the `expect` to succeed. Had to rework this to throw an error if the correct value is not found.

packages/story-editor/src/utils/useApplyTextAutoStyle.js
```diff
@@ -0,0 +1,51 @@
...
+        : null;
+      const content = fgColor
+        ? setColor(element.content, { color: fgColor })
+        : element.content;
+      onApply({ ...highlightProps, content });
+    }
+  };
+
+  return applyTextAutoStyle;
+}
```
barklund: Why does the whole hook take a callback as an argument rather than have `applyTextAutoStyle` return a promise with the resulting styles?

barklund: Hmm, I kind of get that the whole hook returns a handler in this way, that you can directly apply to a button's `onClick` or similar, but it still feels a bit weird from a more generalist perspective. But I'm okay with it.

The callback should perhaps have a different name though. It's not an event, it's the actual handler effectuating the auto styling to some element somewhere.

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
```diff
@@ -420,6 +422,12 @@ const useQuickActions = () => {
     ]
   );
 
+  const applyTextAutoStyle = useApplyTextAutoStyle(selectedElement, (props) => {
+    updateElementsById({
+      elementIds: [selectedElement?.id],
+      properties: props,
+    });
+  });
```
barklund: Nit:

```js
  const applyTextAutoStyle = useApplyTextAutoStyle(selectedElement, (properties) =>
    updateElementsById({
      elementIds: [selectedElement?.id],
      properties,
    })
  );
```

barklund (approved):
Works well, just a few (ignorable 😄) comments on structure.

packages/story-editor/src/utils/useApplyTextAutoStyle.js
miina: Hm, we could rename it to `updater` or `elementUpdater` if that seems to make more sense. Thoughts?

PR #9397, opened 2021-10-12
Taxonomies: Hierarchical input accessibility updates

### LINKED ISSUE #9274: Taxonomy: Hierarchical Focus Trap
## Bug Description

The category list should be ignored with tabs, but if we use the arrow keys we should be able to navigate between options. That way we don’t have to tab through all the categories when trying to get to the `Tags` section

While here, let's update the HTML to be more semantic as well. @barklund points out: 

* Use a non-collapsible, multi-selectable [tree view](https://www.w3.org/TR/wai-aria-practices-1.1/#TreeView) for the widget.
  * A `role="tree"` around the whole thing.
  * A `role="group"` around any list of sibling category nodes
  * A `role="treeitem"` around any single category node
* Remember to set `aria-multiselectable="true"` and set `aria-selected` specifically to true or false for the individual selectable elements (if attribute is missing, it means that the element cannot be selected, so actively set to false).
* Note the quite complex list of keyboard interactions, but let's only implement what we strictly need:
  * We can ignore expanding/collapsing nodes, because we only support all nodes expanded for now (I think that's okay).
  * Also ignore the type-ahead instructions, we have a search field.
  * And finally ignore the somewhat complex instructions for multi-selecting while navigating with <kbd>Shift</kbd> or <kbd>Ctrl</kbd>.
  * We'll only support moving up and down using arrow keys (and optionally end/home) as well as toggling selected state with <kbd>Space</kbd> (not <kbd>Enter</kbd> for some reason 🤷 ).
* Also note initial focus should go to the first element if none are selected or the first selected element if any are selected when receiving focus.
  * Optionally, if receiving focus in reverse tab order (shift-tabbing from a subsequent element), the last element if none selected or the last selected element should receive focus. As [this is not actually in the WAI-ARIA recommendations](https://www.w3.org/TR/wai-aria-practices-1.1/#:~:text=when%20a%20multi-select%20tree%20receives%20focus), it's most definitely skippable.
* Refer to the [example implementations](https://www.w3.org/TR/wai-aria-practices-1.1/#examples-13) for further guidance. 

Fixing this here in tandem with the keyboard nav should make things easier.

## Expected Behaviour

Once you get to the categories/hierarchical component that lists all relevant taxonomies, you should use the arrow keys to navigate through the options not the tab key, if tab is hit then it should move focus past the list of terms.

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

samwhale:
## Context

Bug bash discovered that the hierarchical input needed some accessibility updates.

## Summary

Accessibility updates for the hierarchical input.

## Relevant Technical Choices

- The `listbox` role doesn't allow for nested focusable options, so the [`tree`](https://www.w3.org/TR/wai-aria-practices-1.1/#TreeView) role works better here.

## To-do

n/a

## User-facing changes

Small visual change: the border will be highlighted when any checkbox is focused

## Testing Instructions

1. Hitting `Tab` will no longer go through the options, instead it will move away from the checkbox list
2. Tabbing into the checkbox area from the search box will do one of two things:
    - If there is a selected category then it will be focused
    - If there is not a selected category, then the first category will be focused
3. Tab into the checkbox list, change the focus with the arrow keys. Then tab back out and in. The checkbox that was focused before tabbing out should still receive focus.
4. Should be usable with a screen reader

no

no

no

## Checklist

---

Fixes #9274


packages/story-editor/src/components/form/hierarchical/index.js
```diff
@@ -203,20 +313,38 @@ const HierarchicalInput = ({
...
       {showOptionArea && (
         <Border>
           <DirectionAware>
-            <CheckboxArea role="group">
+            <CheckboxArea
+              id="checkbox_list"
+              ref={checkboxListRef}
+              tabIndex={0}
+              onFocus={handleListboxFocus}
+              role="listbox"
```
samwhale: `listbox` role requires direct descendants to have the `option` role. Our `option`s are nested in groups

On `listbox`:


On `option`s:

packages/story-editor/src/components/form/hierarchical/index.js
```diff
@@ -60,73 +65,101 @@ const Label = styled(Text).attrs({
...
  * Renders a checkbox and all children of the checkbox.
  *
  * @param {Object} option The option to render.
- * @return {Node} The rendered option and children
+ * @param {Object} option.optionRefs Ref used to store refs to checkboxes.
+ * @param {number} option.level The indentation level.
+ * @return {Node} The rendered option and children.
  */
-const Option = (option) => {
+const Option = ({ optionRefs = { current: {} }, level = 0, ...option }) => {
```
maxyinger: should we be forwarding this as a ref? or is there a reason you chose against that?

packages/story-editor/src/components/form/hierarchical/index.js
```diff
@@ -142,15 +174,78 @@ const HierarchicalInput = ({
...
+    if (document.activeElement === checkboxListRef.current) {
+      if (!flattenedOrderedOptions.length) {
+        return;
+      }
+
+      const selectedId = flattenedOrderedOptions.find(
+        (option) => option.checked
+      )?.id;
+
+      setFocusedCheckboxId(selectedId || selectedId === 0 ? selectedId : -1);
```
maxyinger: 💭 wondering if using browser api reduces ref complexity here?
```suggestion
checkboxListRef.current?.querySelector('input[aria-checked="true"]')?.focus();
```

packages/story-editor/src/components/form/hierarchical/index.js
```diff
@@ -142,15 +174,78 @@ const HierarchicalInput = ({
...
+        case KEYS.DOWN:
+          if (currentSelectedIndex < flattenedOrderedOptions.length - 1) {
+            setFocusedCheckboxId(
+              flattenedOrderedOptions[currentSelectedIndex + 1].id
+            );
+          }
+          break;
+        default:
+          evt.propagate();
+          break;
```
maxyinger:
```suggestion
      if(!checkboxListRef.current.contains(document.activeElement)) {
        return;
      }
      switch (evt.key) {
        case KEYS.UP:
          document.activeElement.previousElementSibling?.focus();
          return;
        case KEYS.DOWN:
          document.activeElement.nextElementSibling?.focus();
          return;
        default:
          evt.propagate();
          return;
```

packages/story-editor/src/components/form/hierarchical/index.js
```diff
@@ -208,17 +317,31 @@ const HierarchicalInput = ({
...
+              onFocus={handleListboxFocus}
+              role="tree"
+              aria-activedescendant={
+                focusedCheckboxId !== -1
+                  ? buildOptionId(focusedCheckboxId)
+                  : undefined
+              }
+              aria-multiselectable
+            >
               {filteredOptions.length ? (
```
maxyinger: 💭 if we just used `flattenedOrderedOptions` instead of `filteredOptions` could we remove the recursive call inside of `<Option />` entirely?

packages/story-editor/src/components/form/hierarchical/utils/index.js
```diff
@@ -152,3 +152,28 @@ const countOptions = (option) => {
...
+    result.push(option);
+
+    // add children to result
+    if (Array.isArray(options) && options.length) {
+      result = result.concat(flattenOptionsTree(options));
+    }
+  }
+
+  return result;
+};
```
maxyinger: Sorry for communication issues here, I think I've already created the same util except it adds a `$level` prop to the option:
https://github.com/google/web-stories-wp/blob/8defe9e1d61c2906ec0ad0b0ebc6abdee094bbfb/packages/story-editor/src/components/panels/document/taxonomies/utils.js#L21-L29

Would that work here? Could we unify the two? or is there a difference between them I'm not seeing?

maxyinger: There's some tests for it too if you want to look at the spec some more

maxyinger (approved):
Curious to get your thoughts on some ideas I pitched to reduce complexity. Works great as is tho if you're not vibing with any of the feedback 🍻

packages/story-editor/src/components/form/hierarchical/index.js
samwhale: Man, this is so cool! I wish this worked out of the box but the `input`s are the ones being focused, and they have no siblings

Screen Shot 2021-10-14 at 10 09 39 AM

I got it to work (kind of) by traveling up and down using `parent` and `children`. However, I feel like that's pretty brittle in the event where the markup changes or we add/remove elements + wrappers.

💭 I could at least climb out until we hit a `li` element and then we could search the sibling for an input?

packages/story-editor/src/components/form/hierarchical/index.js
samwhale: Hmm let me see if I can make this work

packages/story-editor/src/components/form/hierarchical/index.js
samwhale: Ah so the `optionRefs` an one single ref that stores the refs of each checkbox by id. This is independent of the tree structure

As long as you know the id of the option, you can query the ref object for the element using the id.

---

This was being used for finding the next/previous input to focus. The flattened array stored everything in order, so using that I could find the next/previous element.

I'd grab the id and then using that object could find the element in question and focus it.

It's complicated for just focusing, so it'd be nice to simplify it

packages/story-editor/src/components/form/hierarchical/index.js
maxyinger: ah yea. didn't realize they were nested. ignore my feedback. my bad.

packages/story-editor/src/components/form/hierarchical/index.js
maxyinger: > This is independent of the tree structure

Ah yup, this is the part I wasn't completely grokking. Thanks for clarifying

packages/story-editor/src/components/form/hierarchical/index.js
```diff
@@ -60,105 +60,184 @@ const Label = styled(Text).attrs({
...
-          name={optionName}
-          onChange={(evt) => onChange(evt, option)}
-        />
-        <Label htmlFor={optionId}>{label}</Label>
-      </CheckboxContainer>
-      {options?.map((child) => (
-        <StepContainer key={child.id}>
-          <Option onChange={onChange} {...child} />
-        </StepContainer>
-      ))}
```
samwhale, line 122: No more recursion!

packages/story-editor/src/components/form/hierarchical/index.js
```diff
@@ -60,105 +60,184 @@ const Label = styled(Text).attrs({
...
+ * @param {string} props.label Input label
+ * @param {string} props.noOptionsText Text to display no options are found
+ * @param {Array} props.options Array of options
+ * @param {Function} props.onChange Change event handler. Passed to checkboxes.
+ * @return {Node} The input
+ */
 const HierarchicalInput = ({
   className,
+  inputValue,
+  onInputChange,
```
samwhale, line 169: Options are filtered externally in `makeFlatOptionTree`, so the state was lifted up.

packages/story-editor/src/components/panels/document/taxonomies/HierarchicalTermSelector.js
```diff
@@ -247,12 +231,37 @@ function HierarchicalTermSelector({
     }
   }, [toggleFocus]);
 
+  const orderedCategories = useMemo(
+    () => makeFlatOptionTree(categories, searchText),
```
samwhale, line 235: Options are filtered here now

packages/story-editor/src/components/form/hierarchical/index.js
```diff
@@ -180,21 +182,6 @@ const HierarchicalInput = ({
...
-
-  const filteredOptions = useMemo(
-    () => filterOptionsByLabelText(filteredOptionTree, inputText),
-    [filteredOptionTree, inputText]
-  );
-
-  const flattenedOrderedOptions = useMemo(
-    () => flattenOptionsTree(filteredOptions),
-    [filteredOptions]
-  );
```
samwhale: No longer need these, thanks max!

packages/story-editor/src/components/form/hierarchical/test/hierarchical.js
```diff
@@ -39,67 +39,16 @@ const OPTIONS = [
   { id: 5, label: 'zebra fish', checked: true },
 ];
 
-const OPTIONS_WITH_NO_CHILDREN = OPTIONS.filter((option) => !option.parent);
-
 describe('Hierarchical', () => {
-  it('typing in the list should filter the first level options', () => {
```
samwhale, line 45: Don't need to test this here - options are filtered outside this component

packages/story-editor/src/components/form/hierarchical/utils/index.js
```diff
@@ -129,51 +129,32 @@ export const filterOptionsByLabelText = (options, labelText) =>
...
-
-/**
- * Recursively count the number of `options` in an array of nested objects.
- *
- * @param {Array.<Object>} tree Array of options to count.
- * @return {number} the count of all options
- */
-export const getOptionCount = (tree = []) =>
-  tree.reduce((count, option) => count + countOptions(option), 0);
+function flattenOptionTree(optionTree = [], level = 0) {
```
samwhale, line 139: just moved

packages/story-editor/src/components/form/hierarchical/utils/index.js
```diff
@@ -129,51 +129,32 @@ export const filterOptionsByLabelText = (options, labelText) =>
...
-
-    // add children to result
-    if (Array.isArray(options) && options.length) {
-      result = result.concat(flattenOptionsTree(options));
-    }
-  }
-
-  return result;
-};
+export function makeFlatOptionTree(categories, labelText = '') {
```
samwhale, line 156: just moved from taxonomy utils

packages/story-editor/src/components/form/hierarchical/index.js
```diff
@@ -108,70 +104,76 @@ const buildOptionName = (name) => `hierarchical_term_${name}`;
...
+  inputValue,
+  onInputChange,
   label,
   noOptionsText,
   options,
   onChange,
   ...inputProps
 }) => {
-  const [inputText, setInputText] = useState('');
+  // const [inputText, setInputText] = useState('');
```
maxyinger: remove?

packages/story-editor/src/components/form/hierarchical/index.js
```diff
@@ -108,70 +104,76 @@ const buildOptionName = (name) => `hierarchical_term_${name}`;
...
+        value={value}
+        checked={checked}
+        name={optionName}
+        onChange={(evt) => onChange(evt, option)}
+        onBlur={onBlur}
+        onFocus={onFocus}
+        tabIndex={-1}
+      />
+      <Label htmlFor={optionId}>{label}</Label>
+    </CheckboxContainer>
```
maxyinger, line 137: 😍

packages/story-editor/src/components/form/hierarchical/stories/hierarchical.js
```diff
@@ -66,9 +73,12 @@ export const _default = () => {
   return (
     <Wrapper>
       <Hierarchical
+        inputValue={searchText}
+        onInputChange={(val) => setSearchText(val)}
```
maxyinger: memoize with `useCallback()` ?

maxyinger (approved):
Looks great. Very clean. Awesome job

packages/story-editor/src/components/form/hierarchical/stories/hierarchical.js
samwhale: Can do! This is just the story file but I can update it

packages/story-editor/src/components/form/hierarchical/index.js
samwhale: Ended up doing something different - so this doesn't apply anymore

packages/story-editor/src/components/form/hierarchical/index.js
samwhale: No prob!

BrittanyIRL (approved):
Nice! thanks for the notes too, makes it really clear.

PR #9398, opened 2021-10-12
Bugfix:  Stuck Swatch Tooltips in Safari

### LINKED ISSUE #9188: Color Tooltips get stuck on screen in Safari
## Bug Description

On Safari, the color tooltips ("Gradients not allowed for text") sometimes get stuck on the screen. Seems to happen on mouse-out with high speed.

image17

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

BrittanyIRL:
## Context

Sometimes in safari this would happen: 
https://user-images.githubusercontent.com/58738530/135152882-c6aba1cb-0fad-4cd4-b5af-9307e84aa656.gif 

## Summary

Where disabled swatch tooltips (gradients for text elements, for example) would get stuck open. Only happens in Safari. Only happens on the color swatches. 

## Relevant Technical Choices

So, after doing some digging it appears that the `disabled` button `Swatch` elements in Safari were swallowing up the pointer event and preventing the `Tooltip`'s `onPointerLeave` from firing (sometimes, not every time). It feels likely that this is due to the 💅  fancy styles happening to get the swatches to look right - with various pseudo classes. This is solved by adding `pointer-events: none` to both the button when it's disabled and also the `:after` pseudo class for buttons that are enabled.  As it's only a problem in the swatch it felt better to solve for this at the swatch level rather than impact the whole tooltip - especially since it would mean swapping to a less performant event set for tracking the cursor. 

While in there trying to figure out what was going on I added in some swatches wrapped in tooltips to storybook, decided to leave them for testing/debugging. Also noticed some redundant props/styles so took those out. 

## User-facing changes

Now in Safari if you hover over some disabled color swatches they should close when you are no longer hovering on the swatch. 

safari swatch tooltips

## Testing Instructions

Verify that the swatch tooltips don't get stuck open any more.

1.

No

No

No

## Checklist

---

Fixes #9188

### DIFF
--- a/packages/design-system/src/components/swatch/swatch.js
+++ b/packages/design-system/src/components/swatch/swatch.js
@@ -61,6 +61,7 @@ const SwatchButton = styled.button`
     css`
       opacity: 0.4;
       cursor: default;
+      /* pointer-events none here fixes an edge case in Safari when swatch is within a tooltip #9188 */
       pointer-events: none;
     `}
   ${({ theme }) => themeHelpers.focusableOutlineCSS(theme.colors.border.focus)};
@@ -68,6 +69,7 @@ const SwatchButton = styled.button`
   ::after {
     content: '';
     position: absolute;
+    /* pointer-events none here fixes an edge case in Safari when swatch is within a tooltip #9188 */
     pointer-events: none;
     left: 0;
     top: 0;
--- a/packages/design-system/src/components/swatch/swatch.js
+++ b/packages/design-system/src/components/swatch/swatch.js
@@ -44,7 +44,7 @@ const Transparent = styled.div`
   ${themeHelpers.transparentBg}
 `;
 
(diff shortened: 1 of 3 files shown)


miina (approved):
⭐  Thanks for the cleanup!

maxyinger (approved):
😎

BrookeGraham (approved):
🛸

packages/design-system/src/components/swatch/swatch.js
```diff
@@ -61,12 +61,16 @@ const SwatchButton = styled.button.attrs({ type: 'button' })`
     css`
       opacity: 0.4;
       cursor: default;
+      /* pointer-events none here fixes an edge case in Safari when swatch is within a tooltip #9188 */
+      pointer-events: none;
```
samwhale, line 65: noice

samwhale (approved):
noice!

PR #9403, opened 2021-10-13
Video Optimization: Reuse existing resource if exists

### LINKED ISSUE #9384: Video Optimization: Reuse existing resource if exists
### Scenario

* You have an original unoptimized video and place it on the canvas
* You optimize it
* You create a new story and place the original unoptimized video on the canvas
* You optimize it
* This will result in another optimized version being created instead of just using the original

### How to possibly fix this

1. When optimizing an existing video, fetch `/web-stories/v1/media/<resource_id>` and check for `meta.web_stories_optimized_id`.
2. If `meta.web_stories_optimized_id` is set, grab `/web-stories/v1/media/<optimized_id>` and replace the existing resource with the optimized one

swissspidy:
## Context

Previously, when the same video has been used in multiple stories and you go edit every story to optimize the video, multiple optimized versions have been generated instead of just reusing the existing version.

## Summary

When optimizing an existing video via the checklist, the editor now first checks whether an optimized version of the resource already exists and just uses that if the case.

## Relevant Technical Choices

* New `getOptimizedMediaById` API callback to find the optimized version of an existing resource.
* JSDoc improvements
* Updates tests

## To-do

N/A

## User-facing changes

N/A

## Testing Instructions

1. In the WordPress media library, upload a large video
2. Create a new story ("Story A") and insert the video there
3. Create another new story ("Story B") and insert the video there as well
3. In story A, add some empty pages so that the checklist shows up
4. In story A, optimize the video via the checklist
5. Preview story A, view source, and note down the video URL there
5. In story B, optimize the video via the checklist
6. Verify that the second "optimization" does not result in a video upload being started, but the checklist warning disappearing
7. Preview story B, view source, verify that video URL is the same as for the one in story A

No

No

No

## Checklist

---

Fixes #9384

### DIFF
--- a/packages/wp-story-editor/src/api/media.js
+++ b/packages/wp-story-editor/src/api/media.js
@@ -101,7 +101,7 @@ export async function getOptimizedMediaById(config, mediaId) {
   const result = await apiFetch({ path });
 
   if (result?.meta?.web_stories_optimized_id) {
-    return getMediaById(config, result?.meta?.web_stories_optimized_id);
+    return getMediaById(config, result.meta.web_stories_optimized_id);
   }
 
   return null;
--- a/packages/wp-story-editor/src/api/media.js
+++ b/packages/wp-story-editor/src/api/media.js
@@ -74,7 +74,7 @@ export function getMedia(
  *
  * @param {Object} config Configuration object.
  * @param {number} mediaId Media ID.
- * @return {Promise} Media object promise.
+ * @return {Promise<import('@web-stories-wp/media').Resource>} Media object promise.
  */
 export function getMediaById(config, mediaId) {
   const path = addQueryArgs(`${config.api.media}${mediaId}/`, {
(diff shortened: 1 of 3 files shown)


packages/wp-story-editor/src/api/media.js
```diff
@@ -85,13 +85,35 @@ export function getMediaById(config, mediaId) {
...
+export async function getOptimizedMediaById(config, mediaId) {
+  const path = addQueryArgs(`${config.api.media}${mediaId}/`, {
+    context: 'edit',
+    _fields: 'meta.web_stories_optimized_id',
+  });
+
+  const result = await apiFetch({ path });
+
+  if (result?.meta?.web_stories_optimized_id) {
+    return getMediaById(config, result?.meta?.web_stories_optimized_id);
```
spacedmonkey: Does this need any await? 
```suggestion
    const response = await getMediaById(config, result?.meta?.web_stories_optimized_id);
    return response;
```

spacedmonkey: Also, do we need the questions marks here, we know it is set, from the line above.

spacedmonkey (approved):
Approved with a nitpic. 

We should 100% of do the same with muting.

packages/wp-story-editor/src/api/media.js
swissspidy: No `await` needed because the Promise is just returned right away and this is an async function.

In short, here `return await Promise` and `return Promise` are identical.

spacedmonkey:
```suggestion
    return getMediaById(config, result.meta.web_stories_optimized_id);
```
Not needed as the line above does this check.

PR #9363, opened 2021-10-11
Layers Panel: Layer Action Scrim Container

maxyinger:
## Context
Part of design updates for layers panel.

## Summary
Adds a container that will house the layer actions (soon to be built). This container appears on hover. Right now there are no actions other than the disabled `lock/unlock` action on the Background element.

## Relevant Technical Choices
Went with a css predominant approach to minimize having to hold button state in js.

Also went with a box-shadow implementation for the scrim instead of an element with a gradient to simplify positioning and nodes.

## User-facing changes
NA (Small right aligned scrim appears on hover in the layers panel now)

## Testing Instructions
Interact with the layers panel and see that there's now a little right aligned scrim that appears on hover. Other than that, just make sure nothing had any regressions.

- [x] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1.

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [x] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [x] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #9140

PR #9405, opened 2021-10-13
Layers panel: Delete layer from panel

### LINKED ISSUE #9138: Layers panel: Delete layer from panel
## Description

Add a button to each layer in the layers panel that allows the layer to be deleted from the layers panel. Behavior should mimic using the delete key on a layer.

https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor-Sprint-Feature-Level-Designs_?node-id=9946%3A422172
Designs as of Oct 4, 2021 (left button)
Screen Shot 2021-10-04 at 11 24 50 AM

## To Do

- [x] Final UX from design team

## Acceptance Criteria 

- new button on each layer panel (besides background) that allows the layer to be deleted, should also work with keyboard
- updated tests

maxyinger:
## Context
Part of new layer actions

## Summary
This adds the `delete` action to the layers panel:
Screen Shot 2021-10-13 at 9 56 21 AM

## Relevant Technical Choices
NA

## To-do
~add a lil karma test~

## User-facing changes
Users now have a delete layer button in the layers panel:
Screen Shot 2021-10-13 at 9 56 21 AM

## Testing Instructions

1. Open the editor 
2. Make sure there are some elements
3. Go to the Layers panel
4. See that you can delete layers with the delete action
5. see that there's no delete action on the bg element

## Checklist

---

Fixes #9138

### DIFF
--- a/packages/story-editor/src/components/panels/design/layer/karma/layer.karma.js
+++ b/packages/story-editor/src/components/panels/design/layer/karma/layer.karma.js
@@ -85,30 +85,19 @@ describe('Layer Panel', () => {
   });
 
   it('should be able to delete elements with delete action', async () => {
-    const insertElement = await fixture.renderHook(() => useInsertElement());
-    await fixture.act(() => {
-      insertElement('text', {
-        font: TEXT_ELEMENT_DEFAULT_FONT,
-        content: 'element_a',
-        x: 40,
-        y: 40,
-        width: 250,
-      });
-
-      insertElement('text', {
-        font: TEXT_ELEMENT_DEFAULT_FONT,
-        content: 'element_b',
-        x: 40,
-        y: 40,
-        width: 250,
-      });
-    });
+    await fixture.editor.library.textTab.click();
+    await fixture.events.click(fixture.editor.library.text.preset('Title 1'));
+    await fixture.events.click(fixture.editor.library.text.preset('Title 2'));
 
     expect(layerPanel.layers.length).toBe(3);
-    const elementALayer = layerPanel.getLayerByInnerText('element_a');
+    const elementALayer = layerPanel.getLayerByInnerText('Title 1');
     await fixture.events.hover(elementALayer);
     const deleteElementAButton = within(elementALayer).getByLabelText('Delete');
     await fixture.events.click(deleteElementAButton);
+
     expect(layerPanel.layers.length).toBe(2);
+    expect(
+      layerPanel.layers.every((layer) => layer.innerText !== 'Title 1')
+    ).toBeTrue();
   });
 });
--- a/packages/story-editor/src/components/panels/design/layer/karma/layer.karma.js
+++ b/packages/story-editor/src/components/panels/design/layer/karma/layer.karma.js
@@ -13,6 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+/**
+ * External dependencies
+ */
+import { within } from '@testing-library/react';
 
 /**
  * Internal dependencies
@@ -78,5 +82,33 @@ describe('Layer Panel', () => {
       up(),
     ]);
     expect(section.clientHeight).toBe(initialHeight);
+  });
+
+  it('should be able to delete elements with delete action', async () => {
+    const insertElement = await fixture.renderHook(() => useInsertElement());
+    await fixture.act(() => {
+      insertElement('text', {
+        font: TEXT_ELEMENT_DEFAULT_FONT,
+        content: 'element_a',
+        x: 40,
+        y: 40,
+        width: 250,
+      });
+
+      insertElement('text', {
+        font: TEXT_ELEMENT_DEFAULT_FONT,
+        content: 'element_b',
+        x: 40,
+        y: 40,
+        width: 250,
+      });
+    });
+
+    expect(layerPanel.layers.length).toBe(3);
+    const elementALayer = layerPanel.getLayerByInnerText('element_a');
+    await fixture.events.hover(elementALayer);
+    const deleteElementAButton = within(elementALayer).getByLabelText('Delete');
+    await fixture.events.click(deleteElementAButton);
+    expect(layerPanel.layers.length).toBe(2);
   });
 });
--- a/packages/story-editor/src/components/panels/design/layer/layer.js
+++ b/packages/story-editor/src/components/panels/design/layer/layer.js
@@ -131,17 +131,13 @@ const LayerDescription = styled.div`
 `;
 
 const IconWrapper = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: flex-start;
   position: absolute;
-  right: 0px;
+  right: 0;
   width: 32px;
   aspect-ratio: 1;
-
-  /*
-   * have to manually center this because this element is taller
-   * than its parent causing flex to not properly center it.
-   */
-  top: 50%;
-  transform: translateY(-50%);
 
   svg {
     display: block;
--- a/packages/story-editor/src/components/panels/design/layer/layer.js
+++ b/packages/story-editor/src/components/panels/design/layer/layer.js
@@ -24,6 +24,7 @@ import {
   BUTTON_TYPES,
   Icons,
   themeHelpers,
+  Tooltip,
 } from '@web-stories-wp/design-system';
 
 /**
@@ -92,6 +93,7 @@ const LayerButton = styled(Button).attrs({
       background: ${theme.colors.interactiveBg.secondaryPress};
       + * {
         --background-color: ${theme.colors.interactiveBg.secondaryPress};
+        --selected-hover-color: ${theme.colors.interactiveFg.brandHover};
       }
     `}
 
@@ -154,6 +156,7 @@ const LayerContentContainer = styled.div`
 
 const LayerAction = styled(Button).attrs({
   type: BUTTON_TYPES.PLAIN,
+  tabIndex: -1,
 })`
   position: relative;
   aspect-ratio: 1;
@@ -184,9 +187,15 @@ const LayerAction = styled(Button).attrs({
     background: var(--background-color);
   }
 
-  :disabled,
-  :hover {
+  :disabled {
     color: ${({ theme }) => theme.colors.fg.secondary};
+  }
+
+  :hover {
+    color: var(
+      --selected-hover-color,
+      ${({ theme }) => theme.colors.fg.secondary}
+    );
   }
 
   & + & {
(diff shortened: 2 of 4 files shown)


packages/story-editor/src/components/panels/design/layer/layer.js
```diff
@@ -31,41 +36,80 @@ import { LayerText } from '../../../../elements/shared/layerText';
...
+  align-items: center;
+  height: 100%;
+  top: 0;
+  right: 0;
+  padding-right: 6px;
+
+  --background-color: ${({ theme }) =>
+    theme.colors.interactiveBg.secondaryNormal};
+  background-color: var(--background-color);
+  box-shadow: 0px 0px 15px 20px var(--background-color);
```
swissspidy:
```suggestion
  ${({ theme }) => `
  background-color: ${theme.colors.interactiveBg.secondaryNormal};
  box-shadow: 0px 0px 15px 20px ${theme.colors.interactiveBg.secondaryNormal};
  `
```

We don't usually use CSS custom properties in our styled components, especially not when it's just to save some characters.

swissspidy: OK, I see you're using the custom property further below for hover&active state. But perhaps you could just use this way of writing there too.

packages/story-editor/src/components/panels/design/layer/layer.js
```diff
@@ -105,38 +156,111 @@ const LayerContentContainer = styled.div`
...
+    position: absolute;
+    width: 32px;
+    height: auto;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+  }
+
+  /*
+   * override base button background color so we can recieve the
```
swissspidy:
```suggestion
   * override base button background color so we can receive the 
```

packages/story-editor/src/components/panels/design/layer/layer.js
```diff
@@ -86,14 +130,21 @@ const LayerDescription = styled.div`
...
 `;
 
-const LockIconWrapper = styled.div`
+const IconWrapper = styled.div`
   position: absolute;
-  right: 0;
-  display: flex;
-  align-items: center;
-  justify-content: flex-start;
+  right: 0px;
```
swissspidy:
```suggestion
  right: 0;
```

packages/story-editor/src/components/panels/design/layer/layer.js
```diff
@@ -86,14 +130,21 @@ const LayerDescription = styled.div`
   color: ${({ theme }) => theme.colors.fg.primary};
 `;
 
-const LockIconWrapper = styled.div`
+const IconWrapper = styled.div`
   position: absolute;
-  right: 0;
-  display: flex;
-  align-items: center;
-  justify-content: flex-start;
```
swissspidy: I just added this today 😅  so it doesn't work as expected? or just not anymore?

swissspidy:
Should this have been based against #9363? 🤔

packages/story-editor/src/components/panels/design/layer/layer.js
maxyinger: IMO, there's less room for error with the css var approach because if color assignment ever updates, with the ccs var we only have to update 1 css property vs targeting every property and child property that consumes it.

Consider the actions down below that consume this `--background-color` to create the relevant focus styles.

What's the aversion to the css var here? I see some benefits, but I'm struggling to see the negatives of using it

packages/story-editor/src/components/panels/design/layer/layer.js
maxyinger: For learning purposes, is there any benefits to omitting units when the value is 0 ? or is this just a stylistic preference?

packages/story-editor/src/components/panels/design/layer/layer.js
maxyinger: I was running into issues with flex centering children that were taller than the parent container earlier. see lines 139-144.

Happy to revert to this if it works

packages/story-editor/src/components/panels/design/layer/layer.js
swissspidy: just stylistic :) I thought our stylelint config actually flags this, but apparently it doesn't 🤔

packages/story-editor/src/components/panels/design/layer/layer.js
swissspidy: No aversion. I just didn't see that you used it in so many places and thought it was only here, in which case it was quite unnecessary.

But for this use case I suppose it's useful 👍

packages/story-editor/src/components/panels/design/layer/layer.js
```diff
@@ -227,15 +242,22 @@ function Layer({ layer }) {
...
       </LayerButton>
       <ActionsContainer>
-        {/*
-         *@TODO #9137 #9138 add layer actions here. only reason we conditionally
-         * render right now is to maintain visual continuity until
-         * the actual actions are present.
-         */}
-        {isBackground && (
-          <LayerAction disabled>
+        {isBackground ? (
```
miina, line 278: Q: If the background is background media, should it be possible to remove that as well?

packages/story-editor/src/components/panels/design/layer/karma/layer.karma.js
```diff
@@ -79,4 +83,32 @@ describe('Layer Panel', () => {
     ]);
     expect(section.clientHeight).toBe(initialHeight);
   });
+
+  it('should be able to delete elements with delete action', async () => {
+    const insertElement = await fixture.renderHook(() => useInsertElement());
+    await fixture.act(() => {
```
miina: Nit: generally in Karma tests we've been trying to imitate the actual user behaviour to make it as close to real experience as possible, meaning clicking on actual buttons to create elements instead of using hooks and insertion by code. E.g in this case a simple way to add two text elements with different content could be something like this:
```
await fixture.editor.library.textTab.click();
await fixture.events.click(fixture.editor.library.text.preset('Title 1'));
await fixture.events.click(fixture.editor.library.text.preset('Title 2'));
```

Thoughts?

miina (approved):

packages/story-editor/src/components/panels/design/layer/karma/layer.karma.js
samwhale: Ah this is good to know! I've been doing it wrong 😮‍💨

packages/story-editor/src/components/panels/design/layer/karma/layer.karma.js
```diff
@@ -79,4 +83,32 @@ describe('Layer Panel', () => {
...
+        width: 250,
+      });
+    });
+
+    expect(layerPanel.layers.length).toBe(3);
+    const elementALayer = layerPanel.getLayerByInnerText('element_a');
+    await fixture.events.hover(elementALayer);
+    const deleteElementAButton = within(elementALayer).getByLabelText('Delete');
+    await fixture.events.click(deleteElementAButton);
+    expect(layerPanel.layers.length).toBe(2);
```
samwhale: Nit: May be a little sturdier if we check to see the correct element was removed?

samwhale (approved):
✨

packages/story-editor/src/components/panels/design/layer/karma/layer.karma.js
maxyinger: Oh awesome. Yea that makes sense. tests the exposed UI more than the internals. Thanks for the example. will update!

packages/story-editor/src/components/panels/design/layer/layer.js
maxyinger: I'll post on the design channel. that makes sense to me

maxyinger: https://xwp.slack.com/archives/C01ESBB8ZN1/p1634243974171700

Thread here. Design currently doesn't want a delete action for the background.

packages/story-editor/src/components/panels/design/layer/layer.js
```diff
@@ -46,7 +51,10 @@ const ActionsContainer = styled.div`
   box-shadow: 0px 0px 15px 20px var(--background-color);
 `;
 
-const LayerContainer = styled.div`
+const LayerContainer = styled.div.attrs({
+  // Because the layer panel is aria-hidden, we need something else to select by
+  'data-testid': 'layer-option',
```
swissspidy, line 75: The `useTemplate.js` e2e test is looking for `button[data-testid="layer-option"]`. Since the attribute is now added to the `LayerContainer ` div instead of the button, the test obviously needs updating.

swissspidy: Fixed in e3cc6754a2d10bbd97957d50ac4bcce2a1923f4f

packages/story-editor/src/components/panels/design/layer/layer.js
```diff
@@ -227,15 +238,22 @@ function Layer({ layer }) {
...
+        {isBackground ? (
+          <LayerAction
+            aria-label={__('Locked', 'web-stories')}
+            aria-describedby={layerId}
+            disabled
+          >
             <Icons.LockClosed />
           </LayerAction>
+        ) : (
+          <LayerAction
```
swissspidy: I think this causes aXe tests in the e2e tests to fail.

> ARIA hidden element must not contain focusable elements

https://dequeuniversity.com/rules/axe/4.3/aria-hidden-focus

Adding `tabindex={-1}` might help pass this test

swissspidy:
@littlemilkstudio When you hover over the delete button, its color changes, making the color contrast very very low.
Screenshot 2021-10-18 at 12 43 19

We should improve this to pass WCAG AA recommendations.

Also, perhaps a tooltip could be added?

samwhale:
> @littlemilkstudio When you hover over the delete button, its color changes, making the color contrast very very low. Screenshot 2021-10-18 at 12 43 19
> 
> We should improve this to pass WCAG AA recommendations.
> 
> Also, perhaps a tooltip could be added?

We could darken the color of the svg on hover too 🤷

maxyinger:
yea, there are a lot of hover color bg variations with the buttons being nested inside the layer. I also am not in love with the fact that the hover and disabled colors are the same, but that's what was specified in the designs 🤷 

Thanks for all the feedback tho. will go through and update and post anything in the design channel that's needed color wise

packages/story-editor/src/components/panels/design/layer/layer.js
maxyinger: oh awesome. thanks!

maxyinger:
Looks like there's some css issues on safari and I'm gonna need to use a psuedo element with a gradient. in the process of updating

PR #9409, opened 2021-10-14
Dashboard: Make API response uniform by using camel case

mohdsayed:
## Context

Part 2 of https://github.com/google/web-stories-wp/pull/9406

## Summary

Dashboard API response [uses a mix of came case and snake case fields](https://github.com/google/web-stories-wp/blob/8defe9e1d61c2906ec0ad0b0ebc6abdee094bbfb/packages/wp-dashboard/src/api/utils/reshapeStoryObject.js#L54-L76). Now that we are going to document the expected API response from all API callbacks  in the next phase of decoupling, the fields casing should be uniform.

## Testing Instructions

1.

## Checklist

---


swissspidy:
I think we need to decide on the desired casing first.

For example in the editor we recently changed a lot of stuff to camelCase and even added a `snakeCaseToCamelCase` function to help with that. But even before that, most of the fields were in camelCase

I think snake_case is fine for the API fields, but the transform/reshape functions in the dashboard should turn them into camelCase

packages/dashboard/src/app/api/useStoryApi.js
```diff
@@ -62,8 +63,10 @@ const useStoryApi = () => {
         // Maybe not making a lot of sense to expect fetchedStoryIds from the api callbacks response.
         // However the order of ids get changed if we try to create that array here
         // which may ( or may not ) cause some regression. @todo Reflect on fetchedStoryIds again in next phase.
+        const response = await fetchStoriesCallback(queryParams);
+
         const { stories, fetchedStoryIds, totalPages, totalStoriesByStatus } =
-          await fetchStoriesCallback(queryParams);
+          snakeToCamelCaseObjectKeys(response);
```
swissspidy: Should that transformation happen on the `wp-dashboard` side instead?

packages/wp-dashboard/src/api/utils/reshapeStoryObject.js
```diff
@@ -61,17 +61,17 @@ export default function reshapeStoryObject(originalStoryData) {
...
-    featuredMediaUrl,
-    editStoryLink,
-    previewLink,
+    bottom_target_action: edit_story_link,
+    featured_media_url,
+    edit_story_link,
+    preview_link,
     link,
-    originalStoryData,
+    original_story_data: originalStoryData, // @todo Do we need it?
```
swissspidy: I don't think we do. Looks like the only references are in `updateStory` to get the author ID (which we could just add to the `author` field above by turning that into an object instead of just the author name) and in `duplicateStory` to get the story ID (which is already available as `id`)

packages/dashboard/src/app/api/useStoryApi.js
swissspidy: On the editor side, `snakeCaseToCamelCase` is also used in `wp-story-editor`

mohdsayed: Doing it in `wp-dashboard` would mean that the core dashboard API response is expecting camel casing instead of snake casing.

I am in favour of using snake casing for the API fields, and I thought you also prefer the same.

> I think snake_case is fine for the API fields, but the transform/reshape functions in the dashboard should turn them into camelCase

I think it should also change for story editor
> On the editor side, snakeCaseToCamelCase is also used in wp-story-editor

swissspidy: > I am in favour of using snake casing for the API fields, and I thought you also prefer the same.

I meant more the API fields you get from the REST API, sorry for not being precise.

> I think it should also change for story editor

So you mean moving the current camelCase logic from `wp-story-editor` to `story-editor`?

-----

Sounds like we need to discuss this a bit first 🙂 

I think we both agree that we should have some consistency across dashboard and editor.

And it makes sense for the `wp-*` packages to bring the REST API responses into a format more suitable for the core apps. `transformGetStoryResponse` in `wp-story-editor` is a good example of that.
That function does not really change casing though.

On the other hand we also use `getResourceFromAttachment` in `wp-story-editor` that turns everything into camel case.

I'd find it odd if `wp-*` does _some_ transformation on the data, but then the core apps do some _more_ transformation on top of that.

Shouldn't all transformations live in the `wp-*` packages? So that we only transform the responses once.

WDYT?

----

Aside: some consistent naming would be good too. In the editor we use `transform...()` functions, but in the dashboard they're named `reshape..*`

mohdsayed: I think I am a little lost here. 

Sorry I still couldn't understand what you meant by that
> I meant more the API fields you get from the REST API

For example, if I think of the core `story-editor` in isolation, ( when I am integrating it with some other platform, not using `wp-story-editor` ), if the `wp-story-editor` is supposed to do the transformation of casing from snake to camel inside  `wp-story-editor` inside `.then()` for example here

https://github.com/google/web-stories-wp/blob/e7fc7071543bc4a768d15e653c9baf50a8c814a2/packages/wp-story-editor/src/api/story.js#L79-L88

Would it not mean that we are expecting camel case fields in response from `getStoryById` inside ~`wp-story-editor`~  `story-editor` ? 

Or do you want the expected API response fields to use camel casing.



I agree it is odd.
> I'd find it odd if wp-* does some transformation on the data, but then the core apps do some more transformation on top of that.

swissspidy: > Would it not mean that we are expecting camel case fields in response from `getStoryById` inside  `story-editor` ?

It would.

So I am proposing to do this change across the board. Curious to hear your opinion on it.

1. WordPress REST API returns snake_case
2. `wp-story-editor`/`wp-dashboard` transforms fields from snake_case into camelCase format expected by `story-editor`/`dashboard` (`getResourceFromAttachment` is prior art for this) 
3. `story-editor`/`dashboard` expect and use camelCase everywhere

mohdsayed: Got it, thanks for clarifying. I think it should be fine. My reason for leaning towards the snake case for the core editor/dashboard was inspired by [WordPress REST API](https://developer.wordpress.org/rest-api/reference/posts/) fields, but I think camel casing should be fine and it would simplify things a little in the core packages. 👍

swissspidy (approved):

PR #9416, opened 2021-10-15
Design panel: add context menu to layer panel

### LINKED ISSUE #9139: Layers Panel: Context menus
## Description

Harness the right click context menus that are available on the canvas by type and show them in the layers panel when a layer is right clicked on. 

Figma: https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor-Sprint-Feature-Level-Designs_?node-id=9954%3A422464
Designs as of Oct 4, 2021
Screen Shot 2021-10-04 at 11 36 45 AM

## Acceptance Criteria 

- right clicking on layers in layer panel should show context menu for that layer type/layer 
- updated tests

samwhale:
## Context

Add our fancy dandy context menu to the layer panel

## Summary

Adds the right click menu to the layer panel. Right clicking on a layer will open the context menu in the same way as right clicking on an element in the canvas.

**NOTE**: Right clicking a layer and the right clicking somewhere else re-opens the _same_ right click menu, since the selected element won't change. This is very apparent in the layer panel. If a user wants to right click a different layer while the context menu is open then it'll re-open the same menu where the user clicked.

This is the same functionality as the right click menu in the canvas.

Clicking an action in the menu will perform that action on the selected element.

## Relevant Technical Choices

Selecting multiple layers and `shift + right click` would still open the context menu. Added some code to stop that nonsense.

## To-do

n/a

## Testing Instructions

1. Right clicking a layer should open the context menu for that element type
2. All right click actions should work as intended
3. Selecting multiple layers while holding `Shift` and then right clicking one while holding `Shift` should not open our cusom context menu 

no

no

no

## Checklist

---

Fixes #9139

### DIFF
--- a/packages/story-editor/src/app/rightClickMenu/provider.js
+++ b/packages/story-editor/src/app/rightClickMenu/provider.js
@@ -174,22 +174,29 @@ function RightClickMenuProvider({ children }) {
    *
    * @param {Event} evt The triggering event
    */
-  const handleOpenMenu = useCallback((evt) => {
-    evt.preventDefault();
-    evt.stopPropagation();
+  const handleOpenMenu = useCallback(
+    (evt) => {
+      if (selectedElements.length > 1) {
+        return;
+      }
 
-    dispatch({
-      type: ACTION_TYPES.OPEN_MENU,
-      payload: {
-        x: evt?.x,
-        y: evt?.y,
-      },
-    });
+      evt.preventDefault();
+      evt.stopPropagation();
 
-    trackEvent('context_menu_action', {
-      name: 'context_menu_opened',
-    });
-  }, []);
+      dispatch({
+        type: ACTION_TYPES.OPEN_MENU,
+        payload: {
+          x: evt?.x,
+          y: evt?.y,
+        },
+      });
+
+      trackEvent('context_menu_action', {
+        name: 'context_menu_opened',
+      });
+    },
+    [selectedElements]
+  );
 
   /**
    * Close the menu and reset the tracked position.
--- a/packages/story-editor/src/app/rightClickMenu/provider.js
+++ b/packages/story-editor/src/app/rightClickMenu/provider.js
@@ -186,8 +186,8 @@ function RightClickMenuProvider({ children }) {
       dispatch({
         type: ACTION_TYPES.OPEN_MENU,
         payload: {
-          x: evt?.x,
-          y: evt?.y,
+          x: evt?.clientX,
+          y: evt?.clientY,
         },
       });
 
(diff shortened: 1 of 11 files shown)


packages/story-editor/src/app/rightClickMenu/provider.js
```diff
@@ -152,9 +151,6 @@ function RightClickMenuProvider({ children }) {
     })
   );
 
-  // Ref for attaching the context menu
-  const rightClickAreaRef = useRef();
-
```
samwhale, line 157: Doing this externally now so we can set up multiple refs for menus

packages/story-editor/src/components/panels/design/layer/layerList.js
```diff
@@ -70,12 +78,29 @@ function LayerPanel({ layers }) {
...
+    }
+
+    node.addEventListener('contextmenu', onOpenMenu);
+
+    return () => {
+      node.removeEventListener('contextmenu', onOpenMenu);
+    };
+    // On initial load there are no layers. Ref update won't trigger effect.
+    // So re-run when layers are fully updated.
+  }, [numLayers, onOpenMenu]);
```
maxyinger: okay, so this is protecting us from when this effect runs on mount and we're rendering `null` here, then we render the actual `LayerList`, but this effect has already run?

If so, I think this could be a good opportunity to use a callback ref to attach these event listeners, then you don't need to infer when `LayerList` is rendered, but can just have a callback associated with it directly:
https://medium.com/@teh_builder/ref-objects-inside-useeffect-hooks-eb7c15198780.

Another option that avoids this entirely could be by passing the `onContextMenu` prop directly on `LayerList`

maxyinger (approved):
Looks and works great! I think we can clean up some of those effect binding hooks by just passing `onContextMenu` directly.

Also I know there's a lot of complexity going into closing the context menu, but created a ticket here to do a little discovery task into possibly being able to right click from element to element and have the proper right click menu show up:
https://app.zenhub.com/workspaces/web-stories-5e94d3aced449034e1e9b226/issues/google/web-stories-wp/9426

packages/story-editor/src/components/panels/design/layer/layerList.js
samwhale: Thanks for teaching me about `onContextMenu`. That's exactly what I need!

BrittanyIRL (approved):
🕺

BrittanyIRL:
🤔  @samwhale on monday can you look into why karma's failing here before this gets merged?

samwhale:
hi hi!

> 🤔 @samwhale on monday can you look into why karma's failing here before this gets merged?

It seems like they are passing 🤔

PR #9418, opened 2021-10-15
Layers Panel: Duplicate layer from panel

### LINKED ISSUE #9137: Layers Panel: Duplicate layer from panel
## Description

Add a button to each layer in the layers panel that allows the layer to be duplicated from the layers panel. Behavior should mimic duplicating a layer from the right click menu.

https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor-Sprint-Feature-Level-Designs_?node-id=9946%3A422172

Designs as of Oct 4, 2021 (middle button)
Screen Shot 2021-10-04 at 11 24 50 AM

## Acceptance Criteria 

- new button on each layer panel (besides background) that allows the layer to be duplicated, should also work with keyboard
- updated tests

maxyinger:
## Context
Part of updates to the layers panel

## Summary
Adds duplicate action to the layers in the layers panel:
Screen Shot 2021-10-15 at 2 39 49 PM

## Relevant Technical Choices
Created a new story reducer action for this.

## To-do
- [ ] Discuss UX. It's a little hard to tell which layer was the original and which was the duplicate once you duplicate:
https://user-images.githubusercontent.com/35983235/137544933-8e23818f-650b-4af4-92d6-7165f4b6e322.mp4

## User-facing changes
Layers now have a `duplicate` action on them
Screen Shot 2021-10-15 at 2 39 49 PM

## Testing Instructions

1. Open a story
2. Make sure there are some elements in it
3. go to the layers panel
4. hover over any layer and see the duplicate layer button
5. see that the element duplicates when you press it

## Checklist

---

Fixes #9137


crypt096 (approved):

packages/story-editor/src/app/story/useStoryReducer/reducers/duplicateElementById.js
```diff
@@ -0,0 +1,79 @@
...
+ * External dependencies
+ */
+import { v4 as uuidv4 } from 'uuid';
+
+/**
+ * Internal dependencies
+ */
+import { getPastedCoordinates } from '../../../../utils/copyPaste';
+
+function duplicateElementById(state, { elementId }) {
```
miina, line 22: Hm, I suppose needing to use the specific index is the reason why we couldn't just use `addElement` with updated props?
Also, I wonder if it would be good to have an util for creating a clone of an element, including updating ID, animations, coordinates? Since we now have cloning, duplicating, copy-paste, and also duplicating a page (since it needs all elements cloned) all needing something similar. Perhaps you already checked that and it wasn't possible but checking just in case.

maxyinger: Yea they're all slightly different, but relatively similar. specific index here, needing to update selection for others, updating position on some, needing it to work across separate pages for others.

We could probably unify them, but feels a little out of scope for this ticket. would probably need a pretty augmented function to handle the different use cases, which at that point is it worth unifying? I wasn't sure.

Maybe worth creating a code quality ticket and looking into 🤷 . I'm open to whatever tho. What are your thoughts?

packages/story-editor/src/app/story/useStoryReducer/test/duplicatElementById.js
```diff
@@ -0,0 +1,170 @@
...
+      ...initialState,
+      pages: [
+        {
+          id: '111',
+          animations: [
+            { id: '1', targets: ['456'], type: 'wild_wacky_animation' },
+            {
+              id: expect.any(String),
+              targets: [newElement.id],
+              type: 'wild_wacky_animation',
```
samwhale, line 130: 😆

samwhale (approved):
Nice, works super bien. Agree with Miina - maybe it would be nice to have a shared util that can create a new element from an existing one - animations and all. It could just return the element and then we can decide what to do with it 🤔

maxyinger:
Cool. I can look into unifying them more. cloning might be a quicker win. copy & paste might require a little more work

maxyinger:
updated all the actions, but running into issues around `isDefaultPage`. Currently looking into this

miina (approved):
Nice work! ⭐  Thanks for creating a duplication util here.

Issue #7337, opened 2021-04-29
Dashboard Settings: Ability to pick publisher logo from WP media library


spacedmonkey:
## Feature Description

In the dashboard settings page, when a user clicks on upload logo, it opens a default file picker dialog. However, in the editor, it opens the WordPress media picker. User should be able to select existing files in the media library. 

### Current implementation 
<img width="1884" alt="Screenshot 2021-04-29 at 18 10 47" src="https://user-images.githubusercontent.com/237508/116591431-b79fa600-a916-11eb-9ac7-85f7c2cc1319.png">

### Publisher logo implementation in editor
<img width="1782" alt="Screenshot 2021-04-29 at 18 14 05" src="https://user-images.githubusercontent.com/237508/116591558-dc941900-a916-11eb-9441-ac5c2994d410.png">

## Additional Context

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

BrittanyIRL:
This seems like a good candidate to bundle with this feature in the roadmap: `Easily toggle between publisher logos from within the workspace` since that's to give continuity in publisher logo behavior between settings and the document panel of the workspace. This feature doesn't have a slated release yet.

spacedmonkey:
After working on https://github.com/google/web-stories-wp/pull/8067, I wonder if [mediaPicker](https://github.com/google/web-stories-wp/tree/main/assets/src/edit-story/components/mediaPicker), should be moved to it's own package. This makes sharing code between easier and it will make decoupling from WordPress a little easier too.

swissspidy:
Related: #4148

Shared utilities like that can be extracted as part of the ongoing decoupling work. It's not 100% clear yet how the code sharing there will look like, so this is not actionable just yet.

PR #9419, opened 2021-10-18
Decoupling: Move dashboard settings page to `wp-dashboard`

mohdsayed:
## Context

"Move settings page ( analytics, video optimization, etc ) to wp-dashboard" task from https://github.com/google/web-stories-wp/issues/2918

## Summary
As discussed ( https://github.com/google/web-stories-wp/pull/9419#issuecomment-945478745 ), this PR moves dashboard settings page to `wp-dashboard` including all its tests. 

As further discussed on this PR,  editor settings karma integration tests were removed and were compensated with e2e tests.

## Testing Instructions

1.

## Checklist

---


mohdsayed:
@swissspidy So far I have only moved the `EditorSettings` component to `wp-dashboard` in this PR. Looking for your opinion on how settings should be configured from `wp-dashboard`. 

I have a couple of options in mind

1. Move the entire [settings fields components](https://github.com/google/web-stories-wp/tree/main/packages/dashboard/src/app/views/editorSettings) to `wp-dashboard`. 
2. We move only WordPress-specific fields ( the whole component of [googleAnalytics](https://github.com/google/web-stories-wp/tree/main/packages/dashboard/src/app/views/editorSettings/googleAnalytics) and [archive](https://github.com/google/web-stories-wp/tree/main/packages/dashboard/src/app/views/editorSettings/archive) ) and export rest of the components from `dashboard` to be added in [EditorSettings](https://github.com/google/web-stories-wp/blob/d75f13d752107b43b4d3fa18c78dee6314be9678/packages/wp-dashboard/src/components/editorSettings/index.js#L316-L371) and we provide more configuration options to each field components ( like `<TelemetrySettings />`, `<PublisherLogoSettings />`  ) for providing API Callbacks and description/help texts.

swissspidy:
> @swissspidy So far I have only moved the `EditorSettings` component to `wp-dashboard` in this PR. Looking for your opinion on how settings should be configured from `wp-dashboard`.
> 
> I have a couple of options in mind
> 
> 1. Move the entire [settings fields components](https://github.com/google/web-stories-wp/tree/main/packages/dashboard/src/app/views/editorSettings?rgh-link-date=2021-10-18T08%3A08%3A23Z) to `wp-dashboard`.
> 2. We move only WordPress-specific fields ( the whole component of [googleAnalytics](https://github.com/google/web-stories-wp/tree/main/packages/dashboard/src/app/views/editorSettings/googleAnalytics?rgh-link-date=2021-10-18T08%3A08%3A23Z) and [archive](https://github.com/google/web-stories-wp/tree/main/packages/dashboard/src/app/views/editorSettings/archive?rgh-link-date=2021-10-18T08%3A08%3A23Z) ) and export rest of the components from `dashboard` to be added in [EditorSettings](https://github.com/google/web-stories-wp/blob/d75f13d752107b43b4d3fa18c78dee6314be9678/packages/wp-dashboard/src/components/editorSettings/index.js#L316-L371) and we provide more configuration options to each field components ( like `<TelemetrySettings />`, `<PublisherLogoSettings />`  ) for providing API Callbacks and description/help texts.

That is a very good question.

FWIW, the `adManagement` setting is also WP specific as it references Site Kit.

Related: the `publisherLogo` setting might use the WP media modal in the future (#7337), so another WP touchpoint.

I currently lean towards moving the entire settings field components, for these reasons:

1. It's easier if components are not split across to components, making it a bit unclear about what goes where
2. It provides the biggest flexibility for other platforms to have a fully custom settings page (or none at all)
2. It's cheap to revert this again and only move specific components again in the future.
3. In the future, with more platforms, we will get a better sense of how the components should be organized. Moving everything for now leaves all options option for us.

mohdsayed:
@swissspidy Now that we are moving [EditorSettings](https://github.com/google/web-stories-wp/tree/main/packages/dashboard/src/app/views/editorSettings) to `wp-dashboard`, its [karma tests](https://github.com/google/web-stories-wp/tree/main/packages/dashboard/src/app/views/editorSettings/karma) would also have to be moved, however, it cannot be tested from `wp-dashboard` because there is no setup for it.

Furthermore,  since the actual integration is happening in `wp-dashboard` ( same for `wp-story-editor` ), shouldn't the integration tests ( including fixtures ) be moved to `wp-dashboard`? That would solve the above problem.

swissspidy:
Thanks for raising this!

Just to clarify, you mean moving **all** the Karma tests to `wp-*`?



The goal of integration tests is to verify that several components/units work together as expected. So it's not related to the _integration into specific platforms_.

Especially when considering that we might have multiple platform integrations and split up the WP plugin and the core JS packages/apps into separate repositories, it's important to have these integration tests for the core apps without the WordPress/platform dependency.

For this reason, I don't think moving karma tests makes sense.

As an alternative we could analyze `editorSettings.karma.js` and identify tests which are better written as unit or e2e tests anyway, and rewrite them accordingly.

That said, some of them might indeed be better suited as Karma tests, begging the question again whether we should move all components... Maybe we should rethink that again?

Also, we might run into this situation again when refactoring the document sidebar in the story editor, but I don't think that part has huge Karma test coverage.

mohdsayed:
@swissspidy Great points

Yes, that's what I meant
> Just to clarify, you mean moving all the Karma tests to wp-*?

Exactly, I am wondering that the `wp-*` packages also have a good amount of code/components, how would the integration test cover if WordPress integration with all the pieces are working together as expected?
> The goal of integration tests is to verify that several components/units work together as expected.

I agree, therefore initially I was thinking of keeping two separate integration tests one for the core package and one for the `wp-*` 
>  it's important to have these integration tests for the core apps without the WordPress/platform dependency.

swissspidy:
> Exactly, I am wondering that the `wp-*` packages also have a good amount of code/components, how would the integration test cover if WordPress integration with all the pieces are working together as expected?
> 
> I agree, therefore initially I was thinking of keeping two separate integration tests one for the core package and one for the `wp-*`
> 

Yeah I am wondering that too.

I think adding Karma tests for `wp-*` is going to be too big of a maintenance burden, considering that we'll likely split up the repositories. Having Karma in the WP plugin repo just for these specific integrations doesn't sound worthwhile.

Hence my initial thinking to replace those Karma tests with some more unit & e2e tests.

The middle ground would be to do option 2️⃣  of https://github.com/google/web-stories-wp/pull/9419#issuecomment-945478745 and keep some Karma tests in `dashboard`.

packages/wp-dashboard/src/components/layout/index.js
```diff
@@ -22,11 +22,12 @@ import { InterfaceSkeleton } from '@web-stories-wp/dashboard';
  * Internal dependencies
  */
 import { useAdminSubMenu } from '../../effects';
+import EditorSettings from '../editorSettings';
 
 function Layout() {
   useAdminSubMenu();
 
-  return <InterfaceSkeleton />;
+  return <InterfaceSkeleton editorSettings={<EditorSettings />} />;
```
swissspidy: WDYT about something more generalized, e.g.

```suggestion
  return <InterfaceSkeleton additionalRoutes={[{ route: '/editor-settings', component: EditorSettings }]} />;
```

packages/wp-dashboard/src/components/editorSettings/test/editorSettings.js
```diff
@@ -19,19 +19,20 @@
  */
 import { fireEvent, within, screen } from '@testing-library/react';
 import { setAppElement } from '@web-stories-wp/design-system';
+import {
+  renderWithProviders,
+  AD_NETWORK_TYPE,
```
swissspidy: This const should be in the same package as the `adManagement` components, so `wp-dashboard`

packages/wp-dashboard/src/components/editorSettings/archive/test/archive.js
```diff
@@ -19,12 +19,11 @@
  */
 import { screen } from '@testing-library/react';
 import { FlagsProvider } from 'flagged';
+import { renderWithProviders, ARCHIVE_TYPE } from '@web-stories-wp/dashboard';
```
swissspidy: This const should also live in `wp-dashboard`

packages/dashboard/src/app/views/editorSettings/index.js
```diff
@@ -13,354 +13,5 @@
...
-          </Main>
-        </Layout.Scrollable>
-      </Wrapper>
-    </Layout.Provider>
-  );
-}
-
-export default EditorSettings;
+export { default as useTelemetryOptIn } from '../shared/useTelemetryOptIn';
+export { default as useMediaOptimization } from '../shared/useMediaOptimization';
```
swissspidy: I think these need to be moved to wp-dashboard as well. WDYT?

mohdsayed: Yes, moved that already

mohdsayed:
@swissspidy You are right, I think this makes sense for now.

> Hence my initial thinking to replace those Karma tests with some more unit & e2e tests.

packages/wp-dashboard/src/components/layout/index.js
mohdsayed: That's a good idea.

packages/wp-dashboard/src/components/editorSettings/test/editorSettings.js
mohdsayed: Yeah, this PR still needs good amount of work including these refactoring, will move.

swissspidy:
Coincidentally, #9482 is adding some more e2e tests for the settings page, which is great timing for this PR :-)

mohdsayed:
> Coincidentally, #9482 is adding some more e2e tests for the settings page, which is great timing for this PR :-)

I see there is not much overlap of  #9482 but surely it adds more to it. 👍 

I have compensated for the removed integration tests with e2e in https://github.com/google/web-stories-wp/pull/9419/commits/408eac38d6ed2b00aebdb217a84791e3251d768c and https://github.com/google/web-stories-wp/pull/9419/commits/b7684811f219c7724714d0052ca0281e74b164ed ( needs some fixes though ).

packages/wp-dashboard/src/components/editorSettings/test/editorSettings.js
mohdsayed: Done in https://github.com/google/web-stories-wp/pull/9419/commits/76a87cea2e2e474dc6a67146a33d04eba393cca9

packages/wp-dashboard/src/components/editorSettings/archive/test/archive.js
mohdsayed: Done in https://github.com/google/web-stories-wp/pull/9419/commits/76a87cea2e2e474dc6a67146a33d04eba393cca9

mohdsayed:
@swissspidy Sorry the PR inadvertently became very big, should have been divided in more than one PRs

mohdsayed:
Integration tests are passing locally, cannot tell why it is failing here.
E2E Tests is failing due to the following error, but this doesn't have any relation with this PR

image

tests/js/jest.config.js
```diff
@@ -47,6 +47,9 @@ export default {
     '<rootDir>/build',
     '<rootDir>/node_modules',
     '<rootDir>/vendor',
+    'story-editor',
+    'wp-story-editor',
+    'wp-dashboard',
```
swissspidy: Won't this cause all tests in these folders to be skipped? 🤔

swissspidy:
> Integration tests are passing locally, cannot tell why it is failing here.

Yeah unfortunately they are super flakey at the moment 😞 

> E2E Tests is failing due to the following error, but this doesn't have any relation with this PR
> image

IIRC Sam reported the same error when working on #9482

It might be related to toggling the telemetry checkbox, causing the Tag Manager script to be inserted, but without `crossorigin=anonymous`. I'll take a look...

tests/js/jest.config.js
mohdsayed: Ooops, this was just to run the test for `dashboard` locally, I accidentally pushed it.

mohdsayed: Reverted in https://github.com/google/web-stories-wp/pull/9419/commits/e66c90dd6fd4d2601c90bde490a4d6312f45d368

mohdsayed:
All passed, YAY!!  🥳

packages/e2e-tests/src/specs/dashboard/settings/adminUser/telemetryBanner.js
```diff
@@ -31,6 +31,14 @@ describe('Telemetry Banner', () => {
     });
   });
 
+  afterAll(async () => {
+    await visitSettings();
+    await disableCheckbox('[data-testid="telemetry-settings-checkbox"]');
+    await page.evaluate(() => {
+      localStorage.removeItem('web_stories_tracking_optin_banner_closed');
```
swissspidy, line 38: Good idea 👍

packages/e2e-tests/src/specs/dashboard/telemetryBanner.js
```diff
@@ -0,0 +1,88 @@
...
+  it('should not display the banner after it has been closed', async () => {
+    await expect(page).toMatch('Help improve the editor!');
+    await expect(page).toClick('[aria-label="Dismiss telemetry banner"]');
+    await page.reload();
+    await expect(page).not.toMatch('Help improve the editor!');
+  });
+
+  it('should keep focus on the checkbox when checking/unchecking via keyboard', async () => {
+    await page.waitForTimeout(300); // Because dashboard takes some time before banner appears.
+    const checkbox = await page.$('#telemetry-banner-opt-in');
```
swissspidy, line 72: Have you tried `expect(page).toMatchElement('#telemetry-banner-opt-in')`?

That will automatically wait a bit for the element to appear.

We should avoid using `waitForTimeout` if possible

swissspidy (approved):

packages/e2e-tests/src/specs/dashboard/telemetryBanner.js
swissspidy: Looks like that didn't work 🤔

mohdsayed: Yeah, unfortunately, it did not. Reverted the change.

swissspidy:
Whoops, looks like merging #9406 caused a bunch of merge conflicts here.

Sorry about that!

swissspidy:
I also just merged #9482 which adds some related e2e tests.

I like how you've organized the tests in this PR, so you might want to move those other tests over to match this structure. 

Overlapping tests can be removed.

mohdsayed:
@swissspidy Ready to be merged!

Issue #8736, opened 2021-08-19
Page Templates: Add Tests


BrittanyIRL:
## Feature Description

Add better test coverage for page templates. In Q1 2021 there were only 1 type of page template and the structure was a little more basic, we had a test to check page templates loading, but as of #8621 it's been discovered as really fragile and inaccurate if we're being honest at this point! 

Now we have default templates and saved templates. We should be using unit tests to check individual aspects of this panel, not the whole thing - that's what Karma tests are for (those are working fine). 

We should be checking: 
- Proper chips showing up, being clickable 
- Filtering templates 
- Do default templates load? 
- Do Saved templates load? 

Let's not do `pageTemplatesPane` tests, let's do tests in `defaultTemplates` and `savedTemplates` , so on. 

## Additional Context

This is follow up because updating react-virtual to 2.8.1 broke the pageTemplatesPane test, but it wasn't really doing anything. That test was removed in https://github.com/google/web-stories-wp/pull/8735

PR #9448, opened 2021-10-19
Page Templates: Fix karma tests

BrittanyIRL:
## Context

Page templates are getting swapped to be static images for perf, this allows some structural updates to the component to simplify it but because custom page templates haven't been updated yet we need to do this gradually. This PR fixes the tests for the update so that it can go to QA and we're unblocked. More refactor to come with #8736 

## Summary

Just updating some tests

## Relevant Technical Choices

- updates template data used for karma tests to include posters and slugs since we need those in the page layouts component now 
- default page template karma tests look for `button` now instead of `listitem` since that is where the necessary data is stored now for all checks. 
- updated image alt text to be less repetitive
- gave page layout buttons aria label so it doesn't catch the image and the span copy by default (making the karma tests fail) 
- swapped the keyboard nav test to just use the 'right' arrow instead of also down because it was being flakey otherwise with how the grid renders specific to karma.

## To-do

Once the custom page templates are updated we'll be able to simplify this whole thing.

## User-facing changes

None

## Testing Instructions

1.

## Checklist

---

Partially addresses #9093

### DIFF
--- a/packages/e2e-tests/src/specs/editor/pageTemplates.js
+++ b/packages/e2e-tests/src/specs/editor/pageTemplates.js
@@ -50,12 +50,12 @@ describe('Page Templates', () => {
     });
 
     // Adding a custom page template automatically switches to the "Saved Templates" view.
-
+    await expect(page).toMatch('Page Template saved.');
     await expect(page).toMatch('Saved templates');
     await expect(page).toMatchElement(
       '[aria-label="Page Template Options"] [role="listitem"]'
     );
 
-    await percySnapshot(page, 'Page Templates');
+    await percySnapshot(page, 'Custom Page Templates');
   });
 });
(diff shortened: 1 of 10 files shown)


BrookeGraham (commented):
looks good to me, but still some failing test 🤔

BrittanyIRL:
> looks good to me, but still some failing test 🤔

yeah, rerunning to see what's up - good news is that the page template karma tests are now passing!

BrookeGraham (approved):

packages/e2e-tests/src/specs/editor/pageTemplates.js
```diff
@@ -50,12 +50,12 @@ describe('Page Templates', () => {
...
     // Adding a custom page template automatically switches to the "Saved Templates" view.
-
+    await expect(page).toMatch('Page Template saved.');
     await expect(page).toMatch('Saved templates');
     await expect(page).toMatchElement(
       '[aria-label="Page Template Options"] [role="listitem"]'
     );
 
-    await percySnapshot(page, 'Page Templates');
+    await percySnapshot(page, 'Custom Page Templates');
```
BrookeGraham, line 59: nice.

PR #9440, opened 2021-10-19
Checklist: Copy updates for low resolution checks

BrittanyIRL:
## Context

Some copy was overlooked as the checklist morphed. Thanks to Pascal and users for reading and making note that it was redundant and unhelpful. 

## Summary

Updates the copy for image and videos with low resolution. Think the phrasing works well for singular or plural since it's indefinite. 

## To-do

Split out work mentioned as an aside in #9401 to take scale into consideration for image resolution (#9439)

## User-facing changes

| Old | New |
| --- | --- |
| Screen Shot 2021-10-18 at 4 56 57 PM | Screen Shot 2021-10-18 at 4 46 23 PM |

## Testing Instructions

Have a tiny image and a tiny video and see the resolution check get flagged and shows new copy.

- [ ] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1.

### Does this PR have a security-related impact?

No

### Does this PR change what data or activity we track or use?

No

### Does this PR have a legal-related impact?

No

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [x] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [x] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #9400 
Fixes #9401

PR #9483, opened 2021-10-20
Checklist: Account for Scale in Image Resolution Check

### LINKED ISSUE #9439: Checklist: Image Resolution Check Should Consider Scale
Currently the PPC check for image resolution isn't considering scale, it should. 

Follow up from #9401.

maxyinger:
## Context
Part of image resolution check in checklist.

## Summary
Accounts for element scale when performing image resolution check in the checklist.

## Relevant Technical Choices
NA

## To-do
NA

## User-facing changes
Now if your image is scaled really large and the uploaded resolution is less than it, it should populate the images scaling recommendation in the checklist.

## Testing Instructions
This is pretty hard to verify by interacting with the UI, so I'm gonna say it doesn't need QA verification. let me know if you think otherwise.

1. upload an image with a resolution that at a scale of 1 would be less that 2x element size, but at a scale of 4x, wouldn't
2. scale the image up all the way and see that this check passes.

## Checklist

---

Fixes #9439

### DIFF
--- a/packages/story-editor/src/components/checklist/checks/imageElementResolution.js
+++ b/packages/story-editor/src/components/checklist/checks/imageElementResolution.js
@@ -37,13 +37,17 @@ import { filterStoryElements, getVisibleThumbnails } from '../utils';
 import { useRegisterCheck } from '../countContext';
 import { useIsChecklistMounted } from '../popupMountedContext';
 
-// TODO: This does not seem to accommodate for scaling.
 export function imageElementResolution(element) {
-  const imageResolutionLow =
-    element.resource?.height < 2 * element.height ||
-    element.resource?.width < 2 * element.width;
+  if (!['image', 'gif'].includes(element.type)) {
+    return false;
+  }
 
-  return ['image', 'gif'].includes(element.type) && imageResolutionLow;
+  const scale = (element.scale || 100) / 100;
+  const imageResolutionLow =
+    element.resource?.height < 2 * (element.height * scale) ||
+    element.resource?.width < 2 * (element.width * scale);
+
+  return imageResolutionLow;
 }
 
 const ImageElementResolution = () => {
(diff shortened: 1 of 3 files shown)


samwhale (approved):
tested and it works well. Wondering if we should update the text shown as well. The issue card doesn't suggest decreasing the scaling:

maxyinger:
I can add something like:
> Consider scaling down the image

But that may be confusing cus you want to scale down the image element inside the editor, not the image you upload to the editor.

but wondering if that's kinda covered by:
> Consider similar pixel density for cropped images

idk. open to whatever here

samwhale:
> I can add something like:
> 
> > Consider scaling down the image
> 
> But that may be confusing cus you want to scale down the image element inside the editor, not the image you upload to the editor.
> 
> but wondering if that's kinda covered by:
> 
> > Consider similar pixel density for cropped images
> 
> idk. open to whatever here

To me it feels like it should point to setting the scale since that's a possible solution, so maybe the first one? I think the second one implies it but what if you don't understand pixel density 🤷

BrittanyIRL:
> > I can add something like:
> > > Consider scaling down the image
> > 
> > 
> > But that may be confusing cus you want to scale down the image element inside the editor, not the image you upload to the editor.
> > but wondering if that's kinda covered by:
> > > Consider similar pixel density for cropped images
> > 
> > 
> > idk. open to whatever here
> 
> To me it feels like it should point to setting the scale since that's a possible solution, so maybe the first one? I think the second one implies it but what if you don't understand pixel density 🤷

i have this in a PR right now, lemme know what you think: https://github.com/google/web-stories-wp/pull/9440

BrittanyIRL (approved):
Works for me!

maxyinger:
> i have this in a PR right now, lemme know what you think: #9440

Oh nice! that looks good to me

PR #9488, opened 2021-10-21
Fix: Media Accessibility Panel UI Simplifications

### LINKED ISSUE #9465: UI Simplifications: Media panels
## Feature Description
See [Figma for background media](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor-Sprint-Feature-Level-Designs_?node-id=10250%3A436589) and for [accessibility panel](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor-Sprint-Feature-Level-Designs_?node-id=10250%3A436662)

 ~Update the style for Removing Background media button~ That was an accidental difference, it can stay the same! - Amy
- [ ] Filter Panel: Add eyedropper next to the Color input (not seen in Figma but it was confirmed by UX that it should be like this everywhere. Makes sense to add it to the Color input directly if not done elsewhere.
- [ ] Accessibility panel: Add description above the textarea for both Video and Image (instead of the current placeholder).

delowardev:
See: #9465

## Context

Media Accessibility Panel UI Simplifications

## Summary

Added description above the textarea for both Video and Image (instead of the current placeholder). 
See: [figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=10250%3A436662)

## To-do

 - [x] Accessibility panel: Add description above the textarea for both Video and Image (instead of the current placeholder).
 - [x] Filter Panel: Add eyedropper next to the Color input (not seen in Figma but it was confirmed by UX that it should be like this everywhere. Makes sense to add it to the Color input directly if not done elsewhere.

## User-facing changes

Description moved to the label from placeholder.

Before | After
---|---
alt text|<br/> alt text
alt text|<br/>alt text
alt text|alt text
<br/>

## Testing Instructions

Important! From Pascal: FYI 3P PRs are not uploaded for QA due to GHA limitations. One would need to download the ZIP file from https://github.com/google/web-stories-wp/runs/4011337300 and install that version manually.

1. Add media (test with both image and video)
2. Open the Filter Panel, verify that in case of "Tint", eyedropper is displayed next to the Color Picker and it works as expected. There should be no eyedropper for Linear / Vignette.
3. Verify the style of the eyedropper looks as in [Figma ](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=10291%3A442662)
4. Open the Accessibility panel. Verify the description of the Accessibility field is displayed above the textarea for both Video and Image
5. Verify the design matches [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=10250%3A436662)

## Checklist

---

Fixes #9465


packages/story-editor/src/components/panels/design/imageAccessibility/imageAccessibility.js
```diff
@@ -68,19 +68,16 @@ function ImageAccessibilityPanel({ selectedElements, pushUpdate }) {
...
-          placeholder={
-            alt === MULTIPLE_VALUE
-              ? MULTIPLE_DISPLAY_VALUE
-              : __(
-                  'Add assistive text for visually impaired users',
-                  'web-stories'
-                )
-          }
+          placeholder={alt === MULTIPLE_VALUE ? MULTIPLE_DISPLAY_VALUE : ''}
+          label={__(
```
miina: As it is now, it's not really a label and is more of a description/hint. Ideally all text area components should work the same way and have consistent UX.
I'll confirm this with UX team, it might still change a bit.

delowardev: Added as a separate text similar to the "Story Description in the Document panel" as you mentioned

delowardev:
@miina could you please confirm how Eyedropper should behave when it's with gradient picker?

miina:
> could you please confirm how Eyedropper should behave when it's with gradient picker?

@delowardev We have the eyedropper feature currently implemented in the Color Picker -- we should just reuse the feature from there (Opening the Color Picker -> Clicking "Custom" button will open the custom view of the Color Picker where Eyedropper is). When using the Eyedropper, it chooses a new color from the Page and overwrites the existing one, so the previous color being gradient is irrelevant.

packages/story-editor/src/components/form/color/color.js
```diff
@@ -22,11 +22,21 @@ import styled from 'styled-components';
...
+  BUTTON_VARIANTS,
+  Icons,
+  themeHelpers,
+} from '@web-stories-wp/design-system';
 
 /**
  * Internal dependencies
  */
 import { MULTIPLE_VALUE } from '../../../constants';
+import useEyedropper from '../../colorPicker/eyedropper';
```
miina: Since the eyedropper will now mostly used outside of the `colorPicker`, it might make sense to separate it from the `colorPicker` component.

packages/story-editor/src/components/form/color/color.js
```diff
@@ -44,9 +54,28 @@ const Space = styled.div`
   background-color: ${({ theme }) => theme.colors.divider.primary};
 `;
 
-// 10px comes from divider
+// 10px comes from divider,
+// extra 2% width to fit color input when eyedropper is enabled
```
miina: The width of the panel is dynamic (changes depending on the screen width), however, the size of the Eyedropper button is always the same, so probably the extra room the eyedropper button should be considered in pixels, as it's static. Where did the 2% come from?

packages/story-editor/src/components/form/color/color.js
```diff
@@ -44,9 +54,28 @@ const Space = styled.div`
...
+`;
+
+const EyeDropperButton = styled(Button).attrs({
+  variant: BUTTON_VARIANTS.SQUARE,
+  type: BUTTON_TYPES.TERTIARY,
+  size: BUTTON_SIZES.SMALL,
+})`
+  margin-right: 4px;
+  ${({ theme }) =>
+    themeHelpers.focusableOutlineCSS(theme.colors.border.focus, '#1d1f20')};
```
miina: Would just `${focusStyle};` not work here?

packages/story-editor/src/components/panels/design/filter/filter.js
```diff
@@ -78,14 +78,13 @@ function FilterPanel({ selectedElements, pushUpdate }) {
         </Row>
       )}
       {overlayType !== OverlayType.NONE && (
-        <Row>
```
miina: Why remove the Row from here?

miina:
> could you please confirm how Eyedropper should behave when it's with gradient picker?

Ah, okay, now I understand, this question is for the Linear and Vignette specifically. We can disable the Eyedropper for these cases for now -- I'll check with UX and we can follow up about this if needed.

packages/story-editor/src/components/form/color/color.js
delowardev: Here the problem is with color input itself, it requires a minimum of 106px to display properly (hex value in the left image), so fixed with `flex-basis & flex-grow`.

packages/story-editor/src/components/form/color/color.js
```diff
@@ -44,9 +55,32 @@ const Space = styled.div`
   background-color: ${({ theme }) => theme.colors.divider.primary};
 `;
 
-// 10px comes from divider
+// color input requires 106px of minimum width to display hex value (try: #000000)
+// 10px comes from divider / 2
+const inputSize = 106;
 const InputWrapper = styled.div`
```
miina, line 66: From Figma, it looks like the Opacity input is always generally shorter than the Hex input, e.g. 100px vs 113px [here](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=10250%3A435845).

Previously, Opacity input did not have an icon and had `Opacity` written out, that's why it needed the same amount of room as the Hex input. Now that it only has an icon, we could always display it a little bit narrower compared to the Hex input, not only when the Eyedropper is present. E.g. using the 100px vs 113px in Figma would be 47% and 53% (- divider), we could always use these values now (as in Figma), this will also reduce the empty space of the opacity field a bit.
Thoughts?

delowardev: Right, it would simply resolve that issue.

miina:
@delowardev Let me know if the PR is ready for review (or re-request a review from me). I saw you marked it ready for review but also added some new commits, so not sure if it's currently just waiting for review.

packages/story-editor/src/components/form/color/color.js
```diff
@@ -44,9 +55,22 @@ const Space = styled.div`
...
+const OpacityWrapper = styled.div`
+  width: calc(47% - 10px);
+`;
+
+const EyeDropperButton = styled(Button).attrs({
+  variant: BUTTON_VARIANTS.SQUARE,
+  type: BUTTON_TYPES.TERTIARY,
+  size: BUTTON_SIZES.SMALL,
+})`
+  margin-right: 4px;
```
miina: The space in Figma is 8px, let's use the same as in Figma.

packages/story-editor/src/components/form/color/color.js
```diff
@@ -75,7 +100,40 @@ const Color = forwardRef(function Color(
...
+  const tooltip = __('Pick a color from canvas', 'web-stories');
+
+  // Enable eyedropper only for solid colors
+  // https://github.com/google/web-stories-wp/pull/9488#issuecomment-950679465
+  const shouldShowEyedropper =
+    hasEyedropper &&
+    value.type !== OverlayType.RADIAL &&
+    value.type !== OverlayType.LINEAR;
+
+  const withEyedropper = (children) => {
```
miina: Q: Why use the wrapper of `withEyedropper` here instead of just adding `<Tooltip>...</Tooltip>` directly before the `InputWrapper` below?

packages/story-editor/src/components/form/color/color.js
```diff
@@ -75,7 +100,40 @@ const Color = forwardRef(function Color(
...
+  const { initEyedropper } = useEyedropper({
+    onChange: (color) => onChange({ color }),
+  });
+  const tooltip = __('Pick a color from canvas', 'web-stories');
+
+  // Enable eyedropper only for solid colors
+  // https://github.com/google/web-stories-wp/pull/9488#issuecomment-950679465
+  const shouldShowEyedropper =
+    hasEyedropper &&
+    value.type !== OverlayType.RADIAL &&
```
miina: Detecting this part should go directly to the Filter panel and `hasEyedropper` not be added if it's radial/linear. Filter panel is the only place where we should avoid displaying the Eyedropper for non-solid colors.

In other cases, the eyedropper value would just overwrite the previous color.

packages/story-editor/src/components/panels/design/imageAccessibility/imageAccessibility.js
```diff
@@ -23,12 +23,19 @@ import { __ } from '@web-stories-wp/i18n';
...
+import { Text, THEME_CONSTANTS } from '@web-stories-wp/design-system';
+import styled from 'styled-components';
 import { Row, TextArea } from '../../../form';
 import { getCommonValue, useCommonObjectValue } from '../../shared';
 import { SimplePanel } from '../../panel';
 import { MULTIPLE_DISPLAY_VALUE, MULTIPLE_VALUE } from '../../../../constants';
 import { useHighlights, states, styles } from '../../../../app/highlights';
 
+const StyledText = styled(Text)`
+  color: ${({ theme }) => theme.colors.fg.tertiary};
```
miina: Should be `theme.colors.fg.secondary` per Figma:
Screenshot 2021-10-26 at 12 33 41

packages/story-editor/src/components/panels/design/imageAccessibility/imageAccessibility.js
```diff
@@ -23,12 +23,19 @@ import { __ } from '@web-stories-wp/i18n';
...
+  margin-bottom: 12px;
```
miina: I'm seeing 10px in Figma, which place did you see 12px? Mb I'm checking other place.

packages/story-editor/src/components/panels/design/videoAccessibility/videoAccessibility.js
```diff
@@ -55,6 +56,11 @@ const InputsWrapper = styled.div`
   margin-left: 16px;
 `;
 
+const StyledText = styled(Text)`
+  color: ${({ theme }) => theme.colors.fg.tertiary};
```
miina: Same here about the color, it's marked `secondary` in Figma.

miina (changes requested):

packages/story-editor/src/components/form/color/color.js
delowardev: @miina in 4px is being used in the "Text panel", couldn't find a focused/hover style for the EyeDropper button in Figma.

packages/story-editor/src/components/form/color/color.js
delowardev: @miina  Just wanted to avoid CSS calculation for the EyeDropper & its margin, thought it would be less confusing to wrap with the container. Do you prefer adding before the `InputWrapper` instead? (the CSS `calc` would be something like `(53% - (eyeDropper + margin / 2) - divider)`

packages/story-editor/src/components/form/color/color.js
miina: @delowardev Let's use 8px as in Figma for the eyedropper:
Screenshot 2021-10-26 at 13 25 32

The focus/hover style comes automatically from the button type (tertiary), so no need to worry about that.

packages/story-editor/src/components/form/color/color.js
miina: If it makes sense, you can still wrap the eyedropper button into a `EyedropperWrapper` or something similar, just not sure why it needs to be using `withEyedropper` instead of just adding it before the `InputWrapper`

packages/story-editor/src/components/panels/design/imageAccessibility/imageAccessibility.js
delowardev: updated to 10px. I think there are some design mismatches in the Figma, for example: [here](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=10149%3A434319) the margin is 6px

miina: That Figma area is outdated, only the area that's linked in the issue is up to date -- the area that has the latest date (10/12-15)

packages/story-editor/src/components/form/color/color.js
```diff
@@ -44,9 +54,30 @@ const Space = styled.div`
   background-color: ${({ theme }) => theme.colors.divider.primary};
 `;
 
-// 10px comes from divider
+// ( 32px eyedropper button + 8px margin ) / 2
+const eyedropperSpacing = '20px';
+// 10px comes from divider / 2
 const InputWrapper = styled.div`
-  width: calc(50% - 10px);
+  width: ${({ hasEyedropper }) =>
```
miina: Previously you mentioned having a wrapper around the `EyeDropperButton` to make this calculation more readable, any reason why you opted for not using it?

Probably the easiest way would be to have something like `ColorInputWrapper` around the color inputs and then `EyedropperWrapper` around the eyedropper button (if it needs one that all, maybe just having a wrapper around both color inputs would work as well). Then we could still just use `'calc(53% - 10px)'` etc., no?

Then it could be something like 
```
<Container>
  <Eyedropper ... />
  <ColorInputsWrapper>
    <InputWrapper ... />
    <OpacityWrapper ... />
  </ColorInputsWrapper>
</Container>
```

miina: (That's the last nit to look into, otherwise looks good)

delowardev: Thanks, that's a good idea actually. Code has been updated

packages/story-editor/src/components/form/color/color.js
```diff
@@ -75,28 +99,51 @@ const Color = forwardRef(function Color(
   const displayOpacity =
     value !== MULTIPLE_VALUE && Boolean(getPreviewText(value));
 
+  const { initEyedropper } = useEyedropper({
+    onChange: (color) => onChange({ color }),
+  });
+  const tooltip = __('Pick a color from canvas', 'web-stories');
+
+  // eslint-disable-next-line react/prop-types
+  const ColorInputsWrapper = ({ children }) =>
```
miina: Let's create a new styled component so that we always use it instead of displaying it conditionally.
The Container is currently a `section` tag, the wrapper around the input can be just a `div`.
```
  const ColorInputsWrapper = styled.div`...`;
```

packages/story-editor/src/components/form/color/color.js
```diff
@@ -18,35 +18,66 @@
...
 import { MULTIPLE_VALUE } from '../../../constants';
+import useEyedropper from '../../eyedropper';
+import Tooltip from '../../tooltip';
+import { focusStyle } from '../../panels/shared';
 import applyOpacityChange from './applyOpacityChange';
 import OpacityInput from './opacityInput';
 import ColorInput from './colorInput';
 
-const Container = styled.section`
+const ContainerStyles = css`
```
miina: Nit: for consistency with code elsewhere, let's use
`const containerCss = ...` (first char is lowercase + using Css as the suffix)

packages/story-editor/src/components/panels/design/imageAccessibility/imageAccessibility.js
```diff
@@ -60,6 +67,9 @@ function ImageAccessibilityPanel({ selectedElements, pushUpdate }) {
       title={__('Accessibility', 'web-stories')}
       isPersistable={!highlight}
     >
+      <StyledText size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.X_SMALL}>
+        {__('Add assistive text for visually impaired users', 'web-stories')}
```
miina: One more nit: in Figma, the new text is `Add a brief description of the image` ("video" in case of video)

packages/story-editor/src/components/panels/design/videoAccessibility/videoAccessibility.js
```diff
@@ -159,6 +165,9 @@ function VideoAccessibilityPanel({ selectedElements, pushUpdate }) {
           canUpload={hasUploadMediaAction}
         />
         <InputsWrapper>
+          <StyledText size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.X_SMALL}>
```
miina: Just noticed that it's currently using `X_SMALL`. It's `SMALL` per Figma, using 14px as the font-size. 
X_SMALL uses 12px as the font size.

Let's change this to `SMALL` size instead.
Screenshot 2021-10-26 at 17 48 56

miina (approved):

miina:
@delowardev Thank you for working on the PR and for the quick responses!

swissspidy:
@miina @csossi FYI 3P PRs are not uploaded for QA due to GHA limitations. One would need to download the ZIP file from https://github.com/google/web-stories-wp/runs/4011337300 and install that version manually.

PR #9490, opened 2021-10-21
Block: Use actual `<img>` for `<amp-story-player>` poster

### LINKED ISSUE #9391: Block: Use actual `<img>` for `<amp-story-player>` poster
## Feature Description

Instead of setting the poster as background image of the `<a>` within `<amp-story-player>`, we should use a proper `` now that this is supported.

```html
<amp-story-player style="width: 360px; height: 600px;" amp-cache="cdn.ampproject.org">
    <a
      href="https://preview.amp.dev/documentation/examples/introduction/stories_in_amp/"
    >
      <img src="https://amp.dev/static/samples/img/story_dog2_portrait.jpg" width="360" height="600" loading="lazy" data-amp-story-player-poster-img>
    </a>
  </amp-story-player>
```

spacedmonkey:
## Summary

Instead of setting the poster as background image of the `<a>` within `<amp-story-player>`, we should use a proper `` now that this is supported.

## Testing Instructions

1.

## Checklist

---

Fixes #9391

### DIFF
--- a/includes/Renderer/Story/Embed.php
+++ b/includes/Renderer/Story/Embed.php
@@ -121,9 +121,11 @@ class Embed {
 									loading="lazy"
 									data-amp-story-player-poster-img
 								/>
-							<?php } else {
+								<?php 
+							} else {
 								echo esc_html( $title );
-							} ?>
+							} 
+							?>
 						</a>
 					</amp-story-player>
 				</div>
@@ -150,9 +152,11 @@ class Embed {
 								loading="lazy"
 								data-amp-story-player-poster-img
 							/>
-						<?php } else {
+							<?php 
+						} else {
 							echo esc_html( $title );
-						} ?>
+						} 
+						?>
 					</a>
 				</amp-story-player>
 			</div>
--- a/includes/Renderer/Story/Embed.php
+++ b/includes/Renderer/Story/Embed.php
@@ -90,12 +90,7 @@ class Embed {
 		$class  = $args['class'];
 		$url    = $this->story->get_url();
 		$title  = $this->story->get_title();
-		$poster = ! empty( $this->story->get_poster_portrait() ) ? esc_url_raw( $this->story->get_poster_portrait() ) : '';
-
-		$amp_player_inner = esc_html( $title );
-		if ( $poster ) {
-			$amp_player_inner = sprintf( '<img src="%s" width="%d" height="%d" alt="%s" loading="lazy" data-amp-story-player-poster-img />', $poster, (int) $args['width'], (int) $args['height'], esc_attr( $title ) );
-		}
+		$poster = ! empty( $this->story->get_poster_portrait() ) ? $this->story->get_poster_portrait() : '';
 
 		$wrapper_style = sprintf(
 			'--aspect-ratio: %F; --width: %dpx; --height: %dpx',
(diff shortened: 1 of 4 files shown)


includes/Renderer/Story/Embed.php
```diff
@@ -92,7 +92,11 @@ public function render( array $args = [] ): string {
 		$title  = $this->story->get_title();
 		$poster = ! empty( $this->story->get_poster_portrait() ) ? esc_url_raw( $this->story->get_poster_portrait() ) : '';
 
-		$poster_style  = ! empty( $poster ) ? sprintf( '--story-player-poster: url(%s)', $poster ) : '';
+		$amp_player_inner = esc_html( $title );
+		if ( $poster ) {
+			$amp_player_inner = sprintf( '<img src="%s" width="%d" height="%d" alt="%s" loading="lazy" data-amp-story-player-poster-img />', $poster, (int) $args['width'], (int) $args['height'], esc_attr( $title ) );
```
swissspidy: I'd use numbered placeholders here and put this on multiple lines.

But could also just print the `` tag directly in the code below to improve readability. Then we also don't need the `phpcs:ignore` comments

includes/Renderer/Story/Embed.php
```diff
@@ -92,7 +92,11 @@ public function render( array $args = [] ): string {
 		$title  = $this->story->get_title();
 		$poster = ! empty( $this->story->get_poster_portrait() ) ? esc_url_raw( $this->story->get_poster_portrait() ) : '';
```
swissspidy: Now we can use `esc_url` instead of `esc_url_raw`.

But I would also suggest to late escape further down before actual output.

```suggestion
		$poster = ! empty( $this->story->get_poster_portrait() ) ? $this->story->get_poster_portrait() : '';
```

includes/Renderer/Story/Embed.php
```diff
@@ -112,10 +116,8 @@ public function render( array $args = [] ): string {
 						width="<?php echo esc_attr( $args['width'] ); ?>"
 						height="<?php echo esc_attr( $args['height'] ); ?>"
 						layout="intrinsic">
-						<a
-							href="<?php echo esc_url( $url ); ?>"
-							style="<?php echo esc_attr( $poster_style ); ?>">
-							<?php echo esc_html( $title ); ?>
+						<a href="<?php echo esc_url( $url ); ?>">
+							<?php echo $amp_player_inner; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
```
swissspidy:
```suggestion
							<?php if ( $poster ) : ?>
								<img src="<?php echo esc_url( $poster ); ?>" width="<?php echo esc_attr( $args['width'] ); ?>" height="<?php echo esc_attr( $args['height'] ); ?>" alt="<?php echo esc_attr( $title ); ?>" loading="lazy" data-amp-story-player-poster-img />
							<?php else : ?>
								<?php echo esc_html( $title ); ?>
							<?php endif: ?>
```

swissspidy (approved):
I'd personally prefer a different style of writing the `img` tag, but otherwise looks good 👍

includes/Renderer/Story/Embed.php
spacedmonkey: Should we use amp-img tag here?

swissspidy: Since there's a `$this->is_amp()` condition below already, yeah why not 👍 

Another reason why it's simpler to print the HTML directly.

spacedmonkey: In my testing, it seems the amp-img is stripped out. Also all the example of posters as image tags are img not amp-img. Let's leave it as it is.

includes/Renderer/Story/Embed.php
```diff
@@ -112,10 +111,21 @@ public function render( array $args = [] ): string {
 						width="<?php echo esc_attr( $args['width'] ); ?>"
 						height="<?php echo esc_attr( $args['height'] ); ?>"
 						layout="intrinsic">
-						<a
-							href="<?php echo esc_url( $url ); ?>"
-							style="<?php echo esc_attr( $poster_style ); ?>">
-							<?php echo esc_html( $title ); ?>
+						<a href="<?php echo esc_url( $url ); ?>">
+							<?php if ( $poster ) { ?>
+								<img
```
swissspidy, line 116: This is where we can use `amp-img`

Probably need to add `layout="responsive"` too

spacedmonkey: I know but this - https://github.com/google/web-stories-wp/pull/9490#discussion_r733747356

includes/Renderer/Story/Embed.php
swissspidy: The examples are for non-AMP pages, thus not using amp-img.

In your testing, the element probably got stripped because `layout="responsive"` wasn't added.

swissspidy: Welp, looks like it's indeed `` only. My bad!

https://github.com/ampproject/amphtml/pull/30753
https://github.com/ampproject/amphtml/pull/32371

Issue #8953, opened 2021-09-06
Decoupling: Platform-agnostic media upload


swissspidy:
## Background

The story editor package uses `useMediaPicker()` in multiple places to handle media uploads.

That hook allows using the WordPress media modal for upload.

To decouple the editor from WordPress, this logic needs to be made more platform-agnostic, with the WordPress media modal handling moved to `wp-story-editor`, allowing other implementations on other platforms.

## Proof-of-concept

@sayedtaqui has worked on a POC for decoupling the media upload, making `MediaUpload` a React component that needs to be passed via the `ConfigProvider`:

https://github.com/rtCamp/web-stories-wp/pull/731

## Context

Gutenberg uses the `@wordpress/hooks` package to allow filtering the `MediaUpload` component. By default it renders nothing, but in the WordPress block editor it's filtered to render a component using the WordPress media modal.

https://github.com/WordPress/gutenberg/blob/3e14c5e812d8077f28e080522f20763bbfb1dccc/packages/block-editor/src/components/media-upload/README.md

https://github.com/WordPress/gutenberg/blob/3da717b8d0ac7d7821fc6d0475695ccf3ae2829f/packages/block-editor/src/components/media-upload/index.js

https://github.com/WordPress/gutenberg/blob/3da717b8d0ac7d7821fc6d0475695ccf3ae2829f/packages/edit-post/src/hooks/components/index.js

We can't use `@wordpress/hooks` directly due to the GPL license, which is incompatible with our Apache license.

## Implementation Brief

We need to find a good way for making media upload platform-agnostic.

The POC above works, but the config React context is better suited for passing data/settings and not React components. So it doesn't feel right to pass `MediaUpload` that way.

There's also another POC for using hooks/filters: https://github.com/rtCamp/web-stories-wp/pull/765

But we can't really use that as-is due to licensing restrictions (uses parts of GPL-licensed code).

An alternative could perhaps be to use a [Slot & Fill technique](https://github.com/camwest/react-slot-fill).

But maybe there are other, better ways.

swissspidy:
@choumx @barklund Would love to hear your thoughts on this

swissspidy:
Current plan is to proceed with the POC and refine later based on feedback. The code needs to be moved anyway.

samwhale:
Hi hi hi! So just opened #9534 and have been learning about this work that's being done. 

[...]

PR #9534, opened 2021-10-26
Quick Actions: Add Replace Media Functionality

### LINKED ISSUE #9196: "Replace media/background" quick action should actually replace
## Feature Description

Expected: The "Replace media/background" quick actions replaces the currently selected media while preserving position and size (and respecting the new media's aspect ratio)

Actual: The "Replace media/background" quick actions just focuses the media panel

## Alternatives Considered

N/A

## Additional Context

This would make templates more usable by making it much easier to swap out the stock images.

Originally reported by `mludwig@` who was assisting an external user.

samwhale:
## Context

People want to replace media, but _for real_

## Summary

Adds `Replace Media` functionality for foreground and background images.

## Relevant Technical Choices

The [`MediaUpload` component](https://github.com/google/web-stories-wp/blob/main/packages/wp-story-editor/src/components/mediaUpload/index.js#L26-L30) requires a render prop.

However, the `useQuickActions` hook does not render the quick action button. It's easier to pass in a click event handler that is later passed to the context menu button.

To do this, the `useMediaPicker` hook was exposed to the app. That way we don't need to use the render prop.

## To-do

- [x] Add replace functionality to background media
- [x] Add karma tests

## User-facing changes

|Replace foreground media|Replace background media|
|--|--|
|replace-foreground|replace-background-media|

## Testing Instructions

**Make sure you do these for foreground media + background media (images, gifs, and videos)
1. Add a piece of media
2. Add styles to the media in the design panel
3. Click the `Replace` action in the quick action menu
4. Add a media and click `insert media`
5. New media should replace existing media but the styles should remain the same
6. Try the same thing by inserting media of different types (image, video, gif)

_Users without the `hasMediaUpload` capability should not be able to see this quick action_ (contributors)

## Checklist

---

Fixes #9196


packages/wp-story-editor/src/index.js
```diff
@@ -66,6 +67,7 @@ const initialize = (id, config, flags) => {
     ...config,
     apiCallbacks: getApiCallbacks(config),
     MediaUpload,
+    useMediaPicker,
```
samwhale: 🤔  Wondering what the motivation here is to load them in through the config rather than importing the module instead.

Would make integration testing a bit easier too 😄

samwhale: >Would make integration testing a bit easier too 😄

Hmm well, we'd probably still have to mock it so that the media isn't actually uploaded 🤔

mohdsayed: In the context of decoupling, when a third party integrates the core editor ( story-editor ) with their CMS/platform, they will pass their own `MediaUpload` component which is supposed to provide certain props including `render`. We are unopinionated about how they created that `MediaUpload` component as long as those props are being provided.

Therefore `useMediaPicker` is an internal implementation of `MediaUpload` within `wp-story-editor` ( WordPress integration in our case ) and cannot be exposed ( or used in ) to `story-editor`. We need to think of some alternative.

spacedmonkey:
If you insert an image block into a post / page in WordPress, then press replace, the WP Picker opens with the current image selected. 
Screenshot 2021-11-01 at 16 00 04

We should do the same here. 

I would also change the title of the WP Picker, something like "Replace media"

packages/story-editor/src/app/mediaUpload/useMediaUpload.js
```diff
@@ -0,0 +1,38 @@
...
+import Context from './context';
+
+function useMediaUpload(selector) {
+  const context = useContextSelector(Context, selector ?? identity);
+
+  if (!context) {
+    throw new Error(
+      'Must use `useMediaUpload()` within <mediaUpload.Provider />'
+    );
+  }
```
maxyinger: I guess within this context, this is fine. But watch out for this pattern with `useContextSelector` because it's the context after it's been selected. So if say you have:
```
const someBool = useSomeContext((context) => context.state.someBool);
```
and someBool happens to be `false`, we wouldn't want an error thrown. 

Anyway, just giving a heads up cus I've ran into this before

packages/story-editor/src/components/layout/index.js
```diff
@@ -81,24 +87,170 @@ function Layout({ header, children }) {
...
     })
   );
+  const {
+    allowedTranscodableMimeTypes,
+    allowedFileTypes,
+    allowedMimeTypes: {
+      image: allowedImageMimeTypes,
+      video: allowedVideoMimeTypes,
+    },
+    // capabilities: { hasUploadMediaAction },
```
maxyinger: remove?

packages/story-editor/src/app/mediaUpload/provider.js
```diff
@@ -0,0 +1,116 @@
...
+ */
+import { useConfig } from '..';
+import Context from './context';
+
+function OpenMediaUploadHandler({ onOpenMediaUpload, ...props }) {
+  /**
+   * `open` needs to be stored in a ref so that it does not cause endless
+   * re-renders for any children.
+   */
+  const openRef = useRef(open);
```
maxyinger: should this be:
```
const openRef = useRef(onOpenMediaUpload);
```

packages/story-editor/src/app/mediaUpload/provider.js
```diff
@@ -0,0 +1,116 @@
...
+
+  return (
+    <MediaUpload
+      {...props}
+      render={(open) => (
+        /* the only way to access `open` is through the render prop of `MediaUpload` */
+        <OpenMediaUploadHandler onOpenMediaUpload={open}>
+          {children}
+        </OpenMediaUploadHandler>
+      )}
```
maxyinger: maybe memoize since this is held at a high level:
```js
const renderMediaProvider = useCallback((open) => (
        /* the only way to access `open` is through the render prop of `MediaUpload` */
        <OpenMediaUploadHandler onOpenMediaUpload={open}>
          {children}
        </OpenMediaUploadHandler>
), []);
// ...
render={renderMediaProvider}
```

packages/story-editor/src/app/mediaUpload/provider.js
maxyinger: I'm not following how this as a ref like this changes the behavior at all vs just passing the `onOpenMediaUpload` prop right through the provider. Mind breaking it down for me?

packages/story-editor/src/components/layout/index.js
```diff
@@ -81,24 +87,170 @@ function Layout({ header, children }) {
...
-            </Editor>
-          </HighlightsProvider>
+          <MediaUploadProvider
+            title={__('Select media', 'web-stories')}
+            buttonInsertText={__('Insert media', 'web-stories')}
+            onSelect={handleMediaSelect}
+            onClose={resetWithFetch}
+            type={allowedMimeTypes}
+            onSelectErrorMessage={onSelectErrorMessage}
+          >
```
maxyinger: I get worried about wrapping the whole editor in this. Would it be worth just wrapping sections of the editor where we need it?

maybe just wrapping up all the args and titles into something like this that we can use around the app?
```
<BaseMediaUploadProvider>
    // ...children
</BaseMediaUploadProvider>
```

Or does this sound silly? idk

maxyinger (commented):
Got a couple Qs, but looking great!

packages/story-editor/src/app/mediaUpload/provider.js
samwhale: oops yeah, good catch

We really should pass the ref through, not the value of the ref. Monday brain 😮‍💨

samwhale: I think this would work
```js
function OpenMediaUploadHandler({ onOpenMediaUpload, ...props }) {
  /**
   * `open` needs to be stored in a ref so that it does not cause endless
   * re-renders for any children.
   */
  const openRef = useRef(onOpenMediaUpload);
  openRef.current = onOpenMediaUpload;

  const handleOpenMediaUpload = useCallback((ev) => openRef.current(ev), []);

  return (
    <Context.Provider
      {...props}
      value={{ onOpenMediaUpload: handleOpenMediaUpload }}
    />
  );
}
OpenMediaUploadHandler.propTypes = {
  onOpenMediaUpload: PropTypes.func.isRequired,
};

```

packages/story-editor/src/components/layout/index.js
samwhale: It just needs to wrap the quick action menu, so that should work

samwhale: The tricky thing is that it needs to live above the highlights provider because the quick actions live in the highlights provider

I agree though, it'd be nice to not wrap the whole app in this.

To fix this:
1. Refactor the quick actions into their own provider
2. Wrap just the quick action menu with the `QuickActionProvider` and the `MediaUploadProvider` 👍 

That would make this pr pretty massive though. I'm happy to follow up this PR with that refactor

packages/story-editor/src/app/mediaUpload/provider.js
samwhale: But the idea is that if the `open` function that's taken in from the render prop changes, then it's added to the ref. Hopefully this wouldn't cause an extra re-render every time something above this re-renders since the `onOpenMediaUpload` function reference never changes

But lmk if I'm thinking about this incorrectly

packages/story-editor/src/components/layout/index.js
maxyinger: awesome. yea agreed. we can make it a code quality follow up

packages/story-editor/src/components/layout/index.js
```diff
@@ -81,24 +88,170 @@ function Layout({ header, children }) {
...
-            </Editor>
-          </HighlightsProvider>
+          <MediaUploadProvider
+            title={__('Replace media', 'web-stories')}
+            buttonInsertText={__('Insert media', 'web-stories')}
+            onSelect={handleMediaSelect}
+            onClose={resetWithFetch}
+            type={allowedMimeTypes}
+            onSelectErrorMessage={onSelectErrorMessage}
+          >
```
swissspidy: That's quite odd, as these props would mean this generically named `MediaUploadProvider` is only used for the "Replace media" use case.

Shouldn't these be passed to `onOpenMediaUpload` if anything?

maxyinger: I think that would be tough, cus you'd need to make sure the `MediaUpload` rendered with those args before the `open` it supplies (and you'd be calling) is valid. I think this is just one of the caveats that comes with making this a component that exposes `open` through a render prop.

Kinda mentioned it here:
https://github.com/google/web-stories-wp/issues/8953#issuecomment-955018168

samwhale: In this case, it is only for the 'replace media' use case. For any other use cases we'd have to create another instance of the provider.

Just like Max said - this is part of having the `MediaUpload` component passed into the config, and not exposing an `openModal` function without first rendering the component. This provider exposes the 'open' function to the rest of the app here: https://github.com/google/web-stories-wp/pull/9534/files#diff-fb9f988abfb4a21e672cc1f4d632c65fe4815ea00db7a5936443cd0b04e83e2eR91-R98.

The problem here is that the 'open' function can't be created until we the `MediaUpload` component is initialized. This was why initially I tried to grab the `useMediaPicker` hook. However, I now know that that isn't possible.

samwhale:
> If you insert an image block into a post / page in WordPress, then press replace, the WP Picker opens with the current image selected.
>
> We should do the same here.

@spacedmonkey that would be great! I'd say that is out of scope of this ticket, so I think we should add a ticket to do this in the future. Thoughts?

packages/story-editor/src/components/layout/index.js
swissspidy: Then at the very least this provider would need to be renamed to have a less generic name

swissspidy:
Not super fond of how we have to solve this, so I'd really like feedback from @choumx and @sayedtaqui here

dreamofabear:
Sorry for the delay, will take a look at this tomorrow.

packages/story-editor/src/app/mediaUpload/provider.js
mohdsayed: I see the problem and why we are doing this. However, where `render` is supposed to render the button, it is weird to render the entire app inside it to make `open` available in the required component. 

Since `MediaUpload` can render the replace background icon, I think it's a matter of how the code is structured in our core story editor. So maybe if we think in a different direction of refactoring the existing code to use the `MediaUpload` component for rendering that icon as it should be, instead of hacking the `open` callback, it would make more sense.

mohdsayed:
Perhaps we need to refactor our existing code to make it more flexible to accept `MediaUpload` as it is, instead of dragging the `open` callback to the point where it is needed. I have not tried but maybe we can directly pass the `MediaUpload` via `Icon` here and render the picture swap icon inside its `render`.

https://github.com/google/web-stories-wp/blob/80f1ccfcb53876aa2706143cd24a2fd2bffac906/packages/story-editor/src/app/highlights/quickActions/useQuickActions.js#L375-L380

and then reactor [MenuItem](https://github.com/google/web-stories-wp/blob/main/packages/design-system/src/components/contextMenu/menuItem.js#L53-L59) to handle the case when `onClick` is not provided.

BrittanyIRL:
> Perhaps we need to refactor our existing code to make it more flexible to accept `MediaUpload` as it is, instead of dragging the `open` callback to the point where it is needed. ...

🤔  This is an interesting thought, but to me it feels a bit dodgy to mold the design system (which is supposed to be agnostic) to this API?

mohdsayed:
> 🤔 This is an interesting thought, but to me it feels a bit dodgy to mold the design system (which is supposed to be agnostic) to this API?

Not sure if I understood the dodgy part for making a small changes for `onClick`, but this is what I was suggesting. 
If we do not want to make any changes to  [MenuItem](https://github.com/google/web-stories-wp/blob/main/packages/design-system/src/components/contextMenu/menuItem.js#L53-L59) at all, we may just pass `noop` for `onClick`

https://github.com/google/web-stories-wp/compare/try/quickaction-media-upload?expand=1

BrittanyIRL:
> > 🤔 This is an interesting thought, but to me it feels a bit dodgy to mold the design system (which is supposed to be agnostic) to this API?
> 
> Not sure if I understood the dodgy part for making a small changes for `onClick`, but this is what I was suggesting. If we do not want to make any changes to [MenuItem](https://github.com/google/web-stories-wp/blob/main/packages/design-system/src/components/contextMenu/menuItem.js#L53-L59) at all, we may just pass `noop` for `onClick`
> 
> https://github.com/google/web-stories-wp/compare/try/quickaction-media-upload?expand=1

Ah, I see. I had thought you meant change the design system's API to fit this structure. I see now what you mean by putting the click event on the icon. Still feels like a break in the structure but this is a tough solve and it's going to be a little weird no matter what. Thank you for the demo of what you meant.

packages/story-editor/src/app/mediaUpload/provider.js
dreamofabear: > refactoring the existing code to use the MediaUpload component for rendering that icon

Hmm, I haven't been in the trenches here but I think I broadly agree with this suggestion.

The current approach is pretty clever to workaround `MediaUpload`'s use of a [render prop](https://reactjs.org/docs/render-props.html). 

However, instead of adding `OpenMediaUploadHandler` and `MediaUploadProvider`, could we instead extend the `useQuickActions` API? 

Add an alternative to the `onClick` property -- maybe a `wrapper` property that decorates an existing component or definition? Then in `useQuickActions.js`, we could have something like:

```js
{
  Icon: PictureSwap,
  label: ACTIONS.REPLACE_BACKGROUND_MEDIA.text,
  onClickWrapper: (ToBeWrapped) => {
    return <MediaUpload
      render={(open) => <ToBeWrapped onClick={open}>},
      // having these properties below keeps media-related code local to its usage
      title={__('Replace media', 'web-stories')},
      onSelect={handleMediaSelect},
      ...
    />;
  },
  ...actionMenuProps,
},
```

And then in `menuItem.js`, something like:

```js
const ButtonWithoutClick = () => 
  <Button
    ref={itemRef}
    aria-label={itemLabel}
    disabled={disabled}
    onFocus={onFocus}
    {...menuItemProps}
  >
    {textContent}
  </Button>;

if (onClick) { // existing behavior
  return <ButtonWithoutClick onClick={handleClick}>
} else if (onClickWrapper) { // new behavior
  return onClickWrapper(ButtonWithoutClick);
}
```

It'd be nice to keep new code more local and avoid adding a root-level provider (especially one for each use of `MediaUpload`).

dreamofabear:
> If we do not want to make any changes to MenuItem at all, we may just pass noop for onClick
> 
> https://github.com/google/web-stories-wp/compare/try/quickaction-media-upload?expand=1

Ah just saw this but I made a similar suggestion here: https://github.com/google/web-stories-wp/pull/9534#discussion_r743227045

Edit: one issue with Sayed's proposal is that it moves some presentation logic (how to render a quick action menu item?) from `menuItem.js` into `useQuickActions.js`. A wrapper approach avoids this and allows us to keep `useQuickActions` to only contain data/configuration of context menu UI.

swissspidy:
> > 🤔 This is an interesting thought, but to me it feels a bit dodgy to mold the design system (which is supposed to be agnostic) to this API?
> 
> Not sure if I understood the dodgy part for making a small changes for `onClick`, but this is what I was suggesting. If we do not want to make any changes to [MenuItem](https://github.com/google/web-stories-wp/blob/main/packages/design-system/src/components/contextMenu/menuItem.js?rgh-link-date=2021-11-04T21%3A03%3A26Z#L53-L59) at all, we may just pass `noop` for `onClick`
> 
> [`try`?expand=1 (compare)](https://github.com/google/web-stories-wp/compare/try/quickaction-media-upload?expand=1)

That would be waay simpler than what's currently in this PR for sure 👍

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
```diff
@@ -382,15 +524,15 @@ const useQuickActions = () => {
       actions.push({
         Icon: PictureSwap,
         label: ACTIONS.REPLACE_MEDIA.text,
-        onClick: (ev) => {
+        onClick: () => {
```
maxyinger, line 554: would `onClick` get overwritten here anyway?

maxyinger (approved):
This works for me. 

I'm not in love with any solution that's been proposed here so far, and I don't think there's a clean one with the way the `MediaUpload` is set up. Looks like there all versions of one of the two approaches I mentioned here originally:
https://github.com/google/web-stories-wp/issues/8953#issuecomment-955018168

I don't think grinding the wheels too much on this one outlier case for both apis will provide much benefit.

Needless to say, if it works and is relatively isolated, I'm cool with it

packages/design-system/src/components/contextMenu/menu.js
```diff
@@ -172,6 +172,19 @@ MenuList.propTypes = {
...
 
+/**
+ * Default wrapper to wrap the context menu's menu item. Necessary to allow
+ * other apps the ability to inject their own media pickers.
+ *
+ * @param {Object} props Component props
+ * @param {Function} props.children Children as a function.
+ * @return {Node} The react component
+ */
+const DefaultWrapper = ({ children, ...props }) => children(props);
```
dreamofabear: Nit: `IdentityItemWrapper` or `NoopItemWrapper`?

- "Identity/noop" because it doesn't do anything
- "Item" to disambiguate with existing `MenuWrapper`

packages/design-system/src/components/contextMenu/menu.js
```diff
@@ -172,6 +172,19 @@ MenuList.propTypes = {
   isIconMenu: PropTypes.bool,
 };
 
+/**
+ * Default wrapper to wrap the context menu's menu item. Necessary to allow
+ * other apps the ability to inject their own media pickers.
+ *
+ * @param {Object} props Component props
+ * @param {Function} props.children Children as a function.
```
dreamofabear:
```suggestion
 * @param {Function} props.render Decorating wrapper function for menu item components.
```

Rename `children` to `render`? To avoid confusion with the normal type of React children, and "render prop" is an [established convention](https://reactjs.org/docs/render-props.html).

Also stumbled across this: https://americanexpress.io/faccs-are-an-antipattern/

packages/design-system/src/components/contextMenu/menu.js
```diff
@@ -303,24 +316,36 @@ const Menu = ({
...
-              index={index}
-              onFocus={(ev) => handleFocusItem(ev, index, onFocus)}
-              onDismiss={onDismiss}
-              {...itemProps}
-            />
-          </li>
-        ))}
+        {items.map(
+          (
+            { Wrapper = DefaultWrapper, separator, onFocus, ...itemProps },
```
dreamofabear:
```suggestion
            { ItemWrapper = DefaultItemWrapper, separator, onFocus, ...itemProps },
```

packages/story-editor/src/app/highlights/quickActions/mediaUploadWrapper.js
```diff
@@ -0,0 +1,68 @@
...
+  );
+};
+MediaUploadWrapper.propTypes = {
+  children: PropTypes.func.isRequired,
+  mediaUploadProps: PropTypes.shape(mediaUploadPropsType),
+};
+
+const WithMediaUploadWrapper = ({ children, ...props }) => (
+  <MediaUploadWrapper mediaUploadProps={props}>{children}</MediaUploadWrapper>
+);
```
dreamofabear: Hmm, what does this extra wrapper buy us? Can it be removed?

packages/story-editor/src/app/highlights/quickActions/mediaUploadWrapper.js
```diff
@@ -0,0 +1,68 @@
...
+
+const MediaUploadWrapper = ({ children, mediaUploadProps, ...props }) => {
+  const { MediaUpload } = useConfig();
+
+  return (
+    <MediaUpload
+      {...mediaUploadProps}
+      // overwrite `onClick` because the only way to access the open
+      // function is to dive into the MediaUpload component in the render prop.
+      render={(open) => children({ ...props, onClick: open })}
```
dreamofabear: Ditto here with `children` as a function. Rename to `render`?

packages/story-editor/src/app/highlights/quickActions/mediaUploadWrapper.js
```diff
@@ -0,0 +1,68 @@
...
+  onPermissionError: PropTypes.func,
+  type: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.arrayOf(PropTypes.string),
+  ]),
+  cropParams: PropTypes.object,
+  multiple: PropTypes.bool,
+};
+
+const MediaUploadWrapper = ({ children, mediaUploadProps, ...props }) => {
```
dreamofabear: This is cool, though do we expect to have multiple uses of `MediaUploadWrapper`? 

I think having it inlined in `useQuickActions.js` would be fine too and reduce one layer of reader indirection. The "meat" of this component is only a couple lines of code.

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
dreamofabear: Yea I think this needs to be called after `open()` in `MediaUploadWrapper`.

packages/story-editor/src/components/layout/index.js
```diff
@@ -88,170 +81,25 @@ function Layout({ header, children }) {
...
-      allowedVideoMimeTypes,
-      insertMediaElement,
-      isTranscodingEnabled,
-      optimizeGif,
-      optimizeVideo,
-      showSnackbar,
-      transcodableMimeTypes,
-      updateVideoIsMuted,
-    ]
-  );
```
dreamofabear: 👏

packages/story-editor/src/components/layout/index.js
```diff
@@ -88,170 +81,25 @@ function Layout({ header, children }) {
...
       <LayoutProvider>
         <ChecklistCheckpointProvider>
-          <MediaUploadProvider
-            title={__('Replace media', 'web-stories')}
-            buttonInsertText={__('Insert media', 'web-stories')}
-            onSelect={handleMediaSelect}
-            onClose={resetWithFetch}
-            type={allowedMimeTypes}
-            onSelectErrorMessage={onSelectErrorMessage}
-          >
```
dreamofabear: 🎉

dreamofabear (commented):
Thanks for refactoring!

dreamofabear:
> This works for me.

+1

> I don't think there's a clean one with the way the `MediaUpload` is set up

We could also consider extending `MediaUpload` with other ways to pass the `open` callback to children.

I couldn't find a simpler way of doing that vs. using a render prop. The alternative suggested by [this SO answer](https://stackoverflow.com/questions/32370994/how-to-pass-props-to-this-props-children) with `cloneElement` seems like a bit much, but it does allow passing children components normally.

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
samwhale: oh this is a good point. It'd be good to be able to call all of these together

packages/story-editor/src/app/highlights/quickActions/mediaUploadWrapper.js
samwhale: `useQuickActions` is pretty large already (1000+ lines), but we could move it to that file. I think it'd be ideal to  keep it as a separate component. I'd say this keeps things a little cleaner. Thoughts?

packages/story-editor/src/app/highlights/quickActions/mediaUploadWrapper.js
samwhale: It separate the props that need to be passed to the `MediaUpload` vs the props that need to be passed to the children.

I'll see if this can be removed.

samwhale:
I appreciate everyone taking the time to really think through our options here in an effort to properly support media upload moving forward.

packages/story-editor/src/app/highlights/quickActions/mediaUploadWrapper.js
samwhale: But now that I've cleaned it up we may not need the extra file actually

samwhale:
Found an issue when replacing a gif/video with an image. Need to figure that out still, but with those refactors I think this is more in line with what y'all were thinking.

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
samwhale: Should be fixed now

samwhale:
> Found an issue when replacing a gif/video with an image. Need to figure that out still, but with those refactors I think this is more in line with what y'all were thinking.

should be fixed now

BrittanyIRL (approved):
lgtm, thank you for the notes in code and for the thoughtful discussion and problem solving, i know this was a weird one.

dreamofabear (approved):
🎉 

Re: length of `useQuickActions.js`, we can consider breaking out certain actions into their own files.

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
```diff
@@ -94,6 +106,144 @@ const useQuickActions = () => {
...
+      isTranscodingEnabled,
+      optimizeGif,
+      optimizeVideo,
+      showSnackbar,
+      transcodableMimeTypes,
+      updateVideoIsMuted,
+    ]
+  );
+
+  const MediaPicker = useCallback(
```
mohdsayed: Even though we are using `useCallback`, isn't it odd to define a new component inside a custom hook? Is it possible to move it out?

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
```diff
@@ -94,6 +106,144 @@ const useQuickActions = () => {
...
+          optimizeVideo,
+          optimizeGif,
+        };
+      }
+    );
+
+  const { isTranscodingEnabled } = useFFmpeg();
+
+  // Media Upload Props
+  const allowedMimeTypes = useMemo(() => {
```
mohdsayed: Curious to know, how `useMemo` would help here?

swissspidy: Probably not much.

Looks like a lot of code here (including this bit) is just copied from `packages/story-editor/src/components/library/panes/media/local/mediaPane.js`

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
samwhale: @choumx wanted it to be inlined here in the same hook: https://github.com/google/web-stories-wp/pull/9534#discussion_r743952018

We can leave it here or move it outside the hook. Either is fine with me.

BrittanyIRL: As this has passed QA, thoughts on keeping as is, inline with @choumx 's request? @sayedtaqui and @swissspidy do you want this moved out or do you want to leave it as is? This PR's been open for a long time, it'd be nice to have it done.

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
samwhale: Yeah, happy to remove if we don't want this.

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
dreamofabear: I think it should be in the same file, either inside or outside of the hook is fine. IMO it's ok as-is.

swissspidy: I'd keep it in the same file but outside the hook if possible.

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
mohdsayed: Yeah, let's remove that. Unfortunately, the [official document](https://reactjs.org/docs/hooks-reference.html#usememo) only says that "You may rely on `useMemo` as a performance optimization" but it doesn't elaborate when to use it. Someone on [this article](https://canimerge.com/should-you-really-use-usememo-in-react-lets-find-out/) had done some experiments and the results are interesting. 

So what we learn from that experiment is that `useMemo` comes at a tiny price and if there is not enough complexity, we are doing more harm than good by using it.

image

mohdsayed: In fact, I have previously thought, that if we remove all unnecessary `useMemo` from the whole project, we should gain some performance improvement.

swissspidy: Yeah we're definitely overusing `useMemo` in many places.

dreamofabear: Surveying use of `useMemo` sounds good. It definitely improved performance for reducing child re-renders in a few places though.

mohdsayed (approved):

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
mohdsayed: Would using `React.memo` for the whole component be a better choice for those special places instead?

dreamofabear: Ah yea, I was referring to use of `React.memo` not `useMemo`.

PR #9535, opened 2021-10-26
UI Simplification: Selection Panel

### LINKED ISSUE #9460: UI Simplifications: Selection Panel
## Feature Description
See [Figma for details](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor-Sprint-Feature-Level-Designs_?node-id=10250%3A435609).

- [ ] Change the name of Size & Position to "Selection"
- [ ] Remove the Layer panel and move the opacity field to the new "Selection" panel
- [ ] Remove the Corner radius panel and move the input to the "Selection" panel. Redesign the input (see locked vs [unlocked](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor-Sprint-Feature-Level-Designs_?node-id=10250%3A436882) in Figma)
- [ ] Ensure the changes are reflected accurately in all element panels.

miina:
## Summary
Simplifies the Selection panel (previously Size & Position)

## To-do

- [x] Fix grid area lint
- [x] Fix tests

## User-facing changes
| Before | After |
| ------ | ----- |
| Screenshot 2021-10-26 at 17 27 41 | Screenshot 2021-10-26 at 17 26 08 |
| Screenshot 2021-10-26 at 17 27 50 | Screenshot 2021-10-26 at 17 25 53 |

## Testing Instructions

1. Verify that the Layer panel is not a separate panel anymore and is now under the new Selection panel (verify for each element type).
2. Verify that Opacity can be changed from the Opacity input (previously was in the Layer panel)
3. Verify that Corner radius panel is not a separate panel anymore and is now an input under the Selection panel.
4. Verify the unlocked and locked corner radius input style per [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor-Sprint-Feature-Level-Designs_?node-id=10250%3A436882) for the corner radius input(s)
5. Verify the corner radius works as expected still.

Note: the opacity input icon (the drop) looks a bit different than in Figma, that will be fixed in another PR.

No

No

No

## Checklist

---

Fixes #9460


packages/story-editor/src/components/panels/design/sizePosition/sizePosition.js
```diff
@@ -221,8 +248,12 @@ function SizePositionPanel({
     };
   }, []);
   return (
-    <SimplePanel name="size" title={__('Size & Position', 'web-stories')}>
-      <Grid isSingleMedia={isMedia && isSingleElement}>
+    <SimplePanel name="size" title={__('Selection', 'web-stories')}>
```
miina, line 250: @barklund I'm having doubts about if to also change the panel name in code instead of continuing using `sizePosition`. In a way, it's just the name that changed, we probably shouldn't change code so much every time we need to change the visible name. However, `sizePosition` is not fully accurate anymore.
At the same time, using "selection" can cause confusion since that's used for element selection most of the time.

I left it as is since that seemed reasonable enough for now.
Any thoughts on this?

Trapsta (approved):
LGTM

packages/story-editor/src/components/panels/design/sizePosition/sizePosition.js
barklund: Let's leave that as-is for now. When we're done moving things around, we can always take a holistic look and rename panels where appropriate.

packages/story-editor/src/components/panels/design/sizePosition/radius.js
```diff
@@ -0,0 +1,287 @@
+/*
```
barklund, line 1: Many of my comments over [here](https://github.com/google/web-stories-wp/pull/9571#pullrequestreview-792473581) apply to this component as well. Only now noticing, that this is probably the "original" for many of the proposed changes over there.

barklund (approved):
Looks good and seems to work across panels/element types. I just have the same overall comments (especially on the horizontally stacked inputs) as I did over in @Trapsta's [PR](https://github.com/google/web-stories-wp/pull/9571#pullrequestreview-792473581).

Also, this does not reduce the size of the general inputs, is that on purpose? In Figma, inputs are now only 32px tall. In this PR, the inputs are still 36px tall.

packages/story-editor/src/components/panels/design/sizePosition/radius.js
miina: All the inputs are still 36px in Figma:
Screenshot 2021-10-29 at 12 23 09

This is the [latest Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=10250%3A435766). Are you seeing something else in the same Figma general area / another element panel?

miina: @barklund Merging this as-is to be followed up on the shared component in Jasper's PR.

PR #9548, opened 2021-10-27
Improve AMP plugin availability detection

swissspidy:
## Context

From what I've seen in some support requests, it looks like there are cases where neither our AMP sanitization nor the one built into the AMP plugin runs.

This might be caused by some other plugin or theme on the user's site disabling some of the AMP plugin's functionality. (don't know yet what the cause usually is).

Right now, we defer AMP sanitization to the plugin if it's installed. However, that doesn't mean it's actually running.

For example, if another plugin does something like `add_filter( 'amp_is_enabled', '__return_false' );`, our plugin would defer sanitization, but the AMP plugin would not do anything.

## Summary

This PR adds some further checks to verify whether the AMP plugin is actually going to sanitize the story or not, so that we will take over if necessary.

## Relevant Technical Choices

* Uses `amp_is_enabled()` despite being marked as internal due to lack of alternatives

## To-do

* Testing
* Maybe use different methods to detect AMP availability

## User-facing changes

N/A

## Testing Instructions

1. Add `add_filter( 'amp_is_enabled', '__return_false' );` somewhere in a plugin
2. Verify that a web story is still being properly sanitized

No

No

No

## Checklist

---

Fixes #

### DIFF
--- a/includes/AMP/Output_Buffer.php
+++ b/includes/AMP/Output_Buffer.php
@@ -119,14 +119,24 @@ class Output_Buffer extends Service_Base implements Conditional {
 	/**
 	 * Check whether the conditional object is currently needed.
 	 *
+	 * If the AMP plugin is installed and available in a version >= than ours,
+	 * all sanitization and optimization should be delegated to the AMP plugin.
+	 * But ONLY if AMP logic has not been disabled through any of its available filters.
+	 *
 	 * @since 1.10.0
 	 *
 	 * @return bool Whether the conditional object is needed.
 	 */
 	public static function is_needed(): bool {
-		// If the AMP plugin is installed and available in a version >= than ours,
-		// all sanitization and optimization should be delegated to the AMP plugin.
-		return ! defined( '\AMP__VERSION' ) || ( defined( '\AMP__VERSION' ) && version_compare( \AMP__VERSION, WEBSTORIES_AMP_VERSION, '<' ) );
+		$current_post = get_post();
+
+		$has_old_amp_version = ! defined( '\AMP__VERSION' ) || ( defined( '\AMP__VERSION' ) && version_compare( \AMP__VERSION, WEBSTORIES_AMP_VERSION, '<' ) );
+		$amp_available       = function_exists( 'amp_is_available' ) && amp_is_available();
+		$amp_enabled         = function_exists( 'amp_is_enabled' ) && amp_is_enabled(); // Technically an internal method.
+		$amp_initialized     = did_action( 'amp_init' ) > 0;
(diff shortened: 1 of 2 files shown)


swissspidy:
@westonruter Curious to hear whether there is a better way to detect this.

An alternative solution might be to _always_ do sanitization in the Web Stories plugin. But that could cause issues with double-sanitization if both plugins run it.

includes/AMP/Output_Buffer.php
```diff
@@ -119,14 +119,24 @@ public static function get_registration_action_priority(): int {
...
 	 */
 	public static function is_needed(): bool {
-		// If the AMP plugin is installed and available in a version >= than ours,
-		// all sanitization and optimization should be delegated to the AMP plugin.
-		return ! defined( '\AMP__VERSION' ) || ( defined( '\AMP__VERSION' ) && version_compare( \AMP__VERSION, WEBSTORIES_AMP_VERSION, '<' ) );
+		$current_post = get_post();
+
+		$has_old_amp_version = ! defined( '\AMP__VERSION' ) || ( defined( '\AMP__VERSION' ) && version_compare( \AMP__VERSION, WEBSTORIES_AMP_VERSION, '<' ) );
+		$amp_available       = function_exists( 'amp_is_available' ) && amp_is_available();
+		$amp_enabled         = function_exists( 'amp_is_enabled' ) && amp_is_enabled(); // Technically an internal method.
```
westonruter, line 135: This is a good check. If someone is turning off AMP by filtering `amp_is_enabled` to `false`, then the constants and functions will still exist but AMP won't initialize. So this makes sense to me.

includes/AMP/Output_Buffer.php
```diff
@@ -119,14 +119,24 @@ public static function get_registration_action_priority(): int {
...
+		$amp_initialized     = did_action( 'amp_init' ) > 0;
+		$amp_supported_post  = function_exists( 'amp_is_post_supported' ) && amp_is_post_supported( $current_post->ID ?? 0 );
```
westonruter, line 137: An alternative to doing these checks would be to check if `\AMP_Theme_Support::is_output_buffering()` returns true. However, to do this you'd need to increase the action priority by one so that it would run at `template_redirect` after `\AMP_Theme_Support::start_output_buffering()`.

swissspidy: Both `\AMP_Theme_Support::start_output_buffering()` and this class/method run at `PHP_INT_MAX` already 👀

westonruter: Oh right, I got that reversed. I first thought it was running as _early_ as possible, but it's running as _late_ as possible.

PR #9551, opened 2021-10-27
Media Library: Remove `aria-label` from image and video

### LINKED ISSUE #7574: Media Library: Incorrect use of `aria-label`
## Bug Description

Images and videos in the media library incorrectly have `aria-label` attributes.

Both should only have `alt` attributes.

Offending code:

https://github.com/google/web-stories-wp/blob/4c891bd2b6b10815c72b06d7bbccd15718da79b1/assets/src/edit-story/components/library/panes/media/common/innerElement.js#L135-L167

## Expected Behaviour

No usage of `aria-label`. Just `alt`

## Screenshots

Screenshot 2021-05-18 at 23 14 35

Screenshot 2021-05-18 at 23 14 15

## Additional Context

- Plugin Version: 1.7
- WordPress Version: 5.7
- Operating System: macOS
- Browser: Chrome

---

## Implementation Brief

* Remove the two `aria-label` props in the code section shared above

samwhale:
## Summary

Remove the `aria-label` attribute from the image and video tags. Instead use:
* `alt` for `<image />`
* `title` for `<video />`

## Relevant Technical Choices

n/a

## To-do

n/a

## User-facing changes

n/a

**Image:**

**Video:**
 

**GIF:**

## Testing Instructions

1. Open dev tools
2. View an element in the media panel (media 3p works the best because they all have lots of attributes)
3. Inspect the `img` or `video` element that is rendered. Should not see `aria-label`. Should see `alt` for an `img` and `title` for a video.

no

no

no

## Checklist

---

Fixes #7574

### DIFF
--- a/packages/story-editor/src/components/library/test/mediaElement.js
+++ b/packages/story-editor/src/components/library/test/mediaElement.js
@@ -18,6 +18,7 @@
  * External dependencies
  */
 import { fireEvent, screen } from '@testing-library/react';
+import { axe } from 'jest-axe';
 
 /**
  * Internal dependencies
@@ -26,6 +27,28 @@ import MediaElement from '../panes/media/common/mediaElement';
 import { renderWithTheme } from '../../../testUtils';
 import CanvasContext from '../../../app/canvas/context';
 import StoryContext from '../../../app/story/context';
+
+const IMAGE_RESOURCE = {
+  id: 789,
+  src: 'http://image-url.com',
+  type: 'image',
+  mimeType: 'image/png',
+  width: 100,
+  height: 100,
+  local: true, // Not yet uploaded
+  alt: 'image :)',
+};
+
+const VIDEO_RESOURCE = {
+  id: 456,
+  src: 'http://video-url.com',
+  type: 'video',
+  mimeType: 'video/mp4',
+  width: 100,
+  height: 100,
+  local: true, // Not yet uploaded
+  alt: 'video :)',
+};
 
 const renderMediaElement = (resource, providerType, canEditMedia = true) => {
   const canvasContext = {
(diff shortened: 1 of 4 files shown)


packages/story-editor/src/components/library/test/mediaElement.js
```diff
@@ -17,7 +17,7 @@
 /**
  * External dependencies
  */
-import { fireEvent } from '@testing-library/react';
+import { fireEvent, screen } from '@testing-library/react';
```
swissspidy, line 20: Some aXe coverage for these tests would be nice (`toHaveNoViolations`)

samwhale: can do!

packages/story-editor/src/components/library/test/mediaElement.js
```diff
@@ -69,122 +92,92 @@ const renderMediaElement = (resource, providerType, canEditMedia = true) => {
...
+          ...resource,
+          local,
+        },
+        'local'
+      );
+
+      const results = await axe(container);
+      expect(results).toHaveNoViolations();
+    }
+  );
```
samwhale: 💭 Probably don't need one of these for each resource type and whether or not it's uploaded. Can remove the `it.each` if we just want one check.

swissspidy: I don't think `local` is relevant here, so I'd remove that from the table.

But I'd add a gif element to the list.

BrittanyIRL:
🤔  these karma failures seem like they might be residual from these updates, are they?

BrittanyIRL (approved):
🕺

swissspidy (approved):

Issue #7365, opened 2021-05-04
E2E Tests: Review disabled aXe rules


swissspidy:
## Task Description

When we initially added aXe testing to the e2e test setup, we've disable some rules that caused frequent errors, either because of known bugs in the editor or issues in WordPress.

https://github.com/google/web-stories-wp/blob/d0ffcff89e8840ad9fd0bbe7fd93d1e4235a635b/packages/e2e-tests/src/config/bootstrap.js#L204-L219

Let's go over each disabled rule to:

1. See whether the rule can be safely removed (i.e. there are no errors)
2. Identify where the issues are coming from (editor, dashboard, WordPress itself, something else) and their nature (valid bug, false positive)
3. Document the reason for it being here in this ticket
4. Create tickets to fix bugs where applicable

The goal is to reduce the number of disabled rules and improve the accessibility of the plugin.

spacedmonkey:
All of these accessibility issues seem to be related to the editor. Many issues are relatively simple to fix. I have list of common issues / low hanging fruit to fix. I would recommend we create tickets for of these rules and re-enable them one by one. 

### aria-allowed-attr
```
    Rule: "aria-allowed-attr" (Elements must only use allowed ARIA attributes)
    Help: https://dequeuniversity.com/rules/axe/4.2/aria-allowed-attr?application=axe-puppeteer
    Affected Nodes:
      svg[aria-label="Media\ Gallery"]
        Fix ALL of the following:
          - ARIA attribute cannot be used: aria-label.
      svg[aria-label="Explore\ Media"]
        Fix ALL of the following:
          - ARIA attribute cannot be used: aria-label.
      #text-tab-icon
        Fix ALL of the following:
          - ARIA attribute cannot be used: aria-label.
      svg[aria-label="Shapes\ library"]
        Fix ALL of the following:
          - ARIA attribute cannot be used: aria-label.
      svg[aria-label="Page\ templates\ library"]
        Fix ALL of the following:
          - ARIA attribute cannot be used: aria-label.
      video
        Fix ALL of the following:
          - ARIA attribute cannot be used: aria-label.
      .frameElement__Wrapper-sc-1pyee9f-0
        Fix ALL of the following:
          - ARIA attribute cannot be used: aria-labelledby.
```

These errors seem to valid. For the most part there incorrect use of aria-label on svg icons.  Also it seems like video tags do not this attribute. 

### color contrast

```
[...]

PR #9566, opened 2021-10-27
Library Tabs: change aria-label to title on svgs

### LINKED ISSUE #7573: Library Tabs: Incorrect use of `aria-label` on SVGs
## Bug Description

The library tabs in the editor use `aria-label` attributes on the SVG icons. Example:

https://github.com/google/web-stories-wp/blob/4c891bd2b6b10815c72b06d7bbccd15718da79b1/assets/src/edit-story/components/library/panes/media/local/mediaIcon.js

While some assistive technologies understand that, it's recommended to use `<title>` elements within the SVG.

Fixing this is as simple as replacing `aria-label` props with `title` props. webpack will turn them into `<title>` elements within the `<svg>`.

## Expected Behaviour

Using `title` instead of `aria-label`

## Screenshots

Screenshot 2021-05-18 at 23 10 43

## Additional Context

- Plugin Version: 1.7
- WordPress Version: 5.7
- Operating System: macOS
- Browser: Chrome

---

BrookeGraham:
## Context

The library tabs in the editor use aria-label attributes on the SVG icons. While some assistive technologies understand that, it's recommended to use <title> elements within the SVG.

## Summary

Replaced svg icons' `aria-label` attribute with the `title` attribute. 

## Testing Instructions

1. Open dev dev tool and inspect library tabs in the editor
2. Note the svgs' now have a title element instead of the aria-label attribute. 

no

no

no

## Checklist

---

Fixes #7573

### DIFF
--- a/packages/story-editor/src/components/library/karma/mediaTab.karma.js
+++ b/packages/story-editor/src/components/library/karma/mediaTab.karma.js
@@ -79,4 +79,11 @@ describe('Library Media Tab', () => {
       ).toBeDefined();
     });
   });
+  //todo: add axe for all tabs in libraryTabs.karma.js once the text tab is no longer a button within a button see issue #7365
+  describe('library Media tab should have no aXe accessibility violations', () => {
+    it('should pass accessibility', async () => {
+      const mediaTab = fixture.container.querySelector(`#library-tab-media`);
+      await expectAsync(mediaTab).toHaveNoViolations();
+    });
+  });
 });
--- a/packages/story-editor/src/components/library/panes/elements/elementsIcon.js
+++ b/packages/story-editor/src/components/library/panes/elements/elementsIcon.js
@@ -21,7 +21,7 @@ import { __ } from '@web-stories-wp/i18n';
 import { Icons } from '@web-stories-wp/design-system';
 
 function ElementsIcon() {
-  return <Icons.Box4 aria-label={__('Elements library', 'web-stories')} />;
+  return <Icons.Box4 title={__('Elements library', 'web-stories')} />;
 }
 
 export default ElementsIcon;
--- a/packages/story-editor/src/components/library/panes/media/local/mediaIcon.js
+++ b/packages/story-editor/src/components/library/panes/media/local/mediaIcon.js
@@ -21,7 +21,7 @@ import { __ } from '@web-stories-wp/i18n';
 import { Icons } from '@web-stories-wp/design-system';
 
 function MediaIcon() {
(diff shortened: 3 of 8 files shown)


BrittanyIRL:
This update should mean that the `label` rule that was disabled here: https://github.com/google/web-stories-wp/issues/7365 can be able to be enabled (that's a mouthful!) - if that's the case, can you enable it and if it's not leave a note what's left?

I might be jumping the gun here since the status of this PR still says 'in progress' but, It would also be great to add in a test that can verify that these aren't an issue any more (and prevent regressions). Since they're all in the panel headers (quick actions has a validation check already), perhaps just one test in the `mediaTab.karma.js` file to validate that there's no aXe issues present (` .toHaveNoViolations()`) there?

samwhale (approved):
Rad - couple things:

1. does this one need a title too? https://github.com/google/web-stories-wp/blob/main/packages/story-editor/src/components/checklist/checklistIcon.js
2. I think this one _does not_ need an `alt` (hehe oops, my bad) https://github.com/google/web-stories-wp/blob/main/packages/story-editor/src/components/library/panes/media/media3p/media3pIcon.js#L26. Could we remove that?

BrookeGraham:
@samwhale  
>does this one need a title too? https://github.com/google/web-stories-wp/blob/main/packages/story-editor/src/components/checklist/checklistIcon.js

Turns out this is nested inside a `button` with an `aria-label`.  Do you think I should still add it in cases where the icon gets used with the button wrapper? 

@BrittanyIRL 
>This update should mean that the label rule that was disabled here: #7365 can be able to be enabled (that's a mouthful!) - if that's the case, can you enable it and if it's not leave a note what's left?

The `label` rule was removed. We _should_ be able to remove the `aria-allowed-attr` very soon though. Once #7574 is pushed we only would need to fix one other issue. We need to remove the `aria-labelledBy` from  `frameElement` at least I think this is [where](https://github.com/google/web-stories-wp/blob/main/packages/story-editor/src/components/canvas/frameElement.js#L190) the issue is stemming from.

BrittanyIRL (approved):
lgtm

BrittanyIRL:
> The `label` rule was removed. We _should_ be able to remove the `aria-allowed-attr` very soon though. Once #7574 is pushed we only would need to fix one other issue. We need to remove the `aria-labelledBy` from `frameElement` at least I think this is [where](https://github.com/google/web-stories-wp/blob/main/packages/story-editor/src/components/canvas/frameElement.js#L190) the issue is stemming from.

awesome! that can be separate from this for sure.

PR #9576, opened 2021-10-29
Added video frame strip to the trim rail background

### LINKED ISSUE #9104: Display frames in the trim rail behind the handles
## Feature Description

Show actual frames from the video in the trim rail.

Things to consider:

* Video dimensions
* Mask dimensions
* Video length
* Number of frames to display

## Additional Context

Current frame-less design:
Screen Shot 2021-09-22 at 10 21 05

Goal design:
Screen Shot 2021-09-22 at 10 22 02

barklund:
## Context

This implements #9104 as per the ticket description.

## Relevant Technical Choices

* I use the actual size, scale, and focal point of the video so the trim rail will only show the same frame snippets that you see in the actual video (e.g. if you zoom in on a small part, only that part of the video will be snapshotted for every frame). I'm not sure if this is standard though.

## Todo

* [ ] Should we display a loader while the strip generates?

## User-facing changes

| Before | After |
|-|-|
| Screen Shot 2021-10-29 at 00 29 44 | Screen Shot 2021-10-29 at 00 29 33 |

## Testing Instructions

1. Add a local video to a page
2. Enter trim mode
3. Observe that the trim rail background is a film strip from the video being trimmed

## Checklist

---

Fixes #9104


packages/media/src/generateVideoStrip.js
```diff
@@ -0,0 +1,146 @@
...
+}
+
+/*
+ * Returns an image of the first frame of a given video.
+ *
+ * @see https://github.com/ampproject/amp-wp/blob/c5fba13dd17d4f713c9889d26898aec6091e421b/assets/src/stories-editor/helpers/uploadVideoFrame.js#L10-L39
+ * @param {string} src Video src URL.
+ * @return {Promise<string>} The extracted image in base64-encoded format.
+ */
+async function generateVideoStrip(element, resource, stripWidth, stripHeight) {
```
barklund, line 54: I didn't find a good way to cut this much shorter, but I feel like it's fairly easy to read with all the comments in there.

packages/story-editor/src/components/videoTrim/useRailBackground.js
```diff
@@ -0,0 +1,38 @@
...
+import { generateVideoStrip } from '@web-stories-wp/media';
+
+function useRailBackground(isReady, videoData, railWidth) {
+  const [railBackgroundImage, setRailBackgroundImage] = useState(null);
+  useEffect(() => {
+    if (!isReady) {
+      return;
+    }
+    const { element, resource } = videoData;
+    generateVideoStrip(element, resource, railWidth, 36).then(
```
miina: Nit: for readability, let's create a constant for the height instead of using magic numbers.

packages/media/src/generateVideoStrip.js
```diff
@@ -0,0 +1,148 @@
...
+  const frameCount = Math.ceil(stripWidth / stripFrameWidth);
+
+  // Calculate what the actual width of each frame will have to be to squeeze the images in
+  const actualStripFrameWidth = stripWidth / frameCount;
+
+  // Scale the original frame width down by the same amount so we don't squish the frames
+  const actualFrameWidth =
+    (actualStripFrameWidth / stripFrameWidth) * frameWidth;
+
+  // get video size based on scale and focal
```
miina, line 79: Just realized that flipping is not considered for trimming at all currently -- if the video is flipped, it displays as originally in trim mode. Probably would be good to create a followup issue for that.

packages/media/src/generateVideoStrip.js
```diff
@@ -0,0 +1,148 @@
...
+  } = element;
+
+  const { src, length } = resource;
+
+  // Calculate the scaled frame width in the strip
+  const stripFrameHeight = stripHeight;
+  const stripFrameWidth = (stripFrameHeight / frameHeight) * frameWidth;
+
+  // Calculate how many frames we can fit in the strip (rounded up)
+  const frameCount = Math.ceil(stripWidth / stripFrameWidth);
```
miina, line 70: Per Figma, it shows having more frames that are overlapping partially (the last time I checked at least). Is this done differently for performance reasons or something else? (Fine by me as is, just curious)

packages/media/src/generateVideoStrip.js
```diff
@@ -0,0 +1,148 @@
...
+  video.width = videoWidth;
+  video.height = videoHeight;
+
+  // Create the target canvas
+  const canvas = document.createElement('canvas');
+  canvas.width = stripWidth;
+  canvas.height = stripHeight;
+  const ctx = canvas.getContext('2d');
+
+  window.dummy = { video, canvas };
```
miina: Nit: Maybe `dummyStripData` or sth similar for clarity / being more specific? (Feel free to ignore)

miina (approved):
Works well!

packages/story-editor/src/components/videoTrim/useRailBackground.js
barklund: Oh yes, forgot about that part. It's used in another file as well.

packages/media/src/generateVideoStrip.js
barklund: Oh, this was for debugging only (so I could append them while making sure I was generating the strip correctly). Will remove!

packages/media/src/generateVideoStrip.js
barklund: Oh, excellent point. I'll file as follow-up. I'm afraid to check rotation, transparency, and other stuff now though 😂 .

barklund: Okay, rotation works, transparency doesn't, but that's probably "by design" by accident, right? It makes sense that the video changes temporarily to full opacity while trimming, and then when exiting switches back to the actual transparency level.

packages/story-editor/src/components/videoTrim/useRailBackground.js
barklund: Done in bec708cc000a209ace8c2c639dcfc4b277850b27.

packages/media/src/generateVideoStrip.js
barklund: Removed in 939d895d7821e01629d01eb48aad691d8ca25dea.

packages/media/src/generateVideoStrip.js
barklund: I don't think the frame strips in Figma are carefully crafted with any specificity in mind - it doesn't show any details about how it works for different size and length videos or anything like that. I implemented what I thought to make sense without more specific instruction. Let's test it out and ask UX if they feel it suffices.

packages/media/src/generateVideoStrip.js
barklund: Filed in 9589.

Issue #9575, opened 2021-10-28
Use Custom Page Template Images


BrittanyIRL:
## Description 

Working off of this branch: https://github.com/google/web-stories-wp/pull/9584#issuecomment-962029745 (#9574), let's figure out the best way to get these images consistent.

Let's do PRs into #9584 to keep things clean and clear. 

To do: 
- [ ] Consistent image capture for new custom page templates as well as backfilling 
- [ ] Make sure they have a big enough size when captured (no thumbnail sizing) 
- [ ] Any refactor to use the same grid codebase as default page templates 
- [ ] Any updated tests for parity with default page templates 
- [ ] Document perf gains

## Acceptance Criteria 
- Static custom page templates

PR #9584, opened 2021-10-29
Page Templates: Save image for custom page templates

### LINKED ISSUE #9574: Save Custom Page Templates as Images
## Description 

When a page is saved as a custom page template, we should use the same `html-to-img` pattern as in the carousel to get an image of that page and save that image to media. 

This is going to need some back and forth between back and front end to make sure that the new data is getting saved properly. Probably we can use the media optimization flow as a starting place for how to store custom page template images without making them part of the main media gallery. Ideally,  the custom page template now has a media id for that img of the page and if that media is available it'll render the img instead of loading a preview of the page. 

If a custom page template loads in the panel and doesn't have an image yet then we should take a snapshot of that preview once it's loaded and save it to media as well. 

## Acceptance Criteria 
- images getting saved in media of custom page templates 
- if existing custom page templates have no image yet, when those load, images are added to media

swissspidy:
## Context

When saving custom page templates we want to create and upload an image of it so that we can display them in the list as an image instead of a "live" preview.

## To-do

- [x] Reorganize / DRY code
- [x] Don't use fullbleedContainer, but a custom node for rendering a 640x853 image (or another specified dimension)
- [x] Tests

## User-facing changes

Saved templates should now be in the 2:3 size like the default templates. It should also render images instead of the element page as needed.

## Testing Instructions

1. Go to the editor
2. make some edits to a page
3. save the page as a page template
4. see that the page shows up in the page template
5. should also eventually be rendered as an image, but that might take some time to do depending on how many saved templates you have that need images generated for them.

No

No

No

## Checklist

---

Fixes #9574


swissspidy:
@BrittanyIRL This is an early draft of the page templates images functionality. PTAL.

Image creation upon initial page template creation works well already. Just the image size is a bit random as it depends on the canvas size.

We need to somehow get a solution where we can render the image at an arbitrary size.

Once we have that, the backfill logic to update existing page templates will be a breeze, as most of the plumbing is already handled.

If Pea pod could take over the remaining parts (for example as PRs against this branch), I wouldn't mind :-)

includes/REST_API/Page_Template_Controller.php
```diff
@@ -102,4 +102,60 @@ public function get_item_schema(): array {
...
+		if ( ! $post_type || ! current_user_can( $post_type->cap->edit_posts ) ) {
+			return new WP_Error(
+				'rest_forbidden_context',
+				__( 'Sorry, you are not allowed to edit page templates.', 'web-stories' ),
+				[ 'status' => rest_authorization_required_code() ]
+			);
+		}
+
+		return true;
+	}
```
spacedmonkey, line 166: Why are these needed?

spacedmonkey:
This looks good. 👍

includes/REST_API/Page_Template_Controller.php
swissspidy: I realized that otherwise page templates were publicly accessible via the REST API, which isn't really intended.

BrittanyIRL:
I'm not ignoring this, i'm really stoked about it and floored you made the time so fast for this. I am trying to knock out some stuff in our sprint and make sure that everything is caught up and then i'll give this a proper look. Should get to it by tomorrow at the latest.

includes/REST_API/Page_Template_Controller.php
spacedmonkey, line 166:
```suggestion
	public function get_item_permissions_check( $request ) {
		return $this->update_item_permissions_check( $request );
	}
```

Could you just do this then?

maxyinger (commented):
Does it make more sense to make the image from saved template thumbnails? Seems like it may have less variability than using the canvas.

also if we're worried about image sizes, we could just check if the image we receive is smaller than the element is, then regenerate it at the larger size if needed.

Is there a reason why the canvas is more beneficial than the page preview in the saved templates panel?

packages/story-editor/src/components/library/panes/pageTemplates/templateSave.js
```diff
@@ -96,26 +98,67 @@ const SaveButton = styled.button`
...
+      let imageId;
+
+      if (hasUploadMediaAction) {
+        // TODO(#9574): Generate image using html-to-image here.
+
+        const htmlToImage = await import(
+          /* webpackChunkName: "chunk-html-to-image" */ 'html-to-image'
+        );
+
+        // TODO: We need something more reliable than fullbleedContainer. What if the canvas is small or zoomed in?
```
BrittanyIRL: 🤔  could it make sense to treat getting the screenshot of a custom template like optimizing media? So you add a page as a custom template and then that shows the same preview node we're used to seeing in the custom page templates view (but maybe we only display the safe area), while this is rendered we show the loading/progress bar in the UI and use the same fallback code you outlined earlier in this PR to grab the screenshot. This would eliminate weird situations where the canvas is zoomed in or super small since the panel won't shrink below 360px, it would also unify the process regardless of backfill/newly added template.

swissspidy:
> Does it make more sense to make the image from saved template thumbnails? Seems like it may have less variability than using the canvas.

It does yeah. At least when backfilling existing ones

Just need to make sure that we have consistent image sizes and good enough quality.

When saving the current page as a template, it would be nice to be able to generate an image right away instead of grabbing it from the thumbnail.

Perhaps I could just render the template somewhere not in the DOM to generate the image. So I don't have to rely neither on the thumbnail nor on the canvas.

> Is there a reason why the canvas is more beneficial than the page preview in the saved templates panel?

This code is mainly referencing the canvas because I copy-pasted stuff.

BrittanyIRL:
> Perhaps I could just render the template somewhere not in the DOM to generate the image. So I don't have to rely neither on the thumbnail nor on the canvas.

@swissspidy do you want pea pod to take this from here? You've done all the infra work to make sure that we can save the images appropriately. We have the bandwidth. I know you mentioned us taking it on at some point. Clearly we'll continue having back and forth with you on the PR but we can totally prioritize this for our next sprint, we have room. 

we'll use #9575 and (as you mentioned in one of these issues) just do review/merge against this branch to keep things clean.

swissspidy:
If you have bandwidth and can work against this branch to make reviews easier, that'd be greatly appreciated!

BrittanyIRL:
> If you have bandwidth and can work against this branch to make reviews easier, that'd be greatly appreciated!

yeah, we got this! what you've done already is incredible.

maxyinger:
@swissspidy Curious to get your thoughts on this approach for generating images offscreen before I move forward with it:
https://github.com/google/web-stories-wp/pull/9700

There still might be some lingering errors that get logged to the console that I have to clean up, but just want to get a pulse check on the general approach before I move forward with it

maxyinger:
@swissspidy just pinging to let you know I merged #9700 into here.

maxyinger (approved):

packages/story-editor/src/utils/storyPageToDataUrl.js
```diff
@@ -0,0 +1,163 @@
...
+        ref={resolverRef}
+        page={page}
+        width={width}
+        height={height}
+        containerHeight={containerHeight}
+      />,
+      bufferRoot
+    );
+  });
+
```
maxyinger, line 146: promise pattern to wait a component and all of it's descendants to render. Could be useful in our karma `fixture.render` method

PR #9586, opened 2021-10-29
Layers Panel: constrain max height of layer panel

### LINKED ISSUE #9542: Layer panel: Remove scrollbar when the layer panel is max height
## Bug Description

Expanding the layer panel to its maximum height makes a scrollbar appear for no reason.

## Expected Behaviour

The layer panel should only show a scrollbar if there are too many layers to display on the screen.

## Steps to Reproduce

Expand the layer panel all the way up.

## Additional Context

- Plugin Version: 1.14.0-alpha
- WordPress Version: 5.8.1
- Operating System: MacOSX Sierra
- Browser: Chrome

BrookeGraham:
## Context
Expanding the layer panel to its maximum height makes a scrollbar appear. 

## Summary

This scroll bar was appearing because the actual height of the layers panel was set to `99999999`. An overriding `maxHeight` prop was set to the `MAX_HEIGHT_DEFAULT` from `./panel`

## Relevant Technical Choices

We want the max height to fill the space underneath the document/design tab bar. Since the document/design tab bar is 50px subtract that from the full height of the inspector and use this value as the `maxHeight` override prop. 

## To-do

n/a

### Old: 
Screen Shot 2021-10-29 at 2 25 39 PM

### New:

Screen Shot 2021-10-29 at 2 26 28 PM
## Testing Instructions

1. open the layers panel
2. drag to resize to the full height
3. notice the scroll bar no longer appears. 

no

no

no

## Checklist

---

Fixes #9542

### DIFF
--- a/packages/story-editor/src/components/panels/design/layer/layerPanel.js
+++ b/packages/story-editor/src/components/panels/design/layer/layerPanel.js
@@ -24,12 +24,8 @@ import styled from 'styled-components';
 /**
  * Internal dependencies
  */
-import {
-  MAX_HEIGHT_DEFAULT,
-  Panel,
-  PanelTitle,
-  PanelContent,
-} from '../../panel';
+import { Panel, PanelTitle, PanelContent } from '../../panel';
+import useInspector from '../../../inspector/useInspector';
 import LayerList from './layerList';
 import useLayers from './useLayers';
 
@@ -56,6 +52,14 @@ const Divider = styled.div`
 function LayerPanel() {
   const layers = useLayers();
 
+  const {
+    state: { inspectorContentHeight },
+  } = useInspector();
+
+  // We want the max height to fill the space underneath the document/design tab bar.
+  // Since the document/design tab bar is 50px subtract that from the full height of the inspector.
+  const maxHeight = inspectorContentHeight - 50;
+
   const initialHeight = useMemo(() => Math.round(window.innerHeight / 4), []);
 
   return (
@@ -72,7 +76,7 @@ function LayerPanel() {
           isSecondary
           isResizable
           count={layers?.length}
-          maxHeight={MAX_HEIGHT_DEFAULT}
+          maxHeight={maxHeight}
         >
           {__('Layers', 'web-stories')}
         </StyledPanelTitle>
--- a/packages/story-editor/src/components/panels/design/layer/layerPanel.js
+++ b/packages/story-editor/src/components/panels/design/layer/layerPanel.js
@@ -26,6 +26,7 @@ import styled from 'styled-components';
  */
 import { Panel, PanelTitle, PanelContent } from '../../panel';
 import useInspector from '../../../inspector/useInspector';
+import { TAB_HEIGHT, TAB_VERTICAL_MARGIN } from '../../../tabview';
 import LayerList from './layerList';
 import useLayers from './useLayers';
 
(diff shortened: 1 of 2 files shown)


packages/story-editor/src/components/panels/design/layer/layerPanel.js
```diff
@@ -56,6 +52,14 @@ const Divider = styled.div`
 function LayerPanel() {
   const layers = useLayers();
 
+  const {
+    state: { inspectorContentHeight },
+  } = useInspector();
+
+  // We want the max height to fill the space underneath the document/design tab bar.
+  // Since the document/design tab bar is 50px subtract that from the full height of the inspector.
+  const maxHeight = inspectorContentHeight - 50;
```
BrittanyIRL: 🤔   i'm trying to think there's a way we can pull in the height of the tab bar instead of assigning it by force here - just for when the designs update. It's set in components/tabview > `TabElement` and it's a combo of 32px height and 19px of vertical margin.

BrookeGraham: I could probably calculate it via `useInspector` -> `tabRef` in the `InspectorProvider` . I'll. look into it.

maxyinger: I think Brittany's thinking something much simpler. just making like `TAB_HEIGHT` & `TAB_MARGIN_VERTICAL` constants and using them both here and in the `TabElement`.

maxyinger (approved):
Looks good.

packages/story-editor/src/components/panels/design/layer/layerPanel.js
BrittanyIRL: thanks, max - yup! keep it simple 😎

samwhale (approved):
Agree with the new way. This works well too, but the simpler approach will be easier ✅

packages/story-editor/src/components/panels/design/layer/layerPanel.js
BrookeGraham: oh duh! Cool

packages/story-editor/src/components/panels/design/layer/layerPanel.js
```diff
@@ -56,6 +53,16 @@ const Divider = styled.div`
...
   const layers = useLayers();
 
+  const {
+    state: { inspectorContentHeight },
+  } = useInspector();
+
+  // We want the max height to fill the space underneath the document/design tab bar.
+  // Since the document/design tab bar has top and bottom margin in addition to a static
+  // height subtract sum from the full height of the inspector.
+  const tabHeight = 2 * TAB_VERTICAL_MARGIN + TAB_HEIGHT;
```
BrittanyIRL, line 63: ❤️  thank youuuu

BrittanyIRL (approved):

packages/story-editor/src/components/panels/design/layer/layerPanel.js
```diff
@@ -57,8 +58,10 @@ function LayerPanel() {
   } = useInspector();
 
   // We want the max height to fill the space underneath the document/design tab bar.
-  // Since the document/design tab bar is 50px subtract that from the full height of the inspector.
-  const maxHeight = inspectorContentHeight - 50;
+  // Since the document/design tab bar has top and bottom margin in addition to a static
+  // height subtract sum from the full height of the inspector.
+  const tabHeight = 2 * TAB_VERTICAL_MARGIN + TAB_HEIGHT;
+  const maxHeight = inspectorContentHeight - tabHeight + 2;
```
samwhale, line 64: nit: could memoize this since these are all static (unless the user resizes the window). We don't save much by memoizing this though, it's only a couple calculations.

BrookeGraham: I think memoizing in this case would end up being more expensive than the simple math 🧐

samwhale: sounds good!

PR #9594, opened 2021-11-01
Convert Types PHP trait to class

### LINKED ISSUE #7844: Convert Types PHP trait to class
## Feature Description

The Types trait is used in 4 classes. It contains a number of methods, used for managing file types. Now that the service container model is in place, we can use injected classes in more places. Converting this trait to a class, should be pretty simple. We can add some unit tests at the same time. 

## Additional Context

---

spacedmonkey:
## Summary

Convert Types PHP trait to class

## Testing Instructions

1.

## Checklist

---

Fixes #7844


includes/Types.php
```diff
@@ -24,14 +24,14 @@
  * limitations under the License.
  */
 
-namespace Google\Web_Stories\Traits;
+namespace Google\Web_Stories;
```
spacedmonkey: Should this class live in the media namespace?

swissspidy: Makes sense to me.

Makes it clearer that it's about "media types" and not other types of "types".

spacedmonkey: Done!

tests/phpunit/integration/tests/Media/Types.php
```diff
@@ -62,13 +78,22 @@ public function test_get_allowed_mime_types() {
...
 	 */
 	public function test_get_allowed_image_mime_types() {
-		$actual   = $this->instance->get_allowed_image_mime_types();
-		$expected = [
-			'image/webp',
-			'image/png',
-			'image/jpeg',
-			'image/gif',
-		];
+		if ( version_compare( get_bloginfo( 'version' ), '5.8.0', '<' ) ) {
```
swissspidy: Why are there now these version checks needed to accommodate `image/webp`?

What if some is using WP < 5.8 and uses a plugin to add `image/webp` support?

Can't we detect support this another way?

I'd really, really like to avoid WP version checks and prefer feature detection over this.

spacedmonkey: Changed check.

swissspidy (approved):

PR #9596, opened 2021-11-01
Decoupling: Remove usage of `wp:*` from taxonomy added in `story-editor`

mohdsayed:
## Context

Remove usage of `wp:*` from taxonomy added in `story-editor`

## Testing Instructions

1.

## Checklist

---


swissspidy (approved):

packages/wp-story-editor/src/api/taxonomy.js
```diff
@@ -43,6 +43,8 @@ export async function getTaxonomies(config) {
   });
 
   return Object.values(result).map((taxonomy) => {
+    taxonomy.rest_path = taxonomy['_links']?.['wp:items']?.[0]?.href;
```
swissspidy, line 46: Using camel case because `snakeCaseToCamelCase` below will convert it anyway?
Using `restPath` already here doesn't work?

And are we actually removing `_links` anywhere?

mohdsayed: Actually, `snakeCaseToCamelCase` converts `restPath` to `restpath`, which is a bug.

Not removing `_links` but we can `delete`  `_links` here.

PR #9599, opened 2021-11-01
Dashboard: 4:6 ratio for story grid item display

### LINKED ISSUE #9583: Dashboard: Update and align aspect ratio for displaying stories and templates
Right now the poster images are getting displayed in 3:4 ratio, let's update it to be 4:6 to be inline with how stories are presented in search -  cropping (in css) the images a bit to get to this is expected.

BrittanyIRL:
## Context

Pascal keenly pointed out last week that the dashboard's displaying stories in 3:4 ratio while the templates got updated to 4:6, they should be consistent. 

## Summary

update grid poster size to 4:6 instead of 3:4

## Relevant Technical Choices

Just updating the ratio and adjusting the object size to fill to maintain ratio in a way that's consistent with other displays of stories (like in Google Search (Mobile)). 
Realized here that 4:6 and 2:3 are the same (_math_ 💥 ) so we can finally gut the other heights we were using in the calculation of story heights for the grid back from when we were previewing the whole story and needed both the fullbleed and the container. Anyway, that's nice to get rid of - most of this PR is just cleaning out the `containerHeight` and the `posterHeight`.

## To-do

- [x] Just realized that this update brings us full circle back to the original PAGE_RATIO (2:3), so i'm going to clean up the dashboard's page size calculation code a bit since we don't need to keep a separate posterHeight as part of the pageSize anymore. 🎉 

## User-facing changes

| Main (3:4) | This Branch (4:6) | 
| --- | --- | 
| Screen Shot 2021-11-01 at 9 50 28 AM | Screen Shot 2021-11-01 at 9 47 37 AM |

## Testing Instructions

See that there's an updated ratio for grid items on the dashboard

1.

## Checklist

---

Fixes #9583

### DIFF
--- a/packages/dashboard/src/components/cardGrid/index.js
+++ b/packages/dashboard/src/components/cardGrid/index.js
@@ -51,7 +51,7 @@ DashboardGrid.propTypes = {
 };
 
 const CardGrid = forwardRef(function CardGrid(
-  { ariaLabel, children, pageSize, isPosterHeight },
+  { ariaLabel, children, pageSize },
   ref
 ) {
   return (
@@ -65,9 +65,7 @@ const CardGrid = forwardRef(function CardGrid(
       tabIndex={0}
       aria-label={ariaLabel}
       columnWidth={pageSize.width}
-      columnHeight={
-        isPosterHeight ? pageSize.posterHeight : pageSize.containerHeight
-      }
+      columnHeight={pageSize.height}
     >
       {children}
     </DashboardGrid>
@@ -78,7 +76,6 @@ CardGrid.propTypes = {
   ariaLabel: PropTypes.string,
   children: PropTypes.node.isRequired,
   pageSize: PageSizePropType.isRequired,
-  isPosterHeight: PropTypes.bool,
 };
 
 export default CardGrid;
--- a/packages/dashboard/src/utils/usePagePreviewSize.js
+++ b/packages/dashboard/src/utils/usePagePreviewSize.js
@@ -57,7 +57,7 @@ import {
 export const getPagePreviewHeights = (width) => {
   const fullBleedHeight = Math.round((width / FULLBLEED_RATIO) * 100) / 100;
   const storyHeight = Math.round((width / PAGE_RATIO) * 100) / 100;
-  const posterHeight = Math.round(((width / (3 / 4)) * 100) / 100);
+  const posterHeight = Math.round(((width / (4 / 6)) * 100) / 100);
   return { fullBleedHeight, storyHeight, posterHeight };
 };
 
--- a/packages/dashboard/src/utils/usePagePreviewSize.js
+++ b/packages/dashboard/src/utils/usePagePreviewSize.js
@@ -46,8 +46,7 @@ import {
  * @param {number} width  width of page to base ratios on
  * @return {Object}       heights to use in pagePreviews { fullBleedHeight: Number, storyHeight: Number}
  */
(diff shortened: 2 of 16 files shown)


packages/dashboard/src/components/cardGrid/index.js
```diff
@@ -51,7 +51,7 @@ DashboardGrid.propTypes = {
 };
 
 const CardGrid = forwardRef(function CardGrid(
-  { ariaLabel, children, pageSize, isPosterHeight },
```
samwhale, line 54: 💭 If we're not using this anymore then it should be removed from the following files too:
- `packages/dashboard/src/app/views/myStories/content/storyGridView/index.js`
- `packages/dashboard/src/app/views/exploreTemplates/content/templateGridView.js`

packages/dashboard/src/utils/usePagePreviewSize.js
```diff
@@ -40,26 +40,14 @@ import {
...
- * This height is used anywhere we need the height of the container holding a pagePreview.
- * It is the true boundary for overflow.
- * It is the 'fullBleedHeight'
- * 2. storyHeight - The height for the actual story (in the dashboard the only actual stories rendered are the templates), used in our shared components with edit-story
- * This means things like the unitsProvider and displayElements that we import to the Dashboard from the editor
- * It's maintaining a 2:3 aspect ratio.
- * When fullbleed is visible (as it is for templates) we use the 2:3 aspect ratio w/ overflow to allow the fullBleed height to be visible
- * 3. posterHeight - This is a height based on 3:4 ratio which is what poster images need, this is used on 'Dashboard' view.
+ * Here we need to calculate height for every pagePreview in use.
+ * The height in 2:3 ratio which is consistent with poster sizes that serve as cover images
```
samwhale, line 44: 👍

samwhale (approved):
Nice! Just one small cleanup thing

packages/dashboard/src/utils/usePagePreviewSize.js
```diff
@@ -40,26 +40,14 @@ import {
...
  * @return {Object}       heights to use in pagePreviews { fullBleedHeight: Number, storyHeight: Number}
  */
-export const getPagePreviewHeights = (width) => {
-  const fullBleedHeight = Math.round((width / FULLBLEED_RATIO) * 100) / 100;
-  const storyHeight = Math.round((width / PAGE_RATIO) * 100) / 100;
-  const posterHeight = Math.round(((width / (3 / 4)) * 100) / 100);
-  return { fullBleedHeight, storyHeight, posterHeight };
-};
+export const getPosterHeight = (width) =>
+  Math.round((width / PAGE_RATIO) * 100) / 100;
```
BrookeGraham: I feel like there is going to be an obvious reason for this, but why do we first multiple by 100 only to then divide by 100?

BrookeGraham (approved):

packages/dashboard/src/components/cardGrid/index.js
BrittanyIRL: thanks, overlooked those!

packages/dashboard/src/utils/usePagePreviewSize.js
BrittanyIRL: i think at one point there was because this was more complicated, but now there's no reason for it. thank you for the fresh eyes on this, i totally looked past it!

maxyinger (approved):

Issue #9070, opened 2021-09-20
Decoupling: Assemble document sidebar in `wp-story-editor`


swissspidy:
## Feature Description

As suggested at https://github.com/google/web-stories-wp/pull/9027#issuecomment-922682077:

What if we export all panels from Document and assemble them in `wp-story-editor` like below

```js
import {
  PublishPanel,
  ExcerptPanel,
  SlugPanel,
  StatusPanel,
  PageAdvancementPanel,
  BackgroundAudioPanel,
  TaxonomiesPanel,
} from '@web-stories-wp/story-editor';

const Document = () => {
  function DocumentInspector() {
    const enabledTaxonomies = useFeature('enableTaxonomiesSupport');
    return (
      <>
        <StatusPanel />
        <PublishPanel />
        <ExcerptPanel />
        <SlugPanel />
        <PageAdvancementPanel />
        <BackgroundAudioPanel />
        {enabledTaxonomies ? <TaxonomiesPanel /> : null}
      </>
    );
  }
}

function Layout() {
  return (
    <MetaBoxesProvider>
      <InterfaceSkeleton
        header={<Header />}
        inspector={ {
            document: {
              component: <Document />,
            },
          }
        }
      >
        <MetaBoxes />
      </InterfaceSkeleton>
    </MetaBoxesProvider>
  );
}

```

That will help us in a couple of ways

1. We would have more control over individual components like `PublishPanel`, so we can easily pass more configuration options ( props ) to handle `isSaving` or passing labels, etc.
2. It would be possible to remove/add panels from a third-party integration.

Just a thought that may be later the entire inspector panel may be configured from the third party integration using the following. 

```js
      <InterfaceSkeleton
        header={<Header />}
        inspector={ {
          design: {
            title: __( 'Design', 'web-stories' ),
            component: <Design />
          },
          document: {
            title: __( 'Document', 'web-stories' ),
            component: <Document />,
          },
          ...
        } }
      >
```

## Additional Context

Gutenberg does this in a similar way, see the `SettingsSidebar` component: https://github.com/WordPress/gutenberg/blob/164bef81386e96330128c6218a66f29c9ed8e865/packages/edit-post/src/components/sidebar/settings-sidebar/index.js

mohdsayed:
[...]

PR #9502, opened 2021-10-22
Support password protected and pending stories 

spacedmonkey:
## Context

The story editor does not currently support password protected stories or submitting stories for review (relevant for the contributor user role)

## Summary

* Adds support for password protected stories.
* Add unit, e2e and JS unit tests.
* Adds support for submitting stories for review.
* Hides the publish time setting if user cannot publish and needs to submit for review insrtead.

## Relevant Technical Choices

* Refactor the check for `is_singular( Story_Post_Type::POST_TYPE_SLUG )` to include checks for `is_embed()` and password requirement. 
* Refactor traits to use new context class to be injected everywhere

## User-facing changes

**Administrator**

Default state (public, draft):

Screenshot 2021-10-28 at 19 37 27

Draft, password protected:

Screenshot 2021-10-28 at 19 37 33

Published, password protected:

Screenshot 2021-10-28 at 19 37 42

Message that shows up when selecting "Private":

Screenshot 2021-10-28 at 19 37 48

**Contributor**

Screenshot 2021-10-28 at 19 37 14

## Testing Instructions

- [ ] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

**Contributor User**

1. Create new story
2. Verify that the user **cannot** choose between different visibility options
3. Verify that user cannot publish a story, but only submit for review
4. No "review checklist before publishing" dialog appears when doing so
5. Also verify that the publish time cannot be chosen

**Admin / Editor / Author**

1. Create new story
2. Verify that the user **can** choose between different visibility options
3. Verify that user can publish a story, and a review dialog is shown
4. Verify that selecting the "private" visibility causes a confirmation for immediately privately publishing
4. A "review checklist before publishing" dialog appears when privately publishing (if applicable)
5. Publish time can be chosen

### Does this PR have a security-related impact?

No

### Does this PR change what data or activity we track or use?

No

### Does this PR have a legal-related impact?

No

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [x] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
[...]

PR #9614, opened 2021-11-02
Decoupling: Assemble document sidebar in `wp-story-editor`

mohdsayed:
## Context

https://github.com/google/web-stories-wp/issues/9070

## Testing Instructions

1.

## Checklist

---

Partially addresses https://github.com/google/web-stories-wp/issues/9070


swissspidy:
Curious, what's the reason for ff303036886e4fe889f49b59bc2731d11973722e?

mohdsayed:
> Curious, what's the reason for [ff30303](https://github.com/google/web-stories-wp/commit/ff303036886e4fe889f49b59bc2731d11973722e)?

I tried to pass an array first so that more tabs can be sent from the configuration but I realized that there would be a couple of issues

1. The core package is not ready for this change, it would need some refactoring.
Currently, we need to provide `refs` for each tab here

https://github.com/google/web-stories-wp/blob/07d041dde2d8f26e3cb63f0822186d16d20512a8/packages/story-editor/src/components/inspector/inspectorProvider.js#L75-L81

We cannot do that dynamically, because of the limitation that a react hook cannot be used inside a loop or conditionally, and if we provide `useRef` from the `inspectorTabs` configuration, it's not making sense.

2. We cannot allow unlimited tabs because there is not enough space for them in the sidebar, and perhaps would not even be needed.

If there is a need to create one more tab later, we may provide that option by adding one spare `additional` tab like this. 

```js
<InterfaceSkeleton
  header={<Header />}
  inspectorTabs={{
    document: {
      title: __('Document', 'web-stories'),
      Pane: DocumentPane,
    },
    additional: {
      title: __('Advert', 'web-stories'),
      Pane: AdvertPane,
    }
  }}
>
```

mohdsayed:
Circling back to the comment in https://github.com/google/web-stories-wp/pull/9569#discussion_r738479055

> Aside: #9070 will reduce this list a bit as we move more stuff out of the core app to wp-story-editor

Looks like only these two are directly being used inside the document panels, ( there are a couple more but they are reused at other places in the core package too ) which we can move.

https://github.com/google/web-stories-wp/blob/07d041dde2d8f26e3cb63f0822186d16d20512a8/packages/story-editor/src/components/panels/document/publish/publish.js#L114-L116

packages/wp-story-editor/src/components/documentPane/index.js
```diff
@@ -13,5 +13,31 @@
...
+  StatusPanel,
+  PageAdvancementPanel,
+  BackgroundAudioPanel,
+  TaxonomiesPanel,
+} from '@web-stories-wp/story-editor';
+
+function DocumentPane() {
+  return (
+    <>
+      <StatusPanel />
```
swissspidy, line 36: We should move this component to `wp-story-editor`

It already contains lots of WP specific references (e.g. "Visible to site admins & editors only").
Plus, #9502 is going to update it to accommodate for WP specific quirks (e.g. WP does not supports private drafts, so private means instantly publishing the story).

packages/wp-story-editor/src/components/documentPane/index.js
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Internal dependencies
+ */
+import * as publisherLogoCallbacks from '../../api/publisherLogos';
+
+function DocumentPane() {
+  return (
+    <>
+      <StatusPanel />
+      <PublishPanel apiCallbacks={publisherLogoCallbacks} />
```
swissspidy: Would it be easier if we'd move this component to `wp-story-editor` as well? Then we don't have to pass callbacks like this.

packages/wp-story-editor/src/components/documentPane/index.js
mohdsayed: Yes, makes sense 👍

packages/wp-story-editor/src/components/documentPane/index.js
mohdsayed: yeah, on it.

mohdsayed:
Just for info, I removed the integration tests which are no longer valid in the core editor in https://github.com/google/web-stories-wp/pull/9614/commits/286fb1cad87139c21a6d17bdf0fbfe0798989cfd . Will create another PR to compensate for it in E2E tests.

mohdsayed:
Looks like E2E won't pass here 🙁  . Will have to skip a couple of tests that are failing on `main` branch also for me locally.

Screenshot 2021-11-08 at 2 27 40 PM
Screenshot 2021-11-08 at 2 28 51 PM

swissspidy:
I don't see them failing in this PR? Tests also seem to be passing on `main` here on GitHub.

Could you perhaps elaborate where these are failing?

mohdsayed:
Passed after merging latest `main` today!  😺

swissspidy (approved):

Issue #9463, opened 2021-10-20
UI Simplifications: Saved Styles


miina:
## Feature Description
See [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor-Sprint-Feature-Level-Designs_?node-id=10250%3A435820)

- [ ] Move Saved styles to the Text Panel using a subheading.
- [ ] Create a popup for displaying Saved Styles (behaves the same way as the Color Picker)
- [ ] Only display one row of styles in the panel, and use the popup for the rest of the styles when clicking on "More styles"
- [ ] Adjust the style when there are no Saved Styles
- [ ] Ensure the Edit mode (now only inside the popup) and Adding style work as previously.

<img width="343" alt="Screenshot 2021-10-20 at 12 28 01" src="https://user-images.githubusercontent.com/3294597/138067168-3b9afcf6-32c2-4402-8f4d-db0790404155.png">

csossi:
Verified in QA

PR #9616, opened 2021-11-02
UI Simplification: Colors

### LINKED ISSUE #9464: UI Simplifications: Colors
## Feature Description
See [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor-Sprint-Feature-Level-Designs_?node-id=10250%3A437034)

- [ ] Change "Style" panel to "Color" for Shape element
- [ ] Remove Saved colors panel from Design Panel
- [ ] Change "Story colors" to "Current story" and reorder the sections
- [ ] Allow saving colors directly from the "Basic" view of the Color Picker by clicking the `+` icon.
- [ ] Add Edit mode to the Color Picker for Current/Saved Colors. When clicking on a Default color in edit mode, it should work as usually + exit the edit mode.
- [ ] Change "Add" in the Custom view to "Add to Saved Colors"
- [ ] Add Eyedropper to the Basic view.

miina:
## Summary
Moves the Saved colors panel functionality to the Color Picker.

## To-do

- [x] Make the confirmation dialogue work
- [x] Refactoring
- [x] Tests

## User-facing changes
| Before | After |
|-------|------|
| Screenshot 2021-11-04 at 11 42 36 | Screenshot 2021-11-04 at 11 34 34 |
| Screenshot 2021-11-04 at 11 43 37 | Screenshot 2021-11-04 at 11 39 15 |
| Screenshot 2021-11-04 at 11 44 45 | Screenshot 2021-11-04 at 11 38 17 |
| Screenshot 2021-11-04 at 11 45 34 | Screenshot 2021-11-04 at 11 37 55 |

## Testing Instructions

1. Add a shape element and very that the "Style" panel is now called the "Color" panel instead.
2. Verify there is no Saved colors panel anymore.
3. Open the Color picker and verify the Basic color picker view now includes "Current story", "Saved Story" sections.
4. Verify that new colors can be added from the Color Picker
5. Verify that the edit mode of the Color Picker works as expected and allows removing colors. Note that "Default colors" can not be deleted.
6. Verify that when clicking on a Default color while in Edit mode, the Edit mode exits and the color is applied to the element.
7. Verify the Eyedropper (Sample color) in the Basic view works as expected.
8. Go to the Custom view of the color picker, verify the button "Add" now says "Add to Saved Colors"

No

No

No

## Checklist

---

Fixes #9464


packages/story-editor/src/components/colorPicker/basicColorPicker.js
```diff
@@ -138,6 +242,16 @@ function BasicColorPicker({
           <StyledPlus />
         </StyledButton>
       </Body>
+      {showDialog && (
+        <ConfirmationDialog
```
miina, line 235: @barklund I moved this over here currently but it's not really working -- the confirmation dialogue and the color picker popup can not be open at the same time. Wondering what would be a good approach for this.

One option could be moving the `ConfirmationDialog` to the `Color` input directly, to the same component that includes the `ColorPicker`, closing the color picker for the time of opening the confirmation dialogue, and once the user has interacted with the dialogue, reopening the color picker again. This could potentially tangle the Color input and the ColorPicker quite a bit. But maybe it's ok.

Not sure if there's a simpler alternative. Let me know if you have any alternative thoughts.

packages/story-editor/src/components/panels/design/preset/stories/confirmationDialog.js
```diff
@@ -24,7 +24,7 @@ import { boolean } from '@storybook/addon-knobs';
  * Internal dependencies
  */
 import ConfirmationDialog from '../confirmationDialog';
```
miina, line 26: Note: More refactoring can be done here for the style presets, this will be part of #9463

packages/story-editor/src/components/colorPicker/basicColorPicker.js
```diff
@@ -28,32 +28,48 @@ import {
...
 const Body = styled.div`
   display: flex;
   flex-direction: column;
   margin: 8px 16px 16px;
-  gap: 16px;
+`;
+
+const SavedColors = styled.div`
+  border-bottom: 1px solid ${({ theme }) => theme.colors.border.defaultNormal};
+  padding-bottom: 24px;
```
miina, line 56: FYI when reviewing: There were some pixel inconsistencies in the Color Picker in Figma (e.g. partial pixels, having 23px in one place for padding and 24px in other place, etc.) -- I chose what made the most sense for consistency.

packages/story-editor/src/components/colorPicker/basicColorList.js
```diff
@@ -107,27 +136,34 @@ function BasicColorList({
...
-          />
+          <Tooltip key={patternAsBackground} title={tooltip}>
+            <StyledSwatch
+              onClick={() => handleClick(pattern, isLocal)}
+              pattern={pattern}
+              isSelected={isSelected}
+              isDisabled={isDisabled}
+              tabIndex={tabIndex}
+              title={patternAsBackground}
+              aria-label={isEditMode ? deleteLabel : applyLabel}
```
miina: @barklund The label here was an accessible name before (e.g. just `#fff`). Now it supports deleting as well though. Any thoughts on what would be a better label now? `Apply local color: #fff` etc.?

Trapsta (approved):
Tested and works great. Minor thing but the height of the  header for the color picker seems a bit more than the design. 52px on the browser but 43px on figma.
Screenshot 2021-11-05 at 05 50 29

Also, could not find this in the designs but the popup to confirm deleting a color is in light mode, not sure if that's how it's supposed to look or if it's a bug on loading the theme.

miina:
> Minor thing but the height of the header for the color picker seems a bit more than the design. 52px on the browser but 43px on figma.

Thanks! Updated to 44px. There are some incorrect values in Figma for the Color Picker currently, e.g. partial pixels (7.2 instead of 8 etc.), but 52px was definitely incorrect as well.

> Also, could not find this in the designs but the popup to confirm deleting a color is in light mode, not sure if that's how it's supposed to look or if it's a bug on loading the theme.

Yep, that being in light mode is correct, thanks for checking.

packages/story-editor/src/components/colorPicker/header.js
```diff
@@ -29,46 +29,97 @@ import {
...
-        <Icons.Cross />
-      </CloseButton>
+      <Actions>
+        {hasPresets && isEditMode && (
+          <StyledButton
+            {...buttonProps}
+            aria-label={__('Exit edit mode', 'web-stories')}
+          >
+            {__('Done', 'web-stories')}
+          </StyledButton>
```
barklund: This button is a bit tight (visible in the hover state mostly) - perhaps the button type should change when the wording is Done? Or we could change it to an icon - we have the floppy-disk save icon already, which might work?

Screen Shot 2021-11-05 at 11 58 57

packages/story-editor/src/components/colorPicker/basicColorPicker.js
```diff
@@ -76,58 +100,134 @@ function BasicColorPicker({
...
+            {__('Default', 'web-stories')}
+          </Label>
+          <BasicColorList
+            color={color}
+            colors={BASIC_COLORS}
+            handleClick={(pattern) => {
+              handleColorChange(pattern);
+              if (isEditMode) {
+                setIsEditMode(false);
+              }
```
barklund: Nit: If `isEditMode` is already false, setting it to false again is ignored (because it's a primitive), so for ease of implementation this `if` could be removed and you could just always set it to false here.

packages/story-editor/src/components/colorPicker/basicColorPicker.js
```diff
@@ -76,58 +100,134 @@ function BasicColorPicker({
...
+                colorType="local"
+                allowsOpacity={allowsOpacity}
+                allowsGradient={allowsGradient}
+                aria-labelledby="colorpicker-story-colors-title"
+                isEditMode={isEditMode}
+              />
+            </>
+          )}
+          {allowsSavedColors && (
+            <>
```
barklund: Couldn't these four lines be removed? It's the same condition guarding the contents, no?

packages/story-editor/src/components/colorPicker/basicColorPicker.js
```diff
@@ -76,58 +100,134 @@ function BasicColorPicker({
...
+      return;
+    }
+    // If deleting a local color, delete without confirmation.
+    if (isLocal) {
+      deleteLocalColor(preset);
+      return;
+    }
+
+    // If the user has dismissed the confirmation dialogue previously.
+    const storageKey = 'DELETE_COLOR_PRESET_DIALOG_DISMISSED';
```
barklund: Is this defined elsewhere as well? Shouldn't it be a `const` somewhere then?

packages/story-editor/src/components/colorPicker/colorPicker.js
```diff
@@ -113,15 +114,15 @@ function ColorPicker({
     [onDebouncedChange, selectedElementIds, changedStyle, pushTransform]
   );
 
-  const closeIfNotEyedropping = () => {
-    if (!isEyedropperActive) {
+  const maybeClose = () => {
+    if (!isEyedropperActive && !showDialog) {
       onClose();
     }
   };
```
barklund: I think this deserves a comment or two - it looks pretty simple, but the reasoning can be a bit deceptive.

packages/story-editor/src/components/colorPicker/confirmationDialog.js
```diff
@@ -0,0 +1,96 @@
...
+      isOpen
+      onClose={onClose}
+      secondaryText={__('Cancel', 'web-stories')}
+      onPrimary={onPrimary}
+      primaryText={__('Delete', 'web-stories')}
+    >
+      <Text size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL}>
+        {__(
+          'This is a global color. Deleting this color will remove it from the Saved Colors panel across all stories and the color will no longer be available to any other users on the site.',
+          'web-stories'
```
barklund: This dialog references the Saved Colors panel in the wording - perhaps it should just say color picker instead now?

Screen Shot 2021-11-05 at 11 59 59

packages/story-editor/src/components/colorPicker/header.js
```diff
@@ -29,46 +29,97 @@ import {
...
+        )}
+        {hasPresets && !isEditMode && (
+          <StyledButton
+            {...buttonProps}
+            aria-label={__('Edit colors', 'web-stories')}
+          >
+            <Icons.Pencil />
+          </StyledButton>
+        )}
```
barklund: Could you move the hasPresets check a bit out? Then it would be a simple ternary conditional on the label and body:

```js
        {hasPresets && (
          <StyledButton
            {...buttonProps}
            aria-label={isEditMode ? __('Exit edit mode', 'web-stories') : __('Edit colors', 'web-stories')}
          >
            {isEditMode ? __('Done', 'web-stories') : <Icons.Pencil />
          </StyledButton>
        )}
```

This would also fix the a11y of the button losing focus when pressed using keyboard currently.

packages/story-editor/src/components/colorPicker/confirmationDialog.js
```diff
@@ -0,0 +1,96 @@
...
+  display: flex;
+  margin-top: 12px;
+  padding: 14px 0 0 9px;
+  border-top: 1px solid ${({ theme }) => theme.colors.divider.primary};
+`;
+
+function ConfirmationDialog({ onClose, onPrimary }) {
+  const storageKey = 'DELETE_COLOR_PRESET_DIALOG_DISMISSED';
+  const [isDialogDismissed, setIsDialogDismissed] = useState(
+    localStore.getItemByKey(LOCAL_STORAGE_PREFIX[storageKey])
```
barklund: This init check can't be necessary? This dialog would never appear if the flag was true, so it must be false, no?

packages/story-editor/src/components/colorPicker/useDeleteColor.js
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * External dependencies
+ */
+import { useCallback } from '@web-stories-wp/react';
+/**
+ * Internal dependencies
+ */
+import { useStory } from '../../app';
+
+function useDeleteColor({ setIsEditMode }) {
```
barklund: I feel this would be a bit cleaner and independent if this callback was just `onEmpty` that would be invoked if there were no more saved colors. Then it isn't mixed up with the notion of edit mode, which is otherwise irrelevant for this hook. And it would just be from the outside:

```js
useDeleteColor({ onEmpty: () => setIsEditMode(false) });
```

(but probably memoized)

packages/story-editor/src/components/colorPicker/basicColorList.js
```diff
@@ -88,15 +93,35 @@ function BasicColorList({
     .map(getPatternAsString)
     .findIndex((c) => colorAsBackground === c);
 
+  const isLocal = 'local' === colorType;
```
barklund: Why is it that we use a string `type` and not just bool throughout? It seems overly complicated because we keep converting the string back to a bool as that makes more sense in many places.

barklund (commented):
Generally looks good and works great.

I didn't go through everything, because there really is a lot in this one, but here's at least a bunch of comments 😄

packages/story-editor/src/components/colorPicker/header.js
miina: This is something that was brought directly over from the panel since the design for that state was missing in Figma. I'll follow up with Amy about it but leave it as is (and as was) for now.

packages/story-editor/src/components/colorPicker/confirmationDialog.js
miina: Hm, it shouldn't appear if it was true though.

PR #9617, opened 2021-11-02
UI Simplifications: Border Panel

### LINKED ISSUE #9461: UI Simplifications: Border Panel
## Feature Description
See [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor-Sprint-Feature-Level-Designs_?node-id=10250%3A436799).

- [ ] Update the design based on Figma for locked and unlocked status
- [ ] Add Eyedropper next to the color input
- [ ] Update the opacity icon (it was adjusted a bit since initially adding it)

Screenshot 2021-10-20 at 12 17 42

Trapsta:
## Summary

Updates and simplifies the border width inputs design for locked and unlocked status

## To-do

Changes for the border color eyedropper  are addressed on a different PR

## User-facing changes

| Before | After |
| ------ | ----- |
| Screenshot 2021-10-26 at 17 27 41 | Screenshot 2021-10-26 at 17 26 08 |
| Screenshot 2021-10-26 at 17 27 50 | Screenshot 2021-10-26 at 17 25 53 |

## Testing Instructions

- Verify that the border width inputs do not have labels at the bottom when unlocked.
- Verify that the border width inputs do not have a separator gap in gutters between each input.
- Verify that each border width input has a corresponding suffix icon denoting the border side it controls as per [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor-Sprint-Feature-Level-Designs_?node-id=10250%3A436799) .
- Verify the border panel matches the design  on [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor-Sprint-Feature-Level-Designs_?node-id=10250%3A436799)

 
- Verify the border width still works as expected.

No

No

No

## Checklist

---

Fixes #9461


swissspidy:
Nit: only toggle the checkbox when actually true :-) 
Screenshot 2021-11-02 at 17 37 20
Screenshot 2021-11-02 at 17 37 17

packages/story-editor/src/components/form/stackedInputs/index.js
```diff
@@ -0,0 +1,193 @@
...
+
+function StackedInputs(props) {
+  const {
+    lockInput,
+    handleChange,
+    inputProps,
+    suffix,
+    firstInputLabel,
+    showLockedSuffixIcon,
+  } = props;
```
barklund: My idea about this component (or set of components) was a bit different actually.

I'd like to see it be more generic (why specific to four inputs? Why look at those specific labels `topLeft` and `left`? etc) and generalized in one of two ways.

It could either be made as a wrapper with inputs inside like a two-component utility (we use a similar method for `<PillGroup />` and `<Pill />` in the design system):

```js
<StackableGroup>
  <StackableInput
    value=...
    suffix=...
    ...
  />
</StackableGroup>
```

Or it could be achieved using a single component, but with the option of providing an array of inputs.

```js
const inputs = [
  {
    key: 'left',
    value: border.left,
    label: _('Left border', 'web-stories'),
    suffix: <BorderLeft />,
    ...
  },
  ...
];
... 
<StackedInputs
  inputs={inputs}
  onChange=...
  ...
/>
```

I personally prefer the first method, as it seems like a more neat API inside the component using the stacked inputs, but it might be a bit trickier.

The only things that should go inside the new component(s) are basically the look and feel of the inputs and their wrapper and nothing about the logic, naming, etc at all.

In either case, all the stuff about locking/unlocking, the keys, names, and labels of the attributes, etc should all be moved outside the component and it should be possible to reuse this for a completely different set of stacked inputs with a different number of inputs, different labels and names, with and without icons, etc.

For both methods, the difference between locked and unlocked inputs is just about how many child elements to render - either by using conditional rendering (as before) or simply adding more elements to the array in the second instance.

packages/story-editor/src/components/form/stackedInputs/index.js
```diff
@@ -0,0 +1,193 @@
...
+
+const styleOverrideBottomLeft = css`
+  ${inputContainerStyle}
+  border-radius: 0;
+`;
+
+const styleOverrideBottomRight = css`
+  ${inputContainerStyle}
+  border-radius: 0 4px 4px 0;
+`;
```
barklund: This bit can be generalized using `:first-of-type` and `last-of-type` so it works for any number of sibling nodes.

packages/story-editor/src/components/form/stackedInputs/index.js
```diff
@@ -0,0 +1,193 @@
...
+     ${
+       side === 'topRight' || side === 'right'
+         ? 'transform: rotate(180deg);'
+         : ''
+     }
+     ${
+       side === 'bottomRight' || side === 'bottom'
+         ? 'transform: rotate(270deg);'
+         : ''
+     }
```
barklund: This is terribly specific for these two instances and should not go in here at all.

packages/story-editor/src/components/form/stackedInputs/utils.js
```diff
@@ -0,0 +1,54 @@
...
+    case 'right':
+      return __('Right border', 'web-stories');
+
+    case 'bottom':
+      return __('Bottom border', 'web-stories');
+
+    default:
+      return key;
+  }
+};
```
barklund: This is also pretty messy to maintain. These labels should be specified in the components using the stacked inputs, not inside the stacked inputs themselves.

packages/story-editor/src/components/panels/design/sizePosition/radius.js
```diff
@@ -185,87 +121,14 @@ function RadiusControls({ selectedElements, pushUpdateForObject }) {
...
-        )}
-      </InputContainer>
+      <StackedInputs
+        lockInput={lockRadius}
+        inputProps={borderRadius}
+        handleChange={handleChange}
+        suffix={<Icons.Corner />}
+        firstInputLabel={firstInputLabel}
+        showLockedSuffixIcon
+      />
```
barklund: I do see how this cleans up this component a lot and I do appreciate that. But I also think it creates a new component super-specific to only these two instances, that can't be used anywhere else ever again, and thus isn't that relevant in the codebase.

barklund: There might be room for a `<LockableStackableInputs />` component somewhere though, that puts the lock inside the components and you could then use conditional rendering of the children object to only render the first input rather than all children in case of locked/unlocked inputs - this might even be reusable for the text box padding inputs actually. But that's a rather specific component, that is a bit much work I think.

barklund (commented):
This is a good effort, but I think the new component is too specific for the current use cases and doesn't leave any real room for reuse. I also feel like important information is now hidden away inside the "general" component, that really should be controlled by the panel (e.g. labels and icon rotations).

packages/story-editor/src/components/panels/design/sizePosition/radius.js
miina: Agreed with Morten here that the new component is currently specific only to these two use cases and doesn't make much sense as a generic form component. Any generic form component should support new use cases and only include the shared parts of the input(s). Labels, icons, etc. would all come from the specific use case and be outside of the generic form component.

Trapsta:
@barklund @miina thanks for the feedback,  I have made the component more generic. Went with the first suggestion from Morten by changing the component to a wrapper as this was also my initial idea but shelved it as I thought it would require refactoring the inputs components but turns out that's not really necessary and seems to work quite well

miina:
@Trapsta From testing:

Works well! One detail -- currently the line between two inputs is a double line (1px line from both inputs), however, it should be just 1px line -- moving all but the first input to the left by 1 px should fix it.

packages/story-editor/src/components/form/stackable/index.js
```diff
@@ -0,0 +1,64 @@
...
+      }
+      &:last-of-type > div {
+        border-radius: 0 4px 4px 0;
+      }
+    }
+  `
+);
+
+function StackableGroup(props) {
+  const { children, containerStyleOverride, locked } = props;
```
barklund: We try to always destruct props in the function definition directly:

```js
function StackableGroup({ children, containerStyleOverride, locked }) {
```

packages/story-editor/src/components/form/stackable/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+  return (
+    <StackableContainer locked={locked} styleOverride={containerStyleOverride}>
```
barklund: No need for a style override here - styled-components generates class names as the first choice, so if you only need a single style override, a class name will do:

```js
<StackableContainer locked={locked} className={className}>
...
const StyledStackableGroup = styled(StackableGroup)`...`
...
<StyledStackableGroup>...
```

However, you actually don't even seem to be using this feature. I do think that allowing a class name to be passed is a good convention anyway.

packages/story-editor/src/components/panels/design/border/borderWidth.js
```diff
@@ -87,24 +56,25 @@ const ToggleWrapper = styled.div`
...
-    </>
-  );
-}
+const StackableInput = styled(NumericInput)`
+  & > div {
+    border-radius: 0;
+    padding: 2px 7px 2px 12px;
+    ${inputContainerStyleOverride};
+  }
+`;
```
barklund: I feel like this should be a generalized component as well as its the same between the two panels?

packages/story-editor/src/components/panels/design/sizePosition/radius.js
```diff
@@ -77,45 +73,32 @@ const inputContainerStyleOverride = css`
...
-  ${inputContainerStyleOverride}
-  border-radius: 0;
+const StackableInput = styled(NumericInput)`
+  svg {
+    ${iconCss}
+  }
+  & > div {
+    border-radius: 0;
+    padding: 2px 7px 2px 12px;
+    ${inputContainerStyleOverride};
```
barklund: You need this one in the border panel as well (or actually in the generalized StackableInput as suggested elsewhere) as it includes the `:focus-within { z-index: 1; }` which sends the focused input to the top avoiding the slight clipping of the focus outline.

Compare the focus outlines here, where you see the outline being behind the neighbor for the border, but not for the radius:

| Radius panel | Border panel |
|-|-|
| Screen Shot 2021-11-03 at 14 04 04 | Screen Shot 2021-11-03 at 14 03 52 |

(notice the two little notches in the right side of the focus outline in the second screenshot)

barklund (approved):
This is a lot better. Just a few nits on the composability and a slight styling detail.

packages/story-editor/src/components/form/stackable/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+    & > div {
+      border-radius: 0;
+
+      &:first-of-type > div {
+        border-radius: 4px 0 0 4px;
+      }
+      &:last-of-type > div {
+        border-radius: 0 4px 4px 0;
+      }
```
barklund: I feel like these should go to the generalized StackableInput as well, which makes the CSS faster as CSS rules are evaluated right-to-left.

The right-most part of this rule is "div" matching thousands of elements, so the CSS parser has to go one step further before discarding this rule for almost all elements.

With styled-components we're almost exclusively using single-item CSS selectors for speed reasons, because we can and only multi-item selectors when unavoidable. And when we do use multi-level, we make them as specific as possible. Here you could have made it more specific if you had a reference to that other div and did:

```
& > ${StackableInput} {
```

Styled-components knowns how to compose those and replace the component with its generated class name, which is pretty neat.

packages/story-editor/src/components/panels/design/border/borderWidth.js
```diff
@@ -21,62 +21,27 @@ import PropTypes from 'prop-types';
 import { useCallback } from '@web-stories-wp/react';
```
miina, line 21: @Trapsta I just realized that this PR doesn't add the eyedropper to the Border color input. Just adding `hasEyedropper` here should do it (but test it out just in case):
https://github.com/google/web-stories-wp/blob/306146c63488ee3d20365a980feeebb2c6067558/packages/story-editor/src/components/panels/design/border/border.js#L72-L79

packages/story-editor/src/components/form/stackable/stackableGroup.js
```diff
@@ -0,0 +1,62 @@
...
+    border-radius: 0;
+
+    &:first-of-type > div {
+      border-radius: 4px 0 0 4px;
+      margin-left: 0;
+    }
+
+    &:last-of-type > div {
+      border-radius: 0 4px 4px 0;
+    }
```
barklund: This fails if there's only one element because it would be both first of type and last of type and you overwrite the left radii in the declaration for the latter statement:

Screen Shot 2021-11-05 at 11 52 16

You could add a third rule capturing both, but if you instead changed it to the following, it would be a bit prettier (and also more specific):
```css
    &:first-of-type > div {
      border-top-left-radius: 4px;
      border-bottom-left-radius: 4px;
      margin-left: 0;
    }

    &:last-of-type > div {
      border-top-right-radius: 4px;
      border-bottom-right-radius: 4px;
    }
```

barklund (commented):
Looks great - just one nit.

PR #9630, opened 2021-11-03
Design System: Snackbar Accessibility

### LINKED ISSUE #6061: Snackbar Accessibility Issues
## Feature Description

Our current snackbars are not really accessible, as they are not announced to screen readers, and it's impossible for users to interact with them.

I know that there are general issues with snackbars (bad location, auto-disappearance, etc.), but let's do one step at a time.

## Additional Context

Recommended reading including documents linked to: https://github.com/WordPress/gutenberg/issues/16549

---

## Acceptance Criteria

Make snackbars more accessible

## Implementation Brief

* [ ] Snackbar is announced to screen readers (#6059)
* [ ] If the snackbar contains an action, there is a global shortcut to perform said action

samwhale:
## Context

Our current snackbars are not really accessible, as they are not announced to screen readers, and it's impossible for users to interact with them.

## Summary

- Use `useLiveRegion` to announce the snackbar messages to users
- `dissmissable` was spelled wrong, so that's fixed here too

## Relevant Technical Choices

- Hiding the snackbar from screen readers for now. If we use `useLiveRegion` then the notification is announced twice. 
- Added a new prop: `actionHelpText`. This is text that is only announced via screen reader. It is not rendered to the page.

## To-do

n/a

## User-facing changes

Nothing visual

## Testing Instructions

1. Use a screen reader to test this change. An option in chrome is the [chromevox chrome plugin](https://chrome.google.com/webstore/detail/screen-reader/kgejglhpjiefppelpmljglcjbhoiplfn)
2. Perform an action that shows a snackbar (ex: right click an element to open the context menu. Then click 'copy styles' or something).
3. The snackbar text should have been announced. If there is an action, there is helper text with that action.

no

no

no

## Checklist

---

Fixes #6061


packages/design-system/src/components/snackbar/constants.js
```diff
@@ -45,11 +45,12 @@ export const SnackbarNotificationThumbnail = PropTypes.shape({
...
   actionLabel: PropTypes.oneOfType([PropTypes.node, PropTypes.string]),
-  dismissable: PropTypes.bool,
+  dismissible: PropTypes.bool,
   onAction: PropTypes.func,
   onDismiss: PropTypes.func,
   preventActionDismiss: PropTypes.bool,
   preventAutoDismiss: PropTypes.bool,
   timeout: PropTypes.number,
   thumbnail: SnackbarNotificationThumbnail,
+  key: PropTypes.string,
```
samwhale: this was missing, but used in the component

packages/design-system/src/components/snackbar/snackbarContainer.js
```diff
@@ -151,13 +165,13 @@ export const SnackbarContainer = ({
               <ChildContainer placement={placement}>
                 <Component
                   {...notificationProps}
-                  aria-label={message}
+                  aria-hidden
```
samwhale, line 172: Don't need this to be visible to screen readers now that the message is being announced. The global shortcut will also remove the need to hit the action button in the snackbar with the keyboard.

Note: it was already really tricky (maybe not possible?) to tab into the snackbar

packages/design-system/src/components/snackbar/constants.js
swissspidy: `key` is a special React prop. I don't think that needs to be added to prop types? Or am I missing something?

packages/story-editor/src/app/rightClickMenu/provider.js
```diff
@@ -61,6 +62,13 @@ import rightClickMenuReducer, {
 } from './reducer';
 import { getDefaultPropertiesForType, getElementStyles } from './utils';
 
+const UNDO_HELP_TEXT = sprintf(
+  /* translators: %s: Command or Control key on keyboard */
+  __('Press %s Z to undo the last change', 'web-stories'),
+  getKeyForOS('mod'),
+  'Z'
+);
```
swissspidy: I'd move the whole key combo out of the translatable string and also use `prettifyShortcut`, e.g. something like

```suggestion
const UNDO_HELP_TEXT = sprintf(
  /* translators: %s: Ctrl/Cmd + Z keyboard shortcut */
  __('Press %s to undo the last change', 'web-stories'),
  prettifyShortcut('mod+z')
);
```

Also that second `Z` arg was not doing anything

Could probably

swissspidy: Ditto for the instance in `useQuickActions.js`

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
```diff
@@ -43,7 +43,7 @@ import { getResetProperties } from './utils';
 import { ACTIONS, RESET_PROPERTIES, RESET_DEFAULTS } from './constants';
 
 const UNDO_HELP_TEXT = sprintf(
-  /* translators: %s: Command or Control key on keyboard */
+  /* translators: %s: Command or Control key and the z key on the keyboard */
```
swissspidy: I'd really prefer something like `Ctrl/Cmd + Z keyboard shortcut` over this.

packages/design-system/src/components/snackbar/constants.js
samwhale: That's true - `key` is a special react prop. It's being used as an id here too: https://github.com/google/web-stories-wp/blob/main/packages/design-system/src/components/snackbar/snackbarContainer.js#L146

Probably best to check for an `id` rather than a `key` since `key` doesn't get propagated as a prop, so the line above is meaningless.

packages/design-system/src/components/snackbar/snackbarContainer.js
```diff
@@ -125,13 +127,29 @@ export const SnackbarContainer = ({
     }
   }, [max, notifications, onRemove]);
 
+  // Announce messages to screen reader when a new message shows up
+  useEffect(() => {
+    notifications.forEach((notification) => {
+      if (!announcedNotifications.current.has(notification)) {
+        // speak the message
+        const message = `${notification.message} ${
+          notification.actionHelpText ? notification.actionHelpText : ''
```
BrittanyIRL:
```suggestion
          notification.actionHelpText || ''
```

packages/design-system/src/components/snackbar/stories/index.js
```diff
@@ -27,6 +29,19 @@ import { theme } from '../../..';
 import { DarkThemeProvider } from '../../../storybookUtils/darkThemeProvider';
 import { SnackbarContainer } from '../snackbarContainer';
 import { THUMBNAIL_STATUS } from '../constants';
+import { Button, BUTTON_TYPES } from '../..';
+
+const buttonText = [
+  'Sorry, your image failed to load',
+  'Success! One cow successfully bought',
```
BrittanyIRL, line 36: 🐮

BrittanyIRL (approved):
Looks good to me, thanks for the storybook and tests! Like the idea to update to not use `key`, otherwise think this is great.

packages/design-system/src/components/snackbar/snackbarContainer.js
littlemilkstudio: yea these feel a little weird to tab into

maxyinger (approved):

packages/design-system/src/components/snackbar/snackbarContainer.js
samwhale: 😂 I should've done this in the first place, nice catch

packages/design-system/src/components/snackbar/snackbarMessage.js
```diff
@@ -258,6 +259,7 @@ const SnackbarMessage = ({
             <CloseButton
               onClick={onDismiss}
               aria-label={__('Close', 'web-stories')}
+              tabIndex={-1}
```
BrookeGraham, line 262: 🧑🏻‍🍳💋

BrookeGraham (approved):

Issue #9582, opened 2021-10-29
Update Template CLI Script Viewport Size


BrittanyIRL:
## Feature Description

As pointed out by Pascal, the current template poster CLI script creates images that are 776x1164px (4:6 ratio). Kinda a weird size. Let's update the viewport that the screenshots are being taken from to give us  images that are inline with what we expect from story posters (640x853) - which is actually 568x853px because we want a 2:3 aspect ratio. 

## Acceptance Criteria 
- PR to update the script 
- PR to static site replacing all of the template images (make sure to optimize the pngs)

PR #9631, opened 2021-11-03
Templates: Update CLI script to generate screenshots with different dimensions

BrittanyIRL:
## Context

The viewport size was arbitrarily set to have a height of 800, not thinking about how big the snapshots would be. 

## Summary

Update template CLI that generates screenshots used in dashboard and page templates to be taller and set scale to 1 so we get as close to the recommended minimum poster height of 853px. Still doing 2:3 aspect ratio.

## Relevant Technical Choices

- Updates the viewport size on launch and then also the additional page that opens for the previews 
- Set the deviceScaleFactor to 1 so that we get an image that is 853px height instead of 1706px. 
- Some tweaks to the script for legibility and making sure all the pages run 
- Adjust how 'Explore templates' link is clicked - copy can change depending on how many "new" templates there are, so just looking for the label that starts with 'explore templates' instead.

## To-do

- will follow up with all the new assets to the static site

## User-facing changes

None 

## Testing Instructions

You can run the script locally, `npm run workflow:render-template-posters` 

1.

No

No

No

## Checklist

---

Partially addresses #9582

### DIFF
--- a/packages/templates/src/cli.js
+++ b/packages/templates/src/cli.js
@@ -48,8 +48,8 @@ fs.mkdirSync(screenshotsPath, { recursive: true });
   await page.click('#wp-submit');
   await page.waitForNavigation({ waitUntil: 'networkidle2' });
 
-  await page.waitForSelector('aria/Explore Templates');
-  await page.click('aria/Explore Templates');
+  await page.waitForSelector('[aria-label^="Explore Templates"]');
+  await page.click('[aria-label^="Explore Templates"]');
 
   // Wait for templates to be rendered
   await page.waitForSelector('div.templateGridItem');
--- a/packages/templates/src/cli.js
+++ b/packages/templates/src/cli.js
@@ -120,7 +120,7 @@ fs.mkdirSync(screenshotsPath, { recursive: true });
       return document.querySelectorAll('amp-story-page').length;
     });
 
-    for (let currentPage = 1; currentPage < totalPages; currentPage++) {
+    for (let currentPage = 1; currentPage <= totalPages; currentPage++) {
       const templatePageSafeArea = await pagePreview.$(
         `amp-story-page:nth-child(${currentPage}) .page-safe-area`
       );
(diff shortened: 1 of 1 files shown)


swissspidy:
What is the resulting image size with this change? That's not really clear to me.

And why is the viewport width still 1600?

To get 640x853 sized images, can't we just set the viewport dimensions to exactly this?

BrittanyIRL:
> What is the resulting image size with this change? That's not really clear to me.
> 
> And why is the viewport width still 1600?
> 
> To get 640x853 sized images, can't we just set the viewport dimensions to exactly this?

I kept the viewport at 1600px wide because we still need to navigate the app beyond the previews and it'd be really hard at 640px wide to do that. I opted for consistency so that we know we have the size we want (height wise) for the screenshots based on the dimensions at launch.  

That said, it's fine if the width on the preview window gets shrunk - compared the 2 versions and there's no difference so I've updated that here. It doesn't impact the screenshot taken though since it's just the browser dimensions and we're grabbing the `page-safe-area`. 

We do need to do 854px though not 853px because we hit a breakpoint and the preview shrinks. 
| 853px | 854px |
| --- | --- |
| Screen Shot 2021-11-04 at 9 29 11 AM | Screen Shot 2021-11-04 at 9 29 32 AM |


The resulting screenshot regardless of the set height/width either 854px or 1145 is still 2:3 and 854px height so no need to rerun anything, this will be the same going forward with these updates.

maxyinger (approved):

packages/templates/src/cli.js
```diff
@@ -97,11 +97,12 @@ fs.mkdirSync(screenshotsPath, { recursive: true });
...
     await page.click('aria/Preview');
 
+    // the viewport height decides the size of the preview, we're aiming to maintain our recommended 853px height, we get the closest to this by setting the viewport height to 1145
     const pagePreview = await browser.newPage();
     await pagePreview.setViewport({
-      width: 1600,
-      height: 800,
-      deviceScaleFactor: 2,
+      width: 640,
+      height: 854, // 853 is a breakpoint and will shrink the preview to height of 630px
```
samwhale, line 104: nice comment 👍

samwhale (approved):

swissspidy (approved):

PR #9632, opened 2021-11-04
Website: Update Template Screenshots Dimensions

BrittanyIRL:
## Context

As pointed out by Pascal, the current template poster CLI script creates images that are 776x1164px. Kinda a weird size.

## Summary

Updating the assets based off of CLI update here: #9631 so that they are 568x853px instead - this takes into consideration the recommended height and grabs the page-safe-area which is 2:3 aspect ratio which is what we want for our use cases in the dashboard and page templates. 

## Relevant Technical Choices

Oh nothing technical, just a lot of webp and optimized pngs 

## Testing Instructions

1.

No

No

No

## Checklist

---

Partially addresses #9582


swissspidy (changes requested):
The image sizes are actually 568x853, not 640x853...?

BrittanyIRL:
> The image sizes are actually 568x853, not 640x853...?

Yes, this is intentional because we want to run with the 2:3 aspect ratio discussed here: https://github.com/google/web-stories-wp/issues/9583 instead of 3:4. 640x853 is in line with a 3:4 aspect ratio. I took the height as the baseline for this to keep it consistent with recommendations and then grabbed the page-safe-area like usual (which is inline with the 2:3).  

This is a little confusing because the poster size recommendations _are_ in 3:4 ratio but you'll note there's also some cut off on those in various 'preview' (Search, our dashboard) - which is what we discussed in that 9583 ticket/corresponding slack convo from last week. So for the purpose of these screenshots, I think this is right.

swissspidy:
The PR description says 640x853, which is very confusing and why I flagged this.

In that case, this should be fine.

Your explanation would have made a great PR description, FWIW.

Out of curiosity, how is the appearance in the dashboard when using 640x853 images?

BrittanyIRL:
> The PR description says 640x853, which is very confusing and why I flagged this.
> 
> In that case, this should be fine.

Yeah, when we talked last week i made the ticket really quick and just didn't bother updating it, this is my bad and I'll go correct it (but also i'm glad to touch base, the sizing of various things the past like year has often been a point of confusion so more talking about it is always good) 

> Out of curiosity, how is the appearance in the dashboard when using 640x853 images?

It's great! It's exactly those screenshots here: https://github.com/google/web-stories-wp/issues/9583#issuecomment-955099157

swissspidy:
> > Out of curiosity, how is the appearance in the dashboard when using 640x853 images?
> 
> It's great! It's exactly those screenshots here: [#9583 (comment)](https://github.com/google/web-stories-wp/issues/9583#issuecomment-955099157)

Aren't those the **568x853** screenshots though?

I'd be curious to compare with **660x853** specifically. Yes there will be some cutoff, but we'd be dealing with consistently sized images. Hence curious to see a comparison.

BrittanyIRL:
> Aren't those the **568x853** screenshots though?

yes, the 2:3 update. 
 
> I'd be curious to compare with **660x853** specifically. Yes there will be some cutoff, but we'd be dealing with consistently sized images. 

Do you mean comparing template screenshots that are this ratio? because that would mean taking screenshots that include the fullbleed area, which I thought we were avoiding. 


I (for one) am getting a little confused. I'm just going to try and level set here. 

**STORIES** 
- Use a poster image to represent them on the dashboard. Up until this PR #9599 they were in 3:4 ratio (640x853px) which is the same size we recommend users have as a minimum on their poster images for the story. 
- In 9599 we update that ratio to be 2:3 (569x853px). We're not altering the poster size, we're just covering the space with the previews and letting a bit be cut off a bit which is what  we discussed in slack. 
- Digging deeper in to the preview shown on Google Search with Visual Stories, the aspect ratio there is actually still 3:4 so now I'm really not sure why we were updating the stories in the first place. 

**Templates** 
- Have always used the page-safe-area as the preview, this is always a 2:3 ratio. The `PAGE_RATIO` used to create the 2:3 ratio is the default page ratio stored in `packages/units`. 

So you're asking 640x853 (3:4) vs 569x853(2:3) in our given grid structure of the dashboard but the grid itself is what's setting the width, we're just using the ratio and the available page space to set the grid items, so it is purely a matter of what consistency of space we want out of these screenshots. Do we want a 2:3 ratio or do we want a 3:4.  

It feels like we're going around in a circle. 

All this PR for the template screenshots is doing is making them smaller, there's no reason to have pngs that are 1164px.  I went with the 2:3 ratio on these because thats what the grid's getting updated to and that is consistent with the safe area on the screenshots that have been what we've used since this feature rolled out. If your point is that you would like to see the screenshots be the fullbleed area then that's different. 

Here's why we use the safe-area: 
| just `<amp-story-page>` | fullbleed | safe area | 
| --- | --- | --- | 
| 1-justpage | 1-fullbleed |  1-safe |
 

So, the fullbleed is a bit taller, yeah, but it's going to include those little gray progress lines, and the full `<amp-story-page>` will have some space on the sides because it's part of the space that that element takes up on the page which is all the screenshot cares about.

swissspidy:
Thanks for bearing with me here! Trying to rephrase...

> Digging deeper in to the preview shown on Google Search with Visual Stories, the aspect ratio there is actually still 3:4 so now I'm really not sure why we were updating the stories in the first place.

That is being updated FWIW.

> In 9599 we update that ratio to be 2:3 (569x853px). We're not altering the poster size, we're just covering the space with the previews and letting a bit be cut off a bit which is what we discussed in slack.

Let me try to rephrase my question:

**Could** we **theoretically** do the same for the templates as well?

With that I mean: Use 640x853 template images but display them in 2:3 ratio, so 569x853 (or was it 568x853? 🤔 )

So I was wondering what this 2:3 display does look like when (again, theoretically) using 640x853 template images (instead of cropped, pixel-perfect 569x853 images).

BrittanyIRL:
Back atcha for bearing with me - what a journey! 

So, I think I've found our disconnect: Templates do not have poster images and can't have a 3:4 ratio screenshot taken in the story preview because the story preview is in 2:3 ratio. 

What we're doing with that script is just zeroing in on the page safe area (sorry i'm a broken record, i keep trying to spell this out just to find my way to the surface as well), this is the first layer of the page that gives us pure template preview (no Story related controls or progress bar). We _could_ hide the  `aside` that is overlayed on the fullbleed area that makes that screenshot option have the gray bar and control buttons a viable option but that's still a 2:3 ratio so it's not solving anything.   

Otherwise, we'd have to cut off some height to get the 3:4 and that's weird. 

I feel like having the templates in 2:3 screenshots of the safe zone will look and feel consistent with the story poster images (which are 3:4) just filling the 2:3 grid space. Looking at the PR that updates the dashboard stories grid to 2:3 (#9599 ) this is how they compare: 

Screen Shot 2021-11-05 at 10 22 24 AM
Screen Shot 2021-11-05 at 10 22 14 AM


I know you mean theoretically grabbing wider template screenshots, but there really isn't a way to do this within our script/the preview structure.

samwhale (approved):

maxyinger (approved):

Issue #9577, opened 2021-10-29
Media: Do not store transient data on resource objects


swissspidy:
## Feature Description

We currently store some transient flags on media resources when performing some tasks on them, e.g. `isTranscoding`, `isTrimming`, `isMuting`, and even `local`. That's really not the best way to handle this, especially when a resource is used in multiple places.

Storing this transient data on the resources themselves might also cause issues with the edit history, see https://github.com/google/web-stories-wp/issues/9550#issuecomment-953828865. Another good reason to change this.

**Idea**

Export some helper function(s) in `useMediaUploadQueue` to get a specific resource's state, and re-export that in the local media provider for use by the rest of the editor.

Perhaps something like this:

```js
// in packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js

isResourceUploading: (resourceId) =>
  state.queue.some(
    (item) =>
      !['UPLOADED', 'CANCELLED', 'PENDING'].includes(item.state) &&
      item.resource.id === resourceId
  ),
isResourceTranscoding: (resourceId) =>
  state.queue.some(
    (item) =>
      item.state === 'TRANSCODING' && item.resource.id === resourceId
  ),
isResourceMuting: (resourceId) =>
  state.queue.some(
    (item) => item.state === 'MUTING' && item.resource.id === resourceId
  ),
isResourceTrimming: (resourceId) =>
  state.queue.some(
    (item) =>
      item.state === 'TRIMMING' && item.resource.id === resourceId
  ),
```

So we don't have to check for `resource.isTranscoding` and the like anymore.

Then, update all the reducers in there which currently set `isTranscoding` et al on the resources. We don't need that anymore.

**Note:** need to update `createResource` and the `Resource` JSDoc accordingly.

**Related:** We should probably generate a resource ID with `uuid` in `getResourceFromLocalFile` and `getResourceFromUrl` so that all resources have an ID. Just to make our lives easier in the future.

spacedmonkey:
Reposting this https://github.com/google/web-stories-wp/pull/9758#discussion_r754113046 as I think it is related. 

I think we should remove the `local` property from resources. We can replace it with a new property called `source`. When a media element is uploaded, instead of changing local from `true` / `false`, you could just change source from `local` to `library`. This makes gets rid of a property that is not useful outside the upload process and adds a useful field for the future. This would of course require a migrate but is possible. 

Thoughts @swissspidy

swissspidy:
[...]

PR #9638, opened 2021-11-04
Media: Improve average color handling & performance 

### LINKED ISSUE #9550: Media: Improve average color handling & performance
## Feature Description

In a few places we use colorthief to get the average/base color of an image or video poster:

* When dropping media elements (IDK why it's done here really 🤔 )
* In the media library, using `useAverageColor`, as soon as images are loaded
* In the Size & Position design panel, when clicking on the "Set as background" button

This average color is used for:

* for skeleton pages in the carousel, to determine the page color based on the background element's average color
* for frontend output, to determine background color based on the background element's average color

Some thoughts based on that:

* The `useAverageColor` usage in the media library is **extremely wasteful**. Right when loading the editor, **all images** will be run through colorthief. When you perform a search or any other action, they will be run through **again**
* ...This will take a loooong time
* At the very least this could happen in a `requestIdleCallback` call instead
* There's no reason why this calculation happens for all media items on load, even when they will most likely never be inserted into the story.
* There's no reason why this calculation needs to happen during drag & drop.
* For frontend output, this background color could be used for _all_ images as a fallback. That would be a nice UX enhancement.
* This calculation could simply happen when first inserting a media element onto the canvas. Just once.
* The media average color could be generated _during upload_ and then stored permanently in the database. It could be backfilled for existing media items.

spacedmonkey:
## Summary

Improve performance, by async generating base color of image and storing it against the attachment. Remove all the places color is generated on fry and only generate in three cases. 

1. Async after inserting a element into page
2. When media library is loaded ( and cached ). 
3. When new assets are uploaded. 

In PHP, I added a new interface, called HasMeta and tweaked some logic and tests. Not needed for this PR, but is nice. 

Base color, is now also a string / hex value. This PR includes a migration from array to hex value. 

## Testing Instructions

1.

## Checklist

---

Fixes #9550


packages/story-editor/src/components/canvas/useInsertElement.js
```diff
@@ -98,6 +132,7 @@ function useInsertElement() {
       if (resource) {
         backfillResource(resource);
         handleRegisterUsage(resource);
+        generateBaseColorExternal(element);
```
miina: Hm, the only issue here is with adding incorrect history entries if the element is changed _after_ insertion and not at the same time.

Pascal created [an issue](https://github.com/google/web-stories-wp/issues/9577) to move the transient data out of the `resource` object, could we perhaps not store the `baseColor` there either anymore to avoid history issues? Or is there any other alternative for not corrupting history? Thoughts?

An example issue it can create:
- User inserts an image
- Two history entires get created: the insertion and adding the base color
- User clicks "Undo" once
- The image stays but the base color is removed, nothing changes visibly for the user.

swissspidy: That issue is for transient state like `isMuting` though, which we only need to know about for a couple of seconds. But `baseColor` is something that will need to be stored **persistently**.

Where else would we store it if not on the resource?

> Or is there any other alternative for not corrupting history?

I was hoping you had some ideas :-)

I see three options:

1. Generate base color **async** after insertion like in this PR -> corrupts history
2. Generate base color **sync** during/before insertion -> not ideal UX
3. Generate base color **async** after insertion, but don't create a new history entry and just add it to the current history entry. -> better UX, no corrupt history  
  Kinda what you mentioned [here](https://github.com/google/web-stories-wp/issues/9550#issuecomment-953828865)

As for (2), you mentioned [alternative options](https://github.com/google/web-stories-wp/issues/9550#issuecomment-953801633) like calculating on hover, but that doesn't seem feasible as it's not super fast. So no guarantee that it works.

IMO (3) would be most compelling: Having a way to update an element/story/page/resource without adding a new history entry.

I think such a functionality would be useful beyond this single use case.

includes/Infrastructure/HasMeta.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+interface HasMeta {
+	/**
+	 * Register meta
+	 *
+	 * @since 1.15.0
+	 *
+	 * @return void
+	 */
+	public function register_meta();
```
swissspidy, line 27: Why public?

includes/Media/Base_Color.php
```diff
@@ -0,0 +1,106 @@
...
+			self::BASE_COLOR_POST_META_KEY,
+			[
+				'type'           => 'array',
+				'description'    => __( 'Attachment base color', 'web-stories' ),
+				'show_in_rest'   => [
+					'schema' => [
+						'type'  => 'array',
+						'items' => [
+							'type' => 'integer',
+						],
```
swissspidy: Yesterday we talked about storing hex colors instead of RGB arrays, discussing advantages like easier sanitization etc.

I just learned that the REST API  (since WP 5.5) supports a special `hex-color` format.

So we could do

```suggestion
						'type'  => 'string',
						'format' => 'hex-color',
```

Just mentioning this here for context.

packages/story-editor/src/app/media/local/reducer.js
```diff
@@ -39,6 +41,7 @@ const INITIAL_STATE = {
  * @typedef {import('./typedefs').LocalMediaReducerState} LocalMediaReducerState
  */
 
+/* eslint-disable complexity */
```
swissspidy: On line 17 above the complexity threshold is already set to 30.

Instead of just disabling the rule entirely here we should increase that threshold there.

Also, let's create a ticket to refactor this file here.

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -217,13 +222,21 @@ function useProcessMedia({
...
       const onUploadSuccess = ({ resource }) => {
         copyResourceData({ oldResource, resource });
-        if ('video' === resource.type && !resource.local) {
-          if (!resource.posterId) {
-            uploadVideoPoster(resource.id, resource.src);
-          }
-          if (resource.isMuted === null) {
-            updateVideoIsMuted(resource.id, resource.src);
-          }
+        if (resource.local || 'video' !== resource.type) {
```
swissspidy: That means `updateBaseColor` below won't run for images.

`updateBaseColor` needs to run for images as well.

packages/story-editor/src/components/canvas/useInsertElement.js
miina: > As for (2), you mentioned alternative options like calculating on hover, but that doesn't seem feasible as it's not super fast. So no guarantee that it works.

We could potentially wait with insertion until the calculation that started on hovering has been completed. However, agreed that (3) would be the best option.

Perhaps we can change the logic of history so that it would generally allow ignoring specific props as needed. This would need a separate issue though for the history logic change, so that this PR could just define the variable to ignore. cc @barklund since you implemented the original history tracking.

Is it acceptable to corrupt the history for now within this PR, or should this new history change be a blocker for the current PR?

packages/story-editor/src/app/media/utils/useProcessMedia.js
spacedmonkey: This function is only ever passed a video resource. This check is just being overly careful.

packages/story-editor/src/components/canvas/useInsertElement.js
spacedmonkey: My vote is for 3. 

If there a way to add a param to `updateElementById` that would skip the adding to history. Something like this. 

```
updateElementById({ elementId: id, properties: { ...changes, ignoreHistory: true } });
```

packages/story-editor/src/app/media/utils/useProcessMedia.js
swissspidy: Ah, just realized we're in `useProcessMedia` / `trimExistingVideo` here.

packages/story-editor/src/components/canvas/useInsertElement.js
miina: Added an issue for it: https://github.com/google/web-stories-wp/issues/9645

There's one issue that would come with this approach as well (editor would not indicate unsaved state), added details to the description.

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -125,19 +126,21 @@ function useProcessMedia({
...
+
         if (
-          'video' === resource.type &&
-          !resource.local &&
-          resource.isMuted === null
+          !Array.isArray(resource?.baseColor) ||
+          !resource?.baseColor.length
         ) {
-          updateVideoIsMuted(resource.id, resource.src);
+          updateBaseColor({ resource });
```
swissspidy: Why are we calling `updateBaseColor` when optimizing an existing video?

Are there concerns the original video doesn't have a baseColor set?

packages/story-editor/src/components/canvas/useInsertElement.js
spacedmonkey: It is also worth noting, this generation of base color on insertion like this, is only done on hotlink / external resources. Base Color is generated once and cache for assets stored / cached. So even if merge this, it will only be a issue for external resources.

swissspidy: > Is it acceptable to corrupt the history for now within this PR

Yes IMHO. Soon this will be a rare scenario, because in the future all media items will have a base color during upload.

includes/Infrastructure/HasMeta.php
spacedmonkey: > All interface methods must be public, while abstract class methods is public or protected

packages/story-editor/src/app/media/local/reducer.js
spacedmonkey: fixed.

spacedmonkey: Created a follow up ticket for this - https://github.com/google/web-stories-wp/issues/9646

packages/story-editor/src/app/media/utils/useProcessMedia.js
spacedmonkey: > Are there concerns the original video doesn't have a baseColor set?

Either the original video didn't have one set or opertizing the video change the video enough that the base color is different. It is a simple post about muting, this should be copied over, just being overly careful here.

includes/Media/Base_Color.php
spacedmonkey: Added some more php level validation.

spacedmonkey: I know about the hex-color validation, but I don't think it makes sense to store as one value and convert it in javascript. If the javascript needs an array

includes/Media/Base_Color.php
```diff
@@ -0,0 +1,109 @@
...
+
+	/**
+	 * Register meta
+	 *
+	 * @since 1.15.0
+	 *
+	 * @return void
+	 */
+	public function register_meta() {
+		register_meta(
```
swissspidy, line 67: Why not using `register_post_meta` here btw?

tests/phpunit/integration/tests/Media/Base_Color.php
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+namespace Google\Web_Stories\Tests\Integration\Media;
+
+use Google\Web_Stories\Tests\Integration\TestCase;
+
+/**
+ * @coversDefaultClass \Google\Web_Stories\Media\Base_Color
+ */
+class Base_Color extends TestCase {
```
swissspidy, line 25: Note that you're registering the post meta as an RGB array, but here you're testing with hex strings

packages/story-editor/src/components/library/panes/media/local/mediaPane.js
```diff
@@ -239,12 +245,15 @@ function MediaPane(props) {
...
 
       if (
-        !resource.local &&
         allowedVideoMimeTypes.includes(resource.mimeType) &&
         resource.isMuted === null
       ) {
         updateVideoIsMuted(resource.id, resource.src);
       }
+
+      if (!resource?.baseColor || !resource?.baseColor.length) {
```
swissspidy: Nit:

```suggestion
      if (!resource?.baseColor?.length) {
```

packages/story-editor/src/utils/getMediaBaseColor.js
```diff
@@ -46,13 +46,18 @@ function getImgNodeKey(elementId) {
...
+
+  return setOrCreateImage({
+    src: type === 'image' ? src : poster,
     width: 10,
     height: 'auto',
-  }).then(
+  });
+}
+
+export function getMediaBaseColor(resource, onBaseColor) {
```
swissspidy: If I am not mistaken, this function is now completely unused and could be removed

packages/story-editor/src/components/dropTargets/provider.js
```diff
@@ -176,39 +175,29 @@ function DropTargetsProvider({ children }) {
           resource,
         });
       }
-      const finalizeDrop = (baseColor = null) => {
-        if (baseColor) {
-          combineArgs.firstElement.resource.baseColor = baseColor;
```
swissspidy, line 181: Is `baseColor` still properly added to `firstElement`?

packages/story-editor/src/components/canvas/useInsertElement.js
```diff
@@ -56,14 +59,45 @@ function useInsertElement() {
...
       }
+      // Backfill base color, if one not set.
+      if (!Array.isArray(baseColor) || !baseColor.length) {
+        updateBaseColor({ resource });
+      }
+    },
+    [uploadVideoPoster, updateBaseColor]
+  );
+
+  const generateBaseColorExternal = useCallback(
```
swissspidy: Why is this limited to external items only?

There could be a scenario where a regular media item doesn't have a baseColor, so this would be very useful to run for those cases as well.

spacedmonkey:
I would like to see https://github.com/google/web-stories-wp/pull/9652 merge before this PR goes in.

swissspidy:
Some thoughts after reviewing the code so far...

Overall the approach looks good. Backfilling really makes things a lot smoother!

There is still 1 usage of `getMediaBaseColor` that should be removed. It's in the Size & Position design panel. The base color is calculated when clicking on "Set as background". That is no longer needed since we backfill this everywhere:

https://github.com/google/web-stories-wp/blob/d4bb3eba1a09542ae91002dcfdaed4a5cdd0c4ba/packages/story-editor/src/components/panels/design/sizePosition/sizePosition.js#L237-L241

**[Decision]** I now we've talked about it already in our sync, but I thought about it a bit more and I **do** think we should update `baseColor` everywhere to be a hex string instead of an RGB array. So let's update this everywhere. Including any existing code. Even if it means adding a data migration.

**[Note]** While testing this PR I found a slightly related bug. See #9653

packages/story-editor/src/components/dropTargets/provider.js
spacedmonkey: baseColor is added to all resources...

packages/story-editor/src/utils/getMediaBaseColor.js
spacedmonkey: It still used here. https://github.com/google/web-stories-wp/blob/a6509894b5da58af92823b607c05751f6c3822a0/packages/story-editor/src/components/panels/design/sizePosition/sizePosition.js#L240

Can we remove the base color out of size position @miina ?

includes/Media/Base_Color.php
spacedmonkey: We don't use `register_post_meta` in other places. It is all the same to me. register_post_meta is just a wrapper function.

spacedmonkey:
> [Decision] I now we've talked about it already in our sync, but I thought about it a bit more and I do think we should update baseColor everywhere to be a hex string instead of an RGB array. So let's update this everywhere. Including any existing code. Even if it means adding a data migration.

Can you explain your reasoning here? What benefit was it being a hex value would add? It maybe a lot of work to make this change, just want to ensure that the work is justified. Could it also be done in another, follow up PR?

swissspidy:
> > [Decision] I now we've talked about it already in our sync, but I thought about it a bit more and I do think we should update baseColor everywhere to be a hex string instead of an RGB array. So let's update this everywhere. Including any existing code. Even if it means adding a data migration.
> 
> Can you explain your reasoning here? What benefit was it being a hex value would add? It maybe a lot of work to make this change, just want to ensure that the work is justified. Could it also be done in another, follow up PR?

Sure can be done in a separate PR, but ideally _before_ this one here, not _after_. Otherwise we need a database migration.

There are multiple reasons that led me to this conclusion:

* If we'd ever want to change the data format, now's the best time to think about it and do it
* There really is no point in `baseColor` being an RGB array in the first place. There's no benefit in using an array for that
* Hex strings take less space, are easier to store (no serialization), validate,
*  `baseColor` will be available as a hex string only in the media3p API (for Unsplash images)
* Storing an RGB array in the database will cause it to be serialized  
  * This will take quite some space, e.g. ` [255, 255, 255]` will be stored as `a:3:{i:0;i:255;i:1;i:255;i:2;i:255;}` (36 characters), whereas `#FFFFFF` is just 7 characters.
  *  Hex strings would also be much easier to handle if ever needed to perform some more detailed

Also, FWIW, it would be trivial to support both arrays and hex strings. Would just need to use `getSolidFromHex` to transform hex strings into RGB arrays, or alternatively add some `Array.isArray()` checks in places like this:

https://github.com/google/web-stories-wp/blob/681dead5acbc5648261a8dfbc20efe58212b78ad/packages/story-editor/src/output/page.js#L48-L52

But maintenance is a bit easier if the data format is consistent.

packages/story-editor/src/utils/getResourceBaseColor.js
```diff
@@ -46,16 +46,19 @@ function getImgNodeKey(elementId) {
...
+
+  const imageSrc = type === 'image' ? src : poster;
+
+  if (!imageSrc) {
+    return Promise.reject(new Error('No source to image'));
+  }
+  return setOrCreateImage({
+    src: imageSrc,
     width: 10,
     height: 'auto',
```
spacedmonkey: @miina Is such a small image here by design? We could use the height / width from the resource and get the full size image and pick from that.

miina: Probably for performance reasons since it's an average color anyway. Would using a full size image make a difference?

packages/story-editor/src/utils/getResourceBaseColor.js
```diff
@@ -46,16 +46,19 @@ function getImgNodeKey(elementId) {
...
-  }).then(
-    (color) => onBaseColor(color),
-    () => onBaseColor([255, 255, 255]) // Fallback color is white.
```
miina, line 57: FI: the Color Thief throws an error when the image is fully white (it's a feature), so in case of a white image we'd need to set it inside a catch.

packages/story-editor/src/utils/getResourceBaseColor.js
```diff
@@ -46,16 +46,19 @@ function getImgNodeKey(elementId) {
...
   const { type, src, poster } = resource;
-  setOrCreateImage({
-    src: type === 'video' ? poster : src,
+
+  const imageSrc = type === 'image' ? src : poster;
+
+  if (!imageSrc) {
+    return Promise.reject(new Error('No source to image'));
+  }
+  return setOrCreateImage({
```
miina: Nit: Maybe change the name of this now for readability to indicate that it's not just setting/creating image but also getting the average color?

packages/story-editor/src/components/library/panes/media/common/innerElement.js
```diff
@@ -244,29 +224,18 @@ function InnerElement({
     handleDrag(resource, event.clientX, event.clientY);
   };
 
+  const thumbnailUrl = type === ContentType.IMAGE ? thumbnailURL : poster;
```
miina: Nit: Any other options here instead of using `thumbnailUrl` and `thumbnailURL` as different constants? Can be a bit confusing. Maybe overwrite the `thumbnailUrl` instead?

packages/story-editor/src/app/media/local/useContextValueProvider.js
```diff
@@ -182,10 +189,33 @@ export default function useContextValueProvider(reducerState, reducerActions) {
...
 
+  const processMediaBaseColor = useCallback(
+    (resource) => {
+      const { baseColorProcessed, baseColorProcessing } = stateRef.current;
+      const { id } = resource;
+
+      const process = async () => {
+        // Simple way to prevent double-uploading.
+        if (
+          baseColorProcessed.includes(id) ||
```
miina, line 198: Wanted to test out how it works and did the following:
- Checked out the PR & reloaded the editor
- Dropped media from the media library to the Page as background (tried both with video and image)
- Saved and checked the preview -- didn't see the background color assigned (seemingly no base color was added in the front-end)
- Went back to the editor and refreshed the page
- Got the following error for this line:
Screenshot 2021-11-09 at 13 34 15

Is there something I'm missing?

packages/story-editor/src/utils/getResourceBaseColor.js
spacedmonkey: I am guessing, that a 10x10 image compared to a 600x600 image would get a different primary color from colorthief? Don't you agree?

miina: Not sure, would need testing out to see if it makes a difference that could actually result in an incorrect color that would not fulfil its purpose.

It's possible to test it here for example (in the "Try it yourself" section): https://lokeshdhakar.com/projects/color-thief/

packages/migration/src/migrations/v0036_changeBaseColorToHex.js
```diff
@@ -0,0 +1,51 @@
...
+}
+
+function reducePage({ elements, ...rest }) {
+  return {
+    elements: elements.map(updateElement),
+    ...rest,
+  };
+}
+
+function getHexFromSolidArray(baseColor) {
```
spacedmonkey, line 31: @swissspidy This function appears twice in the code base. Should it just live in the pattern package and migration imports it? Or the other way around?

includes/Media/Base_Color.php
swissspidy: We _do_ use it in `Story_Post_Type`.

packages/migration/src/migrations/v0036_changeBaseColorToHex.js
swissspidy: Migration should not import from another package, in order to keep it as lightweight as possible and prevent unexpected changes should the other package's implementation ever change.

So what you have now is OK.

packages/media/src/createResource.js
```diff
@@ -27,7 +27,7 @@ import normalizeResourceSizes from './normalizeResourceSizes';
  * @return {import('./types').Resource} Resource object.
  */
 function createResource({
-  baseColor,
+  baseColor = '',
```
swissspidy: Curious, why default to an empty string instead of `null`?

packages/story-editor/src/utils/getResourceBaseColor.js
spacedmonkey: Thanks for this, I have fixed in PR

packages/story-editor/src/components/library/panes/media/common/innerElement.js
spacedmonkey: Fixed

packages/story-editor/src/app/media/local/useContextValueProvider.js
spacedmonkey: Rebase issue. Can you test again?

includes/Media/Base_Color.php
spacedmonkey: @swissspidy You add this line - https://github.com/google/web-stories-wp/commit/9d18d5f52f1b8a88f170df6be099b0a9d69e40c2

Personally, I like using `register_meta`, it gives you full control, I don't really care either way.

packages/media/src/createResource.js
spacedmonkey: So the type of jsdocs are correct. So it is always a string.

packages/story-editor/src/app/media/local/useContextValueProvider.js
miina: The base color is assigned now as expected, thanks!

Still seeing an error (a different one) when reloading the editor after dragging the media to background and saving:
Screenshot 2021-11-09 at 18 20 59

Thoughts?

packages/story-editor/src/utils/getResourceBaseColor.js
swissspidy: Performance > 100% accuracy

packages/story-editor/src/app/media/local/useContextValueProvider.js
spacedmonkey: @miina I am not seeing that. Try again now and ensure to get the latest version. 

If you are still seeing it, steps to replicate / video would be useful.

packages/story-editor/src/components/canvas/useInsertElement.js
```diff
@@ -66,6 +71,40 @@ function useInsertElement() {
...
+      } = resource;
+
+      const imageSrc = type === 'image' ? src : poster;
+      if (local || !imageSrc || currentBaseColor) {
+        return;
+      }
+
+      if (isExternal) {
+        const imageSrcProxy = getProxiedUrl(resource, imageSrc);
+        const baseColor = await getMediaBaseColor(imageSrcProxy);
```
swissspidy: Error is not caught here.

packages/story-editor/src/utils/getResourceBaseColor.js
swissspidy: @spacedmonkey Where is this fixed?

I still see no `catch` block in this function. This should still be there.

packages/story-editor/src/utils/getMediaBaseColor.js
```diff
@@ -46,16 +47,15 @@ function getImgNodeKey(elementId) {
...
 }
 
-export function getMediaBaseColor(resource, onBaseColor) {
-  const { type, src, poster } = resource;
-  setOrCreateImage({
-    src: type === 'video' ? poster : src,
+export function getMediaBaseColor(src) {
+  if (!src) {
+    return Promise.reject(new Error('No source to image'));
+  }
```
swissspidy, line 53: Which scenario is this for? Where is this error caught? Why not a simple fallback?

packages/story-editor/src/utils/getMediaBaseColor.js
```diff
@@ -46,16 +47,15 @@ function getImgNodeKey(elementId) {
...
+  return setOrCreateImage({
+    src,
     width: 10,
     height: 'auto',
-  }).then(
-    (color) => onBaseColor(color),
-    () => onBaseColor([255, 255, 255]) // Fallback color is white.
-  );
+  });
```
swissspidy: Why not `try { ... } catch { return '#ffffff' }` or something?

packages/story-editor/src/app/media/local/useContextValueProvider.js
miina: @spacedmonkey Yep, still seeing. Here's a video:

https://user-images.githubusercontent.com/3294597/141096530-a7090118-5acb-4501-bf11-263b0e373989.mp4

packages/story-editor/src/utils/getMediaBaseColor.js
spacedmonkey: Sometime earlier when a video is created, a poster is not generated, meaning the source passed here might be empty. I believe I have got all the place this might happen, but just be safe, this needs to error, so try catches can work.

packages/story-editor/src/utils/getMediaBaseColor.js
spacedmonkey: I wanted to return a promise. How you handle a failure, might be different in different usage.

packages/story-editor/src/components/canvas/useInsertElement.js
spacedmonkey: Fixed.

packages/story-editor/src/components/carousel/skeletonPage.js
```diff
@@ -36,8 +39,8 @@ const EmptyPage = styled.li.attrs({ role: 'presentation' })`
   border-radius: 4px;
 `;
 
-function getPatternFromArray([r, g, b]) {
-  return { color: { r, g, b } };
+function getPatternFromArray(baseColor) {
+  return getSolidFromHex(baseColor.replace('#', ''));
 }
```
swissspidy: `baseColor` is a string here, why still call the function `...fromArray`? 🤔

packages/story-editor/src/utils/getMediaBaseColor.js
swissspidy: That does not mean you cannot return a Promise.

As @miina pointed out, Color Thief returns an error on all-white images. We need to catch that here in `getMediaBaseColor` and return `#ffffff` as a fallback.

It doesn't make sense for every consumer of `getMediaBaseColor` to add this fallback.

It should be done within this function.

packages/story-editor/src/app/media/local/useContextValueProvider.js
miina: Checked again after new changes -- seems to be fixed now.

packages/story-editor/src/utils/getMediaBaseColor.js
swissspidy: Is this message ever surfaced to the user?

packages/story-editor/src/utils/getResourceBaseColor.js
spacedmonkey: Catch it upstream.

packages/story-editor/src/utils/getMediaBaseColor.js
spacedmonkey: Nope

packages/story-editor/src/components/carousel/skeletonPage.js
spacedmonkey: `generatePatternStyles` used a couple of lines down,  expects a rgba object to be passed it. It is a bigger refactor to have this to expect a string and convert it there.

packages/story-editor/src/utils/getResourceBaseColor.js
spacedmonkey: Try / catch has now been added.

packages/story-editor/src/components/carousel/skeletonPage.js
swissspidy: `baseColor` is not an array. So this function should not be called `...fromArray`.

packages/story-editor/src/utils/getMediaBaseColor.js
```diff
@@ -46,16 +47,25 @@ function getImgNodeKey(elementId) {
...
+  }
+  let color;
+  try {
+    color = await setOrCreateImage({
+      src,
+      width: 10,
+      height: 'auto',
+    });
+  } catch (error) {
+    // Known error of color thief with white only images.
```
swissspidy, line 62: Do I understand this comment correctly in that white-only images result in TypeError?

spacedmonkey: It does, other kinds of errors are throw still.

packages/story-editor/src/components/carousel/skeletonPage.js
spacedmonkey: Removed.

packages/migration/src/migrations/v0036_changeBaseColorToHex.js
```diff
@@ -0,0 +1,51 @@
...
+    .map((s) => s.padStart(2, '0'))
+    .join('');
+  return '#' + color;
+}
+
+function updateElement(element) {
+  if (
+    element?.resource?.baseColor &&
+    typeof element?.resource?.baseColor === 'object'
+  ) {
```
swissspidy:
```suggestion
  if (Array.isArray(element?.resource?.baseColor)) {
```

Such a check is preferred.

All unit tests still pass with this.

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -165,27 +151,33 @@ function useProcessMedia({
...
           return;
         }
+
+        const additionalData = {
+          original_id: oldResource.id,
+          web_stories_is_muted: oldResource.isMuted,
+        };
+
+        if (oldResource?.baseColor) {
+          additionalData.meta.web_stories_base_color = oldResource.baseColor;
```
swissspidy: Pretty sure this will throw an error because `meta` is not defined.

You can try this yourself by running this in the console:

```js
const foo = { bar: 123 }; 
foo.baz.barbar = 456; // Uncaught TypeError: Cannot set properties of undefined (setting 'barbar')
```

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -350,17 +330,23 @@ function useProcessMedia({
...
 
+        const additionalData = {
+          original_id: oldResource.id,
+          web_stories_media_source: oldResource?.isOptimized
+            ? 'video-optimization'
+            : 'editor',
+        };
+
+        if (oldResource?.baseColor) {
+          additionalData.meta.web_stories_base_color = oldResource.baseColor;
```
swissspidy: Ditto https://github.com/google/web-stories-wp/pull/9638/files?w=0#r749386836

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -425,23 +404,29 @@ function useProcessMedia({
           return;
         }
 
+        const additionalData = {
+          original_id: oldResource.id,
+        };
+
+        if (oldResource?.baseColor) {
+          additionalData.meta.web_stories_base_color = oldResource.baseColor;
```
swissspidy: Ditto https://github.com/google/web-stories-wp/pull/9638/files?w=0#r749386836

swissspidy (commented):
Just a couple of nits, otherwise seems solid

packages/migration/src/migrations/v0036_changeBaseColorToHex.js
spacedmonkey: This is the way it was and it didn't work for some reason. We spoke of this in our sync.

packages/story-editor/src/app/media/utils/useProcessMedia.js
spacedmonkey: Migration of base color now happens php / rest api.

packages/migration/src/migrations/v0036_changeBaseColorToHex.js
spacedmonkey: I have changed it.

includes/REST_API/Stories_Media_Controller.php
```diff
@@ -177,13 +178,15 @@ protected function process_post( $post_id, $parent_post, $original_id ) {
...
 
-			if ( ! empty( $image_alt ) ) {
-				// update_post_meta() expects slashed.
-				update_post_meta( $post_id, '_wp_attachment_image_alt', wp_slash( $image_alt ) );
-			}
+				if ( ! empty( $value ) ) {
+					// update_post_meta() expects slashed.
+					update_post_meta( $post_id, $meta_field, wp_slash( $value ) );
+				}
+			}
```
swissspidy:
```suggestion
			}
```

packages/story-editor/src/app/media/utils/useProcessMedia.js
swissspidy: @spacedmonkey Curious, what's the reason for this change?

IIRC you previously mentioned running it client-side is more safe because the base color could change after trimming for example.

spacedmonkey: When we added server side migration, it was to stop stuff like this. Migrating on the server ends up with cleaner logic in javascript and no repeated code.

swissspidy (approved):

packages/story-editor/src/components/dropTargets/provider.js
swissspidy: That's not what I meant.

Setting `combineArgs.firstElement.resource.baseColor` still seems needed.

swissspidy:
Looks like there is some legit Karma test failure:

https://github.com/google/web-stories-wp/runs/4230805407?check_suite_focus=true#step:7:8443

```
Drop-Target integration when there is nothing on the canvas should correctly handle image dragged from library straight to edge FAILED
Error: Expected '#060606' to equal '#2c1c1c'.
```

packages/story-editor/src/components/dropTargets/karma/dropTarget.karma.js
```diff
@@ -100,7 +100,7 @@ describe('Drop-Target integration', () => {
       expect(rep2).toBeEmpty();
 
       // Verify the background base color is handled as expected.
-      expect(bgElement.resource.baseColor).toEqual([201, 201, 219]);
+      expect(bgElement.resource.baseColor).toEqual('#060606');
```
swissspidy: Where did `[201, 201, 219]` come from before? Where is `#060606` coming from now?

I want to make sure that this is not actually a regression.

packages/story-editor/src/components/dropTargets/provider.js
spacedmonkey: Is it needed? We now generate base color for all elements and not just the first element.

packages/story-editor/src/components/dropTargets/karma/dropTarget.karma.js
spacedmonkey: I updating the fixture data seems to have fixed this issue. You have to remember, this is a change of behaviour now. Base color will be generated on update, meaning it will available. So fixture data needed to be updated.

miina: The point of this check was previously to ensure that when media without base color is added to the page, it's base color gets calculated correctly.

Since now it's just preset in the fixture, there's no check for that, it just uses the color from the preset values in the fixture.

There's still a possibility for the base color not to exist beforehand, right? Shouldn't it be calculated then when adding the media from library, perhaps the test could check for that now?

spacedmonkey: Yeah, it is possible for the base color not to exist. But the behaviour has now changed. If an media is inserted without base color, then it is generated async and all elements using that resource are updated. We are trying to avoid generating base color on drop, as it is bad for performance.

spacedmonkey:
Should we force baseColor be generated on upload? At the moment, base color is only asynchronously backfilled. However I could force base color generation after the poster is generated. This would slow down the upload process.

packages/story-editor/src/components/dropTargets/karma/dropTarget.karma.js
miina: > We are trying to avoid generating base color on drop, as it is bad for performance.

Yep, I know that, I'm basically asking if it's still possible to test the current new logic in a Karma test as well. To ensure that if the base color does not exist, it's eventually added.

swissspidy:
> Should we force baseColor be generated on upload? At the moment, base color is only asynchronously backfilled. However I could force base color generation after the poster is generated. This would slow down the upload process.

What's the benefit? When we now have this nice async backfill process.

If there's a big advantage, let's discuss in a follow-up ticket.

spacedmonkey:
> If there's a big advantage, let's discuss in a follow-up ticket.

Follow up ticket created https://github.com/google/web-stories-wp/issues/9739

packages/story-editor/src/components/dropTargets/karma/dropTarget.karma.js
spacedmonkey: I have added this test. https://github.com/google/web-stories-wp/pull/9638/commits/55a992837fd637fcd40459931cfedc50ae3c2bc9 . The fixture data does not contain color for this image. Meaning the color is successfully generated.

packages/story-editor/src/components/dropTargets/provider.js
swissspidy: This is not about color generation.

This is about copying the baseColor from one element to the other

miina (approved):

Issue #9639, opened 2021-11-05
Decoupling: Package for common utility functions


mohdsayed:
## Feature Description
Sometimes there is a need to share some utility functions between packages, for instance, `snakeToCamelCaseObjectKeys` is required to be used in both `wp-*` integration packages for making the API responses of the API callbacks camel case. Currently, the available options are to either duplicate the code or move it to `@web-stories-wp/design-system`. 

## Additional Context

https://github.com/google/web-stories-wp/issues/2918

swissspidy:
If this is only gonna be used for the `wp-*` packages, a private `wp-shared` or `wp-utils` package or something shouldn't be a big deal.

mohdsayed:
Circling back to this ticket after the progress made on https://github.com/GoogleForCreators/web-stories-wp/pull/10452 . There were a few more utility functions dumped in `design-system` only because they were commonly used in more than one package however do not belong to `design-system`. Maybe we can move at least some generic static functions like ( existing and new ) in a utility package?

```js
1. isValidUrl
2. withProtocol
3. stripHTML
4. getKeyboardMovement
5. calcRotatedResizeOffset
6. objectWithout
7. deepMerge
8. domReady
9. uniqueEntriesByKey
10. addQueryArgs
11. noop
12. isNullOrUndefinedOrEmptyString
```

I am expecting more utility functions looking for proper home as we break `story-editor` down into more packages.

swissspidy:
Thanks for this list, very helpful 👍 

For the public packages, if we create a new utility package that's just another dumping ground without cleanly organizing the code. So that would just move a problem around.

> isValidUrl
> withProtocol
> addQueryArgs

There is also `toAbsoluteUrl`

-> a new `url` package?

> stripHTML

There is also `escapeHTML`

-> a new `dom` package?

> domReady

This is only used for `wp-story-editor` and `wp-dashboard`, right? If so, we could:

a) Use the `@wordpress/dom-ready` package there instead. We already use `@wordpress/api-fetch` and this is not a published package.
b) Alternatively, just copy this function to both packages. It's only like 5 lines of code.

> isNullOrUndefinedOrEmptyString

On `main` this is only used within `design-system`, so I think we can just leave it there.

> noop

This is like 20 chars. Every package can just define its own `noop` function for reuse. No need to move that to a shared package.

> getKeyboardMovement

[...]

PR #9644, opened 2021-11-05
Decoupling: Setup storybook playground for dashboard

mohdsayed:
## Context

Since the dashboard is now decoupled, there should be a storybook playground for the dashboard too.

## Summary

image

---

Explore Templates

image

## Testing Instructions

1.

## Checklist

---

Related: https://github.com/google/web-stories-wp/pull/9569


swissspidy:
Random thought: as a **stretch goal**, would it be possible for the dashboard playground to link to the editor playground when clicking on "Create New Story" for example?

Addons like https://storybook.js.org/addons/@storybook/addon-links might allow for this

mohdsayed:
> Random thought: as a stretch goal, would it be possible for the dashboard playground to link to the editor playground when clicking on "Create New Story" for example?

@swissspidy Already doing it by passing the following, but without the use of the addon, and it works.

https://github.com/google/web-stories-wp/blob/c8a33866175a5a1ce56943582a57b2793888860e/.storybook/stories/playground/dashboard/index.js#L186

.storybook/stories/playground/story-editor/header/buttons/preview.js
```diff
@@ -41,7 +46,9 @@ function PreviewButton() {
   const openPreviewLink = () => {
     saveStory();
 
-    const previewLink = `${window.location.origin}/iframe.html?id=playground-preview--default`;
+    const previewLink = getStorybookUrl(
+      'iframe.html?id=playground-preview--default'
+    );
```
swissspidy: Something `@storybook/addon-links` could be used for?

packages/dashboard/src/utils/index.js
```diff
@@ -23,3 +23,4 @@ export {
 } from './usePagePreviewSize';
 export { default as useStoryView } from './useStoryView';
 export { default as useTemplateView } from './useTemplateView';
+export { default as deepMerge } from './deepMerge';
```
swissspidy: Maybe something for design system since it's used by story editor as well?

mohdsayed: Ah okay, I thought you prefer to duplicate it until we decide something in https://github.com/google/web-stories-wp/issues/9639. Will move to design system

.storybook/stories/playground/story-editor/header/buttons/preview.js
mohdsayed: Okay, trying...

packages/dashboard/src/utils/index.js
swissspidy: #9639 is for utils shared by the `wp-*` packages, no? Here we are in the core dashboard, not `wp-dashboard`

mohdsayed: Yes, correct.

.storybook/stories/playground/story-editor/header/buttons/preview.js
mohdsayed: It should be fine here for the preview, but the only problem is that we need to deal with the promise for updating `newStoryURL` in `config` using `hrefTo`, which is currently static and easy to pass. 

https://github.com/google/web-stories-wp/blob/c51869d083ff38a5f09240d18a7eb8898c6d742f/.storybook/stories/playground/dashboard/index.js#L126-L128

mohdsayed: For preview, it is easy though 

```js
const previewLink = await hrefTo('Playground/preview', 'default');
```

mohdsayed: But yeah `@storybook/addon-links` looks like a better choice, will think again.

swissspidy: It certainly feels less hardcoded and more future proof should storybook ever change the URL structure.

packages/dashboard/src/utils/index.js
mohdsayed: Moved in https://github.com/google/web-stories-wp/pull/9644/commits/8eb755628f1ca37180294374a63c9732ce6b1b5e

.storybook/stories/playground/story-editor/header/buttons/preview.js
mohdsayed: If we follow thier API we would need to call `linkTo` method using `onClick` for which we will have to make changes to the core dashboard.

So to keep things simple, I used the `hrefTo` from `@storybook/addon-links` for preview but  created a better utility function called  `linkHrefTo` using `toId` from `'@storybook/csf'` which is internally being used by `@storybook/addon-links`  to convert string to id.

Updated in https://github.com/google/web-stories-wp/pull/9644/commits/a79794fe060b7a3806d08bbd5338128be5f11d78#

swissspidy (approved):

PR #9651, opened 2021-11-08
UI Simplification: Saved Styles

### LINKED ISSUE #9463: UI Simplifications: Saved Styles

miina:
## Summary
Removes the Saved Styles as a separate panel and moves it under Text Style panel, adding a popup.

## To-do

- [x] Fix tests
- [x] Add Confirmation dialogue for deleting a style.

## User-facing changes
| Before | After |
| ------ | ----- |
| Screenshot 2021-11-08 at 19 15 59 | Screenshot 2021-11-08 at 19 13 40 |
| Screenshot 2021-11-08 at 19 16 05 | Screenshot 2021-11-08 at 19 13 47 |

## Testing Instructions

1. Add a text element
2. Verify the Saved Styles panel is not a separate panel anymore and is part of the Text Style panel
3. Verify that it's possible to add new styles by clicking on the + icon (add more than 2 styles).
4. Verify the panel only displays up to 2 styles in the design panel.
5. Verify that clicking on "More styles" opens a popup including all the added styles.
6. Verify that it's possible to apply styles by clicking on the style in the popup and in the panel
7. Verify that it's possible to delete styles in the popup by entering Edit mode (click on the Pencil icon) and then clicking on the style.

No

No

No

## Checklist

---

Fixes #9463


packages/story-editor/src/app/rightClickMenu/provider.js
```diff
@@ -41,13 +41,14 @@ import { useStory } from '..';
 import { createPage, duplicatePage, ELEMENT_TYPES } from '../../elements';
```
miina, line 42: @barklund FYI when starting to review this -- it has a lot of changed files but most of this is cleaning up after removing the Saved Styles panel as it was before so it's not that much really 😛 

About the new `StyleManager` component -- it does have some similarities with the ColorPicker but didn't seem enough to try to create a new component for it, seemed more trouble than benefit.

packages/story-editor/src/components/panels/design/textStyle/stylePresets/styleManager.js
```diff
@@ -0,0 +1,154 @@
...
+          isEditMode={isEditMode}
+          setIsEditMode={setIsEditMode}
+        >
+          <StyledText>{__('Saved Styles', 'web-stories')}</StyledText>
+        </Header>
+        <Body>
+          <StyleGroup
+            styles={styles}
+            handleClick={handleClick}
+            isEditMode={isEditMode}
```
barklund, line 129: I don't really like this edit mode here. It just doesn't look good and a bit confusing IMHO (the small X's in the middle of the button over the text):

Screen Shot 2021-11-08 at 15 47 48

Maybe we should add the small X to the corner of each button instead? Or completely remove the edit button altogether and have a trash can icon in the corner of each style button on hover? Then you don't explicitly enter edit mode, but can at anytime delete any of the presets directly from hover (because the button is so big, there's room for a "sub button" inside).

I know it's a UX thing, but I feel like the current result is so bad UX, that we should re-address it.

packages/story-editor/src/components/panels/design/textStyle/stylePresets/stylePresets.js
```diff
@@ -0,0 +1,188 @@
...
+            placement={PLACEMENT.LEFT_START}
+            spacing={spacing}
+            renderContents={() => (
+              <StyleManager
+                styles={textStyles}
+                applyStyle={handlePresetClick}
+                onClose={() => setIsPopupOpen(false)}
+              />
+            )}
+          />
```
barklund: There's a slight issue with the positioning of this popup, but I guess it's a general issue.

Because the popup animates in by sliding down, the script initially thinks the popup is placed a bit higher than it will be once animated in. This results in the popup dipping lower on mount than it would on hover resulting in a slight jump if you mount low in the viewport and then scroll the design panel. It's a minor UX nit though, probably best addressed in a separate PR:

sidebar-position-offset

packages/story-editor/src/components/panels/design/textStyle/stylePresets/stylePresets.js
```diff
@@ -0,0 +1,188 @@
...
+          aria-label={__('Add style', 'web-stories')}
+        >
+          <Icons.Plus />
+        </StyledButton>
+      </PresetsHeader>
+      {hasPresets && (
+        <>
+          <StylesWrapper>
+            <StyleGroup
+              styles={textStyles.slice(-2)}
```
barklund: I don't feel like this ordering works well. I think we should reverse the order of the text styles displayed in both lists (both panel and flyout) and then just slice the top 2 here instead:

Screen Shot 2021-11-08 at 15 48 10

Currently, the two displayed are the bottom two in the flyout, but it seems a bit counterintuitive to spot that if the total number is odd.

If we reverse the list, it will effectively be "newest first" which I feel makes more sense.

packages/story-editor/src/components/panels/design/textStyle/stylePresets/stylePresets.js
```diff
@@ -0,0 +1,188 @@
...
+  const handlePresetClick = (preset) => {
+    handleApplyStyle(preset);
+  };
+
+  const spacing = useMemo(() => ({ x: 20 }), []);
+  return (
+    <>
+      <PresetsHeader>
+        <SubHeading size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL}>
+          {__('Recent Saved Styles', 'web-stories')}
```
barklund: Isn't this bad English? Shouldn't it be "Recently Saved Styles"?

packages/story-editor/src/components/panels/design/textStyle/stylePresets/styleGroup.js
```diff
@@ -55,7 +59,7 @@ function StyleGroup({ styles, isEditMode, handleClick }) {
   return (
     <Group ref={groupRef}>
       {styles.map((style, i) => (
-        <ButtonWrapper key={JSON.stringify(style)}>
+        <ButtonWrapper key={JSON.stringify(style)} width={buttonWidth}>
           <StyleItem
```
barklund, line 63: Just now noticed there's an issue with the letter-spacing displayed in the style item preview - letter spacing is greatly exaggerated. Here it's set to 3% and only showing a slight letter-spacing on canvas, but a huge spacing in the style panel preview (compare red and blue circled text previews):

Screen Shot 2021-11-08 at 16 07 08

packages/story-editor/src/components/panels/design/textStyle/stylePresets/stylePresets.js
```diff
@@ -0,0 +1,188 @@
...
+  const hasPresets = textStyles.length > 0;
+
+  const handleApplyStyle = useApplyStyle({ pushUpdate });
+  const { addGlobalPreset } = useAddPreset({ presetType: PRESET_TYPES.STYLE });
+
+  const handlePresetClick = (preset) => {
+    handleApplyStyle(preset);
+  };
+
+  const spacing = useMemo(() => ({ x: 20 }), []);
```
barklund: Does it make a difference to memoize this? Is the popup memoized thus requiring the same argument (same object) every time? Since it's static anyway, you could just move it to a `const` outside the component.

packages/story-editor/src/components/panels/design/textStyle/stylePresets/stylePresets.js
```diff
@@ -0,0 +1,188 @@
...
+            )}
+          />
+        </>
+      )}
+      {!hasPresets && (
+        <NoStylesWrapper>
+          <NoStylesText size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL}>
+            {__('No Styles Saved', 'web-stories')}
+          </NoStylesText>
+        </NoStylesWrapper>
```
barklund, line 178: Is this in Figma? It looks a bit wrong - like there's a hole in the panel because of the difference in BG color:

Screen Shot 2021-11-08 at 16 12 32

packages/story-editor/src/karma/fixture/containers/designPanel/textStyle.js
```diff
@@ -103,7 +104,59 @@ export class TextStyle extends AbstractPanel {
...
+    super(node, path);
+  }
+
+  get apply() {
+    return this.getByRole('button', { name: /Apply style/ });
+  }
+
+  get presets() {
+    return this.getAllByRole('button', { name: /Apply style/ });
+  }
```
barklund, line 149: These can't both exist, or? Won't `getByRole` throw if there's more than one? Or do you make sure to only access `.apply` if there's **only one** preset?

barklund: Oh, I see it's the same in the main panel - and was in the old panel as well. I guess you just have to be careful which one you call.

packages/story-editor/src/app/rightClickMenu/provider.js
barklund: I don't mind this - they're significantly different that refactoring would be more work than gain. And I actually recommend removing the whole edit mode (in a different comment), which would make them even more different.

barklund (approved):
This works great and looks really good and sharp. Just a few comments, but most of it is actually UX 🤷

packages/story-editor/src/components/panels/design/textStyle/stylePresets/styleGroup.js
miina: Good catch, looks like it was used just as `px` in the style but as `em` in the workspace.

packages/story-editor/src/components/panels/design/textStyle/stylePresets/styleManager.js
miina: Reached out to Amy about this -- I'll probably merge as-is for now not to block the PR (since that's how it was before as well) but will follow up in a separate issue as soon as we have feedback from Amy.

packages/story-editor/src/components/panels/design/textStyle/stylePresets/stylePresets.js
miina: [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=10250%3A436112). The Alignment panel also has this background. But yep, it's true that it doesn't feel like it's part of the Recent Saved Styles and feels like a separate area in the panel.

cc @agingoldseco for thoughts.

Posting it here with more context:
Screenshot 2021-11-09 at 12 37 41

packages/story-editor/src/components/panels/design/textStyle/stylePresets/stylePresets.js
miina: Hm, I'm not actually seeing it locally. Does it happen with the ColorPicker as well for you?

barklund: Yes, I see the same thing when opening the color picker. And I see the same thing in the color picker [on the staging environment](https://stories-new-wordpress-amp.pantheonsite.io/wp-admin/post-new.php?post_type=web-story).

miina: Ok, I'm using 13'' screen currently and not seeing it, perhaps the screen size plays a role -- will create an issue based on your feedback.

miina: Added an issue: https://github.com/google/web-stories-wp/issues/9666

PR #9653, opened 2021-11-08
Prevent history pollution caused by media backfilling

swissspidy:
## Context

While testing #9638 I noticed that any existing media backfilling logic currently causes history pollution because `updateElementsByResourceId` updates state even if there are no elements in a story with the given resource ID.

## Summary

This PR makes `updateElementsByResourceId` more robust thanks to a short-circuit if there are no elements found with the given resource ID.

This prevents history pollution caused by media backfill logic.

## Relevant Technical Choices

* Updates `updateElementsByResourceId` to do nothing if there's no element with the given resource ID
* Updates `deleteElementsByResourceId` to do nothing if there's no element with the given resource ID
* Updates tests for all other reducers to be more strict to prevent future regressions

## To-do

The `updateElements` reducer could be made more robust in the future.

Right now, the following test fails when using `toBe()`:

https://github.com/google/web-stories-wp/blob/d4bb3eba1a09542ae91002dcfdaed4a5cdd0c4ba/packages/story-editor/src/app/story/useStoryReducer/test/updateElementById.js#L80-L95

## User-facing changes

None

## Testing Instructions

1. Go to Media -> Add New in WordPress and upload a video
2. Go to the story editor and observe the Network tab
3. See that a video poster for the new video has been generated and uploaded
4. See that the undo button below the canvas has become active.

No

No

No

## Checklist

---

Fixes #

### DIFF
--- a/packages/story-editor/src/app/story/useStoryReducer/reducers/deleteElementsByResourceId.js
+++ b/packages/story-editor/src/app/story/useStoryReducer/reducers/deleteElementsByResourceId.js
@@ -21,6 +21,8 @@
  *
  * If an empty id or a no matches with id, state is unchanged.
  *
+ * If no element with the given resource id is found, state is changed.
+ *
  * @param {Object} state Current state
  * @param {Object} payload Action payload
  * @param {string|null} payload.id id Delete all elements with this resource id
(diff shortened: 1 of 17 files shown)


spacedmonkey (approved):
SGTM

PR #9656, opened 2021-11-08
Dashboard: Next and Previous page buttons should retain focus

### LINKED ISSUE #9537: Dashboard: Template Preview Focus Resets
When you're using a keyboard to navigate the single template views in the dashboard the focus resets in a less than ideal way. Morten outlined it as such: 
`When using the keyboard to navigate between templates on the single template view (the right-most or left-most arrow buttons), focus is reset to the close button. Focus should remain on the same button. Focus should only move to the close button the first time the single template page is loaded, not every time the single page template changes to a different template` 

which is pretty on spot. We can improve this.

samwhale:
## Summary

Clicking or using the keyboard to switch between templates should keep the focus on the button.

## Relevant Technical Choices

Removed code that would blur the buttons on click. This was removing focus from the buttons and focusing the body instead. However, this keeps the buttons in the `:active` state.

## To-do

n/a

## User-facing changes

|Before|After|
|--|--|
|old|new|

## Testing Instructions

1. View template details by selecting the `See details` button for a template from the `Explore Templates` page
2. Click the `previous` and `next` buttons. The should keep focus
3. Use your keyboard to hit the same buttons. The buttons should retain focus.
4. Clicking the `See Details` buttons for related templates should scroll the page to the top and focus the `close` button at the top of the page still.

no

no

no

## Checklist

---

Fixes #9537

### DIFF
--- a/packages/dashboard/src/app/views/templateDetails/content/detailsGallery/index.js
+++ b/packages/dashboard/src/app/views/templateDetails/content/detailsGallery/index.js
@@ -106,9 +106,7 @@ function DetailsGallery({
         size={BUTTON_SIZES.SMALL}
         variant={BUTTON_VARIANTS.CIRCLE}
         aria-label={__('View previous template', 'web-stories')}
-        onClick={({ currentTarget }) => {
-          // blurring target here because memoized button remains active on click
-          currentTarget.blur();
+        onClick={() => {
           switchToTemplateByOffset(-1);
         }}
         disabled={!orderedTemplatesLength || activeTemplateIndex === 0}
@@ -123,9 +121,7 @@ function DetailsGallery({
         size={BUTTON_SIZES.SMALL}
         variant={BUTTON_VARIANTS.CIRCLE}
         aria-label={__('View next template', 'web-stories')}
-        onClick={({ currentTarget }) => {
-          // blurring target here because memoized button remains active on click
-          currentTarget.blur();
+        onClick={() => {
           switchToTemplateByOffset(1);
         }}
         disabled={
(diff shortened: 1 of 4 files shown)


packages/dashboard/src/app/views/templateDetails/content/detailsGallery/index.js
```diff
@@ -106,9 +106,7 @@ function DetailsGallery({
         size={BUTTON_SIZES.SMALL}
         variant={BUTTON_VARIANTS.CIRCLE}
         aria-label={__('View previous template', 'web-stories')}
-        onClick={({ currentTarget }) => {
-          // blurring target here because memoized button remains active on click
-          currentTarget.blur();
```
samwhale, line 111: Buttons would still lose focus when clicked. Decided to remove this, even though the button will remain active.

packages/dashboard/src/app/views/templateDetails/content/index.js
```diff
@@ -42,19 +41,6 @@ function Content({
...
-  const {
-    actions: { scrollToTop },
-  } = useLayoutContext();
-
-  useEffect(() => {
-    if (template !== null && template?.id !== previousTemplateId.current) {
-      scrollToTop();
-      previousTemplateId.current = template.id;
-    }
-  }, [template, scrollToTop]);
```
samwhale, line 56: This runs every time the template was changed. Now `scrollToTop` is only called if you click `see details` on a related template

maxyinger, line 56: ❤️ beautiful. good call on the slight behavior update here too to only scroll on grid click. nice touch.

maxyinger (approved):

BrittanyIRL (approved):
thank you, this is so much better. this page has changed several times and i feel like this PR does a good job of gutting the old that used to make more sense and putting in place something that makes way more sense. Thanks for getting to the core of the issue.

BrookeGraham (approved):
🕺🏻

Issue #6581, opened 2021-03-02
WordPress menu overlay new tooltip


spacedmonkey:
## Bug Description

Found here https://github.com/google/web-stories-wp/pull/6566#issuecomment-788133378 . Tooltip should not be overlay by the WordPress menu like this. See screenshot. 

## Expected Behaviour

Tooltip should not be hidden behind menu. 

## Screenshots

![image](https://user-images.githubusercontent.com/237508/109672720-3376af80-7b6d-11eb-974c-3f37d7fd6a94.png)

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

barklund:
It's actually not because it's hidden behind the menu, it's because we hide all overflow from the editor itself (to avoid other issues).

The simplest fix here would be to simply left-aligned the tooltip, so it stays within the bounds of the editor.

We do the same thing with the tooltip for the local WordPress media tab above - simply left-aligned to avoid this hidden overflow issue:

![Screen Shot 2021-05-10 at 13 43 52](https://user-images.githubusercontent.com/637548/117701911-ccd9c780-b195-11eb-9c85-18383c97e586.png)

Would that be an acceptable solution?

swissspidy:
Sounds acceptable to me. Just need to make sure the alignment is flipped for RTL.

spacedmonkey:
I think this a simple and workable solution.

BrittanyIRL:
I think  #9668 fixes this issue.

<img width="265" alt="Screen Shot 2021-11-09 at 1 58 04 PM" src="https://user-images.githubusercontent.com/10720454/141004631-cd53e82f-adec-4b48-a563-52fd94196ed3.png">
<img width="279" alt="Screen Shot 2021-11-09 at 2 06 28 PM" src="https://user-images.githubusercontent.com/10720454/141004695-5558a83a-1b1c-4ff0-8208-50aa69296d4e.png">

miina:
Closing as this seems to be fixed.

PR #9668, opened 2021-11-09
Design System: Improve tooltips position in RTL

### LINKED ISSUE #9598: Tooltips: Watch position while open
A while back @BrookeGraham fixed an issue that was newly exposed with dropdowns where they should switch positions while expanded because their position on the page changed and obscured the menu. That was here: #9060

We should do the same for tooltips. @miina pointed out here: https://github.com/google/web-stories-wp/issues/9460#issuecomment-956072696 that there's a few tooltips that are encountering the same problem.

Acceptance criteria: 
- If a tooltip, while open, changes position and is obscured by the end of the browser window or app space it should automatically move positions so that it is not obscured. Places this happens: `toggle consistent corner radius`

BrookeGraham:
## Context

Tooltip placement is normally repositioned _if_ the tip will be cutoff by the edge of the window. However, in the edge case where a tooltip is on a toggle that reveals a menu, the tooltip can cover newly shown buttons, lists, etc. As is the case with the `toggle consistent corner radius` tip on the corner radius toggle button. 

## Summary

When the tooltip toggle button is switched from one state to another, we are remounting the tooltip. This will reposition the tooltip's anchor. It will also close the tooltip even though the button, ie. tooltip anchor is still in focus. 

While digging into the tooltip used within the `radius` toggle button I stumbled across a couple issues. ~~First, the `Radius` component, as well and the `SizePosition` component were not using the `Tooltip` from `Design System`. By importing it directly, those components were not receiving the RTL mapping for `placement`.~~ [see comment thread](https://github.com/google/web-stories-wp/pull/9668#discussion_r746758391) Secondly,  in RTL mode the tooltips are cutoff on the left side of the screen. There is an offset where if a tooltip is cut off on the right edge of the window we offset the anchor point. This forces it to render up against the edge of the window. However, this doesn't apply in RTL. 

## Relevant Technical Choices

Since we use quite a few tooltips, tracking the anchor's location on the screen and repositioning it every time they change seemed a bit overkill for this edge case. Remounting seemed the smart choice.

Added a switch case in the if statement within `Tooltip`'s `positionPlacement` callback to solve the cutoff in RTL mode. If the popup would normally be rendered outside of the window, we are updating the placement to `{placement}-end`. 

### AFTER

https://user-images.githubusercontent.com/1820266/140964557-239adcd8-4dbe-455c-b7b7-bb128defc068.mov

https://user-images.githubusercontent.com/1820266/140964548-cbac9c6f-19dd-4a8c-b50b-7f6bb226b2e8.mov

https://user-images.githubusercontent.com/1820266/140964532-1f0edb79-98ae-489c-b37b-7dc450c2767a.mov

https://user-images.githubusercontent.com/1820266/140964543-1c3ac0e8-0177-4e5e-8930-c19c1118d4cc.mov

### BEFORE

https://user-images.githubusercontent.com/1820266/140965561-e4075235-1844-48f3-9b5c-206136a7c0c2.mov

https://user-images.githubusercontent.com/1820266/140965571-865783bd-ab8e-4b1e-b5bc-c5e49e60ed7d.mov

https://user-images.githubusercontent.com/1820266/140965578-e343cd56-49ab-4bf8-bc74-69908b490346.mov

## Testing Instructions

1. Select an image on the canvas.
2. In the `Design` tab select the `Selection` dropdown / panel
3. Hover on the Lock Icon for `Radius` - you should see the tooltip 
4. Click on the lock icon. This should reveal/hide a menu with the different radius options. During reveal/hide the tooltip should no longer show until the next time you hover (or select icon via keyboard)
5. Check RTL mode!!
6. In RTL mode the same tooltip for the `radius` icon should no longer be cutoff by the edge of the screen. This goes for all tooltips on the far left ( found by hover on `radius` lock icon, `aspect ratio` lock icon, and `align bottom`) 
7. Lastly, check bottom cutoff still works in both RTL and LTR 
8. Open `Document` tab and go to `Background Audio` panel
9. Adjust window so that the trash can icon is just in view
10. Hover the trash can, tooltip should show above the icon, and not be cutoff the edge of the screen. 

no

no

no

## Checklist

---

Fixes #9598

### DIFF
--- a/packages/design-system/src/components/tooltip/index.js
+++ b/packages/design-system/src/components/tooltip/index.js
@@ -128,6 +128,21 @@ function Tooltip({
     }),
     [placement]
   );
+  // We can sometimes render a tooltip too far to the left, ie. in RTL mode.
+  // when that is the case, we can switch to placement-start and the tooltip will no longer get cutoff.
+  const updatePlacement = useCallback(() => {
+    const currentPlacement = placementRef.current;
+    switch (currentPlacement) {
+      case currentPlacement.endsWith('start'):
+        setDynamicPlacement(currentPlacement.replace('-start', '-end'));
+        break;
+      case currentPlacement.endsWith('end'):
+        setDynamicPlacement(currentPlacement.replace('-end', '-start'));
+        break;
+      default:
+        setDynamicPlacement(`${currentPlacement}-start`);
+    }
+  }, []);
 
   // When near the edge of the viewport we want to force the tooltip to a new placement as to not
   // cutoff the contents of the tooltip.
@@ -138,8 +153,8 @@ function Tooltip({
       const shouldMoveToTop =
         dynamicPlacement.startsWith('bottom') &&
         neededVerticalSpace >= window.innerHeight;
-
-      // in RTL mode, we can sometimes render a tooltip too far to the left
+      // check that the tooltip isn't cutoff on the left edge of the screen.
+      // right-cutoff is already taken care of with `getOffset`
       const isOverFlowingLeft = offset.popupLeft < 0;
 
       if (shouldMoveToTop && !isOverFlowingLeft) {
@@ -147,20 +162,10 @@ function Tooltip({
       } else if (shouldMoveToTop && isOverFlowingLeft) {
         setDynamicPlacement(PLACEMENT.TOP_START);
       } else if (!shouldMoveToTop && isOverFlowingLeft) {
-        const currentPlacement = placementRef.current;
-        switch (currentPlacement) {
-          case currentPlacement.endsWith('start'):
-            setDynamicPlacement(currentPlacement.replace('-start', '-end'));
-            break;
-          case currentPlacement.endsWith('end'):
-            setDynamicPlacement(currentPlacement.replace('-end', '-start'));
-            break;
-          default:
-            setDynamicPlacement(`${currentPlacement}-start`);
-        }
+        updatePlacement();
       }
     },
-    [dynamicPlacement]
+    [dynamicPlacement, updatePlacement]
   );
 
   const positionArrow = useCallback(
--- a/packages/design-system/src/components/tooltip/index.js
+++ b/packages/design-system/src/components/tooltip/index.js
@@ -129,29 +129,30 @@ function Tooltip({
     [placement]
   );
 
-  // When near the bottom of the viewport and the tooltip is placed on the bottom we want to force the tooltip to the top as to not
+  // When near the edge of the viewport we want to force the tooltip to a new placement as to not
   // cutoff the contents of the tooltip.
   const positionPlacement = useCallback(
     ({ offset }) => {
-      // check to see if there's an overlap with the window edge
+      // check to see if there's an overlap with the window's bottom edge
       const neededVerticalSpace = offset.bottom;
-      // if the tooltip was assigned as bottom we want to always check it
-      if (
-        offset &&
+      const shouldMoveToTop =
         dynamicPlacement.startsWith('bottom') &&
-        neededVerticalSpace >= window.innerHeight
-      ) {
-        setDynamicPlacement(PLACEMENT.TOP);
-      }
+        neededVerticalSpace >= window.innerHeight;
+
       // in RTL mode, we can sometimes render a tooltip too far to the left
-      const isOverFlowingX = offset.popupLeft < 0;
-      if (isOverFlowingX) {
-        switch (dynamicPlacement) {
-          case dynamicPlacement.endsWith('start'):
-            setDynamicPlacement(`${placement}-end`);
+      const isOverFlowingLeft = offset.popupLeft < 0;
+
+      if (shouldMoveToTop && !isOverFlowingLeft) {
+        setDynamicPlacement(PLACEMENT.TOP);
+      } else if (shouldMoveToTop && isOverFlowingLeft) {
+        setDynamicPlacement(PLACEMENT.TOP_START);
+      } else if (!shouldMoveToTop && isOverFlowingLeft) {
+        switch (placement) {
+          case placement.endsWith('start'):
+            setDynamicPlacement(placement.replace('-start', '-end'));
             break;
           case dynamicPlacement.endsWith('end'):
-            setDynamicPlacement(`${placement}-start`);
+            setDynamicPlacement(placement.replace('-end', '-start'));
             break;
           default:
             setDynamicPlacement(`${placement}-start`);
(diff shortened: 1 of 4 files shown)


packages/design-system/src/components/tooltip/index.js
```diff
@@ -129,19 +129,35 @@ function Tooltip({
...
+        switch (currentPlacement) {
+          case currentPlacement.endsWith('start'):
+            setDynamicPlacement(currentPlacement.replace('-start', '-end'));
+            break;
+          case currentPlacement.endsWith('end'):
+            setDynamicPlacement(currentPlacement.replace('-end', '-start'));
+            break;
+          default:
+            setDynamicPlacement(`${currentPlacement}-start`);
+        }
```
BrittanyIRL: Just for readability, thoughts on pulling this out into its own function? I'm torn between thinking it'd help read or not, mostly it just feels weird to have a switch nested in an if statement.

packages/story-editor/src/components/panels/design/sizePosition/radius.js
```diff
@@ -207,7 +206,10 @@ function RadiusControls({ selectedElements, pushUpdateForObject }) {
       </StackableGroup>
 
       <LockContainer>
-        <Tooltip title={__('Toggle consistent corner radius', 'web-stories')}>
+        <Tooltip
+          title={__('Toggle consistent corner radius', 'web-stories')}
+          key={borderRadius.locked}
```
BrittanyIRL: 🤔  if borderRadius.locked is a bool, should this maybe be something like 
```
`radius_locked_tooltip_${borderRadius.locked.toString()}`
``` 
so it still retriggers the tooltip render but doesn't pass a bool as a key?

BrittanyIRL (approved):
Looks awesome. Great notes and the solution makes sense and is really robust. 

This also fixes this issue: https://github.com/google/web-stories-wp/issues/6581!

packages/story-editor/src/components/panels/design/sizePosition/radius.js
BrookeGraham: GP!!

packages/design-system/src/components/tooltip/index.js
BrookeGraham: Agreed. It is super weird. At one point I did pull it out, but ended up refactoring it back in, though I had a bit of other stuff going on then. I think I will pull it out and maybe add a little more comments since it is hard to follow.

packages/story-editor/src/components/panels/design/sizePosition/sizePosition.js
```diff
@@ -49,7 +50,6 @@ import {
   inputContainerStyleOverride,
   useCommonObjectValue,
 } from '../../shared';
-import Tooltip from '../../../tooltip';
```
miina, line 52: The editor was previously using a local `Tooltip` component which was based on the `Tooltip` from the design system, and was adjusting the tooltip placement based on if it was RTL or not (using the `TOOLTIP_RTL_PLACEMENT`).

I assume the `TOOLTIP_RTL_PLACEMENT` was not working fully as expected? Is the local `Tooltip` component obsolete now and can be fully replaced?

We're still using the local `Tooltip` in various places, if the RTL logic used in there was not working then perhaps it should be replaced with the design system `Tooltip` everywhere now? Thoughts?

BrittanyIRL: I think this PR should mean that it can be entirely replaced. I'd missed that the Tooltip was localized to the editor this spring when we pulled it out into the DS (it was copied over so it would have been the same structure 🤔 ). Anyways, _yes_ let's replace it! 

I'll make a [new ticket](https://github.com/google/web-stories-wp/issues/9684), we should branch off of this and try replacing all of them there as proof of concept before this gets merged. 

@BrookeGraham would you mind taking this on?

BrookeGraham: Yes! I was just looking at this. I will double check that we can remove and dedupe.

BrittanyIRL: thank you!

BrookeGraham: Ok. Turns out the local version of the tooltip both in `Story-Editor` and `Dashboard`, are wrapping  the Design System `Tooltip` and mapping to the RTL placement **correctly**. What isn't working is _my_ {placement}-end swap. So we don't need to create a new ticket to de-dupe. I just need to account for {placement}-end.  {placement} and {placement}-start are working as expected in both RTL and LTR.

BrookeGraham: Alrighty, I am using the local version of Tooltip, thanks @miina. I have also fixed my inability to properly write a switch /case statement 🤦🏻‍♀️

packages/design-system/src/components/tooltip/index.js
```diff
@@ -128,23 +128,58 @@ function Tooltip({
...
+  // We can sometimes render a tooltip too far to the left, ie. in RTL mode.
+  // when that is the case, we can switch to placement-start and the tooltip will no longer get cutoff.
+  const updatePlacement = useCallback(() => {
+    const currentPlacement = placementRef.current;
+    switch (currentPlacement) {
+      case PLACEMENT.BOTTOM_START:
+      case PLACEMENT.TOP_START:
+      case PLACEMENT.RIGHT_START:
+        // {placement}-START shouldn't ever appear in overflow so do nothing
+        break;
```
samwhale, line 140: good to be explicit. Leaving it as the default without a comment may be confusing.

packages/story-editor/src/components/panels/design/sizePosition/radius.js
```diff
@@ -207,7 +207,10 @@ function RadiusControls({ selectedElements, pushUpdateForObject }) {
       </StackableGroup>
 
       <LockContainer>
-        <Tooltip title={__('Toggle consistent corner radius', 'web-stories')}>
+        <Tooltip
+          title={__('Toggle consistent corner radius', 'web-stories')}
+          key={borderRadius.locked.toString()}
```
samwhale, line 214: good pr description about this - super smart. Should we leave a comment here explaining what it does?

```js
// Remounts the component when `borderRadius.locked` changes
```

samwhale (approved):
excellante

packages/design-system/src/components/tooltip/index.js
BrookeGraham: Yeh, I felt like if I didn't leave a comment others would be confused why they weren't accounted for.

maxyinger (approved):
Looks great. Thanks for the helpful comments!

PR #9669, opened 2021-11-09
3P Media: Update label for search results

### LINKED ISSUE #3903: More appropriate label than "Trending" for search results
## Feature Description

Right now we show "Trending" above the media list when not searched at all and when performed a search. For a selected category we show its name. It might be confusing to the user to show "Trending" also for search results.

Also, when a search is made we should show something like `Results for "cat"`

BrookeGraham:
## Context

In `3pMedia` tab, it can be confusing what is being displayed after entering a search term. The subheader above the results displays `Trending` no matter if you have searched or not. 

## Summary

When searching for 3pMedia, we want display the subheading just above the results with the text `Search Results` since this more inline with what is being displayed. As opposed to showing a subheading of `Trending`. Category name still takes precedent when filtered by a category. 

## Testing Instructions

1. In 3pMedia tab, enter a search term. You should see `Search Results` instead of the `Trending` text. 
2. If you select a category to filter by, the subheading should be the Category text. ie. "Wallpapers`

no

no

no
## Checklist

---

Fixes #3903

### DIFF
--- a/packages/story-editor/src/components/library/panes/media/media3p/providerPanel.js
+++ b/packages/story-editor/src/components/library/panes/media/media3p/providerPanel.js
@@ -117,7 +117,9 @@ function ProviderPanel({ providerType, isActive, searchTerm, ...rest }) {
         data-testid={'media-subheading'}
         shouldDisplay={shouldDisplayMediaSubheading}
       >
-        {searchTerm.length ? searchResultsLabel : displayName}
+        {searchTerm.length && !state.categories.selectedCategoryId
+          ? searchResultsLabel
+          : displayName}
       </MediaSubheading>
       <PaginatedMediaGallery
         providerType={providerType}
--- a/packages/story-editor/src/components/library/panes/media/media3p/providerPanel.js
+++ b/packages/story-editor/src/components/library/panes/media/media3p/providerPanel.js
@@ -100,6 +100,9 @@ function ProviderPanel({ providerType, isActive, searchTerm, ...rest }) {
   const shouldDisplayMediaSubheading = Boolean(
     state.media?.length || state.categories.selectedCategoryId
   );
+  // When displaying search results, subHeading should be updated to reflect so
+  const searchResultsLabel = __('Search Results', 'web-stories');
+
   return (
     <ProviderWrapper {...rest}>
       {PROVIDERS[providerType].supportsCategories && (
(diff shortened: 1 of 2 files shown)


packages/story-editor/src/components/library/panes/media/media3p/test/media3pPane.js
```diff
@@ -241,6 +253,7 @@ describe('Media3pPane', () => {
   it('should render <Media3pPane /> with the category display name when selected', () => {
     useMediaResult.media3p.PROVIDER_1.state.isMediaLoaded = true;
     useMediaResult.media3p.PROVIDER_1.state.media = MEDIA;
+    useMediaResult.searchTerm = '';
```
BrookeGraham, line 256: I just copied over the patterns used in the rest of this test. Because I set `useMediaResult.searchTerm = 'cats';` I had to reset it to an empty string here. I had thought that the 
```beforeEach(() => {
    useMediaResult = DEFAULT_USE_MEDIA_RESULT;
  });
```
on line 148 would have reset it to be empty. BUT that doesn't seem to be the case.

BrookeGraham: Let me know if y'all think I should make this better. 🤷🏻‍♀️

packages/story-editor/src/components/library/panes/media/media3p/providerPanel.js
```diff
@@ -114,7 +117,7 @@ function ProviderPanel({ providerType, isActive, searchTerm, ...rest }) {
         data-testid={'media-subheading'}
         shouldDisplay={shouldDisplayMediaSubheading}
       >
-        {displayName}
+        {searchTerm.length ? searchResultsLabel : displayName}
```
BrittanyIRL: 🤝

packages/story-editor/src/components/library/panes/media/media3p/test/media3pPane.js
BrittanyIRL: think it's fine as is!

BrittanyIRL (commented):
An edge case for sure, but if i have a search term and then i select a filter (like Wallpapers) then the text should update to "Wallpapers" not remain "Search Results" since it's not cactus wallpaper it's just wallpaper.

Screen Shot 2021-11-09 at 2 12 13 PM

BrookeGraham:
> An edge case for sure, but if i have a search term and then i select a filter (like Wallpapers) then the text should update to "Wallpapers" not remain "Search Results" since it's not cactus wallpaper it's just wallpaper.

Nice catch!

samwhale (approved):
This is really dope! Super cool how I can make a search, add a filter, then remove the filter and continue showing `Search Results`.

BrittanyIRL (approved):
thanks for the update!

PR #9672, opened 2021-11-09
Page Templates: Default view set in localStorage

### LINKED ISSUE #9637: Default view for page templates panel
## Feature Description

A user would like to have the ability to change the default view for page templates to their saved page templates instead of the default templates. 

Screen Shot 2021-11-04 at 1 33 20 PM

## Additional Context

https://wordpress.org/support/topic/few-enhancement-requests/

BrookeGraham:
## Context

Users that frequently use `Saved Templates` have to toggle to them every time they open the templates layout panel. 

## Summary

When users toggle to `Saved Templates` in Template Layouts, we want to preserve the last viewed templates as the default view the next time they open the templates panel

## Relevant Technical Choices

Storing current template list in localStorage. If it is not there, we still default to the default template list. 

## Testing Instructions

1. Open `Templates Layout` in editor
2. Select `Saved Templates` 
3. Navigate to any other page, or hard refresh.
4. The next time you come to the editor and open the `Templates Layout` panel, the `Saved Templates `should appear first. 

no

no

no

## Checklist

---

Fixes #9637

### DIFF
--- a/packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
+++ b/packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
@@ -26,6 +26,10 @@ import {
 import styled from 'styled-components';
 import { __ } from '@web-stories-wp/i18n';
 import { FULLBLEED_RATIO, PAGE_RATIO } from '@web-stories-wp/units';
+import {
+  localStore,
+  LOCAL_STORAGE_PREFIX,
+} from '@web-stories-wp/design-system';
 
 /**
  * Internal dependencies
@@ -65,8 +69,9 @@ const ButtonWrapper = styled.div`
 const DEFAULT = 'default';
 const SAVED = 'saved';
 const PAGE_TEMPLATE_PANE_WIDTH = 158;
-const LOCAL_STORAGE_KEY = 'web_stories_default_template_view';
-const DEFAULT_TEMPLATE_VIEW = window.localStorage.getItem(LOCAL_STORAGE_KEY);
+const LOCAL_STORAGE_KEY =
+  LOCAL_STORAGE_PREFIX.DEFAULT_VIEW_PAGE_TEMPLATE_LAYOUT;
+const DEFAULT_TEMPLATE_VIEW = localStore.getItemByKey(LOCAL_STORAGE_KEY);
 
 function PageTemplatesPane(props) {
   const {
@@ -95,6 +100,7 @@ function PageTemplatesPane(props) {
     (page) => {
       setSavedTemplates([page, ...(savedTemplates || [])]);
       setHighlightedTemplate(page.id);
+      localStore.setItemByKey(LOCAL_STORAGE_KEY, SAVED);
     },
     [setSavedTemplates, savedTemplates]
   );
@@ -133,7 +139,7 @@ function PageTemplatesPane(props) {
 
   const handleToggle = (evt, value) => {
     setShowDefaultTemplates(value === DEFAULT);
-    window.localStorage.setItem(
+    localStore.setItemByKey(
       LOCAL_STORAGE_KEY,
       showDefaultTemplates ? SAVED : DEFAULT
     );
--- a/packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
+++ b/packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
@@ -137,8 +137,8 @@ function PageTemplatesPane(props) {
     setNextTemplatesToFetch,
   ]);
 
-  const handleToggle = (evt, value) => {
-    setShowDefaultTemplates(value === DEFAULT);
+  const handleToggle = () => {
+    setShowDefaultTemplates(!showDefaultTemplates);
     localStore.setItemByKey(LOCAL_STORAGE_KEY, !showDefaultTemplates);
   };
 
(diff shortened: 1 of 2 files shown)


packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
```diff
@@ -127,6 +131,14 @@ function PageTemplatesPane(props) {
     setNextTemplatesToFetch,
   ]);
 
+  const handleToggle = (evt, value) => {
+    setShowDefaultTemplates(value === DEFAULT);
+    window.localStorage.setItem(
```
swissspidy: Import `localStore` from `@web-stories-wp/design-system` instead and use `localStore.getItemByKey` and `localStore.getItemByKey`/`localStore.setItemByKey` to get error handling for free. 

There's also `LOCAL_STORAGE_PREFIX`

BrookeGraham: Nice, Thanks @swissspidy

BrittanyIRL (commented):
Curious what you think about storing the user's default view as a boolean to be more in line with the other user settings in local storage? Something along the lines of `web_stories_saved_page_templates_default` - I think it'd make this even more straight forward.

BrookeGraham:
> Curious what you think about storing the user's default view as a boolean to be more in line with the other user settings in local storage? Something along the lines of `web_stories_saved_page_templates_default` - I think it'd make this even more straight forward.

I had initially used a boolean, but on the first visit they won't have it set yet. So the `web_stories_saved_page_templates_default` or whatever it is called, will be `null`. We would end up needing to differentiate between `false` and `null`. I could explicitly check for `null` first and if hasn't been set yet, then set if to `true` then we wouldn't need to worry about it being `null` afterward.

BrittanyIRL:
> > Curious what you think about storing the user's default view as a boolean to be more in line with the other user settings in local storage? Something along the lines of `web_stories_saved_page_templates_default` - I think it'd make this even more straight forward.
> 
> I had initially used a boolean, but on the first visit they won't have it set yet. So the `web_stories_saved_page_templates_default` or whatever it is called, will be `null`. We would end up needing to differentiate between `false` and `null`. I could explicitly check for `null` first and if hasn't been set yet, then set if to `true` then we wouldn't need to worry about it being `null` afterward.

yeah, i feel like that is a better route to take.

packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
```diff
@@ -175,9 +190,7 @@ function PageTemplatesPane(props) {
             <Select
               options={options}
               selectedValue={showDefaultTemplates ? DEFAULT : SAVED}
-              onMenuItemClick={(evt, value) =>
-                setShowDefaultTemplates(value === DEFAULT)
-              }
+              onMenuItemClick={(evt, value) => handleToggle(evt, value)}
```
samwhale: 🤔 I usually prefer pure functions like `handleToggle` but in this case the `value` doesn't change between children or renders right? If it's pretty static - then I think it wouldn't be a problem to reference the `value` that's held in the component scope in the `handleToggle`

Further, that could save us a render because this inline arrow function is referentially different each render. Really just a nit, but this works great too.

samwhale (approved):
Noice

packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
BrookeGraham: Duh! GP. GP.  I just left it when I copied and moved up, didn't even think about it. 🙌🏽

samwhale: Not a big deal 😄

packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
```diff
@@ -83,14 +90,17 @@ function PageTemplatesPane(props) {
     setNextTemplatesToFetch: state.actions.setNextTemplatesToFetch,
   }));
 
-  const [showDefaultTemplates, setShowDefaultTemplates] = useState(true);
+  const [showDefaultTemplates, setShowDefaultTemplates] = useState(
+    DEFAULT_TEMPLATE_VIEW === null ? true : DEFAULT_TEMPLATE_VIEW
```
BrittanyIRL, line 94: 👍

BrittanyIRL (approved):
🚀

BrittanyIRL:
I just hit rerun on the e2e tests but it might be worth checking if this test needs a little love? I know it's been flakey lately. 

Screen Shot 2021-11-12 at 8 50 12 AM

BrookeGraham:
> I just hit rerun on the e2e tests but it might be worth checking if this test needs a little love? I know it's been flakey lately.

Yeah, it has failed on me a couple times for this pr. Should I make a ticket for it?

BrittanyIRL:
> > I just hit rerun on the e2e tests but it might be worth checking if this test needs a little love? I know it's been flakey lately.
> 
> Yeah, it has failed on me a couple times for this pr. Should I make a ticket for it?

Since it's just failing on this PR, can you investigate it here and make sure the test just doesn't need an update to go with your changes?

PR #9673, opened 2021-11-09
Workspace: Fix template usage e2e test and improve color picker accessibility

### LINKED ISSUE #9665: E2E: Fix flaky test: `should be able to use existing template for new story`
## Bug Description
The test `Template > should be able to use existing template for new story` is flaky and was disabled in #9617 

Screenshot 2021-11-09 at 14 40 38

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

samwhale:
## Summary

Fix flakey e2e test and update accessibility so that it passes axe tests.

## Relevant Technical Choices

n/a

## To-do

n/a

## User-facing changes

none

## Testing Instructions

1. Tests should pass
2. Open a story
3. Click some text in the story
4. Open the color picker in the design panel
5. Color swatches should still look and work the same.

n/a

n/a

## Checklist

---

Fixes #9665

### DIFF
--- a/packages/e2e-tests/src/specs/dashboard/templates/useTemplate.js
+++ b/packages/e2e-tests/src/specs/dashboard/templates/useTemplate.js
@@ -21,9 +21,7 @@ import percySnapshot from '@percy/puppeteer';
 import { visitDashboard } from '@web-stories-wp/e2e-test-utils';
 
 describe('Template', () => {
-  // TODO(#9665): Fix flakey test.
-  // eslint-disable-next-line jest/no-disabled-tests
-  it.skip('should be able to use existing template for new story', async () => {
+  it('should be able to use existing template for new story', async () => {
     await visitDashboard();
 
     const dashboardNavigation = await expect(page).toMatchElement(
@@ -60,7 +58,7 @@ describe('Template', () => {
     await page.waitForNavigation();
 
     // Wait for media elements to load before continuing.
-    await page.waitForSelector('[data-testid="mediaElement-image"]');
+    await page.waitForSelector('input[placeholder="Add title"]');
     await expect(page).toMatch('Layers');
     await expect(page).toMatchElement('input[placeholder="Add title"]');
     await expect(page).toMatchElement('[data-element-id]');
@@ -80,8 +78,8 @@ describe('Template', () => {
       text: 'Fresh',
     });
 
-    // Toggle open 'Saved Colors'
-    await expect(page).toClick('button', { text: 'Saved Colors' });
+    // Open the color picker
+    await expect(page).toClick('button[aria-label="Text color"]');
 
     // Get all saved story colors and subtract 1 button for adding other colors
     const editorSavedColors = await page.evaluate(() => {
--- a/packages/e2e-tests/src/specs/dashboard/templates/useTemplate.js
+++ b/packages/e2e-tests/src/specs/dashboard/templates/useTemplate.js
@@ -57,7 +57,7 @@ describe('Template', () => {
     await expect(page).toClick('button', { text: 'Use template' });
     await page.waitForNavigation();
 
-    // Wait for media elements to load before continuing.
+    // Wait for title input to load before continuing.
     await page.waitForSelector('input[placeholder="Add title"]');
     await expect(page).toMatch('Layers');
     await expect(page).toMatchElement('input[placeholder="Add title"]');
--- a/packages/story-editor/src/components/colorPicker/basicColorList.js
+++ b/packages/story-editor/src/components/colorPicker/basicColorList.js
@@ -155,6 +155,7 @@ function BasicColorList({
           )}
         </SwatchList>
       )}
+      {/* The `ColorAdd` button can only live in the listbox if there is at least one element with `role="option"` */}
       {!colors.length && (isLocal || isGlobal) && (
         <ColorAdd isLocal={isLocal} isGlobal={isGlobal} />
       )}
--- a/packages/story-editor/src/components/colorPicker/basicColorList.js
+++ b/packages/story-editor/src/components/colorPicker/basicColorList.js
@@ -44,24 +44,43 @@ const focusStyle = css`
     )};
 `;
 
-const SwatchList = styled.div`
-  display: flex;
-  max-width: 100%;
-  flex-wrap: wrap;
-  gap: 6px;
+const Wrapper = styled.div`
+  display: inline-block;
+`;
+
+const SwatchList = styled.div.attrs({ role: 'listbox' })`
+  width: 100%;
+  display: inline;
+`;
+
+const SwatchWrapper = styled.div.attrs({ role: 'presentation' })`
+  display: inline-block;
 `;
 
 const StyledSwatch = styled(Swatch).attrs(({ isSelected }) => ({
   role: 'option',
   'aria-selected': isSelected,
 }))`
+  display: inline-block;
+  border: 2px solid transparent;
+  height: 36px;
+  width: 36px;
+
   ${focusStyle};
 
   ${({ isSelected, theme }) =>
     isSelected &&
     css`
-      border: 2px solid ${theme.colors.border.defaultActive};
+      border-color: ${theme.colors.border.defaultActive};
     `}
+
+  ::after {
+    content: '';
+    height: 36px;
+    width: 36px;
+    display: inline-block;
+    padding: 6px;
+  }
 `;
 
 function getPatternAsString(pattern) {
@@ -73,7 +92,6 @@ function getPatternAsString(pattern) {
 }
 
 function BasicColorList({
-  'aria-labelledby': ariaLabelledBy,
   color,
   colors,
   handleClick,
@@ -98,8 +116,8 @@ function BasicColorList({
     ? __('local', 'web-stories')
     : __('global', 'web-stories');
   return (
-    <SwatchList ref={listRef} {...rest}>
-      <SwatchList role="listbox" aria-labelledby={ariaLabelledBy}>
+    <Wrapper>
+      <SwatchList ref={listRef} {...rest}>
         {colors.map((pattern, i) => {
           const isTransparentAndInvalid = !allowsOpacity && hasOpacity(pattern);
           const isGradientAndInvalid = !allowsGradient && hasGradient(pattern);
@@ -134,30 +152,31 @@ function BasicColorList({
             tabIndex = -1;
           }
           return (
-            <Tooltip key={patternAsBackground} title={tooltip}>
-              <StyledSwatch
-                onClick={() => handleClick(pattern, isLocal)}
-                pattern={pattern}
-                isSelected={isSelected}
-                isDisabled={isDisabled}
-                tabIndex={tabIndex}
-                title={title}
-              >
-                {isEditMode && <Icons.Cross />}
-              </StyledSwatch>
-            </Tooltip>
+            <SwatchWrapper key={patternAsBackground}>
+              <Tooltip title={tooltip}>
+                <StyledSwatch
+                  onClick={() => handleClick(pattern, isLocal)}
+                  pattern={pattern}
+                  isSelected={isSelected}
+                  isDisabled={isDisabled}
+                  tabIndex={tabIndex}
+                  title={title}
+                >
+                  {isEditMode && <Icons.Cross />}
+                </StyledSwatch>
+              </Tooltip>
+            </SwatchWrapper>
           );
         })}
       </SwatchList>
       {(isLocal || isGlobal) && (
         <ColorAdd isLocal={isLocal} isGlobal={isGlobal} />
       )}
-    </SwatchList>
+    </Wrapper>
   );
 }
 
 BasicColorList.propTypes = {
-  'aria-labelledby': PropTypes.string.isRequired,
   handleClick: PropTypes.func.isRequired,
   allowsOpacity: PropTypes.bool,
   allowsGradient: PropTypes.bool,
--- a/packages/story-editor/src/components/colorPicker/basicColorList.js
+++ b/packages/story-editor/src/components/colorPicker/basicColorList.js
@@ -44,66 +44,24 @@ const focusStyle = css`
     )};
 `;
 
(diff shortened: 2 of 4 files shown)


packages/e2e-tests/src/specs/dashboard/templates/useTemplate.js
```diff
@@ -80,8 +78,8 @@ describe('Template', () => {
       text: 'Fresh',
     });
 
-    // Toggle open 'Saved Colors'
-    await expect(page).toClick('button', { text: 'Saved Colors' });
+    // Open the color picker
+    await expect(page).toClick('button[aria-label="Text color"]');
```
samwhale, line 82: Color presets were moved to the color picker in #9616

packages/story-editor/src/components/colorPicker/basicColorList.js
```diff
@@ -100,54 +106,56 @@ function BasicColorList({
...
-        let tooltip = null;
-        if (isDisabled && !isEditMode) {
-          tooltip = isTransparentAndInvalid
-            ? __(
-                'Page background colors cannot have an opacity.',
-                'web-stories'
-              )
-            : __('Gradient not allowed for Text', 'web-stories');
-        }
+      <ListBox aria-labelledby={ariaLabelledBy}>
```
samwhale: The `SwatchList` contains a button that is not an option: the `Add color` button. `role="listbox"` must only contain elements with `role="option"`

packages/story-editor/src/components/colorPicker/basicColorPicker.js
```diff
@@ -187,6 +187,7 @@ function BasicColorPicker({
                 allowsGradient={allowsGradient}
                 aria-labelledby="colorpicker-story-colors-title"
                 isEditMode={isEditMode}
+                data-testid="saved-story-colors"
```
samwhale, line 190: This was removed in #9616: https://github.com/google/web-stories-wp/pull/9616/files#diff-99ae7f248c2e6cf85c3daaf92de97ce89d78267600444faf894376b86bceb276L81, but was being used by the e2e test

samwhale (commented):
summary

samwhale:
> summary

Ignore this 🤦

packages/story-editor/src/components/colorPicker/basicColorPicker.js
swissspidy: If there's a way to get the colors in the e2e test without relying on testid, that would be great too

packages/e2e-tests/src/specs/dashboard/templates/useTemplate.js
```diff
@@ -60,7 +58,7 @@ describe('Template', () => {
     await page.waitForNavigation();
 
     // Wait for media elements to load before continuing.
```
swissspidy: Now this comment is not accurate anymore

packages/e2e-tests/src/specs/dashboard/templates/useTemplate.js
swissspidy: FYI @miina

miina: Eh, looks like my fault!

miina: Eh, looks like my fault!

packages/story-editor/src/components/colorPicker/basicColorList.js
```diff
@@ -75,6 +73,7 @@ function getPatternAsString(pattern) {
 }
 
 function BasicColorList({
+  'aria-labelledby': ariaLabelledBy,
```
miina: Hm, doesn't this work as part of the `rest`?

miina: Seems to work as-is as well:

Screenshot 2021-11-10 at 11 58 02
Screenshot 2021-11-10 at 11 59 21

Thoughts?

packages/story-editor/src/components/colorPicker/basicColorList.js
```diff
@@ -100,54 +99,56 @@ function BasicColorList({
...
-        let tooltip = null;
-        if (isDisabled && !isEditMode) {
-          tooltip = isTransparentAndInvalid
-            ? __(
-                'Page background colors cannot have an opacity.',
-                'web-stories'
-              )
-            : __('Gradient not allowed for Text', 'web-stories');
-        }
+      <SwatchList role="listbox" aria-labelledby={ariaLabelledBy}>
```
swissspidy: Nested `SwatchList` is a bit confusing 🤔

miina: Indeed, this also changes the style and adds extra space that's not needed. Looks like both the `role` and the `aria-labelledby` were assigned to the `SwatchList` already before. Curious which part of the tests required adding duplicate `SwatchList` here, perhaps the test can be adjusted instead.

miina: I guess you wrapped the options in additional `SwatchList` because `listbox` can't include a `button`?
Maybe create a new wrapper component instead of the original `SwatchList` instead (or change the button to an `option`)? The original `SwatchList` should probably get the same props as before... So it would be something like
```
<Wrapper>
  <SwatchList ref={listRef} {...rest}>
  </SwatchList>
  <ColorAdd ... />
</Wrapper>
```
The `listRef` should still be on the `SwatchList`, otherwise navigation between colors with arrow keys will not work.

Feel free to create a new issue for this particular accessibility issue and assign it to me if it starts taking too much time -- doesn't look like that influences the e2e tests.

samwhale: >Curious which part of the tests required adding duplicate `SwatchList` here, perhaps the test can be adjusted instead.

The automated axe tests were failing once the test was fixed up because the `Add New Color` button was inside of the listbox. This is fine if there are other options in the listbox, but when there are no options (like in the e2e test for saved colors) then it fails the axe test because the listbox has no children with the role `group` or `option`.



> I guess you wrapped the options in additional `SwatchList` because `listbox` can't include a `button`? Maybe create a new wrapper component instead of the original `SwatchList` instead (or change the button to an `option`)? The original `SwatchList` should probably get the same props as before... So it would be something like
> 
> ```
> <Wrapper>
>   <SwatchList ref={listRef} {...rest}>
>   </SwatchList>
>   <ColorAdd ... />
> </Wrapper>
> ```
> 
> The `listRef` should still be on the `SwatchList`, otherwise navigation between colors with arrow keys will not work.

I think this is a great idea and then we wouldn't have to peel `aria-labelledby` off of the props too.

> Feel free to create a new issue for this particular accessibility issue and assign it to me if it starts taking too much time -- doesn't look like that influences the e2e tests.

The e2e test still fails without the accessibility fix, so I'd say it's still in scope to fix that here.

packages/story-editor/src/components/colorPicker/basicColorList.js
samwhale: It used to but the `listbox` was moved off of the top most wrapper: https://github.com/google/web-stories-wp/pull/9673/files#diff-eb2e914ab33bdf308ba539de8ae8c6d6fd708795611ecb60bea4efffa1233647R101-R102

But this won't be necessary with the update to use a `Wrapper` like you suggested

samwhale:
Had to update the color swatch grid so that it would be accessible but look the same as before. 

|Before|After|
|--|--|
|image|image|

Everything should be working now and accessibility is improved 👍 

Another good way to view the visual changes are here in the percy diffs: https://percy.io/google/web-stories-wp/builds/13844956/changed/783655234?browser=chrome&browser_ids=19%2C20&subcategories=approved&viewLayout=overlay&viewMode=new&width=1600&widths=1600. Very minor changes but helps to compare a vs b

packages/story-editor/src/components/colorPicker/basicColorList.js
samwhale: Thanks for the suggestion! Code has been updated.

packages/story-editor/src/components/colorPicker/basicColorPicker.js
BrittanyIRL: For context - this is my fault. I added this when we fixed the saved colors to bring in colors from a template in the dashboard. I tried a few different ways to target just the saved story colors but came up at a loss.

BrittanyIRL (approved):
Nice work around. This works well and looks good on chrome, firefox, safari for me when i pull the code and run locally. 

Think someone from prometheus should def make sure this is cool with them too just so there's no surprises when this changes.

packages/story-editor/src/components/colorPicker/basicColorList.js
```diff
@@ -44,26 +44,66 @@ const focusStyle = css`
...
 
-const SwatchList = styled.div.attrs({
-  role: 'listbox',
-})`
-  display: flex;
-  max-width: 100%;
-  flex-wrap: wrap;
-  gap: 6px;
+const Wrapper = styled.div`
+  display: inline-block;
```
miina: If we'd change this to `display: flex`, couldn't we use an actual grid for the `SwatchList` instead and not have to add the additional `SwatchWrapper`? Navigating between the swatches is still broken currently using just arrow keys, probably because of adding the extra `SwatchWrapper`.

@barklund Pinging you here for thoughts as well since you implemented the original layout of the color swatches.
Some context for the changes here:
`ColorAdd` (button for adding a new color) was added originally to the `listbox` but since it was a `button` and not an `option`, then it was incorrect for accessibility reasons. Sam is now trying to fix it by moving this one button out of the `SwatchList` but still making it group together with other swatches.

packages/story-editor/src/components/colorPicker/basicColorList.js
```diff
@@ -44,26 +44,66 @@ const focusStyle = css`
...
+
+    /* Add a grid similar to css gap in flexbox */
+    padding: 2px 3px 0 3px;
+    padding-bottom: 0;
+    /* Remove padding from swatches on left edge */
+    :nth-child(6n + 1) {
+      padding-left: 0;
+    }
+    /* Remove padding from swatches on right edge */
+    :nth-child(6n) {
```
miina: We can use `COLOR_PRESETS_PER_ROW` instead of using `6` directly anywhere where applicable.

packages/story-editor/src/components/colorPicker/basicColorList.js
```diff
@@ -44,26 +44,66 @@ const focusStyle = css`
...
-  gap: 6px;
+const Wrapper = styled.div`
+  display: inline-block;
+`;
+
+const SwatchList = styled.div.attrs({ role: 'listbox' })`
+  width: 100%;
+  display: inline;
+
+  > div {
```
miina: Could we use the specific element here, e.g.
`> ${SwatchWrapper}` in this case?

Could all this be assigned to the `const SwatchWrapper` directly as well (in case we need to use it)?

packages/story-editor/src/components/colorPicker/basicColorList.js
```diff
@@ -44,26 +44,66 @@ const focusStyle = css`
     )};
 `;
 
-const SwatchList = styled.div.attrs({
-  role: 'listbox',
```
miina, line 48: > This is fine if there are other options in the listbox, but when there are no options (like in the e2e test for saved colors) then it fails the axe test because the listbox has no children with the role group or option.

I think I asked it somewhere already but can't find exactly where / what was the answer -- couldn't we just change the role of the Add button to `option` and leave everything else as it was?

Alternatively, could we display the `listbox` + the Add button inside it only if there are colors, and display just the one button without a listbox if there aren't any? Not sure if this would be ok from accessibility point of view (cc @barklund)

packages/story-editor/src/components/colorPicker/basicColorList.js
samwhale: @miina as far as I can tell the swatch navigation is broken in staging too - I don't think it has to do with this update?

packages/story-editor/src/components/colorPicker/basicColorList.js
samwhale: > > This is fine if there are other options in the listbox, but when there are no options (like in the e2e test for saved colors) then it fails the axe test because the listbox has no children with the role group or option.
> 
> I think I asked it somewhere already but can't find exactly where / what was the answer -- couldn't we just change the role of the Add button to `option` and leave everything else as it was?

For the `Add Color` button to be an `option` it would have to have an `aria-selected` property too - which it can't because it's not selectable. I think making it an option wouldn't make sense to screen reading technologies.

> Alternatively, could we display the `listbox` + the Add button inside it only if there are colors, and display just the one button without a listbox if there aren't any? Not sure if this would be ok from accessibility point of view (cc @barklund)

Oh yeah good idea - I'll investigate this

packages/story-editor/src/components/colorPicker/basicColorList.js
miina: @samwhale You're right! Looks like adding the `Tooltip` wrapper around the `StyledSwatch` broke it.

packages/story-editor/src/components/colorPicker/basicColorList.js
miina: > For the Add Color button to be an option it would have to have an aria-selected property too - which it can't because it's not selectable. I think making it an option wouldn't make sense to screen reading technologies.

Makes sense!

samwhale: > Alternatively, could we display the `listbox` + the Add button inside it only if there are colors, and display just the one button without a listbox if there aren't any? 

This passes axe tests and is much simpler 🥳

miina: Thanks for putting all this effort into fixing something I broke 😛 

I added a separate issue for myself to fix the navigation with arrow keys: https://github.com/google/web-stories-wp/issues/9701

miina (approved):

PR #9679, opened 2021-11-10
Added unit tests for video trimmer

### LINKED ISSUE #9346: Video Trimming: Add tests
## Feature Description
The Video Trimming feature was enabled in #8893 but without tests coverage.

Within this issue we should add proper unit tests and integration tests coverage as well.

barklund:
## Context

This adds exhaustive unit testing for the video trimmer.

Unfortunately, video transcoding does not work in the karma browser runner, so no karma tests are possible. Thus the very extensive unit test setup instead.

## User-facing changes

None

## Testing Instructions

## Checklist

---

Fixes #9346


tests/js/jest.config.js
```diff
@@ -60,8 +60,9 @@ export default {
   coverageReporters: ['lcov'],
   coverageDirectory: '<rootDir>/build/logs',
   collectCoverageFrom: [
-    '<rootDir>/assets/src/**/*.js',
+    '<rootDir>/packages/**/*.js',
     '<rootDir>/bin/**/*.js',
+    '!<rootDir>/packages/fonts/**',
```
barklund: @swissspidy, I updated these, since we weren't actually collecting any coverage from anything (as all the code in `packages` were ignored). I had to exclude fonts though, as Istanbul for some reason didn't like it.

Is coverage still collected somewhere? Is there a report for each PR?

packages/story-editor/src/testUtils/firePointerEvent.js
```diff
@@ -0,0 +1,47 @@
...
+
+pointerEventTypes.forEach((type) => {
+  firePointerEvent[type] = (node, properties) =>
+    firePointerEvent(node, type.toLowerCase(), {
+      bubbles: true,
+      ...properties,
+    });
+});
+
+export default firePointerEvent;
```
swissspidy, line 47: Is this still needed?

How is this different to e.g. `fireEvent.pointerDown()` (which is a thing!)?

tests/js/jest.config.js
swissspidy: Whoops! Good catch.

Do you recall the specific issue with the fonts package? We do have some tests in there, so it would be nice to include them in coverage.

Looks like the codecov config ought to ignore the `cli.js` script in there though.

Code coverage is still collected. There should have been a comment on the PR about it actually 🤔 

Direct link is https://app.codecov.io/gh/google/web-stories-wp though

miina (approved):
🌟

packages/story-editor/src/testUtils/firePointerEvent.js
barklund: `fireEvent.pointerDown`relies on the underlying DOM implementation to have a PointerEvent class, and JSDOM still doesn't, as far as I can tell. I tried going directly through that event, but it doesn't work.

Maybe we need to upgrade JSDOM though, didn't test that out.

tests/js/jest.config.js
barklund: I think it was only `cli.js`, that failed - something about having `await` in non-`async` function?

swissspidy: Cool, then let's only ignore `cli.js` here.

packages/story-editor/src/components/videoTrim/test/videoTrimmer.js
```diff
@@ -0,0 +1,309 @@
...
+      expect(getEndHandle()).toHaveStyle('left: 90.00px');
+    });
+
+    it('should have enabled cancel button', () => {
+      const { getCancel } = setup();
+
+      expect(getCancel()).toBeEnabled();
+    });
+
+    it('should have enabled trim button only once changed', () => {
```
Trapsta: wording is a bit tricky to understand what this is testing

Trapsta (approved):
Looks good just a single nit on the wording

packages/story-editor/src/components/videoTrim/test/videoTrimmer.js
barklund: Fixed in 73cfc68c93.

Issue #9369, opened 2021-10-11
CORS Proxy: Add Tests


swissspidy:
## Feature Description

Add tests for new PHP and JS code

swissspidy:
We've come to the conclusion that it's going to be quite difficult to write useful tests for the proxying part

We'll look into instead adding some complementary e2e tests for hotlinking in general.

PR #9681, opened 2021-11-10
CORS Proxy: Use file stream

### LINKED ISSUE #9368: CORS Proxy: Stream Responses
## Feature Description

From what I've gathered, the current CORS proxy POC saves all data in memory before serving the response.

**(needs confirmation & testing)**

For performance reasons we would want to avoid doing that. Ideally the proxy acts as a simple passthrough stream.

This should be supported by cURL in PHP and apparently `WP_Http` exposes that too, but I haven't tested it.

Something along these lines might work:

```php
$res = wp_safe_remote_get(
	$url,
	[
		// ... other args
		'decompress' => false,
		'stream'     => true,
		'filename'   => 'php://web-stories-cors-proxy', // or something else temporary..?
	]
);

return $res['body'];
```

But I think because WP sets `CURLOPT_RETURNTRANSFER` to `true`, it's not entirely possible out of the box. So perhaps we need a custom WP HTTP handler wrapper around cURL that does something similar to what's outlined at https://stackoverflow.com/a/23078568/1185292 

Again, needs testing & confirmation

swissspidy:
## Context

We want to improve performance of the CORS proxy

## Summary

Updates the cURL calls to send responses to a PHP stream, drastically reducing memory footprint and improving performance.

Certain headers from the proxied resource are just passed through as-is.

## Relevant Technical Choices

* Adds a fallback logic for when cURL is not available.

## To-do

* Further improvements, like forwarding HTTP status / more headers, etc.)
* Ensure `Access-Control-Allow-Origin` headers are sent
* Tests (see also #9369)

## User-facing changes

N/A

## Testing Instructions

1. Use CORS proxy with larger files and see how responses should be quicc

No

No

No

## Checklist

---

Fixes #9368

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,9 @@
     "wp-coding-standards/wpcs": "^2.3",
     "yoast/wp-test-utils": "^1.0.0"
   },
+  "suggest": {
+    "ext-curl": "Used for modifying cURL requests in CORS proxy"
+  },
   "config": {
     "discard-changes": true,
     "platform": {
--- a/includes/REST_API/Hotlinking_Controller.php
+++ b/includes/REST_API/Hotlinking_Controller.php
@@ -65,6 +65,13 @@ class Hotlinking_Controller extends REST_Controller implements HasRequirements {
 	private $types;
 
 	/**
+	 * If streaming to a file, keep the file pointer
+	 *
+	 * @var resource
+	 */
+	protected $stream_handle;
+
+	/**
 	 * Constructor.
 	 *
 	 * @param Story_Post_Type $story_post_type Story_Post_Type instance.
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -32,6 +32,9 @@
     "wp-coding-standards/wpcs": "^2.3",
     "yoast/wp-test-utils": "^1.0.0"
   },
+  "suggest": {
```
spacedmonkey, line 35: Why just suggested?

swissspidy: Because it's not a requirement. There's a fallback in the code if cURL is not avaialble.

spacedmonkey (approved):

PR #456, opened 2020-03-04
Introduce batching callback to avoid sync flush in React

dvoytenko:
Per https://blog.logrocket.com/simplifying-state-management-in-react-apps-with-batched-updates/

PR #9683, opened 2021-11-10
Decoupling: Use camel casing for expected API response from story editor API callbacks

mohdsayed:
## Context

As decided in https://github.com/google/web-stories-wp/pull/9409 we want to use camel casing for the expected API response from all API callbacks. This PR updates `wp-story-editor` and `story-editor` for camel casing.

## Testing Instructions

1.

## Checklist

---


mohdsayed:
Note that I had to move `snakeToCamelCase` to `design-system` ( in https://github.com/google/web-stories-wp/pull/9683/commits/a5c16c1f13bc6a9c54f40993a67a6970a794a797 ) because it was required in `story-editor` inside `taxonomies.js`

here 
https://github.com/google/web-stories-wp/blob/a5c16c1f13bc6a9c54f40993a67a6970a794a797/packages/story-editor/src/components/panels/document/taxonomies/taxonomies.js#L73-L74

and here 
https://github.com/google/web-stories-wp/blob/a5c16c1f13bc6a9c54f40993a67a6970a794a797/packages/story-editor/src/components/panels/document/taxonomies/taxonomies.js#L54-L55

mohdsayed:
Waiting for percy e2e tests to pass.

packages/wp-story-editor/src/api/utils/snakeToCamelCase.js
```diff
@@ -0,0 +1,43 @@
...
+ */
+import { snakeToCamelCase } from '@web-stories-wp/design-system';
+
+/**
+ * Transform a given object keys from snake case to camel case recursively.
+ *
+ * @param {Object} obj Object to be transformed.
+ * @return {any} Transformed object.
+ */
+export function snakeToCamelCaseObjectKeys(obj) {
```
swissspidy: Any reason not to move that to design system as well?

mohdsayed: Yeah, I was tempted to move it to design-system, but I stopped because it's only being used in `wp-*` packages. Ref https://github.com/google/web-stories-wp/pull/9644#discussion_r745695726

swissspidy: Ah right!

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -315,7 +315,7 @@ function useMediaUploadQueue() {
           startUploading({ id });
 
           trackEvent('upload_media', {
-            file_size: newFile.size,
+            fileSize: newFile.size,
             file_type: newFile.type,
```
swissspidy: No camel case here?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/test/useMediaUploadQueue.js
```diff
@@ -44,18 +44,18 @@ const mockResource = {
...
+  sourceUrl: 'http://localhost:9876/__static__/asteroid.ogg',
   title: {
     raw: 'Title',
   },
   description: {
     raw: 'Description',
   },
-  featured_media_src: {},
+  featuredMediaSrc: {},
   meta: {
```
swissspidy, line 59: No camel case for `meta` keys and `web_stories_is_muted` below?

packages/story-editor/src/components/panels/document/taxonomies/taxonomies.js
```diff
@@ -48,10 +49,10 @@ function TaxonomiesPanel(props) {
   }
 
   const availableTaxonomies = taxonomies.filter((taxonomy) => {
-    const isVisible = taxonomy?.visibility?.show_ui;
+    const isVisible = taxonomy?.visibility?.showUi;
     const canAssignTerms = Boolean(
-      capabilities[`assign-${taxonomy?.restBase}`] ||
-        capabilities[`assign-${taxonomy?.slug}`]
+      capabilities[snakeToCamelCase(`assign-${taxonomy?.restBase}`)] ||
+        capabilities[snakeToCamelCase(`assign-${taxonomy?.slug}`)]
```
swissspidy, line 55: Just confirming, this still works as expected?

packages/wp-story-editor/src/api/media.js
```diff
@@ -100,8 +104,8 @@ export async function getMutedMediaById(config, mediaId) {
 
   const result = await apiFetch({ path });
 
-  if (result?.meta?.web_stories_muted_id) {
-    return getMediaById(config, result.meta.web_stories_muted_id);
+  if (result?.meta?.webStoriesMutedId) {
```
swissspidy: I don't think the result we get from the REST API here is camel-cased already.

packages/wp-story-editor/src/api/media.js
```diff
@@ -122,8 +126,8 @@ export async function getOptimizedMediaById(config, mediaId) {
 
   const result = await apiFetch({ path });
 
-  if (result?.meta?.web_stories_optimized_id) {
-    return getMediaById(config, result.meta.web_stories_optimized_id);
+  if (result?.meta?.webStoriesOptimizedId) {
```
swissspidy: Same here, I don't think the result we get from the REST API is camel-cased already.

packages/wp-story-editor/src/api/pageTemplate.js
```diff
@@ -33,8 +37,12 @@ export function getCustomPageTemplates(config, page = 1) {
   return apiFetch({ path }).then(({ headers, body }) => {
     const totalPages = parseInt(headers['X-WP-TotalPages']);
     const templates = body.map((template) => {
-      return { ...template['story_data'], templateId: template.id };
+      const storyData = template['story_data'];
+      storyData.elements = storyData.elements.map(snakeToCamelCaseObjectKeys);
```
swissspidy: Is there a need to specifically camelcase everything inside story_data? 🤔

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
mohdsayed: `fileSize` was mistakenly changed to `fileSize` for `trackEvent` only at this place. Reverted to `file_siz` in https://github.com/google/web-stories-wp/pull/9683/commits/5178599c1879c14b4a6045d0c3da8b837c86a0d0 and verified that it did not accidentally happen anywhere else for `trackEvent`.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/test/useMediaUploadQueue.js
mohdsayed: Thanks, Done in https://github.com/google/web-stories-wp/pull/9683/commits/01becc912951a77dc943238c888682ddf3d347fe

packages/story-editor/src/components/panels/document/taxonomies/taxonomies.js
mohdsayed: Yes, double-checked, and it works as expected.

packages/wp-story-editor/src/api/media.js
mohdsayed: Sorry, my bad. Done in https://github.com/google/web-stories-wp/pull/9683/commits/c29543c002a5f6c407e0d8f56b2d878b44ae7acf

packages/wp-story-editor/src/api/media.js
mohdsayed: Updated in  https://github.com/google/web-stories-wp/pull/9683/commits/c29543c002a5f6c407e0d8f56b2d878b44ae7acf

mohdsayed:
**codecov** is failing with a random negative percentage. Added some more coverage in https://github.com/google/web-stories-wp/pull/9683/commits/30159d9f0392de717811eb8102499699f5a53208 and it started showing even more negative difference

packages/wp-story-editor/src/api/pageTemplate.js
mohdsayed: On the latest `main` branch, I am still getting snake case as a response here.

Some more details would be very helpful.

image

mohdsayed: The response of a newly saved custom template ☝️

swissspidy: But surely not with this branch, no?

Since you've added camelCase to `getResourceFromAttachment`, all resources and their image sizes should be camelCased already.

swissspidy:
> **codecov** is failing with a random negative percentage. Added some more coverage in [30159d9](https://github.com/google/web-stories-wp/commit/30159d9f0392de717811eb8102499699f5a53208) and it started showing even more negative difference

I'd take it with a grain of salt :-)

packages/wp-story-editor/src/api/media.js
```diff
@@ -165,7 +169,7 @@ export function updateMedia(config, mediaId, data) {
     path: `${config.api.media}${mediaId}/`,
     data,
     method: 'POST',
-  });
+  }).then(snakeToCamelCaseObjectKeys);
```
swissspidy: Should this be using `getResourceFromAttachment` instead?

packages/wp-story-editor/src/api/pageTemplate.js
swissspidy: Oh, but not for existing templates.

Let me leave a comment on the PR itself.

packages/wp-story-editor/src/api/story.js
```diff
@@ -167,7 +167,7 @@ export function saveStoryById(config, story) {
       url: embedded?.['wp:featuredmedia']?.[0]?.source_url || '',
     };
 
-    return data;
+    return snakeToCamelCaseObjectKeys(data);
```
swissspidy, line 170: I don't think we can or should do this.

The REST API expects snake_case, but here everything is changed to camelCase

packages/wp-story-editor/src/api/story.js
```diff
@@ -73,7 +73,7 @@ const transformGetStoryResponse = (post) => {
   post.taxonomies = links?.['wp:term']?.map(({ taxonomy }) => taxonomy) || [];
   post.terms = embedded?.['wp:term'] || [];
```
swissspidy: General note:

Seems like we're still returning things like `post._embedded` and `post._links` here, even though we shouldn't 🤔

swissspidy:
As discussed yesterday, writing down some of my thoughts.

First of all, this is looking really good so far!

There's now camelCase in most places already, including `getResourceFromAttachment`, which is good. That means all image sizes keys are properly camelCased for example.

Also, I took another look at `getCustomPageTemplates` as well and now I better understand why you added the camelCasing there.

A few general observations

1. Anything in `story_data` is "versioned". If our data structure changes, we add a migration in `packages/migration` for it.
2. When loading a story in the editor, we run that migration (in `useLoadStory`)
3. When editing an existing story, `story_data` will theoretically still contain snake_case image sizes.  
  a. However, in this PR, the `snakeToCamelCaseObjectKeys` call in `transformGetStoryResponse` actually recursively updates all properties.
  b. That call is a bit greedy _because_ of that recursiveness. I don't think we should do that. We'd only want to camelCase the top level object keys. It should not modify anything inside `story_data`.
4. To update `story_data` and camelCase image sizes for existing stories, we should write a migration for that.
2. When loading existing custom page templates, `story_data` will theoretically still contain snake_case image sizes.  
  a. That said, I just noticed that custom page templates are not currently run through migration ⚠️  That could lead to unexpected behavior when loading a super old page template.
  b. To fix this, I think we need to call `migrate()` in `loadTemplates` within `PageTemplatesPane` or somewhere near there.
3. **That said...** do we even need the image sizes names? It doesn't seem like we do. So we could just simplify this by making `resource.sizes` an array instead (and adding a migration for it)

packages/wp-story-editor/src/api/story.js
mohdsayed: > REST API expects snake_case

Trying to understand this better. Here we are updating the response inside `.then()`. 
The response of `saveStoryById` looks like this

```js
{
    'slug': '130',
    'status': 'publish',
    'link': 'http://localhost:8899/web-stories/130',
    'preview_link': 'http://localhost:8899/web-stories/130',
    'edit_link': 'http://localhost:8899/wp-admin/post.php?post=130&action=edit',
    'embed_post_link': 'http://localhost:8899/wp-admin/post-new.php?from-web-story=130',
    'featured_media': {
      'id': 0,
      'height': 0,
      'width': 0,
      'url': '',
    },
  }
```

which is being used here, after getting converted to camel case.

https://github.com/google/web-stories-wp/blob/c29543c002a5f6c407e0d8f56b2d878b44ae7acf/packages/story-editor/src/app/story/actions/useSaveStory.js#L83-L91

swissspidy: Argh, my bad! Somehow thought this was the request data, not response data.

packages/wp-story-editor/src/api/pageTemplate.js
mohdsayed: > But surely not with this branch, no?

I see, yes, happening on this branch for new templates because of adding `getResourceFromAttachment` to media. 

> Oh, but not for existing templates.

Correct not for existing templates.

packages/wp-story-editor/src/api/media.js
mohdsayed: Why should this be using `getResourceFromAttachment` 🤔

packages/wp-story-editor/src/api/story.js
mohdsayed: Went through all the API callback responses which go to the core editor, and found out that only `_links` is being used for these callbacks, will delete them.

```js
getAuthors: [
  {
    _links: {},
  }
]

updateCurrentUser: {
  _links: {}
}

getTaxonomyTerm: [
  {
    _links: {}
  }
]
```

packages/wp-story-editor/src/api/media.js
swissspidy: Isn‘t this updating a media element and then returning it?

swissspidy:
I don‘t think we should make design-system a dependency of migrations.

We don‘t need that for just updating image sizes names.

Migrations should be self contained.

Also, curious to hear your thoughts on making the image sizes a simple array instead.

mohdsayed:
> I don‘t think we should make design-system a dependency of migrations.

Yes, I realized that, was just working to revert that change.

> Also, curious to hear your thoughts on making the image sizes a simple array instead.

It's being used at a couple of places.

https://github.com/google/web-stories-wp/blob/25d43b402e39f46de36ec4636a1893f8749b932e/packages/story-editor/src/app/highlights/quickActions/useQuickActions.js#L163-L166

https://github.com/google/web-stories-wp/blob/25d43b402e39f46de36ec4636a1893f8749b932e/packages/story-editor/src/components/library/panes/media/local/mediaPane.js#L222-L225

Why do we want to change that to an array?

mohdsayed:
Since PRs are constantly merging in main, one new `source_url` got added in the codebase as snake-case, updated in https://github.com/google/web-stories-wp/pull/9683/commits/206427ce76b15a249234edda17c391943dafcb0a, but going to review the entire change again as there may be more new instances like this.

swissspidy:
> > Also, curious to hear your thoughts on making the image sizes a simple array instead.
> 
> It's being used at a couple of places.
> 
> https://github.com/google/web-stories-wp/blob/25d43b402e39f46de36ec4636a1893f8749b932e/packages/story-editor/src/app/highlights/quickActions/useQuickActions.js#L163-L166
> 
> https://github.com/google/web-stories-wp/blob/25d43b402e39f46de36ec4636a1893f8749b932e/packages/story-editor/src/components/library/panes/media/local/mediaPane.js#L222-L225
> 
> Why do we want to change that to an array?

So, I don't necessarily _want_ to do it, but now is a great opportunity to rethink and refactor how things work in this area.

Some thoughts:

* Resource sizes like `medium`  or `web-stories-thumbnail` are quite WP-specific. It's odd that these are now referenced as `webStoriesThumbnail`.
* The only reason why we reference image sizes by name is convenience, to do things like `resource.sizes?.medium?.source_url || resource.src` so we get a reasonably sized image for insertion
* Instead of accessing image sizes like that, we could just as well have some `getSomeReasonablySizedImageSizeForInsertion(resource)` function or whatever to get this image size.\
* If we have such a function we don't need image sizes names. They could just be an array and we don't need a migration for them.

WDYT?

Related:

* camelCasing things like capabilities and taxonomy names takes some time getting used to. For example `capabilities[snakeToCamelCase(`assign-${taxonomy?.restBase}`)]` is not very readable and also a bit error prone, as the taxonomy is `web-story-tag`, but the capability is `assignWebStoryTag`.
* Not sure how that could be improved though

mohdsayed:
> Instead of accessing image sizes like that, we could just as well have some getSomeReasonablySizedImageSizeForInsertion(resource) function or whatever to get this image size.\

Do you mean `getSomeReasonablySizedImageSizeForInsertion` utility function would be able to find the closest image size? I think that would unnecessarily increase complexity. Would it be easier that we document what sizes keys ( with generic names ) we are expecting? Do you see any problem with it? For example:

```js
'sizes': {
    'small': {
      'file': '',
      'width': 100,
      'height': 150,
      'mimeType': 'image/jpeg',
      'sourceUrl': '',
    },
    'medium': {
      'file': '',
      'width': 300,
      'height': 150,
      'mimeType': 'image/jpeg',
      'sourceUrl': '',
    },
    'large': {
      'file': '',
      'width': 1024,
      'height': 576,
      'mimeType': 'image/jpeg',
      'sourceUrl': '',
    },
````

----

> camelCasing things like capabilities and taxonomy names takes some time getting used to. For example capabilities[snakeToCamelCase(assign-${taxonomy?.restBase})] is not very readable and also a bit error prone, as the taxonomy is web-story-tag, but the capability is assignWebStoryTag.


Agree it's confusing, will work on it in some other PR.

**Comment by swissspidy** [comment]
Converting this to draft as I feel like things like image sizes and caps need a bit more thinking. Let's sync about this.

> I think that would unnecessarily increase complexity.
> Would it be easier that we document what sizes keys ( with generic names ) we are expecting?

Yeah true. The advantage would be that the image sizes names don't matter.

Right now the only image size we use by name seems to be `medium`. And that's just because we don't want to use `resource.src` (the `full` size) when inserting on the canvas for performance reasons (don't wanna display a 4K picture there).

`getResourceFromMedia3p` / `getImageUrls` mentions a bunch more image sizes names, but we don't reference the sizes by names, ever, as far as I can tell.

Aside: We register a `web-stories-thumbnail` image size in WP but all the JS code seems to reference `web_stories_thumbnail` 🤔 

> Agree it's confusing, will work on it in some other PR.

Curious, what did you have in mind here?

I hesitate a bit to camelCase taxonomy names. If I have 1 taxonomy "web-story-tag" and one "webStoryTag", then there would be conflicts.

The same goes for image sizes names too.

**Comment by mohdsayed** [comment]
> Aside: We register a web-stories-thumbnail image size in WP but all the JS code seems to reference 
web_stories_thumbnail 🤔

I think both `web-stories-thumbnail` and `web_stories_thumbnail` are not being used. They were probably used before. Found the PR where `web_stories_thumbnail` was added https://github.com/google/web-stories-wp/pull/3089

**Comment by mohdsayed** [comment]
Do you have something like `getSmallestUrlForWidth` [getSmallestUrlForWidth.js#L22-L30](https://github.com/google/web-stories-wp/blob/main/packages/media/src/getSmallestUrlForWidth.js#L22-L30) in mind for getting the closest image for `getSomeReasonablySizedImageSizeForInsertion` ?

**Comment by mohdsayed** [comment]
>  Curious, what did you have in mind here?
I hesitate a bit to camelCase taxonomy names. If I have 1 taxonomy "web-story-tag" and one "webStoryTag", then there would be conflicts.

I think the easiest is to add an `ignore` param in the `snakeToCamelCaseObjectKeys`,  which will be exempted in camel casing like I just did for `story_data` in https://github.com/google/web-stories-wp/pull/9683/commits/ab3220b196a85a8cd3bc95c6ee98e54283667721

Similarily we can use `capabilities`.
```
snakeToCamelCaseObjectKeys(post, ['story_data', 'capabilities']);
```

The only problem is undoing the changes made for `capabilities`. 


----

Aside: We changed these many snake cases to camel cases, and considering so many changes that merged in `main`, I am not feeling very confident with this PR, I think I will create a fresh PR after we are done with our discussion here.

```js
- permalink_template
- story_data
- style_presets
- preview_link
- edit_link
- embed_post_link
- assign-author
- assign-web_story_tag
- assign-web_story_category
- create-web_story_tag
- create-web_story_category
- lock_user
- featured_media
- publisher_logo
- date_gmt
- modified_gmt
- source_url
- mime_type
- **Remove ping_status and comment_status**
- web_stories_is_muted
- featured_media_src
- alt_text
- media_details
- media_type
- length_formatted
- file_sizes
- file_name
- manage_terms
- edit_terms
- delete_terms
- assign_terms
- add_new_item
- items_list
- separate_items_with_commas
- most_used
- singular_name
- search_items
- all_items
- parent_item
- parent_item_colon
- edit_item
- view_item
- update_item
- add_new_item
- new_item_name
- separate_items_with_commas
- add_or_remove_items
- choose_from_most_used
- not_found
- no_terms
- filter_by_item
- items_list_navigation
- items_list
- most_used
- back_to_items
- item_link
- item_link_description
- menu_name
- name_admin_bar
- publicly_queryable
- show_admin_column
- show_in_nav_menus
- show_in_quick_edit
- show_ui
- source_url
- web-stories-poster-portrait
- web-stories-publisher-logo
- web-stories-thumbnail
- web_stories_media_optimization
- web_stories_onboarding
- web_stories_tracking_optin
- web_stories_muted_id
- web_stories_optimized_id
- web_stories_is_poster
- web_stories_poster_id
- web_stories_trim_data
- web_stories_is_muted
```

mohdsayed:
@swissspidy As discussed on Tuesday, could you please add some more thoughts on media sizes?

packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
```diff
@@ -115,7 +116,16 @@ function PageTemplatesPane(props) {
...
+        const updatedTemplates = [...(savedTemplates || []), ...templates].map(
+          (template) => {
+            return {
+              version: DATA_VERSION,
+              ...migrate(template, template.version),
+            };
+          }
+        );
+        setSavedTemplates(updatedTemplates);
+
```
swissspidy, line 128: ❤️

packages/story-editor/src/app/media/utils/getResourceFromMedia3p.js
```diff
@@ -158,7 +158,7 @@ function getGifUrls(m) {
       .slice(2, imageSizesFromBiggest.length - 1)
       .map((u) => [u.width + '_' + u.height, u]),
     [
-      'web_stories_thumbnail',
+      'web-stories-thumbnail',
```
mohdsayed, line 161: Are we planning to keep the image sizes?

packages/wp-story-editor/src/api/media.js
```diff
@@ -169,7 +165,7 @@ export function updateMedia(config, mediaId, data) {
     path: `${config.api.media}${mediaId}/`,
     data,
     method: 'POST',
-  }).then(snakeToCamelCaseObjectKeys);
+  }).then(getResourceFromAttachment);
```
mohdsayed, line 168: As I can see `updateMedia` is being used at these many places, and nowhere have they used its response. So shouldn't we omit the `.then(getResourceFromAttachment);` part?

https://github.com/google/web-stories-wp/blob/acb3df5148d37e8192023c70ea2d9256cf86241d/packages/story-editor/src/app/media/utils/useDetectBaseColor.js#L69-L71
https://github.com/google/web-stories-wp/blob/acb3df5148d37e8192023c70ea2d9256cf86241d/packages/story-editor/src/app/media/utils/useProcessMedia.js#L83-L88
https://github.com/google/web-stories-wp/blob/acb3df5148d37e8192023c70ea2d9256cf86241d/packages/story-editor/src/app/media/utils/useProcessMedia.js#L95-L99
https://github.com/google/web-stories-wp/blob/acb3df5148d37e8192023c70ea2d9256cf86241d/packages/story-editor/src/app/media/utils/useDetectVideoHasAudio.js#L81-L83
https://github.com/google/web-stories-wp/blob/acb3df5148d37e8192023c70ea2d9256cf86241d/packages/story-editor/src/app/media/utils/useUploadVideoFrame.js#L90-L96
https://github.com/google/web-stories-wp/blob/acb3df5148d37e8192023c70ea2d9256cf86241d/packages/story-editor/src/components/library/panes/media/local/mediaEditDialog.js#L121

packages/wp-story-editor/src/api/media.js
```diff
@@ -189,5 +185,7 @@ export function deleteMedia(config, mediaId) {
     path: addQueryArgs(`${config.api.media}${mediaId}/`, { _method: 'DELETE' }),
     data: { force: true },
     method: 'POST',
-  }).then(snakeToCamelCaseObjectKeys);
+  }).then((result) => {
+    return getResourceFromAttachment(result.previous);
```
mohdsayed, line 189: Similarly, the response is `deleteMedia` is not being used anywhere. I think we may remove the `.then` part.

https://github.com/google/web-stories-wp/blob/acb3df5148d37e8192023c70ea2d9256cf86241d/packages/story-editor/src/components/library/panes/media/local/deleteDialog.js#L62

packages/wp-story-editor/src/api/media.js
swissspidy: Yeah not sure about it, feels like this might change in the future and we'd need the return value. Seems like something one would expect for updating media

packages/wp-story-editor/src/api/media.js
swissspidy: Yeah for this one it probably makes more sense to return void 👍

packages/wp-story-editor/src/components/checklist/accessibility/videoOptimizationCheckbox/index.js
```diff
@@ -63,7 +63,7 @@ function VideoOptimizationCheckbox() {
     })
   );
 
-  const checked = currentUser?.meta?.web_stories_media_optimization;
+  const checked = currentUser?.meta?.webStoriesMediaOptimization;
```
swissspidy, line 66: Post meta is one thing I'm not sure about yet.

For reading meta values (e.g. from media), we need to camelCase, but for updating we obviously need to use the snake_case format. That's slightly confusing.

swissspidy:
> Are we planning to keep the image sizes?

I'm still tinkering with this but it feels more reasonable to leave dynamic object keys like image sizes and taxonomy names (which are both out of our control) untouched.

packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
swissspidy: FYI I split this up into #9897 now

mohdsayed:
Closing in favour of https://github.com/google/web-stories-wp/pull/9906

PR #9694, opened 2021-11-11
Workspace: Don't throw errors for falsey values in context selectors

### LINKED ISSUE #9600: Context: Fix sneaky bug in the hooks across the application
## Bug Description

This app uses the [`useContextSelector` hook](https://github.com/dai-shi/use-context-selector). The current pattern looks like:

```js
export function useHelpCenter(selector) {
  const context = useContextSelector(Context, selector ?? identity);

  if (!context) {
    throw new Error('Must use `useHelpCenter` within <helpCenter.Provider />');
  }

  return context;
}

```

But this could fail if the selector is selection for just one value that ends up being 'falsey', like

```js
const isHelpCenterOpen = useHelpCenter((stuff) => stuff.isHelpCenterOpen);
```

This would throw an error whenever the help center is closed.

## Expected Behaviour

Using the hook should only throw an error if the hook is used outside of the related provider.

samwhale:
## Context

Sneaky bug that was discovered. Selecting for a false-y value could crash the app.

## Summary

Remove error throwing that was a result of previous guardrails with the `useContext` hook. Now that we have `useContextSelector` we don't need these.

no

no

no

## Checklist

---

Fixes #9600


BrookeGraham (approved):
LGTM

swissspidy:
The `should throw an error if used oustide StroyAnimation.Provider` unit test would need to be removed

BrittanyIRL (approved):

swissspidy (approved):

maxyinger (approved):

Issue #7566, opened 2021-05-18
Zoom: Page Attachment not positioned correctly.


miina:
## Bug Description
Page Attachment is not positioned correctly when present and using zoom. The size is also not adjusted.

## Expected Behaviour
Page Attachment should be positioned at the bottom of the page and change size according to zoom, too.

## Steps to Reproduce
1. Add Page Attachment
2. Zoom to 200%.

## Screenshots
![Screenshot 2021-05-18 at 13 18 33](https://user-images.githubusercontent.com/3294597/118634760-30d24080-b7d3-11eb-82c1-699ac08960a1.png)

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

csossi:
Verified in QA

PR #9699, opened 2021-11-12
Media: Improve video captions appearance in editor

### LINKED ISSUE #9344: Video Captions: Use similar captions appearance in editor
## Feature Description

In #9043 we implemented initial support for using `<amp-story-captions>` to display video captions when viewing stories.

This allows showing captions always on top of all other elements so that they're never covered/obstructed by other elements.

Example:

Screenshot 2021-09-29 at 14 29 32

Demo Story: https://wp.stories.google/stories/three-tips-to-build-your-digital-community/

In the editor itself we also show captions, but not in this nice way.

Example:

Screenshot 2021-10-08 at 19 57 50

Notice how the caption is covered by other elements on the page in the editor.

I do wonder whether we should aim for a similar captions appearance in the editor so that it more closely matches the end result.

## Alternatives Considered

* Keep status quo
* Don't display captions in the editor? Would be confusing though and not good for accessibility.

swissspidy:
## Context

We recently improved how captions are displayed on the frontend.

Now we want to do the same in the editor as well.

## Summary

Adapts video captions appearance in the editor so that it matches the frontend experience, using a custom renderer for the video tracks.

## Relevant Technical Choices

* Adapted the logic from `amp-story-video-captions` to parse timestamps and manually render cues.
* Uses `dangerouslySetInnerHTML` to render the result of [`VTTCue.getCueAsHTML()`](https://developer.mozilla.org/en-US/docs/Web/API/VTTCue/getCueAsHTML)

## To-do

* Look into issues with captions not being displayed at all.
* Manual testing.

## User-facing changes

Screenshot 2021-11-25 at 13 07 26

## Testing Instructions

1. Enable editor video captions experiment
2. Add a new video to a story and add captions to it
3. Move it around and maybe add some other elements on top of it so it's only partially visible
4. Play the video in the editor
5. See that the captions are displayed on top of all other elements and always at the same position on the canvas

Using `dangerouslySetInnerHTML` in a couple of places, but the track contents should be considered trusted.

No

No

## Checklist

---

Fixes #9344


swissspidy:
@miina @barklund Perhaps this is related to #8255. In this PR I added some change to not render the `FramesLayer` and `DisplayLayer` components when the canvas dimensions are not yet set (i.e. are 0x0). Avoids some unnecessary renders and perhaps improves performance? Not sure about any negative impact though.

miina:
> In this PR I added some change to not render the FramesLayer and DisplayLayer components when the canvas dimensions are not yet set (i.e. are 0x0). Avoids some unnecessary renders and perhaps improves performance? Not sure about any negative impact though.

Thanks for tagging, we've been just discussing potentially rendering the frames of elements only conditionally as part of the Page performance discovery, so that's related indeed. Shouldn't really have any negative impact, AFAIK. cc @Trapsta since you looked into this.

packages/story-editor/src/components/canvas/canvasLayout.js
```diff
@@ -77,12 +88,13 @@ function CanvasLayout({ header }) {
...
             <SelectionCanvas>
-              <DisplayLayer />
-              <FramesLayer />
+              {hasDimensions && <DisplayLayer />}
+              {hasDimensions && <FramesLayer />}
               <NavLayer header={header} />
             </SelectionCanvas>
             <EditLayer />
             <EyedropperLayer />
+            {hasDimensions && <VideoCaptionsLayer />}
```
miina: Curious why not display it inside the display layer? Maybe it would be more straightforward to keep the captions inside the visible area when using the display layer (e.g. in case of zoomed videos).

Also, I'm wondering what should happen with the captions if the video is partially out of the page, should the captions be forced to the visible area only?

swissspidy: I'll look into moving this to the DisplayLayer, haven't tested that at all.

> Also, I'm wondering what should happen with the captions if the video is partially out of the page, should the captions be forced to the visible area only?

The video captions position should never change, regardless of video position.

swissspidy: Moved it to DisplayLayer now, not sure that makes a difference or brings an advantage. Seems kinda pointless since the captions position should never change anyway.

miina: Hm, I suppose not -- when I first tested it, I actually saw the captions on top of the video (in the bottom area), so I assumed the idea was to display the captions on top of the video that it relates to, but on top of all other elements... That's why I thought that maybe having it in the `DisplayLayer` of that specific element might be beneficial. But it looks like it's actually displayed in the same place related to the page, misunderstood that. Since the display layer includes everything that's related to what the user gets displayed, should still be good to have it there though.

What happens if there are more than one video with captions playing at the same time?

swissspidy: > What happens if there are more than one video with captions playing at the same time?

On the frontend the captions are stacked.

In the editor it's not even possible to play 2 videos at the same time, as only the currently selected video can be played.

miina: Not directly related to this PR but it seems to me that when a video has captions, the element links on that page don't work anymore in the FE. Is it just me or can you also reproduce it? The `captions-area` seems to cover everything and prevent the links from being clicked.

swissspidy: Yikes, yeah I can reproduce it. Opened #9819 for this. Good catch!

miina: One more thing I'm wondering about -- it looks like currently, using a Page Attachment and Video captions on the same page is also problematic (the captions get partially hidden under the Page Attachment) -- is this expected behaviour?

swissspidy: Do you mean in the editor or on the frontend?

I just saw that on the frontend it can happen only on smaller screens:

Screenshot 2021-11-26 at 11 25 08

Not sure how easy (and urgent) it would be to change the position for smaller screens.

On the frontend you'll notice that the page attachment is positioned at the very bottom of the page:

Screenshot 2021-11-26 at 11 31 26

Whereas in the editor, the page attachment has lots of space below it:
Screenshot 2021-11-26 at 11 26 36

So that will lead to captions getting partially hidden:

Screenshot 2021-11-26 at 11 33 46

I think the proper solution here is to move the page attachment further down in the editor, so it matches what you actually see on the frontend.

miina: > Do you mean in the editor or on the frontend?

> I just saw that on the frontend it can happen only on smaller screens:


I meant on the front-end. I'm currently using a 13'' screen so that's a smaller screen and explains why I saw it right away, wasn't aware it works as expected on larger screens.

miina: cc @barklund related to the Page Attachment positioning in the editor since you're currently working on #7566 (in case it's easy to adjust it as part of that issue in Fit mode as well)

packages/story-editor/src/components/canvas/videoCaptions/cue.js
```diff
@@ -0,0 +1,108 @@
...
+                <span>
+                  <CueEl $isFutureCue={isFutureCue}>{node.textContent}</CueEl>
+                </span>
+              ) : (
+                <CueEl>{node.textContent}</CueEl>
+              );
+            }
+
+            return timestamp !== null ? (
+              <span>
```
spacedmonkey, line 84: Why does this need to be wrapped in a span?

swissspidy: For styling purposes, matching the markup that AMP generates on the frontend.

swissspidy:
@miina @choumx Friendly ping for review :-)

packages/story-editor/src/components/canvas/canvasLayout.js
```diff
@@ -67,6 +67,16 @@ function CanvasLayout({ header, footer }) {
...
 
+  const { pageWidth, pageHeight } = useLayout(
+    ({ state: { pageWidth, pageHeight } }) => ({
+      pageWidth,
+      pageHeight,
+    })
+  );
+
+  // If we don't have proper canvas dimensions yet, don't bother rendering element layers.
+  const hasDimensions = pageWidth !== 0 && pageHeight !== 0;
```
miina, line 78: Nit: If either of those is `0`, probably doesn't make sense to render? Although not sure if that ever happens.

packages/story-editor/src/components/canvas/videoCaptions/videoCaptionsLayer.js
```diff
@@ -0,0 +1,116 @@
...
+  }));
+
+  const videoElement = useStory(({ state }) => {
+    const { currentPage, selectedElementIds } = state;
+
+    if (selectedElementIds.length !== 1) {
+      return null;
+    }
+
+    return (currentPage?.elements || [])
```
miina: If we're only checking for the single selected element anyway, is it necessary to filter all the elements on the current page? Maybe just use `selectedElements[0]` to check if it's a video and has `tracks.length`? WDYT?

packages/story-editor/src/components/canvas/videoCaptions/trackRenderer.js
```diff
@@ -0,0 +1,133 @@
...
+
+    updateCues();
+
+    setTrack(videoTrack);
+
+    videoEl.addEventListener('timeupdate', () => {
+      setVideoTime(videoEl.currentTime);
+    });
+
+    // TODO: Figure out why this doesn't work after moving the video on the canvas.
```
miina, line 99: Does this need to be figured out within this PR? :)

miina (approved):

packages/story-editor/src/components/canvas/videoCaptions/trackRenderer.js
swissspidy: It would be nice to, but not critical. If you have any ideas or pointers, that would be great.

packages/story-editor/src/components/canvas/videoCaptions/videoCaptionsLayer.js
swissspidy: Totally 🤦 

I used `elements` first before realizing that one can only play the selected element anyway.

packages/story-editor/src/components/canvas/canvasLayout.js
swissspidy: Yeah doesn't really make sense to render in that case; which is why this checks for both to be non-zero.

miina: Ah, right, it is already checking for that 💡 Misread the code.

packages/story-editor/src/components/canvas/canvasLayout.js
swissspidy: @miina Is this resolved now with #7566?

packages/story-editor/src/components/canvas/videoCaptions/trackRenderer.js
swissspidy: If not, not a big deal since this is still behind a feature flag.

packages/story-editor/src/components/canvas/canvasLayout.js
miina: Should be, the Page Attachment is now fixed close to the bottom edge as in the front-end, however, on smaller screens these (captions + the link) still overlap.

packages/story-editor/src/components/canvas/videoCaptions/trackRenderer.js
miina: Ah, forgot to answer -- I did debug it briefly that day but couldn't come up with anything, would need more time probably to investigate.

packages/story-editor/src/components/canvas/videoCaptions/videoCaptionsLayer.js
```diff
@@ -0,0 +1,111 @@
...
+
+  const videoElement = useStory(({ state }) => {
+    const { selectedElementIds } = state;
+
+    if (selectedElementIds.length !== 1) {
+      return null;
+    }
+
+    return (selectedElementIds || [])
+      .filter(({ type, tracks }) => type === 'video' && tracks?.length > 0)
```
miina: Hm, doesn't `selectedElementIds` only include the ids though? Probably should use `selectedElements` instead?
Also, if we check for `selectedElements.length === 1` then we could just check the `selectedElements[0]` as well instead of using `filter` and `pop`.

dreamofabear (approved):

PR #9702, opened 2021-11-12
Hotlinking: Fix CORS Proxy issue and link validation bug

### LINKED ISSUE #9696: Hotlinking: False positive for invalid link
## Bug Description

Valid urls are being flagged as invalid url, when they are not. 

## Steps to Reproduce

1. Create a new story. 
2. Click in media from url. 
3. Input this URL: https://spacedmonkey.ams3.cdn.digitaloceanspaces.com/uploads/2021/10/sample_5184×3456-1-scaled.jpeg
4. See error.

## Screenshots

Screenshot 2021-11-12 at 12 40 37

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

swissspidy:
## Context

There was an issue with the CORS proxy not returning the correct response code, as well as an issue with URLs not being encoded before validated on the client-side.

Also, insertion when using the proxy was potentially quite slow, which was not very clear to users.

## Summary

This PR contains a few (significant) changes to improve the overall UX and performance of hotlinking.

1. **Accessibility**: The hotlinking dialog can now be submitted by pressing enter.
2. **User Feedback**: The "Submit" button is now disabled while insertion happens.  
  A spinner could perhaps be added in the future.
3. **Performance**: Overhaul insertion logic to reduce the number of times the video is fetched.  
  Mostly relevant when loading larger video files via the CORS proxy.

Regarding the latter, a video is now _preloaded_ only once, so that dimensions, length, and audio information can be retrieved.

If the user can upload media, the video is _seeked_ to the initial frame to generate a poster image, even _before_ the video is actually inserted.

Doing that _before_ insertion has two main reasons:

1. Doing it _after_ and updating the video using `updateElementsByResourceId` would cause an unexpected history entry.
2. Using `updateElementsByResourceId` for that would also mean the video gets re-rendered, which means playback stops.

## Relevant Technical Choices

Refactored a lot of the logic for preloading videos and extracting metadata, to reduce repetition in multiple areas.

## To-do

N/A

## User-facing changes

* The hotlinking dialog can now be submitted by pressing enter
* The "Submit" button is now disabled while insertion happens

## Testing Instructions

1.  Verify that external videos and images can be inserted via hotlinking
2. Verify that URLs like `https://foo` are rejected
2. Verify that the "Insert" button changes to "Inserting" while insertion is in progress

Some possible test files:

* https://c.tenor.com/Mc-DPcAOaNsAAAPo/lol-omg.mp4
* https://sample-videos.com/video123/mp4/720/big_buck_bunny_720p_30mb.mp4 (large)
* https://spacedmonkey.ams3.cdn.digitaloceanspaces.com/uploads/2021/10/sample_640%C3%97426.jpeg
* https://spacedmonkey.ams3.cdn.digitaloceanspaces.com/uploads/2021/10/sample-30s-1.webm (large)
* https://spacedmonkey.ams3.cdn.digitaloceanspaces.com/uploads/2021/10/sample_960x400_ocean_with_audio.mp4 (large)
* https://stream.mux.com/r4lT81CEvYAIutD01Zp7KAZAMnk2dbafb/high.mp4?token=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6InQ5UHZucm9ZY0hQNjhYSmlRQnRHTEVVSkVSSXJ0UXhKIn0.eyJleHAiOjE2Mzc1OTUyNTcsImF1ZCI6InYiLCJzdWIiOiJyNGxUODFDRXZZQUl1dEQwMVpwN0tBWkFNbmsyZGJhZmIifQ.GedMsn0dzJHhNl7nj6jKpMgc7G4u7-hOUt98HBeiI5OAiiGw5_cWPBFMqscaUo7WA9MrNDpMVk6aSsXOvtGlaWvj58Tjh5C08Z1Wjabioba2zMN8MED-ZfYoqj5KOIkrDYV2hY0ZzT83LBCwL83uLbYCK2xtjvjFRUBsMimQtksLxVy6KN28x-ahHmexOwh-u5zdc3I9eEIRGi1x56KOWnOavOZVtJREiBZvZs7ApCvxGvydit5IoMUleJxia_TtitZHBbuLV-SHqAPyLc5-gYKYVRnFP3Z4brHHF2nkSerdA3EfMw6Dpcj85UjYzlk875rih9VNtqikPDdw0ki4_A

**Notes**

* Hotlinking large media files can be slow to insert because they're loaded through the WP site (proxy)

No

No

No

## Checklist

---

Fixes #9696


spacedmonkey:
This is working great!

The only example I have found that doesn't work is this url - https://filesamples.com/samples/video/mp4/sample_1280x720_surfing_with_audio.mp4

spacedmonkey:
https://user-images.githubusercontent.com/237508/141507013-60778fc8-9908-48ad-b3b2-58c7f8c8a077.mov

This fix does not address this looping issue. https://sample-videos.com/video123/mp4/720/big_buck_bunny_720p_30mb.mp4

miina (approved):
Thanks for the validation fix!

LGTM pending Jonny's looping comment.

miina:
@swissspidy A hotlinking-related Karma change was merged meanwhile, that might fix the failing Karma test.

spacedmonkey:
I think we can merge this change, as it does fix the issue and work well. 

We can create another issue / PR for the looping issue.

spacedmonkey:
Is that a valid karma test failure?

miina:
> Is that a valid karma test failure?

Hm, it seems so, looks like the behaviour changed within this PR. Previously it just showed `Invalid link` when inserting `d` (used in the tests) as the link to insert, but now it seems to initially display a message about incorrect file extension instead which fails the Karma test. 🤔 The message changes to "Invalid link" after the link is passed to the endpoint for validation but Karma test catches the initial message.

It seems that `isValidUrl` is generally working a bit oddly -- it shouldn't probably consider `https://d` a valid URL even when encoded as this PR is now doing, no?

swissspidy:
> It seems that `isValidUrl` is generally working a bit oddly -- it shouldn't probably consider `https://d` a valid URL even when encoded as this PR is now doing, no?

Technically that's a valid URL (run `new URL('https://d')` to confirm). So `valid-url` correctly allows it.

For the sake of hotlinking we could add some stricter check on top of that though, for example disallow empty paths (`"/"`) like we already do on the server side. After all without a path we don't have a valid file name.

I'll take a look.

packages/story-editor/src/components/library/panes/media/local/hotlink/useInsert.js
```diff
@@ -54,6 +60,23 @@ function getErrorMessage(code, description) {
   }
 }
 
+function seekVideo(video) {
```
spacedmonkey: This can could be reused in `preloadVideo`.

swissspidy: Yeah maybe, it's not super straightforward to reuse though because it just works off an existing video el. Not the most beautiful code :)

packages/story-editor/src/components/library/panes/media/local/hotlink/useInsert.js
```diff
@@ -78,40 +106,87 @@ function useInsert({ link, setLink, setErrorMsg, onClose }) {
...
+          resourceLike.length = Math.round(video.duration);
+          resourceLike.lengthFormatted = getVideoLengthDisplay(
+            resourceLike.length
+          );
+
+          resourceLike.isMuted = !(
+            video.mozHasAudio ||
+            Boolean(video.webkitAudioDecodedByteCount) ||
+            Boolean(video.audioTracks?.length)
           );
```
spacedmonkey: We could just export the `hasAudio` function.

packages/story-editor/src/components/library/panes/media/local/hotlink/useInsert.js
spacedmonkey: You could move seekVideo function to the media package and export it. Then use seek video in preloadVideo.

spacedmonkey:
```
function preloadVideo(src) {
  const video = document.createElement('video');
  video.muted = true;
  video.crossOrigin = 'anonymous';
  video.preload = 'auto';

  return seekVideo(video);
}
```

swissspidy:
@spacedmonkey @miina Re-requesting review because I ended up doing a bigger overhaul than originally anticipated.

I tried to slightly improve UX and also performance by only creating one `<video>` element to get all the metadata we need for a video. If only inserting media with reasonable size these two points shouldn't really be a concern though (happy case).

I'll leave holistic QA for hotlinking of various media types with different file sizes etc. to a dedicated QA ticket that I have yet to create.

Ideally we merge this PR as-is to unblock other ongoing work and — most importantly — unbreak the CORS proxy.

packages/media/src/generateVideoStrip.js
```diff
@@ -18,29 +18,8 @@
...
-  if (video.currentTime === offset) {
-    return Promise.resolve();
-  }
-  return new Promise((resolve) => {
-    video.addEventListener('seeked', resolve, { once: true });
-    video.currentTime = offset;
-  });
-}
+import preloadVideo from './preloadVideo';
+import seekVideo from './seekVideo';
```
spacedmonkey, line 22: Is seekVideo used in this file?

packages/story-editor/src/app/media/utils/useDetectVideoHasAudio.js
```diff
@@ -55,7 +55,7 @@ function useDetectVideoHasAudio({ updateMediaElement }) {
         return;
       }
       try {
-        const hasAudio = await hasVideoGotAudio(src);
+        const hasAudio = hasVideoGotAudio(await preloadVideo(src));
```
spacedmonkey: I find this more readable. 

```
 const video = await preloadVideo(src);
 const hasAudio = hasVideoGotAudio(video);
```

includes/REST_API/Hotlinking_Controller.php
```diff
@@ -276,9 +288,9 @@ public function proxy_url( $request ) {
 			// Other than php://memory, php://temp will use a temporary file once the amount of data stored hits a predefined limit (the default is 2 MB).
 			// The location of this temporary file is determined in the same way as the {@see sys_get_temp_dir()} function.
 			if ( WP_DEBUG ) {
-				$stream_handle = fopen( 'php://temp', 'wb' ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_read_fopen
+				$stream_handle = fopen( 'php://memory', 'wb' ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_read_fopen
```
spacedmonkey, line 321: What is the benefit of using memory of temp here?

packages/story-editor/src/app/media/utils/getResourceFromLocalFile.js
```diff
@@ -89,8 +89,8 @@ const getVideoResource = async (file) => {
     length = videoLength.length;
     lengthFormatted = videoLength.lengthFormatted;
   }
+  const hasAudio = hasVideoGotAudio(videoEl);
   const posterFile = await getFirstFrameOfVideo(src);
```
spacedmonkey: We could do this as well. 
```suggestion
  await seekVideo(videoEl);
  const posterFile = await getImageFromVideo(videoEl);
```

swissspidy: `getFirstFrameOfVideo` already calls `seekVideo`

includes/REST_API/Hotlinking_Controller.php
swissspidy: The comment above explains the difference well.

`php://temp` will use a temporary file once the amount of data stored hits a predefined limit  (the default is 2 MB), so touching the disk. `php://memory` is all in-memory.

packages/story-editor/src/components/library/panes/media/local/hotlink/useInsert.js
```diff
@@ -78,37 +87,79 @@ function useInsert({ link, setLink, setErrorMsg, onClose }) {
...
+        resource.src = link;
+
         insertElement(type, {
           resource,
         });
 
+        setIsInserting(false);
         setErrorMsg(null);
         setLink('');
         onClose();
```
miina, line 162: Below, in the catch, should it also do `setIsInserting(false);`?

swissspidy: I thought that was handled by the `catch` in `onInsert`, but you're right, it's needed here as well.

packages/story-editor/src/app/media/utils/useDetectVideoHasAudio.js
swissspidy: Really indifferent to me, both are equally readable to me. Personal preference I suppose 🙂

includes/REST_API/Hotlinking_Controller.php
spacedmonkey: Yes, I read the 2 line comment. I wanted more details than that.

packages/story-editor/src/app/media/utils/getResourceFromLocalFile.js
spacedmonkey: This is about reusing the video tag.

packages/story-editor/src/components/library/panes/media/local/hotlink/useInsert.js
```diff
@@ -78,41 +87,84 @@ function useInsert({ link, setLink, setErrorMsg, onClose }) {
...
       } catch (e) {
+        setIsInserting(false);
         setErrorMsg(getErrorMessage());
       }
```
spacedmonkey:
```suggestion
        setErrorMsg(null);
        setLink('');
        onClose();
      } catch (e) {
        setErrorMsg(getErrorMessage());
      } finally {
        setIsInserting(false);
      }
```

Might be cleaner to do this.

packages/story-editor/src/components/library/panes/media/local/hotlink/useInsert.js
```diff
@@ -78,41 +87,84 @@ function useInsert({ link, setLink, setErrorMsg, onClose }) {
...
+          id: uuidv4(),
           src: proxiedUrl,
           mimeType,
           needsProxy,
           alt: originalFileName,
-        });
-        resource.src = link;
+        };
+
+        let video;
```
spacedmonkey: Why is this a let, is never changed.

packages/story-editor/src/components/library/panes/media/local/hotlink/useInsert.js
```diff
@@ -78,33 +87,74 @@ function useInsert({ link, setLink, setErrorMsg, onClose }) {
...
+
+        // We need to gather some metadata for videos, but efficiently.
+        // Thus only loading it once to speed up insertion.
+        if (isVideo) {
+          // preloadVideoMetadata would suffice, except for audio detection
+          // which requires loading more than just metadata.
+          video = await preloadVideo(proxiedUrl);
```
spacedmonkey:
```suggestion
         const video = await preloadVideo(proxiedUrl);
```

spacedmonkey:
Screenshot 2021-11-18 at 15 09 54

Inserting needs to be reset on closing the dialog.

spacedmonkey:
https://user-images.githubusercontent.com/237508/142442587-fcf8dee2-487c-4446-83e3-be12a0862e30.mov

packages/story-editor/src/components/library/panes/media/local/hotlink/useInsert.js
swissspidy: Ah, some leftover

swissspidy:
> Inserting needs to be reset on closing the dialog.

Fixed now

packages/story-editor/src/app/media/utils/useDetectVideoHasAudio.js
spacedmonkey: Don't we also need to seek the video here as well?

spacedmonkey (dismissed):
Audio detection is broken in this PR now. You need to ensure that you seek the video first before doing the check.

swissspidy:
> Audio detection is broken in this PR now. You need to ensure that you seek the video first before doing the check.

🤔 

Seeking is already happening:

https://github.com/google/web-stories-wp/blob/87c840f7798f895a6e174ef3a2183fc9578f0d22/packages/story-editor/src/components/library/panes/media/local/hotlink/useInsert.js#L111-L121

swissspidy:
I can't reproduce any issues with audio detection.

Inserting `https://c.tenor.com/Mc-DPcAOaNsAAAPo/lol-omg.mp4` -> video is marked as muted

Inserting `https://sample-videos.com/video123/mp4/720/big_buck_bunny_720p_30mb.mp4` -> video is marked as having sound

With that said, I think 50720bcb9cefe1f843ca08dfa4e52f4d9b5950d7 makes this more robust.

Please test this again and provide any steps to reproduce, including video URL used.

spacedmonkey:
Sorry I wasn't clear. Uploading media, has audio is broken.

swissspidy:
Audio detection for uploads is now fixed

packages/story-editor/src/app/media/utils/useDetectVideoHasAudio.js
spacedmonkey:
```suggestion
        const videoEl = await preloadVideo(src);
        await seekVideo(videoEl);
        const hasAudio = hasVideoGotAudio(videoEl);
```

packages/story-editor/src/app/media/utils/getResourceFromLocalFile.js
```diff
@@ -74,23 +76,23 @@ const getVideoResource = async (file) => {
...
-    const videoLength = await getVideoLength(src);
-    length = videoLength.length;
-    lengthFormatted = videoLength.lengthFormatted;
-  }
-  const posterFile = await getFirstFrameOfVideo(src);
-  const hasAudio = await hasVideoGotAudio(src);
+
+  const videoLength = getVideoLength(videoEl);
+  const length = videoLength.length;
+  const lengthFormatted = videoLength.lengthFormatted;
```
spacedmonkey, line 91:
```suggestion
  const { length, lengthFormatted } = videoLength;
```

spacedmonkey:
```suggestion
  const { length, lengthFormatted } = getVideoLength(videoEl);
```

packages/story-editor/src/app/media/utils/getResourceFromUrl.js
```diff
@@ -44,37 +45,67 @@ import {
...
+    const video = await preloadVideo(src);
+    await seekVideo(video);
+
+    additionalData.width = video.videoWidth;
+    additionalData.height = video.videoHeight;
+
+    const videoLength = getVideoLength(video);
+
+    additionalData.length = videoLength.length;
+    additionalData.lengthFormatted = videoLength.lengthFormatted;
```
spacedmonkey, line 85:
```suggestion
    const { length, lengthFormatted } = getVideoLength(video);

    additionalData.length  = length;
    additionalData.lengthFormatted = lengthFormatted;
```

swissspidy: That does not work because `length` and `lengthFormatted` are already declared.

packages/media/src/seekVideo.js
```diff
@@ -27,6 +27,7 @@ function seekVideo(video, offset = 0.99) {
   }
 
   return new Promise((resolve, reject) => {
+    video.addEventListener('suspend', reject);
```
spacedmonkey, line 30: Should we have this everywhere we adding video is a promise?

swissspidy: In my testing earlier I didn‘t see the need for it. But I can do more testing next week.

spacedmonkey: I think we do this in a follow up PR.

spacedmonkey:
This needs QA instructions. This pr has changed before now, there are UX changes, like disabling the insert button and changing of text on button etc. I would also like to see this tested, with different files.

swissspidy:
@spacedmonkey We've previously discussed moving QA to its separate ticket. There's also bug bash tomorrow which is an ideal opportunity to get even more eyes.

spacedmonkey:
> We've previously discussed moving QA to its separate ticket. There's also bug bash tomorrow which is an ideal opportunity to get even more eyes.

I think that was a old conversion, before UX was added to this ticket. Also, why would we bug bash a ticket that has even been QAed. That feels like a waste of time. Why not QA these changes?

spacedmonkey (approved):

PR #9703, opened 2021-11-12
Supporting qualifying outbound links (rel attribute)

### LINKED ISSUE #4440: Supporting qualifying outbound links (`rel` attribute)
## Feature Description

_Originally reported at https://github.com/ampproject/amphtml/issues/30140_

For certain links users might want to tell search engines their relationship with the linked page using [the `rel` attribute](https://support.google.com/webmasters/answer/96569?hl=en). Examples include:

* `rel="sponsored"` - Mark links that are advertisements or paid placements
* `rel="ugc"` - Mark links as user-generated content (UGC) such as comments and forum posts
* `rel="nofoloow"` - For when you'd rather Google not associate your site with, or crawl the linked page from, your site

These can also be combined.

## Additional Context

---

spacedmonkey:
## Summary

For certain links users might want to tell search engines their relationship with the linked page using the rel attribute. Examples include:

rel="sponsored" - Mark links that are advertisements or paid placements
rel="ugc" - Mark links as user-generated content (UGC) such as comments and forum posts
rel="nofoloow" - For when you'd rather Google not associate your site with, or crawl the linked page from, your site

## User-facing changes

https://user-images.githubusercontent.com/237508/141502140-33134a6f-a6b8-4516-bca2-628c8a7ed8e4.mov

## Testing Instructions

1. Create a story.
2. Insert image into story. 
3. In design panel, add link, example. http://www.google.com
4. Checkbox no follow or other option. 
5. Publish story. 
6. View story on front end. 
7. View source. 
8. Find `<a` tag and see rel value. 

## Checklist

---

Fixes #4440


packages/story-editor/src/components/panels/design/link/link.js
```diff
@@ -63,6 +69,30 @@ const Error = styled.span`
...
+`;
+
+const StyledText = styled(Text)`
+  color: ${({ theme }) => theme.colors.fg.secondary};
+  padding: 8px 0;
+`;
+
+const LinkTypes = [
+  { key: 'nofollow', title: __('No follow', 'web-stories') },
+  { key: 'ugc', title: __('User-generated content', 'web-stories') },
```
swissspidy: As per discussion on #4440 we don't want a checkbox for ugc

packages/story-editor/src/components/panels/design/link/link.js
```diff
@@ -63,6 +69,30 @@ const Error = styled.span`
...
+  margin: 12px 0;
+`;
+
+const StyledText = styled(Text)`
+  color: ${({ theme }) => theme.colors.fg.secondary};
+  padding: 8px 0;
+`;
+
+const LinkTypes = [
+  { key: 'nofollow', title: __('No follow', 'web-stories') },
```
swissspidy: In the mocks it's `Nofollow`, not `No follow`.

packages/story-editor/src/components/panels/design/link/link.js
```diff
@@ -201,8 +228,24 @@ function LinkPanel({ selectedElements, pushUpdateForObject }) {
...
+    [handleChange, link]
+  );
+
   const isMultipleUrl = MULTIPLE_VALUE === link.url;
   const isMultipleDesc = MULTIPLE_VALUE === link.desc;
+  const showRelOption =
+    MULTIPLE_VALUE !== link.rel && link.url && !isMultipleUrl;
+
+  const relHelpLink =
+    'https://developers.google.com/search/docs/advanced/guidelines/qualify-outbound-links?hl=en';
```
swissspidy: This should be translatable

packages/story-editor/src/components/panels/design/link/link.js
```diff
@@ -284,6 +327,40 @@ function LinkPanel({ selectedElements, pushUpdateForObject }) {
           </Row>
         </>
       )}
+      {showRelOption && (
+        <CheckboxContainer>
+          <StyledText size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL}>
+            {__('About this link.', 'web-stories')}
```
swissspidy:
```suggestion
            {__('Qualify outbound links, 'web-stories')}
```

As per https://github.com/google/web-stories-wp/issues/4440#issuecomment-956361256

packages/story-editor/src/components/panels/design/link/link.js
```diff
@@ -284,6 +327,40 @@ function LinkPanel({ selectedElements, pushUpdateForObject }) {
...
+            </CheckboxWrapper>
+          ))}
+
+          <Link
+            rel="noopener noreferrer"
+            target="_blank"
+            href={relHelpLink}
+            size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.X_SMALL}
+          >
+            {__('What are these?', 'web-stories')}
```
swissspidy: As per https://github.com/google/web-stories-wp/issues/4440#issuecomment-956361256

```suggestion
            {__('Learn more', 'web-stories')}
```

packages/story-editor/src/components/elementLink/output.js
```diff
@@ -31,14 +31,16 @@ function WithLink({ element, children, ...rest }) {
   if (!link?.url?.length) {
     return children;
   }
+  const rel = link.rel ? link.rel.join(' ') : 'noreferrer';
```
swissspidy: We need to always add `noreferrer`

So could be `noreferrer nofollow sponsored` for instance.

packages/story-editor/src/components/elementLink/output.js
```diff
@@ -31,14 +31,16 @@ function WithLink({ element, children, ...rest }) {
...
   const urlWithProtocol = withProtocol(link.url);
   return (
+    // eslint-disable-next-line react/jsx-no-target-blank
```
swissspidy, line 39: Comment explaining this would be good

packages/story-editor/src/components/elementLink/output.js
swissspidy: Let's add a unit tests for this change

packages/story-editor/src/components/panels/design/link/link.js
```diff
@@ -284,6 +327,40 @@ function LinkPanel({ selectedElements, pushUpdateForObject }) {
           </Row>
         </>
       )}
+      {showRelOption && (
```
miina: Should these options be before the icon/optional description perhaps? Otherwise it can cause quite a lot of jumpiness during validation, e.g.
jumpiness

packages/story-editor/src/components/panels/design/link/link.js
```diff
@@ -88,10 +118,7 @@ function LinkPanel({ selectedElements, pushUpdateForObject }) {
     getElementsInAttachmentArea(selectedElements).length > 0 &&
     currentPage?.pageAttachment?.url?.length > 0;
 
-  const defaultLink = useMemo(
-    () => createLink({ url: '', icon: null, desc: null }),
-    []
-  );
+  const defaultLink = useMemo(() => createLink({ icon: null, desc: null }), []);
```
miina, line 92: Why the change here?

packages/story-editor/src/components/panels/design/link/link.js
```diff
@@ -201,8 +228,24 @@ function LinkPanel({ selectedElements, pushUpdateForObject }) {
...
+        ? link.rel.filter((el) => el !== value)
+        : [...link.rel, value];
+      handleChange({ rel }, true);
+    },
+    [handleChange, link]
+  );
+
   const isMultipleUrl = MULTIPLE_VALUE === link.url;
   const isMultipleDesc = MULTIPLE_VALUE === link.desc;
+  const showRelOption =
```
miina: We already have `displayMetaFields` for detecting if the URL is set / if it's valid, maybe just use that here as well instead of `link.url`? Or maybe the whole new checkbox area should also be inside the `displayMetaFields` condition when rendering as the rest of the meta fields. Thoughts?

packages/story-editor/src/components/panels/design/link/link.js
spacedmonkey: CreateLink, will default url to an empty string for you. So this not needed.

packages/story-editor/src/components/elementLink/output.js
spacedmonkey: Fixed.

spacedmonkey: Fixed.

packages/story-editor/src/components/panels/design/link/link.js
spacedmonkey: Fixed.

packages/story-editor/src/components/panels/design/link/link.js
spacedmonkey: Fixed.

spacedmonkey: I was working of the figma.

packages/story-editor/src/components/panels/design/link/link.js
spacedmonkey: Fixed.

packages/story-editor/src/components/panels/design/link/link.js
spacedmonkey: Fixed.

packages/story-editor/src/components/panels/design/link/link.js
spacedmonkey: Removed.

packages/story-editor/src/components/elementLink/output.js
spacedmonkey: Fixed.

packages/story-editor/src/components/panels/design/link/link.js
spacedmonkey: Yeah, good point. I am pushed a fix.

packages/story-editor/src/components/panels/design/link/link.js
spacedmonkey: Fixed.

packages/story-editor/src/components/panels/design/link/link.js
miina: Nit: If instead we'd wrap all this into `displayMetaFields` as well together with the `Row` above, `showRelOptions` could be just `MULTIPLE_VALUE !== link.rel`.

Not a big difference but might improve readability.

packages/story-editor/src/components/panels/design/link/link.js
```diff
@@ -282,6 +321,43 @@ function LinkPanel({ selectedElements, pushUpdateForObject }) {
               </IconText>
             </IconInfo>
           </Row>
+          {!isMultipleRel && (
```
miina, line 292: Should this be added to the Page Attachment as well? (Per Pascal't comment: https://github.com/google/web-stories-wp/issues/4440#issuecomment-963277350)

spacedmonkey: Added.

packages/story-editor/src/components/elementLink/output.js
```diff
@@ -31,14 +31,17 @@ function WithLink({ element, children, ...rest }) {
...
   return (
+    // False positive.
+    // eslint-disable-next-line react/jsx-no-target-blank
     <a
       href={urlWithProtocol}
       data-tooltip-icon={link.icon}
       data-tooltip-text={link.desc}
       target="_blank"
-      rel="noreferrer"
+      rel={`${rel} noreferrer`}
```
swissspidy: Nit: when `rel` is empty, this will result in `rel=" noreferrer"`.

Might want to trim. And perhaps add a test.

packages/story-editor/src/components/elementLink/output.js
```diff
@@ -31,14 +31,17 @@ function WithLink({ element, children, ...rest }) {
   if (!link?.url?.length) {
     return children;
   }
+  const rel = link?.rel.join(' ') ?? '';
```
swissspidy: I don't think the null coalescing operator does anything here.

If `rel` is not an array, this will throw an error.

If it's an empty array, it's never null, so the `''` fallback is never hit.

This would be safer:

```suggestion
  const rel = link?.rel?.join(' ') || '';
```

packages/story-editor/src/components/elementLink/test/output.js
```diff
@@ -67,6 +68,20 @@ describe('WithLink', () => {
...
+    });
+
+    it('should use rel=noreferrer nofollow', async () => {
+      render(withLink({ rel: ['nofollow'] }));
+      const a = screen.getByRole('link');
+      await expect(a.rel).toBe('nofollow noreferrer');
+    });
+  });
+
   describe('AMP validation', () => {
```
swissspidy, line 85: Bonus points for also adding a test case here for AMP validation

packages/story-editor/src/components/panels/design/link/link.js
```diff
@@ -46,6 +52,7 @@ import {
   useCommonObjectValue,
 } from '../../shared';
 import { states, styles, useHighlights } from '../../../../app/highlights';
+import { LinkTypes, relHelpLink, relHelpText, relTitle } from '../constants';
```
swissspidy: FYI: Lint error

packages/story-editor/src/components/panels/design/pageAttachment/pageAttachment.js
```diff
@@ -285,6 +307,39 @@ function PageAttachmentPanel() {
               {__('Link icon', 'web-stories')}
             </Text>
           </StyledRow>
+          <Row>
```
miina: If it's exactly the same here as in the Link panel, should we perhaps reuse instead of duplicating? Maybe even combine it with the `LinkInput` component?

spacedmonkey: The link and page attachment panels are like this in general. It should be more generic component. 

Anyway, I have refactored the logic to be it's own component.

packages/story-editor/src/components/elementLink/output.js
```diff
@@ -27,18 +27,23 @@ import { withProtocol } from '../../utils/url';
...
-  const link = getLinkFromElement(element);
-  if (!link?.url?.length) {
+  const link = getLinkFromElement(element) || {};
+  const { url, icon, desc, rel = [] } = link;
+  if (!url) {
     return children;
   }
-  const urlWithProtocol = withProtocol(link.url);
+  rel.push('noreferrer');
+  const relOutput = rel.join(' ').trim();
```
swissspidy: Now that you're using `push()` above to add `noreferrer`, no need for the `trim` anymore.

```js
['noreferrer'].join(' ') // 'noreferrer'
['noreferrer', 'foo'].join(' ') // 'noreferrer foo'
[].join(' ') // ''
```

packages/story-editor/src/components/elementLink/test/output.js
```diff
@@ -67,9 +68,26 @@ describe('WithLink', () => {
...
     it('should produce valid AMP output', async () => {
       await expect(withLink()).toBeValidAMPStoryElement();
     });
+    it('should produce link with rel valid AMP output', async () => {
```
swissspidy:
```suggestion
    it('should produce valid AMP output for link with rel', async () => {
```

miina (approved):

packages/story-editor/src/components/panels/design/link/link.js
```diff
@@ -201,8 +199,20 @@ function LinkPanel({ selectedElements, pushUpdateForObject }) {
...
 
+  const onChangeRel = useCallback(
+    (value) => {
+      const rel = link?.rel?.includes(value)
+        ? link?.rel?.filter((el) => el !== value)
+        : [...link?.rel, value];
+      handleChange({ rel }, true);
+    },
+    [handleChange, link]
+  );
```
swissspidy: All the optional chaining here is superfluous and not done anywhere else in this file.

If `link` or `link.rel` can be `null`, then `[...link?.rel, value]` (same as `[...null, value]`) would throw an error.

Assuming that `link.rel` can be `null` for existing links in existing stories, then it'd be safer to write something like 

```js
const linkRel = link.rel || [];
const rel = linkRel.includes(value) ? linkRel.filter((el) => el !== value) : [...linkRel, value];
handleChange({ rel }, true);
```

packages/story-editor/src/components/panels/design/link/link.js
```diff
@@ -282,6 +292,9 @@ function LinkPanel({ selectedElements, pushUpdateForObject }) {
               </IconText>
             </IconInfo>
           </Row>
+          {!isMultipleRel && (
+            <LinkRelations onChangeRel={onChangeRel} rel={link?.rel || []} />
```
swissspidy: Same here, optional chaining is superfluous, `link` doesn't seem to be nullable.

swissspidy (approved):
Just a couple of nits

packages/story-editor/src/components/panels/design/link/link.js
spacedmonkey: Went a different direction with this.

PR #9706, opened 2021-11-12
Context Menu: Placement Update

### LINKED ISSUE #9563: Right Click Menu Animation
## Feature Description

Brought up by @agingoldseco in bug bash:
> I think I’d like to see the right click menu fade in a little faster—it feels like it’s delayed when that may just be an animation choice? Or it is actually slightly slow and the animation makes it feel a bit slower 

Eng to work with design to figure out improvements. 

### Update from design sync 11/9/2021

Going to remove the animation completely, set the default direction of the menu to open _down_ and if there's not room for that, have it open _up_, justified to the bottom right (LTR) - left (RTL) -  corner.

maxyinger:
## Context
Part of design feedback on the current context menu.

## Summary
- removes animation from context menu
- mimics chromes positioning:
    - bottom-right justification of cursor by default
    - stick to bottom if not enough room vertically
    - switches justification to bottom-left of cursor if not enough room horizontally

## Relevant Technical Choices
Using a callback ref with some css vars was quick and effective here. Is clean, simple and don't have to worry about having any frames where the menu is open, but not positioned correctly while state is taking a render to set

## To-do
We could attempt to do a karma test where we click on the layers panel and see if this causes any overflow scroll to the body, but feels superfluous to me. Thoughts?

## User-facing changes
- right click menu will now appear to the bottom-right of the cursor (unless it can't fit in the viewport)
- right click menu will no longer animate open.

## Testing Instructions

1. Open the editor
2. Make sure there's some elements in the story
3. Right click an element and see that the context menu no longer has an animation on open
4. See that the right click menu is positioned to the bottom right of the cursor by default (like chrome)

## Checklist

---

Fixes #9563

### DIFF
--- a/packages/design-system/src/components/contextMenu/contextMenu.js
+++ b/packages/design-system/src/components/contextMenu/contextMenu.js
@@ -30,6 +30,7 @@ const ContextMenu = ({
   animate,
   isAlwaysVisible,
   items,
+  isRTL,
   isInline = false,
   ...props
 }) => {
@@ -44,6 +45,7 @@ const ContextMenu = ({
         isInline={isInline}
         role={isAlwaysVisible ? null : 'dialog'}
         isOpen={isAlwaysVisible || props.isOpen}
+        isRTL={isRTL}
       >
         <Menu aria-expanded={props.isOpen} items={items} {...props} />
         {/* <AnimationContainer /> has a <Shadow />. Don't double the shadow. */}
@@ -58,6 +60,7 @@ ContextMenu.propTypes = {
   animate: PropTypes.bool,
   isOpen: PropTypes.bool,
   isAlwaysVisible: PropTypes.bool,
+  isRTL: PropTypes.bool,
 };
 
 export default ContextMenu;
--- a/packages/design-system/src/components/contextMenu/contextMenu.js
+++ b/packages/design-system/src/components/contextMenu/contextMenu.js
@@ -21,7 +21,7 @@ import { useMemo } from '@web-stories-wp/react';
 /**
  * Internal dependencies
  */
-import { Popover, Shadow } from './styled';
+import { SmartPopover, Shadow } from './styled';
 import Menu, { MenuPropTypes } from './menu';
 import AnimationContainer from './animationContainer';
 import Mask from './mask';
@@ -34,7 +34,7 @@ const ContextMenu = ({
   ...props
 }) => {
   const Wrapper = useMemo(
-    () => (animate ? AnimationContainer : Popover),
+    () => (animate ? AnimationContainer : SmartPopover),
     [animate]
   );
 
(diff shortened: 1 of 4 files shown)


packages/design-system/src/components/contextMenu/contextMenu.js
```diff
@@ -21,14 +21,14 @@ import { useMemo } from '@web-stories-wp/react';
...
+import { SmartPopover, Shadow } from './styled';
 import Menu, { MenuPropTypes } from './menu';
 import AnimationContainer from './animationContainer';
 import Mask from './mask';
 
-const ContextMenu = ({ animate, isAlwaysVisible, items, ...props }) => {
+const ContextMenu = ({ animate, isAlwaysVisible, items, isRTL, ...props }) => {
   const Wrapper = useMemo(
-    () => (animate ? AnimationContainer : Popover),
+    () => (animate ? AnimationContainer : SmartPopover),
```
samwhale, line 38: :smart:

packages/design-system/src/components/contextMenu/styled.js
```diff
@@ -47,3 +43,43 @@ export const Shadow = styled.div`
...
+    },
+    [props.isRTL]
+  );
+
+  return props.isOpen && <Popover ref={setRef} {...props} />;
+}
+
+SmartPopover.propTypes = {
+  isOpen: PropTypes.bool,
+  isRTL: PropTypes.bool,
```
samwhale, line 84: Would we need `isRTL` here if we use the `DirectionAware` component? I see that the math is being done manually here so we probably can't rely on the magic of the `DirectionAware`

packages/story-editor/src/components/canvas/rightClickMenu.js
```diff
@@ -58,23 +60,25 @@ const RightClickMenu = () => {
   }, [ref]);
 
   return createPortal(
-    <DirectionAware>
+    <StyleSheetManager stylisPlugins={[]}>
```
samwhale, line 63: This is needed because of the portal, right?

samwhale (approved):
Works great!

packages/design-system/src/components/contextMenu/styled.js
maxyinger: yeaup exactly

packages/story-editor/src/components/canvas/rightClickMenu.js
maxyinger: yea if you check it out rtl on staging right now, and try and click the layers panel and stuff, you can see the positioning bug this is fixing. 

Basically we're explicitly setting the coordinate based on mouse click, but if the positioning component is direction aware, it changes `right` -> `left`. So this removes that weird behavior.

BrittanyIRL (approved):
this is great!

PR #9707, opened 2021-11-15
Color Picker: Fix navigation with arrow keys

### LINKED ISSUE #9701: Color Picker: fix navigation with arrow keys
## Bug Description
Since adding a `Tooltip` wrapper around the `StyledSwatch` component, navigation with arrow keys doesn't work anymore.

## Expected Behaviour
The user should be able to navigate between the color swatches using arrow keys.

## Steps to Reproduce
- Open the color picker after ensuring it has a few saved colors
- Try to navigate between the colors using arrow keys.

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

miina:
## Summary
Fixes navigation by arrow keys in the color picker by adding the `Tooltip` wrapper only for disabled swatches.

## Testing Instructions

1. Open the color picker
2. Verify it's possible to navigate between the colors in a group of colors using arrow keys.

No

No

No

## Checklist

---

Fixes #9701

### DIFF
--- a/packages/story-editor/src/components/colorPicker/basicColorList.js
+++ b/packages/story-editor/src/components/colorPicker/basicColorList.js
@@ -72,6 +72,18 @@ function getPatternAsString(pattern) {
   return styles.backgroundColor || styles.backgroundImage;
 }
 
+function ConditionalTooltip({ tooltip, children }) {
+  if (!tooltip) {
+    return children;
+  }
+  return <Tooltip title={tooltip}>{children}</Tooltip>;
+}
+
+ConditionalTooltip.propTypes = {
+  tooltip: PropTypes.string,
+  children: PropTypes.node.isRequired,
+};
+
 function BasicColorList({
   color,
   colors,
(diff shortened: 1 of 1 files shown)


packages/story-editor/src/components/colorPicker/basicColorList.js
```diff
@@ -135,19 +135,25 @@ function BasicColorList({
...
-                  isDisabled={isDisabled}
-                  tabIndex={tabIndex}
-                  title={title}
-                >
-                  {isEditMode && <Icons.Cross />}
-                </StyledSwatch>
+                {renderedSwatch}
               </Tooltip>
+            ) : (
+              renderedSwatch
```
barklund: I think I would prefer something like a <MaybeTooltip /> component in between - so this would just be:

```js
          return (
            <MaybeTooltip key={patternAsBackground} tooltip={tooltip}>
              <StyledSwatch
                onClick={() => handleClick(pattern, isLocal)}
                pattern={pattern}
                isSelected={isSelected}
                isDisabled={isDisabled}
                tabIndex={tabIndex}
                title={title}
              >
                {isEditMode && <Icons.Cross />}
              </StyledSwatch>
            </MaybeTooltip>
          );
```

And then `MaybeTooltip` would be:

```js
function MaybeTooltip({ tooltip, children }) {
  if (!tooltip) return children;
  return <Tooltip title={tooltip}>{children}</Tooltip>
}
```

It simplifies the code a bit and also makes the `key` a bit simpler as it only goes one place.

Naming is not great though - please be more creative than me 😄 .

Alternatively, this could be baked into the existing tooltip component or be made a new component alongside the tooltip but still inside the design system. Maybe an `OptionalTooltip` with a `hasTooltip` property (so it doesn't generally directly depend on the truthiness of the `tooltip` property).

This might be useful in other places in the codebase as well, as I seem to remember similar optional tooltip rendering elsewhere.

barklund (approved):
Works great. I think it could deserve generalization though, as I seem to remember we have similar logic elsewhere in the code. But if that's a lot of work, it could be a new ticket of course.

packages/story-editor/src/components/colorPicker/basicColorList.js
miina: Sure! I actually tried out that version as well (with `ConditionalTooltip` as the name) but somehow liked the current one more. I'm 50-50 though so if you prefer this one, let's go with that!

PR #9710, opened 2021-11-15
Decoupling: Refactor and assemble checklist in `wp-story-editor`

mohdsayed:
## Context

Refactor and assemble checklist in `wp-story-editor` as discussed in https://github.com/google/web-stories-wp/pull/9569#discussion_r739065299

[](https://github.com/google/web-stories-wp/pull/9569#discussion_r739065299)

## Testing Instructions

1.

## Checklist

---


packages/story-editor/src/defaultConfig.js
```diff
@@ -110,7 +110,6 @@ const defaultConfig = {
   },
   isRTL: false,
   storyId: 1,
-  // @todo https://github.com/google/web-stories-wp/pull/9569#discussion_r739066940
   dashboardSettingsLink: '',
```
mohdsayed: @swissspidy Can you elaborate more on what you want to do with this `dashboardSettingsLink`.  `PublishPanel` is moved and `VideoOptimizationCheckbox` is being assembled with other checks in `wp-story-editor`

https://github.com/google/web-stories-wp/pull/9569#discussion_r739066940

swissspidy: That's a good question 🙂 

`VideoOptimizationCheckbox` is just a way to tell the user they need to enable a setting so that they can optimize videos. Some platforms might not even have a setting for that (e.g. in a self-hosted playground it could always be enabled).

So I think it's safe to just move this component to `wp-story-editor` entirely.

mohdsayed: Moved in https://github.com/google/web-stories-wp/pull/9710/commits/fe65b43573491cf9783cbcbe481b329e78cba31d

swissspidy: So can we now remove `dashboardSettingsLink` from `defaultConfig`?

packages/wp-story-editor/src/components/checklist/accessibility/videoOptimizationCheckbox/index.js
```diff
@@ -25,18 +25,17 @@ import {
...
+  useCurrentUser,
+  useIsChecklistMounted,
+} from '@web-stories-wp/story-editor';
 
 const Container = styled.div`
   display: flex;
   align-items: center;
   flex-direction: column;
-  width: ${CARD_WIDTH}px;
+  width: 272px;
```
swissspidy: I wonder why this width was used in the first place.

Other checks don't need it because they use `<ChecklistCard>`.

Could we use `<ChecklistCard>` here as well instead of `<Container>`? Any style differences?

I'd want to avoid hardcoding 272px here.

packages/wp-story-editor/src/components/checklist/priority/index.js
```diff
@@ -0,0 +1,67 @@
...
+      {canManageSettings && <StoryMissingPublisherName />}
+      <StoryTitleLength />
+      <StoryMissingExcerpt />
+      {hasUploadMediaAction && <StoryPosterAttached />}
+      {hasUploadMediaAction && <StoryPosterSize />}
+      {hasUploadMediaAction && <PublisherLogoMissing />}
+      {hasUploadMediaAction && <PublisherLogoSize />}
+      {hasUploadMediaAction && <VideoElementMissingPoster />}
+      {/* `isTranscodingEnabled` already checks for `hasUploadMediaAction` */}
+      {isTranscodingEnabled && <VideoOptimization />}
```
swissspidy: If we'd move this `isTranscodingEnabled` check within `VideoOptimization` then we wouldn't have to import `useFFmpeg` here.

Though that could result in some unnecessary renders 🤔

packages/wp-story-editor/src/components/checklist/accessibility/videoOptimizationCheckbox/index.js
mohdsayed: After and before respectively 👇 

image

Done https://github.com/google/web-stories-wp/pull/9710/commits/de15679a9ea6259eacb4228bce2d9ee6dbbeb649

packages/wp-story-editor/src/components/checklist/priority/index.js
mohdsayed: Yeah, there [are quite a few things happening](https://github.com/google/web-stories-wp/blob/main/packages/story-editor/src/components/checklist/checks/videoOptimization.js#L114-L195) in that component, I would want to avoid running those. But I have a middle ground, I created a wrapper component for `VideoOptimization`, and this way we would not have to export `useFFmpeg`.

Done in https://github.com/google/web-stories-wp/pull/9710/commits/8553a4356e5bb5129c57ee0cb3c6f7b161b95732

mohdsayed:
E2E "Inserting 3P Media › should insert an Unsplash image" is failing here also.

mohdsayed:
@swissspidy  E2E is probably failing due to this PR https://github.com/google/web-stories-wp/pull/9757/files

swissspidy:
That PR only updated Karma fixtures and did not touch anything else though 🤔

swissspidy:
Does it fail locally for you? Would be good to know what triggers that 404 error.

Perhaps it's just flakey, or maybe it's actually a bug.

swissspidy (approved):

PR #9713, opened 2021-11-15
Use CORS Proxy for Link Icons

### LINKED ISSUE #9492: Use CORS Proxy for Link Icons
## Feature Description

For page attachments and links we automatically fetch the site's favicon to display it.

That works fine on the frontend, but in the editor, when video optimization is enabled (and thus cross-origin isolation), it often doesn't work because of lacking CORS headers:

Screenshot 2021-10-21 at 16 49 47
Screenshot 2021-10-21 at 16 49 33
Screenshot 2021-10-21 at 16 49 28

By using our new CORS proxy for that we could easily fix this.

Only needed when video optimization is enabled though. Otherwise it works fine.

spacedmonkey:
## Summary

Ensure that link / attachment icons are loaded via the cors proxy. 

## Relevant Technical Choices

Added new property of needsProxy to links, to "cache" the result of the head request, similar to how it currently works 

## Testing Instructions

#### Element links. 

1. Open story editor
2. Insert an image into page.
3. In documents side panel, under link, add a link to website. ( try different links )
4. See icons display. 

#### Page attachment links.
1. Open story editor
2. Click on canvas / page. 
3. In documents side panel, under page attachment, add a link to website. ( try different links )
4.  See icons display. 

## Checklist

---

Fixes #9492


packages/story-editor/src/components/panels/design/link/link.js
```diff
@@ -201,6 +204,8 @@ function LinkPanel({ selectedElements, pushUpdateForObject }) {
     link.url,
   ]);
 
+  const linkIcon = getProxiedUrl({ needsProxy: true }, encodeURI(link.icon));
```
spacedmonkey: @swissspidy In this PR, I am forcing all link icon request from through the proxy. But this may not be needed in some cases. We could add a new property to link object, called `needsProxy`. This maybe overkill. What do you think?

spacedmonkey:
Submitting as draft for now, as https://github.com/google/web-stories-wp/pull/9702 is blocking this change.

packages/story-editor/src/components/panels/design/link/link.js
swissspidy: It may be overkill, but we went the same route with hotlinking as well. So let's do the same thing here for consistency and to avoid sending _everything_ through the proxy.

We can move the `checkResourceAccess` function from `useInsert.js` to `useCORSProxy.js` and then use that here.

packages/story-editor/src/utils/checkResourceAccess.js
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Makes a HEAD request, which in turn triggers a CORS preflight request
+ * in the browser.
+ *
+ * If the request passes, we don't need to do anything.
+ * If it doesn't, it means we need to run the resource through our CORS proxy at all times.
+ *
+ * @type {function(): boolean}
+ */
+async function checkResourceAccess(link) {
```
swissspidy: My recommendation was to move this function within `useCORSProxy` hook as it's related to that. Whats' the rationale for making this a standalone util function?

spacedmonkey: You didn't make a recommendation. There is no comment on this PR. It makes sense a util, makes it re-usable.

spacedmonkey: To be clear, we need this as a function, as need to value from it to cache it, in the resource, link and page attachment.

swissspidy: > You didn't make a recommendation. There is no comment on this PR. It makes sense a util, makes it re-usable.

https://github.com/google/web-stories-wp/pull/9713#discussion_r750198995

swissspidy: > To be clear, we need this as a function, as need to value from it to cache it, in the resource, link and page attachment.

I don't understand how moving this function to a different file causes problems with any of this.

spacedmonkey: I don't see the point of it is living in that file. You just like it there?

swissspidy: Checking resource access for proxying and getting the proxy URL are coupled, you wouldn't use one without the other. It makes sense for this functionality to live together.

spacedmonkey: Moved.

swissspidy (approved):
Needs rebase and some QA instructions, otherwise looks good

swissspidy (approved):

PR #9715, opened 2021-11-15
Editor: Update Karma fixture to fix flakey `waitFor()`

### LINKED ISSUE #9381: Karma: Flakey drop target integration test
## Task Description

The following test appears to be flakey:

https://github.com/google/web-stories-wp/blob/38ba121a8bd6062e85290ccca6c80a298cea5910/packages/story-editor/src/components/dropTargets/karma/dropTarget.karma.js#L41

Example run:

https://github.com/google/web-stories-wp/runs/3868257464#step:7:5681

### LINKED ISSUE #9387: Karma: Flakey PageMenu test
## Task Description

The following test appears to be flakey:

https://github.com/google/web-stories-wp/blob/fe6e87efa45dfd30a26db8afb40fe56e22835741/packages/story-editor/src/components/canvas/karma/pageMenuActions.karma.js#L86-L107

Example run:

https://github.com/google/web-stories-wp/runs/3869354710?check_suite_focus=true#step:7:5838

### LINKED ISSUE #9389: Karma: Flakey gradient overlay test
## Task Description

The following test appears to be flakey:

https://github.com/google/web-stories-wp/blob/ded12add6848c9dc6706d86462d1e5fa5f215dc1/packages/story-editor/src/components/panels/design/filter/karma/filter.karma.js#L126-L135

Example run:

https://github.com/google/web-stories-wp/runs/3871620200?check_suite_focus=true#step:7:5744

BrittanyIRL:
## Context

Solvin' some flakey tests. 

## Summary

Collapsing the help center while rendering the fixture for story editor karma tests has become responsible for some flakey tests that were tragically disabled 😿 . This removes that and adds a function available on the fixture to collapse the help center when desired. Now when you want to collapse the help center just `await fixture.collapseHelpCenter()`

## Relevant Technical Choices

- Updates karma tests to collapseHelpCenter after render in `beforeEach` where applicable (pretty much everywhere except the karma tests for the help center and popups - now those are actually testing the initial state!) 
- Updates the popups tests now that help center can start expanded.
- carousel clickOnThumbnail was being flakey, gave it a tiny sleep (1 second) and added `await` to the scroll into view and it's working great now

## User-facing changes

None

## Testing Instructions

Do the karma tests for story editor now pass? if yes, then we're good.

1.

## Checklist

---

Fixes #9389
Fixes #9387
Fixes #9381

### DIFF
--- a/packages/story-editor/src/app/history/karma/history.karma.js
+++ b/packages/story-editor/src/app/history/karma/history.karma.js
@@ -25,6 +25,7 @@ describe('CUJ: Creator can View and Modify Document Settings: Navigating without
   beforeEach(async () => {
     fixture = new Fixture();
     await fixture.render();
+    await fixture.collapseHelpCenter();
   });
 
   afterEach(() => {
--- a/packages/story-editor/src/components/canvas/karma/carousel.karma.js
+++ b/packages/story-editor/src/components/canvas/karma/carousel.karma.js
@@ -76,15 +76,15 @@ fdescribe('Carousel integration', () => {
   }
 
   async function clickOnThumbnail(index) {
-    await fixture.events.sleep(300);
+    await fixture.events.sleep(100);
     const { pages } = await fixture.editor.footer.carousel;
-    expect(pages[index]).toBe('not this but i need to see inside here remote!');
+    // expect(pages[index]).toBe('not this but i need to see inside here remote!');
     const thumb = await pages[index];
     await expect(thumb).toBeDefined();
 
     const thumbNode = !thumb?.node ? thumb._node : thumb.node;
     await expect(thumbNode).toBeDefined();
-    await thumbNode?.scrollIntoView();
+    await thumbNode.scrollIntoView();
     await fixture.events.mouse.clickOn(thumbNode, 5, 5);
   }
 
--- a/packages/story-editor/src/components/canvas/karma/carousel.karma.js
+++ b/packages/story-editor/src/components/canvas/karma/carousel.karma.js
@@ -77,11 +77,12 @@ fdescribe('Carousel integration', () => {
 
   async function clickOnThumbnail(index) {
     const { pages } = await fixture.editor.footer.carousel;
-    expect(pages).toBe('not this but i need to see inside here remote!');
+    // expect(pages).toBe('not this but i need to see inside here remote!');
     const thumb = pages[index];
-    await expect(thumb).toBeDefined();
-    await thumb.node.scrollIntoView();
-    await fixture.events.mouse.clickOn(thumb.node, 5, 5);
+    const thumbNode = thumb.node || thumb._node;
+    await expect(thumbNode).toBeDefined();
+    await thumbNode.scrollIntoView();
+    await fixture.events.mouse.clickOn(thumbNode, 5, 5);
   }
 
   it('should select the current page', async () => {
--- a/packages/story-editor/src/components/canvas/karma/carousel.karma.js
+++ b/packages/story-editor/src/components/canvas/karma/carousel.karma.js
@@ -17,7 +17,6 @@
 /**
  * External dependencies
  */
-import { waitFor } from '@testing-library/react';
 import { createSolid } from '@web-stories-wp/patterns';
 /**
  * Internal dependencies
@@ -27,7 +26,7 @@ import { useStory } from '../../../app/story';
 import { useInsertElement } from '..';
 import { TEXT_ELEMENT_DEFAULT_FONT } from '../../../app/font/defaultFonts';
 
-fdescribe('Carousel integration', () => {
+describe('Carousel integration', () => {
   let fixture;
   let element1;
 
@@ -78,14 +77,12 @@ fdescribe('Carousel integration', () => {
   async function clickOnThumbnail(index) {
     await fixture.events.sleep(100);
     const { pages } = await fixture.editor.footer.carousel;
-    // expect(pages[index]).toBe('not this but i need to see inside here remote!');
-    const thumb = await pages[index];
+    // node is found fine running locally while _node is present when running remote.
+    const page = await pages[index];
+    const thumb = !page?.node ? page._node : page.node;
     await expect(thumb).toBeDefined();
-
-    const thumbNode = !thumb?.node ? thumb._node : thumb.node;
-    await expect(thumbNode).toBeDefined();
-    await thumbNode.scrollIntoView();
-    await fixture.events.mouse.clickOn(thumbNode, 5, 5);
+    await thumb.scrollIntoView();
+    await fixture.events.mouse.clickOn(thumb, 5, 5);
   }
 
   it('should select the current page', async () => {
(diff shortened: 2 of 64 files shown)


BrittanyIRL:
```
Carousel integration should click into carousel on the first page FAILED
	TypeError: Cannot read properties of undefined (reading 'node')
	    at clickOnThumbnail (webpack:///packages/story-editor/src/components/canvas/karma/carousel.karma.js:80:11 <- packages/story-editor/src/karma-tests.js:499974:11)
```

Screen Shot 2021-11-15 at 3 47 30 PM


```
Checklist integration checklist should have no aXe accessibility violations should pass accessibility tests with a open non-empty checklist FAILED
	Error: Expected 0 to be 2.
	    at <Jasmine>
	    at webpack:///packages/story-editor/src/components/checklist/karma/checklist.karma.js:54:61 <- packages/story-editor/src/karma-tests.js:574647:61
```
Screen Shot 2021-11-15 at 3 48 02 PM

Hey @miina I'm curious if you've seen these failing the past few days with the carousel re-work ya'll are doing? I tried a new approach to collapse the help center - i moved it so that we can collapse when we want and not have it part of the actual render process. So now you have:
```
beforeEach(async () => {
await fixture.render();
...
})
```

I'll keep digging deeper but i wanted to get your eyes on it in case i'm attempting to solve for something that you or morten are already on top of.

miina:
> I'll keep digging deeper but i wanted to get your eyes on it in case i'm attempting to solve for something that you or morten are already on top of.

Thanks for tagging! I haven't personally seen these tests failing and as far as I know then Morten hasn't really started making changes to the carousel itself yet, even though he did do some refactoring and renaming. @barklund Could you confirm / take a look at the failures Brittany brought up to check if you see it familiar / if your work so far might have influenced the failures?

miina:
@BrittanyIRL I started seeing Carousel integration test failures now on an unrelated PR, example failure: https://github.com/google/web-stories-wp/runs/4224632638?check_suite_focus=true#step:7:4719
cc @barklund FYI

barklund:
> > I'll keep digging deeper but i wanted to get your eyes on it in case i'm attempting to solve for something that you or morten are already on top of.
> 
> Thanks for tagging! I haven't personally seen these tests failing and as far as I know then Morten hasn't really started making changes to the carousel itself yet, even though he did do some refactoring and renaming. @barklund Could you confirm / take a look at the failures Brittany brought up to check if you see it familiar / if your work so far might have influenced the failures?

Thanks for flagging. I haven't noticed any such failure, but I'll keep an eye on it in my work with the carousel!

packages/story-editor/src/components/canvas/karma/carousel.karma.js
```diff
@@ -74,10 +75,14 @@ describe('Carousel integration', () => {
...
   async function clickOnThumbnail(index) {
-    await fixture.editor.footer.carousel.waitReady();
-    const thumb = fixture.editor.footer.carousel.pages[index];
-    thumb.node.scrollIntoView();
-    await fixture.events.mouse.clickOn(thumb.node, 5, 5);
+    await fixture.events.sleep(100);
+    const { pages } = await fixture.editor.footer.carousel;
+    // node is found fine running locally while _node is present when running remote.
+    const page = await pages[index];
+    const thumb = !page?.node ? page._node : page.node;
```
swissspidy: That is odd 🤔 

@barklund Do you perhaps know what's up there?

I'd want to avoid checks like that wherever possible, so curious to know what's going on here.

BrittanyIRL: yeah odd right? i wanted to see if this solves the failures. this is what a `PageThumb` looks like when run remote: 
```
PageThumb({
      _node: (
        <button
          aria-label="Page 1 (current page)"
          tabindex="0"
          role="option"
          data-page-id="page1"
          width="40"
          height="60"
          class="pagepreview__Page-sc-i45m4c-0 dGxJop"
        >
          ...
        </button>
      ),
      _path: 'editor.footer.carousel.pages[page1]',
    }),
```

BrittanyIRL: anyways if this next run works i'll continue down this path to figure out the discrepancy and remove the ternary.

swissspidy: `_node` and `_path` are "internal" class properties on the `Container` class (`packages/story-editor/src/karma/fixture/containers/container.js`).

`node` and `path` are just getters for `_node` and `_path`. So `PageThumb.node` _should_ actually just return `_node`.

Are you saying that `page.node` is `null`/`undefined`?

BrittanyIRL: yeah, they were at least before i added the sleep.  wanted to see if this allowed the tests to be successful. it did so now it's time to remove and make it work consistently. fear not 🙃  it won't stay. it just only fails remotely so i have to keep pushing up to try things.

packages/story-editor/src/app/history/karma/history.karma.js
```diff
@@ -25,6 +25,7 @@ describe('CUJ: Creator can View and Modify Document Settings: Navigating without
   beforeEach(async () => {
     fixture = new Fixture();
     await fixture.render();
+    await fixture.collapseHelpCenter();
```
miina, line 28: @BrittanyIRL I remember it was under discussion previously as well, but can't recall right now exactly what was the conclusion -- thoughts on just closing it by setting the local storage? I know it's not the correct scenario for when the user opens the editor the first time but one could argue that most of the times, the user would actually use the editor with the Help Center already closed, and it happens less often that the user uses the editor for the first time with the Help Center opened, so this could perhaps be tested separately instead by resetting the local storage for those specific tests.

Thoughts?

BrittanyIRL: yes, so this was what i tried initially when we ended up with adding it to the fixture! 

The problem is that when you first load the editor (which we treat every time as the first time for per test for consistency) that the help center will show up because that local storage isn't there. Then, if we update the value to have isOpen be false then we'd need to _reload_ the entire thing to get the help center to close. Whereas with this proposed solution, it's just a click that's waited for before going on our merry way to run those sweet sweet tests.

packages/story-editor/src/components/canvas/karma/carousel.karma.js
```diff
@@ -74,9 +75,10 @@ describe('Carousel integration', () => {
   }
 
   async function clickOnThumbnail(index) {
+    await fixture.events.sleep(100);
     await fixture.editor.footer.carousel.waitReady();
     const thumb = fixture.editor.footer.carousel.pages[index];
-    thumb.node.scrollIntoView();
+    await thumb.node.scrollIntoView();
```
swissspidy: Since when is [`scrollIntoView`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) async?

swissspidy: I also see `await vaPanel.panelTitle.scrollIntoView();` in `videoPoster.karma.js`, which would also be incorrect

BrittanyIRL: thanks, yah - removed both of 'em.

miina (approved):

swissspidy (approved):

PR #9721, opened 2021-11-16
Smart text style: ignore selected element in contrast calculation

### LINKED ISSUE #9546: Smart Text Color re-triggering includes the element itself to the reference image
## Bug Description
When generating an image for contrast calculation, the text element itself is also included into the image, that causes unexpected and inaccurate results.

## Expected Behaviour
The element itself should not be included into the generated image.

## Steps to Reproduce
1. Add a text element on a white background
2. Click on the "Magic wand" button for auto-styling.
3. The text should stay black as it is but it gets a highlight instead.

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

Trapsta:
## Context

When generating an image for contrast calculation, the text element itself is also included into the image, that causes unexpected and inaccurate results.

## Summary

Improve how the pixel data for contrast calculation is generated before calculating the contrast by preventing the text element from being included in the pixel data that we use to calculate the contrast. 

## Relevant Technical Choices

The [html-to-image](https://github.com/bubkoo/html-to-image) library we use to extract the pixels for contrast calculation provides a filter option which we can use to filter out nodes from the DOM which we want to exclude from the generated canvas.

## To-do

n/a

## User-facing changes
Before
Screen-Recording-2021-11-16-at-0

After
Screen-Recording-2021-11-16-at-0 (1)

## Testing Instructions

1. Add a text element on a white background
2. Click on the "Magic wand" button for auto-styling.
3. The text should stay black as it is without any highlight.
4. Verifying that auto-styling still works correctly in other cases e.g. Adds a highlight incase of an applicable background image

No

No

No

## Checklist

- [] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))

---

Fixes #9546


packages/story-editor/src/utils/contrastUtils.js
```diff
@@ -158,3 +158,26 @@ export function getAccessibleTextColorsFromPixels(pixelData, fontSize) {
...
+ * @param {Object} pixelData Array of pixel data.
+ * @param ctx
+ * @param {number} x x-axis coordinate of the top-left corner of the area from which the ImageData will be extracted.
+ * @param {number} y y-axis coordinate of the top-left corner of the area from which the ImageData will be extracted.
+ * @param {number} width width of the area from which the ImageData will be extracted.
+ * @param {number} height height of the area from which the ImageData will be extracted.
+ * @param {number} sharpness resolution to use in getting the pixels data.
+ * @return {Object} Returns ImageData object containing the image data for the canvas context and coordinates specified
+ */
+export function getPixelsFromCoords(ctx, x, y, width, height, sharpness) {
```
miina: @Trapsta Could you explain what exactly is the change here? I'm not sure I understand how this excludes the selected text element from the contrast calculation.

Currently, when testing, the contrast is doesn't seem right anymore, e.g. adding a new Title in this PR doesn't result in good contrast, part of the title is on a darker background and that doesn't seem to be considered (generally I'm not able to see a highlight being added anymore even when it's needed):
Screenshot 2021-11-16 at 09 35 53

Adding a title in `main` with the same image -- highlight is added to the background for ensuring the text to stand out:
Screenshot 2021-11-16 at 09 36 29

The contrast calculation was fine as it was before, there's no need to change that part. The problem was that when re-triggering the logic, the element itself was also included to the image that was used for calculations, influencing the result. As far as I can see that still happens -- the element is still included in the image and it still influences the result.

There shouldn't be a need to change the general contrast calculation, we should find a way how to exclude the selected element from the image that `htmlToImage` is generating for contrast calculation in case of re-triggering the logic specifically.

Please let me know if you have any questions about it / if anything is unclear.

miina (changes requested):

packages/story-editor/src/utils/contrastUtils.js
Trapsta: Thanks @miina , that provides more clarity but I think the steps to reproduce on the issue description might be a bit misleading. We might need to update the description with the context you provided here.

miina: @Trapsta The steps to reproduce usually just show one example of how to reproduce the issue (the steps to reproduce might even be missing in some cases from an issue)

The description and the expected behaviour are the things to check in case of any issue. E.g. here the expected behaviour is:
"The element itself should not be included into the generated image."

miina: Anytime you have doubts or are thinking of a solution that does not satisfy the expected behaviour / does not fully fix the described issue, please reach out to discuss or leave a comment under the issue.

Trapsta: Cool, noted. Thanks

packages/story-editor/src/utils/usePageAsCanvas.js
```diff
@@ -27,7 +27,7 @@ import { ZOOM_SETTING } from '../constants';
 import { getAccessibleTextColorsFromPixels } from './contrastUtils';
 import { calculateTextHeight } from './textMeasurements';
 
-function usePageAsCanvas() {
+function usePageAsCanvas(element) {
```
miina: We should probably rename this to something more clear, e.g. `toExclude` or `elementToExclude`.

Alternatively we could make this a flag instead, e.g. `excludeSelectedElement` and then we could get the selected element here instead.

packages/story-editor/src/utils/usePageAsCanvas.js
```diff
@@ -98,6 +98,11 @@ function usePageAsCanvas() {
               const promise = htmlToImage.toCanvas(fullbleedContainer, {
                 fontEmbedCss: '',
                 pixelRatio: 1,
+                // exclude selected element from generated image to prevent interfereing with contrast calculation. true includes and false excludes
+                filter: (node) =>
```
miina: We could maybe move this into a separate function and only use if if the element to exclude is actually set.

Note that we also have `nodesById` available via `useCanvas`, we could get the node directly by the element ID and use that in the filter for the comparison instead.

packages/story-editor/src/utils/usePageAsCanvas.js
barklund: In the case of multiple selected text elements, what do we do currently and what should we do? Should we do each of them as if that text did not exist, but the other selected texts do exist, or should we do them all as if none of them exist? Does this even work for multi-selection?

packages/story-editor/src/utils/usePageAsCanvas.js
```diff
@@ -98,6 +98,11 @@ function usePageAsCanvas() {
...
+                  !node.dataset ||
+                  !node.dataset.elementId ||
+                  node.dataset.elementId !== element.id,
```
barklund: With [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), this just becomes:

```js
                  node?.dataset?.elementId !== element.id,
```

barklund: Hey, this was the first time I used [the new "linkify-on-paste" feature that Github just added](https://github.blog/changelog/2021-11-10-linkify-selected-text-on-url-paste/).

It's a nice feature that I think Slack might have invented and I've used it a lot there.

(and this was the second time 😄 )

packages/story-editor/src/utils/usePageAsCanvas.js
miina: Currently the feature doesn't support multi-selection, so there's no clear way how to account for multi-selection since due to the same question of if to consider each text as separate or all together, it was decided not to support multi-selection (for now).

packages/story-editor/src/utils/usePageAsCanvas.js
swissspidy: > It's a nice feature that I think Slack might have invented and I've used it a lot there.

Slack has actually copied it from WordPress, which has had this for years :-)

packages/story-editor/src/utils/usePageAsCanvas.js
Trapsta: Good point , using the `selectedElementIds` property from the `useStory` hook was enough because the node that `html-to-image` was giving me was a bit different from the node I was getting from `nodesById(selectedElementIds[0]` so I couldn't compare the two node directly using `isEqualNode` but doing it with just the id from `selectedElementIds` is still sufficient and maybe a little faster

packages/story-editor/src/utils/usePageAsCanvas.js
```diff
@@ -95,9 +98,15 @@ function usePageAsCanvas() {
         if (!pageCanvasPromise && fullbleedContainer) {
           import(/* webpackChunkName: "chunk-html-to-image" */ 'html-to-image')
             .then((htmlToImage) => {
+              const filter = (node) => {
+                // exclude selected element from generated image to prevent interfereing with contrast calculation. true includes and false excludes
+                return node?.dataset?.elementId !== selectedElementIds[0];
```
miina, line 106: @Trapsta We still need to add a flag for skipping the element, it shouldn't be always skipped, only if we're using the "Magic wand" button. So let's add `skipSelectedElement` or something like that as a boolean to the `usePageAsCanvas` hook, and only filter if that is set.

packages/story-editor/src/utils/usePageAsCanvas.js
```diff
@@ -95,9 +98,15 @@ function usePageAsCanvas() {
...
+              };
+
               const promise = htmlToImage.toCanvas(fullbleedContainer, {
                 fontEmbedCss: '',
                 pixelRatio: 1,
+                filter: selectedElementIds[0] && filter,
```
barklund: I feel like `selectedElementIds.length === 1` is a better test. Also please move to a const like `hasSingleSelection` or similar.

barklund (approved):
Approved pending comments.

packages/story-editor/src/utils/usePageAsCanvas.js
```diff
@@ -95,9 +98,18 @@ function usePageAsCanvas() {
         if (!pageCanvasPromise && fullbleedContainer) {
           import(/* webpackChunkName: "chunk-html-to-image" */ 'html-to-image')
             .then((htmlToImage) => {
+              const hasSingleSelection = selectedElementIds.length === 1;
+              const filterSelectedElement =
+                skipSelectedElement && hasSingleSelection;
+              const filter = (node) => {
+                // exclude selected element from generated image to prevent interfereing with contrast calculation. true includes and false excludes
```
miina:
```suggestion
                // Exclude selected element from generated image to prevent interfering with contrast calculation: `true` includes and `false` excludes.
```

miina (approved):

miina:
Thinking of merging this PR with unrelated Karma errors, looks like it's failing for more than one time in a row with the media elements not being loaded in `fixture`:
`Error: Not ready: only found 0 media elements`

Example run: https://github.com/google/web-stories-wp/runs/4286674056?check_suite_focus=true#step:7:8190

@swissspidy Is this something that would go for Infra pod to look into? Have you seen this issue previously?

PR #9723, opened 2021-11-16
Code Quality: Adopt PHPStan level 9

### LINKED ISSUE #9629: Adopt PHPStan level 9
## Feature Description

PHPStan 1.0.0 has a new maximum level of 9 that is more strict.

We're currently on level 8, so it would be nice if we could possibly adopt level 9.

There are currently hundreds of errors being reported when switching to level 9. As usual, some are easy fixes, some require more sophisticated solutions.

For instance, at first glance it seems like we might need to write custom extensions for some of our code with dynamic return types, like `Settings::get_setting()`.

Right now it seems like this is also blocked by https://github.com/szepeviktor/phpstan-wordpress/issues/65 due to return types of `apply_filters` calls being flagged.

swissspidy:
## Context

PHPStan 1.0.0 has a new maximum level of 9 that is more strict and helps catch more issues.

We're currently on level 8, so it would be nice if we could possibly adopt level 9.

## Summary

This PR bumps the PHPStan level to 9 and overhauls the code base to fix all reported errors.

## Relevant Technical Choices

* Bundled `ApplyFiltersDynamicFunctionReturnTypeExtension` from upstream
* Added lots of PHPDoc comments

## To-do

N/A

## User-facing changes

N/A

## Testing Instructions

1.

No

No

No

## Checklist

---

Fixes #9629


includes/Integrations/Jetpack.php
```diff
@@ -274,7 +279,7 @@ public function filter_rest_api_response( WP_REST_Response $response, WP_Post $p
 
 		$metadata = wp_get_attachment_metadata( $post->ID );
 
-		if ( $metadata && isset( $metadata['videopress']['duration'] ) ) {
+		if ( $metadata && isset( $metadata['videopress']['duration'], $data['media_details'] ) && is_array( $data['media_details'] ) ) {
```
spacedmonkey, line 282:
```suggestion
		if ( $metadata && isset( $metadata['videopress']['duration'] ) ) {
		   if( ! isset( $data['media_details']  ) || ! is_array( $data['media_details']  )  ) {
		        $data['media_details'] = [];
		   }
```

includes/REST_API/Embed_Controller.php
```diff
@@ -137,15 +145,19 @@ public function get_proxy_item( $request ) {
...
-			return rest_ensure_response( $response );
+		if ( is_string( $data ) && ! empty( $data ) ) {
+			/**
+			 * Decoded cached embed data.
+			 *
+			 * @var array|null $embed
+			 */
+			$embed = json_decode( $data, true );
+
+			if ( $embed ) {
```
spacedmonkey, line 157: If there json is invalid or this is empty, should we return a WP_error here?

includes/REST_API/Embed_Controller.php
```diff
@@ -293,10 +317,27 @@ private function url_to_post( $url ) {
...
-			$home_url_host = str_replace( 'www.', '', wp_parse_url( home_url(), PHP_URL_HOST ) );
+			/**
+			 * The URL's hostname.
+			 *
+			 * @var string|false|null $url_host
+			 */
+			$url_host = wp_parse_url( $url, PHP_URL_HOST );
+			if ( $url_host ) {
+				$url_host = str_replace( 'www.', '', $url_host );
+			}
```
spacedmonkey, line 328: These lines are repeated. Worth creating a little method on the class.

includes/REST_API/Link_Controller.php
```diff
@@ -136,10 +142,18 @@ public function parse_link( $request ) {
...
-			return rest_ensure_response( $response );
+		if ( is_string( $data ) && ! empty( $data ) ) {
+			/**
+			 * Decoded cached link data.
+			 *
+			 * @var array|null $link
+			 */
+			$link = json_decode( $data, true );
+
+			if ( $link ) {
```
spacedmonkey, line 153: Same point as above.

includes/REST_API/Stories_Media_Controller.php
```diff
@@ -128,17 +134,43 @@ public function get_items( $request ) {
...
+		 *
+		 * @var int $original_id
+		 */
+		$original_id = ! empty( $request['original_id'] ) ? $request['original_id'] : null;
+
 		unset( $request['post'] );
 
 		$response = parent::create_item( $request );
-		if ( is_wp_error( $response ) || ( ! $parent_post && ! $original_id ) ) {
+		if ( ( ! $parent_post && ! $original_id ) || is_wp_error( $response ) ) {
```
spacedmonkey, line 155: Why does this have to change?

includes/Renderer/Stories/FieldStateFactory/Factory.php
```diff
@@ -119,8 +119,6 @@ public function get_field( $view = 'grid' ) {
 				 * @param FieldState $default_field_state Field state object.
 				 */
 				$field_state = apply_filters( 'web_stories_default_field_state', $default_field_state );
-
-				$field_state = $field_state instanceof FieldState ? $field_state : $default_field_state;
```
spacedmonkey, line 123: Why was this removed? This seems valid to me. If someone filters a field_state to false, we need to reset it back to a class that extends `FieldState`

includes/Integrations/AMP.php
```diff
@@ -388,9 +388,16 @@ protected function get_validated_url_post_type( $post_id ) {
 			return null;
 		}
 
-		$queried_object = get_post_meta( $post->ID, '_amp_queried_object', true );
+		$queried_object = (array) get_post_meta( $post->ID, '_amp_queried_object', true );
```
spacedmonkey: I thought were trying to remove the castings?

includes/Integrations/Core_Themes_Support.php
```diff
@@ -142,7 +142,7 @@ public function register() {
 
 		$this->extend_theme_support();
 
-		$options = get_option( Customizer::STORY_OPTION );
+		$options = (array) get_option( Customizer::STORY_OPTION );
```
spacedmonkey:
```suggestion
		$options = (array) get_option( Customizer::STORY_OPTION, [] );
```

includes/Integrations/Site_Kit.php
```diff
@@ -95,7 +95,7 @@ protected function is_plugin_active(): bool {
 	 */
 	protected function is_adsense_module_active(): bool {
 		$adsense_module_active       = in_array( 'adsense', $this->get_site_kit_active_modules_option(), true );
-		$adsense_options             = get_option( 'googlesitekit_adsense_settings' );
+		$adsense_options             = (array) get_option( 'googlesitekit_adsense_settings' );
```
spacedmonkey, line 98: Should this use the settings object?

includes/Media/Video/Poster.php
```diff
@@ -235,6 +235,11 @@ public function get_thumbnail_data( int $thumbnail_id ): array {
 	 * @return void
 	 */
 	public function delete_video_poster( int $attachment_id ) {
+		/**
+		 * Post ID.
+		 *
+		 * @var int $post_id
```
spacedmonkey: Should this also be false?

includes/REST_API/Stories_Media_Controller.php
swissspidy: Microperformance, checking two variables is faster than a function call.

includes/Integrations/AMP.php
swissspidy: Maybe some leftover, I'll have to check.

spacedmonkey (changes requested):
Some nits but otherwise, this looks good.

includes/Renderer/Stories/FieldStateFactory/Factory.php
swissspidy: With `ApplyFiltersDynamicFunctionReturnTypeExtension` in place, the filter's PHPDoc becomes the source of truth.

The docblock says this can only ever be a `FieldState` instance, so that's the only thing PHPStan expects and it would yell at us if we'd _still_ do an `instanceof` check.

If someone filters this to be something else, they deserve to get an error.

The alternative is to do something like `@param FieldState|mixed $default_field_state Field state object.` but I don't see a point in doing that.

includes/Integrations/Site_Kit.php
swissspidy: The `Settings` object is used for retrieving our own plugin's settings that we've registered with `registered_setting`, We don't currently use it for other settings, especially not some from other plugins.

Rethinking that is best done in a dedicated ticket.

includes/REST_API/Embed_Controller.php
swissspidy: Then it would never be possible to clear that cache and fetch that data anew.

It makes sense to just fetch the data again if there's an error.

includes/Integrations/Jetpack.php
swissspidy: Just a readability preference?

includes/REST_API/Embed_Controller.php
swissspidy: A method for a 1 line `str_replace` call?

includes/Integrations/Jetpack.php
swissspidy: This would mean a change in behavior ⚠️ 

Right now, the code only adds duration to `media_details` if `media_details` is actually a requested field. Otherwise `media_details` is not provided.

Your suggested change would make it so that `media_details` is always provided.

includes/Media/Video/Poster.php
swissspidy: Just updated the PHPDoc because the result of `get_post_meta` is either `int 0` or a non-empty string.

spacedmonkey (approved):

PR #9741, opened 2021-11-18
Pre-Publish Checklist: Show contrast error when there's a contrast problem

### LINKED ISSUE #7724: Fix contrast checking for SVG rendering
# Bug Description
Unexpected behavior seen while working on [SVG rendering](https://app.zenhub.com/workspaces/web-stories-5e94d3aced449034e1e9b226/issues/google/web-stories-wp/7519) in the pre-publish checklist. Contrast checking seems to be showing contrast issues when it shouldn't. (see screenshot)

# Expected Behaviour
The contrast check should only appear when there is insufficient contrast between the foreground text and background color.

# Steps to Reproduce

[After SVG rendering is being used in the editor] 
1. Start a new story
2. Add a background color to the page (e.g. #397AED) and some text (e.g. #fff)
3. Open the pre publish checklist
4. There are too many recommendations for fixing the contrast between the text and the background.

# Screenshots

Additional Context

    Plugin Version: 1.1.0
    Operating System: macOS
    Browser: Chrome 86

samwhale:
## Context

Contrast checking seems to be showing contrast issues when it shouldn't.

## Summary

Pass the real font size in pixels rather than the font size in editor pixels.

## Relevant Technical Choices

The font size when converted to editor pixels comes out to be much less than the font size given to the element. Text that should pass the contrast check fails due to the change in units.

## To-do

n/a

## User-facing changes

none

## Testing Instructions

This is a good website to see which colors should pass/fail together https://webaim.org/resources/contrastchecker/. **We are testing against the `WCAG AA` standard**.

1. Start a new story
2. Add a background color to the page (e.g. #397AED) and some text (e.g. #fff)
3. Open the pre publish checklist
4. No contrast recommendation should show
5. Change the text color to something with low contrast (e.g. #1263FF).

The warning should be shown when:
1. The contrast is less than `3.0`
2. The contrast is between `3.0` and `4.5` but the font size is less than `18px`

no

no

no

## Checklist

---

Fixes #7724

### DIFF
--- a/packages/story-editor/src/components/checklist/checks/pageBackgroundLowTextContrast/check.js
+++ b/packages/story-editor/src/components/checklist/checks/pageBackgroundLowTextContrast/check.js
@@ -19,8 +19,6 @@
 import {
   PAGE_RATIO,
   FULLBLEED_RATIO,
-  dataToFontSizeY,
-  dataFontEm,
   getBox,
   getBoundRect,
 } from '@web-stories-wp/units';
@@ -48,16 +46,6 @@ import { getSpansFromContent } from '../../utils';
  * @property {number} g green value
  * @property {number} b blue value
  */
-
-/**
- *
- * @param {number} fontSize The text element's font size in editor pixels
- * @return {number} the true font size for measuring against accessibility standards
- */
-function getPtFromEditorFontSize(fontSize) {
-  // 1 point = 1.333333 px
-  return dataFontEm(dataToFontSizeY(fontSize, 100)) * 1.333333;
-}
 
 /**
  *
@@ -318,7 +306,7 @@ function getTextShapeBackgroundColor({ background }) {
  * @param {Object} arguments The arguments
  * @param {RGB} arguments.backgroundColor The r, g, b object representing a background color to compare to the text colors
  * @param {Array} arguments.textStyleColors The array of style colors of the text being checked
- * @param {number} arguments.fontSize The font size (in editor pixels) of the text being checked
+ * @param {number} arguments.fontSize The font size (in pixels) of the text being checked
  * @return {boolean} If true, there is a contrast issue with some text
  */
 function textBackgroundHasLowContrast({
@@ -340,7 +328,7 @@ function textBackgroundHasLowContrast({
     const contrastCheck = checkContrastFromLuminances(
       textLuminance,
       backgroundLuminance,
-      fontSize && getPtFromEditorFontSize(fontSize)
+      fontSize
     );
     return !contrastCheck.WCAG_AA;
   });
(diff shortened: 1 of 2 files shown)


packages/story-editor/src/components/checklist/checks/pageBackgroundLowTextContrast/check.js
```diff
@@ -49,16 +47,6 @@ import { getSpansFromContent } from '../../utils';
...
 
-/**
- *
- * @param {number} fontSize The text element's font size in editor pixels
- * @return {number} the true font size for measuring against accessibility standards
- */
-function getPtFromEditorFontSize(fontSize) {
-  // 1 point = 1.333333 px
-  return dataFontEm(dataToFontSizeY(fontSize, 100)) * 1.333333;
-}
```
samwhale, line 60: `22pt` in the editor would become `6.3999984pt` according to this fn. 
Screen Shot 2021-11-18 at 11 07 22 AM
The calculated value ends up being about `19px`.

I removed this because we're already using `px` but this function returns the font size in `pt`. The `aa` check calculates contrast using the font size in `px` 

---

For reference: The `aa` function in the `hues` package checks font size in pixels:
image

```
WCAG 2.0 level AA requires a contrast ratio of at least 4.5:1 for normal text and 3:1 for large text. 
WCAG 2.1 requires a contrast ratio of at least 3:1 for graphics and user interface components (such as form input borders). 
WCAG Level AAA requires a contrast ratio of at least 7:1 for normal text and 4.5:1 for large text.

Large text is defined as 14 point (typically 18.66px) and bold or larger, or 18 point (typically 24px) or larger.
```

samwhale: @miina @swissspidy Wanted to check with you two specifically, I'm hoping you have context here.

Should we be passing in the value that users have set in the design panel or the calculated value on the element? So in the above example, it would be `22px` vs `19.22px` 🤔

miina: @samwhale We should be using the attribute value of the element directly (`element.fontSize`), that would be 22 in this case. The "real" displayed value at that point (19.22px) depends completely on the screen size, so if we'd use that, it would create inconsistent results depending on the screen size.

samwhale: Awesome, so this should stay removed all together. Thanks!

packages/story-editor/src/utils/contrastUtils.js
```diff
@@ -27,12 +27,12 @@ import * as hues from '@ap.cx/hues';
  * @return {number} Luminance
  */
 export function calculateLuminanceFromRGB(rgb) {
-  const { r, g, b, a } = rgb;
+  const { r, g, b, a = 1.0 } = rgb;
```
BrookeGraham, line 30: Nice.

BrookeGraham (approved):

BrittanyIRL (approved):
Works as you defined, thanks for getting Miina's eyes on the update too.

PR #9743, opened 2021-11-18
Media: Add `decoding=async` to all images

### LINKED ISSUE #9740: Media: Add `decoding=async` to all images
## Feature Description

Background [Trac #53232](https://core.trac.wordpress.org/ticket/53232). 

In WordPress core, there is a plans to add `decoding=async` to all images. This improves page render performance. We should consider doing the same for images found in the media library.

spacedmonkey:
## Summary

Add `decoding=async` to all images rendered in the web stories editor. By adding this attribute, is improves rendering and time to first paint. See user facing change to compare results. 

## User-facing changes

|Request|Before  | After |
--- | --- | ---|
|Normal|main|branch|
|Slow CPU|main-slow-cpu|branch-slow-cpu|
|Fast 3G|main-fast-3g|branch-fast-3g|

## Testing Instructions

1.

## Checklist

---

Fixes #9740


packages/story-editor/src/elements/media/imageDisplay.js
```diff
@@ -76,6 +76,7 @@ function ImageDisplay({ element, box, previewMode }) {
   );
 
   imgProps.crossOrigin = 'anonymous';
+  imgProps.decoding = 'async';
```
swissspidy: FWIW, this overrides `decoding="sync"`

spacedmonkey:
```suggestion
```

spacedmonkey:
@swissspidy What are your recommendations on profiling this? I have tried react profiler and lighthouse, but because the image rendering in after the editor is loaded, I can't get a good reading on them. Async used in places, where height and width are set and no need to calulated. In all cases I have highlighted in our codebase, these values are set. 

How much faster does it have to be to get this merged? Is it okay, we just believe it will be faster?

swissspidy:
https://github.com/google/web-stories-wp/issues/9740#issuecomment-981927414 mentioned to enable CPU throttling before measuring LCP.

Regardless of any numbers, are there any changes _visible_ to the user? Can you _see_ any changes with CPU throttling enabled?

spacedmonkey:
I have done some testing. See the before and after results. The best result is in time to first paint. Meaning this is a meaningful and worthwhile change.

swissspidy:
Is the trend still there even if you run the tests a number of times?

spacedmonkey:
> Is the trend still there even if you run the tests a number of times?

Each one of these tests were run 3 times

swissspidy (approved):

PR #9744, opened 2021-11-18
Explore Templates: Reduce Noise for Colors in Search

### LINKED ISSUE #9624: Add primary/secondary color families to template tags
## Feature Description

Right now template colors are really specific (burnt sienna i'm lookin' at you) which is great for the detailed template view but really unhelpful for the new template search (#9508 ) - let's add in the primary color families as tags (and make 'em translatable).

## Alternatives Considered

- should this be it's own `color-tag` section in template metadata or just strings in the array of tags? If the latter, updating the template search component to also consider color-tags on search matching. 

## Acceptance Criteria 
- Update template package unit tests to make sure color groups aren't skipped on new template additions.

BrittanyIRL:
## Context

We're adding in a free form search for templates on the dashboard (explore templates).

## Summary

Previous to this update, the template colors displayed as search options in explore templates have been super specific because they are based on the color label (which is used to describe the color in the template details view). We want to be more generic so users don't see options for both 'Swamp Green' and  'Dark Moss Green' - rather, just 'Green'. 

## Relevant Technical Choices

There's nothing technical here, just metadata updates to the templates package so that all colors have a `family`. Added a new test to templates to make sure there's a family added to new templates going forward, tweaked the label stored for the template search results in the dashboard package.

There were a few different names thrown out for what to call this color grouping - alias, primitive, family. I opted for family because all of the variants of orange belong to the same color family so it seemed the most straight forward. If anyone feels really strongly in a different way that's fine we can update it. 

## User-facing changes

| New | Old | 
| --- | --- |
| Screen Shot 2021-11-18 at 10 33 24 AM | Screen Shot 2021-11-18 at 11 08 38 AM |

## Testing Instructions

- Enable the template search experiment 
- Go to dashboard -> Explore Templates 
- Search some colors. For instance, 'Green' and see that 'Swamp Green' and 'Dark Moss Green' are no longer additional options in the list of results.

## Checklist

---

Fixes #9624


packages/templates/src/raw/a-day-in-the-life/metaData.js
```diff
@@ -31,9 +31,21 @@ export default {
...
+    },
+    {
+      label: _x('Arctic White', 'color', 'web-stories'),
+      color: '#fcfcfc',
+      family: _x('White', 'color family', 'web-stories'),
+    },
+    {
+      label: _x('Honey Pot Yellow', 'color', 'web-stories'),
+      color: '#ffc864',
+      family: _x('Yellow', 'color family', 'web-stories'),
```
BrittanyIRL: 👋  as resident expert, @swissspidy this usage ok? `_x('Yellow', 'color family', 'web-stories'),`

packages/templates/src/raw/12-hours-in-barcelona/metaData.js
```diff
@@ -30,13 +30,26 @@ export default {
     _x('Blue', 'template keyword', 'web-stories'),
   ],
   colors: [
-    { label: _x('Winter Blue', 'color', 'web-stories'), color: '#a4deff' },
+    {
+      label: _x('Winter Blue', 'color', 'web-stories'),
+      color: '#a4deff',
+      family: _x('Blue', 'color family', 'web-stories'),
```
swissspidy: Let's use `color` instead of `color family` as context to reduce the amount of duplicate strings

packages/templates/src/test/getTemplates.js
```diff
@@ -69,4 +69,20 @@ describe('getTemplate', () => {
       });
     });
   });
+
+  it('should have keys for color, label, and family for each template color', async () => {
```
swissspidy, line 73: 👍

packages/templates/src/raw/album-releases/metaData.js
```diff
@@ -31,10 +31,26 @@ export default {
...
-    { label: _x('Cola Brown', 'color', 'web-stories'), color: '#3d3225' },
+    {
+      label: _x('Hot Orange', 'color', 'web-stories'),
+      color: '#f7722f',
+      family: _x('Orange', 'color', 'web-stories'),
+    },
+    {
+      label: _x('Peach Bisque', 'color', 'web-stories'),
+      color: '#fee2c6',
+      family: _x('Red', 'color', 'web-stories'),
```
BrookeGraham: Just curious, how did you decide the family? I would think a `Peach` would fall under `Orange`🤔

maxyinger: I'm assuming there's a lot scenarios where some of these could go several ways, and there's a lot of colors here to go through.

imo if they're not glaringly wrong, I'm fine with them. the only way I'd know how to do this is eyeballing it, so I'm assuming there's a lot of subjectivity involved here

BrittanyIRL: Max is spot on, some of these are subjective. This one is probably better off orange, i will update it.

packages/templates/src/raw/belly-fat-workout/metaData.js
```diff
@@ -31,10 +31,26 @@ export default {
...
+    {
+      label: _x('Lime Green', 'color', 'web-stories'),
+      color: '#e0ff8c',
+      family: _x('Green', 'color', 'web-stories'),
+    },
+    {
+      label: _x('Slime Green', 'color', 'web-stories'),
+      color: '#d6fd56',
+      family: _x('Green', 'color', 'web-stories'),
+    },
```
samwhale, line 43: Two greens 🍏 Is it okay for a template to be defined with two colors that have the same family?

samwhale (approved):
wowzers

packages/templates/src/raw/belly-fat-workout/metaData.js
maxyinger: yea it's fine. having one or more results in the same behavior of it just showing up when searched for

packages/dashboard/src/app/views/utils/getTemplateFilters.js
```diff
@@ -33,7 +33,7 @@ function getTemplateFilters(templates) {
   const colors = templatesArr
     .flatMap((t) => t.colors)
     .map((c) => ({
-      ...c,
+      label: c.family,
       value: c.label,
```
maxyinger: might be nice to make these both the family:
```
label: c.family,
value: c.family,
```

I don't think we're even using the value, but just for safe keeping.

maxyinger (approved):
Looks great! Thanks for going through all of these

packages/dashboard/src/app/views/utils/getTemplateFilters.js
BrittanyIRL: yeah good point, was trying to be the least invasive - like keeping it tight though.

packages/templates/src/raw/belly-fat-workout/metaData.js
BrittanyIRL: slime and lime are clearly nuanced greens 😂  it'll only show up once - like max said.

packages/templates/src/raw/album-releases/metaData.js
BrookeGraham: That was my assumption, I was just curious if there was some magic way I didn't know of. Going through these by hand must have been brutal. 💪 Thanks for doing it!

BrookeGraham (approved):

PR #9748, opened 2021-11-18
Editor: Improve behavior when right-clicking on multiple elements

### LINKED ISSUE #9426: Right Click Menu: Right Click from Element to Element Discovery
Currently the right click context menu opens up based on the selected element. This ticket is to:
1. Explore passing in an element to the callback handler to generate the menu items vs generating them based on the selected element
2. Explore alternatives to the fullscreen clickable background element approach to dismiss the context menu. The new approach should be geared towards allowing for right clicking between elements, but not cause any regressions from the previous approach.

Task 1 should be pretty straight forward to implement, however, task 2 could be a rabbit hole. Mainly setting this up as a 3 point Discovery ticket in case there's no fruitful approaches for task 2 (which is very much a possible outcome).

Please try to time box this. It's okay if nothing gets discovered. This is just a possible UX enhancement, so it's not worth spending a week on if initial discovery seems like there's no possible paths

maxyinger:
## Context
Part of investigative work to see if we could better the right click menu ux with a better pattern for dismissing it when clicking outside.

## Summary
Adds a new approach to detect clicking outside of the context menu to dismiss it.

## Relevant Technical Choices
Used the `mousedown` capture event on the document to always detect the `mousedown` event even if components have disabled bubbling or prevented default. This seemed to cover all needed mouse interaction scenarios to dismiss the context menu.

## To-do
NA

## User-facing changes
You should now be able to right click between elements in the canvas without needing to have an intermediary click to dismiss the context menu. Other than that, should behave the same as before

## Testing Instructions

1. Open the editor and make sure you have several elements of different types
2. right click one element and see the relevant right click menu appear
3. right click an element of another type and see the correct right click menu appear
4. click anywhere off the right click menu and see that the right click menu gets dismissed (this should behave the same as before)

## Checklist

---

Fixes #9426


BrittanyIRL (approved):
this is super cool, i like it a lot. Kinda weird that the dashboard karma tests are complaining, i'm out tomorrow so I'm approving this ahead of fixing that. great idea!

packages/design-system/src/utils/useComposeRefs.js
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Takes refs as a series of arguments and composes them to
+ * return a single callback ref.
+ *
+ * @param  {...*} refs - ref arguments
+ * @return {(node:HTMLElement) => void} composed callback ref
+ */
+function useComposeRefs(...refs) {
+  const refsRef = useRef(refs);
```
samwhale, line 29: refs ref lmao

packages/design-system/src/utils/useMouseDownOutsideRef.js
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * External dependencies
+ */
+import { useRef, useCallback } from '@web-stories-wp/react';
+/**
+ * Internal dependencies
+ */
+import { noop } from './noop';
+
+function useMouseDownOutsideRef(onPointerDownOutside = noop) {
```
samwhale: 💭 Let's add a little docstring description here to describe what this does. Specifically that it returns a ref 😄

samwhale (approved):
Wow this is so much better. Cool thing is that right clicking another element now selects that element too since we don't have the mask. This is excellante
Screen Shot 2021-11-18 at 11 07 09 AM
Screen Shot 2021-11-18 at 11 07 22 AM

packages/design-system/src/utils/useMouseDownOutsideRef.js
maxyinger: ope. woops. thanks for pointing out

maxyinger:
> Wow this is so much better. Cool thing is that right clicking another element now selects that element too since we don't have the mask. This is excellante Screen Shot 2021-11-18 at 11 07 09 AM Screen Shot 2021-11-18 at 11 07 22 AM

Not following what you're trying to point out with these screenshots. Mind elaborating?

samwhale:
> > Wow this is so much better. Cool thing is that right clicking another element now selects that element too since we don't have the mask. This is excellante 
> 
> Not following what you're trying to point out with these screenshots. Mind elaborating?

hahaha shot wrong things. I mean to add my screen recordings 😮‍💨

maxyinger:
> > > Wow this is so much better. Cool thing is that right clicking another element now selects that element too since we don't have the mask. This is excellante
> > 
> > 
> > Not following what you're trying to point out with these screenshots. Mind elaborating?
> 
> hahaha shot wrong things. I mean to add my screen recordings 😮‍💨  

lol right on

maxyinger:
wow you're quick. Thanks for merging Pascal! 😄

PR #9752, opened 2021-11-19
Decoupling: Make dashboard API Callbacks optional

mohdsayed:
## Context

While working on https://github.com/google/web-stories-wp/pull/9644, we realized that we have to provide some default API callbacks for the dashboard to work here https://github.com/google/web-stories-wp/blob/ad567f6777ac9d419029b184b7f671f100909609/.storybook/stories/playground/dashboard/index.js#L38-L45

To simplify the integration layer, we should make these optional.

## Summary

In this PR, we have tried to do two things
1. Do not render the related button or link if the underlying API callback is not provided.
2. If possible, handle the case when the function is directly called elsewhere like inside some `useEffect`

- [x] Make `getAuthor` optional ( https://github.com/google/web-stories-wp/pull/9752/commits/1f3140bf5b5a079be0d13c414caaf349d73eb273 ) 
- [x] Make `createStoryFromTemplate` optional ( https://github.com/google/web-stories-wp/pull/9752/commits/9b503829866345ae604b6e6a324cb0bdbff9140e )
- [x] Make `duplicateStory`, `trashStory`, `updateStory` optional ( https://github.com/google/web-stories-wp/pull/9752/commits/fa5626b7f70f42e6c544f6e7f6a130a78c7f0c29 )
- [x] Update unit and integration tests for this change.

## Relevant Technical Choices

If checking conditions using the API callbacks doesn't make a lot of sense, just a rough idea that perhaps we may introduce some more custom capabilities and use those as conditionals to not render those buttons and links.

## Testing Instructions

1.

## Checklist

---


packages/dashboard/src/app/api/useStoryApi.js
```diff
@@ -184,15 +177,19 @@ const useStoryApi = () => {
         });
       }
     },
-    [createStoryFromTemplateCallback]
+    [apiCallbacks]
   );
 
+  if (!apiCallbacks.createStoryFromTemplate) {
+    createStoryFromTemplate = noop;
+  }
```
swissspidy: Why is this needed?

I'd tend to think that `createStoryFromTemplate` should never be called in the first place when the API callback isn't provided, because any button to create a story from a template would not be visible.

mohdsayed: Not needed, it was only for handling some future case influenced by how the function is being called inside `useEffect` here.

https://github.com/google/web-stories-wp/blob/5b713a3f908660330014bfbac951c95d23f092df/packages/dashboard/src/app/views/myStories/index.js#L106-L112

Removed in https://github.com/google/web-stories-wp/pull/9752/commits/8dce4460fc58a70786f8eac72ab86672636925fc

mohdsayed:
Karma "MediaPane fetching should fetch 2nd page" is also failing on other PRs like https://github.com/google/web-stories-wp/pull/9764 , probably after this PR https://github.com/google/web-stories-wp/pull/9757/files

mohdsayed:
All passed 🎉

mohdsayed:
@swissspidy If no other feedback, it's ready to be merged.

packages/dashboard/src/app/views/exploreTemplates/content/templateGridItem.js
```diff
@@ -113,20 +115,22 @@ const TemplateGridItem = forwardRef(
...
-                  tabIndex={tabIndex}
-                  ariaLabel={sprintf(
-                    /* translators: %s: template title.*/
-                    __('Create new story from %s', 'web-stories'),
-                    title
-                  )}
-                >
-                  {__('Use template', 'web-stories')}
-                </Button>
+                {apiCallbacks?.createStoryFromTemplate && (
```
swissspidy: From a readability perspective it's odd that this UI component is checking for API callbacks.

Maybe that's better handled at a different level?

For example, in `TemplateGridView` we could do something like this:

```js
const canCreateStory = Boolean(templateActions.createStoryFromTemplate);

// ...

<TemplateGridItem
              detailLink={resolveRoute(centerTargetAction)}
              onCreateStory={canCreateStory ? () => handleUseStory({ id, title }): null}

// ...
```

Then, here we just check for `onCreateStory` instead of `apiCallbacks?.createStoryFromTemplate`

packages/dashboard/src/app/views/exploreTemplates/content/templateGridItem.js
```diff
@@ -113,20 +115,22 @@ const TemplateGridItem = forwardRef(
...
+                  <Button
+                    size={BUTTON_SIZES.SMALL}
+                    type={BUTTON_TYPES.PRIMARY}
+                    onClick={onCreateStory}
+                    disabled={!onCreateStory}
```
swissspidy: While we're at it: This seems a bit pointless 🤔 

Right now, `onCreateStory` is never false.

And with the proposed change above we don't render that component in the first place when it's not provided.

So we could just remove this.

packages/dashboard/src/app/views/myStories/index.js
```diff
@@ -100,19 +103,25 @@ function MyStories() {
...
   useEffect(() => {
-    fetchStories({
-      page: page.value,
-      searchTerm: search.keyword,
-      sortDirection: sort.direction,
-      sortOption: sort.value,
-      status: filter.value,
-      author: author.filterId,
-    });
+    if (apiCallbacks?.fetchStories) {
```
swissspidy: Should we really make this one optional? Seems kinda pointless to have a dashboard without any stories.

Maybe it should default to a function returning an empty array?

packages/dashboard/src/app/views/myStories/content/storiesView/index.js
```diff
@@ -174,12 +176,29 @@ function StoriesView({
...
+          [STORY_CONTEXT_MENU_ACTIONS.DELETE]: handleDeleteStory,
+        }),
+        ...(apiCallbacks?.duplicateStory && {
+          [STORY_CONTEXT_MENU_ACTIONS.DUPLICATE]: handleDuplicateStory,
+        }),
         [STORY_CONTEXT_MENU_ACTIONS.OPEN_STORY_LINK]: handleOpenStoryInEditor,
-        [STORY_CONTEXT_MENU_ACTIONS.RENAME]: handleRenameStory,
+        ...(apiCallbacks?.updateStory && {
+          [STORY_CONTEXT_MENU_ACTIONS.RENAME]: handleRenameStory,
+        }),
```
swissspidy: Are these checks for the `menuItemActions` really needed? Menu items without an API callback are filtered out below anyway.

packages/dashboard/src/app/views/myStories/content/storiesView/index.js
```diff
@@ -174,12 +176,29 @@ function StoriesView({
...
+          case STORY_CONTEXT_MENU_ACTIONS.DELETE:
+            return typeof apiCallbacks?.trashStory === 'function';
+          case STORY_CONTEXT_MENU_ACTIONS.DUPLICATE:
+            return typeof apiCallbacks?.duplicateStory === 'function';
+          case STORY_CONTEXT_MENU_ACTIONS.RENAME:
+            return typeof apiCallbacks?.updateStory === 'function';
+          default:
+            return true;
+        }
+      }),
```
swissspidy: Do we really need this?

Instead of checking for the API callbacks here, should we just rely on the capabilities being set?

Menu items are filtered based on user capabilities here:

https://github.com/google/web-stories-wp/blob/7849181a9f4ba9a5b56a9123bcb3dc845c9cda42/packages/dashboard/src/components/popoverMenu/story-menu-generator.js#L48-L50

packages/dashboard/src/app/views/templateDetails/header/index.js
```diff
@@ -70,18 +73,20 @@ function Header({ onHandleCtaClick, templateTitle }) {
...
-          size={BUTTON_SIZES.SMALL}
-          aria-label={sprintf(
-            /* translators: %s: template title */
-            __('Use %s template to create new story', 'web-stories'),
-            templateTitle
-          )}
-        >
-          {__('Use template', 'web-stories')}
-        </Button>
+        {apiCallbacks?.createStoryFromTemplate && (
```
swissspidy: Same point here https://github.com/google/web-stories-wp/pull/9752/files#r757598665

packages/dashboard/src/app/views/myStories/index.js
mohdsayed: Yeah, I think it makes sense, `fetchStories` shouldn't be optional.

packages/dashboard/src/app/views/myStories/content/storiesView/index.js
mohdsayed: Yes, good point, "Duplicate" story actually didn't have any capability assigned.

 
image


Removed this filter and added `hasEditAction` capability in https://github.com/google/web-stories-wp/pull/9752/commits/1621c6ae519652c8babaed335ea1ea9110bcd83f

packages/dashboard/src/app/views/templateDetails/header/index.js
mohdsayed: Done https://github.com/google/web-stories-wp/pull/9752/commits/439ac976e6a4711d47ba23e781357173353f198a

swissspidy (approved):

PR #9753, opened 2021-11-19
Add video playback shortcut for canvas element

### LINKED ISSUE #2350: Keyboard support for video playback
## Feature Description

Keyboard support for video playback (elements on the canvas).
At minimum play/pause toggle (spacebar?).
Not yet in [spec](https://docs.google.com/spreadsheets/d/1VNv8O5SnlqZfgYXo5ua2mCc-MWTN9ow6OtYWxml5Ka0/edit#gid=0).

## Additional Context

---

amovar18:
## Context

A feature was requested for adding a shortcut for video playback on canvas elements.

## Summary

Hitting spacebar while focus is on a video element on canvas should change the state from playing to pause or from pause to playing.

## User-facing changes
Yes this is a user facing changes, it will allow users to play/pause video playback on a video element on the canvas

## Testing Instructions

1. Open the story-editor.
2. Place a video on the canvas.
3. When video element is in focus hit spacebar to play/pause the video.

## Checklist

---

Fixes [#2350](https://github.com/google/web-stories-wp/issues/2350)

### DIFF
--- a/packages/story-editor/src/elements/video/controls.js
+++ b/packages/story-editor/src/elements/video/controls.js
@@ -44,7 +44,6 @@ function VideoControls({
       isTransforming={isTransforming}
       elementRef={elementRef}
       element={element}
-      allowKeyboardSupport
     />
   );
 }
--- a/packages/story-editor/src/elements/video/controls.js
+++ b/packages/story-editor/src/elements/video/controls.js
@@ -33,7 +33,6 @@ function VideoControls({
   isTransforming,
   elementRef,
   element,
-  videoRef,
 }) {
   const isActive =
     isSelected && !isTransforming && isSingleElement && !isEditing;
@@ -45,7 +44,7 @@ function VideoControls({
       isTransforming={isTransforming}
       elementRef={elementRef}
       element={element}
-      videoRef={videoRef}
+      allowKeyboardSupport
     />
   );
 }
@@ -58,7 +57,6 @@ VideoControls.propTypes = {
   isTransforming: PropTypes.bool.isRequired,
   elementRef: PropTypes.object.isRequired,
   element: StoryPropTypes.element.isRequired,
-  videoRef: PropTypes.object,
 };
 
 export default VideoControls;
--- a/packages/story-editor/src/elements/video/playPauseButton.js
+++ b/packages/story-editor/src/elements/video/playPauseButton.js
@@ -28,7 +28,7 @@ import {
 import { CSSTransition } from 'react-transition-group';
 import { __ } from '@web-stories-wp/i18n';
 import { rgba } from 'polished';
-import { Icons } from '@web-stories-wp/design-system';
+import { Icons, useKeyDownEffect } from '@web-stories-wp/design-system';
 
 /**
  * Internal dependencies
@@ -120,6 +120,7 @@ function PlayPauseButton({
   elementRef,
   element,
   videoRef = null,
+  allowKeyboardSupport = false,
 }) {
   const { isRTL } = useConfig();
   const hasVideoSrc = Boolean(element.resource.src);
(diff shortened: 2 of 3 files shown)


packages/story-editor/src/elements/video/controls.js
```diff
@@ -45,7 +44,7 @@ function VideoControls({
       isTransforming={isTransforming}
       elementRef={elementRef}
       element={element}
-      videoRef={videoRef}
```
amovar18, line 48: Removed `videoRef` since it does not receive anything and passes `undefined` to `playpausebutton`.

packages/story-editor/src/elements/video/playPauseButton.js
```diff
@@ -120,6 +120,7 @@ function PlayPauseButton({
   elementRef,
   element,
   videoRef = null,
+  allowKeyboardSupport = false,
```
swissspidy: When would we _not_ want keyboard support?

packages/story-editor/src/elements/video/playPauseButton.js
```diff
@@ -209,6 +210,33 @@ function PlayPauseButton({
...
   }, [checkMouseInBBox]);
 
+  const handleKeyDown = useCallback(
+    ({ key }) => {
+      if (allowKeyboardSupport && key === ' ') {
+        const videoNode = getVideoNode();
+        if (videoNode) {
+          if (videoNode.paused) {
+            videoNode.play();
+            setIsPlaying(true);
```
swissspidy: [`play()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play) returns a Promise that can be rejected if playback fails for whatever reason.

It'd be safer to do something like this:

```js
videoNode
  .play()
  .then(() => setIsPlaying(true))
  .catch(() => {});
```

packages/story-editor/src/elements/video/playPauseButton.js
amovar18: @swissspidy the `playPauseButton` is also being used in trim component. My thought was we dont want user to use spacebar to toggle between play/pause while trimming the video.

miina: Actually, it's possible to play/pause while trimming the video using the mouse, so it can be allowed with keyboard as well if the focus is on the play button.

packages/story-editor/src/elements/video/playPauseButton.js
```diff
@@ -209,6 +210,35 @@ function PlayPauseButton({
     };
   }, [checkMouseInBBox]);
 
+  const handleKeyDown = useCallback(
+    ({ key }) => {
+      if (allowKeyboardSupport && key === ' ') {
```
miina: Since `useKeyDownEffect` below is already handling the `space` part, there's no need to check for the key here.

packages/story-editor/src/elements/video/playPauseButton.js
```diff
@@ -209,6 +210,35 @@ function PlayPauseButton({
...
+        const videoNode = getVideoNode();
```
miina: Couldn't we just reuse `handlePlayPause` logic here?

packages/story-editor/src/elements/video/playPauseButton.js
amovar18: By default the focus while trimming would be on `cancel` button if user needs to use the keyboard support then they might have to click on the video element to be able to use the keyboard support if they do that wouldnt it be easier to use the mouse for play/pause button to toggle the video?
Edit: play button will appear into focus once user hovers on the video element.

miina: Sure, but the editor generally also aims to support keyboard-only users. There's no need to specifically forbid keyboard support in trim mode.

amovar18: ok then i guess there wouldnt be any need to have allowKeyboard support prop and also no need to have any separate logic to handleKeydown, will update the code.

miina (approved):
Thanks for working on this!

miina:
@swissspidy Since this can't be easily QA-ed (in the QA environment) and it's a quite small change, maybe we can just do another QA check by an engineer (in addition to me) and then merge, WDYT?

swissspidy (approved):
Works as expected!

PR #9758, opened 2021-11-19
Media: Adjust `needsProxy` check for poster images for base color generation

spacedmonkey:
## Summary

For hotlinked videos, poster images are uploaded to the media library. Meaning that poster could be served from mysite.com and video from an external site. Meaning that needsproxy on the resource only covers the asset not the poster.  

## Testing Instructions

1.

## Checklist

---

Fixes #

### DIFF
--- a/packages/story-editor/src/components/canvas/useInsertElement.js
+++ b/packages/story-editor/src/components/canvas/useInsertElement.js
@@ -71,7 +71,10 @@ function useInsertElement() {
       }
 
       try {
-        const needsProxy = await checkResourceAccess(imageSrc);
+        const needsProxy =
+          type === 'image'
+            ? resource?.needsProxy
+            : await checkResourceAccess(imageSrc);
         const imageSrcProxied = getProxiedUrl({ needsProxy }, imageSrc);
         const baseColor = await getMediaBaseColor(imageSrcProxied);
         updateElementById({
@@ -87,7 +90,7 @@ function useInsertElement() {
         // Do nothing for now.
       }
     },
-    [getProxiedUrl, updateElementById]
+    [getProxiedUrl, checkResourceAccess, updateElementById]
   );
 
   /**
--- a/packages/story-editor/src/components/canvas/useInsertElement.js
+++ b/packages/story-editor/src/components/canvas/useInsertElement.js
@@ -47,7 +47,7 @@ function useInsertElement() {
   const {
     actions: { registerUsage },
   } = useMedia3pApi();
-  const { getProxiedUrl } = useCORSProxy();
+  const { getProxiedUrl, checkResourceAccess } = useCORSProxy();
 
   const { setZoomSetting } = useLayout(({ actions: { setZoomSetting } }) => ({
     setZoomSetting,
@@ -71,7 +71,8 @@ function useInsertElement() {
       }
 
       try {
-        const imageSrcProxied = getProxiedUrl(resource, imageSrc);
+        const needsProxy = await checkResourceAccess(imageSrc);
+        const imageSrcProxied = getProxiedUrl({ needsProxy }, imageSrc);
         const baseColor = await getMediaBaseColor(imageSrcProxied);
         updateElementById({
           elementId: id,
--- a/packages/story-editor/src/utils/useCORSProxy.js
+++ b/packages/story-editor/src/utils/useCORSProxy.js
@@ -30,10 +30,38 @@ function useCORSProxy() {
     actions: { getProxyUrl },
   } = useAPI();
   const enableCORSProxy = useFeature('enableCORSProxy');
+
+  /**
+   * Check if the resource can be accessed directly.
+   *
+   * Makes a HEAD request, which in turn triggers a CORS preflight request
+   * in the browser.
+   *
+   * If the request passes, we don't need to do anything.
+   * If it doesn't, it means we need to run the resource through our CORS proxy at all times.
+   *
+   * @type {function(): boolean}
+   */
+  async function checkResourceAccess(link) {
+    let shouldProxy = false;
+    if (!link) {
+      return shouldProxy;
+    }
+    try {
+      await fetch(link, {
+        method: 'HEAD',
+      });
+    } catch (err) {
+      shouldProxy = true;
+    }
+
+    return shouldProxy;
+  }
+
   const getProxiedUrl = useCallback(
     (resource, src) => {
       const { needsProxy } = resource;
-      if (enableCORSProxy && needsProxy) {
+      if (enableCORSProxy && needsProxy && src) {
         return getProxyUrl(src);
       }
       return src;
@@ -43,6 +71,7 @@ function useCORSProxy() {
 
   return {
     getProxiedUrl,
+    checkResourceAccess,
   };
 }
 export default useCORSProxy;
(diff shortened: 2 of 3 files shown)


packages/story-editor/src/utils/useCORSProxy.js
```diff
@@ -30,10 +30,38 @@ function useCORSProxy() {
     actions: { getProxyUrl },
   } = useAPI();
   const enableCORSProxy = useFeature('enableCORSProxy');
+
```
spacedmonkey, line 33: This change is wholesale copied from https://github.com/google/web-stories-wp/pull/9713.

swissspidy: If you add this here you also need to remove it from `useInsert`

packages/story-editor/src/utils/useCORSProxy.js
```diff
@@ -30,10 +30,38 @@ function useCORSProxy() {
...
+    }
+
+    return shouldProxy;
+  }
+
   const getProxiedUrl = useCallback(
     (resource, src) => {
       const { needsProxy } = resource;
-      if (enableCORSProxy && needsProxy) {
+      if (enableCORSProxy && needsProxy && src) {
```
swissspidy, line 64: I don't particularly like this. If `src` is falsey/null, `getProxiedUrl` should probably not be called on it in the first place

packages/story-editor/src/components/canvas/useInsertElement.js
```diff
@@ -71,7 +71,8 @@ function useInsertElement() {
       }
 
       try {
-        const imageSrcProxied = getProxiedUrl(resource, imageSrc);
+        const needsProxy = await checkResourceAccess(imageSrc);
```
swissspidy: Poster images are never hotlinked, so they don't need that. We can really avoid that HTTP request.

packages/story-editor/src/components/canvas/useInsertElement.js
```diff
@@ -47,7 +47,7 @@ function useInsertElement() {
   const {
     actions: { registerUsage },
   } = useMedia3pApi();
-  const { getProxiedUrl } = useCORSProxy();
+  const { getProxiedUrl, checkResourceAccess } = useCORSProxy();
```
swissspidy, line 50: Note that lint error in this file now

packages/story-editor/src/components/canvas/useInsertElement.js
swissspidy:
```suggestion
        const needsProxy = type === 'image' && await checkResourceAccess(imageSrc);
```

a simple fix, albeit not  the best

spacedmonkey: No this is the wrong way around. We want to check for poster images here not images. A poster image on an hot linked image could is a local image ( not needing a proxy ) but a poster image on a coverr or tenor resource may need a proxy.

packages/story-editor/src/utils/useCORSProxy.js
spacedmonkey: I know, but there is nothing wrong with a little bit of harden here. This came up with a issue in unit tests.

packages/story-editor/src/components/canvas/useInsertElement.js
swissspidy: 3P media never needs a proxy

packages/story-editor/src/utils/useCORSProxy.js
swissspidy: We should update the tests to use proper data

packages/story-editor/src/components/canvas/useInsertElement.js
spacedmonkey: Okay, but we dont have a simple check for is hotlinked.

swissspidy: Sounds like adding `isHotlinked` to resources will make our lives a lot easier

spacedmonkey: > Sounds like adding isHotlinked to resources will make our lives a lot easier 

Let's no add another property like this. We already `local` and `isExternal` which are boolean. We should just add `source`. Values would be `local`, `library`, `hotlink`, `unsplash`, `tenor` or `coverr`. We could then even get rid of  `local` and `isExternal`. We can even add a little helper function called, isExternalResource

packages/story-editor/src/utils/useCORSProxy.js
spacedmonkey: So this function should only work if propper data is passed to it? We should not handle invalid data?

packages/story-editor/src/components/canvas/useInsertElement.js
spacedmonkey: ?

swissspidy: There was a lint error in this file

https://github.com/google/web-stories-wp/runs/4266687792?check_suite_focus=true

You just fixed it in 76425572f524e1715ba3b89b9bfe40f85e767008

packages/story-editor/src/components/canvas/useInsertElement.js
swissspidy: Interesting suggestion. Something I'd be willing to explore in a dedicated ticket.

packages/story-editor/src/utils/useCORSProxy.js
swissspidy: That should be handled at a different level

swissspidy: If you could point to the unit test you mentioned, that would be helpful.

packages/story-editor/src/components/canvas/useInsertElement.js
```diff
@@ -71,7 +71,11 @@ function useInsertElement() {
       }
 
       try {
-        const imageSrcProxied = getProxiedUrl(resource, imageSrc);
+        const needsProxy =
+          type === 'image'
+            ? resource?.needsProxy
+            : await checkResourceAccess(imageSrc);
+        const imageSrcProxied = getProxiedUrl({ needsProxy }, imageSrc);
```
swissspidy, line 78: Why not just

```js
const imageSrcProxied =
          type === 'video' ? imageSrc : getProxiedUrl(resource, imageSrc);
```

swissspidy: I don't see the need for `checkResourceAccess` when we already have `resource.needsProxy`

spacedmonkey: What if you are filtering in ( like we know some third parties are doing ) from a different media source? The fact is, poster and video could be served from different domains. It is wise here to be careful and just call `checkResourceAccess` for posters, as we do not know if they need a proxy or not. 

> const imageSrcProxied = type === 'video' ? imageSrc : getProxiedUrl(resource, imageSrc);

This feels extremely brittle.

packages/story-editor/src/utils/useCORSProxy.js
spacedmonkey: > That should be handled at a different level

That isn't DRY at all? And may result in someone forget to add this check upstream.

PR #9759, opened 2021-11-19
Video: Show correct flip state of video when in trim mode 

### LINKED ISSUE #9589: Video trimmer ignores flipped videos
## Bug Description

The video trimmer should display videos with the flipping settings, it has in the element on the canvas. And the video frame strip should do the same.

## Expected Behaviour

Element flip should be preserved.

## Steps to Reproduce

1. Add a video to the canvas
2. Flip it using the flip tools
3. Enter video trim mode
4. Observe that the video is no longer flipped

Trapsta:
## Context

The video trimmer should display videos with the flipping settings, it has in the element on the canvas. And the video frame strip should do the same.

## Relevant Technical Choices

For flipping the frames strip am using  `context.setTransform` instead of save and restore which [can be generally slow ](https://stackoverflow.com/a/38079796)

## To-do

n/a

 ## User-facing changes

Before

before

After
fixed

## Testing Instructions

1. Add a video to the canvas
2. Flip it using the flip tools
3. Enter video trim mode
4. Video and the video frame strip should be still flipped as before
5. Verify that trimming video still works 

n/a

n/a

n/a

## Checklist

---

Fixes #9589


packages/media/src/generateVideoStrip.js
```diff
@@ -112,6 +113,23 @@ async function generateVideoStrip(element, resource, stripWidth, stripHeight) {
...
+    const dx = frame * actualStripFrameWidth;
+
+    if (flip.vertical || flip.horizontal) {
+      // translate context to center of canvas
+      ctx.translate(
+        dx + actualStripFrameWidth / 2,
+        0 + actualStripFrameWidth / 2
+      );
+      // flip context
+      ctx.scale(flip.horizontal ? -1 : 1, flip.vertical ? -1 : 1);
```
barklund, line 101: Uh nice, didn't actually know this function. TIL!

packages/media/src/generateVideoStrip.js
```diff
@@ -112,6 +113,23 @@ async function generateVideoStrip(element, resource, stripWidth, stripHeight) {
...
+      // translate context back to orgin
+      ctx.translate(
+        -(dx + actualStripFrameWidth / 2),
+        -(0 + actualStripFrameWidth / 2)
+      );
```
barklund: It is however a little annoying, that we have to do this after the `scale()`. I guess it could be done in a single operation using `setTransform()`, but this is more obvious, so works for me!

packages/story-editor/src/components/videoTrim/trimmerComponents.js
```diff
@@ -49,6 +50,7 @@ export const Rail = styled.div`
   border-radius: 4px;
   border: 1px solid ${({ theme }) => theme.colors.standard.white};
   background-color: ${({ theme }) => theme.colors.standard.white};
+  ${elementWithFlip}
```
barklund: Is this necessary? Isn't the rail correctly rendered? I guess this doesn't do anything because we don't pass a `flip` property to the `<Rail />` element?

packages/media/src/generateVideoStrip.js
```diff
@@ -122,13 +140,16 @@ async function generateVideoStrip(element, resource, stripWidth, stripHeight) {
       actualFrameWidth,
       frameHeight,
       // Destination position
-      frame * actualStripFrameWidth,
+      dx,
```
barklund, line 116: Hm, somehow it now doesn't correctly render the frame inside the strip - if you resize the original video in only one dimension (by squeezing it longer or thinner), the strip ends up with extra whitespace either above or below the images and the frames themselves aren't fully inside the strip:

| Before | After |
|-|-|
| flipped-video-trim-before | flipped-video-trim-after |

This particular video is flipped in both directions, but I get the same issue if only flipped in one direction (and resized in one dimension). No issue if not flipped and resized.

barklund (commented):
There's a bug if you flip and resize the video in one dimension (so skewing the proportions). The rendered frame on the strip is offset for some reason.

packages/story-editor/src/components/videoTrim/trimmerComponents.js
miina: As Morten mentioned then this doesn't do anything without the correct properties passed to the `Rail` component.

Using flipping directly on `Rail` wouldn't work in case of horizontal flip since it would flip the whole vide strip and not just the single images inside it, it would also likely have some side effects for the other elements inside the the `Rail`.

Trapsta: Yes, this is unnecessary, would flip the entire canvas instead of the individual frames. Removing

packages/media/src/generateVideoStrip.js
Trapsta: @barklund  Good catch, didn't notice this at first. I was thinking it's because of the `ctx.translate()` but am seeing it also without any flip and on main as well. It might be related to the size of the video we are using to extract the frames. looking into it but in the meantime , could you confirm if you are not getting the same on main?

packages/media/src/generateVideoStrip.js
Trapsta: Yes, I see what you mean. I did try using `ctx.setTransform()` initially but I couldn't get it to work consistently

packages/media/src/generateVideoStrip.js
miina: @Trapsta I was testing it out as well (with the same video as Morten) and don't see the issue on `main` but can see it on this PR.
(In Morten's comment, the *Before* column of the table is from `main`.)

miina:
@Trapsta When you have a moment, could you merge `main` meanwhile? Checking out this PR can cause errors currently due to base color changing to hex instead of rgb on `main`.

packages/media/src/generateVideoStrip.js
```diff
@@ -112,6 +113,20 @@ async function generateVideoStrip(element, resource, stripWidth, stripHeight) {
     // Seek to the next offset
     await seekVideo(video, timeOffset);
 
+    const dx = frame * actualStripFrameWidth;
+
+    if (flip.vertical || flip.horizontal) {
+      // translate context to center of canvas
```
miina: Nit (for the comments below as well):
```suggestion
      // Translate context to the center of the canvas.
```

packages/media/src/generateVideoStrip.js
```diff
@@ -112,6 +113,20 @@ async function generateVideoStrip(element, resource, stripWidth, stripHeight) {
...
+      ctx.translate(dx + actualStripFrameWidth / 2, 0 + stripFrameHeight / 2);
```
miina: Is `0` needed here (and below)? 😛

packages/media/src/generateVideoStrip.js
```diff
@@ -58,6 +58,7 @@ async function generateVideoStrip(element, resource, stripWidth, stripHeight) {
     scale,
     focalX,
     focalY,
+    flip,
```
miina, line 40: Hm, there's one issue currently with the video strip -- it looks like if flipping while in trim mode, it doesn't update the strip, only the image on the Page gets updated.

packages/media/src/generateVideoStrip.js
Trapsta: @miina the issue was I was translating vertically using the width instead of the height, after fixing that am now getting the same thing am seeing on main which is a little strip of white and the frames in between (Video is resized to 100 width x 300 height)
Screenshot 2021-11-23 at 15 27 09

miina: Which browser are you using? I can see everything correctly on `main`, the feature also has passed QA and bug-bash without anyone reporting this issue 🤔 

Could you create a story in the QA environment where this issue is visible and send the link to the story for checking?

Trapsta: Am using chrome, was actually just discussing it with Morten -  try this story on QA https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/post.php?post=15287&action=edit

packages/media/src/generateVideoStrip.js
Trapsta: Got it. Should be fixed now although takes a while to regenerate the strip in the correct orientation if you change the flip settings while in flip mode

miina: Hm, I wonder if there's a way to reuse the previous images (store the originals after generation) and flip as needed before putting together?
I assume the frame generation is the main thing that takes time? Since we should need the same frames (but just flipped) then could we reuse these. WDYT?

Trapsta: Yes, correct generating the frames takes a bit of time but am thinking there are way too many variables to make it useful to store the generated images. position, size , rotation, length and flip settings of both the video and canvas  all factor in the generated frames.

miina: Isn't it possible to store the frames as they were originally before the changes in this PR, and only flip it afterwards for strip generation if needed?

miina: (We can live with it taking a bit of time to flip the strip if it gets very complicated, asking just in case though!)

Trapsta: I do think it would be possible but not sure if the benefits outweigh the added complexity since we would need to regenerate the frames anyway incase any of the other variables changed  but lets ping @barklund to see if he has more clarity on this since he knows the feature a bit more incase am wrong

barklund: I'm not sure I understand the question here. What is the exact goal and what are the possible solutions and their pros and cons?

I need this spelled out in big cardboard letters here to follow along 🙈 .

Trapsta: @barklund we are just discussing with Miiina if it would be good to save and reuse the generated frames for the video strip. So we would save the frames the first time we generate them and then instead of re-generating them on subsequent renders, we would use the stored frames. The  pros would be the frames strip would render faster since we would eliminate the need to generate the frames all the time but the catch is that the frames generated depend on so many variables like the position of the video in the canvas, the size, orientation , rotation such that  am thinking we would mostly need to generate the frames most of the time anyway due to some of these variables changing. Too much rumbling, hope am making sense 😄

miina: @barklund While a user is in trim mode and changes something of the video (e.g. flips the video), the strip is fully regenerated and updating the strip takes some time. However, the origin video is actually the same, and the generated frames as well, so the idea was that perhaps we could store the initial frames and then reuse those if possible instead.

(If doing that gets too complicated, we can just leave it as-is, too, it's already working well as it is)

barklund: Hmm, as Jasper mentions, there are too many things that would go into the cache as parameters, so I don't think it's worth it.

miina: Alright, sgtm! @Trapsta is the PR otherwise ready for review (since it was moved back to In Progress yesterday)?

Trapsta: @miina yes,  will update the status

packages/media/src/generateVideoStrip.js
miina: @Trapsta If this still remains an issue after this PR, would you mind creating a new issue for fixing it?

Trapsta: Cool, np

miina (approved):

packages/media/src/generateVideoStrip.js
Trapsta: @miina added https://github.com/google/web-stories-wp/issues/9870

PR #9724, opened 2021-11-16
Editor: Add collapsing carousel

barklund:
## Context

This adds a collapsing carousel as per Figma designs.

## To-do

- [x] Tests

## User-facing changes

| State | Video |
|-|-|
| Wide page | carousel-collapse-wide-fast |
| Narrow page with many thumbnails | carousel-collapse-narrow-multipage-fast |
| Narrow page with a single thumbnail | carousel-collapse-narrow-singlepage-fast |

## Testing Instructions

This PR can be tested by following these steps:

1. Observe there is a new bar above the page carousel with the current page number and the total page count
2. Click this bar to collapse the page carousel
3. Observe the collapse state with the page count bar perfectly aligning in the middle of the buttons to either side and that the canvas above has expanded to use the extra space
4. Click the bar again to expand the page carousel
5. Observe that everything works as expected and that the canvas has shrunk to make room for the carousel.

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [x] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [x] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #9641

Issue #9549, opened 2021-10-27
30+ page stories causes image/template browsing feature to use 4X client memory


brianpetro:
To reproduce, in a story with 30+ pages,

1) Hard refresh.
2) Scroll through the images or templates section.
3) Observe substantially increased memory usage when compared to a new/blank story.

Note: the difference in memory usage between the blank story and the 30+ page story is less than 2X.

These features work well when the story is short. But then stop working, or crash the client's computer, when the story is long.  

What might be causing this?

Thanks

swissspidy:
Thanks for raising this!

We're painfully aware of the performance shortcomings in the editor, especially in those two areas.

That's why we're overhauling the templates section in the next release to replace the dynamic previews with simple static images (see #9289). Especially when scrolling through the templates list the experience will be much smoother.

Besides that, we're actively working on addressing memory and performance issues throughout the editor, from the canvas to the media library.

We'll take a look at this specific scenario that you're mentioning to identify possible causes for this.

cc @choumx @miina - let's discuss this further to get to the bottom of this.

brianpetro:
Thanks for the always-speedy response @swissspidy!

dreamofabear:
Thanks for the report, Brian!

> Scroll through the images or templates section.

Can you clarify which sections? 

Does "images" mean the "WordPress media" tab on the left-side in the editor? Or the second "Third-party media" tab?

<img width="347" alt="Screen Shot 2021-10-27 at 6 12 35 PM" src="https://user-images.githubusercontent.com/20525523/139154859-e70ede61-e2c0-4b48-afaf-0d2f0299f1b8.png">

Does "templates" mean the templates on the dashboard or the "Page templates" tab in the editor?

brianpetro:
@choumx yeah, of course, I was referring to both the "Third-party media" and the "Page templates" tabs in the editor.

Thanks!

brianpetro:
Similarly, "rearranging pages in Grid view" may be added to the list.

swissspidy:
Ah yes, that's a good point. I think our reordering components (in grid view and the carousel) as well as the Moveable integration in the library are really not performant

miina:
> I think our reordering components (in grid view and the carousel) as well as the Moveable integration in the library are really not performant

Moveable should in theory make a difference in the library only for the element that's being hovered on since it's only displayed for a hovered/active element. Or do you mean the dragging itself is not performant?

[...]

PR #9760, opened 2021-11-19
Fixing carousel cache

### LINKED ISSUE #9749: Fix thumbnail image generation
## Feature Description

EDIT: A quick fix was implemented meanwhile (#9760), this issue now focuses on the long-term solutions, see also https://github.com/GoogleForCreators/web-stories-wp/issues/9749#issuecomment-992667669

Thumbnail images are generated when a preview page is displayed and there's idle time in the browser.

This happens in the carousel, the checklist, and the grid view.

However, these thumbnail images are generated locally in each component needing them, regardless of whether another preview of the exact same page has been generated elsewhere in the application. Also, this happens even if the thumbnail isn't visible (e.g. in the case of a carousel thumbnail not currently visible due to scroll).

The "generated locally" in the component is extra bad for the new collapsing carousel, as all the pages are unmounted on collapse, which means that on expansion, all the images have to be re-generated.

Furthermore, if multiple page previews are visible at the same time, they might all trigger their idle callback at the same time, and then all the thumbnails will be generated in the same JS frame because there's no queuing or synchronization between the components.

## Solutions considered

Short-term solutions include:

1. Generate images of each page on story load in a queue when the browser is idle (only one image at a time, then request a new idle callback)
  – Then store these images locally in the browser so it's accessible to all components.
  – Update images when a page changes (but in an idle callback).
2. Only generate thumbnail images for the carousel, not in the checklist nor the grid view
3. Only generate thumbnail images for visible pages in the carousel, not the current invisible ones

One or more of the above solutions can be implemented at the same time, last they complement each other rather than compete with each other.

Long-term solutions include:

4. Store images in WP media storage
  – This has some issues with multiple revisions of large stories potentially taking up 100 or even 1000s of physical files on the server
  – Solutions to this include storing only images for the current revision deleting for old revisions as well as storing all images in a single physical file either as a sprite or as a multi-frame image (gif or webp).
5. Store images as base64 in the database along with the story JSON.
  – Base64 wastes about 33% overhead, so images will be even larger than physical files
  – However, the load will be much faster, as only a single load will load both the story data and all the images

For either of the above, image generation scheduling still has to be figured out - when should the image be generated and how do we ensure the browser doesn't hang while it happens? And the above short-term solutions might still all be relevant in the long term.

Both the above have the added advantage of being able to preview stories (as flat images only though, no animation or video) which can be used in both the dashboard and in the Gutenberg blocks inserting stories into pages.

barklund:
## Context

This PR moves the thumbnail image cache storage from the individual thumbnail to the carousel provider and abstracts it out of the thumbnail class itself.

This fixes a number of issues:

- Cache images are generated for the pages in the checklist for no good reason. The checklist no longer used image caching after this PR.
- (Very large and expensive) cache images are generated for the pages in the grid view. The grid view no longer uses image caching after this PR.
- The page carousel would regenerate all cached images when the carousel remounts, which does not make sense if the pages haven't actually changed. (only relevant post-merge of #9724)

## Summary

* This is not a great solution. It's a temporary short-term fix and it has some minor issues, but it works for the purpose it serves - solving the issue here and now.

## User-facing changes

- Less memory-intensive when opening checklist or grid view
- Less memory-intensive when re-expanding the page carousel (post-merge of #9724)

## Testing Instructions

1. Create a story with 30 pages (just use 30 random default page templates)
2. Open the grid view
3. Observe that the browser no longer struggles to function while all the image caches are being generated, because this does not happen anymore.

**NB**: Observe this [known bug](https://github.com/google/web-stories-wp/pull/9760#discussion_r753579779).

## Checklist

---
Maybe helps with #9549 
Partial for #9749

### DIFF
--- a/packages/story-editor/src/components/footer/pagepreview/index.js
+++ b/packages/story-editor/src/components/footer/pagepreview/index.js
@@ -117,11 +117,15 @@ function PagePreview({
   // Whenever the page is re-generated
   // remove the old (and now stale) image blob
   useEffect(() => {
-    if (enableThumbnailCaching && pageAtGenerationTime.current !== page) {
+    if (
+      enableThumbnailCaching &&
+      isActive &&
+      pageAtGenerationTime.current !== page
+    ) {
       setCachedImage({ pageId: page.id, cachedImage: null });
       pageAtGenerationTime.current = null;
     }
-  }, [page, setCachedImage, enableThumbnailCaching]);
+  }, [page, setCachedImage, isActive, enableThumbnailCaching]);
 
   useEffect(() => {
     // If this is not the active page, there is a page node, we
--- a/packages/story-editor/src/components/footer/pagepreview/index.js
+++ b/packages/story-editor/src/components/footer/pagepreview/index.js
@@ -97,36 +97,47 @@ const Image = styled.img`
 `;
 
 // PagePreview is used in the editor's Carousel as well as in the Checklist and GridView
-function PagePreview({ page, label, ...props }) {
+function PagePreview({
+  page,
+  label,
+  isCacheable = false,
+  cachedImage = null,
+  setCachedImage = null,
+  ...props
+}) {
   const { backgroundColor } = page;
   const { width, height, isActive } = props;
 
-  const [imageBlob, setImageBlob] = useState();
   const [pageNode, setPageNode] = useState();
   const setPageRef = useCallback((node) => node && setPageNode(node), []);
   const pageAtGenerationTime = useRef();
-  const enableThumbnailCaching = useFeature('enableThumbnailCaching');
+  const enableThumbnailCaching =
+    useFeature('enableThumbnailCaching') && isCacheable;
 
   // Whenever the page is re-generated
   // remove the old (and now stale) image blob
   useEffect(() => {
-    if (pageAtGenerationTime.current !== page) {
-      setImageBlob(null);
+    if (enableThumbnailCaching && pageAtGenerationTime.current !== page) {
+      setCachedImage({ pageId: page.id, cachedImage: null });
       pageAtGenerationTime.current = null;
     }
-  }, [page]);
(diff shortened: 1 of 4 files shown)


packages/story-editor/src/components/footer/pagepreview/index.js
```diff
@@ -97,36 +97,51 @@ const Image = styled.img`
...
+    if (
+      enableThumbnailCaching &&
+      isActive &&
+      pageAtGenerationTime.current !== page
+    ) {
+      setCachedImage({ pageId: page.id, cachedImage: null });
       pageAtGenerationTime.current = null;
     }
-  }, [page]);
+  }, [page, setCachedImage, isActive, enableThumbnailCaching]);
```
barklund, line 128: This is a bit ugly. Basically, I had to add `isActive` here because otherwise when the carousel remounted, all the pages would report themselves as being "new", because they basically are - they all forgot which page they snapshotted, so they all just think they have to re-snapshot, which would re-introduce the problem I was trying to fix anyway.

This particular "fix" can lead to stale images being shown in the carousel because if you:

* close the carousel
* switch page
* make some changes
* switch page again
* re-expand the carousel

The now inactive page, that changed while the carousel was collapsed, will be showing the wrong (old) thumbnail.

This is for sure a bug but solving it requires moving more of the snapshot logic to the carousel provider, which is only a temporary location for the caching logic anyway, as all the logic has to go further up the stack to be useful throughout the application. I did not want to go that far in this PR, so decided this was an acceptable middle-ground.

The above bug is probably pretty rare to see and will fix itself as soon as you switch to the page with the wrong thumbnail to make it the active page.

I felt this was a reasonable "bug" to leave in to improve the rendering without doing too much work, which would have to be re-done anyway at a later point when we implement the proper long-term fix for the image cache.

miina (approved):
Thanks for creating this PR so fast.
LGTM as a short-term fix - Improves the current situation.

PR #9775, opened 2021-11-23
Decoupling: Introduce `initialEdits` and remove `getDemoStoryById`

### LINKED ISSUE #9628: Decoupling: Move demo story logic from `story-editor`
## Feature Description

As discussed here https://github.com/google/web-stories-wp/pull/9569#discussion_r739076535.
Since `getDemoStoryById` is WordPress specific, it should not be part of the core story editor. 

> That probably means reverting most of #5507.

> I've always felt that using `get_default_post_to_edit()` or having something that works more akin to how Gutenberg loads the post would be cleaner than the current `useLoadStory` logic.

> It's essentially this story: https://wp.stories.google/stories/tips-web-stories-editor/

> Admittedly, it's a very hidden gimmick that probably has very low usage. So this is not a priority at all.

> In fact, I might even suggest to the team that we remove this feature entirely.

## Additional Context

Related https://github.com/google/web-stories-wp/issues/2918

mohdsayed:
## Context

Currently, the core `story-editor` requires `getStoryById` or `getDemoStoryById` API callback to fetch story for the editor. In the context of WordPress, we are [preloading few API Paths ](https://github.com/google/web-stories-wp/blob/777b83a5ca3e7556f8e4b303cd08fe7d6f1640a1/includes/templates/admin/edit-story.php#L39-L117) including the story, however, this is not ideal for a third-party integration or a standalone story editor for instance a playground which only has to load the story from localStorage instead of fetching it using some API Path.

See related discussion
https://github.com/google/web-stories-wp/pull/9569#discussion_r739173847

## Summary

This PR removes the concept of demo story from `story-editor` as raised in https://github.com/google/web-stories-wp/issues/9628 and introduces `initialEdits` which can be used for demo story as well as for a standalone editor.

Also, note that we are still preloading the story path because we are also doing it for other items like media, users, taxonomies [here](https://github.com/google/web-stories-wp/blob/777b83a5ca3e7556f8e4b303cd08fe7d6f1640a1/includes/templates/admin/edit-story.php#L39-L117), I think it would make sense to remove it when we remove all other items.

## Testing Instructions

1.

## Checklist

---

Fixes https://github.com/google/web-stories-wp/issues/9628


packages/story-editor/src/app/story/effects/useLoadStory.js
```diff
@@ -28,134 +28,130 @@ import { useHistory } from '../../history';
...
-          pages,
-          story,
-          selection: [],
-          current: null, // will be set to first page by `restore`
-          capabilities,
+      if (story) {
+        loadStory(storyId, story, restore, clearHistory);
+      } else {
+        getStoryById(storyId).then((post) => {
+          loadStory(storyId, post, restore, clearHistory);
```
swissspidy, line 150: So we need to do this if no story was provided via `initialEdits`? When would that be the case?

includes/templates/admin/edit-story.php
```diff
@@ -116,6 +83,48 @@
...
+		),
+	]
+);
+
+if ( ! wp_validate_boolean( $demo ) ) {
+	$preload_paths[] = $story_path;
+} else {
+	$story_data             = \Google\Web_Stories\rest_preload_api_request( [], $story_path );
+	$initial_edits['story'] = ( ! empty( $story_data[ $story_path ]['body'] ) ) ? $story_data[ $story_path ]['body'] : [];
+}
```
swissspidy, line 126: Observation:

I originally thought `initialEdits` would be a nice opportunity to move the demo story logic out of the REST API controller:

https://github.com/google/web-stories-wp/blob/f7a7f7a2b8159b0c4c4ff59457bca2c2792ca607/includes/REST_API/Stories_Controller.php#L81-L88

Plus we wouldn't need to make auto-drafts available via the REST API too:

https://github.com/google/web-stories-wp/blob/f7a7f7a2b8159b0c4c4ff59457bca2c2792ca607/includes/REST_API/Stories_Controller.php#L229

Similar to how `initial_edits` works in Gutenberg:

https://github.com/WordPress/wordpress-develop/blob/cb67c52be2971af8a304dbbf27e9cd78f1bfaee9/src/wp-admin/edit-form-blocks.php#L76-L96

https://github.com/WordPress/wordpress-develop/blob/cb67c52be2971af8a304dbbf27e9cd78f1bfaee9/src/wp-admin/edit-form-blocks.php#L76-L96

But now that I see it in action, it seems less possible. Hmm.

includes/Admin/Editor.php
```diff
@@ -279,15 +284,17 @@ public function admin_enqueue_scripts( $hook ) {
...
 
 		wp_localize_script(
 			self::SCRIPT_HANDLE,
-			'webStoriesEditorSettings',
-			$this->get_editor_settings()
+			'webStories',
+			[
+				'publicPath' => $this->assets->get_base_url( 'assets/js/' ), // Required before the editor script is enqueued.
+				'localeData' => $this->assets->get_translations( self::SCRIPT_HANDLE ), // Required for i18n setLocaleData.
+				'metaBoxes'  => $this->meta_boxes->get_meta_boxes_per_location(), // Because meta boxes are not available inside edit-story.php.
```
swissspidy: Odd. So the `$wp_meta_boxes` global is `null`?

includes/Admin/Editor.php
```diff
@@ -279,15 +284,17 @@ public function admin_enqueue_scripts( $hook ) {
...
 		$this->assets->enqueue_style_asset( self::SCRIPT_HANDLE, [ $this->google_fonts::SCRIPT_HANDLE ] );
 
 		wp_localize_script(
 			self::SCRIPT_HANDLE,
-			'webStoriesEditorSettings',
-			$this->get_editor_settings()
+			'webStories',
+			[
+				'publicPath' => $this->assets->get_base_url( 'assets/js/' ), // Required before the editor script is enqueued.
+				'localeData' => $this->assets->get_translations( self::SCRIPT_HANDLE ), // Required for i18n setLocaleData.
```
swissspidy, line 295: As an alternative we could think about exposing `setLocaleData` under `window.webStories.setLocaleData`, and then make these `setLocaleData` calls inline. Then we wouldn't to pass this huge list around.

Not sure that's really an advantage though 🤔

packages/story-editor/src/app/story/effects/useLoadStory.js
mohdsayed: In the context of WordPress, we are only providing the demo story via `initialEdits` ( playground can also use `initialEdits` to load story ). Regular stories are still being loaded via `getStoryById` using the preloaded API data in. the same way.

includes/Admin/Editor.php
mohdsayed: Yes, but thanks for pointing that out. I think its because we are trying to get it before `meta-boxes.php` is included. Let me try to move after the file inclusion. 

https://github.com/google/web-stories-wp/blob/444e2cdf625a8c45aa3348fa071f4c0ba175f98d/includes/templates/admin/edit-story.php#L154-L173

mohdsayed: Yes, that was the reason, done in https://github.com/google/web-stories-wp/pull/9775/commits/1a8db57d898caa79b448c9f1fa18438cf5e64f4c

includes/Admin/Editor.php
mohdsayed: But shouldn't `setLocaleData(localeData);` be called before the script is loaded. 
comment 👇 

https://github.com/google/web-stories-wp/blob/d7c7f712510516c13277d128fd907080dab7bfdd/packages/wp-story-editor/src/index.js#L26-L28

and `setLocalData` is included in the package, meaning it will not be available before the script is loaded.

https://github.com/google/web-stories-wp/blob/d7c7f712510516c13277d128fd907080dab7bfdd/packages/wp-story-editor/src/setLocaleData.js#L20

So how can we add an inline script?

swissspidy: Ah, you're right 🤦  never mind then

includes/templates/admin/edit-story.php
mohdsayed: Yeah, currently it looks like the easiest way to get demo story, maybe we will revisit this in the future, not sure about it now.

packages/wp-story-editor/src/index.js
```diff
@@ -47,14 +47,19 @@ import {
...
  * Initializes the web stories editor.
  *
- * @param {string} id       ID of the root element to render the screen in.
- * @param {Object} config   Story editor settings.
+ * @param {string} id           ID of the root element to render the screen in.
+ * @param {Object} config       Story editor settings.
+ * @param {Object} initialEdits Initial edits.
  */
-const initialize = (id, config) => {
+window.webStories.initializeStoryEditor = (id, config, initialEdits) => {
```
swissspidy, line 62: We might want to do the same for the dashboard in the future, for consistency.

mohdsayed: Absolutely!

swissspidy (approved):

PR #9776, opened 2021-11-23
Editor: Allow disabling drop-targeting using the modifier key

### LINKED ISSUE #6383: Disable automatic drop targeting by shortcut
## Feature Description

Per user feedback, it would be nice to be able to place images on top of each other without them dropping into each other because of the automatic drop targeting currently in function.

* [ ] Disable drop targeting by shortcut: <kbd>Cmd</kbd> on Mac, <kbd>Win</kbd> on Windows.
* [ ] Add description of shortcut to the list of shortcuts (by mentioning it under "disable snapping" as per below design)

Screen Shot 2021-02-12 at 17 35 16

---

miina:
## Summary
Deactivates drop-targeting and snapping both while holding down the modifier key.

## User-facing changes
No drop-targeting nor snapping happening when dragging while holding down the modifier:
modifier

## Testing Instructions

1. Press down the modifier key (`⌘` for Mac,  `Ctrl` for Windows)
2. Drag elements around that would usually start using the drop-targeting feature (media / shapes).
3. Verify that both Snapping and drop-targeting are disabled when holding down the key.
4. Verify that the snapping and drop-targeting work as expected otherwise.

No

No

No

## Checklist

---

Fixes #6383

### DIFF
--- a/packages/story-editor/src/components/canvas/utils/useSnapping.js
+++ b/packages/story-editor/src/components/canvas/utils/useSnapping.js
@@ -19,7 +19,7 @@
  */
 import { useCallback } from '@web-stories-wp/react';
 import { FULLBLEED_RATIO } from '@web-stories-wp/units';
-import { useGlobalIsKeyPressed } from '@web-stories-wp/design-system';
+
 /**
  * Internal dependencies
  */
@@ -45,15 +45,17 @@ function useSnapping({
       pageHeight,
     })
   );
-  const { activeDropTargetId } = useDropTargets((state) => ({
-    activeDropTargetId: state.state.activeDropTargetId,
-  }));
+  const { activeDropTargetId, isDropTargetingDisabled } = useDropTargets(
+    (state) => ({
+      activeDropTargetId: state.state.activeDropTargetId,
+      isDropTargetingDisabled: state.state.isDropTargetingDisabled,
+    })
+  );
 
   const triggerOnboarding = useUserOnboarding(({ SAFE_ZONE }) => SAFE_ZONE);
 
-  // ⌘ key disables snapping
-  const snapDisabled = useGlobalIsKeyPressed('meta');
-  canSnap = canSnap && !snapDisabled && !activeDropTargetId;
+  // Drop-targeting is disabled with ⌘ key, we also disable snapping in this case.
+  canSnap = canSnap && !isDropTargetingDisabled && !activeDropTargetId;
 
   const handleSnap = useCallback(
     ({ elements }) => {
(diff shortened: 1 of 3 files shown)


packages/story-editor/src/components/canvas/utils/useSnapping.js
```diff
@@ -45,15 +45,17 @@ function useSnapping({
...
+    })
+  );
 
   const triggerOnboarding = useUserOnboarding(({ SAFE_ZONE }) => SAFE_ZONE);
 
-  // ⌘ key disables snapping
-  const snapDisabled = useGlobalIsKeyPressed('meta');
-  canSnap = canSnap && !snapDisabled && !activeDropTargetId;
+  // Drop-targeting is disabled with ⌘ key, we also disable snapping in this case.
+  canSnap = canSnap && !isDropTargetingDisabled && !activeDropTargetId;
```
miina, line 58: Using `useGlobalIsKeyPressed` doesn't work when used in 2 separate places, so opted for moving it to the drop-targeting provider (since that's accessible in both cases).

packages/story-editor/src/components/dropTargets/provider.js
```diff
@@ -203,20 +205,25 @@ function DropTargetsProvider({ children }) {
...
     actions: {
-      registerDropTarget,
+      registerDropTarget: isDropTargetingDisabled ? noop : registerDropTarget,
       unregisterDropTarget,
-      isDropSource,
-      isDropTarget,
-      handleDrag,
-      handleDrop,
-      setDraggingResource,
+      isDropSource: isDropTargetingDisabled ? () => false : isDropSource,
```
miina, line 222: Since drop-targeting actions are used in various places, the easiest way seemed to be just disabling these directly from the provider instead of changing each usage separately + tracking potential future cases. Doesn't look the cleanest but seemed to be the cleanest approach. LMK in case of alternative ideas.

barklund: Hmm, it does the job very efficiently. But it's a bit dirty and non-transparent, but I also cannot come up with a cleaner solution.

barklund: One minor "bug" though:

If you start dragging without pressing mod, hover a drop target and observe the potential drop, then press mod, then drag around more, the item is stuck to the drop target and can't "get out" until you release mod and the press mod again. But I'm not sure it's worth fixing, as it's so easy to get around.

barklund (approved):
🎆

packages/story-editor/src/components/dropTargets/provider.js
miina: >  But I'm not sure it's worth fixing, as it's so easy to get around.

Hm, I'm not even certain what should happen in this case. Probably it's an edge case that we can address if it gets reported.

PR #9780, opened 2021-11-23
Media: Fix Media fetching karma test

### LINKED ISSUE #9779: Media fetching test failures
## Bug Description

When `fixture.screen` isn't ready, we throw an error causing the `MediaPane fetching should fetch 2nd page` karma test to fail. 

## Expected Behaviour

It should wait for the `fixture.screen` results to be present before moving forward. It shouldn't throw an error. 

## Steps to Reproduce

Reported [here](https://github.com/google/web-stories-wp/pull/9757#issuecomment-976565540) 

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

BrookeGraham:
## Context

Media fetching karma test is throwing an error when the fixture has yet to be set up causing the test to fail.

## Summary

When the mediaFetch call is initially made it takes a little bit of time before the elements are actually rendered into their container. In some instances this meant we were making an additional api call to grab more media items because we thought there wasn't enough items present to fully fill the container's height. This unnecessary second call would cause the media fetch tests to fail in ci. 
## Relevant Technical Choices

Now we are ensuring an additional page call is only made if the container isn't full after a 500 ms delay. Using a debouncedCallback to set the delay and then checking if the second call is still needed before calling `setNextPage`.

## Testing Instructions

1. Make sure the all the media call are still being made by clicking through both local media and 3P media tabs. 
2. On the 3p video tab, if you make your screen resolution really really small, you can see the slight time delay between the first and second media batches. 

no

no

no
## Checklist

---

Fixes #9779

### DIFF
--- a/packages/story-editor/src/components/library/panes/media/common/paginatedMediaGallery.js
+++ b/packages/story-editor/src/components/library/panes/media/common/paginatedMediaGallery.js
@@ -80,12 +80,6 @@ function PaginatedMediaGallery({
       return;
     }
 
-    // Load the next page if the page isn't full, ie. scrollbar is not visible.
-    if (node.clientHeight === node.scrollHeight) {
-      setNextPage();
-      return;
-    }
-
     // If scrollTop is zero, we know we do not need to fetch an additional page yet.
     if (node.scrollTop === 0) {
       return;
--- a/packages/story-editor/src/components/library/panes/media/common/paginatedMediaGallery.js
+++ b/packages/story-editor/src/components/library/panes/media/common/paginatedMediaGallery.js
@@ -65,14 +65,15 @@ function PaginatedMediaGallery({
   // State and callback ref necessary to load on scroll.
   const refContainer = useRef();
 
-  const isNextPageNeeded = () => {
+  const isNextPageNeeded = useCallback(() => {
+    // Load the next page if the container still isn't full, ie. scrollbar is not visible.
     if (
       refContainer.current.clientHeight === refContainer.current.scrollHeight
     ) {
-      return setNextPage;
+      return setNextPage();
     }
     return () => {};
-  };
+  }, [setNextPage]);
 
   const debouncedSetNextPage = useDebouncedCallback(isNextPageNeeded, 500);
 
@@ -91,9 +92,13 @@ function PaginatedMediaGallery({
       return;
     }
 
-    // Load the next page if the page isn't full, ie. scrollbar is not visible.
+    // When the node.scrollHeight is first calculated it may not be accurate
+    // depending on if the elements have fully render on the page. If the container
+    // height and scroll height are the same, debounce the call to allow time
+    // for the elements to paint before making an additional setNextPage call.
     if (node.clientHeight === node.scrollHeight) {
       debouncedSetNextPage();
+      return;
     }
 
     // Load the next page if we are "close" (by a length of ROOT_MARGIN) to the
--- a/packages/story-editor/src/components/library/panes/media/common/paginatedMediaGallery.js
+++ b/packages/story-editor/src/components/library/panes/media/common/paginatedMediaGallery.js
@@ -65,6 +65,17 @@ function PaginatedMediaGallery({
   // State and callback ref necessary to load on scroll.
   const refContainer = useRef();
 
+  const isNextPageNeeded = () => {
+    if (
+      refContainer.current.clientHeight === refContainer.current.scrollHeight
+    ) {
+      return setNextPage;
+    }
+    return () => {};
+  };
+
+  const debouncedSetNextPage = useDebouncedCallback(isNextPageNeeded, 500);
+
   const loadNextPageIfNeeded = useCallback(() => {
     const node = refContainer.current;
     if (
@@ -80,9 +91,9 @@ function PaginatedMediaGallery({
       return;
     }
 
-    // If scrollTop is zero, we know we do not need to fetch an additional page yet.
-    if (node.scrollTop === 0) {
-      return;
+    // Load the next page if the page isn't full, ie. scrollbar is not visible.
+    if (node.clientHeight === node.scrollHeight) {
+      debouncedSetNextPage();
     }
 
     // Load the next page if we are "close" (by a length of ROOT_MARGIN) to the
@@ -93,7 +104,14 @@ function PaginatedMediaGallery({
       setNextPage();
       return;
     }
-  }, [resources, hasMore, isMediaLoaded, isMediaLoading, setNextPage]);
+  }, [
+    resources.length,
+    hasMore,
+    isMediaLoaded,
+    isMediaLoading,
+    debouncedSetNextPage,
+    setNextPage,
+  ]);
 
   // Scroll to the top when the searchTerm or selected category changes.
   useEffect(() => {
--- a/packages/story-editor/src/components/library/panes/media/common/paginatedMediaGallery.js
+++ b/packages/story-editor/src/components/library/panes/media/common/paginatedMediaGallery.js
@@ -73,6 +73,8 @@ function PaginatedMediaGallery({
       // This condition happens when the component is hidden, and causes the
       // calculation below to load a new page in error.
       node.clientHeight === 0 ||
+      // If scrollTop is zero, we know we do not need to fetch an additional page yet.
+      node.scrollTop === 0 ||
       !hasMore ||
       !isMediaLoaded ||
       isMediaLoading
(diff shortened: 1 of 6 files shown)


packages/story-editor/src/components/library/panes/media/local/karma/mediaFetching.karma.js
```diff
@@ -49,11 +49,12 @@ describe('MediaPane fetching', () => {
     const initialElements =
       within(mediaGallery).queryAllByTestId(/^mediaElement-/);
 
+    // ensure fixture.screen has loaded before calling expect to prevent immediate failure
+    if (initialElements.length !== LOCAL_MEDIA_PER_PAGE) {
+      await fixture.events.sleep(500);
+    }
+
```
samwhale: instead of sleeping here we could use `waitFor`. If that time is less than 500 ms then this won't fail, but this will also pass if it takes longer than 500 ms.
```suggestion
    await waitFor(() => {
      if (initialElements.length !== LOCAL_MEDIA_PER_PAGE) {
        throw new Error('Not ready - initial media elements not loaded yet');
      }
    });

```

packages/story-editor/src/components/library/panes/media/local/karma/mediaFetching.karma.js
```diff
@@ -62,15 +63,15 @@ describe('MediaPane fetching', () => {
...
-      ) {
-        throw new Error('wait');
-      }
+    // ensure fixture.screen has loaded before calling expect to prevent immediate failure
+    if (
+      fixture.screen.queryAllByTestId(/^mediaElement-/).length <
+      LOCAL_MEDIA_PER_PAGE * 2
+    ) {
+      await fixture.events.sleep(500);
+    }
```
samwhale: Same here 😄

packages/story-editor/src/components/library/panes/media/local/karma/mediaFetching.karma.js
BrookeGraham: That is what we are already doing. So when the if statement is true the error gets thrown resulting in a failed test.

BrookeGraham: At least I think that would be essentially the same thing as we are currently doing.

packages/story-editor/src/components/library/panes/media/local/karma/mediaFetching.karma.js
```diff
@@ -38,42 +37,34 @@ describe('MediaPane fetching', () => {
...
-      expect(initialElements.length).toBe(LOCAL_MEDIA_PER_PAGE);
-    });
+    expect(initialElementsLength).toBeGreaterThanOrEqual(LOCAL_MEDIA_PER_PAGE);
 
-    await mediaGallery.scrollTo(
-      0,
-      mediaGallery.scrollHeight - mediaGallery.clientHeight - ROOT_MARGIN / 2
-    );
+    // Scroll all the way down.
+    await mediaGallery.scrollTo(0, 9999999999999999999);
```
BrookeGraham: As @samwhale mentioned, the viewport for headless versus not headless is different sizes. So scrolling wasn't happening as expected. Now we are just scrolling all the way to the bottom.

samwhale (approved):
🥳

packages/story-editor/src/components/library/panes/media/common/paginatedMediaGallery.js
```diff
@@ -80,6 +80,17 @@ function PaginatedMediaGallery({
...
 
+    // Load the next page if the page isn't full, ie. scrollbar is not visible.
+    if (node.clientHeight === node.scrollHeight) {
+      setNextPage();
+      return;
+    }
+
+    // If scrollTop is zero, we know we do not need to fetch an additional page yet.
+    if (node.scrollTop === 0) {
+      return;
```
BrookeGraham: So here is the issue with this test being flakey. When this is first ran `node` gets defined, but sometimes the media elements are still being fetched resulting in the `scrollTop`  being set at `0`. ie. The container that is the node _is_ there, but it doesn't have any contents yet. The early return on line 90 `if node.scrollTop === 0` would solve the issue, _but_ the early return above on line 84 prevents it. This is because when the `node.ScrollTop` is `0` the `node.clientHieght === node.scrollHieght` is true. Therefore we call `setNextPage` in order to fill the container space with media elements. This means we make two fetch media calls when really we only want to make one. I think in the before times when the api was fetching 20 items, it was possible that 20 item may not have entirely filled the container's height. Which would totally make sense that we would want to make an additional call. We always want to fill that space. However, now that the api returns 50 items, it is very unlikely the container's height would not be full once the first batch of items has been called. In fact, I had to turn my external monitor vertical and  set resolution to super tiny in order to see that the second call would be needed. And the site as a whole is unusable with it that small. So I propose that we remove the check which makes the second call on line 90 altogether. Or I somehow wait for the elements to actually be present and from there check to see if we still need a second to fill the remaining space. @BrittanyIRL thoughts?

BrittanyIRL: thank you for such an insightful summary and for the time you've spent getting to the bottom of this. I'm fine removing that check, but I want to run this past Miina since her team has more experience and context for the media panels. @miina does the above make sense to ya?

miina: SGTM, however, pinging @swissspidy here as well since Infra pod has done a lot of work with everything related to Media.

packages/story-editor/src/components/library/panes/media/local/karma/mediaFetching.karma.js
swissspidy: Were was this mentioned / is this difference documented somewhere?

packages/story-editor/src/components/library/panes/media/common/paginatedMediaGallery.js
swissspidy: Unfortunately I don't have more info here as this logic here was implemented by our temp media pod last year

packages/story-editor/src/components/library/panes/media/local/karma/mediaFetching.karma.js
samwhale: > Were was this mentioned / is this difference documented somewhere?

This hasn't been mentioned or documented anywhere other than right here. This was just discovered, so it would be good to add a comment about this 👍

samwhale:
Seems like that fixed it? Other tests are failing that don't seem to be related 🤔

BrittanyIRL:
Looks like it's grumpy about the initial fetch 

https://github.com/google/web-stories-wp/runs/4400521926?check_suite_focus=true#step:6:15497
Screen Shot 2021-12-03 at 9 01 45 AM

samwhale:
> Looks like it's grumpy about the initial fetch

Oh maybe I was looking at the wrong tests

BrookeGraham:
So I just need to add a sleep for the debounce on the local test, but now I am running into the media3p fetching test failing.

samwhale:
> So I just need to add a sleep for the debounce on the local test, but now I am running into the media3p fetching test failing.

Weird - they shouldn't be dependent on each other 😬

packages/story-editor/src/components/library/panes/media/common/paginatedMediaGallery.js
```diff
@@ -88,13 +109,14 @@ function PaginatedMediaGallery({
       setNextPage();
       return;
     }
-
-    // Load the next page if the page isn't full, ie. scrollbar is not visible.
-    if (node.clientHeight === node.scrollHeight) {
-      setNextPage();
-      return;
-    }
-  }, [resources, hasMore, isMediaLoaded, isMediaLoading, setNextPage]);
```
BrookeGraham, line 97: @samwhale  this is what broke the media3p test. It wasn't that the test are dependent on one another. Because in ci the screen is set to 1600:1000 meaning that the media3p video tab needs to make 2 api calls to get enough items to fill the container. 3pMedia is set to only gran 20 items, not 50 like local is. Since I ripped this out, it broke your test because with a screen that size, the second api call wasn't ever made. Additionally that meant that you couldn't ever scroll down to get more items. I fixed this by adding the debounced callback and then checking again if we actually still need the second api call.

BrittanyIRL (approved):
✅ ✔️ ☑️ ✅ ✔️ ☑️ ✅ ✔️ ☑️ ✅ ✔️ ☑️ 
It's working!!! 
✅ ✔️ ☑️ ✅ ✔️ ☑️ ✅ ✔️ ☑️ ✅ ✔️ ☑️

packages/story-editor/src/components/library/panes/media/common/paginatedMediaGallery.js
samwhale: OOF my bad. Great find - that is super duper tricky. I owe you one 😓

BrookeGraham: If that comment hadn't of been left I would have glossed right over it. 🙌🏽 Go team.

Issue #9739, opened 2021-11-18
Media: Generate base color on upload


spacedmonkey:
## Feature Description

Follow up from https://github.com/google/web-stories-wp/pull/9638. 

Generate base color for images / videos, when the asset is uploaded instead of waiting for the backfill.

swissspidy:
Adding labels because it would be good to know how long generation takes for local files

PR #9785, opened 2021-11-23
Media: Do not store transient data on resource objects

### LINKED ISSUE #9577: Media: Do not store transient data on resource objects

spacedmonkey:
## Summary
Remove the `local`, `isMuting`, `isTrimming` and `isTranscoding` properties from the resource object stored in elements. This means, a large refactor, meaning all the references to this properties, in code and in tests. Use the media queue hook to work out if an resource is transcoding or local. 

There are two kinds of resources lookups required. By ID or by original id. For muting/transcoding/trimming resources, there are cases where need to lookup by id ( like in the media library ) and cases where we need to lookup state by the original id. This is because, while a resource is processing, it doesn't have an id yet, it uses a temp id. So we have to lookup via the old resources id. 

I have fixed all tests and added a couple of new ones for the new functionality found in the reducer. 

## Testing Instructions

This change effects all parts of uploading assets, ALL upload interactions will need to be tests. Including. 

- Upload of Image, video and gifs. 
- Muting of videos. 
- Trimming of videos and trimming a trimmed video. 
- Transcoding videos. 
- Uploading via the WP media dialog. 
- Dragging into canvas. 
- Pasting image into canvas
- Dragging animated gif onto canvas and convert to mp4. 

All video uploads, should include a time stamp, is muted icon and poster image. 

### Upload of Image.
1. Open editor. 
2. Drag a png/jpeg file into canvas / media library. 
3. See progress bar appear on new media item added to library and save button disabled.
4. Progress bar finishes. 
5. Save button re-enabled 

### Upload of Video.
1. Open editor. 
2. Drag a mp4/webm file into canvas / media library. 
3. See progress bar appear on new media item added to library and save button disabled. You should be able to hover over save button / new item in media library, saying that transcoding is taking place.  Trimming / muting options should be hidden. 
4. Progress bar finishes. 
5. Save button re-enabled / Trimming / muting buttons should appear

### Upload of animated gif.
1. Open editor. 
2. Drag a animated fig file into canvas / media library. 
3. See progress bar appear on new media item added to library and save button disabled. You should be able to hover over save button / new item in media library, saying that transcoding is taking place. 
4. Progress bar finishes. 
5. Save button re-enabled 

### Muting a video. 
1. Open editor.
2. Insert a video ( that has sound ) into canvas. 
3. In documents sidebar, under video options click mute video. 
4. See progress bar appear on new media item added to library and save button disabled. You should be able to hover over save button / new item in media library, saying that transcoding is taking place. 
5. Progress bar finishes. 
6. Save button re-enabled 

### Trimming video
1. Open editor.
2. Insert a video into canvas. 
3. In documents sidebar, under video options click trim. 
4. Trimming video to 10 seconds. 
5. Click trim to start trimming process. 
6. See progress bar appear on new media item added to library and save button disabled. You should be able to hover over save button / new item in media library, saying that transcoding is taking place. 
7. Progress bar finishes. 
8. Save button re-enabled. 

### Retrimming video
1. Follow steps in trimming video. 
2. On video video, retrim video, following steps in trimming video.

### Pasting file
1. Copy jpg/mp4/gif into clip from desktop / folder. 
2. Open editor. 
3. Paste file into canvas. 
4. See progress bar appear on new media item added to library and save button disabled.
5. Progress bar finishes. 
6. Save button re-enabled 

### Upload via the WP media dialog. 
1. Open editor. 
2. Click on upload button. 
3. See WP media dialog open. 
4. Upload jpg/animated gif/video into WP media dialog. 
5. Insert new asset into page. 
6. See dialog close. 
7. See new asset in media library and element in canvas. 

### Transcode existing video. 
1. Go to Media > Upload
2. Upload 1080p / 4K mp4 video. 
3. Open editor. 
4. Insert 1080p / 4K mp4 video into canvas.
5. Add title to story. 
6. Publish story. 
7. Open checklist. 
8. See option to opertize video. 
9. Click opertize video. 
10. See progress bar appear on new media item added to library and save button disabled. You should be able to hover over save button / new item in media library, saying that transcoding is taking place. 
11. Progress bar finishes. 
12. Save button re-enabled. 

## Checklist

---

Fixes #9577


packages/migration/src/migrate.js
```diff
@@ -54,6 +54,7 @@ import removeTitleFromResources from './migrations/v0033_removeTitleFromResource
 import removeUnusedBackgroundProps from './migrations/v0034_removeUnusedBackgroundProps';
 import markVideoAsExternal from './migrations/v0035_markVideoAsExternal';
 import changeBaseColorToHex from './migrations/v0036_changeBaseColorToHex';
+import removeMediaProperties from './migrations/v0037_removeMediaProperties';
```
swissspidy: Nit: let's rename to `removeTransientMediaProperties` to make this more descriptive

packages/migration/src/migrations/v0037_removeMediaProperties.js
```diff
@@ -0,0 +1,55 @@
...
+    delete element.resource.isTrimming;
+  }
+
+  if ('isTranscoding' in element.resource) {
+    delete element.resource.isTranscoding;
+  }
+
+  if ('isMuting' in element.resource) {
+    delete element.resource.isMuting;
+  }
```
swissspidy: Nit: there's no need for the `if`s here. `delete` does not do anything if the property doesn't exist.

packages/story-editor/src/app/media/local/useContextValueProvider.js
```diff
@@ -210,19 +223,10 @@ export default function useContextValueProvider(reducerState, reducerActions) {
...
-        id,
-        posterId,
-        mimeType,
-        poster,
-      } = resource;
+      const { type, isMuted, baseColor, src, id, posterId, mimeType, poster } =
+        resource;
 
-      if (local || !id) {
+      if (!id || isResourceUploadingById(id)) {
```
swissspidy: Do you recall why we have these `id` checks?

Seems like `isResourceUploadingById` ought to catch all relevant cases here, so the `id` check seems superfluous.

I'd like to get to a point where all resources always have an `id`.

packages/story-editor/src/components/canvas/useInsertElement.js
```diff
@@ -101,13 +117,16 @@ function useInsertElement() {
    */
   const handleRegisterUsage = useCallback(
     (resource) => {
-      if (!resource.local && resource?.attribution?.registerUsageUrl) {
+      if (
+        !isResourceUploadingById(resource.id) &&
+        resource?.attribution?.registerUsageUrl
```
swissspidy: I _think_ we can remove the `isResourceUploadingById` check here.

Only 3P media have a `registerUsageUrl` and those can't be uploaded. So the `isResourceUploadingById()` would always return `false` anyway

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
```diff
@@ -125,6 +125,10 @@ export function finishUploading(state, { payload: { id, resource } }) {
     revokeBlob(queueItem.resource.poster);
   }
 
+  if ('originalResourceId' in resource) {
+    delete resource.originalResourceId;
+  }
```
swissspidy: Nit: no need for the `if`.

`delete` won't do anything if the property doesn't exist.

packages/story-editor/src/app/media/local/useContextValueProvider.js
```diff
@@ -261,7 +276,7 @@ export default function useContextValueProvider(reducerState, reducerActions) {
   // generate missing posters / has audio / base color if needed.
   useEffect(() => {
     media?.forEach((mediaElement) => postProcessingResource(mediaElement));
-  }, [media, mediaType, searchTerm, postProcessingResource]);
+  }, [media, mediaType, searchTerm, postProcessingResource, uploaded]);
```
spacedmonkey: `uploaded` is not needed, as the media library doesn't upload on final upload, we need something to trigger reprocessing on final upload.

packages/story-editor/src/components/header/buttons/preview.js
```diff
@@ -49,8 +49,8 @@ function PreviewButton({ forceIsSaving = false }) {
       actions: { autoSave, saveStory },
     }) => ({ isSaving, status, previewLink, autoSave, saveStory })
   );
-  const { isUploading } = useLocalMedia((state) => ({
-    isUploading: state.state.isUploading,
+  const { isUploading } = useLocalMedia(({ state: { isUploading } }) => ({
```
spacedmonkey, line 52: I went through an updated all references like this, as I wanted to be sure all the `useLocalMedia` references were still working. I can revert these changes, but I see it as an improvement.

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
```diff
@@ -109,9 +115,19 @@ function useVideoTrimMode() {
...
     }
-    const { local, isExternal } = selectedElement.resource || {};
-    return isVideoTrimEnabled && isTranscodingEnabled && !isExternal && !local;
-  }, [selectedElement, isVideoTrimEnabled, isTranscodingEnabled]);
+    const { id, isExternal } = selectedElement.resource || {};
+    return (
+      isVideoTrimEnabled &&
+      isTranscodingEnabled &&
+      !isExternal &&
+      !(id && isResourceUploadingById(id))
```
swissspidy: Related to my other question: When would there be no `id`?

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
```diff
@@ -109,9 +115,19 @@ function useVideoTrimMode() {
     if (selectedElement?.type !== 'video' || !selectedElement?.resource) {
       return false;
     }
-    const { local, isExternal } = selectedElement.resource || {};
-    return isVideoTrimEnabled && isTranscodingEnabled && !isExternal && !local;
-  }, [selectedElement, isVideoTrimEnabled, isTranscodingEnabled]);
+    const { id, isExternal } = selectedElement.resource || {};
```
swissspidy:
```suggestion
    const { id, isExternal } = selectedElement.resource;
```

There's already a `!selectedElement?.resource` check above, so at this point here we know that it exists

packages/story-editor/src/components/checklist/checks/videoOptimization.js
```diff
@@ -51,24 +51,27 @@ const OptimizeButton = styled(Button)`
   margin-top: 4px;
 `;
 
-export function videoElementsNotOptimized(element = {}) {
-  if (!element.resource) {
+export function videoElementsNotOptimized(
```
swissspidy: Instead of having to pass `isResourceTranscoding` and `isResourceProcessing` arguments, we can make this easier to use by turning this into a custom hook.

Something like this (untested):

```js
export function useVideoElementsNotOptimized() {
  const { isResourceTranscoding, isResourceProcessing } =
    useLocalMedia(({ actions, state: mediaState }) => ({
      isResourceTranscoding: mediaState.isResourceTranscoding,
      isResourceProcessing: mediaState.isResourceProcessing,
    }));

  return useCallback((element = {}) => {
    // ...
  }, [isResourceTranscoding, isResourceProcessing]);
}
```

packages/story-editor/src/app/rightClickMenu/provider.js
```diff
@@ -775,6 +782,18 @@ function RightClickMenuProvider({ children }) {
     ]
   );
 
+  const canTranscodeResource = useCallback(
+    (resource) => {
+      const { id, isExternal } = resource || {};
```
swissspidy:
```suggestion
      const { id, isExternal } = resource;
```

Where would `resource` be `null`?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/test/useMediaUploadQueue.js
```diff
@@ -97,6 +109,15 @@ describe('useMediaUploadQueue', () => {
...
         isTrimming: false,
+        isResourceMuting,
+        isResourceMutingById,
+        isResourceProcessing,
+        isResourceProcessingById,
+        isResourceTranscoding,
+        isResourceTranscodingById,
+        isResourceTrimming,
+        isResourceTrimmingById,
+        isResourceUploadingById,
```
swissspidy: It's a bit pointless to get functions from `result.current.state` and then compare them with `result.current.state`...

I'd rather use `expect.any(Function)` here.

```suggestion
        isResourceMuting: expect.any(Function),
        isResourceMutingById: expect.any(Function),
        isResourceProcessing: expect.any(Function),
        isResourceProcessingById: expect.any(Function),
        isResourceTranscoding: expect.any(Function),
        isResourceTranscodingById: expect.any(Function),
        isResourceTrimming: expect.any(Function),
        isResourceTrimmingById: expect.any(Function),
        isResourceUploadingById: expect.any(Function),
```

Same for the test below

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -386,6 +386,72 @@ function useMediaUploadQueue() {
...
+            (item) =>
+              [
+                'PENDING',
+                'UPLOADING',
+                'TRANSCODED',
+                'TRANSCODING',
+                'MUTING',
+                'MUTED',
+                'TRIMMING',
+                'TRIMMED',
```
swissspidy: Related:

Since we use these states in a lot of places, in order to avoid typos we could create a simple const in `packages/story-editor/src/app/media/utils/useMediaUploadQueue/constants.js`. Something like this:

```js
const ITEM_STATUS = {
  PENDING: 'PENDING',
  UPLOADING: 'UPLOADING',
  TRANSCODED: 'TRANSCODED',
  TRANSCODING: 'TRANSCODING',
  MUTING: 'MUTING',
  MUTED: 'MUTED',
  TRIMMING: 'TRIMMING',
  TRIMMED: 'TRIMMED',
};
```

Then everywhere we reference a status we'd use `ITEM_STATUS.MUTING`, etc.

packages/story-editor/src/app/media/local/useContextValueProvider.js
swissspidy: Can you elaborate on this? Not sure I follow 100%. What was the issue, why is it needed now, how can this be reproduced?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
```diff
@@ -62,6 +62,7 @@ export function addItem(
     state: 'PENDING',
     resource: {
       ...resource,
+      originalResourceId: resource?.id,
```
swissspidy:
```suggestion
      originalResourceId: resource.id,
```
No need for the optional chaining. `resource` is never `null` here.

If `resource` would be `null`, `...resource` above would throw an error.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -386,6 +386,72 @@ function useMediaUploadQueue() {
         isTranscoding: state.queue.some((item) => item.state === 'TRANSCODING'),
         isMuting: state.queue.some((item) => item.state === 'MUTING'),
         isTrimming: state.queue.some((item) => item.state === 'TRIMMING'),
+        isResourceProcessing: (resourceId) =>
```
swissspidy: Let's add tests for all these new functions in `packages/story-editor/src/app/media/utils/useMediaUploadQueue/test/useMediaUploadQueue.js`

swissspidy: Some inline documention / JSDoc blocks for these new functions would go a long way to help explain the differences.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -386,6 +386,72 @@ function useMediaUploadQueue() {
...
+                'TRANSCODED',
+                'TRANSCODING',
+                'MUTING',
+                'MUTED',
+                'TRIMMING',
+                'TRIMMED',
+              ].includes(item.state) &&
+              item.resource.originalResourceId === resourceId
+          ),
+        isResourceProcessingById: (resourceId) =>
```
swissspidy: Since we talked about improving naming of these functions (because the `ById` suffix is quite confusing), some ideas off the top of my head:

* Instead of `isResourceProcessingById`, use `isCurrentResourceProcessing`

(same would go for the other functions)

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -386,6 +386,72 @@ function useMediaUploadQueue() {
...
+        isResourceMutingById: (resourceId) =>
+          state.queue.some(
+            (item) => item.state === 'MUTING' && item.resource.id === resourceId
+          ),
+        isResourceTrimmingById: (resourceId) =>
+          state.queue.some(
+            (item) =>
+              item.state === 'TRIMMING' && item.resource.id === resourceId
+          ),
+        isResourceTranscoding: (resourceId) =>
```
swissspidy: Nit: could we move those functions together, i.e. move this below/above `isResourceTranscodingById`? Helps with readability

packages/story-editor/src/app/media/local/useContextValueProvider.js
spacedmonkey: The issue was, that now the media resources is not longer updated, with isTrimming etc, this effect was not firing when uploads were complete. State is not being updated as much, so this processing is not being trigger. Using this bit of state, that changes when an upload finishes, it trigger the backfill on upload finishes, which is what we expect to happen. If you remove this, then with it doesn't backfill when new files are uploaded, or backfills at the wrong time. Either way, it works fine.

packages/story-editor/src/app/media/local/useContextValueProvider.js
spacedmonkey: Id is a required field, as without an id ( like hotlinked images ) then most of the code below would fail.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
spacedmonkey: This is needed, unit test failed without it.

packages/migration/src/migrations/v0037_removeMediaProperties.js
spacedmonkey: We have lots of other examples where we do a check.  Example. 

https://github.com/google/web-stories-wp/blob/1dda476acc4157c399b7e6a5a483ac3a394d9a8a/packages/migration/src/migrations/v0029_unifyGifResources.js#L38-L53

Why do this here but not other places.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
spacedmonkey: https://github.com/google/web-stories-wp/pull/9785#discussion_r757629577

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
spacedmonkey: Hotlinked image.

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
spacedmonkey: There was a unit tests failure because of this line. Which is why it failed. Nothing wrong with being careful.

packages/story-editor/src/components/checklist/checks/videoOptimization.js
spacedmonkey: It would mean refactor all the tests.

packages/story-editor/src/app/rightClickMenu/provider.js
spacedmonkey: It is called with a condiontal, `element?.resource`, meaning it could be nulll

packages/migration/src/migrations/v0037_removeMediaProperties.js
swissspidy: We should remove this check everywhere then.

packages/story-editor/src/app/media/local/useContextValueProvider.js
swissspidy: I know that `id` is required.

My point is: all resources are always guaranteed to have an ID. (and if not, we should fix that, as mentioned on the ticket #9577)

Because of that, there is no need for these `id` checks everywhere.

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
swissspidy: Hotlinked images _do_ have an ID though.

https://github.com/google/web-stories-wp/blob/f22db1e7ddc307a07fabc020c536eb6f8d8a8904/packages/story-editor/src/components/library/panes/media/local/hotlink/useInsert.js#L98-L104

packages/story-editor/src/app/media/local/useContextValueProvider.js
swissspidy: And yes, hotlinked images _do_ have an ID:

https://github.com/google/web-stories-wp/blob/f22db1e7ddc307a07fabc020c536eb6f8d8a8904/packages/story-editor/src/components/library/panes/media/local/hotlink/useInsert.js#L98-L104

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
swissspidy: I tend to disagree, one could be overly careful.

Can you point me to the unit test?

packages/story-editor/src/app/rightClickMenu/provider.js
swissspidy: That optional chaining usage is wrong though, there's no need for it.

Check the code in question:

`canTranscodeResource` is only ever called when `isVideo` is true, which is only the case when `selectedElement` is not `null`.

https://github.com/google/web-stories-wp/blob/ca85940900e608639796e5d9833017109a1b409e/packages/story-editor/src/app/rightClickMenu/provider.js#L953-L962

That means we can remove that `?` from that call, and remove the `resource || {}`.

(Aside, instead of `(resource) => { const { id, isExternal } = resource || {};` one could just do `(resource ={}) => {
      const { id, isExternal } = resource;` anyway)

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
swissspidy: Can you point to the relevant unit test?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
miina: Just wanted to ask about the naming here -- since `isResourceProcessing` and `isResourceProcessingById` both seem to be used by ID.

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
```diff
@@ -88,24 +87,52 @@ const HelperText = styled(Text).attrs({
...
+    resource.isExternal;
 
   const shouldDisplayMuteButton =
     isTranscodingEnabled &&
     isSingleElement &&
-    ((!isMuted && canTranscodeResource(resource)) || isMuting);
+    ((!isMuted &&
+      !isResourceProcessing(resourceId) &&
+      !isResourceProcessingById(resourceId)) ||
+      isMuting);
```
miina: Q: Doesn't `isResourceProcessing` already include `isMuting` as well? Meaning that doesn't `!isResourceProcessing` already mean it's not muting?

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
```diff
@@ -88,24 +87,52 @@ const HelperText = styled(Text).attrs({
...
   const isSingleElement = selectedElements.length === 1;
 
   const handleMuteVideo = useCallback(() => {
     muteExistingVideo({ resource });
   }, [resource, muteExistingVideo]);
 
-  const shouldDisableVideoActions = !canTranscodeResource(resource);
+  const shouldDisableVideoActions =
+    isResourceProcessing(resourceId) ||
+    isResourceProcessingById(resourceId) ||
```
miina: Isn't there a way to just check `isResourceProcessing` once? It seems a bit add that it needs to be checked twice by the same resource ID. What's the difference there?

packages/story-editor/src/components/library/test/mediaElement.js
```diff
@@ -187,6 +192,14 @@ describe('MediaElement', () => {
   });
 
   it("should not render dropdown menu's more icon for not uploaded image", () => {
+    useLocalMedia.mockReturnValue({
```
miina, line 218: It seems that this is already done in `beforeEach(() => {` above, wouldn't it work for all cases?

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
```diff
@@ -267,9 +269,30 @@ Element.propTypes = {
...
+      isResourceMutingById,
+      isResourceTrimmingById,
+      isResourceTranscodingById,
+    })
+  );
+  const { id } = props.resource;
 
-  if (isTranscoding || isMuting || isTrimming) {
+  if (
+    isResourceTrimmingById(id) ||
```
miina: Q: Wouldn't `isResourceProcessing` already check for all these? Although, I'm also seeing `MUTED` in the list of the statuses for `isResourceProcessing`. I probably don't have enough context here right away, but is media still processing if it's already `muted`? I thought it would only be processing if it's any of the statuses with `-ING` ending.

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
swissspidy: This has to do with naming; a reason why we need to change it.

Currently, the `byId` function means "check if this current media item is being processed", where es the one without the suffix means "check if a new version of this resource is being processed".

When muting an existing video in a story, you are at the same time "processing" that existing video, but also uploading a new one (with its own unique ID), which is being displayed in the media library.

Now, we want two be able to do two things:

1. show a progress bar for the newly uploading muted video in the library
2. Show the "Muting..." button for the old, existing video in the story

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
swissspidy: Has to do with naming, see https://github.com/google/web-stories-wp/pull/9785#discussion_r758331003 and other thread at https://github.com/google/web-stories-wp/pull/9785#discussion_r758309753

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: https://github.com/google/web-stories-wp/pull/9785#discussion_r758331003

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
miina: Thanks for the explanation, makes sense now. The naming should be changed indeed :)

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
spacedmonkey: It is still processing, as after is muted, it still needs to be uploaded.

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
spacedmonkey: One is for if the current resource is being processed and the other new resource that is linked original resource.

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
miina: Okay, but wouldn't the status then be `UPLOADING`?

spacedmonkey: In this case we don't want tooltip to display.

miina: I'm just generally trying to understand why are `MUTED`, `TRIMMED` etc. statuses necessary. If after `TRIMMING` we need to upload and change to `UPLOADING`, why do we still need separately `MUTED`, `TRIMMED`, etc.?

packages/story-editor/src/app/rightClickMenu/provider.js
spacedmonkey: After changing this, I get this error. 

should contain enabled "trim video"
Screenshot 2021-11-29 at 14 18 33

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
spacedmonkey: Seems I cant replicate.

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
swissspidy: > why do we still need separately `MUTED`, `TRIMMED`, etc.?

That is a great question. You are right, we probably don't need `MUTED`, `TRIMMED` etc., since we immediately go over to `UPLOADING` after that by calling `startUploading`:

https://github.com/google/web-stories-wp/blob/4a614ce56910ca9efb00bcd1eb49c4406a37b456/packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js#L316

Reducers like `finishMuting` just update `isMuted` on the current resource, which is what we mostly care about. So theoretically we would not need to update `state` here:

https://github.com/google/web-stories-wp/blob/4a614ce56910ca9efb00bcd1eb49c4406a37b456/packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js#L261-L272

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: @miina With the additional context, do you have any suggestions for improved naming here?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
spacedmonkey: Fixed.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
miina: Maybe `isResourceProcessingById` => `isOriginResourceProcessing`?

If `isResourceProcessing` is only used for the new resource, then maybe actually changing that to `isNewResourceProcessing` or `isUpdatedResourceProcessing` or `isGeneratedResourceProcessing` could also make things more clear. If it's used in some other cases as well then this wouldn't work.

packages/story-editor/src/components/canvas/useInsertElement.js
spacedmonkey: Fixed.

packages/story-editor/src/components/library/test/mediaElement.js
spacedmonkey: No, I am changing the value of the mocks to return true. I made a change that makes this cleaner. 
https://github.com/google/web-stories-wp/pull/9785/commits/833a39d9135b1152f6ddea3c3b7f712aecb08e67

packages/story-editor/src/app/rightClickMenu/provider.js
spacedmonkey: Also you are wrong here. 
> (resource) => { const { id, isExternal } = resource || {}; one could just do (resource ={}) => { const { id, isExternal } = resource;

There are the same. 

`(resource ={})` sets a default value. If null is passed, the default value will not be set. 
```
const { id, isExternal } = resource || {};
```
Ensure that if resource is null, it falls back to an empty object.

packages/story-editor/src/app/media/local/useContextValueProvider.js
swissspidy: Okay, thanks for the explanation.

Follow-up question:if we implement #9739, backfilling for newly uploaded items would theoretically not be needed, right?

spacedmonkey: > Follow-up question:if we implement #9739, backfilling for newly uploaded items would theoretically not be needed, right?

No, that isn't the case, images uploaded via wp media library always need backfill. The backfill is done to be sure that the many ways we upload assets, that that are always backfilled.

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
miina: Alright, I guess at least it's not necessary to include those statuses into `isResourceProcessing` and `isResourceProcessingById` (per current naming)? Since these states will be changed right away, hence not really used + during that tiny moment when those states are assigned, nothing is processing.
It would make the code a bit more readable / understandable as well.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/test/useMediaUploadQueue.js
```diff
@@ -18,26 +18,41 @@
...
 import useMediaUploadQueue from '..';
 import useFFmpeg from '../../useFFmpeg';
 
+const canTranscodeFile = (file) => {
+  return ['video/mp4'].includes(file.type);
+};
+
+const waitfor100 = async () => {
+  await new Promise((res) => setTimeout(res, 100));
+};
```
swissspidy: Why artificially make the tests slower with some sleep? 🤔 

What's it that this achieves? Is there a way to test this without this sleep?

packages/story-editor/src/app/media/local/useContextValueProvider.js
swissspidy: Images uploaded via the WP media library change `uploaded` then?

spacedmonkey: Media is also changed in that case. But we should look into this in a follow up ticket.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/test/useMediaUploadQueue.js
spacedmonkey: It replicates the real functions and replicate the real functions. Also in my testing, without this slow down, the call to `waitForNextUpdate` would skip to UPLOADED, skipping state change like isTranscoding.

packages/story-editor/src/app/rightClickMenu/provider.js
swissspidy: > After changing this, I get this error.

That's odd 🤔

packages/story-editor/src/components/library/panes/media/local/mediaPane.js
```diff
@@ -209,7 +212,7 @@ function MediaPane(props) {
    */
   const onSelect = (resource) => {
     try {
-      if (isTranscodingEnabled && canTranscodeResource(resource)) {
+      if (isTranscodingEnabled && canTranscodeResource(resource.id)) {
```
spacedmonkey:
```suggestion
      if (isTranscodingEnabled && canTranscodeResource(resource)) {
```

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
spacedmonkey: So should I removed muted, trimmed and transcoded states then?

packages/story-editor/src/app/media/local/useContextValueProvider.js
spacedmonkey: Removed this now, it is not required

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
spacedmonkey: Fixed.

packages/story-editor/src/components/checklist/checks/videoOptimization.js
spacedmonkey: Should I make this into a hook then?

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
miina: Yes, looks like those states are not useful for `isResourceProcessing` and `isResourceProcessingById`.

swissspidy: Yes; would be great if you could give that a shot.

packages/story-editor/src/components/checklist/checks/videoOptimization.js
swissspidy: I suppose it's alright for now; something to improve in the future

packages/story-editor/src/app/media/utils/useMediaUploadQueue/test/useMediaUploadQueue.js
swissspidy: Does it work with a slower timeout, e.g. 50ms?

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
swissspidy: 5c375b2038251c049903fd16208ffcd45f0444aa is not enough

My understanding is that we want to remove lines like this as well:

https://github.com/google/web-stories-wp/blob/23092d2987419a6c61c8c52e794cd08b79f7d3dd/packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js#L267

And of course remove `MUTED` from `ITEM_STATUS`.

Completely getting rid of `ITEM_STATUS.MUTED` et al.

spacedmonkey: Once a mute / transcode / trim is complete, should the state return to pending?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/test/useMediaUploadQueue.js
spacedmonkey: It does. I changed it to 10ms, which seems to be min speed I can push it.

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
swissspidy: No, `PENDING` means a file pending any processing/upload whatsoever; a file at the very end of the queue.

We don't need to change the state at all. The item can stay in its current state as it will be changed to `UPLOADING` right away on this line:

https://github.com/google/web-stories-wp/blob/4a614ce56910ca9efb00bcd1eb49c4406a37b456/packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js#L316

spacedmonkey: We do needed transcoded state, as it used in current implementation. I think we should leave this for now and remove this in a follow up PR.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -87,7 +89,11 @@ function useMediaUploadQueue() {
         state.queue.map(async (item) => {
           const { id, file, state: itemState, resource } = item;
           if (
-            itemState !== 'TRANSCODED' ||
+            ![
+              ITEM_STATUS.TRIMMED,
+              ITEM_STATUS.MUTED,
+              ITEM_STATUS.TRANSCODED,
+            ].includes(itemState) ||
```
spacedmonkey, line 113: I am going to leave this as it is for now and follow up with another ticket to remove this as needed.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -336,15 +342,20 @@ function useMediaUploadQueue() {
...
               newPosterFile,
               resource,
               id,
             });
+            return;
           }
+          finishUploading({
+            id,
+            resource: newResource,
+          });
```
spacedmonkey, line 441: This improved performance in testing. This is because if no poster is genrated, then just don't try to upload one. I can do this another PR if required.

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -195,11 +190,12 @@ function useProcessMedia({
       const onUploadStart = () =>
         updateExistingElements(resourceId, {
           trimData,
-          isTrimming: true,
         });
 
       const onUploadError = () =>
-        updateExistingElements(resourceId, { isTrimming: false });
+        updateExistingElements(resourceId, {
+          trimData: oldResource?.trimData || {},
```
spacedmonkey: This should always have been this way.

packages/story-editor/src/components/checklist/checks/videoOptimization.js
```diff
@@ -52,24 +52,13 @@ const OptimizeButton = styled(Button)`
 `;
 
 export function videoElementsNotOptimized(element = {}) {
```
spacedmonkey, line 54: Refactored this logic to not worry about state at all. All we care about with this function now, is it `isOptimized` and is the correct size.

packages/story-editor/src/components/library/panes/media/common/innerElement.js
```diff
@@ -117,7 +117,7 @@ function InnerElement({
 
   useEffect(() => {
     // assign display poster for videos
-    if (resource.poster && resource.poster.includes('blob')) {
+    if (resource.poster) {
```
spacedmonkey, line 145: This fixes an issue with trimmed videos not showing poster images, even through one is set. Not sure why is this is happening, this it might be related to less state changes means less renders. With way, this works.

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
```diff
@@ -224,7 +227,9 @@ function Element({
           active={active}
         />
         {attribution}
-        {(local || isTranscoding || isMuting || isTrimming) && (
+        {(!src ||
+          isBlobURL(src) ||
```
spacedmonkey: When a empty placeholder is added to state, I wasn't see the progress bar. Instead of local, we need to check for both empty and blob url states.

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
swissspidy: @spacedmonkey And where is it used? But then you'd need to undo 5c375b2038251c049903fd16208ffcd45f0444aa, no? 🤔

packages/story-editor/src/app/media/utils/useProcessMedia.js
swissspidy: `oldResource` is not `nulll` though and we don't use optional chaining for it in other places here.

```suggestion
          trimData: oldResource.trimData || {},
```

packages/story-editor/src/components/checklist/checks/videoOptimization.js
swissspidy: IIRC, this function used to do `if (isTranscoding) { return true; }`, so that videos which are currently being transcoded are still shown in the checklist. Is that still the case?

miina:
@spacedmonkey There's currently something off with the states while processing. E.g. try trimming or muting a video.

Previously, when trimming a video, the Trim button got disabled in the design panel and showed that it was trimming + showed the circular progress. Similar for Muting IIRC. Currently, visibly nothing seems to happen in the design panel, the trim mode just exits. (It can be seen in the Media panel that the video is processing, so the functionality is working in the background but there's no indication of it anymore in the design panel).

packages/story-editor/src/components/checklist/checks/videoOptimization.js
spacedmonkey: It is in my testing.

spacedmonkey:
> There's currently something off with the states while processing. E.g. try trimming or muting a video.

@miina Wow, great catch. The issue was around the resource id being wrong. It only happened when you trimmed and trimmed video. But it turns out the fix was super simple. https://github.com/google/web-stories-wp/pull/9785/commits/b7f7a95dd8db26ca5b0b5a99d159960fe9a982ef . That fixed the issue for me, can you retest.

packages/story-editor/src/components/checklist/checks/videoOptimization.js
spacedmonkey: https://user-images.githubusercontent.com/237508/144456628-82f6c3c5-b29a-47e1-a4c0-687eaa50b11e.mov

miina:
> That fixed the issue for me, can you retest.

Unfortunately, I'm still seeing some issues with re-trimming, specifically when re-trimming a second time.

When re-trimming a video for the second time, it should show the total length of the trimmer according to the original video, and the current selected length based on the new trimmed video. Currently, when re-trimming a video shorter, and then trying to re-trim, it shows the maximum length of the video right away (instead of the trimmed length).

Steps to reproduce:
1. Add a non-trimmed video.
2. Trim the video (e.g. from 15 seconds to 6 seconds) - works
3. Wait until trimming finishes and open the trimmer again - works
4. Now re-trim the video again (e.g. from 6 seconds to 2 seconds) - visibly seems to work
5. Open the trimmer again. The trimmer still shows 6 seconds as the length instead of 2 seconds. - incorrect value

So the initial trimming works. The first re-trimming works. And then trying to re-trim the second time, there's something not right.

Thoughts?

miina:
> Unfortunately, I'm still seeing some issues with re-trimming, specifically when re-trimming a second time.

@spacedmonkey I'm testing on `main` just in case and seeing the same issue there. I'll do some more testing but looks like it's not related to this PR.


Edit: Added an issue for it: https://github.com/google/web-stories-wp/issues/9878

packages/story-editor/src/components/library/panes/media/common/innerElement.js
miina: At which moment is a trimmed video not showing poster image on `main`? Tried out the following:
- Add a video and trim it
- Wait until trimming finishes
- Check the Accessibility panel -- the poster image is visible.

Or did you mean something else?

packages/story-editor/src/components/checklist/checks/videoOptimization.js
```diff
@@ -246,7 +273,10 @@ const BulkVideoOptimization = () => {
           <Thumbnail
             key={element.resource.id}
             onClick={handleClickThumbnail(element)}
-            isLoading={element.resource.isTranscoding}
+            isLoading={
+              isResourceTranscoding(element.resource.id) ||
```
miina: Q: Why not change it to `isNewResourceTranscoding` or something similar if it's always used for new resources only?

It's still a bit unclear to differentiate between `isResource...` and `isCurrentResource...`.

Alternatively, maybe let's make the function comments a bit more detailed? E.g.
`Is the current resource being processed.` -> it would be good to mention that this is the original resource that's used for generating the new resource. Thoughts?

Edit:

And maybe also mention in the comment for `isResourceProcessing` that a new resource means the new resource that gets uploaded and replaces the original resource after trimming / muting, etc.?

packages/migration/src/migrations/v0037_removeTransientMediaProperties.js
```diff
@@ -0,0 +1,44 @@
...
+    ...rest,
+  };
+}
+
+function updateElement(element) {
+  if (!element.resource) {
+    return element;
+  }
+
+  delete element.resource.local;
```
miina, line 36: @spacedmonkey General question to understand this better:
In which case is a resource not `local` and not `isExternal` either? Just wondering that if it's one or another, why keep track of both and not just `isExternal` which could be `true` or `false`. 

Also, not for this PR but should it be renamed to `isLocal` for consistency?

spacedmonkey: Local is being removed here. We can workout if a resource is local by checking if it is in the upload queue or not.

packages/story-editor/src/components/library/panes/media/common/innerElement.js
spacedmonkey: This is not a issue in main, just the PR. This is the only thing I could find to fix it. Without this change, poster is rendered by hidden for some reason.

spacedmonkey: https://user-images.githubusercontent.com/237508/144619452-4190f548-2f2f-474b-a8bb-4f5dbae9b881.mov

This video, show the behaviour I am trying to fix. The poster is there but not shown.

packages/story-editor/src/components/checklist/checks/videoOptimization.js
spacedmonkey: This change does make things more confusing for sure. But to untangle this, would mean writing lots of this code. Refactoring for readability with no part of the scope of this change. I would like I can add comments and change function names. Refactoring of this code to make is clearer, should be done in another follow up ticket.

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
spacedmonkey: @swissspidy trancoded state is used here https://github.com/google/web-stories-wp/blob/91d7271af6a7f17b42e0be4bbaef5332fecd7170/packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js#L89-L95

The relays on transcode to be complete before regenerating the poster image. 

I do agree that the trancoded, muted and trimmed states seem a little pointless, however, this PR is already massive. Can we please remove this in another PR?

packages/story-editor/src/components/checklist/checks/videoOptimization.js
miina: Updating just the comments sounds good to me for now as well!

swissspidy: I have some local changes in this regard ready; will create a PR for this in a bit.

swissspidy (approved):

miina (approved):

miina:
@spacedmonkey FYI this has some conflicts now.

swissspidy:
@spacedmonkey FYI there are some lint errors.

PR #9787, opened 2021-11-23
Fixed missing ref causing karma errors

barklund:
## Context

This fixed the canvas focus and re-enables relevant tests.

## Relevant Technical Choices

* Missing `forwardRef` on new component between layers and DOM element.

## User-facing changes

Auto-focus works again

## Testing Instructions

This PR can be tested by following these steps:

1. Add a new text element using the + button in the element library tab bar
1. Press <kbd>enter</kbd> to edit text
1. Observe the edit mode is working and text is select

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [x] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [x] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #9781

PR #9795, opened 2021-11-24
Media: Enable Hotlinking by default

### LINKED ISSUE #9466: Enable Media Hotlinking by default
## Feature Description
Once the Proxy related issues have been completed, we can enable the feature by default.

- [ ] Enable the feature (both Hotlinking and CORS Proxy)
- [x] Add Karma tests if/where needed

Note: more issues might need to be resolved before enabling the feature once it gets tested.

miina:
## Summary

Enables media hotlinking (and  the CORS proxy) feature by default.

## User-facing changes

Screenshot 2021-11-24 at 13 32 39

## Testing Instructions

1. Use the Hotlinking button in the Media tab (next to the Upload button) to insert different images and videos into the Page
2. For example try linking media from the site itself; from Unsplash; from wp.stories.google
3. Verify that a new media element is created to the page, and while the media is being inserted, the Insert button is disabled.
4. Verify that inserting incorrect URLs (such as faulty URLs, links that point to a file that's not image/video, etc.) display an error message and are not inserted.

## Checklist

---

Fixes #9466


miina:
As it didn't seem that blocking bugs came out of the bug-bash, enabling the features by default for sending to QA.

swissspidy (approved):

miina:
Hm, looks like there are some valid E2E failures:
`TimeoutError: Element button (text: "Upload") not found`

Also, for some reason `Upload ZIP files to storage` seems to fail (on another PR as well).

packages/e2e-test-utils/src/uploadMedia.js
```diff
@@ -27,7 +27,7 @@ import uploadFile from './uploadFile';
  */
 async function uploadMedia(file, exit = true) {
   // Clicking will only act on the first element.
-  await expect(page).toClick('button', { text: 'Upload' });
+  await expect(page).toClick('button[aria-label="Upload"]');
```
spacedmonkey, line 30: Is this related to this changed?

miina: Previously, there was a button with text `Upload`, now the button is an icon but still has a label `Upload`, so the selector had to change a bit. 
I just added this change since E2E tests were failing before.

swissspidy:
> As it didn't seem that blocking bugs came out of the bug-bash, enabling the features by default for sending to QA.

Since there was thorough QA on Monday and bug bash on Tuesday we might even skip that part.

> Hm, looks like there are some valid E2E failures: `TimeoutError: Element button (text: "Upload") not found`

Can you point to a test run?

> Also, for some reason `Upload ZIP files to storage` seems to fail (on another PR as well).

Can you point to a test run?

I did some work in that area yesterday.

miina:
> Can you point to a test run?

Test run on another PR for failing ZIP Upload: https://github.com/google/web-stories-wp/runs/4304527451?check_suite_focus=true

On E2E tests: The Upload button selectors have been modified in the PR (from `'button', { text: 'Upload' }` to `'button[aria-label="Upload"]'`, hopefully it works now.
Test run: https://github.com/google/web-stories-wp/runs/4310265391?check_suite_focus=true#step:11:35

miina:
@swissspidy Now uploading to storage failed here, too: https://github.com/google/web-stories-wp/runs/4310923768?check_suite_focus=true

(E2E tests are passing now after the selector update)

miina:
Most (if not all) of the Karma failures in this PR should get fixed by https://github.com/google/web-stories-wp/pull/9787

miina:
@swissspidy I suppose uploading the build to storage is not important if we don't send this PR to QA?

In this case it should be ready for merging.

swissspidy:
True :-)

In any case, I just fixed this on the `main` branch

spacedmonkey (approved):

PR #9822, opened 2021-11-26
History: Allow ignoring element properties

### LINKED ISSUE #9645: History: allow ignoring properties of elements
## Feature Description
There are some properties of elements that should not influence the History and should not be allowed to be undone/redone by users. 

For example, for performance reasons, it makes sense to calculate the average color of media (`baseColor`) once it has already been inserted and not delay inserting the media due to waiting for calculation.

We should consider making the history more flexible to allow this.

Note that if a history entry is not created and a change is not registered, the editor will also not indicate to the user that the story is unsaved. This means that potentially, the properties might "get lost" this way if changed and the user does not save the Story afterwards.

miina:
## Context
There are some properties that should not create new history entries, such as `baseColor`. Adding this happens in the background and can confuse the user and make the Undo/Redo functionality not to work as expected.

## Summary
This PR adds a possibility to define properties to ignore via adding the properties to a constant in a format of `foo.bar.a`. E.g. for the `baseColor` property inside `resource`, we can define it `resource.baseColor`.

If only the ignored properties change, a new history entry is simply not created.

## Relevant Technical Choices
The check if to skip history entry is done outside of the history reducer -- history itself should not be aware of when adding a new entry is necessary and when not.

Note: This issue currently adds the props to ignore as an empty array since the issue just adds the logic for allowing to do that. Props will be added within #10009 

## To-do

- [x] Tests
- [x] Verify approach
- [x] Remove test data / cleanup / refactor

## User-facing changes
N/A

## Testing Instructions

1.

No

No

No

## Checklist

---

Fixes #9645


packages/story-editor/src/app/story/effects/useHistoryEntry.js
```diff
@@ -23,28 +23,83 @@ import { useEffect, useRef } from '@web-stories-wp/react';
...
+        ...page,
+        elements: elements.map((element) =>
+          deleteNestedKeys(element, ELEMENT_PROPS_TO_IGNORE)
+        ),
+      };
     });
+  };
+
+  useEffect(() => {
+    // There are some element properties that should not influence history.
```
miina, line 63: @barklund WDYT of this approach? It exposes `currentEntry` to the "world" but at the same time leaves the reducer completely as it is -- since the reducer shouldn't be aware of the reasons for adding a history entry / not.

This basically allows defining any (nested or not) element properties that should not create a new history entry. Since these propeties should still be saved into history though, the change is only for the case when _only_ the ignored property changes.

Currently, for testing purposes, `rotationAngle` is ignored. 

LMK if you have any thoughts on this approach.

packages/story-editor/src/app/story/effects/useHistoryEntry.js
```diff
@@ -23,28 +23,83 @@ import { useEffect, useRef } from '@web-stories-wp/react';
  * Internal dependencies
  */
 import { useHistory } from '../../history';
+import deleteNestedKeys from '../utils/deleteNestedKeys';
+
+// @todo REMOVE ROTATION ANGLE! For testing only!
+const ELEMENT_PROPS_TO_IGNORE = ['resource.baseColor', 'rotationAngle'];
```
spacedmonkey: I would also add `resource.isMuted`, `resource.poster` and `resource.posterId`, as these are also generated in the background. 

We may also want to think about `resource.src`, which is updated from a blob to a url which an upload is completed. 

It is also work noting that properties on resource  like isTrimming, isMutting and isTrancoding are being removed as part of https://github.com/google/web-stories-wp/pull/9785

packages/story-editor/src/app/story/effects/useHistoryEntry.js
```diff
@@ -23,28 +23,83 @@ import { useEffect, useRef } from '@web-stories-wp/react';
...
-  useEffect(() => {
-    stateToHistory({
-      story,
-      current: currentPageIndexRef.current,
-      selection: selectedElementIdsRef.current,
-      pages,
-      capabilities,
+  const deleteKeysFromPages = (list) => {
+    // Create a copy of the list not to influence the original.
+    return JSON.parse(JSON.stringify(list)).map((page) => {
```
barklund: Have you checked how expensive this is? Maybe a regular `deepClone()` function is better because I fear this is quite computationally expensive.

packages/story-editor/src/app/story/effects/useHistoryEntry.js
```diff
@@ -23,28 +23,83 @@ import { useEffect, useRef } from '@web-stories-wp/react';
...
+        (key) =>
+          JSON.stringify(withoutPages[key]) ===
+          JSON.stringify(currentHistoryEntryRef.current[key])
+      );
+      // If only pages have changed, check if relevant properties have changed.
+      if (onlyPagesChanged) {
+        const adjustedPages = deleteKeysFromPages(pages);
+        const adjustedEntryPages = deleteKeysFromPages(
+          currentHistoryEntryRef.current.pages
+        );
```
barklund, line 87: Couldn't we do this before storing the ref? In fact, couldn't we do the stringify too - so the ref is a map from the three properties (`story`, `capabilities`, and `pages`) to JSON strings?

packages/story-editor/src/app/story/effects/useHistoryEntry.js
```diff
@@ -23,28 +23,83 @@ import { useEffect, useRef } from '@web-stories-wp/react';
...
+  const deleteKeysFromPages = (list) => {
+    // Create a copy of the list not to influence the original.
+    return JSON.parse(JSON.stringify(list)).map((page) => {
+      const { elements } = page;
+      return {
+        ...page,
+        elements: elements.map((element) =>
+          deleteNestedKeys(element, ELEMENT_PROPS_TO_IGNORE)
+        ),
+      };
```
barklund: Because we're working on a cloned object, we can actually mutate it, so we can do `page.elements = page.elements.map()`. That would make this a bit faster I think.

barklund: Oh, you are actually mutating it in place in the `deleteNestedKeys` function anyway. So no need to set anything, you can just do:

```
page.elements.forEach(deleteNestedKeys(ELEMENT_PROPS_TO_IGNORE));
```

And then you can change `deleteNestedKeys` to be _curryable_ as in:

```js
const deleteNestedKeys = paths => object => {};
```

And finally there's no need to have the return in `deleteNestedKeys` as it just mutates the object in place.

packages/story-editor/src/app/story/utils/deleteNestedKeys.js
```diff
@@ -0,0 +1,46 @@
...
+      return;
+    }
+    if (keys.length === 1) {
+      delete object[keys[0]];
+      return;
+    }
+    const lastKey = keys.pop();
+    const nextLastKey = keys.pop();
+    const nextLastObj = keys.reduce((a, key) => a?.[key] || a, object);
+    if (nextLastObj?.[nextLastKey]?.[lastKey]) {
```
barklund: What if the value is falsy or even `null`? We should still delete it, no? It should only skip deletion if (already) undefined.

You can use [`object.hasOwnProperty`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty) to check if a property is set at all.

barklund (commented):
The approach is good - checking before even adding to history. I am just very worried about speed here. This is a lot of code that runs on every single tiny change of properties, so it has to be very fast. We need to check this on 30+ page stories with many elements (especially media elements, as they take a lot of space in the story data).

Even if it just takes 20ms on a large story, that's still quite expensive because it can run so often.

packages/story-editor/src/app/story/effects/useHistoryEntry.js
miina: No, I haven't yet, I'll try it out.

I was also wondering about the speed here but it needs to be done somehow anyway -- within reducer or not. I'll do some testing.

packages/story-editor/src/app/story/utils/deleteNestedKeys.js
miina: Yep, good point.

packages/story-editor/src/app/story/effects/useHistoryEntry.js
miina: Tested it out using `clone-deep` package vs stringifying using a 40 page story with various elements including a ridiculous amount of media on the pages, and it looks like even in these extreme conditions, the cloning takes approximately 2ms for both. Sometimes, however, when adding a new element, the stringifying spiked to 5ms.

When using a normal story (20 pages, only a few media elements per page), it the tracker consistently got 1ms for this. Considering that it will skip creating a new history entry, that'll probably be an improvement.

I'll go with using `clone-deep` since `JSON.stringify` can potentially break if we should start storing other type of data in the objects (other than strings and numbers, that is).

packages/story-editor/src/app/story/effects/useHistoryEntry.js
miina: We could, however, I was trying to save some time here -- this runs only if `onlyPagesChanged`, meaning that it doesn't have to always run. Thoughts?

packages/story-editor/src/app/story/effects/useHistoryEntry.js
miina: Thanks for these, any others from media elements you can think of that are mutating history?

packages/story-editor/src/app/story/effects/useHistoryEntry.js
```diff
@@ -17,34 +17,90 @@
...
  */
 import { useHistory } from '../../history';
+import deleteNestedKeys from '../utils/deleteNestedKeys';
+
+// Changes to these properties of elements do not create a new history entry.
+const ELEMENT_PROPS_TO_IGNORE = [
+  'resource.baseColor',
+  'resource.isMuted',
+  'resource.poster',
+  'resource.posterId',
```
miina: Hm, I'm not sure what would be the best way to test this change 🤔 

Since there are still the transient props then history is still being mutated in a lot of cases.

E.g. on `main`, when I drag a video from my computer directly onto the Page, it takes 7 clicks of Undo to get back to the state without video. On this PR, however, it takes "only" 4 clicks -- there are still the transient props changes being tracked there.

Not sure what would be a good way to QA this though. Probably a combination of #9785 and this PR would be easier to test.

Thoughts here? cc @spacedmonkey @swissspidy

swissspidy: We can certainly wait until #9785 lands, it's fairly close.

After that it should be easier to test.

packages/story-editor/src/app/story/effects/useHistoryEntry.js
barklund: But isn't that most updates? Aren't almost all updates to pages in everyday workflow?

packages/story-editor/src/app/story/utils/deleteNestedKeys.js
```diff
@@ -0,0 +1,50 @@
...
+        return;
+      }
+      if (keys.length === 1) {
+        delete object[keys[0]];
+        return;
+      }
+      const lastKey = keys.pop();
+      const nextLastKey = keys.pop();
+      const nextLastObj = keys.reduce((a, key) => a?.[key] || a, object);
+      if (Object.prototype.hasOwnProperty.call(nextLastObj, nextLastKey)) {
```
barklund, line 43: You're **not** trusting that object 😂 .

barklund (approved):
Works for me and doesn't seem (too) expensive.

packages/story-editor/src/app/story/effects/useHistoryEntry.js
miina: It is most updates indeed, but there are still some cases where it doesn't happen (when changing the story properties), and the ref is updating every time anyway when the original values change. Just not sure what would be the benefit of moving it to ref if it would still run the same (or more times) then. Thoughts?

swissspidy:
@miina Is this intentionally in the ready for QA column? It's not really user facing and there are no QA instructions.

miina:
> Is this intentionally in the ready for QA column? It's not really user facing and there are no QA instructions.

Yes, it's waiting for the Transient props PRs to get merged for being able to QA it better, see https://github.com/google/web-stories-wp/pull/9822#discussion_r760156008

We can also just merge without QA though since it has been eng-tested. WDYT?

PR #9825, opened 2021-11-26
Decoupling: Help Center

mohdsayed:
## Context

Since there are some WordPress-specific items in quick tips, they need to be moved to `wp-story-editor`

## Testing Instructions

1.

## Checklist

---


packages/story-editor/src/components/helpCenter/constants.js
```diff
@@ -150,24 +133,7 @@ export const Z_INDEX = {
...
-      [tipKey]: PropTypes.bool,
-    }),
-    {}
-  ),
-});
-
-export const TIP_KEYS_MAP = Object.keys(TIPS).reduce((keyMap, key) => {
-  keyMap[key] = true;
-  return keyMap;
-}, {});
```
swissspidy, line 173: After reading the current code on `main` I really don't understand why `TIPS_KEY_MAP` is needed.

Everywhere it's just really used with `Object.keys(TIP_KEYS_MAP)`. This could just be replaced with `Object.keys(TIPS)` or `TIPS` (or well, `TIPS + additionalTips`) directly.

packages/story-editor/src/app/helpCenter/provider.js
```diff
@@ -106,10 +110,12 @@ export const initialState = {
   readError: false,
   unreadTipsCount: persisted?.unreadTipsCount ?? 0,
   isHydrated: false,
-};
+  tips: { ...TIPS, ...additionalInitialState.additionalTips },
+  tipKeys: Object.keys({ ...TIPS, ...additionalInitialState.additionalTips }),
```
maxyinger, line 114: 👍

packages/story-editor/src/app/helpCenter/useHelpCenter/effects.js
```diff
@@ -77,25 +76,26 @@ export const createDynamicNavigationFlow = (previous, next) => {
   if (!isComingFromMenu(previous, next)) {
     return {};
   }
-  const appendedTips = BASE_NAVIGATION_FLOW
+  const appendedTips = next.navigationFlow
```
maxyinger: I think the expected behavior could be documented a little better here, sorry.

So one of the UX features we have to accommodate here is appending all the unread tips to the end so a user can click into any tip, then press `next` all the way until they get to the last tip and have viewed all unread tips.

ie:
- lets say tips are [tip1, tip2, tip3, tip4] and they're all unread
- user opens the main menu
- user clicks into tip3 first
- the resulting navigation flow should look like: [tip1, tip2, tip3, tip4, tip1, tip2]

I think you can just update this to be the following and it should be fine:
 ```
   const baseNavigationFlow = next.tipKeys;
   const appendedTips = baseNavigationFlow;
    // Tips before current index
    .slice(0, next.navigationIndex)
    // Filter out read tips
    .filter((tip) => !next.readTips[tip]);
  return {
    navigationFlow: [
      ...baseNavigationFlow,
      ...appendedTips,
      DONE_TIP_ENTRY[0],
    ],
  };
 ```

packages/story-editor/src/components/helpCenter/constants.js
maxyinger: yea I think the original intention was to keep a consistent type with the read tips of `{ [k in keyof tips]: boolean }`, but I think you're right and you could substitute out `TIPS + additionalTips` it would work the same

packages/story-editor/src/components/helpCenter/menu/tips.js
```diff
@@ -90,8 +90,12 @@ Tip.propTypes = {
   unread: PropTypes.bool,
 };
 
-const TIPS_ITERABLE = Object.entries(TIPS);
 export function Tips({ onTipSelect = () => {}, readTips }) {
+  const {
+    state: { tips },
+  } = useHelpCenter();
+  const TIPS_ITERABLE = Object.entries(tips);
```
maxyinger: do we want to memoize this at all just to prevent any re-renders?

I'd also be open to camel casing it since it's not a module scoped constant

packages/story-editor/src/app/helpCenter/useHelpCenter/effects.js
maxyinger: you can test this behavior out manually by clearing all your read tips as a user by typing this into the console to clear all your read tips on the backend:
```
await wp.apiFetch({ path: '/web-stories/v1/users/me', method: 'POST', data: { meta: { web_stories_onboarding: {} }}})
```

maxyinger (commented):
Awesome work! Looks great so far. Let me know if you got any Q's along the way

packages/story-editor/src/components/helpCenter/constants.js
mohdsayed: Removed in https://github.com/google/web-stories-wp/pull/9825/commits/edf162cfd348f117591a13adac662541b54198e1

packages/story-editor/src/app/helpCenter/useHelpCenter/effects.js
mohdsayed: Thanks updated

packages/story-editor/src/components/helpCenter/menu/tips.js
mohdsayed: Thanks, yes it should have been camel case, updated. 

Curious to know how memoizing it will prevent re-renders?

swissspidy (approved):
LGTM; just waiting for the others' approval

BrittanyIRL (approved):
This looks great and works as expected in the editor, updates make sense to me.

packages/story-editor/src/components/helpCenter/menu/tips.js
maxyinger: Oh yea I guess you're right. It won't. never mind

maxyinger (approved):

PR #9826, opened 2021-11-26
Editor: Support publishing pending stories

### LINKED ISSUE #9798: Publishing: Support publishing pending stories
## Feature Description

When opening a pending story as an admin, it's not currently possible to publish it, continue saving it as pending, or switching to a draft.

Compare this with Gutenberg:

Screenshot 2021-11-24 at 14 39 31

To fix this, we need to update the buttons in the header like this for pending stories:

Screenshot 2021-11-24 at 14 54 11

Bonus points for an e2e test.

swissspidy:
## Context

It wasn't previously possible for admins to publish a pending story through the editor.

## Summary

This change gives admin users the expected ability to publish pending stories, or revert them to regular drafts again.

## Relevant Technical Choices

- Some slight refactorings to make code better readable

## User-facing changes

Contributor users will have two buttons: Save (as draft) and Submit for review:

Screenshot 2021-11-26 at 15 38 37

Admins, when editing a pending story will have three buttons: Switch to draft, save (as pending), publish

Screenshot 2021-11-26 at 15 38 32

## Testing Instructions

1. As a contributor user, create a new story
2. Submit it for review
3. Log in as admin user and edit the pending story
4. Verify that the story can be saved, switched to draft, or published

No

No

No

## Checklist

---

Fixes #9798


packages/story-editor/src/components/header/buttons/update.js
```diff
@@ -75,38 +82,53 @@ function UpdateButton({ hasUpdates = false, forceIsSaving = false }) {
...
+          type={BUTTON_TYPES.QUATERNARY}
+          size={BUTTON_SIZES.SMALL}
+          onClick={() => saveStory({ status: 'draft' })}
+          disabled={!isEnabled}
+          aria-label={__('Save draft', 'web-stories')}
+        >
+          <Icons.FloppyDisk />
+        </Button>
+      </Tooltip>
+    );
```
spacedmonkey:
```suggestion
  if ('pending' === status || 'draft' === status) {
    const text = 'pending' === status ? __('Save as pending', 'web-stories') : __('Save draft', 'web-stories');
    const saveState = 'pending' === status ? {} : { status: 'draft' };
    return (
      <Tooltip title={text} hasTail>
        <Button
          variant={BUTTON_VARIANTS.SQUARE}
          type={BUTTON_TYPES.QUATERNARY}
          size={BUTTON_SIZES.SMALL}
          onClick={() => saveStory(saveState)}
          disabled={!isEnabled}
          aria-label={text}
        >
          <Icons.FloppyDisk />
        </Button>
      </Tooltip>
    );
```

This means a lot less repeated code.

swissspidy: I'll just extract the component

spacedmonkey:
Niced a weird behavior, going from scheduled to private ends in showing publish dialog again.   

https://user-images.githubusercontent.com/237508/143611306-cf65013b-9219-496e-9355-3832dea7b47c.mov

Seems unrelated to this change.

spacedmonkey:
This is looking good, one weird thing I noticed, but once answered, will approve. Great work!

swissspidy:
> Niced a weird behavior, going from scheduled to private ends in showing publish dialog again.
> 
> Seems unrelated to this change.

Correct, it's unrelated—but it should be trivial to address in this PR. I'll take a look.

If it turns out to be a bigger issue, I'll create a ticket for it.

packages/wp-story-editor/src/components/header/buttons/index.js
```diff
@@ -101,20 +105,24 @@ function Buttons() {
           <PreviewButton forceIsSaving={isSavingMetaBoxes} />
           {isSaving && <Loading />}
         </IconWithSpinner>
-        {isDraft ? (
+        {isPending && <SwitchToDraftButton forceIsSaving={isSavingMetaBoxes} />}
```
miina: Shouldn't the `SwitchToDraftButton` be just displayed only if `!isDraft`?

Currently, it's displayed when `isPending` or when `!isDraftOrPending`, it's a bit confusing -- that it's displayed if it's pending and if it's not pending. Sounds like just `!isDraft` would be sufficient.

packages/story-editor/src/components/header/buttons/update.js
```diff
@@ -75,38 +108,45 @@ function UpdateButton({ hasUpdates = false, forceIsSaving = false }) {
...
+    hasFutureDate && !status === 'private'
+      ? __('Schedule', 'web-stories')
+      : __('Update', 'web-stories');
+
   return (
     <ButtonWithChecklistWarning
       text={text}
       onClick={() => saveStory()}
       disabled={isSaving || isUploading}
+      isUploading={isUploading}
```
miina, line 148: Should it be disabled for other media processes as well?

spacedmonkey: +1

spacedmonkey:
Unit tests failure seems valid here.

packages/story-editor/src/components/header/buttons/update.js
swissspidy: Worth noting that this is not doing anything new, just fixing a bug.

`ButtonWithChecklistWarning` accepts an `isUploading` prop:

https://github.com/google/web-stories-wp/blob/f22db1e7ddc307a07fabc020c536eb6f8d8a8904/packages/story-editor/src/components/header/buttons/buttonWithChecklistWarning.js#L111-L115

In one place the prop is passed as expected:

https://github.com/google/web-stories-wp/blob/f22db1e7ddc307a07fabc020c536eb6f8d8a8904/packages/story-editor/src/components/header/buttons/publish.js#L111-L117

But in this component here it was simply forgotten.

Any change in behavior beyond that warrants its own ticket.

packages/wp-story-editor/src/components/header/buttons/index.js
swissspidy: The difference is the order how the buttons are being rendered.

Draft stories get the following buttons, in this order:

1. Save as draft
2. Publish post (or submit for review if you're a contributor)

Published stories get:

1. Switch to draft
2. Update story 

Pending stories get:

1. Switch to draft
2. Update (if you're an admin; means continue saving as pending)
3. Publish post (or submit for review if you're a contributor)

Open for any suggestions to improve readability while maintaining the same order

spacedmonkey (approved):
LGTM

packages/wp-story-editor/src/components/header/buttons/index.js
miina: Hm, since now we have so many different statuses, maybe we should just display the buttons per status for readability, instead of trying to match the conditions? Even though some of those would repeat, it might still be worthwhile.

E.g.
```
 case 'pending`:
    return (... all buttons for pending ...);
 case `draft`:
    return (... all buttons for draft ...);
 ...etc.
```

(Preview button could be outside of the `switch` since that's always displayed for all statuses)
Thoughts?

spacedmonkey (changes requested):
Once unit tests are passing, I will reapprove.

packages/wp-story-editor/src/components/header/buttons/index.js
swissspidy: Improved now to make more readable 👍

swissspidy:
The karma test failure is unrelated / flakey

miina (approved):

PR #9830, opened 2021-11-29
Carousel: Fix incorrect attribute usage for disabling the button

### LINKED ISSUE #9814: Update carousel to indicate that the user is at the first / last page in the story
## Feature Description
Originally reported by @BrookeGraham 

> When scrolling through all the pages within the carousel it is hard to determine when you are at the end. The right arrow doesn’t change, / there is no indication you’ve reached the last page. Same when scrolling to the first page. 

We should disable the carousel arrows accordingly if we have reached the first / last page.

miina:
## Summary
Looks like the buttons were using `isDisabled` instead of `disabled` which caused the buttons not being properly disabled.

## Testing Instructions

1. Add a few pages so that the carousel would have arrows for scrolling
2. Verify the arrows get disabled if the carousel is already scrolled to the beginning/end.

No

No

No

## Checklist

---

Fixes #9814


barklund (approved):
🤦

swissspidy (approved):
Works like a charm

PR #9832, opened 2021-11-29
Right Click Menu: Add `Duplicate Element` action

### LINKED ISSUE #4616: Right Click Menus: Add Duplicate Element Action
## User Request

User requested an icon to duplicate elements or text like the "duplicate page" icon underneath the canvas.

## Feature Description 
- Adding a "duplicate" action into right click menus 
- Creating a new right click menu for when multiple elements are selected and giving that menu  the "duplicate" action as well as "Clear Styles" action. 

## Additional Context

>The ability to duplicate an element within the story. ie text box. Just like you duplicate the entire story slide, same icon but to duplicate an item within a slide.

---

## Acceptance Criteria

- Ability to duplicate elements via right click menu
- New right click menu for when multiple elements are selected that has the ability to duplicate elements. 

## Implementation Brief

Initially this feature request included the ability to "Clear Styles"- pulling that out to be separate and avoid scope creep.

samwhale:
## To-do

- [x] add karma test

## User-facing changes

|Mouse|Keyboard + undo|
|--|--|
|duplicate-mouse|duplicate-keyboard|

## Testing Instructions

1. Select multiple elements (hold the `Shift` key and click elements in the canvas)
2. Right click the area of the selected elements
3. Should see a context menu with one item: `Duplicate Elements`
4. Clicking that should duplicate all selected elements. Should not work on the background
5. Opening the context menu using the keyboard should work with the keyboard shortcut when multiple elements are selected: `mod+alt+shift+m`
6. Opening the context menu by right clicking multiple selected layers in the layer panel should work.

## Checklist

---

Fixes #4616

### DIFF
--- a/packages/story-editor/src/app/rightClickMenu/provider.js
+++ b/packages/story-editor/src/app/rightClickMenu/provider.js
@@ -740,6 +740,17 @@ function RightClickMenuProvider({ children }) {
     [handleMouseDown]
   );
 
+  const duplicateElementAction = useMemo(
+    () => ({
+      label: RIGHT_CLICK_MENU_LABELS.DUPLICATE_ELEMENTS(
+        selectedElements?.length
+      ),
+      onClick: handleDuplicateElements,
+      ...menuItemProps,
+    }),
+    [handleDuplicateElements, menuItemProps, selectedElements?.length]
+  );
+
   const layerItems = useMemo(
     () => [
       {
@@ -871,6 +882,7 @@ function RightClickMenuProvider({ children }) {
 
   const textItems = useMemo(
     () => [
+      { ...duplicateElementAction, separator: 'bottom' },
       ...layerItems,
       {
         label: RIGHT_CLICK_MENU_LABELS.COPY_STYLES,
@@ -898,6 +910,7 @@ function RightClickMenuProvider({ children }) {
       },
     ],
     [
+      duplicateElementAction,
       layerItems,
       handleAddTextPreset,
       menuItemProps,
@@ -925,6 +938,7 @@ function RightClickMenuProvider({ children }) {
       : RIGHT_CLICK_MENU_LABELS.CLEAR_IMAGE_STYLES;
 
     return [
+      { ...duplicateElementAction, separator: 'bottom' },
       ...layerItems,
       {
         label: RIGHT_CLICK_MENU_LABELS.SET_AS_PAGE_BACKGROUND,
@@ -969,6 +983,7 @@ function RightClickMenuProvider({ children }) {
     ];
   }, [
     copiedElement,
+    duplicateElementAction,
     handleClearElementStyles,
     handleCopyStyles,
     handleOpenScaleAndCrop,
@@ -983,6 +998,7 @@ function RightClickMenuProvider({ children }) {
 
   const shapeItems = useMemo(
     () => [
+      { ...duplicateElementAction, separator: 'bottom' },
       ...layerItems,
       {
         label: RIGHT_CLICK_MENU_LABELS.COPY_SHAPE_STYLES,
@@ -1011,6 +1027,7 @@ function RightClickMenuProvider({ children }) {
     ],
     [
       copiedElement?.type,
+      duplicateElementAction,
       handleAddColorPreset,
       handleClearElementStyles,
       handleCopyStyles,
@@ -1021,17 +1038,14 @@ function RightClickMenuProvider({ children }) {
     ]
   );
 
-  const stickerItems = useMemo(() => [...layerItems], [layerItems]);
+  const stickerItems = useMemo(
+    () => [{ ...duplicateElementAction, separator: 'bottom' }, ...layerItems],
+    [duplicateElementAction, layerItems]
+  );
 
   const multipleElementItems = useMemo(
-    () => [
-      {
-        label: RIGHT_CLICK_MENU_LABELS.DUPLICATE_ELEMENTS,
-        onClick: handleDuplicateElements,
-        ...menuItemProps,
-      },
-    ],
-    [handleDuplicateElements, menuItemProps]
+    () => [duplicateElementAction],
+    [duplicateElementAction]
   );
 
   const menuItems = useMemo(() => {
--- a/packages/story-editor/src/app/rightClickMenu/provider.js
+++ b/packages/story-editor/src/app/rightClickMenu/provider.js
@@ -109,6 +109,7 @@ function RightClickMenuProvider({ children }) {
     currentPage,
     currentPageIndex,
     deleteCurrentPage,
+    duplicateElementById,
     pages,
     setBackgroundElement,
     selectedElements,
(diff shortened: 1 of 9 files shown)


packages/story-editor/src/components/canvas/karma/rightClickMenu.karma.js
```diff
@@ -1017,4 +1058,79 @@ describe('Right Click Menu integration', () => {
       });
     });
   });
+
+  describe('right click menu: multiple elements selected', () => {
+    it('should duplicate all selected elements', async () => {
```
samwhale, line 1206: 💭 this also duplicates animations. It would be good to verify that in this test too.

packages/story-editor/src/components/canvas/karma/rightClickMenu.karma.js
```diff
@@ -1017,4 +1058,79 @@ describe('Right Click Menu integration', () => {
...
+            selectedElements: state.selectedElements,
+            initialElements: state.currentPage.elements,
+          }))
+      );
+
+      expect(selectedElements.length).toBe(2);
+      expect(initialElements.length).toBe(4);
+
+      // open right click menu
+      await rightClickOnTarget(imageFrame);
```
samwhale, line 1243: here I'm right clicking on the image but you can technically right click anywhere in the selection area

Decided against testing that since selection using the multi select area is tested in the multi select karma tests. thoughts?

packages/story-editor/src/components/canvas/selection.js
```diff
@@ -84,7 +84,9 @@ function Selection() {
   }
 
   // Multi-selection.
-  return <MultiSelectionMoveable selectedElements={selectedElements} />;
+  return (
+    <MultiSelectionMoveable {...props} selectedElements={selectedElements} />
+  );
```
samwhale, line 89: Just above here there is the `SingleSelectionMoveable` too. That component didn't need the `onContextMenu` prop since the `FrameLayer` handles the right click menu for single select.

💭 Probably should add a comment here about that, but still feels pretty weird.

packages/story-editor/src/app/rightClickMenu/provider.js
```diff
@@ -230,6 +226,17 @@ function RightClickMenuProvider({ children }) {
...
 
+  /**
+   * Duplicate all selected elements.
+   */
+  const handleDuplicateElements = useCallback(() => {
+    if (!selectedElements.length) {
+      return;
+    }
+
+    selectedElements.map(({ id }) => duplicateElementById({ elementId: id }));
```
maxyinger, line 237: oh nice. this works nice here

packages/story-editor/src/components/canvas/karma/rightClickMenu.karma.js
maxyinger: Makes sense as an integration test I guess, but also `duplicateElementById` has test coverage to verify that animations get copied, so not a huge deal either way.

🤔 I guess it's good just from a regression testing perspective if the implementation ever changes

packages/story-editor/src/components/canvas/karma/rightClickMenu.karma.js
maxyinger: yea I'm fine with that. I think it's kinda one half dozen the other. Seems like the fixture makes it easier to grab this node vs the multi select area so I'd stick with this since it's more convenient

packages/story-editor/src/components/canvas/selection.js
maxyinger: Would it be better just to just handle the `onContextMenu` in `SingleSelectionMoveable` instead of the `FrameLayer` to consolidate the approach? or would that cause regressions with the bg or some other behavior?

maxyinger (approved):
lgtm. nice work n tests

packages/story-editor/src/components/canvas/selection.js
samwhale: Yeah so I tried adding `onContextMenu` to the single select but it wasn't working for some reason 🤔 wondering if the frame (when single selected) renders under the canvas frame area.

Thinking about it more - I think the `onContextMenu` handler belongs here so it'd be better to make it work on `SingleSelectMovable`

BrittanyIRL (changes requested):
Works great! 

To catch this PR up to slack convos, since this is a new feature that we're kinda building as we discuss: 

Screen Shot 2021-12-02 at 2 55 22 PM

--------------- 
Since it's kinda weird having just 1 option for when multiple elements are selected let's add in the ability to send backward or bring forward (are there other actions that are also agnostic to element type?). 

We could also, after that, cross the bridge of if you've selected say, 4 elements that are all font and applying changes at once. Let's start off with element type agnostic actions though. I'll make a ticket for sendback/bring forward to avoid scope creep here. 

For this, i think it's important that you be able to duplicate 1 element as well - which I'm not seeing? 
Screen Shot 2021-12-02 at 2 58 46 PM

Could you add that in here so that duplicate element is available as a right click option in a single element right click menu and also multiple element right click menus?

samwhale:
>Since it's kinda weird having just 1 option for when multiple elements are selected let's add in the ability to send backward or bring forward (are there other actions that are also agnostic to element type?).

These are the actions that are element type agnostic:
- *Layer actions*: Bring to front, bring forward, send backward, send to back
- *Page actions*: Add new page after, add new page before, duplicate page, delete page

I'm guessing if we send layers all the way back or forward it should preserve order, even with layers in between selected layers? 

This also smells like scope creep and we'd have to write some extra logic that doesn't exist yet. Maybe we can do all the layer actions together so that product can also decide what the behavior should look like ([@barklund's comment about it](https://xwp.slack.com/archives/C01ESBB8ZN1/p1638499479029800?thread_ts=1638402374.016000&cid=C01ESBB8ZN1))

>For this, i think it's important that you be able to duplicate 1 element as well - which I'm not seeing?

Sure I can add it to each element menu! I didn't realize it should be shown for single selection as well, but that will be fixed soon.

samwhale:
Can now duplicate all elements (as long as they aren't set to the background): [5c5d463](https://github.com/google/web-stories-wp/pull/9832/commits/5c5d463daec28ac14e10e601643bed0f1cafbc31). Selecting multiple elements works as before. Notice the singular vs the plural version of `Duplicate Elements` depending on how many elements are selected

duplicate-errythang

BrittanyIRL:
> I'm guessing if we send layers all the way back or forward it should preserve order, even with layers in between selected layers?
> 
> This also smells like scope creep and we'd have to write some extra logic that doesn't exist yet. Maybe we can do all the layer actions together so that product can also decide what the behavior should look like ([@barklund's comment about it](https://xwp.slack.com/archives/C01ESBB8ZN1/p1638499479029800?thread_ts=1638402374.016000&cid=C01ESBB8ZN1))

Morten's context was so great here! Let's not do any of that here, it is absolutely scope creep (i was going to make new issues) but it's also work that isn't needed to get the requested functionality. The 1 option menu looks a little goofy to me but it's a good starting point and it adds the desired feature.

BrittanyIRL (approved):
thanks for the update!

PR #9833, opened 2021-11-29
Editor: Fix position and layering of page attachment layer

### LINKED ISSUE #7566: Zoom: Page Attachment not positioned correctly.

barklund:
## Context

This moves the page area overlay (the page attachment) to the proper position in the new zoomed canvas layering as well as fixes the positioning of the page attachment, so it's always a fixed margin from the bottom of the page regardless of zoom (just like in an actual published story).

## User-facing changes

| Zoom | Scroll | Image |
|-|-|-|
| Fit | N/A |  |
| Fill | Not bottom |  |
| Fill | Bottom |  |
| 100% | Not bottom |  |
| 100% | Bottom |  |
| 200% | Not bottom |  |
| 200% | Bottom right |  |

## Testing Instructions

1. Add a page attachment
2. Change the zoom to 200%
3. Observe that the page attachment button correctly moves around with the zoomed page as it scrolls.

## Checklist

---

Fixes #7566

### DIFF
--- a/packages/story-editor/src/components/canvas/layout.js
+++ b/packages/story-editor/src/components/canvas/layout.js
@@ -136,6 +136,15 @@ const PageAreaContainer = styled(Area).attrs({
         100% - ${hasHorizontalOverflow ? themeHelpers.SCROLLBAR_WIDTH : 0}px
       );
     `}
+
+      overflow: ${({ showOverflow }) => (showOverflow ? 'visible' : 'hidden')};
+      width: calc(
+        100% - ${hasVerticalOverflow ? themeHelpers.SCROLLBAR_WIDTH : 0}px
+      );
+      height: calc(
+        100% - ${hasHorizontalOverflow ? themeHelpers.SCROLLBAR_WIDTH : 0}px
+      );
+    `}
 `;
 
 const Layer = forwardRef(function Layer({ children, ...rest }, ref) {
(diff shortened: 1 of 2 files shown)


miina (approved):
🐶

miina:
@barklund I already approved the PR but just noticed that the original issue also reported that the size of the Page Attachment doesn't change when zooming in.

I'm actually not sure what's the best approach here, e.g. should 100% Zoom be the currently used size, and in any other case, it should be transformed accordingly? We might need to follow up with UX separately, unless you have a specific idea.
Thoughts?

barklund:
> @barklund I already approved the PR but just noticed that the original issue also reported that the size of the Page Attachment doesn't change when zooming in.

I checked what happens in the published story, and the size of the page attachment button stays fixed even when the whole story scales up and down, so I felt it would be misleading if we scaled the button up and down in the canvas.

miina:
>  checked what happens in the published story, and the size of the page attachment button stays fixed even when the whole story scales up and down, so I felt it would be misleading if we scaled the button up and down in the canvas.

Alright, I think zooming is a bit different than scaling though -- agreed that if if the viewport size changes (scaling) but the Zoom level stays the same, then the size should stay the same but when zooming, it would probably still be nice to adjust together with the other elements. Anyway -- not a priority, the main issues got fixed in this PR already!

Trapsta (approved):
✅

Issue #9676, opened 2021-11-10
Prevent layout shift while loading dashboard


BrittanyIRL:
There's some layout shift occurring in the dashboard on initial load that could be updated to help our perf. Mostly this is the telemetry banner + headline. This ticket is to explore preventing that. 

You can see this by running lighthouse perf on the dashboard before and after making the update. Remember that locally running lighthouse is _not_ the same as running it on staging (our dev environment is wayyyyy slower than prod).

PR #9844, opened 2021-11-29
Dashboard: Remove Excess Header Render

BrittanyIRL:
## Context

There's some layout shift occurring in the dashboard on initial load that could be updated to help our perf. Mostly this is the telemetry banner + headline. This ticket is to explore preventing that.

## Summary

So, turns out that in the year since the telemetry banner got added into the dashboard's code base, it was pulled out into wp-dashboard package and it's getting rendered as a portal based on `#body-view-options-header` so it has to get added into the dashboard after the initial load, so we can't get rid of that shift. Looking into this, I did notice that I could cut down 1 render on the header though, so this PR is for that. 

## Relevant Technical Choices

Add new prop `initialPageReady` in the `useStoryView` hook. This is updated once `totalPages` isn't null - which means that initial data is present. Passing this boolean into the `Header` allows us to prevent excess renders on the pills and search inputs while leaving the header visible to avoid content shift. While in there cleaning up the 2 props variables needed for loading state in the stories content and just combining them ahead of time into the object they became further along the trail.

Now the Header component is rendered 3 times initially instead of 4 (wowwwww).

## User-facing changes

Shouldn't be anything noticeable. The Stories pills and search on the Header will show up after there's data there, so the pills always have an accurate count and ya can't jump the gun searching stories without the initial data present. 

## Testing Instructions

Does the dashboard load like normal?

1.

## Checklist

---

See #9676

### DIFF
--- a/packages/dashboard/src/app/views/myStories/content/index.js
+++ b/packages/dashboard/src/app/views/myStories/content/index.js
@@ -51,14 +51,13 @@ import StoriesView from './storiesView';
 function Content({
   allPagesFetched,
   filter,
-  isLoading,
+  loading,
   page,
   search,
   sort,
   stories,
   storyActions,
   view,
-  showStoriesWhileLoading,
 }) {
   return (
     <Layout.Scrollable>
@@ -67,19 +66,15 @@ function Content({
           <>
             <StoriesView
               filterValue={filter.value}
-              isLoading={isLoading}
               sort={sort}
               storyActions={storyActions}
               stories={stories}
               view={view}
-              loading={{
-                isLoading,
-                showStoriesWhileLoading,
-              }}
+              loading={loading}
             />
             <InfiniteScroller
               canLoadMore={!allPagesFetched}
-              isLoading={isLoading}
+              isLoading={loading?.isLoading}
               allDataLoadedMessage={__('No more stories', 'web-stories')}
               onLoadMore={page.requestNextPage}
             />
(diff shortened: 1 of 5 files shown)


swissspidy:
> So, turns out that in the year since the telemetry banner got added into the dashboard's code base, it was pulled out into wp-dashboard package and it's getting rendered as a portal based on `#body-view-options-header` so it has to get added into the dashboard after the initial load, so we can't get rid of that shift.

If you have any suggestions for improving the way the telemetry banner is added to the dashboard, please let me know 🙂 

The reason for this change is the ongoing decoupling from WordPress. The "core" dashboard app might not have any settings page or telemetry option, so this was moved to the WordPress integration layer.

BrittanyIRL:
> If you have any suggestions for improving the way the telemetry banner is added to the dashboard, please let me know 🙂
> 
> The reason for this change is the ongoing decoupling from WordPress. The "core" dashboard app might not have any settings page or telemetry option, so this was moved to the WordPress integration layer.

Thanks, @swissspidy  - i tinkered a little bit to see if i could come up with something that helps but honestly, it's such a small content shift and the way this is handled makes sense and it reads clearly and it's specific to the platform. I'm a fan of the separation - just hadn't realized it was there when i made the ticket to prevent it! Which honestly, speaks to how smoothly your work of decoupling has been.

packages/dashboard/src/utils/useStoryView.js
```diff
@@ -50,6 +50,7 @@ export default function useStoryView({
   const [authorFilterId, _setAuthorFilterId] = useState(null);
   const [queriedAuthors, setQueriedAuthors] = useState([]);
   const showStoriesWhileLoading = useRef(false);
+  const initialPageReady = useRef(false);
```
maxyinger: I don't think I'm completely following. What's the behavior we get from this being a ref?

packages/dashboard/src/app/views/myStories/content/index.js
```diff
@@ -67,19 +66,15 @@ function Content({
...
-              isLoading={isLoading}
               sort={sort}
               storyActions={storyActions}
               stories={stories}
               view={view}
-              loading={{
-                isLoading,
-                showStoriesWhileLoading,
-              }}
+              loading={loading}
```
maxyinger, line 73: good catch

maxyinger (approved):

packages/dashboard/src/utils/useStoryView.js
BrittanyIRL: oh just me following the same pattern as loading, updating it to just be state instead.

samwhale (approved):
Works as you described. Sorry for such a late review!

Issue #5325, opened 2020-11-17
Allow selecting element by clicking on the outer border


miina:
## Feature Description
Initially, the outer border was implemented so that it wouldn't be possible to select an element clicking on it, however, after consulting with Sam, the expected behavior is to allow selecting an element by clicking on the border as well.

Additionally, in discussion with George, we should include the border to the selection line as well.

- [ ] Include the border into selection and allow clicking on the border to select it.

## Additional Context

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance Criteria
- The user should be able to select element by clicking on the outer border.

## Implementation Brief

The element needs to look exactly the same in the editor and in the front-end, the only thing that changes visually is the selection line including the border now, and the user being able to select the element by clicking on the border.

This can be achieved in a similar way as we're adding a padding:
- When border is added, the width and height of the element are increased accordingly, and decreased when the border is removed.
- The position is also adjusted accordingly, also considering rotation

We should additionally add some padding to the element to allow the size of the element content to stay the same.

The text size calculation needs to consider the text element content's size only (see `calculateTextHeight` and `calculateFitTextFontSize`)

To investigate: Edit modes for media and text -- probably should be as currently since only the content can be manipulated.

To investigate: option of adjusting the frame only? Probably better to keep frame and display in sync due to snapping, dragging, and generally not to complicate the readability.

To investigate: Still need to investigate how the shapes work since those have different implementation. It's possible that non-rectangular shapes will stay as they are -- the selection line is currently in the middle anyway so it's partially clickable and different from rectangular elements already now.

miina:
@barklund FYI added this issue after discussion with Sam.

miina:
Since #6171 is now being worked on, will mark this blocked by that issue as we should ensure the two will work together ok. Moving it out of the sprint for now as #6171 was pulled in again.

miina:
@merapi I've updated the implementation brief for a general overview.
Two points:
1. Edit mode for media: it seems to me that it might make sense to leave the media edit mode as it is since cropping and panning is for the media part only anyway so might be good to leave the border out of it, especially if it should be easier technically. WDYT?
[...]

PR #9851, opened 2021-11-30
Shapes: Add simple border option

### LINKED ISSUE #6171: Simple border for non-rectangular shapes
## Feature Description
Enable the border panel for non-rectangular shapes with the possibility to add a simple border that supports:
- One border-width value only. If different border values exist before and the user changes that element to a shape, the left border value is used.
- Allows choosing color, including opacity.
- Supports outside border only (as the rest of the elements).

## Additional Context

---

## Implementation Brief
For adding border to the shape, we could use an additional svg with stroke and stroke-width, in a similar way that the drop-targeting blue line is currently implemented when hovering over a shape / starting dropping into a shape:
```
      <svg
        viewBox={`0 0 1 ${1 / mask.ratio}`}
        width="100%"
        height="100%"
        preserveAspectRatio="none"
      >
        <path
          vectorEffect="non-scaling-stroke"
          stroke="yellow"
          strokeWidth="5"
          fill="none"
          d={mask?.path}
        />
      </svg>
```

Since adding a border directly to the existing element would create an inner border instead, we should add a separate element for the border.

We can add a new wrapper (or separate element) with the same clip-path, and increase and position this new element based on the border width (e.g. if border-width is 5px, increase it by 10px and move it to maintain visually the same position, too). We can then add the border svg (described above) into the wrapper with increased size. The currently existing mask could also be inside the new wrapper element, or could be separately as well.

This needs to be done in the `DisplayLayer`, note that we are already doing similar positioning for the outer border with the rest of the elements as well.

The structure could be something like this:
```
<IncreasedSizeWrapper clipPath=... width="calc(100% + 10px)" height="...">
	<svg .... ><path stroke="yellow" strokeWidth="5px" .... /></svg>
	<CurrentlyExistingMaskElement .... />
</IncreasedSizeWrapper />
```

or this:
```html
<svg .... ><path stroke="yellow" strokeWidth="5px" .... /></svg>
<NotIncreasedSizeWrapper clipPath="...">
	<CurrentlyExistingMaskElement ... />
</NotIncreasedSizeWrapper />
```

See the conversation and considered alternatives in more detail [here](https://github.com/google/web-stories-wp/issues/5536#issuecomment-771542974).

Trapsta:
## Context

Enables the border panel for non-rectangular shapes with the possibility to add a simple border that supports:

- A single border-width value only.
- Allows choosing color **but not** opacity for the border.
- Only a middle-positioned border (unlike for rectangular elements).

This is all tested and confirmed working with output, flip, rotation, focal changes, focal zoom, page zoom, RTL, layer opacity, and more.

### Todo

- [x] Fix media snapping bug (currently media has a hard time snapping into shapes - not sure what broke) Must be fixed in this PR!
- [x] Fix media edit mode when in non-rect bordered shape
- [x] Fix the weird scaling bug
- [ ] Figure out what happens when dropping a multi-bordered and/or semitransparent-bordered media element into a shape that does not support it (follow-up bug probably).
- [ ] File follow-up bug for the color picker opacity error text discrepancy mentioned below.

### Design panel
The new border width panel for non-rectangular shapes (note the single input, no lock button and no opacity for border color):
Screen Shot 2022-01-22 at 00 02 47

Note the following "bug" in the basic color picker, where transparency is not allowed (which is correct), but the error message assumes that this is because it's about setting page background color, which this isn't.
Screen Shot 2022-01-22 at 00 02 57

A follow-up bug should be filed when this is merged and the text should be customized for page background and non-rectangular shape/mask border-color respectively. Or it should be made generic enough to support both cases (e.g. _"The color on this element does not support opacity"_).

### Canvas
Here are the various elements with their borders increasing and decreasing again:

| border-rect | border-circ | border-tri | border-heart |
|-|-|-|-|
| border-star | border-pent | border-hex | border-blob1 |
| border-blob2 | border-blob3 |border-blob4 |border-blob5 |

### Output
Compare canvas with output:
| Canvas | Output |
|-|-|
| Screen Shot 2022-01-22 at 00 08 49 | Screen Shot 2022-01-22 at 00 08 59 |

## Testing Instructions

1. Add  a non-rectangular shape to the page
2. Select the shape and verify that you can edit the border width  from the border panel
3. verify that you can update the border width and color, but not the opacity
4. observe that you cannot toggle the border width to input multiple values
5. set layer opacity to 50% and observe that the whole element correctly fades down and there's no weirdness happening in the border overlap region

## Checklist

---

Fixes #6171


packages/story-editor/src/components/canvas/displayElement.js
```diff
@@ -70,6 +70,10 @@ const Wrapper = styled.div.attrs(
       border-radius: ${theme.borders.radius.small};
       overflow: hidden;
     `}
+
+      border-radius: ${theme.borders.radius.small};
+      overflow: hidden;
+    `}
```
barklund: This looks like the same annoying linter "fixes" that I've also experienced. Try removing these lines locally (only this) and commit it. If the linter adds it again and then says "nothing to commit", it's clearly the same issue.

A short-term workaround is to remove it in a commit and do "git commit --no-verify" to circumvent the linters. But as soon as you edit this file again, the linter will re-add this crap in pre-commit hook.

Long-term we really need to figure out why this happens.

packages/story-editor/src/components/panels/design/border/borderWidth.js
```diff
@@ -71,8 +72,13 @@ function WidthControls({ selectedElements, pushUpdateForObject }) {
     DEFAULT_BORDER
   );
 
+  // Some shapes i.e all non-rectangular shapes only support a single border width value
+  const singleBorderSupport = selectedElements.some((el) =>
+    singleBorderMask(el)
+  );
```
barklund: This can be collapsed to just this:

```js
  const singleBorderSupport = selectedElements.some(singleBorderMask);
```

I would also prefer the function (and the variable) having a "has" or "is" prefix given that it returns (or is) a boolean, just to make it clear. So maybe just "hasSingleBorderedMask" or something like that? We might have a similar function elsewhere that checks for non-rectangular masks.

barklund: Maybe the two are even misnamed? All shapes support a single border, the issue is that non-rectangular shapes don't support multiple border values, so perhaps it should be:

 ```js
  const hasMultiBorderSupport = selectedElements.every(hasMultiBorderedMask);
```

And then reverse the logic throughout. I feel that makes more sense when skimming the code later.

File: packages/story-editor/src/utils/elementBorder.js [hunk 3]
```diff
@@ -194,3 +198,28 @@ export function getResponsiveBorder(border, previewMode, converter) {
...
+ * Gets style for the border svg wrapper.
+ *
+ * @param {string} maskId Clipping mask id.
+ * @param {number} borderWidth border width value
+ * @return {Object} SVG border wrapper style.
+ */
+export function getBorderWrapperStyle(maskId, borderWidth) {
+  if (!maskId || !borderWidth) {
+    return {};
+  }
```
barklund: Could this check for the rectangular mask as well and just return nothing here?

barklund (commented):
Just some minor nits on style, but also a linter issue.

swissspidy:
Please don't simply remove half of the PR template; it's meant for filling out :-)

If you have suggestions for improving it, I'd be happy to hear them.

packages/story-editor/src/components/canvas/displayElement.js
swissspidy: @barklund Can you please create a ticket for this? Thanks.

miina:
From testing:

Otherwise it works great but I'm still seeing some issues with some of the shapes, it's especially visible on smaller screens and with the heart shape.
Screenshot 2021-12-01 at 11 07 11

Thoughts?

packages/story-editor/src/masks/output.js
```diff
@@ -77,35 +89,66 @@ export default function WithMask({
     );
   }
 
+  const borderWidth = border?.left;
+  const showSingleBorder = Boolean(singleBorderMask(element) && borderWidth);
+  const borderColor = border?.color
```
miina: If there's no `borderColor`, should we still display the border at all?

Agreed with Morten that `singleBorderMask` isn't the most self-explanatory name here, and it would be better to rename it to something like `hasMultiBorderSupport` or `canHaveMultiBorder` or sth similar.

Probably just using `showBorder` is ok here as well for the name.

packages/story-editor/src/masks/display.js
```diff
@@ -85,28 +107,53 @@ export default function WithMask({
...
-    >
-      <svg width={0} height={0}>
-        <defs>
-          <clipPath
-            id={maskId}
-            transform={`scale(1 ${mask.ratio})`}
-            clipPathUnits="objectBoundingBox"
+    <>
+      {showSingleBorder && (
+        <div style={{ ...style, ...wrapperStyle }}>
```
miina: Is there difference between the `display` and `output` for displaying the SVG border? If not, then maybe we could even move this to an util as well, e.g. `getMaskBorder` or something like that. WDYT?

miina:
There's one more issue just noticed: the border being partially below the element itself. E.g. here the element's background is gray with opacity 0.3, the black border can be seen partially under the element:
Screenshot 2021-12-01 at 17 06 57

miina:
@Trapsta 

IIRC then one alternative that was considered was not having border as a separate SVG but instead, increasing the size of the displayed masked element directly and adding the border to the same SVG instead. However, that should be done only in the display layer — the frame layer should stay the same so that the border would stay outside the selection box.

I’m not sure how complicated that would be and if it would influence some other parts as well, or if it would cause issues due to the element frame/display having different sizes, but if it starts seeming that making a separate border SVG match with the element mask is impossible, then it might be worth testing out quickly to see if that could be an option.

(Maybe you have a different alternative in mind as well.)

packages/story-editor/src/components/canvas/displayElement.js
barklund: * Done in https://github.com/google/web-stories-wp/issues/9865

packages/story-editor/src/masks/display.js
Trapsta: yeah, no difference really so moving to a util would eliminate some duplication

packages/story-editor/src/masks/output.js
Trapsta: Okay, gotcha

packages/story-editor/src/components/panels/design/border/borderWidth.js
Trapsta: Yeah, I agree with that, will make the variables names more meaningful and easy to understand. On it

packages/story-editor/src/utils/elementBorder.js
Trapsta: Yes, it could

Trapsta:
> There's one more issue just noticed: the border being partially below the element itself. E.g. here the element's background is gray with opacity 0.3, the black border can be seen partially under the element: Screenshot 2021-12-01 at 17 06 57

@miina , this is probably because am using double the border , I will tweak the calculations and see if that sorts this out. Haven't considered any alternatives really but I did come across this while working on this [https://css-tricks.com/exploring-the-css-paint-api-polygon-border/](https://css-tricks.com/exploring-the-css-paint-api-polygon-border/) but haven't really evaluated it as a possible solution yet.

Trapsta:
@miina , like I communicated on slack  there doesn't seem to be an all-round solution for implementing the border for all the non-rectangular shapes. Most solutions come a pixel or two off for some shapes especially ones with concave curves e.g. Heart and blob shapes. Using the paintApi (https://css-tricks.com/exploring-the-css-paint-api-polygon-border/) seems the far most promising but browser support is scanty (chrome > 65 and Edge). We might need to fragment and rethink our approach as per each shape or compromise on browser support.

miina:
@Trapsta What about trying an alternative solution like mentioned in https://github.com/google/web-stories-wp/pull/9851#issuecomment-983754583 -- to apply the border (stroke) to the original SVG directly instead of adding border as a separate element?

cc @barklund as well in case you have alternative ideas for implementing border for the shapes (the issue currently is that the path of the element doesn't really match with the inner path of the border, meaning there will be some space in case of some elements + when the element has semi-transparent color, the border can be partially seen under the element).

The heart shape had quite visible area between the border and the content the last time I tested + when assigning semi-transparent color to some shapes, the border can partially be seen under the shape. If this approach is not working then we should look for an alternative. We also support Safari and Firefox so probably https://css-tricks.com/exploring-the-css-paint-api-polygon-border/ is not an option unfortunately.

packages/story-editor/src/masks/output.js
```diff
@@ -77,16 +94,86 @@ export default function WithMask({
...
+    ? getBorderColor({ color: border.color })
+    : 'none';
+
   // @todo: Chrome cannot do inline clip-path using data: URLs.
   // See https://bugs.chromium.org/p/chromium/issues/detail?id=1041024.
 
   const maskId = `mask-${mask.type}-${element.id}-output`;
 
+  if (showSingleBorder) {
```
miina: Same question as before, should we show this at all if `borderColor === 'none'`?

Maybe something like this:
```
    const borderWidth = border?.left;
    const borderColor = border?.color ? getBorderColor({ color: border.color }) : undefined;
  const showSingleBorder = Boolean(
    !checkMultiBorderSupport(element) && borderWidth && borderColor
  );
```

packages/story-editor/src/masks/index.js
```diff
@@ -43,3 +48,34 @@ function getDefaultElementMask(type) {
   const { isMaskable } = getDefinitionForType(type);
   return isMaskable ? DEFAULT_MASK : null;
 }
+
+const BORDER_MULTIPLIER = 3;
+export function getBorderedMaskProperties(
```
miina, line 59: Would be great to leave a general comment somewhere explaining how this works.

There's a slight difference currently with "normal" borders -- the frame line is shown in the middle of the border as opposed to around the origin element. Probably this could be unified in #5325 where all the border will be included inside the frame for all elements?
Screenshot 2022-01-24 at 11 57 32

packages/story-editor/src/masks/index.js
```diff
@@ -32,6 +32,11 @@ export function canMaskHaveBorder(element) {
   return !mask || mask.supportsBorder;
 }
 
+export function checkMultiBorderSupport(element) {
```
miina: Nit: Would be nice to have this start with `can...` as well, e.g. `canHaveMultiBorder`. `check...` doesn't really say it will return a boolean, starting with `has...` or `can...` would be great.

packages/story-editor/src/masks/display.js
```diff
@@ -85,6 +114,68 @@ export default function WithMask({
     previewMode ? '-preview' : ''
   }-${randomId}`;
 
+  if (showSingleBorder) {
```
miina: If this is exactly the same as in `output.js`,  could be move it to a helper (DRY)?

miina (commented):
🎉 Omg, this seems to work! 🎉 🌊 

Some unit tests are failing now and would be good to have Karma added to the current border Karma, too.

There's actually one case that is not working currently (due to the border partially covering the element itself as well) -- when the element is small and border is huuuuge:
Screenshot 2022-01-24 at 12 07 19

This is quite an edge case so maybe something to confirm with UX -- e.g. if we should just leave it / limit the max border width, thoughts?

packages/story-editor/src/masks/index.js
barklund: Oh, the middle-aligned border (rather than the outer border) is a significant change, but it's the only way this could work. It changes workflow quite a bit for end-users and is definitely undesirable.

But there is no other way to do this, as it is way too hard to synchronize the two SVG's on top of each other if we want to "punch out" the shape of the element "inside" the border. It just didn't work reliably enough to be useful.

So this is what we have. It's this or nothing, and I feel that this is better.

miina: Ah, yes, definitely, just wanted to confirm mainly if this will still work for #5325 as well (where the frame would be moved to outside line of the border) -- to make the border clickable, too.

miina (approved):
Approving (pending the open comments) based on the discussion with product/UX it's ok to have the edge cases and follow up separately (not spending too much extra time though).

barklund:
This comment serves as documentation for the calculations in this PR (and will be linked in the code).

The border extends more than just half a border-width outside of the element. This is caused by pointy corners near the border. Such pointy corners will cause a miter, that can extend pretty far beyond the element. If the SVG was only half a border-width larger than the original element, the border miter of pointy corners near the edge would be cut off.

Instead, we extend the size of the SVG 3 times the necessary amount (to account for even very acute pointy corners near the edge). Compare the table below where the top right corner clearly shows, that half a border-width works fine for borders where the shape edge is tangential to frame, but not for borders where the shape is pointy towards the frame (like the bottom "tip" of the heart).

| SVG expansion | Plain | Outlined |
|-|-|-|
| 1 half border‑width | Screen Shot 2022-01-24 at 14 19 12 | Screen Shot 2022-01-24 at 14 19 25 |
| 3 half border‑widths | Screen Shot 2022-01-24 at 14 17 15 | Screen Shot 2022-01-24 at 14 18 44 |

packages/story-editor/src/masks/display.js
barklund: This was a tough refactor (so many things were similar but not exactly the same), but managed it in 1c2a42a8c

packages/story-editor/src/masks/index.js
barklund: Renamed to `canSupportMultiBorder()` in 9165f3f9e.

packages/story-editor/src/masks/index.js
barklund: Added documentation in 4193c5d1b.

packages/story-editor/src/masks/output.js
barklund: Simplified it in ce4980aea.

PR #9864, opened 2021-12-01
[WIP] Editor: Make the edit layer scrollable

### LINKED ISSUE #6180: Unable to scroll when zoomed while editing text or media
## Bug Description

Due to layer issues, the initial zoom implementation does not include the ability to scroll the page canvas if editing text or image.

This might be annoying if a user clicks to edit a text partially outside the screen. If this happens, the user would have to exit edit-mode, scroll and then re-enter edit mode. Ditto for images/videos.

## Expected Behaviour

* Scroll should work while editing text or media
* Adjust UX as needed for the menus / scale slider ([Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=10719%3A448782))

## Steps to Reproduce

* Set a zoom level where scroll is required to see the entire page
* Add a text element
* Click to edit the text element
* Try to scroll the page canvas using mouse or scrollbars and observe, that it's not possible.

## Screenshots
zoom-scroll-edit-bug

---

## Implementation Brief

* Fixing this requires allowing either `EditLayer` or `FrameLayer` to be the controlling layer, which will update the other 2 layers so when the edit layer is controlling, the frame layer is passive and vice versa. `DisplayLayer` will always be passive.
* This again probably requires changing the current updating of passive layers to actually scroll the content rather than just offset it using `left` and `top`.

barklund:
## Context

This is a working-draft attempt at making the edit layer scrollable. It is not complete yet.

## To-do

- [ ] Fix weird delayed-scroll bug
- [ ] Move media edit scale slider to tools area
- [ ] Display ghost image in a separate layer

## User-facing changes

_TBD_

## Testing Instructions

1. Add a text element to a page
2. Change zoom to 200%
3. Double-click text to enter edit mode
4. Attempt to scroll the canvas
5. Observe that the canvas does scroll even while in edit mode.

## Checklist

---

Fixes #6180

### DIFF
--- a/packages/story-editor/src/components/canvas/layout.js
+++ b/packages/story-editor/src/components/canvas/layout.js
@@ -139,15 +139,6 @@ const PageAreaContainer = styled(Area).attrs({
         100% - ${hasHorizontalOverflow ? themeHelpers.SCROLLBAR_WIDTH : 0}px
       );
     `}
-
-      overflow: ${({ showOverflow }) => (showOverflow ? 'visible' : 'hidden')};
-      width: calc(
-        100% - ${hasVerticalOverflow ? themeHelpers.SCROLLBAR_WIDTH : 0}px
-      );
-      height: calc(
-        100% - ${hasHorizontalOverflow ? themeHelpers.SCROLLBAR_WIDTH : 0}px
-      );
-    `}
 `;
 
 const Layer = forwardRef(function Layer({ children, ...rest }, ref) {
@@ -189,23 +180,6 @@ const PageClip = styled.div`
       justify-content: center;
       align-items: center;
     `}
-
-      overflow: hidden;
-      width: ${hasHorizontalOverflow
-        ? 'calc(var(--page-width-px) + var(--page-padding-px))'
-        : `calc(var(--viewport-width-px) - ${themeHelpers.SCROLLBAR_WIDTH}px)`};
-      flex-basis: ${hasHorizontalOverflow
-        ? 'calc(var(--page-width-px) + var(--page-padding-px))'
-        : `calc(var(--viewport-width-px) - ${themeHelpers.SCROLLBAR_WIDTH}px)`};
-      height: ${hasVerticalOverflow
-        ? 'calc(var(--fullbleed-height-px) + var(--page-padding-px))'
-        : `calc(var(--viewport-height-px) - ${themeHelpers.SCROLLBAR_WIDTH}px)`};
-      flex-shrink: 0;
-      flex-grow: 0;
-      display: flex;
-      justify-content: center;
-      align-items: center;
-    `}
 `;
 
 const FullbleedContainer = styled.div`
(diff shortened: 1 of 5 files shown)


packages/story-editor/src/components/canvas/layout.js
```diff
@@ -380,26 +379,78 @@ const PageArea = forwardRef(function PageArea(
...
+        lastScrollableLayer.current = null;
+        lastOnScroll.current = null;
+      }
+      if (node) {
+        // Manually attach listener so we get the event early
+        const onScroll = (evt) => {
+          const { scrollLeft: left, scrollTop: top } = evt.target;
+          displayLayer.style.left = `-${left}px`;
+          displayLayer.style.top = `-${top}px`;
+          setScrollOffset({ left, top });
```
miina, line 418: @barklund I was debugging this to try to understand what's happening here with the delay and thinking that perhaps it might be related to assigning `style` here directly while using the native scroll event for the edit layer?

Some things noticed:
1. Even when commenting out `scrollableLayer.scrollTop = scroll.current.top;` and `scrollableLayer.scrollTop = scrollTop;` and not changing anything else in the code, it still continues working as before (the lagging version). It seems like `scrollTop` is assigned to the `scrollableLayer` via native scroll event anyway. (Didn't test out with `scrollLeft`).
2. Initially I tried assigning the `top` value here directly to `scrollableLayer` to test it out without using `scrollTop` but since the `scrollTop` value seems to be assigned automatically anyway, then it doesn't work like it -- it got assigned double (top + scrollTop):
test

Also tried nulling `scrollTop` and tracking the scrolled  `top` manually, however, since it still can't be tested fully without the `scrollTop`, there's still a lag (different kind of lag though and almost not noticeable when scrolling slower than on this gif):
testing

Tried this for scrolling down only (and commented out assigning `scrollTop` below from both `useEffect`s):
```
const onScroll = (evt) => {
          evt.preventDefault();
          evt.stopPropagation();
          const { scrollLeft: left, scrollTop: top } = evt.target;
          if (!lastScrollTopRef.current) {
            lastScrollTopRef.current = top;
          } else {
            lastScrollTopRef.current = lastScrollTopRef.current + top;
          }
          displayLayer.style.left = `-${left}px`;
          displayLayer.style.top = `-${lastScrollTopRef.current}px`;
          if (scrollableLayer && !isControlled && isScrollable) {
            scrollableLayer.scrollTop = 0;
            scrollableLayer.style.top = `-${lastScrollTopRef.current}px`;
          }
          setScrollOffset({ left, top });
          return false;
        };
```

Wondering if there's a way to unify the layers more to apply either style or `scrollTop` to both instead of using native `scrollTop` on the edit layer and `style` on the display?

barklund:
Neither this approach nor any other seemed to work out all that well.

PR #9866, opened 2021-12-01
Bugfix: Prevent Shape Layer Icons getting cut off in Firefox

### LINKED ISSUE #9704: Firefox: Layers Panel Icons Cut Off
Found this while fixing the circle icon to not be cut off (#9559). In firefox the shapes are overflowing a bit in their box and the clip paths are getting cut off. Only in Firefox. Only in the layers panel, not actually impacting usage or the shapes panel.

BrittanyIRL:
## Context

In the layers panel the shapes icons were getting cut off in Firefox (and a little bit in Chrome). 

## Summary

Tweak the shapes layer icon component to make sure it isn't cut off in Firefox and Chrome.

## Relevant Technical Choices

Ended up solving this by adjusting the clip-path `transform: scale` to shrink it just slightly .9 - the issue was that we were using the maskDef.ratio for the scale here when a) sometimes the ratio would boost it over 1 so the clip path would then grow to be too big and the overflow (as is the nature of clip paths) is just cut off and b) the icon size is contained anyways. So, that in tandem with just allowing the div that has the clip path applied to be 100% of the container (always 21px) gives the paths room to render their whole path without impacting layout else where. 

## To-do

`clip-path` isn't well supported in Safari, that's why it renders like this:
Screen Shot 2021-12-01 at 4 08 54 PM
https://stackoverflow.com/questions/41860477/why-doesnt-css-clip-path-with-svg-work-in-safari
https://caniuse.com/css-clip-path 
I moved the `clip-path` from inline to the `ShapePreview` styled component so that it could get auto prefixed with webkit - so when that's fixed in Safari it should "just work" now.  The only other way I thought of around this is to not use clip path for this icon anymore but then that's a whole thing with gradients and it doesn't seem worth it. 

## User-facing changes

| | Chrome | Firefox | Safari | 
| --- | --- | --- | --- |
| Before | Screen Shot 2021-12-01 at 4 14 33 PM | Screen Shot 2021-12-01 at 4 15 04 PM | Screen Shot 2021-12-01 at 4 15 31 PM |
| After |  Screen Shot 2021-12-01 at 4 12 19 PM | Screen Shot 2021-12-01 at 4 12 50 PM | Screen Shot 2021-12-01 at 4 13 19 PM |

## Testing Instructions

1.

No

No

No

## Checklist

---

Fixes #9704

### DIFF
--- a/packages/story-editor/src/components/panels/design/layer/layer.js
+++ b/packages/story-editor/src/components/panels/design/layer/layer.js
@@ -142,6 +142,37 @@ const LayerButton = styled(Button).attrs({
     --background-color-opaque: ${({ theme }) =>
       rgba(theme.colors.interactiveBg.secondaryPress, 0)};
   }
+
+      background: ${theme.colors.interactiveBg.secondaryPress};
+      + * {
+        --background-color: ${theme.colors.interactiveBg.secondaryPress};
+        --background-color-opaque: ${rgba(
+          theme.colors.interactiveBg.secondaryPress,
+          0
+        )};
+        --selected-hover-color: ${theme.colors.interactiveBg.tertiaryHover};
+      }
+    `}
+
+  :hover {
+    background: ${({ theme }) => theme.colors.interactiveBg.secondaryHover};
+  }
+  :hover + * {
+    --background-color: ${({ theme }) =>
+      theme.colors.interactiveBg.secondaryHover};
+    --background-color-opaque: ${({ theme }) =>
+      rgba(theme.colors.interactiveBg.secondaryHover, 0)};
+  }
+
+  :active {
+    background: ${({ theme }) => theme.colors.interactiveBg.secondaryPress};
+  }
+  :active + * {
+    --background-color: ${({ theme }) =>
+      theme.colors.interactiveBg.secondaryPress};
+    --background-color-opaque: ${({ theme }) =>
+      rgba(theme.colors.interactiveBg.secondaryPress, 0)};
+  }
 `;
 
 const LayerIconWrapper = styled.div`
@@ -198,7 +229,7 @@ const LayerAction = styled(Button).attrs({
   /*
    * all of our Icons right now have an embedded padding,
    * however the new designs just disregard this embedded
-   * padding, so to accomodate, we'll make the icon its
+   * padding, so to accommodate, we'll make the icon its
    * intended size and manually center it within the button.
    */
   svg {
--- a/packages/story-editor/src/components/panels/design/layer/layer.js
+++ b/packages/story-editor/src/components/panels/design/layer/layer.js
@@ -142,37 +142,6 @@ const LayerButton = styled(Button).attrs({
     --background-color-opaque: ${({ theme }) =>
       rgba(theme.colors.interactiveBg.secondaryPress, 0)};
   }
-
-      background: ${theme.colors.interactiveBg.secondaryPress};
-      + * {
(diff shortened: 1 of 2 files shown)


maxyinger (approved):
Awesome work. great finds here

packages/story-editor/src/components/panels/design/layer/layer.js
```diff
@@ -198,7 +198,7 @@ const LayerAction = styled(Button).attrs({
   /*
    * all of our Icons right now have an embedded padding,
    * however the new designs just disregard this embedded
-   * padding, so to accomodate, we'll make the icon its
+   * padding, so to accommodate, we'll make the icon its
```
samwhale, line 201: noice

packages/story-editor/src/elements/shape/icon.js
```diff
@@ -40,11 +38,21 @@ const Container = styled.div`
...
 `;
 
+/* 
+clip-path isn't stable in Safari yet, so these 
+Shape Layer Icons are going to show up as various quadrilateral 
+until that stabilizes for inline SVGs as clip paths.
+More info here:  
+https://stackoverflow.com/questions/41860477/why-doesnt-css-clip-path-with-svg-work-in-safari
+https://caniuse.com/css-clip-path
+*/
```
samwhale, line 56: Helpful comment

packages/story-editor/src/elements/shape/icon.js
```diff
@@ -55,21 +63,18 @@ function ShapeLayerIcon({
...
+      <ShapePreview maskId={maskId} backgroundColor={backgroundColor}>
         <svg width={0} height={0}>
           <defs>
             <clipPath
               id={maskId}
-              transform={`scale(1 ${maskDef.ratio})`}
+              // Bring the path scale down a bit from 1
+              // so that we can make sure the entire SVG path is visible when the mask ratio is > 1
+              // this is important for Firefox's interpretation of clip paths
+              transform={`scale(1 0.9)`}
```
samwhale: Don't think we need string interpolation here?
```suggestion
              transform="scale(1 0.9)"
```

samwhale (approved):
The screenshots are very helpful and this works great!

Probably not part of this ticket, but:
1. A few shapes seem to not be vertically centered? 
image
image
image

2. Seems like some shapes overlap their containing box a little
image
image

BrittanyIRL:
> 1. A few shapes seem to not be vertically centered?

Indeed, and this bothers me but I cannot change it. This has to do with the path of the SVG and where it's lined up. If we weren't using clip paths we could use a viewBox instead and position them in the top right corner consistently and then center them from there but because of the clip path this is as good as we're gonna get. Some of the paths have more white space than others. 

> 2. Seems like some shapes overlap their containing box a little

Great eye, i'll tinker with this a little bit, should be able to control this one.

packages/story-editor/src/elements/shape/icon.js
BrittanyIRL: yup! tried this out and it worked and i didn't clean up. on it! 👍

BrookeGraham: Oh, gtk!

BrookeGraham (approved):
This is so smart! Thanks for figuring it out!

BrittanyIRL:
> This is so smart! Thanks for figuring it out!

this was tricky, thanks for letting me grab it from you.

PR #9867, opened 2021-12-02
Media: Re-use existing `isMuted` value on upload

spacedmonkey:
## Summary

We generate `isMuted` property for local assets, we need it for the editor. So on upload, if we know this value, save it, so that the backfill is not needed in post cases. 

## Testing Instructions

1.

## Checklist

---

Fixes #


packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -321,6 +321,10 @@ function useMediaUploadQueue() {
 
           const trackTiming = getTimeTracker('load_upload_media');
 
+          if (resource.type !== 'image' && resource.isMuted !== null) {
```
swissspidy: Why not `resource.type === 'video'`?

We have other types like `gif` too (those are always muted).

spacedmonkey: Yeah and we are setting isMuted on gifs. That is what we want no?

swissspidy: GIFs are always considered to be muted. We don't need to set `isMuted` on GIFs.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -321,6 +321,10 @@ function useMediaUploadQueue() {
 
           const trackTiming = getTimeTracker('load_upload_media');
 
+          if (resource.type === 'video' && resource.isMuted !== null) {
+            additionalData.web_stories_is_muted = resource.isMuted;
```
swissspidy: `additionalData.web_stories_is_muted` might have already been set above (e.g. on line  284, when muting).

Won't this override that?

`resource` at this point is the original resource added to the queue, which means it's not yet muted. So seems like this would break things.

To prevent this, seems like this code should be moved to line ~232

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -281,7 +289,6 @@ function useMediaUploadQueue() {
               try {
                 newFile = await stripAudioFromVideo(file);
                 finishMuting({ id, file: newFile });
-                additionalData.web_stories_is_muted = true;
```
swissspidy: Why do we want to remove this?

swissspidy (approved):

PR #9868, opened 2021-12-02
Media: Measure time to generate base color

### LINKED ISSUE #9739: Media: Generate base color on upload

spacedmonkey:
## Summary

When generating local resource for upload, also generate base color using blob. Then on upload, use the generated base color along with the file. This saves on the need for backfill and it also means that generation is faster as the blob is faster and loading in a image for a url. 

## Relevant Technical Choices

Add some time trackers to aid testing. 

## Testing Instructions

1.

## Checklist

---

Fixes #9739

### DIFF
--- a/packages/story-editor/src/app/media/utils/getResourceFromLocalFile.js
+++ b/packages/story-editor/src/app/media/utils/getResourceFromLocalFile.js
@@ -31,6 +31,10 @@ import {
   seekVideo,
   preloadVideo,
 } from '@web-stories-wp/media';
+/**
+ * Internal dependencies
+ */
+import { getMediaBaseColor } from '../../../utils/getMediaBaseColor';
 
 /**
  * Create a local resource object.
@@ -153,6 +157,14 @@ const getResourceFromLocalFile = async (file) => {
   } catch {
     // Not interested in the error here.
     // We simply fall back to the placeholder resource.
+  }
+  const { src, poster } = resource;
+  const imageSrc = type === 'image' ? src : poster;
+  try {
+    const color = await getMediaBaseColor(imageSrc);
+    resource.baseColor = color;
+  } catch {
+    // Do nothing for now
   }
 
   return { resource, posterFile };
--- a/packages/story-editor/src/app/media/utils/getResourceFromLocalFile.js
+++ b/packages/story-editor/src/app/media/utils/getResourceFromLocalFile.js
@@ -31,10 +31,6 @@ import {
   seekVideo,
   preloadVideo,
 } from '@web-stories-wp/media';
-/**
- * Internal dependencies
- */
-import { getMediaBaseColor } from '../../../utils/getMediaBaseColor';
 
 /**
  * Create a local resource object.
@@ -157,14 +153,6 @@ const getResourceFromLocalFile = async (file) => {
   } catch {
     // Not interested in the error here.
     // We simply fall back to the placeholder resource.
-  }
-  const { src, poster } = resource;
-  const imageSrc = type === 'image' ? src : poster;
-  try {
-    const color = await getMediaBaseColor(imageSrc);
-    resource.baseColor = color;
-  } catch {
-    // Do nothing for now
   }
 
   return { resource, posterFile };
--- a/packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
+++ b/packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
@@ -321,6 +321,13 @@ function useMediaUploadQueue() {
 
           const trackTiming = getTimeTracker('load_upload_media');
 
+          if (resource?.baseColor) {
+            additionalData.meta = {
+              ...additionalData.meta,
+              web_stories_base_color: resource.baseColor,
+            };
+          }
+
           try {
             // The newly uploaded file won't have a poster yet.
             // However, we'll likely still have one on file.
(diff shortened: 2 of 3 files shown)


packages/story-editor/src/app/media/utils/getResourceFromLocalFile.js
```diff
@@ -154,6 +158,14 @@ const getResourceFromLocalFile = async (file) => {
     // Not interested in the error here.
     // We simply fall back to the placeholder resource.
   }
+  const { src, poster } = resource;
+  const imageSrc = type === 'image' ? src : poster;
+  try {
+    const color = await getMediaBaseColor(imageSrc);
```
swissspidy: We should use `getTimeTracker` in `getMediaBaseColor` to measure how long it takes to get the color of an image.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -321,6 +321,13 @@ function useMediaUploadQueue() {
 
           const trackTiming = getTimeTracker('load_upload_media');
 
+          if (resource?.baseColor) {
+            additionalData.meta = {
+              ...additionalData.meta,
+              web_stories_base_color: resource.baseColor,
+            };
```
swissspidy, line 328:
```suggestion
            additionalData.meta.web_stories_base_color = resource.baseColor;
```

spacedmonkey: No, can't do this, as meta is always set / an object.

packages/story-editor/src/app/media/utils/getResourceFromLocalFile.js
spacedmonkey: I did some testing and results were very interesting!. 

Generated base color from blob. 
Image - 177.85000002384186
Video - 428.8799999952316

Generated base color from image. 
Image - 1494.105 / 1360.30
Video - 652.7250000238419

Worth noting, that used the same image / video for both. I could do more testing, but results are positive.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: Could theoretically do

```suggestion
						additionalData.meta = additionalData.meta || {};
            additionalData.meta.web_stories_base_color = resource.baseColor;
```

but both work

packages/story-editor/src/app/media/utils/getResourceFromLocalFile.js
swissspidy: That's in milliseconds I presume?

Can you add the `getTimeTracker` telemetry in `getMediaBaseColor` in this PR?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
spacedmonkey: Honestly don't mind either way, spread operator made sense to me and left very readable, but willing to change it.

packages/story-editor/src/app/media/utils/getResourceFromLocalFile.js
spacedmonkey: Sure!

spacedmonkey: I add time tracking at the base color generation. I saw one example, that was 33ms compared to 532ms on the same video. So much better!

spacedmonkey:
i am wondering if we should not do the base color generation out of local resource geneation. In my testing, inserting a large image, like 30MB image, it takes long to insert the local resource in.

spacedmonkey:
I think we should put this on hold. base color can be slow and SHOULD be uploaded async after the fact.

swissspidy:
> I think we should put this on hold.

I'd only revert the change to `getResourceFromLocalFile.js`

We can keep the changes to `useMediaUploadQueue/index.js` and `getMediaBaseColor.js`

> base color can be slow and SHOULD be uploaded async after the fact.

FWIW, we already do that, since `postProcessingResource` is called after upload.

spacedmonkey:
> I'd only revert the change to getResourceFromLocalFile.js

If I do that, then when would base color be set? 

> We can keep the changes to useMediaUploadQueue/index.js and getMediaBaseColor.js

Those changes are valid, so let's keep them. 

> FWIW, we already do that, since postProcessingResource is called after upload.

We do, which is how we should keep it.

swissspidy:
> If I do that, then when would base color be set?

_After upload_, in `postProcessingResource`. Which is already happening.

spacedmonkey:
> After upload, in postProcessingResource. Which is already happening.

This change, is designed to force base color to be generated on upload, by adding it to local resource generation. That means that baseColor is set on the resource object in the media upload queue. If we don't set the base color, then these lines, 

https://github.com/google/web-stories-wp/blob/b70c764588dcd8caa51fc5ed4183db13bc24e3df/packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js#L324-L329

Would do nothing, as the likelyhood of base color being set before upload is basically none.

swissspidy:
Right, then let's also undo the change to `packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js` then.

We can merge the time tracking only.

packages/story-editor/src/app/media/utils/getResourceFromLocalFile.js
spacedmonkey: Removed after feedback.

swissspidy (approved):

PR #9869, opened 2021-12-02
Storybook: Fix broken build

mohdsayed:
## Context

Storybook build broke recently ( https://google.github.io/web-stories-wp/storybook/ ).

## Testing Instructions

1.

## Checklist

---

Fixes #


packages/story-editor/src/components/panels/shared/flipControls.js
```diff
@@ -32,7 +32,7 @@ import {
  * Internal dependencies
  */
 import Tooltip from '../../tooltip';
-import { focusStyle } from './styles';
+import { focusStyle } from '.';
```
swissspidy: This now seems to cause a cycle, as it looks for `focusStyle` in `index.js`, and `index.js` also references `flipControls.js`

mohdsayed: Yes, noticed that. 

Storybook build was working when I https://github.com/google/web-stories-wp/pull/9569 got merged. Don't know what changed after that it stopped working, trying to figure out. 🤔

mohdsayed: Got another solution. If I import `styles` on top, the build doesn't throw that error. There seems to be some bug in storybook which I couldn't find out.

mohdsayed: `linkRelations.js` is exported from `shared/index.js` before `./styles.js` was exported.

Now `linkRelations.js` imports `focusStyle` from `'../../panels/shared'` from `components/form/index.js` => `components/form/color/index.js` here.

https://github.com/google/web-stories-wp/blob/c210a3c9307a224ed800551c04df93e3b066a0ac/packages/story-editor/src/components/form/color/color.js#L39

So maybe storybook has a problem because `color.js` is importing `focusStyle` before it is exported. Though the order of export shouldn't be an issue but there could be some bug in storybook.

swissspidy: Great investigation!

swissspidy (approved):

PR #9873, opened 2021-12-02
Text element display: add missing units for padding in Text Display

### LINKED ISSUE #9872: Highlight & Fill text editor option issues
## Bug Description

A user has pointed out in the support forums that there are issues with Highlight & Fill text editor options since updating to version 1.15.0. 

**Highlight option:**
This now renders differently since the update. See the screenshots of before 1.15.0 update and after the update from the user.
**Before:** https://onedrive.live.com/?authkey=%21AGJXeukq7slyiN0&cid=F3781633E408D2AC&id=F3781633E408D2AC%21155802&parId=F3781633E408D2AC%21155587&o=OneUp
**After:** https://onedrive.live.com/?authkey=%21AJv0NnjXVKZW%2DTk&cid=F3781633E408D2AC&id=F3781633E408D2AC%21155803&parId=F3781633E408D2AC%21155587&o=OneUp

I have been able to recreate this using both versions also. See results of both here: 
image

The editor panel in the middle is larger since the update so not sure if this is related to that but it's clear that these are not consistent. Note that this issue only shows in the editor and once published looks as expected. 

**Fill option:**
When using the fill option for text, if you select a padding of 10px it used to put 10px of padding all around the text, now the editor sticks the text to the top of the box, as if padding-top was set to 0. This again is only while in the editor and is as expected once published.

It looks like any Fill textbox actually loses any bottom and top padding when double-clicking to edit.
So every time you edit a textbox with padding, you have to go and change the padding in the editor to revert.

## Expected Behaviour

Highlight should show as shown in the previous version screenshots and the fill option should apply the padding when selected.

## Steps to Reproduce

**Highlight option:**
- Add some text to your story. It helps to add a few rows like a list.
- From the design panel select the highlight option.

**Fill option:**
– Create a text box with Fill mode
– Add a padding of 10px
– Add some text
– Double Click the textbox to edit the text
=> All vertical padding is gone (but it still shows 10 in the settings)
– Change to 9px and then to 10px
=> It goes back to “normal” (even though the text has a weird top align instead of being centred)

## Screenshots

Provided above.

## Additional Context

**Related Support topic:**
https://wordpress.org/support/topic/highlight-text-box-style-issue-in-1-15/

- Plugin Version: 1.15.0

miina:
## Summary
Ensures that when using `generateParagraphTextStyle`, the unit is always added to the padding. By default, the util does not add any units and it was missing from displaying the text in the editor, causing display issues.

## User-facing changes
Padding is displaying correctly now:
Screenshot 2021-12-02 at 15 12 57

## Testing Instructions

1. Add a text element
2. Set Fill as the background mode and set padding (e.g. 15)
3. Verify that padding is added correctly to each side.
4. Test with Highlight mode as well, verify there is padding between the text and the highlighted background

No

No

No

## Checklist

---

Fixes #9872


swissspidy:
Why didn't existing tests, like the `generateParagraphTextStyle` ones for example, not catch this? 🤔 

We should add a new one to prevent regressions in the future.

miina:
> Why didn't existing tests, like the generateParagraphTextStyle ones for example, not catch this?

Because in the tests, a different `dataToPadding` is used compared to what the display uses -- it returns `%` as the unit by default whereas in case of display `dataToEditor` is used which does not have units appended by default.

Hm, now the current tests are failing actually because both `%` and `px` are appended -- `%` from the custom `dataTo*` in the tests and `px` added by default.

It's a bit confusing, the system currently... It should either never add units or always add units. That's why the padding was probably overridden for the output as well -- not to use the default `px` as it previously was.

Anyway -- let me check if I can change it generally so that it would never add the units by default and only take it from the specific usage.

barklund (approved):
Looks good. Weird that we never picked up on this before.

miina:
Unrelated Karma failures:
- `CUJ: Page Templates: Creator can Apply a Default Page Template Keyboard Navigation should navigate templates via keyboard` [Example run](https://github.com/google/web-stories-wp/runs/4430822814?check_suite_focus=true#step:6:8313)

miina:
Unrelated Karma failures:
- `CUJ: Page Templates: Creator can Apply a Default Page Template Keyboard Navigation should navigate templates via keyboard` [Example run](https://github.com/google/web-stories-wp/runs/4430822814?check_suite_focus=true#step:6:8313)

PR #9882, opened 2021-12-03
Media Panel: Reduce re-rendering of `InnerElement`

### LINKED ISSUE #9580: Performance: reduce re-rendering in Media panel
## Feature Description
Currently, when doing anything on canvas, there is excess re-rendering happening in the Media Library panel, at least the `LibraryMoveable` and `InnerElement` components re-render hundreds of times while dragging for example.

Try either by logging out something in the `InnerElement` or checking the "Highlight updates when components render." checkbox in the React dev tools, and then dragging an element on canvas.

There is no reason these should re-render while doing something unrelated on canvas.

We should detect what's causing it and avoid unnecessary rendering.

See also discussion in https://github.com/google/web-stories-wp/issues/9549#issuecomment-953732053 for context.

miina:
## Summary
Reduces re-rendering elements in the media panel.

## Relevant Technical Choices
- Removes `draggingResource` dependency and uses a local ref for tracking if the resource has been set to be dragged instead.
- Uses conditional update for `handleDrag` and `handleDrop` using the equality function in the `useContextSelector` to only update `handleDrag` and `handleDrop` if these actually needs an update.
- Adds `useCallback` to functions that other components depend on to reduce re-rendering

## To-do

- [x] The `InnerElement` is still re-rendering due to `onClick` and `onLoad` changing every time when doing something on the page.

## User-facing changes
N/A

## Testing Instructions

1. Log out something in the `InnerElement` in the media panel on `main` and on this PR (e.g on different tabs)
2. Add a Media element. Now click arrow key to move that media element.
3. Compare the amount of re-renders between the two for moving the media element via arrow keys.

In my testing this is on `main` (depends on how much media you have, too!):
Screenshot 2021-12-03 at 17 06 07

And in my testing no re-renders happened on this PR for clicking an arrow key to move the element on the Page.

No

No

No

## Checklist

---

Fixes #9580


packages/story-editor/src/utils/useCORSProxy.js
```diff
@@ -42,7 +42,7 @@ function useCORSProxy() {
    *
    * @type {function(): boolean}
    */
-  async function checkResourceAccess(link) {
+  const checkResourceAccess = useCallback(async (link) => {
```
miina, line 45: @swissspidy Any concerns with this change here?

Apparently this was a dependency for `generateBaseColor` which is a dep for `insertElement` which caused `onInsert` to be updated every time when something happened which also caused the `InnerElement` etc. re-rendering all the time.

packages/story-editor/src/components/library/panes/media/common/innerElement.js
```diff
@@ -109,11 +110,35 @@ function InnerElement({
   isMuted,
 }) {
   const newVideoPosterRef = useRef(null);
+  // Track if we have already set the dragging resource.
+  const hasSetResourceTracker = useRef(null);
+
+  // Note: This `useDropTargets` is purposefully separated for the one below since it
```
miina:
```suggestion
  // Note: This `useDropTargets` is purposefully separated from the one below since it
```

packages/story-editor/src/utils/useCORSProxy.js
swissspidy: No concerns.

I'm actually looking into removing `generateBaseColor` from `insertElement` anyway.

spacedmonkey: >I'm actually looking into removing generateBaseColor from insertElement anyway. 

Are you? Why, it is need for third party resources that don't have base color.

swissspidy: @spacedmonkey I'll share more information soon. **tl;dr:** `useInsertElement` is the wrong place to do this.

Trapsta (approved):
Tested and  takes care of those re-renders nicely. Do you think we could use the same strategy for `onClick` and `onLoad`? 

| Before | After |
|--------|-------|
|   Screenshot 2021-12-06 at 09 26 24     |  Screenshot 2021-12-06 at 09 19 57     |

miina:
> Tested and takes care of those re-renders nicely. Do you think we could use the same strategy for onClick and onLoad?

Probably when clicking or loading the media element, it doesn't harm updating all the necessary hooks for dragging / drop-targeting. The problem was that when doing anything on canvas, the components kept re-rendering (now it should only update those the first time or when interacting with a specific element).
Or did you mean something else? :)

packages/story-editor/src/components/library/panes/media/common/innerElement.js
```diff
@@ -109,11 +110,35 @@ function InnerElement({
...
+      if (hasSetResourceTracker.current) {
+        return false;
+        // If we're rendering the first time, init `handleDrag` and `handleDrop`.
+      } else if (hasSetResourceTracker.current === null) {
+        hasSetResourceTracker.current = false;
+        return false;
+      }
+      // Otherwise ignore the changes in the actions.
+      return true;
+    }
```
barklund, line 134: Uh, that's mighty fancy. We should probably use similar tricks elsewhere. I hadn't even considered using the equality function for this purpose... could definitely save some re-renders!

barklund (approved):
Excellent work here - great optimization.

I know this is just _one_ element and it takes quite a lot of time to do, but it's a start! And once we get going, we learn some new tricks so we can do this better the first time around for new functionality.

packages/story-editor/src/utils/useCORSProxy.js
swissspidy: See #9907

swissspidy (approved):

miina:
Hm, some of the Karma tests that are failing are also failing locally on `main` for me, can't see those failing on other recent PRs though (`Background Copy Paste integration`) 🤔 

Will look into it tomorrow to see if it's a real issue on this PR, although odd that it fails on `main`, too, for me.

Edit: looks like it's failing locally for others, too, see [Slack](https://xwp.slack.com/archives/C0135V8LYJE/p1638810688060000?thread_ts=1638810313.059800&cid=C0135V8LYJE)

miina:
Ok, this PR is now skipping a bunch of tests.

I checked some recent PRs that passed the tests previously and failed with the same tests after merging to `main` to ensure that those were really unrelated tests to this PR and generally failing, too. (Passing locally flakily as well, except for some that always seem to fail on `main`.)

PR #9886, opened 2021-12-03
Saved Templates: Fix keyboard shortcut conflict

### LINKED ISSUE #9884: Saved Template Wrong Keyboard Shortcut
## Bug Description

When a saved template is in focus, pressing on `Space` key does multiple jobs, it opens the delete modal and fires scroll event at the same time. It should do one job at a time (either it should scroll or handle the delete modal).

IMO pressing on `Delete/Backspace` makes more sense to open the delete modal since the entire template gets focused and pressing on `Space` should handle the scroll.

## Expected Behaviour

Pressing on `Space` button shouldn't fire multiple events at a time.

## Steps to Reproduce

1. Go to Saved templates
2. Create 10+ templates
3. Focus on a template using the keyboard.
4. Press on `Space` key

## Screenshots

https://user-images.githubusercontent.com/17360543/144630866-132a9292-0d19-4eaa-bf29-595a8cbfd04d.mov

## Additional Context

- WordPress Version: 5.8.2
- Browser: Chrome

delowardev:
## Context

When a saved template is in focus, pressing on `Space` key opens the delete modal and fires scroll event at the same time.

But in this context `Backspace/Delete` makes more sense for opening the delete modal since entire templates get focused.

See: #9884 

## Summary

Updated the keyboard shortcut for saved template delete modal.

## Testing Instructions

1.

## Checklist

---

Fixes #9884

### DIFF
--- a/packages/story-editor/src/components/library/panes/pageTemplates/karma/savedPageTemplates.karma.js
+++ b/packages/story-editor/src/components/library/panes/pageTemplates/karma/savedPageTemplates.karma.js
@@ -161,7 +161,7 @@ describe('CUJ: Page Templates: Custom Saved Templates', () => {
       // navigate to newly saved template and open delete dialog
       await fixture.events.keyboard.press('Tab');
       await fixture.events.keyboard.press('Tab');
-      await fixture.events.keyboard.press('Space');
+      await fixture.events.keyboard.press('Backspace');
 
       await fixture.events.sleep(200);
 
--- a/packages/story-editor/src/components/library/panes/pageTemplates/templateList.js
+++ b/packages/story-editor/src/components/library/panes/pageTemplates/templateList.js
@@ -125,7 +125,7 @@ function TemplateList({
       if (isGridFocused) {
         if (code === 'Enter') {
           handlePageClick(page);
-        } else if (code === 'Space') {
+        } else if (code === 'Backspace' || code === 'Delete') {
           handleDelete?.(page);
         }
       }


delowardev:
`E2E Tests / chrome - WP latest` is failing everywhere.

BrittanyIRL (dismissed):
Thanks for the fix! Looks good, we'll grab 1 more review and then run it through QA.

miina:
Q: How is it working in the Media library? Does the same issue happen there as well? The reason for using `Space` for the button in the templates was consistency -- in Media Library, using `Space` also clicks on the icon on the right corner (which in that case opens the menu). If Space scrolls in the Saved Templates, should it do the same in the Media Library as well instead of opening the menu? Thoughts? cc @swissspidy as this is a question about Media.

BrittanyIRL:
I was thinking about this over the weekend, this is a break from consistency with the other keyboard actions in the other panels - we should be consistent. I do like this update, it makes sense. But we should discuss what we want before issuing any immediate fixes to specific panels.

BrittanyIRL:
In staging, local media and saved templates have the same panel patterning for the edit/delete functionality available on click or spacebar. Looks like this: 

| Local Media | Saved Templates | 
| --- | --- | 
| spacebar opens dialog | spacebar opens dialog and scrolls | 
| delete/backspace does nothing |  delete/backspace does nothing | 


I think there's 2 options here: 
1 ) Follow the proposed solution in this PR -  use delete/backspace to open the dialog and let spacebar scroll the view. This would necessitate a mirroring update to the local media panel so it's the same functionality. 
2) don't let the saved templates scroll on spacebar and leave the keyboard shortcuts as they are. 

@miina thoughts?

delowardev:
In the local media panel, pressing on `Backspace/Delete` does nothing. Maybe add `Backspace/Delete` support in the media panel? (and wherever applicable) 

So, it would be something like this:
- `Space` key will open the dropdown if there's any ( in saved templates, there's no dropdown, so the space key will do nothing)
- `Backspace/Delete` key will open the delete modal if it's removable

> Observation: in some places like **Default Template Panel**, the `Space` key acts like the `Enter` key.

miina:
> Space key will open the dropdown if there's any ( in saved templates, there's no dropdown, so the space key will do nothing)

`Space` is currently for interacting with any button in the corner, and not specifically for opening the dropdown, so having `Space` click on the button for Media Panel but doing nothing in the Saved Templates still sounds inconsistent.

> 2) don't let the saved templates scroll on spacebar and leave the keyboard shortcuts as they are.

I'd actually go this way for now (or hold off a bit) and maybe ask @gajasinski / @agingoldseco to review the current behaviour generally across the panels and interacting with the panels using keyboard to see which changes make sense, perhaps they have more insights to what's the standard.
Adding `Backspace` to the Media Panel items as a quicker option to delete an item in addition to allowing opening the menu with `Space` could be a reasonable option but would be great to have a confirmation from the Product team for that first. Probably would need a confirmation modal as well then. Thoughts?

BrittanyIRL:
Great ideas, @miina - thanks for pushing the breaks a little - things like this can feel so simple but it's so important to pause and think about the bigger picture. 

> 2\. don't let the saved templates scroll on spacebar and leave the keyboard shortcuts as they are.

I say we go with the above, second option, just because it is jarring when you hit spacebar and open a dialog AND scroll past the template you selected. I don't think now is the right time to really shake up how our keyboard shortcuts work and it feels like this gets to the base of the issue that prompted this bug ticket. 

Agree?

miina:
> Agree?

SGTM!

BrittanyIRL:
closing as there's been no movement on this in over 2 weeks

PR #9887, opened 2021-12-03
Design System: add `onClear` to search input

### LINKED ISSUE #9667: Design System: Update Search component event props
Currently the Search component from the design system calls `onMenuItemClick` with empty string values when you click the clear button. This makes it a little opaque when you're trying to pass in an `onClear` handler:
https://github.com/google/web-stories-wp/pull/9648#discussion_r745146704

This ticket is just to add an `onClear` handler to the search component and use that when the user presses the x button on search instead of calling `onMenuItemClick` to make this component's interface a little more transparent:
https://github.com/google/web-stories-wp/blob/a169d4c8adee9822d6ea9ab370ca3fcb98c2d38a/packages/design-system/src/components/search/search.js#L179-L183

BrookeGraham:
## Context

Currently the Search component from the design system calls `onMenuItemClick` with empty string values when you click the clear button. This makes it a little opaque when you're trying to pass in an `onClear` handler.

## Summary

Separating out on `onClear` handler to call when in the search component's `handleClearInput`. 

## Testing Instructions

1. Type into a search input (on explore templates or my templates) and make you can still clear a search entry. 

no

no

no

## Checklist

---

Fixes #9667

### DIFF
--- a/packages/design-system/src/components/search/search.js
+++ b/packages/design-system/src/components/search/search.js
@@ -61,6 +61,7 @@ const SearchInputWrapper = styled.div``;
  * @param {string} props.label If present, will display a text label above input. ariaLabel controls accessibility label of input since there are times search components are present without label text visible.
  * @param {Object} props.menuStylesOverride should be formatted as a css template literal with styled components. Gives access to completely overriding menu styles (container div > ul > li).
  * @param {Function} props.onMenuItemClick Triggered when a user clicks or presses 'Enter' on an option, or 'Enter' is pressed on input. Returns an object containing label and value keys.
+ * @param {Function} props.onClear Triggered when a user clicks clear 'x' button on the search input.
  * @param {Array} props.options All options, should contain either 1) objects with a label, value, anything else you need can be added and accessed through renderItem or 2) Objects containing a label and options, where options is structured as first option with array of objects containing at least value and label - this will create a nested list.
  * @param {string} props.placeholder Placeholder text to display in input if no value is present.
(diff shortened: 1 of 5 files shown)


packages/design-system/src/components/search/search.js
```diff
@@ -178,9 +180,9 @@ export const Search = ({
 
   const handleClearInput = useCallback(() => {
     inputState.set(undefined);
-    onMenuItemClick?.(null, { label: '', value: '' });
+    onClear();
```
maxyinger, line 183: If removing this happens to break anything else consuming this, I'm fine leaving it in with the `onClear` event ie.
```
onMenuItemClick?.(null, { label: '', value: '' });
onClear();
```

maxyinger (approved):
looks great. Thanks for the cleanup 🎉

packages/design-system/src/components/search/search.js
BrookeGraham: I checked all the search inputs I could find. Seemed to not break anything. And I didn't get any failing tests, so 🤷

samwhale (approved):
Nice makes sense to me and works well!

packages/design-system/src/components/search/search.js
BrittanyIRL: if memory serves this was added because of edge cases in Storybook, so just make sure you also update the Search storybook components so that doesn't break

BrookeGraham: oh good call.

BrittanyIRL (approved):

PR #9892, opened 2021-12-06
E2E Tests: Add tests for deleting/renaming/duplicating stories

### LINKED ISSUE #5043: E2E Tests: Story Actions on Dashboard
## Feature Description

To avoid regressions like #5020 and #5031 we should add e2e tests, as flagged by @swissspidy 

Let's add e2e tests for these actions

* Delete story
* Rename story
* Duplicate story

## Additional Context

---

amovar18:
## Context

This PR addresses the issue where e2e tests for actions:
1. Delete story
2. Rename story
3. Duplicate story

Above tests were requested to avoid regressions.

## Summary
This PR has 3 files which will test the 3 dashboard actions individually.

## Testing Instructions

## Checklist

---

Fixes [#5043](https://github.com/google/web-stories-wp/issues/5043)


packages/e2e-tests/src/specs/dashboard/deleteStory.js
```diff
@@ -0,0 +1,57 @@
...
+    await publishStory();
+    await visitDashboard();
+
+    await expect(page).toMatchElement('h2', { text: 'Dashboard' });
+    await page.hover('div[data-testid^=story-grid-item]');
+    const story = await page.evaluate(
+      'document.querySelector("div[data-testid^=story-grid-item]").getAttribute("data-testid")'
+    );
+    //need to get story id to click on the context menu of the story
+    const storyId = story.split('-');
```
BrittanyIRL: Definitely works as is, but you could isolate the id here and not have to specify which index it is on line 43 
`story.split('-')[3]`

packages/e2e-tests/src/specs/dashboard/deleteStory.js
```diff
@@ -0,0 +1,57 @@
...
+    const [deleteMenuButton] = await page.$x(
+      "//button[contains(.,'Delete Story')]"
+    );
+    await page.waitForTimeout(100);
+    await deleteMenuButton.click();
+    const [deleteButton] = await page.$x("//button[text()='Delete']");
+    await deleteButton.click();
+    await expect(page).not.toMatchElement('h3', {
+      text: 'Delete this story',
+    });
```
BrittanyIRL: What do you think about after this confirming that the story is gone by looking for the matching story title?

BrittanyIRL (commented):
Thanks for jumping into this, @amovar18

packages/e2e-tests/src/specs/dashboard/deleteStory.js
amovar18: @BrittanyIRL yes I am deleting the story after that I am checking that the story should not exist by looking for matching story title. I guess it would perform correctly since there is only one story being created in this test.

packages/e2e-tests/src/specs/dashboard/deleteStory.js
```diff
@@ -0,0 +1,58 @@
...
+ * External dependencies
+ */
+import {
+  createNewStory,
+  publishStory,
+  visitDashboard,
+  insertStoryTitle,
+} from '@web-stories-wp/e2e-test-utils';
+
+describe('Stories Dashboard', () => {
```
swissspidy: I feel like we could just combine all three tests into 1 file, e.g. `specs/dashboard/myStories.js`.

packages/e2e-tests/src/specs/dashboard/deleteStory.js
```diff
@@ -0,0 +1,58 @@
...
+  it('should delete story', async () => {
+    //need to create new story as all stories are deleted before test runs.
+    await createNewStory();
+    await insertStoryTitle('Delete this story');
+    await publishStory();
+    await visitDashboard();
+
+    await expect(page).toMatchElement('h2', { text: 'Dashboard' });
+    await page.hover('div[data-testid^=story-grid-item]');
```
swissspidy: Maybe selecting based on the `aria-label` is a bit more specific / self-documenting? Since there could be multiple grid items on the page.

`page.hover('[role="listitem"][aria-label="Details about Delete this story"]')`

WDYT?

packages/e2e-tests/src/specs/dashboard/deleteStory.js
```diff
@@ -0,0 +1,58 @@
...
+    const story = await page.evaluate(
+      'document.querySelector("div[data-testid^=story-grid-item]").getAttribute("data-testid")'
+    );
+    //need to get story id to click on the context menu of the story
+    const contextMenuId = story.split('-');
+    const storyId = contextMenuId[3];
+    //contextMenuId= ['story', 'grid', 'item', 'storyId']
+    await expect(page).toClick(
+      `button[data-testid="story-context-button-${storyId}"]`
+    );
```
swissspidy: Could simplify like this:

```js
await expect(page).toClick('button[aria-label="Context menu for Delete this story"]');
```

This has the advantage that we don't have to fiddle with testid, `page.evaluate`, etc. and can do it in a a more self-documenting way that also better reflects what a user would do.

packages/e2e-tests/src/specs/dashboard/deleteStory.js
```diff
@@ -0,0 +1,58 @@
...
+    const [deleteMenuButton] = await page.$x(
+      "//button[contains(.,'Delete Story')]"
+    );
+    await page.waitForTimeout(100);
+    await deleteMenuButton.click();
```
swissspidy: Could simplify:

```js
await expect(page).toClick('button', { text: 'Delete Story' });
```

packages/e2e-tests/src/specs/dashboard/deleteStory.js
```diff
@@ -0,0 +1,58 @@
...
+    const [deleteButton] = await page.$x("//button[text()='Delete']");
+    await deleteButton.click();
```
swissspidy: Similarly, could use `expect(page).toClick()` here.

packages/e2e-tests/src/specs/dashboard/deleteStory.js
amovar18: @swissspidy  yes we could do that too, but i dont think there will be multiple grid items on the page since before every test all the stories are deleted. Still would make this change to make it simple

packages/e2e-tests/src/specs/dashboard/myStories.js
```diff
@@ -0,0 +1,95 @@
+/*
+ * Copyright 2020 Google LLC
```
BrittanyIRL:
```suggestion
 * Copyright 2021 Google LLC
```

packages/e2e-tests/src/specs/dashboard/myStories.js
```diff
@@ -0,0 +1,95 @@
...
+  insertStoryTitle,
+} from '@web-stories-wp/e2e-test-utils';
+
+describe('Stories Dashboard', () => {
+  it('should delete story', async () => {
+    //need to create new story as all stories are deleted before test runs.
+    await createNewStory();
+    await insertStoryTitle('Delete this story');
+    await publishStory();
+    await visitDashboard();
```
BrittanyIRL: you can simplify this since it's the same series before each action you're testing by adding 
```
beforeEach(async () => {
    await createNewStory();
    await insertStoryTitle('Delete this story');
    await publishStory();
    await visitDashboard();
}); 
```

then you can don't need to have it within each individual test.

I know you're naming the story specific to the action but if you wanted to, you could make it something generic, store it as a constant and then reference that value where you're finding by labels with some interpolation to decrease potential typos. 

So like: 

`const storyName = 'Test Story';`

`'[role="listitem"][aria-label="Details about ${storyName}"]'`

BrittanyIRL (commented):
Looks good! Left a few suggestions to clean this up further but really appreciate the effort.

swissspidy (approved):

Issue #9647, opened 2021-11-05
Media: Use base color while loading in assets


spacedmonkey:
## Feature Description

In https://github.com/google/web-stories-wp/issues/9550, we now store the base color of the image / video with the resource.  Currently when resources are loading a grey box is rendered. See screenshot. 

<img width="348" alt="Screenshot 2021-11-05 at 15 52 27" src="https://user-images.githubusercontent.com/237508/140539813-dec62f78-b833-4464-b138-834728086a92.png">

But now we have the base color, we use this color as the background color. This is what other sites like unsplash do. See this example. 

<img width="1371" alt="Screenshot 2021-11-05 at 15 49 47" src="https://user-images.githubusercontent.com/237508/140540038-4590da7b-5091-43ba-8b04-63bc57d01aae.png">

We could do the same in the editor.

swissspidy:
Good idea. Makes sense to use the avg color in more places where possible.

Though worth noting that Unsplash uses [BlurHash](https://blurha.sh/), not just a solid color placeholders.

Reminds me that I actually once [proposed](https://xwp.slack.com/archives/CCCAUAH6F/p1588667580251200) using BlurHash.

Perhaps something worth revisiting?

cc @choumx @gajasinski @agingoldseco WDYT about using blurred placeholders?

gajasinski:
+1 a more polished experience compared to grey boxes

dreamofabear:
Love it.

spacedmonkey:
<img width="1904" alt="Screenshot 2021-11-08 at 11 18 44" src="https://user-images.githubusercontent.com/237508/140733234-f3786c87-46af-4a77-9560-4f4896e80028.png">

Google images also does this.

swissspidy:
Summary:

1. Using base color in media library is a no-brainer. We'll add this.
2. Blurhash is a nice enhancement on top of that, will create a new ticket for that.

spacedmonkey:
Related: #1194

csossi:
Verified in QA

PR #9905, opened 2021-12-06
Media: Generate and backfill BlurHash for media items

### LINKED ISSUE #9685: Media: Generate and backfill BlurHash for media items
## Feature Description

As a follow-up to #9550 and #9647, we want to adopt [BlurHash](https://blurha.sh/) as progressive enhancement on top of `baseColor` to get even nicer previews when media is loading.

* [ ] Register post meta to store blurhash strings on attachments
* [ ] Use [`blurhash`](https://github.com/woltapp/blurhash/tree/master/TypeScript) to create blurhash during upload and backfill blurhash for existing media items in the library
* [ ] Display blurhash while loading media items in the media library, perhaps using [`react-blurhash`](https://github.com/woltapp/react-blurhash), which seems pretty lightweight
* [ ] Update `Resource` JSDoc to add blurhash string
* [ ] Get `blurHash` from media3p (currently Unsplash-only) as well

spacedmonkey:
## Summary

As a follow-up to #9550 and #9647, we want to adopt BlurHash as progressive enhancement on top of baseColor to get even nicer previews when media is loading.

## Relevant Technical Choices

Blurhash can be expensive to generate, so only one is generated at a time and only after the upload is complete. 

## User-facing changes

https://user-images.githubusercontent.com/237508/145074451-1360b1bb-91c9-44ef-a700-9327c4f39ea5.mov

## Testing Instructions

1. Open editor.
2. Go 3P media. 
3. Click on image tab ( Unsplash ). 
4. See blurhashes as they load. 

## Checklist

---

Fixes #9685


packages/media/src/getBlurFromImage.js
```diff
@@ -0,0 +1,43 @@
...
+  ctx.drawImage(image, 0, 0);
+  return ctx.getImageData(0, 0, width, height);
+};
+
+const encodeImageToBlurhash = async (src) => {
+  const image = await preloadImage(src);
+  const imageData = getImageData(image);
+  const { data, width, height } = imageData;
+  // todo workout why 1 and 1 the only options that are not painfully slow.
+  return encode(data, width, height, 4, 4);
```
spacedmonkey: @swissspidy 4, 4 are the default values here. However, using these values, the editor is unable. If you change to 9,9, same story. 1,1 is usable, but does not generate a valid blurhash.

packages/story-editor/src/app/media/local/useContextValueProvider.js
```diff
@@ -205,7 +212,25 @@ export default function useContextValueProvider(reducerState, reducerActions) {
...
+  );
+
+  const processMediaBlurhash = useCallback(
+    (resource) => {
+      const { blurHashProcessed, blurHashProcessing } = stateRef.current;
+      const { id } = resource;
+
+      (async () => {
+        // Simple way to prevent double-uploading.
+        if (blurHashProcessed.includes(id) || blurHashProcessing.includes(id)) {
```
spacedmonkey: I tweaked this, to only generate on one media element, still didn't help the issue.

spacedmonkey:
I have been playing around with trying to make perform better with little luck.

packages/media/src/createResource.js
```diff
@@ -28,6 +28,7 @@ import normalizeResourceSizes from './normalizeResourceSizes';
  */
 function createResource({
   baseColor,
+  blurHash,
```
swissspidy: Nit: also need to update JSDoc

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
```diff
@@ -65,6 +66,12 @@ const InnerContainer = styled.div`
   }
 `;
 
+const BlurHastContainer = styled(Blurhash)`
```
swissspidy: `BlurHast` -> `BlurHash`

packages/story-editor/src/app/media/utils/useDetectBlurhash.js
```diff
@@ -0,0 +1,103 @@
...
+  );
+
+  const updateBlurHash = useCallback(
+    async ({ resource }) => {
+      const { type, src, poster } = resource;
+      const imageSrc = type === 'image' ? src : poster;
+      if (!imageSrc) {
+        return;
+      }
+      const imageSrcProxied = getProxiedUrl(resource, imageSrc);
```
swissspidy, line 89: Media items in the media library never need proxying

Hotlinked images never need blurhashes

Let's remove this.

packages/media/src/getBlurFromImage.js
swissspidy: The `useDetectBlurHash` hook calls this function with the full image size (`resource.src`).

Can we try to use a smaller thumbnail size instead. Can use `getSmallestUrlForWidth(0, resource)` for that.

Ideally with some quick time tracking to know how long it takes.

If that doesn't help, we can explore other options as discussed (not implementing, or doing it server-side)

swissspidy: As discussed, another option would be to use a web worker (fairly easy to do with [worker-loader](https://v4.webpack.js.org/loaders/worker-loader/)), but we'd need to measure performance to compare all possible options

packages/story-editor/src/app/media/utils/useDetectBlurhash.js
spacedmonkey: This isn't true. My personal site has access sorted in S3 and they do need proxying to work.

packages/media/src/getBlurFromImage.js
spacedmonkey: In my testing, I managed to the speed to much faster.

packages/story-editor/src/app/media/utils/useDetectBlurhash.js
```diff
@@ -0,0 +1,110 @@
...
+          resource: {
+            ...resource,
+            blurHash,
+          },
+        });
+        setProperties(id, newState);
+        updateMediaElement({
+          id,
+          data: { blurHash },
+        });
```
spacedmonkey, line 63: Do we need to update the resource on the page?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -230,20 +256,7 @@ function useMediaUploadQueue() {
...
-            additionalData.web_stories_is_muted = resource.isMuted;
-          }
-
-          if (resource?.baseColor) {
-            additionalData.meta = {
-              ...additionalData.meta,
-              web_stories_base_color: resource.baseColor,
-            };
-          }
+          const additionalData = addAdditionalData(_additionalData, resource);
```
spacedmonkey: Needed as function was too complex.

packages/story-editor/src/app/media/utils/useDetectBlurhash.js
swissspidy: You should set up CORS headers on your S3 bucket then.

I know it's convenient, but the CORS proxy was not meant to be used like this.

It is designed for use with hotlinking only.

Using it for _everything_ is going to have an unnecessary and huge performance impact on your server —without you ever knowing that you could just add CORS headers to improve things

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: That doesn't feel very readable. I think there's a bigger opportunity to reduce complexity in this function more holistically.

Let's do `// eslint-disable-next-line complexity` for now and solve complexity in a separate ticket.

packages/media/src/getBlurFromImage.js
swissspidy: What made it faster?

Do you have any numbers?

packages/media/src/getBlurFromImage.js
```diff
@@ -0,0 +1,43 @@
...
+};
+export default encodeImageToBlurhash;
```
swissspidy: Nit: let's rename both the file name and the function to `getBlurHashFromImage`

packages/media/src/getBlurFromImage.js
spacedmonkey: My tweaks to make it generate only one at a time. Which is not here. Also no numbers, the different between, editor completely locked and working fine.

packages/story-editor/src/app/media/utils/useDetectBlurhash.js
spacedmonkey: Do we nearly expect everyone to do this? The average user will not know how to do this...

packages/media/src/getBlurFromImage.js
swissspidy: How fast is the blurhash generation _now_?

packages/story-editor/src/app/media/local/useContextValueProvider.js
```diff
@@ -205,7 +213,26 @@ export default function useContextValueProvider(reducerState, reducerActions) {
...
+      const { blurHashProcessed, blurHashProcessing } = stateRef.current;
+      const { id } = resource;
+      // Don't process more than one item at a time and do not process items already processed/processing.
+      if (
+        blurHashProcessed.includes(id) ||
+        blurHashProcessing.includes(id) ||
+        blurHashProcessing.length > 0
+      ) {
+        return;
+      }
```
swissspidy, line 231: A comment explaining _why_ (performance) would be helpful.

packages/media/src/getBlurFromImage.js
spacedmonkey: Now that we are using smaller images too, the average blurhash is around 100-150ms, over 1000 ms of full sized resources. Not to mention the need to down the full size image, which is wasteful. But poster images could be pretty large, which is why we limit to one being processed at a time.

packages/story-editor/src/app/media/local/useContextValueProvider.js
```diff
@@ -205,7 +213,26 @@ export default function useContextValueProvider(reducerState, reducerActions) {
         removeBaseColorProcessing({ id });
       })();
     },
-    [setBaseColorProcessing, removeBaseColorProcessing, updateBaseColor]
+    [setBaseColorProcessing, updateBaseColor, removeBaseColorProcessing]
+  );
+
+  const processMediaBlurhash = useCallback(
+    async (resource) => {
```
swissspidy, line 221: This should work more like `processMediaBaseColor` above.

The callback itself is not async, but the code in it is wrapped in `(async () => { ... })();`

packages/story-editor/src/app/media/local/useContextValueProvider.js
```diff
@@ -263,6 +308,15 @@ export default function useContextValueProvider(reducerState, reducerActions) {
     media?.forEach((mediaElement) => postProcessingResource(mediaElement));
   }, [media, mediaType, searchTerm, postProcessingResource]);
 
+  useEffect(() => {
+    (async () => {
+      for (const mediaElement of media) {
+        // eslint-disable-next-line no-await-in-loop
+        await postProcessingBlurHashResource(mediaElement);
+      }
+    })();
```
swissspidy, line 319: awaits in loop is really bad for performance.

Hence my other comment above for not making `processMediaBlurhash` async.

Then this can be changed to a simple `media?.forEach((mediaElement) => postProcessingBlurHashResource(mediaElement));`

packages/story-editor/src/app/media/local/useContextValueProvider.js
```diff
@@ -249,6 +276,24 @@ export default function useContextValueProvider(reducerState, reducerActions) {
     ]
   );
 
+  const postProcessingBlurHashResource = useCallback(
```
swissspidy, line 281: Could you elaborate on why this is not part of `postProcessingResource`?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -245,6 +246,13 @@ function useMediaUploadQueue() {
             };
           }
 
+          if (resource?.blurHash && !resource?.trimData) {
```
swissspidy, line 249: Why the `trimData` check? 🤔

packages/media/src/getBlurFromImage.js
swissspidy: From the [documentation](https://github.com/woltapp/blurhash):

> We usually go with 4 by 3, which seems to strike a nice balance.
> However, you should adjust the number of components depending on the aspect ratio of your images. For instance, very wide images should have more X components and fewer Y components.

We have width & height here, so we could change the numbers based on the aspect ratio.

packages/story-editor/src/app/media/utils/useDetectBlurhash.js
swissspidy: That is a great question!

Since we only really use the blurhash in the media library, probably not.

On the other hand, we might want to use it on the page as well in the future, e.g. to show the blurhash while the editor is loading.

Plus, Miina is working on making it possible to ignore such updates in history.

So it won't hurt to do it.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
spacedmonkey: If you are trimming the video, it will have a different poster and different blurhash.

packages/story-editor/src/app/media/local/useContextValueProvider.js
spacedmonkey: Because we need to one fire one at a time here. We make it part of `postProcessingResource` then it will fire all the blurhash async.

packages/story-editor/src/app/media/local/useContextValueProvider.js
spacedmonkey: If we don't wait for each one to complete, then they all fire at once.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: Would be good adding a comment with this explanation.

Didn't we originally say that we don't generate a new poster for trimmed videos, because this difference is negligible?

packages/story-editor/src/app/media/local/useContextValueProvider.js
swissspidy: `processMediaBlurhash` makes sure that only one is being processed at a time.

packages/story-editor/src/app/media/utils/useDetectBlurhash.js
swissspidy: Yes.

I know that in support this happens very very rarely and can be resolved very quickly every time.
Part of the reason is that CDNs are usually set up by more technical people, so they know how to fix it.

I understand your concern, I really do.

If circumstances should change in the future (e.g. the number of support requests increasing significantly), this can be revisited.

packages/media/src/getBlurFromImage.js
swissspidy: Also, feels like this is better off living in `story-editor/src/utils`, just like `getMediaBaseColor.js`

packages/media/src/getBlurFromImage.js
swissspidy: Could you add `getTimeTracker` here?

spacedmonkey: I have a tracker elsewhere, want me to move it here?

spacedmonkey:
I have created a breakout PR for just rendering blurhash. See https://github.com/google/web-stories-wp/pull/9941. Let's merge that first.

packages/media/src/getBlurFromImage.js
swissspidy: Where is it? 👀

swissspidy:
Just converting to draft again to save some Percy snapshots :-)

swissspidy:
@spacedmonkey So I briefly looked into how we could possibly optimize this by reducing main thread work.

Some observations:

From what I can tell, we can't put `color-thief` in a worker, because it requires the DOM. Its API is designed so that it takes an image URL and creates a Canvas from that.

On the other hand, `blurhash` doesn't do that as it just accepts an `Uint8ClampedArray` as its first argument. So we could load the image into a canvas ourselves and then [transfer control over it to a worker](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/transferControlToOffscreen). So `blurhash` would do the encoding on this [`OffscreenCanvas`](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas) all in the worker.
The downside: this method is not available in Firefox and Safari. So using it might be considered progressive enhancement and we could fall back to some other method.

That said, we could still optimize `color-thief` usage — and in fact all resource backfilling (muted, poster generation) by using [`requestIdleCallback`](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback) so we prevent latency as much as possible.

Coincidentally, Max is working on some simple task queue handler in #9700 using `requestIdleCallback`, which is exactly what we'd need.

https://github.com/google/web-stories-wp/blob/cea8fd3081a26a786e2186c3569e660a6bc85b9c/packages/story-editor/src/app/pageDataUrls/pageDataUrlsProvider.js

Maybe we could turn it into a reusable `useTaskQueue()` hook or something that we use for all media backfilling.

PR #9907, opened 2021-12-06
Media: Various improvements to upload handling

swissspidy:
While reviewing #9785 I noticed several quirks in the media upload logic that we should fix.

From what I can tell, most have been there before and were simply uncovered by that PR.

A summary of the changes:

* Set a resource ID in `getResourceFromLocalFile`
  This guarantees that every newly uploaded resource gets a temporary, unique ID.
* After upload finishes, any resources with a previously temporary ID are replaced by the final resources with a permanent ID from the WP database. (this is what `previousResourceId` is used for)
* Remove any base color generation or post-processing logic from `useInsertElement`
  This is quite critical. When uploading a new item, `insertElement()` is called at the very start of the upload queue, when no file has been uploaded yet. This is way too early for calling these functions.  
  
  Also, since these are memoize callbacks, calling `isCurrentResourceUploading()` would have never worked and always return `false` because the callback was memoized before the upload happened.
  
  Calling `isBlobURL()` was just a workaround to still give the impression that it works.
* Any post-processing and color generation still happens for newly uploaded media items, as expected. Thanks to various `postProcessingResource` calls.
* Relatedly, any registerUsage logic is moved to the media3p provider, since that's the only place where this is really needed. `useInsertElement` was too generic for that.
* Various inline documentation changes

I think after this #9785 should be in good enough shape to be able to send it to QA.

If it helps, here's a quick diagram of how uploading a new item works:

Untitled document (1)


packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -409,100 +437,158 @@ function useMediaUploadQueue() {
...
       state.queue.some(
         (item) =>
-          [
-            ITEM_STATUS.PENDING,
-            ITEM_STATUS.UPLOADING,
-            ITEM_STATUS.TRANSCODING,
-            ITEM_STATUS.MUTING,
-            ITEM_STATUS.TRIMMING,
-          ].includes(item.state) && item.resource.id === resourceId
+          item.state !== ITEM_STATUS.CANCELLED &&
```
spacedmonkey: Not cancelled and not uploaded right?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -524,12 +614,21 @@ function useMediaUploadQueue() {
...
+     * @param {import('@web-stories-wp/media').Resource} resource Resource object.
+     * @return {boolean} Whether a given resource can be transcoded.
+     */
     const canTranscodeResource = (resource) => {
       const { isExternal, id, src } = resource || {};
+
       return (
         !isExternal &&
         src &&
-        !isBlobURL(src) &&
```
spacedmonkey, line 548: Why are you removing this?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
```diff
@@ -63,14 +63,16 @@ export function addItem(
   }
 ) {
   const id = uuidv4();
+
+  if (!resource.id) {
```
spacedmonkey, line 67: What if resource is null?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -409,100 +437,158 @@ function useMediaUploadQueue() {
...
      */
     const isCurrentResourceUploading = (resourceId) =>
       state.queue.some(
         (item) =>
-          item.state === ITEM_STATUS.UPLOADING &&
-          item.resource.id === resourceId
+          [
+            ITEM_STATUS.PENDING,
+            ITEM_STATUS.UPLOADING,
+            ITEM_STATUS.UPLOADED,
```
spacedmonkey, line 627: Naming of this function seemed wrong now? Why is this change needed?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
```diff
@@ -139,6 +141,7 @@ export function finishUploading(state, { payload: { id, resource } }) {
         ? {
             ...item,
             resource,
+            previousResourceId: item.resource.id,
```
spacedmonkey, line 144: What if resource is null?

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
```diff
@@ -87,11 +87,9 @@ function Element({
...
   } = resource;
 
-  const { isCurrentResourceProcessing } = useLocalMedia(
-    ({ state: { isCurrentResourceProcessing } }) => ({
-      isCurrentResourceProcessing,
-    })
-  );
+  const { isCurrentResourceProcessing } = useLocalMedia(({ state }) => ({
+    isCurrentResourceProcessing: state.isCurrentResourceProcessing,
+  }));
```
spacedmonkey: Why change this?

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
```diff
@@ -227,9 +225,7 @@ function Element({
           active={active}
         />
         {attribution}
-        {(!src ||
-          isBlobURL(src) ||
```
spacedmonkey, line 231: The is blob url check is a good way to see if it something is local. Why remove it?

packages/story-editor/src/components/library/panes/media/media3p/providerPanel.js
```diff
@@ -60,10 +63,60 @@ const ProviderWrapper = styled.div`
...
+  // Like useDetectBaseColor, but without updating the local media library.
+  const updateBaseColor = useCallback(
+    async (resource) => {
+      const { id, type, src, poster } = resource;
+      const imageSrc = type === 'image' ? src : poster;
+      if (!imageSrc) {
+        return;
+      }
+      try {
+        const color = await getMediaBaseColor(imageSrc);
```
spacedmonkey: You need to ensure that url is passed through the proxy.

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
```diff
@@ -115,20 +115,17 @@ function useVideoTrimMode() {
...
-    const { id, src, isExternal } = selectedElement.resource;
+
     return (
       isVideoTrimEnabled &&
       isTranscodingEnabled &&
-      src &&
-      !isBlobURL(src) &&
-      !isExternal &&
-      !isCurrentResourceUploading(id)
+      canTranscodeResource(selectedElement.resource)
```
spacedmonkey, line 122: 👍

packages/story-editor/src/app/media/utils/useDetectBaseColor.js
```diff
@@ -38,29 +38,23 @@ function useDetectBaseColor({ updateMediaElement }) {
     capabilities: { hasUploadMediaAction },
   } = useConfig();
   const { getProxiedUrl } = useCORSProxy();
-  const setProperties = useCallback(
```
spacedmonkey, line 41: This same snippet is in a number places. If we remove here, should be removed everywhere!

spacedmonkey:
https://user-images.githubusercontent.com/237508/144891437-710175a9-cc4d-4481-b319-d81c9a362181.mov

I am seeing longer times with no poster for trimmed videos. Any idea why this might be?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
swissspidy: Then this should still kick in and add an ID. That's intended.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: Because it's redundant. 

If it's a blob URL then the resource **must** be `isCurrentResourceProcessing` or `isResourceProcessing` already.

There's no need to check for blob URLs. We only must check on what's in the queue.

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
swissspidy: Improved readability. Three lines instead of five.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
swissspidy: No issue. Then `useUploadMedia` simply won't use it.

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
swissspidy: If we want to know whether a resource is not yet uploaded we have utilities like `isResourceProcessing` & co.

packages/story-editor/src/components/library/panes/media/media3p/providerPanel.js
swissspidy: This is 3p media.

3p media never needs the proxy.

packages/story-editor/src/app/media/utils/useDetectBaseColor.js
swissspidy: That's a nice to have.

I only changed it here because I was already touching this file.

Any other changes can be done in other PRs. P3 though.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: No, just checking for cancelled here because "uploaded" means the resource still needs to be updated everywhere.

It's only finished when it's either no longer in the queue or cancelled

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: I'll check if more tweaks are needed, but this comparison allows checking whether the media item is being uploaded (or preparing to be uploaded), but not currently being transcoded/trimmed/muted.

"Uploaded" still means there's some unfinished business, so that's included here as well.

swissspidy:
> I am seeing longer times with no poster for trimmed videos. Any idea why this might be?

Are you referring to seconds ~6-8 of your video or seconds ~23-26?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -409,100 +437,158 @@ function useMediaUploadQueue() {
...
     /**
-     * Is the current resource transcoding.
+     * Determine whether the current resource is being transcoded.
      *
-     * @param {number} resourceId Resource id.
-     * @return {boolean} if resource with id is found.
+     * This is the case when uploading a new media item that first
+     * needs to be transcoded.
+     * This is also the case when optimizing an existing video in the story,
+     * which will cause a new resource to be uploaded (the "current" one).
```
miina: Below, for `isResourceTranscoding`, the comment is the following: 

> When optimizing an existing video in the story, which will cause a new resource to be uploaded, this returns the state of this new resource.

So this means that `isResourceTranscoding` is for checking the newly added resource, right?

Here, it says this:

> This is also the case when optimizing an existing video in the story, which will cause a new resource to be uploaded (the "current" one)

I read out from here that the new resource to be uploaded is the "current" one (since it's mentioned in the parenthesis). Should `(the "current" one)` be after the `"...when optimizing an existing video"` instead, if `isCurrentResourceTranscoding` is for the origin video?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
miina: Would be nice to have it written out somewhere what each status means and in which order things happen.

Does `UPLOADED` mean that it goes back to `PENDING` since it's waiting for new unfinished business to be completed now? Or is the next step happening out of the upload queue and thus not `PENDING`, e.g. just waiting to replace the origin resource? Is there another state after `UPLOADED`?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
miina: Jonny was asking about the case when `resource` itself is `null` -- then there wouldn't be anything to add the ID to.

packages/story-editor/src/components/canvas/useInsertElement.js
```diff
@@ -140,11 +59,7 @@ function useInsertElement() {
       const element = createElementForCanvas(type, props);
       const { id, resource } = element;
       addElement({ element });
-      if (resource) {
-        postProcessingResource(resource);
-        handleRegisterUsage(resource);
-        generateBaseColor(element);
```
miina, line 146: If `baseColor` is missing from some previously existing local media, will it still be added somewhere?

miina (commented):
From testing:

When Trimming a Video, the Trim button and Mute button seem to be just fully hidden from the Panel instead of displaying it disabled with circular progress:
Screenshot 2021-12-07 at 11 22 06

Probably not intended?

packages/story-editor/src/components/canvas/useInsertElement.js
swissspidy: Yes, `postProcessingResource` is invoked in a few other places, mainly in the media library panel whenever items change there

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
swissspidy: Ah, got it.

The `uploadMedia` callback in `useUploadMedia` always passes a `resource` to `addItem()`. It's expected to never be null:

https://github.com/google/web-stories-wp/blob/4c23ee8ef2d5b38185d7f974751401987df5e367/packages/story-editor/src/app/media/useUploadMedia.js#L287-L296

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: I suppose I might want to extend my diagram.

> Does UPLOADED mean that it goes back to PENDING since it's waiting for new unfinished business to be completed now?

No. States never occur twice. `PENDING`  is only at the very beginning before anything happens.

> Or is the next step happening out of the upload queue and thus not PENDING, e.g. just waiting to replace the origin resource?

Correct

>  Is there another state after UPLOADED?

No

----

It's basically:

`PENDING`  -> `TRANSCODING` -> `TRANSCODED` -> `TRIMMING` -> `TRIMMED` -> `MUTING` -> `MUTED` -> `UPLOADING` -> `UPLOADED`

transcoding/trimming/muting steps are of course optional, depending on what has been passed.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
spacedmonkey: Is it worth checking it is an object anyway, just to be careful?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
spacedmonkey: But then you are trying to access an object that doesn't exist?

packages/story-editor/src/components/canvas/useInsertElement.js
```diff
@@ -18,116 +18,35 @@
...
-  const {
-    actions: { registerUsage },
-  } = useMedia3pApi();
-  const { getProxiedUrl, checkResourceAccess } = useCORSProxy();
 
   const { setZoomSetting } = useLayout(({ actions: { setZoomSetting } }) => ({
     setZoomSetting,
   }));
 
-  const generateBaseColor = useCallback(
```
spacedmonkey, line 63: @swissspidy Part of the reason this was where it was, was to handle base color for hotlinked images. Where is this handled now?

packages/story-editor/src/components/canvas/useInsertElement.js
spacedmonkey: What about hotlinked images? Or images inserted for WP media library?

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
spacedmonkey: I made a point of changing all these to be the other style for readability. I liked it better the way it was before. We should pick one and stick with it.

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
spacedmonkey: In my testing, the progress bar took half a second to show, as the media library and queue were not in sync. This check means that the progress bar show, straight away, instead of half a second later. 

It is an extremely inexpensive check, why not leave it in.

packages/story-editor/src/components/library/panes/media/media3p/providerPanel.js
spacedmonkey: I would still personally put it through the proxy anyway, just incase.

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
spacedmonkey: @miina @swissspidy I believe this line why the trim button is hidden now instead of disable on trimming. Trim mode need to enabled for the button to show to be disabled. This change needs to be reverted.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
spacedmonkey: Maybe we could add documentation in this [file](https://github.com/google/web-stories-wp/blob/d8c96c5ae1122cdd9210cf559c4aac56a7f2e1b2/packages/story-editor/src/app/media/utils/useMediaUploadQueue/constants.js).

packages/story-editor/src/components/canvas/useInsertElement.js
swissspidy: I thought I changed that, but seems like I forgot it.

This needs to be done in the hotlink `media/local/hotlink/useInsert.js` hook.

Definitely not here.

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
swissspidy: I'll take a look

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: Sounds like the naming is still confusing despite (or because of) the comments.

I'll revisit :)

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
swissspidy: It's never not an object.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
swissspidy: Misread; here, `resource` is never `null` as we're finishing uploading a resource. It's always an object at this point.

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
swissspidy: The other style makes sense when there are more things to return.

Not worth enforcing a specific style for everything IMO.

packages/story-editor/src/components/library/panes/media/media3p/providerPanel.js
swissspidy: Strongly disagree; the proxy is only needed for the use case of hotlinked media.

packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
```diff
@@ -115,14 +115,12 @@ function useVideoTrimMode() {
...
+
     return (
       isVideoTrimEnabled &&
       isTranscodingEnabled &&
-      src &&
-      !isBlobURL(src) &&
-      !isExternal &&
-      !isCurrentResourceUploading(id)
+      !selectedElement.resource.isExternal &&
+      !isCurrentResourceUploading(selectedElement.resource.id)
```
spacedmonkey: Why not just put this back to the way it was. 

```
!isExternal &&
      !isCurrentResourceUploading(id)
```

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: Add some documentation on top of the `useMediaUploadQueue` hook

spacedmonkey:
Still seeing the trim button disappear while trimming not disabled and displaying trimming message.

swissspidy:
> Still seeing the trim button disappear while trimming not disabled and displaying trimming message.

Are you refering to the button in the design panel?

That one changes correctly for me:

Screenshot 2021-12-08 at 13 18 02

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: Renamed functions & improved docs

spacedmonkey:
> Are you refering to the button in the design panel?

My bad, my pull failed before of the force push you did.

swissspidy:
One Jest test that seems to be flakey:

https://github.com/google/web-stories-wp/runs/4457385226?check_suite_focus=true#step:6:11

packages/story-editor/src/components/library/panes/media/media3p/providerPanel.js
```diff
@@ -60,10 +63,60 @@ const ProviderWrapper = styled.div`
...
+          registerUsageUrl: resource.attribution.registerUsageUrl,
+        });
+      }
+    },
+    [registerUsage]
+  );
+
+  // Update the base color for an existing media3p resource on the page.
+  // Like useDetectBaseColor, but without updating the local media library.
+  const updateBaseColor = useCallback(
```
spacedmonkey: Why do we need a new version of this function? Can we reuse the existing version?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -69,6 +83,9 @@ function useMediaUploadQueue() {
   const { uploadVideoPoster } = useUploadVideoFrame({
     updateMediaElement: noop,
   });
+
+  const isMounted = useRef(false);
```
spacedmonkey, line 87: Why is this is needed?

swissspidy: To prevent state updates after the component has already unmounted.

Mostly relevant for the tests to clean things up properly.

Otherwise you get errors, example:

1. Init stuff
2. Run async process
3. Hook is unmounted
4. Async process finishes, tries to update state
5. Error because component has unmounted already

I'm sure you've seen this message in the wild before at some point:

```
Warning: Can 't perform a React state update on an unmounted component. This is a no - op, but it indicates a memory leak in your application
```

packages/story-editor/src/components/library/panes/media/media3p/providerPanel.js
swissspidy: I've since updated the existing version to make this possible 👍

spacedmonkey:
Seeing a bug, where base color is generated and post rest api request is sent to a id that is invalid.

https://user-images.githubusercontent.com/237508/145312165-36174d22-6214-42cd-a879-5bed20f40071.mov

spacedmonkey:
We need to update this 
```
const canTranscodeResource = (resource) => {
      const { isExternal, id, src } = resource || {};
      return (
        !isExternal &&
        src &&
        !isCurrentResourceProcessing(id) &&
        !isNewResourceProcessing(id)
      );
    };
```
to this
```
const canTranscodeResource = (resource) => {
      const { isExternal, id, src } = resource || {};
      return (
        !isExternal &&
        src &&
        !isCurrentResourceProcessing(id) &&
        !isNewResourceProcessing(id) &&
       !isCurrentResourceUploading(id)
      );
    };
```

To ensure that currently uploading resources, are not transcoded in anyway.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
spacedmonkey, line 144: What is the difference between `previousResourceId` and `originalResourceId`?

swissspidy: `originalResourceId`:

You have an existing video ("A") on the page and you click "mute". A new video ("B") will be added to the queue for muting.  It references "A" as the `originalResourceId`. Maybe `parentResourceId` is a better term, not sure.

`previousRecordId`:

A new item is added to the queue. At first it's a local file with blob URL and a random ID.
After uploading to WordPress, the random ID gets replaced by the WP attachment ID. The previous one is stored as `previousRecordId`, so that we can update all instances of the local file with the final resource + attachment ID.

spacedmonkey: So it is a localResourceId then? Maybe we could rename it to make clear what it is?

swissspidy: It's the ID of the previous instance of this resource, which happened to be a local resource or placeholder.

`localResourceId` doesn't hit the nail on the head IMO.

Given that this is only used in few places, inline documentation should suffice for this.

spacedmonkey (approved):

PR #9924, opened 2021-12-06
Edit Media Dialog: Use Text Area for Video Description

### LINKED ISSUE #4576: Edit Image Pop-Up - Long File Name
## Bug Description

Clicked "edit" in the "dot" menu for an image withj a long filename

## Expected Behaviour

All content in popup should be visible to user

## Steps to Reproduce

- hover over image with long file name (i.e. see image)
- click the "three dot" menu
- select Edit

- note content cutoff

## Screenshots

image

## Additional Context

- Plugin Version: RC1
- Operating System: Win10
- Browser: Opera

---

BrookeGraham:
## Context

When editing an uploaded video in the editor, the description can get cut off if it is really long. 

## Summary

When selecting the ellipsis to edit media from the editor a `Dialog` pops up.  We want to use a text area as opposed to an input element. 

## Relevant Technical Choices

~The `Dialog` component in `Design Systems` didn't have a way to override the `DialogBox`'s styles. Added a prop to allow custom styling of the dialog box. If the file name is 21 characters or more, the dialog box is now given 100% max-width. This allows for approximately a minimum of 49 characters to be displayed before needing to scroll. This is due to the content below.  I could make `max-width` determined by `vw`, but 49-ish characters seems enough, thoughts?~

Above changes are obsolete. 

Since it is likely a video description be more than a few short words, I've updated the component to use `TextArea` as opposed to `Input` for this.  

## User-facing changes

Screen Shot 2021-12-06 at 2 45 02 PM
Screen Shot 2021-12-06 at 2 44 40 PM

## Testing Instructions

1. upload an image with a very long name. 
2. In the editor select the `...` on a video
3. Select 'edit' from the menu
4. You should see a Modal appear and the  video description should now be a text area box instead of a single line input. 

no

no

no

## Checklist

---

See #4576

### DIFF
--- a/packages/story-editor/src/components/library/panes/media/local/mediaEditDialog.js
+++ b/packages/story-editor/src/components/library/panes/media/local/mediaEditDialog.js
@@ -30,6 +30,7 @@ import {
 import {
   Input,
   Text,
+  TextArea,
   THEME_CONSTANTS,
   useSnackbar,
 } from '@web-stories-wp/design-system';
@@ -42,6 +43,10 @@ import { useLocalMedia } from '../../../../../app/media';
 import Dialog from '../../../../dialog';
 
 const THUMBNAIL_WIDTH = 152;
+
+const styledDialogBox = css`
+  max-width: 100%;
+`;
 
 const styledMediaThumbnail = css`
   display: flex;
@@ -74,6 +79,10 @@ const DateText = styled(Text)`
 `;
 
 const AssistiveTextInput = styled(Input)`
+  margin: 20px 0 4px;
+`;
+
+const AssistiveTextArea = styled(TextArea)`
   margin: 20px 0 4px;
 `;
 
@@ -141,6 +150,7 @@ function MediaEditDialog({ resource, onClose }) {
       secondaryText={__('Cancel', 'web-stories')}
       onPrimary={updateMediaItem}
       primaryText={__('Save', 'web-stories')}
+      dialogBoxStyleOverride={altText.length >= 21 ? styledDialogBox : null}
     >
       <DialogBody>
         {type === 'image' ? (
@@ -182,13 +192,24 @@ function MediaEditDialog({ resource, onClose }) {
               height
             )}
           </Text>
-          <AssistiveTextInput
-            value={altText}
-            aria-label={isImage ? imageInputTitle : videoInputTitle}
-            type="text"
-            placeholder={isImage ? imageInputTitle : videoInputTitle}
-            onChange={handleAltTextChange}
-          />
+
+          {isImage ? (
+            <AssistiveTextInput
+              value={altText}
+              aria-label={imageInputTitle}
+              type="text"
+              placeholder={imageInputTitle}
+              onChange={handleAltTextChange}
+            />
+          ) : (
+            <AssistiveTextArea
+              value={altText}
+              aria-label={videoInputTitle}
+              type="text"
+              placeholder={videoInputTitle}
+              onChange={handleAltTextChange}
+            />
+          )}
           <Text size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL}>
             {isImage ? imageDialogDescription : videoDialogDescription}
           </Text>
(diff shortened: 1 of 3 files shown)


packages/story-editor/src/components/library/panes/media/local/mediaEditDialog.js
```diff
@@ -182,13 +192,24 @@ function MediaEditDialog({ resource, onClose }) {
...
+            />
+          ) : (
+            <AssistiveTextArea
+              value={altText}
+              aria-label={videoInputTitle}
+              type="text"
+              placeholder={videoInputTitle}
+              onChange={handleAltTextChange}
+            />
+          )}
```
BrittanyIRL: Big fan of this improvement to use a text area for video descriptions.

packages/story-editor/src/components/library/panes/media/local/mediaEditDialog.js
```diff
@@ -141,6 +150,7 @@ function MediaEditDialog({ resource, onClose }) {
       secondaryText={__('Cancel', 'web-stories')}
       onPrimary={updateMediaItem}
       primaryText={__('Save', 'web-stories')}
+      dialogBoxStyleOverride={altText.length >= 21 ? styledDialogBox : null}
```
BrittanyIRL: So, I'm looking at the screenshot of the reported bug and what we've got here now and I think we've lost the display version of the media title. 

| Reported bug | Main now | 
| --- | --- | 
| image | Screen Shot 2021-12-06 at 4 42 33 PM |

With that in mind, I'm curious what you think about just updating the input here to have an ellipsis for text-overflow and leaving the width as is. The title isn't getting displayed as display text anymore so there's nothing there to add text wrap to.

BrookeGraham: Yeah, I am for it.

BrookeGraham: Though. The title was removed so now only the input is what is being cut off. You can see the whole name of the image, you just have to scroll. If we throw in some ellipsis, the name would get forever cutoff. If a user has something like `two otters telling secrets on an iceberg or something else` and `two otters telling secrets on an iceberg or something else 2` they wouldn't know the difference. So maybe this ticket is actually obsolete now that the title isn't being used above? Or are you thinking the ellipsis would be clickable and would show the rest of the text somehow?

BrittanyIRL: I was more thinking the latter - since the ellipsis would say "hey, there's more copy here" and then either scrolling or clicking into it you get the chance to see it. But on second thought, there's nothing really getting cut off as is and you know the moment you scroll that the title's longer. None of our other inputs show an ellipsis if the content gets longer. I think this is obsolete, like you said! 

Let's confirm with design today in our sync.

BrittanyIRL: Just looping back to this post design sync - we all agree that this is obsolete since the dialog style has changed since this ticket was originally created. No need to update anything with an ellipsis, this is all within our existing patterns and expectations. 

Sorry for not catching the design change before throwing this into the sprint. I'm glad that in it you found the idea to update to a text area when the media's a video instead of a text input - so still a nice use of time.

BrittanyIRL (approved):
What a journey for this "bug". Thanks for finding a silver lining in an out of date no longer relevant bug and improving the dialog!

packages/story-editor/src/components/library/panes/media/local/mediaEditDialog.js
```diff
@@ -182,13 +187,24 @@ function MediaEditDialog({ resource, onClose }) {
...
+          {isImage ? (
+            <AssistiveTextInput
+              value={altText}
+              aria-label={imageInputTitle}
+              type="text"
+              placeholder={imageInputTitle}
+              onChange={handleAltTextChange}
+            />
+          ) : (
+            <AssistiveTextArea
```
swissspidy: Any downside of just always using a textarea?

samwhale (approved):
Nice this looks great. Yeah maybe we should always use a text area?

maxyinger (approved):

packages/story-editor/src/components/library/panes/media/local/mediaEditDialog.js
BrookeGraham: Not really. Will update since that will make it a bit cleaner.

PR #9935, opened 2021-12-07
Media: Use smaller images for base color generation

spacedmonkey:
## Summary
While working https://github.com/google/web-stories-wp/pull/9905, I got better performance from blurhash with smaller images. Base color will be the same. Let's use for base color to.

## Testing Instructions

1.

## Checklist

---

Fixes #

### DIFF
--- a/packages/story-editor/src/app/media/local/useContextValueProvider.js
+++ b/packages/story-editor/src/app/media/local/useContextValueProvider.js
@@ -18,6 +18,7 @@
  * External dependencies
  */
 import { useEffect, useCallback, useRef } from '@web-stories-wp/react';
+import { getSmallestUrlForWidth } from '@web-stories-wp/media';
 import { getTimeTracker } from '@web-stories-wp/tracking';
 
 /**
@@ -236,7 +237,8 @@ export default function useContextValueProvider(reducerState, reducerActions) {
         processVideoAudio(id, src);
       }
 
-      const imageSrc = type === 'image' ? src : poster;
+      const imageSrc =
+        type === 'image' ? getSmallestUrlForWidth(0, resource) : poster;
       if (imageSrc && !baseColor) {
         processMediaBaseColor(resource);
       }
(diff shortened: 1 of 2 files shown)


swissspidy:
Any before/after numbers?

miina:
Also curious to see the before/after numbers, especially since the during the base color calculation a 10px width image is used, wondering the difference this change creates.

spacedmonkey:
The benefits of this one seem pretty clear. My testing, I tested a 10MB image. 

#### Before 
Screenshot 2021-12-08 at 12 20 09

#### After
Screenshot 2021-12-08 at 12 22 45

The benefit here is clear. If the user has large images, like 10MB+, that is going to be slow to download and waste a lot of resources to download. 100 media items x 10MB is 1GB of data. 

The downside is also clear, 1. smallest image is less likely to already be loaded and will have to load, 2. the base color will not be as accurate, as it is a smaller sample size. 

There is benefits here, but I am willing to close this PR out, if don't think it makes sense.

miina:
> the base color will not be as accurate, as it is a smaller sample size.

It's using a smaller image anyway when creating an image tag so that shouldn't be an issue. Also, probably accuracy is not that super important in case of this feature anyway as long as it still generally gets a similar color.

packages/story-editor/src/app/media/local/useContextValueProvider.js
```diff
@@ -236,7 +237,8 @@ export default function useContextValueProvider(reducerState, reducerActions) {
         processVideoAudio(id, src);
       }
 
-      const imageSrc = type === 'image' ? src : poster;
+      const imageSrc =
+        type === 'image' ? getSmallestUrlForWidth(0, resource) : poster;
```
swissspidy, line 241: Something for the future:

Do we still store `posterId` for resources?

If so, we could theoretically fetch `/wp-json/web-stories/v1/media/<posterId>` and get a smaller image size for poster images.

swissspidy (approved):

PR #9942, opened 2021-12-08
Text style: adjust margin calculation

### LINKED ISSUE #9940: Text not adhering to line-height from version 1.15 onwards
## Bug Description

Text is not adhering to the set line-height from Web Stories version 1.15.0 onwards. This has been pointed out by a user of the plugin and confirmed with testing of my own. 

In the screenshots section you will see an image of versions 1.14 and 1.15.1 both using a 'Title 1' text box with a line-height value of 4 set.

## Expected Behaviour

Line height should display as seen in 1.14 screenshot. 

## Steps to Reproduce

1. While editing a story add a new text box to a page.
2. From the text options set a line-height value of 4.

## Screenshots

WS-line-height

## Additional Context

This is the behaviour as of version 1.15.0.

miina:
## Summary
15.1 introduced a regression where margin is not properly calculated when assigning line-height.

## Relevant Technical Choices
Margin was incorrectly calculated inside `generateParagraphTextStyle` (without units), and the generated `margin` was not actually used anywhere at all, and was always calculated again in `elementWithTextParagraphStyle`.

This PR calculates the margin correctly the first time, and removes the additional calculation from `elementWithTextParagraphStyle`.

## To-do

- [x] Additional tests

## User-facing changes

| Before | After |
|-------|-------|
| Screenshot 2021-12-09 at 13 10 21 | Screenshot 2021-12-09 at 13 08 48 |
## Testing Instructions

1. Add a text element so that it would have one line only.
2. Add line-height, e.g. 4. Verify the text element looks similar to as it was before (line-height is not visibly applied).
3. Now make the text element two rows.
4. Verify that line-height is properly added between the two lines but not to the top and bottom of the element iself.

No

No

No

## Checklist

---

Fixes #9940

### DIFF
--- a/packages/story-editor/src/elements/shared/index.js
+++ b/packages/story-editor/src/elements/shared/index.js
@@ -19,6 +19,7 @@
  */
 import { css } from 'styled-components';
 import { generatePatternStyles } from '@web-stories-wp/patterns';
+
 /**
  * Internal dependencies
  */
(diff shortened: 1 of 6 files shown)


packages/story-editor/src/elements/shared/index.js
```diff
@@ -95,16 +96,13 @@ export const elementWithFont = css`
...
-    margin: ${-dataToEditorY(marginOffset / 2)}px 0;
-    padding: ${({ padding }) => padding || 0};
-    line-height: ${({ lineHeight }) => lineHeight};
-    text-align: ${({ textAlign }) => textAlign};
-    overflow-wrap: break-word;
-  `;
-};
+export const elementWithTextParagraphStyle = css`
+  margin: ${({ margin }) => margin};
+  padding: ${({ padding }) => padding || 0};
```
barklund, line 101: Margin doesn't need a fallback here as padding has?

Trapsta (approved):
✅   Line-height looks fine in both editor and preview

barklund (approved):
Looks good. Surprised that this bug snuck in there without anyone noticing.

packages/story-editor/src/elements/shared/index.js
miina: It shouldn't really since it's always calculated in the paragraph style generation, even if 0.

Issue #8809, opened 2021-08-26
Karma: improve error output


swissspidy:
## Task Description

Error output in Karma tests ought to be improved by #6900 / #6545, but it doesn't seem to work as expected or is not good enough yet.

We don't need all this bloated error output on CI, making the logs impossible to use.

swissspidy:
Regarding the `Blocked attempt to create a WebMediaPlayer as there are too many WebMediaPlayers already in existence. See crbug.com/1144736#c27` warning, it sounds like we could do a better job with cleaning up media players after tests, e.g. by setting `mediaElement.src = ''`. See https://bugs.chromium.org/p/chromium/issues/detail?id=1232649#c18

We also might be able to set `--max-web-media-player-count=<some high number>` as a workaround.

BrittanyIRL:
See #6545 for more context

PR #9944, opened 2021-12-08
Karma: Improve output to reduce noise

### LINKED ISSUE #10006: Canvas: `isRTL` being passed to svg
## Bug Description

`isRTL` is not an html attribute, but `isRTL` is being passed to the svg tags in the video element (pause and play)

## Expected Behaviour

`isRTL` should not be passed to the svg element.

## Steps to Reproduce

1. Add video to canvas
2. Hover or click the video to show svg
3. Should see error in console

swissspidy:
See #8809.
Fixes #10006.


packages/story-editor/src/components/inspector/design/getDesignPanelsForSelection.js
```diff
@@ -85,10 +85,13 @@ function getDesignPanelsForSelection(elements) {
       });
       // In case of default background without media:
     } else {
-      panels.unshift({
-        type: PanelTypes.PAGE_BACKGROUND,
-        Panel: PageBackgroundPanel,
-      });
+      // Avoids adding the pageBackground panel twice.
+      if (!isBackgroundMedia) {
```
miina: It checks for `video` and `image` and `gif` as the element type in the `if`s above, doesn't it already mean that it's not background media if it gets to this `else`?

packages/react/src/useResizeEffect.js
```diff
@@ -51,10 +53,15 @@ function useResizeEffect(ref, handler, deps = undefined) {
         // requestAnimationFrame prevents the 'ResizeObserver loop limit exceeded' error
         // https://stackoverflow.com/a/58701523/13078978
         window.requestAnimationFrame(() => {
+          if (!mounted) {
```
miina: Wouldn't checking for `ref?.current` work?

packages/story-editor/src/components/canvas/singleSelectionMoveable/index.js
```diff
@@ -154,6 +163,10 @@ function SingleSelectionMoveable({
    */
   const resetMoveable = useBatchingCallback(
     (target) => {
+      if (!isMounted.current) {
```
miina: Same here, I wonder if checking for one of the refs here wouldn't work, e.g. `moveable.current` (or `target`).

packages/karma-fixture/src/init.js
```diff
@@ -396,5 +396,20 @@ beforeEach(async () => {
 });
 
 afterEach(() => {
+  // Try to clean up any lingering media elements.
```
spacedmonkey: I am still seeing warning in the log. 

```
[1208/203114.673605:INFO:CONSOLE(0)] "Blocked attempt to create a WebMediaPlayer as there are too many WebMediaPlayers already in existence. See crbug.com/1144736#c27", source: http://localhost:9876/?id=60207859#page=%25222a164aef-3fc7-40d2-a32e-dc916896de2c%2522 (0)
```

packages/karma-puppeteer-launcher/src/index.cjs
```diff
@@ -50,6 +50,9 @@ function puppeteerBrowser(baseBrowserDecorator, config) {
         // See http://crbug.com/715363
         // We use this flag to work-around this issue.
         '--disable-dev-shm-usage',
+        // Try to prevent "Blocked attempt to create a WebMediaPlayer" warnings in Chrome.
+        // See crbug.com/1144736#c27
+        '--max-web-media-player-count=10000'
```
spacedmonkey: Can we make this number higher?

packages/story-editor/src/components/inspector/design/getDesignPanelsForSelection.js
spacedmonkey: Should this be an elseif?

swissspidy: @miina It should, and in WordPress it seems to work fine. But when running Karma tests there are a few warnings about `pageBackground` being used two times as the `key` prop, which would indicate that the panel is displayed twice.

packages/karma-fixture/src/init.js
swissspidy: Yes I know 😞

packages/karma-puppeteer-launcher/src/index.cjs
swissspidy: I don't think we have more than even 100 video players to be honest.

This number is already extremely high.

packages/story-editor/src/components/inspector/design/getDesignPanelsForSelection.js
miina: Hm, does this change fix it? Could you point to an example run with this warning?

packages/karma-fixture/src/init.js
miina: Would it be possible to try cleaning it up generally with `onended` event listener, not just in `afterEach()`?
Not sure if it's possible to add general listeners like this for Karma.

packages/story-editor/src/components/inspector/design/getDesignPanelsForSelection.js
swissspidy: It does, yes. I don't have a run URL handy for that, but try running the tests yourself with this branch both with and without this particular change here.

packages/karma-fixture/src/init.js
swissspidy: Wouldn't that break tests? because the media element is potentially removed in the middle of a test.

That would probably mean adding a mutation observer to detect any time a new media element is added and then attach an `ended` listener for that media element. And immediately removing an element at that po

Initial research indicates that cleanup like in this PR in an `afterEach` _should_ work.

miina: Hm, good point that this would potentially remove it in the middle of the test. Was thinking of alternatives since it doesn't seem to be working in `afterEach` for some reason 🤷

swissspidy: The weird thing is that these warnings pop up already after the first few tests 🤔

miina: Maybe there's something off with how media is set up for each test?

Even if we'd be creating a lot of videos for each test, still seems very odd that the limit 10000 is exceeded after the first few tests 😕 

I wonder, when doing `document.querySelectorAll('video,audio')` inside a test and logging out the count, does it show unexpectedly high numbers?

miina: If I understand correctly from `fixture.js` then we're creating 104 media items each time `getMedia` is used, 25% of those are videos. In theory, if all those would be created and stay lingering, then we'd still need do this almost 400 times to reach the 10000 limit.

packages/story-editor/src/components/inspector/design/getDesignPanelsForSelection.js
miina: @swissspidy I was looking into this a bit more and it looks like the issue comes from [`formattedTemplatesArray.js`](https://github.com/google/web-stories-wp/blob/3b0af267c70a30b736d3f5cd44db2ccdf532a5c9/packages/story-editor/src/dataUtils/formattedTemplatesArray.js) that's used in the tests having incorrect background element values. There are cases where the background element type is `shape`, but `isDefaultBackground` is not set to `true` -- this never happens in the editor. In the editor, the background element is either media, or default background (shape, `isDefaultBackground: true`).

packages/karma-fixture/src/init.js
swissspidy: The amount of media items in the library is one thing, the number of actual video elements created because of re-renders is another. Chrome will create a new _player_ every time in that case.

You can see that by opening `chrome://media-internals/` in Chromium when running tests in non-headless and observing the log. There are tons and tons of players being created.

Most are discarded really quickly as tests progress, but some players seem to be lingering for up to a minute before being destroyed.

swissspidy: cc @choumx perhaps you have some ideas here to help declutter the logs from these messages?

packages/story-editor/src/components/inspector/design/getDesignPanelsForSelection.js
swissspidy: Cool, will revert in that case and update that fixture.

packages/design-system/src/components/checkbox/index.js
```diff
@@ -95,7 +95,7 @@ const BaseCheckbox = ({ checked, disabled, className = '', ...props }, ref) => (
     <input
       type="checkbox"
       ref={ref}
-      checked={checked}
+      checked={Boolean(checked)}
```
samwhale, line 98: 🤔 Do we need this? Having the prop type of `checked: PropTypes.bool` should be enough to warn us.

But I get that we could still get past that pretty easily.

packages/design-system/src/components/snackbar/snackbarMessage.js
```diff
@@ -271,7 +271,7 @@ const SnackbarMessage = ({
 };
 
 SnackbarMessage.propTypes = {
-  'aria-label': PropTypes.string.isRequired,
+  'aria-label': PropTypes.string,
```
samwhale, line 274: 👍

packages/story-editor/src/components/checklist/checklistContent/designChecks.js
```diff
@@ -38,7 +38,7 @@ function DesignPanel({
   return isChecklistMounted ? (
     <StyledTablistPanel
       badgeCount={badgeCount}
-      isExpanded={badgeCount && isOpen}
+      isExpanded={Boolean(badgeCount) && isOpen}
```
samwhale, line 41: 👍

packages/story-editor/src/components/helpCenter/navigator/bottomNavigation.js
```diff
@@ -118,7 +118,7 @@ export function BottomNavigation({
           size={BUTTON_SIZES.SMALL}
           disabled={!hasBottomNavigation}
         >
-          <ArrowWrap isRTL={isRTL}>
+          <ArrowWrap $isRTL={isRTL}>
```
samwhale, line 121: 👍

packages/design-system/src/components/checkbox/index.js
swissspidy: There were lots of prop types warnings in the karma tests that nobody paid attention to, so 🤷‍♂️

packages/karma-fixture/src/init.js
swissspidy: Hmm #10023 might fix this actually

maxyinger: oh nice. since it'll just render a poster for the video instead of the video?

yea I've been looking into the media players error in the karma logs. noticed when we start getting the `WebMediaPlayer` error it increments by 12 every time we remount the fixture cus theres 12 media items that are videos rendered in the media tab.

I couldn't figure out why the detached video nodes weren't getting GCed, but seems to be an issue for a lot of ppl with react/chrome:
https://github.com/facebook/react/issues/15583

Would be nice to figure out how to have less of a memory leak around dom nodes in general when restoring and remounting the fixture between tests. Definitely gets north of 100,000 nodes by some of the later tests if you run the performance monitor while the karma tests are running (and that's after forcing a garbage collection run in the memory tab)

I tried using [cleanup()](https://testing-library.com/docs/react-testing-library/api/#cleanup) in our `restore` method for the fixture, but still wasn't quite doing the trick

maxyinger: This is right after forcing GC and having the karma tests run for 20 minutes or so
Screen Shot 2022-01-06 at 1 19 46 PM

maxyinger: I'm looking at a lot of the leftover `Detached HTMLDivElement` that should be cleaned up after the fixture is restored. `SelectionCanvas` and `updateLasso` seem to come up a lot when inspecting references. Seems to be the only thing for a lot of them beyond react internals.

I'm curious if there's an event listener that's not getting cleaned up or some type of closure issue that's leaking and holding onto references here.

Seems to be an area of the codebase with a high probability for this to happen cus there's a lot going on with regard to event listeners, closures and refs

packages/karma-puppeteer-launcher/src/index.cjs
maxyinger: it's just the 12 video media elements in the media library panel. They just stick around after we restore the fixture. If you pay close attention the the errors while you're watching a running test, the `Blocked attempt to create a WebMediaPlayer` will increment by 12 every time the fixture remounts.

Trying to look at memory profiles, but they're super hard to track. Could possibly be a react/chrome error like mentioned above

packages/karma-fixture/src/init.js
maxyinger: Maybe it's not from the `SelectionCanvas` specifically 🤷 . Commented all the event handlers out and a lot of detached HTMLDivElements are still being kept around from other stuff.

packages/story-editor/src/components/checklist/karma/checklist.karma.js
```diff
@@ -82,6 +82,10 @@ describe('Checklist integration', () => {
           type: 'image',
           mimeType: 'image/jpg',
           src: 'http://localhost:9876/__static__/earth.jpg',
+          alt: 'earth',
```
BrittanyIRL, line 85: so i'm pretty sure that this is causing the test failure for `should open the design inspector panel and focus the text input` because we're relying on having that alt value as missing to generate accessibility issues.

swissspidy: Hmm then it should probably be an empty string instead of `undefined`. I'll give it a try.

packages/story-editor/src/components/checklist/karma/prepublishSelect.karma.js
```diff
@@ -259,11 +259,14 @@ describe('Pre-publish checklist select offending elements onClick', () => {
           width: 640 / 2,
           height: 529 / 2,
           resource: {
-            width: 640,
-            height: 529,
+            id: 10,
             type: 'image',
             mimeType: 'image/jpg',
             src: 'http://localhost:9876/__static__/earth.jpg',
+            alt: 'earth',
```
BrittanyIRL: Sorry, my earlier comment was for _this_ alt text. we need this to be missing for the test to pass because we want to flag the missing alt text. 

```suggestion
            alt: '',
```

packages/story-editor/src/components/checklist/karma/checklist.karma.js
BrittanyIRL:
```suggestion
          alt: '',
```

This needs to be missing so that the image is flagged in the accessibility issue check. Will fix this: https://github.com/google/web-stories-wp/runs/4781311704?check_suite_focus=true#step:6:91

BrittanyIRL (approved):
Damn Javascript tests!! 

Aside from those, the karma updates look great - so I'm approving based on that and I will look more closely at the JS issues and see if I find anything that can move this along faster. 


This is not fun task, thank you for fixing so much noise. 

Looking at the failing tests, I left changes to fix the checklist ones (where we have intentionally crummy alt text). Not sure how to fix this, but it looks like [the tooltip's just not ever showing on hover](https://github.com/google/web-stories-wp/runs/4781311704?check_suite_focus=true#step:6:953)? I can reproduce the issue locally. @miina is this a known thing? 

first run is `Executed 312 of 461 (2 FAILED) (skipped 51)` so LGTM. https://github.com/google/web-stories-wp/runs/4780977349?check_suite_focus=true#step:6:1830

I definitely see reducing the noise in karma output as a progressive improvement and this is a grand start.

swissspidy:
Dashboard tests looking great already 😻 

Screenshot 2022-01-12 at 13 36 32

swissspidy:
Output is still not great and I had to disable one test, but the current state is already a great improvement.

BrittanyIRL (approved):
So much less scrolling through logs! Did a bit of smoke testing too and ran karma locally.

PR #9945, opened 2021-12-08
Editor: Fixed updating existing resource when retrimming a video

### LINKED ISSUE #9878: Video: Re-trimming a video for the second time not working
## Bug Description
Trimming a video works well and re-trimming it once seemingly works as well but re-trimming the re-trimmed video for the second time doesn't seem to work as expected. The trimmer handles show incorrect value.

## Steps to Reproduce

1. Add a non-trimmed video.
2. Trim the video (e.g. from 15 seconds to 6 seconds) - works
3. Wait until trimming finishes and open the trimmer again - works
4. Now re-trim the video again (e.g. from 6 seconds to 2 seconds) - visibly seems to work
5. Open the trimmer again. The trimmer still shows 6 seconds as the length instead of 2 seconds. - incorrect value

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

barklund:
## Context

When retrimming a video, the logic that updated properties on the element already on stage was broken, as it tried to update properties of an element with the resource id of the original video that this instance is a trimmed version of.

This was evident from the fact, that when retrimming, the current video on-stage never changed to be in trimming state nor did the trim button disable for the element. And finally, then trimming was done, the on-stage element was not correctly updated to refer to the newly created trim video.

## User-facing changes

| Before | After |
|-|-|
| retrim-fix-before | retrim-fix-after |

## Testing Instructions

1. Add an original video to a canvas
2. Trim it to a new length and observe that the trim button in the design panel is disabled while trimming
3. Trim the newly created video again and observe that the trim button in the design panel is once again disabled while trimming
4. Start trimming the new-newly created video and observe, that the trim handles are correctly placed as you placed them after the second trim.

## Checklist

---

Fixes #9878

### DIFF
--- a/packages/story-editor/src/app/media/utils/useProcessMedia.js
+++ b/packages/story-editor/src/app/media/utils/useProcessMedia.js
@@ -193,16 +193,20 @@ function useProcessMedia({
       };
 
       const onUploadStart = () =>
-        updateExistingElements(resourceId, {
+        updateExistingElements(canvasResourceId, {
           trimData,
           isTrimming: true,
         });
 
       const onUploadError = () =>
-        updateExistingElements(resourceId, { isTrimming: false });
+        updateExistingElements(canvasResourceId, { isTrimming: false });
 
       const onUploadSuccess = ({ resource }) => {
-        copyResourceData({ oldResource, resource });
+        const properOldResource = {
+          alt: oldResource.alt,
+          id: canvasResourceId,
+        };
+        copyResourceData({ oldResource: properOldResource, resource });
         postProcessingResource(resource);
       };
 
--- a/packages/story-editor/src/app/media/utils/useProcessMedia.js
+++ b/packages/story-editor/src/app/media/utils/useProcessMedia.js
@@ -202,11 +202,11 @@ function useProcessMedia({
         updateExistingElements(canvasResourceId, { isTrimming: false });
 
       const onUploadSuccess = ({ resource }) => {
-        const properOldResource = {
+        const oldCanvasResource = {
           alt: oldResource.alt,
           id: canvasResourceId,
         };
-        copyResourceData({ oldResource: properOldResource, resource });
+        copyResourceData({ oldResource: oldCanvasResource, resource });
         postProcessingResource(resource);
       };
 


packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -193,16 +193,20 @@ function useProcessMedia({
...
           isTrimming: true,
         });
 
       const onUploadError = () =>
-        updateExistingElements(resourceId, { isTrimming: false });
+        updateExistingElements(canvasResourceId, { isTrimming: false });
 
       const onUploadSuccess = ({ resource }) => {
-        copyResourceData({ oldResource, resource });
+        const properOldResource = {
```
miina: `properOldResource` 🤔

miina (approved):
Works like a charm!

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -193,16 +193,20 @@ function useProcessMedia({
       };
 
       const onUploadStart = () =>
-        updateExistingElements(resourceId, {
+        updateExistingElements(canvasResourceId, {
```
miina, line 196: cc @swissspidy for visibility (since you're also doing some refactoring work in these files)

spacedmonkey:
I strongly recommend reviewing my PR https://github.com/google/web-stories-wp/pull/9785 and not to merge this PR until that one is merged as well.

miina:
> I strongly recommend reviewing my PR #9785 and not to merge this PR until that one is merged as well.

This is a small functionality fix here specific to trimming that could be brought over to the other PRs, doesn't seem to make sense to hold off a small change like this, as opposed to #9785 that's a massive PR and touches all parts of media processing.

If you see specific places in this PR that will be influenced / invalid, please leave comments into the relevant files (as far as I can see, there aren't big conflicts here, just switching to using the correct ID).

packages/story-editor/src/app/media/utils/useProcessMedia.js
barklund: Yeah, I know... `oldCanvasResource` is probably a better name, as that's what it is.

miina: Yep, `oldCanvasResource` or something similar sounds better 😄

barklund: Changed in d020e80139.

kkalarickal:
Tested on [stories-qa-server](https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/post.php?post=15625&action=edit#page=%2522fa4e3b4e-acd6-4261-a4ae-eec722f1c99e%2522) using #9945 code base

Status: Passed with Comments ❓ 

Notes:
- problem associated with #9989 was observed but ignored since there is a fix for it ✔️ 
- Trim it to a new length and observe that the trim button in the design panel is disabled while trimming ✔️ 
image

- There was new console error observed with trimming ❌ 
image

- Trim the newly created video again and observe that the trim button in the design panel is once again disabled while trimming ✔️ 
- Start trimming the new-newly created video and observe, that the trim handles are correctly placed as you placed them after the second trim. ✔️

Side Effects:
- When a video is trimmed, it creates a new upload entry in the media library ✔️ 
- However, when the 1st trimmed video is re-trimmed (i.e. a 2nd trim), the 1st embedded video is *ALSO* trimmed ❓ 
- As a end user, if I trim a new embed, i would expect the original embed to retain its original trim. but it looks like the same reference is maintained and a re-re-trim affects both embeds ❗   
- In a similar fashion, if I duplicate a page containing a video and trim an embed, the video on both pages are trimmed. ❗

barklund:
@kkalarickal (cc @swissspidy, @miina)
> There was a new console error observed with trimming

Hmm, not seeing this locally. Not sure why it only happens in the QA environment. But it seems like this also happened before, no? If I go to the QA env on the main branch and (re)trim any video, I'm seeing the same console error.

> However, when the 1st trimmed video is re-trimmed (i.e. a 2nd trim), the 1st embedded video is _ALSO_ trimmed ❓

As far as I can see, that is not a new bug introduced by this PR? And it even exists for a non-trimmed video as well:

1. Add a completely new video (dragged from desktop) to a page
2. Add one more instance of the same video (dragged from the library)
3. Trim one instance
4. Observe that both are now trimmed once completed

So I don't think neither of these bugs should hold up this PR. Though I agree both are bugs, the latter being the more important one.

swissspidy:
There are tons of outstanding media-related PRs which might inadvertently fix that bug.

Let's re-evaluate once they're merged.

swissspidy:
Ticket created: #10409

PR #9946, opened 2021-12-08
Media3p: flakey karma tests

### LINKED ISSUE #9928: Flaky Karma tests related to Media Fetching
## Bug Description
The following Karma tests are flaky:
`Media3pPane fetching Gallery navigation should handle pressing left when focused` ([Example run](https://github.com/google/web-stories-wp/runs/4441826329?check_suite_focus=true#step:6:14303))
`Media3pPane fetching Gallery navigation should handle pressing up`
`Media3pPane fetching Gallery navigation should handle pressing Home` [Example run](https://github.com/google/web-stories-wp/runs/4437212276?check_suite_focus=true#step:6:386)
`should handle pressing right when at the end of a row`

Looks like the Media is not always loaded by the time the tests start running.

Also seeing occasional failures with `should render the second media provider`:
`Error: Not ready: 40 != 20`

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

BrookeGraham:
## Context

Media fetching tests can be flakey. The Media3p doesn't always load before the first expect statement causing the tests to either throw an error, or the mediaElements are not found. 

## Summary

Once the media tab is clicked we want to wait for `mediaElements` to load before proceeding with the tests. 

## Relevant Technical Choices
Decided to `await` a `waitFor` and throw an error if the elements haven't loaded, this way the test will wait up to 5 seconds before proceeding. However, as soon as the elements are load we can continue on without having to take up the entire 5 second timeout. 

## User-facing changes
na

## Testing Instructions

1.

na

na

na

## Checklist

---

Fixes #9928

### DIFF
--- a/packages/story-editor/src/components/library/panes/media/media3p/karma/mediaFetching.karma.js
+++ b/packages/story-editor/src/components/library/panes/media/media3p/karma/mediaFetching.karma.js
@@ -264,6 +264,21 @@ describe('Media3pPane fetching', () => {
     expect(mediaElements?.length).toBe(expectedCount);
   }
 
+  async function waitForInitialMediaLoad() {
+    await waitFor(
+      () => {
+        const mediaElements = within(
+          fixture.editor.library.media3p.unsplashSection
+        ).queryAllByTestId(/^mediaElement/);
+
+        if (!mediaElements || mediaElements.length === 0) {
+          throw new Error('mediaElements did not load');
+        }
+      },
+      { timeout: 5000 }
+    );
+  }
+
   it('should render no results message', async () => {
     listMediaSpy.and.callFake(() => ({ media: [] }));
     await fixture.events.click(fixture.editor.library.media3pTab);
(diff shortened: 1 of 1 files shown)


BrittanyIRL (approved):
Nice!

packages/story-editor/src/components/library/panes/media/media3p/karma/mediaFetching.karma.js
```diff
@@ -333,7 +353,8 @@ describe('Media3pPane fetching', () => {
       0,
       mediaGallery.scrollHeight - mediaGallery.clientHeight - ROOT_MARGIN / 2
     );
-
+    // Wait for debounce
+    await fixture.events.sleep(700);
```
samwhale: 💭 I think the fetching function is spied-on using a jasmine spy. Using this, we could wait for it to be called rather than waiting 700 ms. This would be a little more resilient, even on a machine that's a little slower

This sleep happens in 3 places. Would be dope if we could remove it!

spy: https://github.com/google/web-stories-wp/blob/main/packages/story-editor/src/components/library/panes/media/media3p/karma/mediaFetching.karma.js#L209

May look something like
```
await waitFor(() => {
    expect(listMediaSpy).toHaveBeenCalledTimes(2);
});
```

samwhale (approved):
Nice, this is great! One little suggestion.

packages/story-editor/src/components/library/panes/media/media3p/karma/mediaFetching.karma.js
BrookeGraham: Yes I love it! Adding.

PR #9949, opened 2021-12-09
Decoupling: Make some `story-editor` api callbacks optional

mohdsayed:
## Context
This PR makes some API callbacks optional to simplify the standalone `story-editor` configuration.

## Summary

After closely looking at the API callbacks individually, I realized that a lot of the API callbacks are anyways optional because some will not be called when a certain capability is not provided, and some will not be called because they are integrated from the `wp-story-editor`. 

## Testing Instructions

1.

## Checklist

---


mohdsayed:
@swissspidy I think I was unnecessarily worried about these optional callbacks initially. After all the recent decoupling changes, it looks a lot easier now. I think we can remove some more.

For example, for a standalone editor, this save template wouldn't be required as there is no place to save template and therefore make its API callback optional.

image

packages/story-editor/src/app/story/effects/useLoadStory.js
```diff
@@ -56,7 +56,9 @@ function loadStory(storyId, post, restore, clearHistory) {
     terms,
   } = post;
 
-  const [prefix, suffix] = permalinkTemplate.split(/%(?:postname|pagename)%/);
+  const [prefix, suffix] = permalinkTemplate
+    ? permalinkTemplate.split(/%(?:postname|pagename)%/)
+    : '';
```
swissspidy: Intuitively, I would have used `[]` instead of `''`

packages/story-editor/src/app/media/utils/useDetectBaseColor.js
```diff
@@ -86,7 +86,9 @@ function useDetectBaseColor({ updateMediaElement }) {
       if (type === 'image') {
         imageSrc = getSmallestUrlForWidth(0, resource);
       } else if (!isExternal) {
-        const posterResource = await getPosterMediaById(id);
+        const posterResource = getPosterMediaById
+          ? await getPosterMediaById(id)
+          : '';
```
swissspidy:
```suggestion
          : null;
```

Seems more intuitive to use `null` instead of an empty string to indicate that this value is absent.

packages/story-editor/src/app/currentUser/currentUserProvider.js
```diff
@@ -33,7 +33,7 @@ function CurrentUserProvider({ children }) {
   } = useAPI();
 
   useEffect(() => {
-    if (!Object.keys(currentUser).length) {
+    if (getCurrentUser && !Object.keys(currentUser).length) {
```
swissspidy, line 36: Would it make sense for `getCurrentUser` to default to a function that returns an empty object or `null` or something?

packages/story-editor/src/app/story/effects/useLoadStory.js
```diff
@@ -30,16 +30,16 @@ import getUniquePresets from '../../../utils/getUniquePresets';
...
 function loadStory(storyId, post, restore, clearHistory) {
   const {
-    title: { raw: title },
+    title: { raw: title = '' } = '',
     status,
     slug,
     date,
     modified,
-    excerpt: { raw: excerpt },
+    excerpt: { raw: excerpt = '' } = '',
```
swissspidy:
```suggestion
    excerpt: { raw: excerpt = '' } = {},
```

Both work the same I suppose.

packages/story-editor/src/app/story/effects/useLoadStory.js
```diff
@@ -30,16 +30,16 @@ import getUniquePresets from '../../../utils/getUniquePresets';
 
 function loadStory(storyId, post, restore, clearHistory) {
   const {
-    title: { raw: title },
+    title: { raw: title = '' } = '',
```
swissspidy:
```suggestion
    title: { raw: title = '' } = {},
```

Both work the same I suppose

packages/story-editor/src/components/library/libraryProvider.js
```diff
@@ -44,7 +44,13 @@ const LIBRARY_TAB_IDS = new Set(
...
 
 function LibraryProvider({ children }) {
-  const [tab, setTab] = useState(MEDIA.id);
+  const { showMedia3p } = useConfig();
+  const {
+    actions: { getMedia },
+  } = useAPI();
+  const showMedia = Boolean(getMedia); // Do not show media tab if getMedia api callback is not provided.
+  const defaultCurrentTabId = showMedia ? MEDIA.id : MEDIA3P.id;
+  const [tab, setTab] = useState(defaultCurrentTabId);
```
swissspidy: This won't work if `showMedia3p` is false.

Might be easier to just default to `const [tab, setTab] = useState(tabs[0])`

packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
```diff
@@ -78,6 +78,7 @@ function PageTemplatesPane(props) {
   const {
     actions: { getCustomPageTemplates },
   } = useAPI();
+  const canShowSaveTemplate = Boolean(getCustomPageTemplates);
```
swissspidy: Maybe rename to `canSaveTemplate` or `supportsCustomTemplates`

packages/story-editor/src/app/currentUser/currentUserProvider.js
mohdsayed: You mean like this `() => Promise.resolve( {} );` . But why only for `getCurrentUser` ?

packages/story-editor/src/components/library/libraryProvider.js
mohdsayed: Good idea, updated, thanks.

packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
mohdsayed: I couldn't come up with a better name, so I was expecting it in the code review 🙂 . `supportsCustomTemplates` looks good 👍

packages/story-editor/src/app/currentUser/currentUserProvider.js
swissspidy: Yeah. And that is a good question 🙂   Doesn't really make sense now that I think about it.

I was just wondering whether this one should be considered mandatory, but we don't use it for any critical stuff in the editor, so should be OK.

swissspidy (approved):
LGTM, pending tests

Issue #9880, opened 2021-12-03
Custom Fonts: Allow inserting custom fonts via URL


miina:
Note: The UX of this issue is still in progress and the feature might change a little bit
## Feature Description
Figma link to be added once ready
- [ ] Add the necessary input(s) to the Dashboard Settings for inserting custom font via URL
- [ ] Extract the font metrics and save the URL and relevant data about the URL (needs to be checked which library to use, see https://github.com/google/web-stories-wp/issues/4717#issuecomment-981943916)
- [ ] Display the custom fonts in a separate section in the Font Picker
- [ ] Allow removing fonts
- [ ] Consider invalid URLs

Note: This issue does NOT include any API work and assumes that all that part is already in place.

miina:
Although this issue can't be completed before the API is ready, we could already do the following using dummy data before to make things quicker once the API is done:
- Implement the UX
- Implement extracting the font metrics utils (find the suitable library)
- Add a new section to the Font Picker using dummy URLs

swissspidy:
Yes, a lot of this is not blocked by the API work at all.

kkalarickal:
Tested this on https://stories-qa-wordpress-amp.pantheonsite.io/ running PR #9966 

Status: LGTM 👍 

Scenarios:
1. Confirmed experimental feature flag (on/off) ✔️ 
2. Confirmed success case for adding custom font via URL ✔️ 
3. Confirmed negative case of trying to add woff2 font or using a font URL that is behind CORS ✔️ 
4. Confirmed ability to use the said custom font inside the editor ✔️ 
5. Confirmed deleting font removes it from the available font list in editor ✔️

kkalarickal:
@miina ... some things to note that might need future polish:

1. when I use the custom font in the editor, there is some weird flickering. I saw it when i changed color of the custom font.... like there is a temp font style applied and then the custom one. this is not consistently reproducible, but it only happened for the custom font.
2. when a custom font is applied to some text in the story, and then the custom font is deleted, the original text continues to show the deleted custom font even though the drop down no longer has it.

swissspidy:
> when a custom font is applied to some text in the story, and then the custom font is deleted, the original text continues to show the deleted custom font even though the drop down no longer has it.

That is expected until we have some mechanism in place to notify users about deleted fonts. See #10014

PR #9952, opened 2021-12-09
Fonts: Add baseline support for custom fonts in editor

### LINKED ISSUE #9856: Fonts: Store Custom Fonts' data
## Feature Description
As part of allowing the user to add Custom Fonts via URLs, we also need a way to store the font URLs and the font metrics in the DB.

From discussion with Pascal:
What we probably want to do is
1. Store font information in database (`wp_options` or `wp_posts`)
2. Maybe create a new REST API endpoint for managing the custom fonts (CRUD)
3. Create a new REST API endpoint for retrieving a list of all fonts (custom fonts + fonts.json) for the font picker, so the editor only has 1 data source for that

Note: this issue is for the infra only and does not include the actual interface of adding fonts,  calculating the necessary font metrics nor displaying the custom fonts in the font picker.

## Additional Context
See #4717 for the user-side implementation.

swissspidy:
## Context

We need the ability to create, read, and delete custom fonts. In the editor, we want to surface these custom fonts alongside the built-in list of Google fonts.

## Summary

This PR implements the following changes:

* New `web-story-font` post type for storing custom fonts
* New `/wp-json/web-stories/v1/fonts` REST API endpoint for getting all fonts
  It accepts params like `include`, `search`, and `service` to filter the list as needed. No pagination added because we don't really need it and it would increase complexity quite a bit.
* Updates font provider in the editor to load appropriate styles for custom fonts
* Updates story output to load appropriate styles for custom fonts as well
* In Karma tests, the list of fonts continues to be mocked to include only a few.
* Updates storybook to continue loading Google fonts there

This PR does **not** make any visual changes to the font picker, such as adding a separate section for the custom fonts.

For testing purposes, one can add some custom fonts to the database like so:

```bash
$ wp post create --post_type=web-story-font --post_status=publish --post_title="Overpass Regular" --post_content='{"family":"Overpass Regular","fallbacks":["sans-serif"],"weights":[400],"styles":["regular"],"variants":[[0,400]],"url":"https://overpass-30e2.kxcdn.com/overpass-regular.ttf","service":"custom","metrics":{"upm":1000,"asc":982,"des":-284,"tAsc":750,"tDes":-250,"tLGap":266,"wAsc":1062,"wDes":378,"xH":511,"capH":700,"yMin":-378,"yMax":1062,"hAsc":982,"hDes":-284,"lGap":266}}'

$ wp post create --post_type=web-story-font --post_status=publish --post_title="Gingrich Small Regular" --post_content='{"family":"Gingrich Small Regular","fallbacks":["sans-serif"],"weights":[400],"styles":["regular"],"variants":[[0,400]],"url":"https://www.barrons.com/fonts/woffs/gingrich/GingrichSmall-Regular.woff","service":"custom","metrics":{"upm":1000,"asc":830,"des":-170,"tAsc":830,"tDes":-170,"tLGap":200,"wAsc":1060,"wDes":226,"xH":522,"capH":750,"yMin":-226,"yMax":1060,"hAsc":830,"hDes":-170,"lGap":200}}'

$ wp post create --post_type=web-story-font --post_status=publish --post_title="Vazir Regular" --post_content='{"family":"Vazir Regular","fallbacks":["sans-serif"],"weights":[400],"styles":["regular"],"variants":[[0,400]],"url":"https://cdn.jsdelivr.net/gh/rastikerdar/vazir-font@v30.1.0/dist/Vazir-Regular.ttf","service":"custom","metrics":{"upm":2048,"asc":2200,"des":-1100,"tAsc":2200,"tDes":-1100,"tLGap":0,"wAsc":2200,"wDes":1100,"xH":1082,"capH":1638,"yMin":-1116,"yMax":2163,"hAsc":2200,"hDes":-1100,"lGap":0}}'
```

Note that Gingrich Small doesn't have the required CORS headers, so the font won't display.

## Relevant Technical Choices

The new fonts REST controller is very simple and only really adds what is needed for dealing with (custom) fonts.

## To-do

* [x] Ensure font creation and deletion works
* [x] Preload styles when displaying custom fonts in the font dropdown
* [x] Tests

## User-facing changes

None yet.

## Testing Instructions

1. Verify that the fonts dropdown in the design panel is still working as intended

New REST API endpoint, but only available to users who can manage options or edit stories.

No

No

## Checklist

---

Fixes #9856


swissspidy:
@miina @spacedmonkey This is an early version of custom fonts support in the REST API, in case you'd like to take a look already.

I totally overdid it, as I also added support for custom fonts in the editor to make sure the API endpoint works as expected.

packages/fonts/src/test/curatedFonts.js
```diff
@@ -1,44 +0,0 @@
...
- */
-import { readFileSync } from 'fs';
-import { resolve } from 'path';
-
-/**
- * Internal dependencies
- */
-import { curatedFontNames } from '..';
-
-describe('Curated fonts', () => {
```
swissspidy, line 28: Maybe I'll revisit this again.

I initially moved the list of curated fonts to the story-editor, but there this test obviously doesn't work.

Perhaps it's better to move it back.

packages/story-editor/src/karma/fixture/fixture.js
```diff
@@ -948,6 +821,220 @@ class APIProviderFixture {
         []
       );
 
+      const getFonts = useCallback(() => {
```
swissspidy: Todo: add support for search.

includes/Font_Post_Type.php
```diff
@@ -0,0 +1,164 @@
...
+			'delete_private_posts'   => 'manage_options',
+			'delete_published_posts' => 'manage_options',
+			'delete_others_posts'    => 'manage_options',
+			'edit_private_posts'     => 'manage_options',
+			'edit_published_posts'   => 'manage_options',
+			'create_posts'           => 'manage_options',
+		];
+
+		return [
+			'labels'                => [
```
swissspidy: We could just get rid of the labels since the post type is not user-facing.

packages/story-editor/src/components/panels/design/textStyle/fontPicker.js
```diff
@@ -121,6 +119,7 @@ const FontPicker = forwardRef(function FontPicker(
     return map;
   }, [fonts]);
 
+  // TODO: Add support for custom fonts as well.
```
miina: Thanks for pushing the PR in WIP state, I can take a more thorough look in a few hours. Just to confirm -- this TODO of adding the custom fonts to the Dropdown here is intended for #9880, right? I was planning to look into this today (unless you already secretly did it yourself).

swissspidy: This TODO is for enqueueing font stylesheets for custom fonts when displaying them in the dropdown.

The custom fonts themselves are already visible and searchable in the dropdown.

swissspidy: What‘s left to be done in that other ticket is adding a dedicated section for the custom fonts though.

miina: Yep, that dedicated section was what I had in mind for looking into :)

packages/story-editor/src/utils/getFontCSS.js
```diff
@@ -0,0 +1,61 @@
...
+  @font-face {
+    font-family: "${name}";
+    src: url('${src}') format('${format}');
+    font-weight: normal;
+    font-display:swap;
+  }
+  `;
+}
+
+export default getFontCSS;
```
miina: Should these go to the fonts package instead?

miina:
@swissspidy Anything in addition to building + composer update I have to do to test this out?
Getting `rest_no_route` error currently for `https://amp.wordpress.test/wp-json/web-stories/v1/fonts/?per_page=100&include=Karla%2CLato...`

The fonts endpoint file gets loaded and `register_routes` runs when loading the editor (can stop the code there when debugging).

swissspidy:
@miina The route is not public for everyone and is only available for logged in users with the capability to edit stories.

packages/story-editor/src/utils/getFontCSS.js
swissspidy: Yeah maybe. I need to get a better sense for what should be extracted into a different package. With #8778 it probably makes sense to move this util.

miina:
> The route is not public for everyone and is only available for logged in users with the capability to edit stories.

Hm, I'm logged in and in the editor (with my admin user), so it should work then, right? The `rest_no_route` error happens when opening the editor, not trying to make a request directly to the endpoint from browser.

swissspidy:
Yeah it should. And actually I just saw that it still is public, which I'd need to fix. Have you tried `composer install`?

miina:
> Yeah it should. And actually I just saw that it still is public, which I'd need to fix. Have you tried composer install?

Yep, tried that. Hm, not sure what can be the issue. @spacedmonkey Does it work for you without issues?

(Getting `Could not apply patch! Skipping. The error was: The "patches/amp-wp-allow-amp-story-captions.patch" file could not be downloaded: Failed to open stream: No such file or directory` when doing `composer install` but that doesn't seem to be related at all).

Edit: Figured the issue out, looks like `$this->namespace = 'web-stories/v1';` is missing from registering the route and it uses `wp/v2` instead.

includes/REST_API/Font_Controller.php
```diff
@@ -0,0 +1,470 @@
...
+	/**
+	 * Registers the routes for posts.
+	 *
+	 * @since 1.16.0
+	 *
+	 * @see register_rest_route()
+	 *
+	 * @return void
+	 */
+	public function register_routes() {
```
spacedmonkey, line 69: Why do we need a custom register route here? Isn't use the same as the one found in `WP_REST_Posts_Controller`

includes/REST_API/Font_Controller.php
```diff
@@ -0,0 +1,470 @@
...
+
+	/**
+	 * Returns a list of Google fonts.
+	 *
+	 * @since 1.16.0
+	 *
+	 * @return array List of Google fonts.
+	 */
+	protected function get_builtin_fonts() {
+		$file = WEBSTORIES_PLUGIN_DIR_PATH . 'includes/data/fonts/fonts.json';
```
spacedmonkey, line 118: Maybe we should convert this to a PHP file with an array for performance reasons.

includes/REST_API/Font_Controller.php
```diff
@@ -0,0 +1,470 @@
...
+	public function prepare_item_for_response( $item, $request ): WP_REST_Response {
+		// Restores the more descriptive, specific name for use within this method.
+		$post            = $item;
+		$GLOBALS['post'] = $post; // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
+
+		setup_postdata( $post );
+
+		$data = [];
+
+		$data['id']     = $post->ID;
```
spacedmonkey:
```suggestion
		if ( rest_is_field_included( 'id', $fields ) ) {
			$data['id'] = $post->ID;
		}
```

includes/REST_API/Font_Controller.php
```diff
@@ -0,0 +1,470 @@
...
+		$data['family'] = $post->post_title;
```
spacedmonkey:
```suggestion
		if ( rest_is_field_included( 'family', $fields ) ) {
			$data['family'] = $post->post_title;
		}
```

includes/REST_API/Font_Controller.php
```diff
@@ -0,0 +1,470 @@
...
+
+		/**
+		 * Font data.
+		 *
+		 * @var array|null $font_data
+		 */
+		$font_data = json_decode( $post->post_content, true );
+
+		if ( $font_data ) {
+			$data = array_merge( $data, $font_data );
```
spacedmonkey:
```suggestion
			$raw_data = array_merge( $data, $font_data );
			$output = [];
			foreach( $raw_data as $key => $value ){
			    if ( rest_is_field_included( $key, $fields ) ) {
			           $output[$key] = $value;
		            }
			} 
```

packages/story-editor/src/utils/loadStylesheet.js
```diff
@@ -14,11 +14,13 @@
...
  */
 
-function loadStylesheet(url) {
+function loadStylesheet(url, id) {
   return new Promise((resolve, reject) => {
     const link = document.createElement('link');
+    link.id = id;
     link.rel = 'stylesheet';
     link.href = url;
+    link.crossOrigin = 'anonymous';
```
spacedmonkey, line 23: Is there error handling for font with cross origin issues?

includes/REST_API/Font_Controller.php
```diff
@@ -0,0 +1,470 @@
...
+	 * Registers the routes for posts.
+	 *
+	 * @since 1.16.0
+	 *
+	 * @see register_rest_route()
+	 *
+	 * @return void
+	 */
+	public function register_routes() {
+		register_rest_route(
```
spacedmonkey, line 70: Will we support batching here?

includes/REST_API/Font_Controller.php
swissspidy: I thought about that too. I'll do some micro benchmarking to see how big of an impact that would make.

packages/story-editor/src/utils/loadStylesheet.js
swissspidy: It will not be possible for users to add new fonts with CORS issues. They will be prevented from doing that on the settings page.

includes/REST_API/Font_Controller.php
swissspidy: There are no `GET` (`get_item`) and `POST` (`update_item`) routes as we don't need them.

Trying to make this controller as simple as possible and only adding the stuff we need.

includes/REST_API/Font_Controller.php
swissspidy: There are no UI features that would require any batching capabilities, so no.

includes/REST_API/Font_Controller.php
swissspidy: Quick test measuring how long it takes to load the fonts list:

Loading and parsing a JSON file: avg. 0.005 seconds

From PHP file returning an array: avg 0.00005 seconds

While the PHP file is a tad faster, it doesn't seem worth changing this at the expense of a more complicated build step.

spacedmonkey: Tad faster, that seems a lot faster...

swissspidy: 0.005 is still fast enough. That's only 5 milliseconds.

packages/story-editor/src/utils/loadStylesheet.js
spacedmonkey: This is documented in a ticket anywhere?

.storybook/stories/playground/story-editor/api/fonts.js
```diff
@@ -0,0 +1,41 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export async function getFonts(params) {
+  let fonts = await import(
+    /* webpackChunkName: "chunk-fonts" */ '@web-stories-wp/fonts/src/fonts.json'
+  ).then((newFonts) =>
```
spacedmonkey: Await and then?

includes/Admin/Editor.php
```diff
@@ -390,6 +390,7 @@ public function get_editor_settings(): array {
 				'link'           => '/web-stories/v1/link/',
 				'statusCheck'    => '/web-stories/v1/status-check/',
 				'taxonomies'     => '/web-stories/v1/taxonomies/',
+				'fonts'          => '/web-stories/v1/fonts/',
```
spacedmonkey:
```suggestion
				'fonts'          =>trailingslashit( $this->font_post_type->get_rest_url() ),
```

packages/fonts/src/utils/getFontCSS.js
```diff
@@ -0,0 +1,59 @@
...
+function getFontFormat(src) {
+  const fileExtension = src.split(/[#?]/)[0].split('.').pop().trim();
+
+  switch (fileExtension) {
+    case 'woff':
+    case 'woff2':
+      return fileExtension;
+    case 'ttf':
+      return 'truetype';
+    default:
```
miina, line 34: Is `.otf` not supported?

miina (commented):
LGTM generally, found one issue from testing:
- When using a custom font on a text and then switching to another font, I get the following error message:
Screenshot 2021-12-14 at 17 56 04

Steps:
1. Assign Vazir Regular as the font (assuming it's added to the DB)
2. Now assign some other font (I used Bungee from the default list).

I suppose that since it's not possible to add custom fonts, there's no need for a feature flag either in this PR.

packages/fonts/src/utils/getFontCSS.js
swissspidy: It should. Just an oversight.

miina:
@swissspidy Is loading the fonts in the font picker working as expected for you currently? Looks like the observer is not working as expected for me, only the first fonts get loaded, when scrolling down, the others remain with default font:
Screenshot 2021-12-15 at 09 53 30

includes/REST_API/Font_Controller.php
```diff
@@ -0,0 +1,586 @@
...
+					'description' => __( 'Font URL.', 'web-stories' ),
+					'type'        => 'string',
+					'format'      => 'uri',
+					'context'     => [ 'view', 'edit', 'embed' ],
+					'readonly'    => true,
+				],
+			],
+		];
+
+		return $schema;
```
spacedmonkey:
```suggestion
		$this->schema = $schema;

		return $this->add_additional_fields_schema( $this->schema );
```

spacedmonkey:
Question: Should the font api be preloaded?

includes/REST_API/Font_Controller.php
```diff
@@ -0,0 +1,586 @@
...
+	 *
+	 * Overrides the namespace.
+	 *
+	 * @since 1.16.0
+	 *
+	 * @param string $post_type Post type.
+	 */
+	public function __construct( $post_type ) {
+		parent::__construct( $post_type );
+		$this->namespace = 'web-stories/v1';
```
spacedmonkey:
```suggestion
		$obj             = get_post_type_object( $post_type );
		$this->namespace = ! empty( $obj->rest_namespace ) ? $obj->rest_namespace : 'web-stories/v1';
```

includes/REST_API/Font_Controller.php
```diff
@@ -0,0 +1,586 @@
...
+
+	/**
+	 * Prepares a single post for create.
+	 *
+	 * @since 1.16.0
+	 *
+	 * @param WP_REST_Request $request Request object.
+	 * @return stdClass|WP_Error Post object or WP_Error.
+	 */
+	protected function prepare_item_for_database( $request ) {
```
spacedmonkey, line 422: Should we check that font family exists before saving. I am worried that someone will try and create two fonts with the same name. Either by uploading two fonts with the same name or by adding one that already appears in list of fonts.

spacedmonkey:
Got some questions, but we are looking good here.

swissspidy:
> Question: Should the font api be preloaded?

Not worth it IMHO but not possible anyway.
The first call to the REST API is for the list of curated fonts. We don't have that list available in PHP.

includes/REST_API/Font_Controller.php
swissspidy: Sounds reasonable 👍

swissspidy:
> @swissspidy Is loading the fonts in the font picker working as expected for you currently? Looks like the observer is not working as expected for me, only the first fonts get loaded, when scrolling down, the others remain with default font: Screenshot 2021-12-15 at 09 53 30

@miina That's odd. I'll take a look. But you might be more familiar with that observer, so would appreciate you taking a look as well.

miina:
> But you might be more familiar with that observer, so would appreciate you taking a look as well.

Checking now.

spacedmonkey (approved):
LGTM

miina (approved):
🎉

PR #9953, opened 2021-12-09
Media: Generate and backfill BlurHash for media items

### LINKED ISSUE #9685: Media: Generate and backfill BlurHash for media items

spacedmonkey:
## Summary

Version 2 of this PR.  Version 1 https://github.com/google/web-stories-wp/pull/9935

Generate blurhashs for all media and existing media elements using the blurhash library. As the blurhash is slow process, use a web worker, to do the calculation in another thread.  

## Testing Instructions

1. Create a new story. 
2. Upload a video or image. 
3. Wait until upload is complete. 
4. Wait for 20 seconds. 
5. Reload the editor. 
6. See blurhash for newly uploaded asset. 

## Checklist

---

Fixes #9685


webpack.config.cjs
```diff
@@ -105,6 +105,15 @@ const sharedConfig = {
         use: ['source-map-loader'],
         enforce: 'pre',
       },
+      {
+        test: /\.worker\.js$/,
+        use: {
+          loader: 'worker-loader',
+          options: {
+            inline: 'fallback',
```
spacedmonkey, line 113: This is required, as it convert worker a blob, to get around cross origin issues. [See](https://github.com/webpack-contrib/worker-loader#cross-origin-policy)

swissspidy:
So work is properly offloaded to a worker now?

Any insights into how well this performs, how long it takes?

`transferControlToOffscreen` didn't work or was not needed?

spacedmonkey:
@swissspidy In my testing, the web worker worked in firefox, safari and chrome. It offloaded out of the main thread and processed in that new thread. 

In my testing, blurhashes took around 300ms to 3 seconds to process, depending on size. But as it is very async now, it does make the editor slower. It just takes as long as it takes. 

> transferControlToOffscreen didn't work or was not needed?

As this works in many browsers, is this needed?

swissspidy:
> But as it is very async now, it does make the editor slower.

So the editor interaction is more sluggish despite the async processing in a worker? 🤨

spacedmonkey:
>But as it is very async now, it does make the editor slower.

* which doesn't make the editor slower.

packages/story-editor/src/app/media/local/useContextValueProvider.js
```diff
@@ -210,7 +217,24 @@ export default function useContextValueProvider(reducerState, reducerActions) {
...
-    [setBaseColorProcessing, removeBaseColorProcessing, updateBaseColor]
+    [setBaseColorProcessing, updateBaseColor, removeBaseColorProcessing]
+  );
+
+  const processMediaBlurhash = useCallback(
+    (resource) => {
+      const { blurHashProcessed, blurHashProcessing } = stateRef.current;
+      const { id } = resource;
+      (async () => {
+        // Don't process more than one item at a time and do not process items already processed/processing.
```
swissspidy: This comment seems outdated now since it's not limited to one item at a time anymore.

packages/story-editor/src/utils/getBlurHashFromImage.js
```diff
@@ -0,0 +1,68 @@
...
+      image: data,
+      width,
+      height,
+      compontentX: 4,
+      compontentY: 4,
+    });
+    worker.addEventListener('message', function (event) {
+      worker.terminate();
+      if (event.data.type === 'success') {
+        trackTiming();
```
swissspidy: Let's move this one line up so it also stops when rejecting.

packages/story-editor/src/utils/getBlurHashFromImage.js
```diff
@@ -0,0 +1,68 @@
...
+        resolve(event.data.blurHash);
+      } else {
+        reject(event.data.error);
+      }
+    });
+    worker.addEventListener('error', (e) => {
+      reject(e);
```
swissspidy, line 68: `worker.terminate()`?

swissspidy: Also: `trackError('blurhash_generation', e.type);`

swissspidy: Also maybe `trackTiming()` as well

swissspidy (approved):
Overall works like a charm! Just a couple of nits.

packages/story-editor/src/utils/getBlurHashFromImage.js
spacedmonkey: Fixed in https://github.com/google/web-stories-wp/pull/9953/commits/88162dcec1faa48a1ae156e711d9ca15e2a09aaf

swissspidy:
Can't hurt to run this past QA too

packages/story-editor/src/utils/getBlurHashFromImage.js
spacedmonkey: Fixed

spacedmonkey:
@swissspidy Any idea what the issue with  LGTM.com?

swissspidy:
You can click on the "View on LGTM.com" link to see all the errors. 

It seems to flag the `new Worker` usage because the worker file has no default export due to how webpack-loader works.

In this case I'd consider this a false positive and we could [suppress this message](https://help.semmle.com/lgtm-enterprise/user/help/alert-suppression.html#javascript) using a `// lgtm[js/call-to-non-callable]` comment.

PR #9951, opened 2021-12-09
Fix `findDOMNode` warnings that clutter the console

amovar18:
## Context

`findDOMNode` warnings take up a lot of space in the console and make it difficult to read anything in console. Since `findDOMNode` has been deprecated in `StrictMode` it would be better to move away from it. This problem occurs due to improper props for `CSSTransition` component

## Summary

This PR corrects the `nodeRef` prop to the `CSSTransition` component and its immediate child.

## Testing Instructions

- [x] This is a non-user-facing change and requires no QA

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [x] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [x] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #9950

PR #9959, opened 2021-12-10
Hotfix `scheduledtransition` not able to get ref

### LINKED ISSUE #9950: `findDOMNode` warnings make it difficult to debug
## Bug Description

`finDOMNode` warnings are taking up a lot of console space and make it difficult to debug if we want to debug or log something. With `findDOMNode` being deprecated in `StrictMode` it might be better to solve this issue.

## Expected Behaviour

`findDOMNode`  warning shouldn't be logged in the console.

## Additional Context

- WordPress Version: 5.8.2
- Operating System: MacOs monterrey (m1)
- Browser: Chrome

amovar18:
## Context
After recent PR #9951 got merged we see an error in console 
image

## Summary

It came to light that `ScheduledTransition` was not able to get the ref of its child. The `ScheduledTransition` uses the component `Transition` from the `react-transition-group` library. `Transition` requires the `nodeRef` for detecting which element has to be transitioned ([reference link](https://reactcommunity.org/react-transition-group/transition)). Since `Interpreter` component just returned children, there is no way possible to attach `ref` to the children in the same file. Therefore we need to clone the element and pass `ref` as config. The other way is to pass `nodeRef` to `ScheduledTransition` and `ref` to the first child of `ScheduledTransition`.

## Testing Instructions

## Checklist

---

Fixes #9950
Related #9951

### DIFF
--- a/packages/story-editor/src/components/transition/scheduledTransition.js
+++ b/packages/story-editor/src/components/transition/scheduledTransition.js
@@ -16,7 +16,14 @@
 /**
  * External dependencies
  */
-import { useReducer, useEffect, useState, useRef } from '@web-stories-wp/react';
+import {
+  useReducer,
+  useEffect,
+  useState,
+  useRef,
+  forwardRef,
+  cloneElement,
+} from '@web-stories-wp/react';
 import PropTypes from 'prop-types';
 import { Transition } from 'react-transition-group';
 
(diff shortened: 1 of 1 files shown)


maxyinger (approved):
I was originally thinking to just pass the ref in scheduled transition directly to the child cus then `ScheduledTransition` mirrors the same API as `Transition` and just alters the state passed to children as function. (and you'd have to do this with `CSSTransition` or `Transition` if we were directly using them)

However, I guess I see no issues with this approach being a little opinionated with how we're currently using it.

I think the only thing that this approach removes is the ability to pass a ref to the child since the cloned child ref will overwrite any original ref present:

Example:
https://codesandbox.io/s/lucid-bash-r3jgh?file=/src/App.js

React Docs:
Screen Shot 2021-12-10 at 10 26 31 AM

We're not currently passing any refs to the direct child in any instance of `ScheduledTransition`, but if we were to, this might be a little opaque as to why it's not receiving any refs.

Anyway, I think this is fine cus I imagine our use of `ScheduledTransition` is relatively stable, but something to make note of and keep an eye out for.

Thanks for the fix. Always love less errors in the console 🎉

PR #9960, opened 2021-12-10
Design System: Context Menu Refactor

### LINKED ISSUE #9750: Design System: Refactor context menu
## Feature Description

The context menu in the design system generates a menu from a list of objects. At first this was thought to be easier since we could create a context menu without needing to build out too much JSX.

However, with the introduction of the `MediaPicker` it has become apparent that it would become easier to define whatever UI we'd like in the context menus.

## Alternatives Considered

n/a

## Additional Context

Ex: The context menu in the [radix-ui](https://www.radix-ui.com/docs/primitives/components/context-menu#anatomy) builds a context menu out of smaller pieces every time it's used. This seems to be a better approach in the event that we'd want to wrap individual items with something (such as a media picker)

```tsx
import * as ContextMenu from '@radix-ui/react-context-menu';
export default () => (

  <ContextMenu.Root>

    <ContextMenu.Trigger />
    <ContextMenu.Content>

      <ContextMenu.Label />

      <ContextMenu.Item />
      <ContextMenu.Group>

        <ContextMenu.Item />

      </ContextMenu.Group>
      <ContextMenu.CheckboxItem>

        <ContextMenu.ItemIndicator />

      </ContextMenu.CheckboxItem>
      <ContextMenu.RadioGroup>

        <ContextMenu.RadioItem>

          <ContextMenu.ItemIndicator />

        </ContextMenu.RadioItem>

      </ContextMenu.RadioGroup>
      <ContextMenu.Root>

        <ContextMenu.TriggerItem />

        <ContextMenu.Content />

      </ContextMenu.Root>
      <ContextMenu.Separator />

    </ContextMenu.Content>

  </ContextMenu.Root>

);
```

samwhale:
## Context

A while back the functionality to replace media using the `MediaPicker` which is injected into the application - #9534 

This PR had a lot of back and forth - and through the conversation we realized that the current structure of the context menu isn't flexible enough. @littlemilkstudio shared a context menu whose structure would be better suited to the needs of the app: [radix-ui](https://www.radix-ui.com/docs/primitives/components/context-menu#anatomy).

## Summary

Refactor the context menu. Now more composable and able to put anything you want in the context menu rather than generating one based off of the `items` prop.

## Relevant Technical Choices

- The context menu is using the [`menu` wai-aria accessibility design](https://www.w3.org/TR/wai-aria-practices-1.2/#menu) 
    - Arrow Keys: focus different options
    - Tab: Closes the menu if possible, but focuses the next/previous element in the tab order
    >Moves focus to the next element in the tab sequence, and if the item that had focus is not in a menubar, closes its menu and all open parent menu containers.
- Keeping the hooks (`useQuickActions`, `RightClickMenuProvider`) returning the `items` arrays rather than the JSX markup that should be in the menu. Hooks shouldn't return JSX, but rather just the business logic. The JSX can be created using the resulting values.

## To-do

- [x] Figure out tabbing problem. Tabbing will not focus the next focusable item in the DOM, but will jump to the searchbar (when tabbing forward. tabbing backward is fine)
- [x] Use query selectors rather than recursively grabbing all children of menu
- [x] Fix broken karma tests

## User-facing changes

Everything should look _exactly the same_ as before

## Testing Instructions

Will Update these soon. There is a lot to test. For those that are eager the locations are:
1. Publisher logo menu (settings page) when adding 2 or more publisher logos
2. Story page (list and grid) have context menus when the elipses are clicked
3. Right click menu in the editor when right clicking an element
4. The quick action menu next to the canvas

## Checklist

---

Fixes #9750


packages/dashboard/src/components/popoverMenu/story-menu-generator.js
```diff
@@ -49,7 +49,7 @@ export const generateStoryMenu = ({
     (item) => !item?.capability || story?.capabilities?.[item?.capability]
   );
 
-  return menuItemsFiltered.map(({ value, ...menuItem }) => {
+  return menuItemsFiltered.map(({ capability, value, ...menuItem }) => {
```
samwhale, line 52: remove this off of the menu items - should not be passed into the context menu. Only necessary for this function.

packages/wp-dashboard/src/effects/useMediaOptimization.js
```diff
@@ -51,7 +51,11 @@ export default function useMediaOptimization() {
     trackEvent('video_optimization_optin', {
       status: mediaOptimization ? 'off' : 'on',
     });
-    showSnackbar({ message: SUCCESS.SETTINGS.UPDATED, dismissible: true });
+    showSnackbar({
+      'aria-label': SUCCESS.SETTINGS.UPDATED,
```
samwhale, line 55: there's a prop type error that was showing up in some snackbars. Same as below.

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
```diff
@@ -418,171 +419,205 @@ const useQuickActions = () => {
...
+              element: 'none',
+            });
+          }}
+          {...actionMenuProps}
+        >
+          <ContextMenuComponents.Icon
+            title={ACTIONS.CHANGE_BACKGROUND_COLOR.text}
+          >
+            <Bucket />
+          </ContextMenuComponents.Icon>
```
samwhale: In my opinion hooks shouldn't return react elements. I think I'll revert this to what it was and used the returned arrays of objects to render the UI we want.

packages/design-system/src/components/contextMenu/menu.js
```diff
@@ -17,379 +17,205 @@
...
-    background-color: transparent;
-    text-decoration: none;
+/**
+ * Finds all focusable children in an html tree.
+ *
+ * @param {Array.<HTMLElement>} all Array of html elements
+ * @param {HTMLElement} element The element to check
+ * @return {Array.<HTMLElement>} Array of html elements
+ */
+export function findFocusableChildren(all, element) {
```
samwhale: 💭 is there a query selector that we could use instead of doing this thing

packages/design-system/src/components/contextMenu/menu.js
```diff
@@ -17,379 +17,205 @@
...
+  });
+  const menuRef = useRef(null);
+  const composedListRef = useComposeRefs(mouseDownOutsideRef, menuRef);
+
+  const handleFocus = useCallback(
+    (evt) => {
+      onFocus(evt);
+
+      const focusableChildren = Array.from(
+        menuRef.current?.children || []
```
samwhale: 💭 Have somewhere defined in the codebase that has all tags that are focusable. Just grab all focusable elements!

packages/design-system/src/components/contextMenu/menu.js
```diff
@@ -17,379 +17,167 @@
...
+    })
+  );
+  const mouseDownOutsideRef = useMouseDownOutsideRef(() => {
+    isOpen && onDismiss?.();
+  });
+  const menuRef = useRef(null);
+  const composedListRef = useComposeRefs(mouseDownOutsideRef, menuRef);
+
+  const getFocusableChildren = (parent) =>
+    Array.from(parent.querySelectorAll(FOCUSABLE_SELECTORS.join(', ')));
```
maxyinger: should we just define this outside the function scope? did the linter not flag the useCallback's deps below for not including it?

packages/wp-dashboard/src/effects/useMediaOptimization.js
maxyinger: Awesome. I've been seeing this in the console. thanks for cleaning up

maxyinger (approved):
Looks great! I think the code is much more easy to reason about and should provide us much more flexibility in the future for nested navigation or any other feature requests for the menus here. Great work!

Just checked it out on my local. I think the only thing that's missing rn is when you're inside a popup and you tab out of it. You already mentioned this issue earlier, but just noting here

samwhale:
>I think the only thing that's missing rn is when you're inside a popup and you tab out of it

~Turns out that it's working exactly as intended! Surprisingly the next item in the tab order in the story grid _is_ the url bar because pressing `Tab` will move focus out of the story grid (since the story grid still has focus).~

~However, as a user I'd expect that when I press `Tab` that the menu closes and focus goes back to the story that I'm focused on. I think we could fix that, but it should be in a separate ticket.~

~Thoughts @littlemilkstudio?~

Note: Ignore! See below

samwhale:
> > I think the only thing that's missing rn is when you're inside a popup and you tab out of it
> 
> Turns out that it's working exactly as intended! Surprisingly the next item in the tab order in the story grid _is_ the url bar because pressing `Tab` will move focus out of the story grid (since the story grid still has focus)
> 
> However, as a user I'd expect that when I press `Tab` that the menu closes and focus goes back to the story that I'm focused on. I think we could fix that, but it should be in a separate ticket.
> 
> Thoughts @littlemilkstudio ?

Actually lmao I fixed it, guess that functionality was written but didn't work until I hooked up `tab` to the keydown effect 😓 

So just ignore! Everything works!

BrittanyIRL (approved):
This refactor makes a lot of sense with the menu usages of this app now vs 2 years ago when it was just for the dashboard. Thanks for taking this on. 

Works great but let's confirm that all the tests are good to go before you merge it.

packages/story-editor/src/components/canvas/pageSideMenu.js
```diff
@@ -63,36 +68,78 @@ function PageSideMenu() {
...
+        isAlwaysVisible
+        isIconMenu
+        disableControlledTabNavigation
+        aria-label={__(
+          'Group of available options for selected element',
+          'web-stories'
+        )}
+        onMouseDown={(e) => {
+          // Stop the event from bubbling if the user clicks in between buttons.
+          // This prevents the selected element in the canvas from losing focus.
```
BrookeGraham, line 89: Oh thanks for the comment!

BrookeGraham (approved):
This is so good. 🎉

PR #9961, opened 2021-12-10
Media: Use smaller video poster image size for base color generation

spacedmonkey:
## Summary

Make a api call to get the poster image and using a smaller image size, so it make the base color generation much faster. 

### Before
Screenshot 2021-12-10 at 00 23 09

### After
Screenshot 2021-12-10 at 00 21 46

## Testing Instructions

1.

## Checklist

---

Fixes #

### DIFF
--- a/packages/wp-story-editor/src/api/media.js
+++ b/packages/wp-story-editor/src/api/media.js
@@ -139,13 +139,13 @@ export async function getOptimizedMediaById(config, mediaId) {
 export async function getPosterMediaById(config, mediaId) {
   const path = addQueryArgs(`${config.api.media}${mediaId}/`, {
     context: 'edit',
-    _fields: 'meta.web_stories_poster_id',
+    _fields: 'featured_media',
   });
 
   const result = await apiFetch({ path });
 
-  if (result?.meta?.web_stories_poster_id) {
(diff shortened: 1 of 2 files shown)


spacedmonkey:
This change should be added too https://github.com/google/web-stories-wp/pull/9953 as well.

packages/wp-story-editor/src/api/media.js
```diff
@@ -129,6 +129,28 @@ export async function getOptimizedMediaById(config, mediaId) {
...
+  const path = addQueryArgs(`${config.api.media}${mediaId}/`, {
+    context: 'edit',
+    _fields: 'meta.web_stories_poster_id',
+  });
+
+  const result = await apiFetch({ path });
+
+  if (result?.meta?.web_stories_poster_id) {
+    return getMediaById(config, result.meta.web_stories_poster_id);
+  }
```
swissspidy: What if the user chose a custom poster image for the video? Shouldn't we use the `featured_media` field to get the poster ID?

spacedmonkey: Good catch! Fixed in https://github.com/google/web-stories-wp/pull/9961/commits/d2781d000e567ce40d5b61a1918b776726c614a4

swissspidy (approved):

Issue #9733, opened 2021-11-17
WordPress 5.9 Fixes


swissspidy:
## Task Description

Let's update our local environments to WP 5.9 Beta once it comes out, so we can do some manual testing ahead of the RC.

swissspidy:
One tiny issue I noticed with the embed block in the Twenty Twenty-Two theme:

The story poster placeholder is too big and thus cut off:
![Screenshot 2021-11-17 at 16 48 31](https://user-images.githubusercontent.com/841956/142233985-445cf41c-5e6d-4437-9aea-c4dd599b8fb5.png)

This is easily fixable with the following CSS though:

```css
.web-stories-list__story-poster-placeholder {
    box-sizing: border-box;
}
```

swissspidy:
`'who' => 'authors'` user queries will be deprecated in 5.9.

We currently use those for the `getAuthors` utils:

https://github.com/google/web-stories-wp/blob/401fc75cb94fceb9e54e6ad1b1b945d23b46d76d/packages/wp-story-editor/src/api/authors.js#L26-L34

https://github.com/google/web-stories-wp/blob/401fc75cb94fceb9e54e6ad1b1b945d23b46d76d/packages/wp-dashboard/src/api/authors.js#L26-L34

Possible solution:

1. Pass capability query arg there
2. In `Stories_Users_Controller::get_items()`, wrap around the `parent::get_items()` call and filter `rest_user_query` while doing so. If both `$prepared_args['who']` and `$prepared_args['capabilities']` are set, it means we're on 5.9 and can unset the former.

spacedmonkey:
We will also need to update the reference found herer. 

https://github.com/google/web-stories-wp/blob/8d2f327fd3473d82d3902a4758dc0e6646415283/includes/templates/admin/edit-story.php#L71-L76

As much as I don't like version checks, I thinkwe should do a PHP check in the `Stories_Users_Controller` class, that converts `who` into`capabilities` depending on the version of WP. This way, we change it in one place and works everywhere in all versions.

spacedmonkey:
By default, WordPress 5.9 with a block base theme do not show the customizer. However, if a theme / plugin hooks into `customize_register` it shows the customizer.  [See](https://github.com/WordPress/wordpress-develop/blob/8bd1509ea8d1120b72256929c57bab6701d7ee58/src/wp-admin/menu.php#L217-L224). 

We could make the Customize class conditional or make the hook into `customize_register` conditional.

Issue #9622, opened 2021-11-02
WordPress 5.9 Compatibility


swissspidy:
## Task Description

WordPress 5.9 development timeline: https://make.wordpress.org/core/5-9/

5.9 RC 1 is slated for 30 November 2021

5.9 stable is slated for 14 December 2021.

Before release of WordPress 5.9:

* [x] Do some testing against 5.9 Beta releases (#9733)
* [x] Run E2E tests against 5.9 RC once published  
  * [ ] Add test for #9971
* [x] Update staging environments to use 5.9 RC
* [x] Bump `Tested up to` in `readme.txt`
* [ ] Add customizer support for Twenty Twenty-Two

After release of WordPress 5.9:

* [ ] Update `php-stubs/wordpress-stubs` (perhaps submit a PR to update stubs)

Notable things to test:

* Twenty Twenty-Two
* Web Stories block in Full Site Editing

spacedmonkey:
What about compatiblity with the new default theme 2022? 

https://github.com/google/web-stories-wp/blob/03a51f7df190a40c217d02ffec8b5d2863262ef7/includes/Integrations/Core_Themes_Support.php#L45-L57

swissspidy:
Good point 👍

swissspidy:
Proposed schedule shared on Slack:

Nov 30 - Beta 1 (2 weeks from now)
Dec 7 - Beta 2
Dec 14 - Beta 3
Dec 21 - Beta 4 (optional)
Jan 4 - RC 1 (5 weeks from Beta)
Jan 11 - RC2
Jan 18 - RC 3
Jan 25 - Release date

So we might have some more time after all

spacedmonkey:
Should we run E2E, unit tests again WP 5.9 RC1?

swissspidy:
Yes we should run e2e tests against RC1, as also noted in the ticket description. Like we did in #8136

PR #9971, opened 2021-12-13
Embeds: Disable Customizer integration for block themes

spacedmonkey:
## Summary

Make Customizer class conditional. This class is not needed in block based themes. 

## Testing Instructions

1.

## Checklist

---

See #9622 
See #9733

### DIFF
--- a/includes/Admin/Customizer.php
+++ b/includes/Admin/Customizer.php
@@ -26,6 +26,7 @@
 
 namespace Google\Web_Stories\Admin;
 
+use Google\Web_Stories\Infrastructure\Conditional;
 use Google\Web_Stories\Settings;
 use Google\Web_Stories\Stories_Script_Data;
 use Google\Web_Stories\Story_Post_Type;
@@ -42,7 +43,7 @@ use WP_Error;
  *
  * @package Google\Web_Stories
  */
-class Customizer extends Service_Base {
+class Customizer extends Service_Base implements Conditional {
 
 	/**
 	 * Customizer section slug.
@@ -122,6 +123,17 @@ class Customizer extends Service_Base {
 	 */
 	public function register() {
 		add_action( 'customize_register', [ $this, 'register_customizer_settings' ] );
+	}
+
+	/**
+	 * Check whether the conditional object is currently needed.
+	 *
+	 * @since 1.16.0
+	 *
+	 * @return bool Whether the conditional object is needed.
+	 */
+	public static function is_needed(): bool {
+		return ! function_exists( 'wp_is_block_theme' ) || ! wp_is_block_theme();
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


includes/Admin/Customizer.php
```diff
@@ -124,6 +125,17 @@ public function register() {
...
 
+	/**
+	 * Check whether the conditional object is currently needed.
+	 *
+	 * @since 1.16.0
+	 *
+	 * @return bool Whether the conditional object is needed.
+	 */
+	public static function is_needed(): bool {
+		return ! function_exists( 'wp_is_block_theme' ) || ! wp_is_block_theme();
```
swissspidy, line 136: FYI, this lint error reported by PHPStan is a false positive that should get resolved by #9973

includes/Admin/Customizer.php
```diff
@@ -124,6 +125,21 @@ public function register() {
...
+	 * @since 1.16.0
+	 *
+	 * @return bool Whether the conditional object is needed.
+	 */
+	public static function is_needed(): bool {
+		if ( function_exists( '\wp_is_block_theme' ) ) {
+			return ! wp_is_block_theme();
+		}
+
+		return true;
```
swissspidy: What was wrong with `return ! function_exists( 'wp_is_block_theme' ) || ! wp_is_block_theme();`?

includes/Admin/Customizer.php
```diff
@@ -124,6 +125,21 @@ public function register() {
...
 	}
 
+	/**
+	 * Check whether the conditional object is currently needed.
+	 *
+	 * @since 1.16.0
+	 *
+	 * @return bool Whether the conditional object is needed.
+	 */
+	public static function is_needed(): bool {
```
swissspidy, line 135: Worth adding a test?

includes/Admin/Customizer.php
spacedmonkey: Readability.

swissspidy: I found it quite readable 🤷

includes/Admin/Customizer.php
spacedmonkey: What would it test? That is returns false?

swissspidy: I was thinking of a small unit test, but also an e2e test that verifies that the customizer is properly hidden.

includes/Admin/Customizer.php
spacedmonkey: Changed back.

includes/Admin/Customizer.php
spacedmonkey: How would I do that? I think that should be added once we test against 5.9? 
This change is simple and clear, no reason to block it.

spacedmonkey: Create a follow up ticket to add a e2e test?

swissspidy: > How would I do that?

Adding an e2e test would be pretty simple, no?

But yes, this needs to be done once we test against RC.

> This change is simple and clear, no reason to block it.

It's still a change that needs to be tested :-) Hence my original question whether it's worth adding tests.

swissspidy (approved):
Works as expected

PR #9974, opened 2021-12-13
REST API: Cater for changes to author queries in WordPress 5.9

spacedmonkey:
## Summary

Who param on the WP_User_Query was removed as part of WP 5.9. This change, stops a doing it wrong notice. 

Part of fixes for 5.9 https://github.com/google/web-stories-wp/issues/9733

## Testing Instructions

- [x] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1.

## Checklist

- [ ] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [ ] I have tested this code to the best of my abilities
- [ ] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [ ] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [ ] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [ ] I have added documentation where necessary
- [ ] I have added a matching `Type: XYZ` label to the PR

---

Fixes #

PR #9990, opened 2021-12-14
Use capabilities query if available

swissspidy:
See #9974

### DIFF
--- a/includes/REST_API/Stories_Users_Controller.php
+++ b/includes/REST_API/Stories_Users_Controller.php
@@ -33,7 +33,6 @@ use Google\Web_Stories\Infrastructure\Service;
 use WP_Error;
 use WP_REST_Request;
 use WP_REST_Response;
-use WP_Meta_Query;
 use WP_REST_Users_Controller;
 
 /**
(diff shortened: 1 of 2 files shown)


includes/REST_API/Stories_Users_Controller.php
```diff
@@ -111,34 +123,15 @@ public function get_items( $request ) {
...
-		$blog_id = get_current_blog_id();
-		if ( isset( $prepared_args['blog_id'] ) ) {
-			$blog_id = absint( $prepared_args['blog_id'] );
-		}
+		// Capability queries were added in 5.9, and the 'who' param was deprecated.
+		if ( isset( $prepared_args['who'], $registered['capabilities'] ) && 'authors' === $prepared_args['who'] ) {
+			$capabilities   = $prepared_args['capabilities'] ?? [];
+			$capabilities[] = $this->story_post_type->get_cap_name( 'edit_posts' );
+
+			$prepared_args['capabilities'] = $capabilities;
```
spacedmonkey, line 133: Should this change be made in the core user endpoint

swissspidy: This would prevent devs from seeing the deprecation notice and updating their code to use the new capabilities query.

swissspidy: Also, why the thumbs down? 🤔 

This is more forward compatible than just doing a user level meta query despite that practice technically being deprecated.

spacedmonkey (approved):

PR #9991, opened 2021-12-14
Media: Add filter to disable potentially expensive tax query

### LINKED ISSUE #9853: Media: tax query affects WP media library performance
## Bug Description

> The Web Stories plugin is causing a performance issue with the WordPress Media Library.  The issue is a tax query that's being added in the function `get_exclude_tax_query` in file `plugins/web-stories/includes/Media/Media_Source_Taxonomy.php`
> 
> This tax query is causing the underlying SQL query to be too slow and causing long delays with working in the media library.  This is most noticeable in the preprod environment due to lower database resources, but is slowing down the media library in production as well. 
> 
> We would like to have the plugin add a filter that would allow us to disable the addition of this tax query.  We tested changing the plugin directly to remove this tax query and that appears to resolve the issue.  There are three functions where we had to make code changes for this: `filter_ajax_query_attachments_args`, `filter_generated_media_attachments` and `filter_rest_generated_media_attachments`
> 
> We would need this filter to apply to all of these contexts.  

## Expected Behaviour

Media library performance is good.

## Steps to Reproduce

See above.

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

spacedmonkey:
## Summary

Add a filter to disable attachment filtering. 

## Testing Instructions

1.

## Checklist

---

Fixes #9853

### DIFF
--- a/includes/Media/Media_Source_Taxonomy.php
+++ b/includes/Media/Media_Source_Taxonomy.php
@@ -214,7 +214,7 @@ class Media_Source_Taxonomy extends Taxonomy_Base {
 	 */
 	private function get_exclude_tax_query( array $args ): array {
 		/**
-		 * Filter whether the query should be filtered.
+		 * Filter whether generated attachments should be hidden in the media library.
 		 *
 		 * @since 1.16.0
 		 *
--- a/includes/Media/Media_Source_Taxonomy.php
+++ b/includes/Media/Media_Source_Taxonomy.php
@@ -221,7 +221,7 @@ class Media_Source_Taxonomy extends Taxonomy_Base {
 		 * @param bool  $enabled Whether the taxonomy check should be applied.
 		 * @param array $args    Existing WP_Query args.
 		 */
-		$enabled = apply_filters( 'web_stories_enable_exclude_query', true, $args );
+		$enabled = apply_filters( 'web_stories_hide_auto_generated_attachments', true, $args );
 		if ( true !== $enabled ) {
 			return $args;
 		}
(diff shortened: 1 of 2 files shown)


includes/Media/Media_Source_Taxonomy.php
```diff
@@ -213,6 +213,19 @@ public function update_callback_media_source( $value, $object ) {
...
 	private function get_exclude_tax_query( array $args ): array {
+		/**
+		 * Filter whether the query should be filtered.
+		 *
+		 * @since 1.16.0
+		 *
+		 * @param bool  $enabled Whether the taxonomy check should be applied.
+		 * @param array $args    Existing WP_Query args.
+		 */
+		$enabled = apply_filters( 'web_stories_enable_exclude_query', true, $args );
```
swissspidy: WDYT about a filter name like `web_stories_hide_auto_generated_attachments` that explains a bit more what this actually does?

includes/Media/Media_Source_Taxonomy.php
```diff
@@ -213,6 +213,19 @@ public function update_callback_media_source( $value, $object ) {
 	 * @return array  Tax query arg.
 	 */
 	private function get_exclude_tax_query( array $args ): array {
+		/**
+		 * Filter whether the query should be filtered.
```
swissspidy:
```suggestion
		 * Filter whether generated attachments should be hidden in the media library.
```

swissspidy (approved):
Just a question about naming

includes/Media/Media_Source_Taxonomy.php
spacedmonkey: I like it. https://github.com/google/web-stories-wp/pull/9991/commits/1379c516123f70eb5b83dee7316f8f84a27de3c9

swissspidy:
Haven't heard back yet whether this filter suffices, but seems reasonable to add nonetheless.

PR #9999, opened 2021-12-14
Media: Only transcode 1 resource at a time

### LINKED ISSUE #8779: Video Optimization: reduce number of simultaneous processes
## Bug Description

If 5+ videos are dragged into the browser and 5+ videos are transcoded at the same time, the browser will crash. 

Found while testing https://github.com/google/web-stories-wp/pull/8770, which did not fix the issue. 

## Expected Behaviour

Videos are queued and browser does not crash. 

## Screenshots

https://user-images.githubusercontent.com/237508/130645611-d8aa7dca-bc94-4afe-aff3-5180635cbaf8.mov

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

swissspidy:
## Context

Too many ongoing ffmpeg process can cause the browser to crash due to resource usage being too high.

## Summary

This PR changes the media upload logic so that only 1 video is ever transcoded at once.

The ffmpeg poster generation (which is a fallback if the browser can't display a poster natively) is also limited to 1 process at a time.

So if multiple videos are uploaded, it might take a tad longer due to this limitation, but at least nothing is crashing.

## Relevant Technical Choices

Updated tests to make them a bit more robust hopefully.

## User-facing changes

No visual changes.

## Testing Instructions

Video upload:

1. Create a new story.
1. Drag & drop multiple (e.g. 5) MOV files onto the canvas at once
2. Observe the videos uploading
3. The browser should not crash

Video optimization:

1. Go to the media library in WordPress
2. Upload a few MP4 videos
1. Create a new story with a few pages
2. Add all these new MP4 videos to the story
3. Open the pre-publish checklist and open the "Accessibility" tab
4. There should be a note about optimizing videos
5. Optimize all videos
6. The browser should not crash

No

No

No

## Checklist

---

Fixes #8779


spacedmonkey:
Might worth doing a merge from main. This PR is 27 days old.

spacedmonkey:
One thing to note, the pending videos, do not have a messaging on them on hover. 

Hoving over trancoding video
Screenshot 2022-01-10 at 15 19 26

Hoving over pending video, shows no message at all. 
Screenshot 2022-01-10 at 15 19 30


Not sure if it should say, pending processing or just processing through.

swissspidy:
> Might worth doing a merge from main. This PR is 27 days old.

I did that 10 minutes before your comment.

spacedmonkey:
https://user-images.githubusercontent.com/237508/148792273-55e5ebef-89f0-4dea-aceb-f9a9281d7884.mov

I am seeing a browser crash here. 

I believe what is happening here, is that the first video is processed, then when it finishes, all the other video all start transcoding at once.

swissspidy:
Sounds like I need to come up with some good unit tests to cover this 🤔

maxyinger:
is there any worth in extracting the simple async task queue from this PR and using it here?
https://github.com/google/web-stories-wp/pull/9584

There's already some unit test coverage for it running in sequence correctly. at the very least, can probably use the unit tests as prior art for testing this behavior

swissspidy:
I've been considering re-using that one for some other tasks where we'd needed to perform 1 thing at a time (hence my comment on your PR last month), but I'm not sure it's a good fit _here_.

This one here is part of a larger media upload queue, where all steps should run in parallel except this one task should be sequentially. Adopting your work would basically mean creating a separate queue just for this, which might make things unnecessarily complex. But I'll take another look for sure.

swissspidy:
@spacedmonkey friendly ping

packages/story-editor/src/app/media/utils/useMediaUploadQueue/test/useMediaUploadQueue.js
```diff
@@ -23,19 +23,21 @@ import { isAnimatedGif } from '@web-stories-wp/media';
 /**
  * Internal dependencies
  */
+import createResource from '@web-stories-wp/media/src/createResource';
```
spacedmonkey: Why do it this way? 
Why not
```
import { isAnimatedGif, createResource } from '@web-stories-wp/media';
```

swissspidy: Totally. Not intentional.

spacedmonkey:
I am still seeing the issue locally. Anything more than 7 videos being uploaded at once, will still see the browser crash. 

https://user-images.githubusercontent.com/237508/149515332-14addd37-885a-4e0d-bf55-ce033f26ece5.mov

spacedmonkey:
The following lines 
https://github.com/google/web-stories-wp/blob/7440358cb9cd4115550a5b2f51319fdfba747e6a/packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js#L145-L192

Should also be limited to one at a time. This code generates the poster image using ffmpeg for files that need to be transcoded. This spins up ffmpeg to get a poster image. But this would be done for all uploading video. Which maybe why the above issue is happening. Sadly, if don't generate poster image, the preview will not look great, but I don't see a way around this.

spacedmonkey:
I have confirmed, if you drag 7 mp4 into the canvas, browser does not crash. 7 mov files and it does. It must be poster generation via FFMPeg is the issue.

swissspidy:
Interesting! Totally forgot about the ffmpeg poster generation.

Should be easy to address that. Will take a look.

spacedmonkey:
Also not seeing any changes related to this comment https://github.com/google/web-stories-wp/pull/9999#issuecomment-1008985240

Is this going to actioned as part of this PR or not?

swissspidy:
> Hoving over pending video, shows no message at all.

Pending items are just pending and we don't know whether they are going to need transcoding or not.
We've never shown a tooltip for pending items, so this is more a feature request. But I suppose it makes sense.

I'm going to generalize the tooltip text a bit for it to make more sense.

spacedmonkey:
There different now, is that pending items used to just be pending for half a second while the join the queue. Now a pending item could be queued for a while, minutes while other video transcode in front of it. If a video has a progress bar for 5 minute with the tooltip, might be confusing. Maybe the message should be like this. 

- Pending - "Video is await processing".
- Trasncoding - "Video is being processed"
- Uploading - "Video is uploading". 

That way it clear to the user

swissspidy:
Users don't need to know about how media upload is working under the hood, i.e. whether it's currently pending, uploading, transcoding, generating a poster, finishing uploading, etc. This generalized message hides this complexity from them on purpose.

Can you confirm that there are no crashes anymore when uploading multiple MOV files?

spacedmonkey:
https://user-images.githubusercontent.com/237508/149549539-c40fb997-1ee5-4108-aeb5-1d14d99e70af.mov

I am not longer seeing the crashing, which is good. However, I am seeing some poster images not generating at all. There are some weird gaps.

swissspidy:
Not at all or just one after another? To me it looks like the latter.

spacedmonkey:
After a couple of tries. There is no order I can see. It is not the first / last element. It can be any of them.

swissspidy:
I don‘t think an order is (or should be) guaranteed for the ffmpeg poster generation.

spacedmonkey:
@swissspidy I am not sure I understand your last two comments. The video I uploaded, shows a regression in functionality. This change messages that some poster images not loading at all. Can you take a look?

swissspidy:
So I found one sorta related bug where the temporary poster image generated by ffmpeg was later dropped after uploading. I fixed this in 55f5a8b5bbee6384300559c31b8bb71aedfecffc. But worth noting that this was also a bug on `main` and not introduced here. Maybe it was just made more obvious.

Apart from that, in my testing the ffmpeg poster generation works fine with 1 poster being generated at a time.

I'll do some more testing though.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -177,6 +191,8 @@ function useMediaUploadQueue() {
             });
           } catch {
             // Not interested in errors here.
+          } finally {
+            currentPosterGenerationItem.current = null;
```
spacedmonkey, line 195: Do we also need to call this 
```
currentPosterGenerationItem.current = null;
```

Above? 

```
if (!isMounted.current) {
       return;
}
```

spacedmonkey:
In my testing, it seems like it only one or two of files that are failing. [Example](https://user-images.githubusercontent.com/237508/149820735-77269b91-20bd-4d08-9e1b-97b7f0fbf4d2.MOV). 

Otherwise this is working well.

swissspidy:
That's actually the file I was testing with.

spacedmonkey (approved):

spacedmonkey:
Approved!

This needs QA instructions, as this testing here will be valuable.

PR #10000, opened 2021-12-14
Karma: Fixed tests using background images

### LINKED ISSUE #9910: Flaky Karma tests: Background Copy Paste integration
## Bug Description
Various Karma tests for `Background Copy Paste integration` seem to fail locally on `main` and also in the CI

`should correctly copy pattern background to page with image`
`should correctly copy image to page without image`
`should correctly copy image to page with existing image`
`works for all background animations`

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

barklund:
## Context

This fixes a number of tests using background images (dragging over the canvas edge apparently doesn't work anymore inside karma, but we have the "set as background" button back again, which is much simpler anyway).

## User-facing changes

_None_

## Testing Instructions

## Checklist

---

Fixes #9910

### DIFF
--- a/packages/story-editor/src/components/canvas/karma/backgroundCopyPaste.karma.js
+++ b/packages/story-editor/src/components/canvas/karma/backgroundCopyPaste.karma.js
@@ -48,15 +48,10 @@ describe('Background Copy Paste integration', () => {
 
   it('should correctly copy pattern background to page with pattern background', async () => {
     // Arrange the backgrounds
-    //await karmaPause();
     await gotoPage(1);
-    //await karmaPause();
     await setBackgroundColor('FF0000');
-    //await karmaPause();
     await gotoPage(2);
-    //await karmaPause();
     await setBackgroundColor('00FF00');
-    //await karmaPause();
 
     // Verify setup - 1 element on each page in the right color
     await gotoPage(1);
@@ -105,7 +100,6 @@ describe('Background Copy Paste integration', () => {
     expect(await getNumElements()).toBe(1);
     await gotoPage(2);
     expect(await getCanvasBackgroundElement()).not.toBeEmpty();
-
     expect(await getCanvasBackgroundImage()).toHaveProperty(
       'src',
       /blue-marble/
--- a/packages/story-editor/src/components/canvas/karma/backgroundCopyPaste.karma.js
+++ b/packages/story-editor/src/components/canvas/karma/backgroundCopyPaste.karma.js
@@ -15,6 +15,11 @@
  */
 
 /**
+ * External dependencies
+ */
+import { waitFor } from '@testing-library/react';
+
+/**
  * Internal dependencies
  */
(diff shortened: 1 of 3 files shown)


packages/story-editor/src/components/canvas/karma/backgroundCopyPaste.karma.js
```diff
@@ -234,13 +239,7 @@ describe('Background Copy Paste integration', () => {
...
   }
   async function gotoPage(index /* 1-indexed */) {
-    const carousel = getElementByQueryAndMatcher(
-      'div[role="listbox"]',
-      getByAttribute('aria-label', /Pages List/i)
-    );
-    const pageAtIndex = carousel.querySelectorAll('button[role="option"]')[
-      index - 1
-    ];
+    const pageAtIndex = fixture.editor.footer.carousel.pages[index - 1].node;
```
barklund, line 242: This wasn't actually required, but I thought the error had to do with this at first, so updated it anyway.

packages/story-editor/src/components/canvas/karma/backgroundCopyPaste.karma.js
```diff
@@ -29,6 +34,12 @@ describe('Background Copy Paste integration', () => {
     await fixture.collapseHelpCenter();
 
     await addNewPage();
+
+    await waitFor(() => {
+      if (fixture.editor.footer.carousel.pages.length === 0) {
+        throw new Error('Carousel pages not loaded yet');
+      }
+    });
```
barklund, line 42: Likewise here.

swissspidy:
@barklund Congrats on reaching the **#10000** milestone! 🎉

miina (approved):
🏅 🎉 🥳 #10000 

Any idea why dragging the background doesn't work anymore here?

barklund:
> Any idea why dragging the background doesn't work anymore here?

I was wondering a lot about that. I can manually do it even in the Phantom browser, but regardless of how I tried to drag media over the trigger point using script, it would not trigger the drop target. I tried waiting, dragging around in multiple steps, etc, but no luck. No replacement would trigger on hover and no drop target would trigger on release.

I wasn't sure if that's a bad thing or not. I don't think it reflects that the feature is broken, just that for some reason, it doesn't work using script (either because of some dependency update or because of a change in our app behavior). I didn't spend too much time on it, as the whole "drag to place bg image" was only a workaround introduced back when we removed the "set as background" button, but now that we have it back, we can revert to using that.

miina:
> I didn't spend too much time on it, as the whole "drag to place bg image" was only a workaround introduced back when we removed the "set as background" button, but now that we have it back, we can revert to using that.

Yep, it's not really directly related to the tests that this PR fixes. 

(I didn't merge the PR just to leave you the honour of the 10000th, not because of the dragging question 😛)

PR #10001, opened 2021-12-14
Media: refactor upload logic to reduce complexity

swissspidy:
## Context

The logic in the media upload queue was getting harder and harder to read as we basically shoved everything into one big function.

## Summary

This PR splits up the logic in more digestible smaller functions, reducing complexity

## Testing Instructions

1.

## Checklist

---

Fixes #


packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -325,186 +477,56 @@ function useMediaUploadQueue() {
             currentTranscodingItem.current = id;
 
             if (isGifThatNeedsTranscoding) {
```
spacedmonkey, line 487: I don't like the multiple elses here. Makes it a little hard to read. How about. 

```
 if (isTranscodingEnabled) {
            currentTranscodingItem.current = id;

            if (isGifThatNeedsTranscoding) {
              convertGifItem(item);
              return;
            } else if (canTranscodeFile(file)) {
              if (trimData) {
                trimVideoItem(item);
                return;
              } else if (muteVideo) {
                muteVideoItem(item);
                return;
              } else {
                // Transcode/Optimize videos before upload.
                // TODO: Only transcode & optimize video if needed (criteria TBD).
                // Probably need to use FFmpeg first to get more information (dimensions, fps, etc.)
                optimizeVideoItem(item);
                return;
              }
            } 
          
            uploadItem(item);
        }
```

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -207,52 +203,216 @@ function useMediaUploadQueue() {
...
+      } catch (error) {
+        // Cancel uploading if there were any errors.
+        cancelUploading({ id, error });
+
+        trackError('upload_media', error?.message);
+      } finally {
+        trackTiming();
+      }
+
       try {
```
spacedmonkey, line 344: I would see poster upload to move to it's own effect. It could even have it's own state. So
```
UPLOADING -> UPLOADED -> FINISHED
```
Becomes 
```
UPLOADING -> UPLOADED -> UPLOADING-POSTER-> FINISHED
```

Thoughts.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -207,52 +203,216 @@ function useMediaUploadQueue() {
...
         const { poster, posterId } = await uploadVideoPoster(
```
spacedmonkey, line 346: Before calling this, you should check if `posterFile` is not null. Like in the case of image uploads. 

This makes it exit quicker. See 

https://github.com/google/web-stories-wp/blob/ecdc7c487b3a75b86ddb022700659818713786dd/packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js#L467

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: That's just multiple else with returns in them? 🤔 

I suppose you mean just returning early and avoiding the elses?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: Hmm not sure it warrants its own separate status 🤔  especially since it only applies to videos.

What I like about the current way is we only consider a video to be fully uploaded once we have at least attempted to upload a poster.

That said, `uploadItem` definitely could need some tweaking, because right now it looks like it tries to upload a poster even for images 🤔

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: As per my comment above, I think I'm going to refactor the whole `uploadItem` function anyway to separate image and video upload logic here. It's quite confusing.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: Fixed

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: Reworked that no a little bit; resolving.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: Reworked that no a little bit; resolving.

packages/story-editor/src/app/media/utils/getResourceFromLocalFile.js
```diff
@@ -81,7 +88,11 @@ const getVideoResource = async (file) => {
 
   await seekVideo(videoEl);
   const hasAudio = hasVideoGotAudio(videoEl);
-  const posterFile = await getImageFromVideo(videoEl);
+  const posterFile = blobToFile(
+    await getImageFromVideo(videoEl),
+    getPosterName(getFileName(file)),
+    MEDIA_POSTER_IMAGE_MIME_TYPE
+  );
```
spacedmonkey, line 95: Why is this change needed?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -207,52 +203,226 @@ function useMediaUploadQueue() {
...
+      const { id, file, resource, additionalData = {} } = item;
+      let { posterFile } = item;
+
+      let newResource;
+
+      // The newly uploaded file won't have a poster yet.
+      // However, we'll likely still have one on file.
+      // Add it back so we're never without one.
+      // The final poster will be uploaded later by uploadVideoPoster().
+      newResource = await uploadFile(file, additionalData);
```
spacedmonkey: Can't we just do this?
```suggestion
      let newResource = await uploadFile(file, additionalData);
```

packages/story-editor/src/app/media/utils/getResourceFromLocalFile.js
swissspidy: It simplified the `posterFile` handling in the media upload queue which manually set a poster file name here:

https://github.com/google/web-stories-wp/blob/417c819d3c12adaf5e33a6c29f59c2050fabeaeb/packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js#L280

This way now we know that `posterFile` always has a proper name that we can use.

spacedmonkey:
https://user-images.githubusercontent.com/237508/150521275-dd2e9b09-eb9f-484a-a36f-2dc921525586.mov

I am seeing the transcoding hanging forever. As you can see from this video, the transcode very kicks in.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: We can. Just an oversight from refactoring.

swissspidy:
Does it also happen when only adding 1 video? Does it consistently happen?

spacedmonkey:
> I am seeing the transcoding hanging forever. As you can see from this video, the transcode very kicks in.

See this on main :(

spacedmonkey:
I am also trancoding not triggering with one video as well.

swissspidy:
As per my message on Slack, if this also happens on `main` it's important to verify that there's actually a bug that can be reproduced by others as well. Personally I cannot reproduce the issue and there are also no user reports about this issue (which we would certainly get since it would be quite a major bug)

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
spacedmonkey:
```suggestion

      // The newly uploaded file won't have a poster yet.
      // However, we'll likely still have one on file.
      // Add it back so we're never without one.
      // The final poster will be uploaded later by uploadVideoPoster().
      let newResource = await uploadFile(file, additionalData);
```

spacedmonkey (approved):

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: I already made this change locally and was about to push it...

With your change you caused a lint error.

PR #10005, opened 2021-12-15
Add filter to allow custom attributes on Web Stories rendering

freibergerb:
## Context

We want to be able to add custom attributes to the `a` tag when Web Stories Widgets are added to a page.
Our use case is that we also have another Analytics platform that isn't Google Analytics and this is the way we thought to integrate.

## Summary

We are adding a filter called `story_render_attributes` passing 3 different parameters to it, `$this->current` which is an instance of the `Story` object, its `position` and the `get_view_type` which is driven by the widget selected.

## Relevant Technical Choices

We thought that a filter would be the most appropriated way to achieve that.

## To-do

None I can think of at this stage but let's see what the contributors think and if they have any other ideas. 

## User-facing changes

None, this is transparent for users.

## Testing Instructions

We do have this patch current implemented live with the filter been added on our end (outside this plugin). It's not on our best interests though to have a modified version of this plugin in our codebase as it will be a problem to update as you release new functionalities. I hope this can be useful for other use cases as well.

1.

## Checklist

---

Fixes #


swissspidy:
Thanks for for your contribution!

Please note that you'll need to [sign the Contributor License Agreement](https://cla.developers.google.com/about) before we could accept it.

I'm curious to learn more about this analytics solution you are using. What kind of attributes do you need to set?

For the stories embeds we  actually prevent clicks on the `<a>` elements so we can open stories in a lightbox. Doesn't that interfere with your analytics? If so, it might be more preferable for you to add a JavaScript integration instead so you can better track things like the lightbox opening, story play/pause/mute, etc.

As for the filter, are all the arguments necessary for your integration to work? Or do you perhaps only need the story object as context?

freibergerb:
hey @swissspidy,

No worries I just signed.

Thank you for looking into this. We have Adobe Analytics as our main Analytics solution. The `href` would look like below:
`
<a href="https://www.theaustralian.com.au/web-stories/free/the-australian/taliban-gunmen-steal-australian-food-aid-in-afghanistan" data-tgev="event290" data-tgev-metric="ev" data-tgev-order="1" data-tgev-container="web-stories-carousel" data-tgev-label="Taliban gunmen steal Australian food aid in Afghanistan">
`

These are the exact attributes the Analytics team has passed to us and asked to be implemented. The reason why we have 3 parameters is because we do use the order + widget name/id as `data-tgev-order` and `data-tgev-container` respectively.

If you would like to see this live you can go on https://www.theaustralian.com.au/ and scroll down or search for "Visual Stories".

If I'm honest I'm unaware how Adobe Analytics work in detail and if this could be handled any other way. I'm happy to raise with the Analytics team though.

What are your main concerns around this addition? A potentially very specific use case?

Thanks!

swissspidy:
Awesome, thanks a lot for the context!

The use case does sound reasonable for sure, no concerns there 🙂 

I assume you have heard from the Analytics team already that this works like intended?

----


I am asking because there seem to be some styling issues with the Web Stories embeds currently that prevent the clicks from being tracked properly. Something that we'll need to fix in the plugin asap. Right now the `<a>` tags have height of 0.

If you run `document.querySelectorAll('.web-stories-list__story-poster a').forEach(link => link.addEventListener('click', console.log))` in dev tools and click on some stories, you'll see that nothing is getting logged right now because of this.

includes/Renderer/Stories/Renderer.php
```diff
@@ -624,6 +624,21 @@ protected function render_story_with_poster() {
 		}
 	}
 
+	/**
+	 * Allows customization of html attributes in the web stories widget anchor tag loop
+	 * Converts array into escaped inline html attributes
+	 *
+	 * @return void
+	 */
```
swissspidy:
```suggestion
	/**
	 * Render additional link attributes.
	 *
	 * Allows customization of html attributes in the web stories widget anchor tag loop
	 * Converts array into escaped inline html attributes.
	 *
	 * @since 1.16.0
	 *
	 * @return void
	 */
```

includes/Renderer/Stories/Renderer.php
```diff
@@ -624,6 +624,21 @@ protected function render_story_with_poster() {
...
+	protected function render_attributes() {
```
swissspidy: Maybe rename to `render_link_attributes()` to be more specific?

includes/Renderer/Stories/Renderer.php
```diff
@@ -624,6 +624,21 @@ protected function render_story_with_poster() {
...
+		$story_render_attributes = apply_filters( 'story_render_attributes', [], $this->current(), $this->position, $this->get_view_type() );
```
swissspidy: Let's rename the filter and add some inline docs:

```suggestion
		/**
		  * Filters the link attributes added to a story's <a> tag.
		  *
		  * @since 1.16.0
		  *
		  * @param array  $attributes Key value array of attribute name to attribute value.
		  * @param Story  $story      The current story instance.
		  * @param int    $position   The current story's position within the list.
		  * @param string $view_type  The current view type.
		$story_render_attributes = apply_filters( 'web_stories_renderer_link_attributes', [], $this->current(), $this->position, $this->get_view_type() );
```

includes/Renderer/Stories/Renderer.php
```diff
@@ -624,6 +624,21 @@ protected function render_story_with_poster() {
...
+		if ( ! empty( $story_render_attributes ) ) {
+			foreach ( $story_render_attributes as $attribute => $value ) {
+				echo ' ' . esc_attr( $attribute ) . '="' . esc_attr( $value ) . '"';
```
swissspidy: `esc_attr()` should only be used for attribute values, not attribute names.

I think we could use `wp_kses_one_attr()` to escape the output here:

```suggestion
				echo ' ' . wp_kses_one_attr( $attribute . '="' . esc_attr( $value ) . '"' );
```

includes/Renderer/Stories/Renderer.php
freibergerb: Yes it does, we have it currently live and tracking as expected.

I see, it makes sense. We are keeping a close eye on the releases so I will check when this gets sorted to see if somehow helps our tracking as well.

Thank you for clarifying!

includes/Renderer/Stories/Renderer.php
freibergerb: Thank you, I actually didn't know about this method, very useful!

includes/Renderer/Stories/Renderer.php
freibergerb: I will address this shortly!

includes/Renderer/Stories/Renderer.php
freibergerb: Sounds good!

includes/Renderer/Stories/Renderer.php
freibergerb: @swissspidy I was just checking this, it seems like we need a second param as the element we are adding it for.
`echo ' ' . wp_kses_one_attr( $attribute . '="' . esc_attr( $value ) . '"', 'a' );`

I was just testing to see if it would render it correctly and it does! 
I'm getting _phpcs_ complaints on my end, I think that's why I ended up going for `esc_attr` on the attribute name.

Are you happy to add as it's above? I will wait on your response before I commit.

freibergerb: fyi I went ahead and committed the changes!

swissspidy (approved):

PR #10008, opened 2021-12-15
Templates: Keyboard Nav Karma tests

### LINKED ISSUE #9912: Flaky Karma test: `should navigate templates via keyboard`
## Bug Description
The following Karma test is flaky:
`CUJ: Page Templates: Creator can Apply a Default Page Template Keyboard Navigation should navigate templates via keyboard`

[Example run](https://github.com/google/web-stories-wp/runs/4433714640?check_suite_focus=true#step:6:8235)

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

BrookeGraham:
## Context

Keyboard navigation karma test for page templates has been flakey, and failing unexpectedly. 

## Summary

To be honest I couldn't get this to fail regularly. When it did fail, it was because all the templates were still present when we were expecting them to have been filtered already.  

## Relevant Technical Choices

Added a  sleep to give it some time before for moving on and evaluating how many times a 'cover' template was still present after filtering.

## Testing Instructions

1.

no

no

no

## Checklist

---

Fixes #9912


swissspidy:
FYI just converted to a draft again since it seems to be in WIP, just so we don't unnecessarily create Percy snapshots and the like :-)

packages/story-editor/src/components/library/panes/pageTemplates/karma/defaultPageTemplates.karma.js
```diff
@@ -215,6 +213,10 @@ describe('CUJ: Page Templates: Creator can Apply a Default Page Template', () =>
       await fixture.events.keyboard.press('right');
       await fixture.events.keyboard.press('right');
       await fixture.events.keyboard.press('Enter');
+
+      // allow filtered templates to be removed from DOM
+      await fixture.events.sleep(500);
```
BrookeGraham: To be honest I couldn't get this to fail regularly. When it did fail, it was because all the templates were still present when the below expect would run. Adding the sleep to give it some time for moving on seemed to help as I couldn't get it to fail once I did so.

maxyinger: I wonder if just increasing the `timeout` option param on the `waitFor` below would have a similar effect on making this robust without always making this test take an extra 500ms. Also tho I feel like `waitFor` doesn't always behave as expected so understand if this needs to be a `sleep` call

maxyinger (approved):
I wonder if there's another approach we could take to make this test more stable 🤔 . I'm fine with the `sleep` call if we can't find anything else, just not the biggest fan of sprinkling in `sleep` calls even tho sometimes it feels like the only approach

I can't think of anything other than increasing the timeout of the waitFor, but even tho in theory that should make this just as robust, I feel like it doesn't always behave as expected 🤷 

Regardless, thanks for the fix 🙌

packages/story-editor/src/components/library/panes/pageTemplates/karma/defaultPageTemplates.karma.js
BrookeGraham: Totally agree. The `sleep` feels cheap. I'll try a few more test rounds with the `waitFor` cause I do like that approach better.

samwhale (approved):
Noice. I think the `waitFor`s are going to be a little more sturdy than the `sleep`s 😄

BrittanyIRL (approved):
looks good but let's wait for a solution to the karma fixture timeout before merging so we can make sure it's actually fixing remote.

BrookeGraham:
> looks good but let's wait for a solution to the karma fixture timeout before merging so we can make sure it's actually fixing remote.

For sure.

PR #10019, opened 2021-12-16
Revert "Media: Generate and backfill BlurHash for media items"

swissspidy:
Reverts #9953

Karma tests began failing starting with that PR and were also super slow due to tons of workers being spun up during tests.

While adding missing baseColor to fixture data helped a bit (see #10016), it did not solve the random browser disconnection errors.

More time is needed to properly investigate this.

A future version of said PR needs to:

* Add missing baseColor to all fixture data
* Not slow down tests
* Ideally improve worker usage so that it doesn't spin up dozens of threads.
  It might be more performant to have only 1 worker for all the blurhash generation/backfilling and pass messages back and forth, instead of creating 1 per image 

https://www.npmjs.com/package/@shopify/react-web-worker might help with the latter, not sure.

**In any case, we will revisit this in the new year.**


miina:
Looks like the Karma tests are still not completing here:
First run with the disconnected error: https://github.com/google/web-stories-wp/runs/4551529472?check_suite_focus=true#step:6:1453
Second run with a different error: https://github.com/google/web-stories-wp/runs/4552157294?check_suite_focus=true#step:6:1605

Really odd since the BlurHash PR was the first commit where these issues started happening 😕 Would have guessed reverting it would fix things.

spacedmonkey:
Wouldn't it just be easier to just update the a blurhash for all the fixture media elements?

swissspidy:
See #10016 for that, as mentioned above.

Neither solution currently works.

That means the bug is somewhere else.

spacedmonkey:
Should we remove this change?

spacedmonkey:
Should we close this PR?

PR #10022, opened 2021-12-17
Media: Avoid taking an existing element's base color

spacedmonkey:
## Summary

While working on another ticket, I noticed that sometimes, the base color was repeated the same color. This was becuase it was reusing the existing element on the page. By passing a unique id, it creates a new element and fixes the issue. 

## Testing Instructions

1.

## Checklist

---

Fixes #

### DIFF
--- a/packages/story-editor/src/utils/getMediaBaseColor.js
+++ b/packages/story-editor/src/utils/getMediaBaseColor.js
@@ -47,7 +47,7 @@ function getImgNodeKey(elementId) {
   return `${IMG_NODE}_${elementId}`;
 }
 
-export async function getMediaBaseColor(src) {
+export async function getMediaBaseColor(src, id) {
   if (!src) {
     return Promise.reject(new Error('No source to image'));
   }
@@ -55,6 +55,7 @@ export async function getMediaBaseColor(src) {
   const trackTiming = getTimeTracker('load_get_base_color');
   try {
     color = await setOrCreateImage({
+      id,
       src,
       width: 10,
       height: 'auto',
--- a/packages/story-editor/src/app/media/utils/useDetectBaseColor.js
+++ b/packages/story-editor/src/app/media/utils/useDetectBaseColor.js
@@ -100,7 +100,7 @@ function useDetectBaseColor({ updateMediaElement }) {
       }
       const imageSrcProxied = getProxiedUrl(resource, imageSrc);
       try {
-        const color = await getMediaBaseColor(imageSrcProxied);
+        const color = await getMediaBaseColor(imageSrcProxied, id);
         await saveBaseColor(resource, color);
       } catch (error) {
         // Do nothing for now.


swissspidy:
This makes me wonder: Why are the img elements added to the DOM in the first place? Seems unnecessary. Avoiding that would fix this too.

spacedmonkey:
> This makes me wonder: Why are the img elements added to the DOM in the first place? Seems unnecessary. Avoiding that would fix this too.

@miina I wonder this as well. Can you explain why this is needed?

miina:
> @miina I wonder this as well. Can you explain why this is needed?

Perhaps the base color was previously created based on some URLs of images when they weren't anywhere yet?
I don't remember why it was added like this, by all means, feel free to change it! The `baseColor` logic has been completely reworked meanwhile, we should def opt for the most suitable approach.

swissspidy:
I‘d recommend opening an accompanying ticket if this requires more work.

spacedmonkey:
I now just create a unquie id for every base color generation. I will create a ticket for refactoring the logic. It is a little hard to understand as it. Not sure I understand why we create a div, we can just pass in a node.

swissspidy:
If we're gonna fix it, let's fix it properly in the new year

packages/story-editor/src/utils/getMediaBaseColor.js
```diff
@@ -93,17 +94,16 @@ export function setOrCreateImage(
   imageData,
   getOnloadCallback = getDefaultOnloadCallback
 ) {
-  const { src, id, height, width } = imageData;
+  const { src, height, width } = imageData;
+  const id = uuidv4();
   return new Promise((resolve, reject) => {
     const NODE_KEY = getImgNodeKey(id);
-    let imgNode = document.body[NODE_KEY];
```
miina: One reason for re-using the same image node was not to create a bunch of nodes hanging around in the DOM but just reusing the same image tag. Why is adding a new node every time better than using the same one node?

If I understood correctly, Pascal's question was generally about adding the node to the DOM, this still does it, no?

miina:
> It is a little hard to understand as it. Not sure I understand why we create a div, we can just pass in a node.

@spacedmonkey the `div` was previously created _because_ we added the node to the DOM -- it was a wrapper that visually hid the added image from the screen, and always the same wrapper was used to just add it once (not a separate one per every element). If we can just pass the image node, there's no reason at all for creating that `div`.

spacedmonkey:
I had a quick look into this and this code needs refactoring now. It is very confusing to follow. I have created issue for this. https://github.com/google/web-stories-wp/issues/10025

For now, I am going back to simple fix, which means passing an id, which fixes the duplicate color issues, I was seeing. Longer term, I really want to refactor this code.

swissspidy:
I'd like to properly fix and refactor this in the current/upcoming sprint.

spacedmonkey:
Created issue https://github.com/google/web-stories-wp/issues/10099 and closing as larger refactor is needed.

PR #3935, opened 2020-08-14
Optimize Image element loading

merapi:
## User-facing changes

When you click on the image in the media library you should instantly see the image on the canvas.

## Testing Instructions

When you click on the image in the media library you should instantly see the image on the canvas. It will use thumbnail so looking closely you will see low quality of the image, after a while (depending on your connection speed, DevTools connection limiter can be useful) you will see a better version of the image.

When image caching is disabled on your WordPress instance then you will see no instant effect.
Like on our QA env? Please check it on a local/clean WordPress instance.

---

Fixes #2615

PR #10024, opened 2021-12-17
Animations: Refactor animation state to improve performance

### LINKED ISSUE #9742: Animations: Unnecessary re-renders caused by `WAAPIWrapper`
## Bug Description

This finding was kindly shared with us by rtCamp engineers:

**Issue:** Any element wrapped by `StoryAnimation.WAAPIWrapper` / `StoryAnimation.Provider` get's re-rendered whenever an element gets moved to another place. This also causes videos or images to be loaded again due to the re-renders, resulting in tons of HTTP requests.

**Cause:** currently, all the animation properties of display elements are being stored in a `Map`, so whenever the (x, y) position of an element changes, that `Map` gets updated and all of those elements depending on animations get re-rendered.

Relevant code:

https://github.com/google/web-stories-wp/blob/31776678fb986ad1f11461ca9febf900ec4f7c3d/packages/animation/src/components/provider.js#L81-L118

## Expected Behaviour

No unexpected re-renders

## Steps to Reproduce

Check renders in React dev tools, see also video below showing the HTTP requests

## Screenshots

https://share.vidyard.com/watch/PTXDn2R1L7XqPyX57Y3tg4?

## Additional Context

- Plugin Version: 1.14.0
- WordPress Version: 5.8
- Operating System: macOS
- Browser: Chrome

maxyinger:
## Context
Original Issue: https://github.com/google/web-stories-wp/issues/9742

Found problem and solution proposed here: https://github.com/google/web-stories-wp/issues/9742#issuecomment-974466215

## Summary
Updates the animation provider and WAAPI animation format to work better with react and media.

## Relevant Technical Choices
- No more composed or changing wrappers from the animation provider makes react not remount the children whenever they change
- No more unnecessary animation regenerations (only generates new animations that have actually updated)
- Increased use-context-selector specificity (coupled with only necessary animation generation) allows for less WAAPIWrapper rerenders (only wrapper that received update rerenders)

## To-do
NA

## User-facing changes
NA

## Testing Instructions

1. Click through all background and foreground animations and see that they still work as they did before

(This isn't user facing, but devs can verify this) 
Add a bunch of media with animations to a page and open up the developer tools network tab. Filter network requests by media. See that

 the media isn't reloading when you move an elements position.

**before**
https://user-images.githubusercontent.com/35983235/146572545-cd44e7f0-db03-4fc6-a6ad-8f5f1907f504.mp4

**after**

https://user-images.githubusercontent.com/35983235/146572593-0334fafa-cf49-4cc4-82aa-a1dc80f5f693.mp4

## Checklist

---

Fixes #9742


packages/animation/src/components/WAAPIWrapper.js
```diff
@@ -13,67 +13,91 @@
...
   );
 
+  // Parents/Wrappers need to stay consistent here and only have prop changes
+  // to allow react's reconcilliation algorithm to function properly
+  // and not generate a new subtree of DOM nodes:
+  // https://github.com/facebook/react/issues/3965
+  //
+  // To Accomodate for this, we're setting a max of 3 levels deep. All stories built recently should
+  // only have 1 animation per element, however this allows for backward compatibility with old stories
+  // created from legacy templates that use stacked animation parts on single elements
```
samwhale, line 83: 🤯

packages/animation/src/components/provider.js
```diff
@@ -71,37 +71,96 @@ const filterWAAPIAnimations = ({ animations, selectedElementIds }) =>
       )
     : animations;
 
+const STABLE_ARRAY = [];
+function mapHasReferencialEntry(map, [key, val]) {
```
samwhale: I think this is spelled `Referential`, no?

packages/animation/src/components/test/WAAPIWrapper.js
```diff
@@ -17,84 +17,160 @@
...
+import { memo } from '@web-stories-wp/react';
+import { render, act } from '@testing-library/react';
 
 /**
  * Internal dependencies
  */
-import * as useStoryAnimationContext from '../useStoryAnimationContext';
+/**
+ * Internal dependencies
+ */
```
samwhale:
```suggestion
```

samwhale (approved):
After chatting with you yesterday about performance and how to approach this stuff, I feel like I understood this PR a lot better. I see exactly what you said we'd see in the PR description. Nice work ✨

BrittanyIRL (approved):
LGTM and I love the test additions.  Great find.

packages/animation/src/components/provider.js
maxyinger: ope. yup. grazi

BrookeGraham (approved):
🤯

BrittanyIRL:
🤔  
https://github.com/google/web-stories-wp/runs/4791701120?check_suite_focus=true#step:6:3694

```
Chrome Headless 93.0.4577.0 (Linux x86_64): Executed 410 of 464 (5 FAILED) (skipped 54) (38 mins 28.961 secs / 37 mins 55.952 secs)
TOTAL: 5 FAILED, 405 SUCCESS
```

None of these are related to this update

maxyinger:
> None of these are related to this update

Yea I'm a little stumped on how these are failing. They all seem to be similar in that they are all image srcs not updating. In the tests they visually appear to work correctly because they have the proper srcset, just not the right src attribute. I'll dig around a bit

maxyinger:
> Chrome Headless 93.0.4577.0 (Linux x86_64): Executed 410 of 464 (5 FAILED) (skipped 54) (38 mins 28.961 secs / 37 mins 55.952 secs) TOTAL: 5 FAILED, 405 SUCCESS

Ah so I think this is an existing issue, it was just never caught until we made this update.

If you go to main, you can replicate this issue by replacing one background image with another and inspecting the carousel which will have the old background image as `src`, but the new background image as `srcset`. I believe this is because the carousel doesn't have the animation wrapper, so the image element never gets remounted, whereas prior to this PR, elements with an animation wrapper got re-mounted, so the issue of src & srcset getting out of sync just never showed up on the canvas.

Now that this PR doesn't remount elements when the animation updates, we're seeing the issue of `src` & `srcset` getting out of sync on the canvas, and the integration tests are highlighting that.

We could just update the integration tests to compare `element.currentSrc` instead of `element.src` but I feel like that's kind of glossing over the problem.

I'm looking at `elements/media/imageDisplay.js` right now and feel like I'm missing a lot of context around decisions that were made here. Also there's no git history around them because all of this behavior was implemented before moving around the packages

swissspidy:
cc @miina perhaps you have insights into the above?

maxyinger:
@miina @swissspidy @BrittanyIRL

What plan of attack do you want on this? Happy to just skip the tests and file a bug, or lump a quick bug fix in with this PR. Open to whatever.

swissspidy:
> If you go to main, you can replicate this issue by replacing one background image with another and inspecting the carousel which will have the old background image as `src`, but the new background image as `srcset`.

Yep, I can see reproduce it 👍 

> I'm looking at `elements/media/imageDisplay.js` right now and feel like I'm missing a lot of context around decisions that were made here.
> Also there's no git history around them because all of this behavior was implemented before moving around the packages

Yeah that's a bit unfortunate, we somehow broke Git history with one of the moves there 😞 

But the most important change is still discoverable in git blame:

#3935 introduce the most relevant changes there. Made by none other than @merapi 🙂 

It would be great to get a better sense of this bug and and an estimate for fixing it.

So perhaps @merapi remembers this and can share more details or has an idea for this?

maxyinger:
created followup ticket here:
https://github.com/google/web-stories-wp/issues/10192

PR #10028, opened 2021-12-19
Fix Color Picker: Saving text background color saves text color instead

### LINKED ISSUE #9891: Saving text background color saves text color instead
## Bug Description

When I want to save the text element's background color, it saves the text color instead.

## Expected Behaviour

The bg color should be saved.

## Steps to Reproduce

1. Insert text element
2. Change bg color
3. Try to save bg color

## Screenshots

https://user-images.githubusercontent.com/841956/144714656-96795a83-f5e4-44a5-81fe-2f772cd04dec.mov

## Additional Context

- Plugin Version: 1.16.0-alpha
- WordPress Version: 5.9 Beta
- Operating System: macOS
- Browser: Chrome

Trapsta:
## Context

Currently, when you click a text element's background color and try to save the background color to the local or global palettes, the element's text color is saved instead. 

## Summary

I am making use of the `changedStyle` prop in the `ColorPicker` component which is currently used for implementing changes in the element's attributes, to detect from which attribute the color picker was triggered from and save the correct color / background color accordingly. 

## User-facing changes

| Before | After |
|--------|-------|
|   144714656-96795a83-f5e4-44a5-81fe-2f772cd04dec |  Screen Recording 2021-12-19 at 22 33 05  |

## Testing Instructions

1. Insert text element
2. Change bg color
3. Try to save bg color
4. make sure the correct color is added to both the local and global colors.
5. Make sure saving the text color also works correctly.

No

No

No

## Checklist

---

Fixes #9891


packages/story-editor/src/components/colorPicker/colorAdd.js
```diff
@@ -51,9 +51,15 @@ const AddColorAction = styled.button`
...
-function ColorAdd({ isLocal = false, isGlobal = false, ...props }) {
+function ColorAdd({
+  isLocal = false,
+  isGlobal = false,
+  isBackground = false,
+  ...props
+}) {
   const { addGlobalPreset, addLocalPreset } = useAddPreset({
     presetType: PRESET_TYPES.COLOR,
+    getBackground: isBackground,
```
miina: Maybe `isBackgroundColor`? `getBackground` sounds like a function, not a bool.

packages/story-editor/src/components/colorPicker/basicColorList.js
```diff
@@ -108,6 +109,9 @@ function BasicColorList({
   const colorType = isLocal
     ? __('local', 'web-stories')
     : __('global', 'web-stories');
+
+  const isBackground = changedStyle !== 'color';
```
miina: Wouldn't `changedStyle === 'background'` work here?

miina (approved):
Just 2 nits, otherwise looks good.

miina:
@Trapsta Forgot to add that we should also add a new Karma test for it.

packages/story-editor/src/components/colorPicker/basicColorList.js
miina: (Since there are other `changedStyle` values as well possible, such as "border-color", would be better if we would check directly if the `changedStyle` is `background`.)

PR #10044, opened 2021-12-20
Performance: Clean Up Excessive pageHash Updates

### LINKED ISSUE #10054: Performance: pageHash memoization
It seems that `pageHash`, which is memoized and stringify-ing `currentPage` is getting called about 40 times per each actual update to `currentPage` resulting in an excess of components receiving updates. Let's fix it!

BrittanyIRL:
## Summary

`pageHash` was getting set in `usePageAsCanvas`, while it was memoized it will always see the `currentPage` as a new object whenever `currentPage` gets updated. Resulting in `pageHash` running about an average of 40 times per each actual `currentPage` update. The performance improvement here isn't night and day but it drastically cuts down on the excess rewrites and perceived "new" props getting passed around.  It cuts commits per data update according to react profiler by about half, this results in some good scripting performance updates as well. 

_all tests done on 2018 macbook pro_ 

**Going from one selected element to another** 
| Before | After |
| --- | --- | 
| 69 commits | 38 commits | 

**Updating the color of a selected element**
| Before | After |
| --- | --- | 
| 36 commits | 19 commits | 

**Positioning an existing element that is already selected**
| Before | After |
| --- | --- | 
| 144 commits | 89 commits | 

## Relevant Technical Choices

Stringify the `pageHash` when setting `currentPage` in `useStory` hook to avoid excess "updates" based on not being able to deeply compare `currentPage` in memoization. 

## To-do

- [x] Document metrics 
- [x] Create and attach issue for tracking 

## Testing Instructions

1.

## Checklist

---

Fixes #10054

### DIFF
--- a/packages/story-editor/src/app/story/storyProvider.js
+++ b/packages/story-editor/src/app/story/storyProvider.js
@@ -56,26 +56,33 @@ function StoryProvider({ storyId, initialEdits, children }) {
   useEffect(() => setHashPageId(current), [current, setHashPageId]);
 
   // Generate current page info.
-  const { currentPageId, currentPageIndex, currentPageNumber, currentPage } =
-    useMemo(() => {
-      if (!current) {
-        return {
-          currentPageId: null,
-          currentPageIndex: null,
-          currentPageNumber: null,
-          currentPage: null,
-        };
-      }
-      const index = pages.findIndex(({ id }) => id === current);
-      const number = index + 1;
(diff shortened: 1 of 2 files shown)


maxyinger (approved):
nice! lgtm

BrookeGraham (approved):
Great find, and decent gains. 🎉

samwhale (approved):
Nice this is super smart

PR #10050, opened 2021-12-22
Design System: Switch - allow axe test to pick up background color 

### LINKED ISSUE #9943: Design System: Update Toggle Component to Avoid aXe Contrast False Positive
## Context

While adding some no violation checks to karma, the document tab flagged a bad color contrast for the active radio option in the toggle button: 

It's flagging  as `color contrast of 1.59 (foreground color: #131516, background color: #373a3b,` the[ real contrast is 7.02](https://webaim.org/resources/contrastchecker/?fcolor=C8CBCC&bcolor=373A3B)

So the ask here is to take a look at the toggle button in the design system to adjust is and get the real color contrast to be picked up in aXe so we can enable this check on the entire document tab and avoid other false positives where this component is used.

You can check this by enabling the disabled test in components/inspector/karma that correlates to this ticket number.

BrookeGraham:
## Context

The axe test coverage is failing in the Design System's Switch component. This is due to background-color being set on an absolutely positioned span which is flagging as color contrast of 1.59 (foreground color: #131516, background color: #373a3b, the real contrast is 7.02. 

## Summary

Allows the axe test coverage to grab the correct background color from the Switch component. 

## Relevant Technical Choices

Changing the background color on the `RadioButtonLabel` when the button `isActive`. Added the radius here as to keep the rounded edges. Also, changed the `RadioButtonLabel` to have `all` on the transition in order to maintain the nice little slidey animation. 

### Before
before

### After
after

## Testing Instructions

1.

no

no

no

## Checklist

---

Fixes #9943

### DIFF
--- a/packages/design-system/src/components/switch/index.js
+++ b/packages/design-system/src/components/switch/index.js
@@ -79,19 +79,6 @@ const SwitchContainer = styled.div`
             background-color: ${theme.colors.interactiveBg.primaryHover};
           }
         `};
-
-          cursor: default;
-          background: ${theme.colors.divider.tertiary};
-
-          ${SlidingButton} {
-            background-color: ${theme.colors.interactiveBg.disable};
-          }
-        `
-      : css`
-          :hover ${SlidingButton} {
-            background-color: ${theme.colors.interactiveBg.primaryHover};
-          }
-        `};
 `;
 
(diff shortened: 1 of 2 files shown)


BrittanyIRL (approved):
Smart fix! Error's gone and switch works the same 🎉

samwhale (approved):
🥳

maxyinger (approved):

PR #10051, opened 2021-12-22
Performance: Limit Renders in the Layers Panel

maxyinger:
## Context
Performance optimization & documentation around several key interactions in the Layers Panel. Makes key actions only render relevant components in the Layers Panel. Basically now `O(updated layers)` vs `O(all layers)` for these interactions

## Summary
Improving rendering performance of the layers panel with 3 interactions. 

Tested on a page with 50 square shape elements (51 including the background element) with the Layers panel open:
Screen Shot 2021-12-22 at 1 55 36 PM

*Tested on a fully specced 2018 Macbook Pro with no CPU Throttling*

### Element Selection
Test: Changing the selected element in the editor by clicking a different layer in the layers panel. 
Result: **~49.14x faster** 

| Before | After |
| --- | --- |
| ~68.8ms | ~1.4ms |
|  Screen Shot 2021-12-22 at 1 28 19 PM |  Screen Shot 2021-12-22 at 1 28 29 PM |

### Reordering Layers
Test: Drag the selected layer in the layers panel to a new position. 
Result: **~45.48x faster** 

| Before | After |
| --- | --- |
| ~104.6ms | ~2.3ms |
| Screen Shot 2021-12-22 at 1 42 27 PM |  Screen Shot 2021-12-22 at 1 42 38 PM |

### Updating an elements style properties
Test: Drag the previously selected element on the canvas to update its x & y coordinates
Result: **~42.75x faster** 

| Before | After |
| --- | --- |
| Screen Shot 2021-12-22 at 2 01 56 PM |  Screen Shot 2021-12-22 at 2 13 26 PM |

## Relevant Technical Choices
Introduces better (more sensitive to referential equality) patterns around `useContextSelector` more inline with the library's recommended usage:
https://github.com/dai-shi/use-context-selector#usage

## Testing Instructions

1.

## Checklist

---

Fixes #10051

### DIFF
--- a/packages/story-editor/src/app/story/useStory.js
+++ b/packages/story-editor/src/app/story/useStory.js
@@ -35,8 +35,8 @@ const DELIMITER = ':::';
  * @return {Array} array of current pages story element ids in order
  */
 export function useCurrentPageElementIds() {
-  const elementIds = useStory((v) => {
-    const elements = v.state.currentPage?.elements || [];
+  const elementIds = useStory((value) => {
+    const elements = value.state.currentPage?.elements || [];
     return elements.map((element) => element.id).join(DELIMITER);
   });
   return elementIds.split(DELIMITER);
--- a/packages/story-editor/src/app/story/useStory.js
+++ b/packages/story-editor/src/app/story/useStory.js
@@ -27,4 +27,19 @@ function useStory(selector) {
   return useContextSelector(Context, selector ?? identity);
 }
 
+const DELIMITER = ':::';
+/**
+ * Returns an array of current page's story element ids. Only rerenders when there are
+ * element reorders, deletions or additions
+ *
+ * @return {Array} array of current pages story element ids in order
+ */
+export function useCurrentPageElementIds() {
+  const elementIds = useStory((v) => {
+    const elements = v.state.currentPage?.elements || [];
+    return elements.map((element) => element.id).join(DELIMITER);
+  });
+  return elementIds.split(DELIMITER);
+}
+
 export default useStory;
--- a/packages/story-editor/src/app/story/useStory.js
+++ b/packages/story-editor/src/app/story/useStory.js
@@ -27,23 +27,4 @@ function useStory(selector) {
   return useContextSelector(Context, selector ?? identity);
 }
 
-const DELIMITER = ':::';
-/**
- * Returns an array of current page's story element ids. Only rerenders when there are
- * element reorders, deletions or additions
- *
- * @return {Array} array of current pages story element ids in order
- */
-export function useCurrentPageElementIds() {
-  // Joining and splitting with delimiter so that selector returns
-  // a strictly equal result when we have a new id array instance that
-  // is shallowly equal. This prevents empty forced re-renders from
-  // useContextSelector
-  const elementIds = useStory((value) => {
-    const elements = value.state.currentPage?.elements || [];
-    return elements.map((element) => element.id).join(DELIMITER);
-  });
-  return elementIds.split(DELIMITER);
-}
-
 export default useStory;
--- a/packages/story-editor/src/app/story/useStory.js
+++ b/packages/story-editor/src/app/story/useStory.js
@@ -35,6 +35,10 @@ const DELIMITER = ':::';
  * @return {Array} array of current pages story element ids in order
  */
 export function useCurrentPageElementIds() {
+  // Joining and splitting with delimiter so that selector returns
+  // a shallow equal result when we have a new id array instance that
+  // is deeply equal. This prevents empty forced re-renders from
+  // useContextSelector
   const elementIds = useStory((value) => {
     const elements = value.state.currentPage?.elements || [];
     return elements.map((element) => element.id).join(DELIMITER);
--- a/packages/story-editor/src/app/story/useStory.js
+++ b/packages/story-editor/src/app/story/useStory.js
@@ -36,8 +36,8 @@ const DELIMITER = ':::';
  */
 export function useCurrentPageElementIds() {
   // Joining and splitting with delimiter so that selector returns
-  // a shallow equal result when we have a new id array instance that
-  // is deeply equal. This prevents empty forced re-renders from
+  // a strictly equal result when we have a new id array instance that
+  // is shallowly equal. This prevents empty forced re-renders from
   // useContextSelector
   const elementIds = useStory((value) => {
     const elements = value.state.currentPage?.elements || [];
--- a/packages/story-editor/src/app/story/useStoryReducer/reducers/setSelectedElements.js
+++ b/packages/story-editor/src/app/story/useStoryReducer/reducers/setSelectedElements.js
@@ -41,11 +41,14 @@ import { intersect } from './utils';
  * @return {Object} New state
  */
 function setSelectedElements(state, { elementIds }) {
-  if (!Array.isArray(elementIds)) {
+  const newElementIds =
+    typeof elementIds === 'function' ? elementIds(state.selection) : elementIds;
+
+  if (!Array.isArray(newElementIds)) {
     return state;
   }
 
-  const uniqueElementIds = [...new Set(elementIds)];
+  const uniqueElementIds = [...new Set(newElementIds)];
 
   // They can only be similar if they have the same length
   if (state.selection.length === uniqueElementIds.length) {
--- a/packages/story-editor/src/components/panels/design/layer/layerList.js
+++ b/packages/story-editor/src/components/panels/design/layer/layerList.js
@@ -58,8 +58,8 @@ const ReorderableLayer = memo(function ReorderableLayer({
   position,
   handleStartReordering,
 }) {
-  const element = useStory((v) =>
-    v.state.currentPage?.elements.find((el) => el.id === id)
+  const element = useStory(({ state }) =>
+    state.currentPage?.elements.find((el) => el.id === id)
   );
   return element ? (
     <Fragment key={id}>
--- a/packages/story-editor/src/components/panels/design/layer/layerList.js
+++ b/packages/story-editor/src/components/panels/design/layer/layerList.js
@@ -82,12 +82,12 @@ ReorderableLayer.propTypes = {
 };
 
 function LayerPanel({ layers }) {
-  const arrangeElement = useStory((v) => v.actions.arrangeElement);
+  const arrangeElement = useStory((value) => value.actions.arrangeElement);
   const setSelectedElementsById = useStory(
-    (v) => v.actions.setSelectedElementsById
+    (value) => value.actions.setSelectedElementsById
   );
 
-  const onOpenMenu = useRightClickMenu((v) => v.onOpenMenu);
+  const onOpenMenu = useRightClickMenu((value) => value.onOpenMenu);
 
   const numLayers = layers.length;
 
--- a/packages/story-editor/src/components/panels/design/layer/layerList.js
+++ b/packages/story-editor/src/components/panels/design/layer/layerList.js
@@ -18,7 +18,7 @@
  * External dependencies
  */
 import styled from 'styled-components';
-import { Fragment, useCallback } from '@web-stories-wp/react';
+import { Fragment, useCallback, memo } from '@web-stories-wp/react';
 import PropTypes from 'prop-types';
 import { __ } from '@web-stories-wp/i18n';
 
@@ -53,15 +53,43 @@ const LayerSeparator = styled(ReorderableSeparator)`
   padding: ${LAYER_HEIGHT / 2}px 0;
 `;
 
+const ReorderableLayer = memo(function ReorderableLayer({
+  id,
+  position,
+  handleStartReordering,
+}) {
+  const element = useStory((v) =>
+    v.state.currentPage?.elements.find((el) => el.id === id)
+  );
+  return element ? (
+    <Fragment key={id}>
+      <LayerSeparator position={position + 1} />
+      <ReorderableItem
+        position={position}
+        onStartReordering={handleStartReordering(id)}
+        disabled={element.isBackground}
+      >
+        <Layer element={element} />
+      </ReorderableItem>
+    </Fragment>
+  ) : null;
+});
+
+ReorderableLayer.propTypes = {
+  id: PropTypes.string.isRequired,
+  position: PropTypes.number.isRequired,
+  handleStartReordering: PropTypes.func.isRequired,
+};
+
 function LayerPanel({ layers }) {
-  const { arrangeElement, setSelectedElementsById } = useStory((state) => ({
-    arrangeElement: state.actions.arrangeElement,
-    setSelectedElementsById: state.actions.setSelectedElementsById,
-  }));
+  const arrangeElement = useStory((v) => v.actions.arrangeElement);
+  const setSelectedElementsById = useStory(
+    (v) => v.actions.setSelectedElementsById
+  );
 
-  const { onOpenMenu } = useRightClickMenu();
+  const onOpenMenu = useRightClickMenu((v) => v.onOpenMenu);
 
-  const numLayers = layers && layers.length;
+  const numLayers = layers.length;
 
   const focusCanvas = useFocusCanvas();
   const handleStartReordering = useCallback(
@@ -89,16 +117,12 @@ function LayerPanel({ layers }) {
       getItemSize={() => LAYER_HEIGHT}
     >
       {layers.map((layer) => (
-        <Fragment key={layer.id}>
-          <LayerSeparator position={layer.position + 1} />
-          <ReorderableItem
-            position={layer.position}
-            onStartReordering={handleStartReordering(layer.id)}
-            disabled={layer.isBackground}
-          >
-            <Layer layer={layer} />
-          </ReorderableItem>
-        </Fragment>
+        <ReorderableLayer
+          key={layer.id}
+          id={layer.id}
+          position={layer.position}
+          handleStartReordering={handleStartReordering}
+        />
       ))}
     </LayerList>
   );
--- a/packages/story-editor/src/components/panels/design/layer/layerList.js
+++ b/packages/story-editor/src/components/panels/design/layer/layerList.js
@@ -82,9 +82,11 @@ ReorderableLayer.propTypes = {
 };
 
 function LayerPanel({ layers }) {
-  const arrangeElement = useStory((value) => value.actions.arrangeElement);
-  const setSelectedElementsById = useStory(
-    (value) => value.actions.setSelectedElementsById
+  const { arrangeElement, setSelectedElementsById } = useStory(
+    ({ actions }) => ({
+      arrangeElement: actions.arrangeElement,
+      setSelectedElementsById: actions.setSelectedElementsById,
+    })
   );
 
   const onOpenMenu = useRightClickMenu((value) => value.onOpenMenu);
--- a/packages/story-editor/src/components/panels/design/layer/useLayerSelection.js
+++ b/packages/story-editor/src/components/panels/design/layer/useLayerSelection.js
@@ -22,17 +22,15 @@ import { useCallback } from '@web-stories-wp/react';
 /**
  * Internal dependencies
  */
-import { useStory, storyReducers } from '../../../../app';
+import { useStory } from '../../../../app';
 import useFocusCanvas from '../../../canvas/useFocusCanvas';
 
 function useLayerSelection(layer) {
   const { id: elementId } = layer;
-  const { isSelected, updateStateWithReducer } = useStory(
-    ({ state, actions }) => ({
-      isSelected: state.selectedElementIds.includes(elementId),
-      updateStateWithReducer: actions.updateStateWithReducer,
-    })
-  );
+  const { isSelected, toggleLayer } = useStory(({ state, actions }) => ({
+    isSelected: state.selectedElementIds.includes(elementId),
+    toggleLayer: actions.toggleLayer,
+  }));
 
   const focusCanvas = useFocusCanvas();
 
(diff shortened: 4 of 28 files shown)


BrittanyIRL (approved):
This rules! I like the patterning you're introducing and think the tests you did are a great baseline and show the gains really well.

dreamofabear:
🔥

packages/story-editor/src/components/reorderable/reorderableItem.js
```diff
@@ -36,9 +36,9 @@ function ReorderableItem({
   disabled,
   ...props
 }) {
-  const {
-    actions: { handleStartReordering },
-  } = useReorderable();
+  const handleStartReordering = useReorderable(
+    (v) => v.actions.handleStartReordering
```
swissspidy: What does v stand for everywhere? What‘s wrong with `state`?

packages/story-editor/src/app/story/useStory.js
```diff
@@ -27,4 +27,19 @@ function useStory(selector) {
...
+/**
+ * Returns an array of current page's story element ids. Only rerenders when there are
+ * element reorders, deletions or additions
+ *
+ * @return {Array} array of current pages story element ids in order
+ */
+export function useCurrentPageElementIds() {
+  const elementIds = useStory((v) => {
+    const elements = v.state.currentPage?.elements || [];
+    return elements.map((element) => element.id).join(DELIMITER);
```
swissspidy: What‘s the secret sauce behind joining and splitting by this delimiter?
Can this be made simpler?
If not, can it be explained in a comment?

packages/story-editor/src/components/panels/design/layer/layerList.js
```diff
@@ -53,15 +53,43 @@ const LayerSeparator = styled(ReorderableSeparator)`
...
+  handleStartReordering: PropTypes.func.isRequired,
+};
+
 function LayerPanel({ layers }) {
-  const { arrangeElement, setSelectedElementsById } = useStory((state) => ({
-    arrangeElement: state.actions.arrangeElement,
-    setSelectedElementsById: state.actions.setSelectedElementsById,
-  }));
+  const arrangeElement = useStory((v) => v.actions.arrangeElement);
+  const setSelectedElementsById = useStory(
```
swissspidy: What‘s the benefit of splitting this into two function calls? Just because the lib‘s usage example does it like this?

packages/story-editor/src/components/panels/design/layer/useLayerSelection.js
```diff
@@ -22,75 +22,67 @@ import { useCallback } from '@web-stories-wp/react';
...
-        }
-        setSelectedElementsById({ elementIds });
-      } else if (evt.metaKey) {
-        // Meta pressed. Toggle this layer in the selection.
-        toggleElementInSelection({ elementId });
-      } else {
-        // No special key pressed - just selected this layer and nothing else.
-        setSelectedElementsById({ elementIds: [elementId] });
-      }
+      updateStateWithReducer({
```
swissspidy: So how does this change here improve/decrease the number of re-renders?
Why is it needed? Can‘t this be avoided?
This happens at the cost of exposing reducers nonetheless.

packages/story-editor/src/components/reorderable/reorderableItem.js
maxyinger: `v` is just shorthand for `value`. I think it makes more sense than `state` cus a context holds a `value` which can have refs, state, actions, config constant, etc. Though I agree that `value` can seem a little too vague, I think there's a reason that react makes providers hold and expose a prop named `value`, and I'm just following that convention. I'm assuming the library author is as well with the `v` shorthand.

Also have never liked `state.<state | actions>...` and think that `value.<state | actions>` doesn't conflate the term `state` so much.

Anyway, that's my rationale. At the end of the day it's just a name, so I'm fine with whatever. Just thought I'd pitch it as I propose a pattern where we might be updating the use of it.

packages/story-editor/src/app/story/useStory.js
maxyinger: Yeah, good call, I can absolutely explain it in a comment.

The idea behind this is the same rationale as the rest of the changes here which is that `useContextSelector` limits re-renders by doing strict equality checks amongst return values of the selector. If that equality check is false, it forces a re-render with the updated state:
Screen Shot 2021-12-24 at 11 04 06 AM

src (don't have to focus on it too much, but can see through the comments that it's basically forcing a re-render with the updated selected value if referential equality changes):
https://github.com/dai-shi/use-context-selector/blob/7ca35cf68d08ba930a8b91c129ab00d1798300b9/src/index.ts#L144-L180

I tried to explain it in the doc string, but can put more technical detail comments in the implementation.

Basically if a component doesn't need element properties for its implementation, but does need something like the number of elements and their respective identifiers, this setup prevents that component from rendering when it doesn't need to.

ie
- there are 3 elements `[{ id: 'a', ... }, { id: 'b', ... }, {id: 'c', ... }]`
- a component can use this and get the info that the elements follow this `['a', 'b', 'c']`
- we update element `b` to have a new `backgroundColor` property
- the component we're using this in would otherwise have an empty re-render that costs some time and would trigger a possible empty re-render in all of it's descendants. (could be less than 1ms per element, but if our component has a subtree of 100 elements, that adds up. This is the primary performance issue we're running into in the codebase right now)
- with this hook use context selector would check the string `a:::b:::c` to the string `a:::b:::c` and not trigger a re-render in the component we're using this in, while still retaining accurate information that the current element structure looks like `['a', 'b', 'c']`

**tldr;** it's a matter of limiting renders to only places that need a visual update on the screen.

packages/story-editor/src/components/panels/design/layer/layerList.js
maxyinger: yea so the library is recommending this usage cus of the referential equality blurb I mentioned above.

if the selector returns a new object, this hook will force a re-render of this component every time the provider this context is coming from re-renders. Basically by returning a new object from the selector we're opting out of all the performance benefits `useContextSelector` provides.

packages/story-editor/src/components/panels/design/layer/useLayerSelection.js
maxyinger: yea I considered prefixing this action with `dangerous` something like `dangerouslyUpdateStateWithReducer` like how react has `dangerouslySetInnerHtml`. Intent behind this is it would discourage people from using it unless they were really confident they needed it for their use case.

Basic idea tho is limiting re-renders caused by `useLayerSelection`. If we reference story state in this hook (which is ultimately just providing an action) that story state will cause a re-render in the component consuming this every time the state updates. 

Specifically here, we only want one `Layer` element to update when there's a change to an element instead of having all `Layer` elements update when there's a change to an element. If we were to pull in all story elements to reference when creating an action for each `Layer`, all `Layers` would update when a single element updated when in actuality, there was only 1 or no visual changes for the user.

As far as other approaches we could use to achieve the same benefits, we could create a new reducer that basically takes a key as it's payload and share code between the composed reducers here, however I went with the approach above cus it was the quickest way to share code and achieve the benefits mentioned here.

I imagine we'll need to compose reducers more throughout the codebase as we audit their performance, and this setup seems like a quicker win for all of those possible cases instead of writing new reducers every time we need to compose existing ones.

maxyinger:
@swissspidy @choumx If y'all are at all interested, I'd be happy to do a little meeting over the thought process behind our `use-context-selector` patterns introduced here.

Writing about it can get a little verbose, but it's relatively simple concepts.

Happy to write up a reference doc too if that's preferred, or if you'd like it in addition to a little 20-30 minute meeting.

Let me know. Happy Holidays 🥂

packages/story-editor/src/components/reorderable/reorderableItem.js
samwhale: Makes sense - maybe we write `value` instead of `v` just for readability? Small nit but that way we don't conflate `state` too much too.

packages/story-editor/src/app/story/useStoryReducer/reducers/setSelectedElements.js
```diff
@@ -41,11 +41,14 @@ import { intersect } from './utils';
...
-  if (!Array.isArray(elementIds)) {
+  const newElementIds =
+    typeof elementIds === 'function' ? elementIds(state.selection) : elementIds;
+
+  if (!Array.isArray(newElementIds)) {
     return state;
   }
 
-  const uniqueElementIds = [...new Set(elementIds)];
+  const uniqueElementIds = [...new Set(newElementIds)];
```
samwhale, line 51: Sets for the win 😤

packages/story-editor/src/components/panels/design/layer/layerList.js
```diff
@@ -89,16 +117,12 @@ function LayerPanel({ layers }) {
       getItemSize={() => LAYER_HEIGHT}
     >
       {layers.map((layer) => (
-        <Fragment key={layer.id}>
-          <LayerSeparator position={layer.position + 1} />
```
samwhale, line 93: 👍

packages/story-editor/src/components/panels/design/layer/layerList.js
```diff
@@ -53,15 +53,43 @@ const LayerSeparator = styled(ReorderableSeparator)`
...
+  id,
+  position,
+  handleStartReordering,
+}) {
+  const element = useStory((v) =>
+    v.state.currentPage?.elements.find((el) => el.id === id)
+  );
+  return element ? (
+    <Fragment key={id}>
+      <LayerSeparator position={position + 1} />
```
samwhale, line 66: 👍

packages/story-editor/src/components/panels/panel/shared/title.js
```diff
@@ -88,7 +88,6 @@ const IconWrapper = styled.div`
   position: relative;
   width: 32px;
   height: 32px;
-
```
samwhale, line 91: 😕 what do you think about the spaces here? I like spaces for readability but it's not a big deal.

samwhale (approved):
Walking through these changes and how to identify them was very helpful. Gave me a lot of context about why these changes were made.

Specifically splitting up grabbing certain pieces of state off of `useContextSelector` so that we don't re-render children on every render.

packages/story-editor/src/components/panels/panel/shared/title.js
maxyinger: oh thanks for pointing out. this wasn't intentional. was a byproduct of fixing a bad format from the pre-commit hook. will add those spaces back in. thanks

packages/story-editor/src/components/panels/design/layer/layerList.js
```diff
@@ -53,15 +53,43 @@ const LayerSeparator = styled(ReorderableSeparator)`
   padding: ${LAYER_HEIGHT / 2}px 0;
 `;
 
+const ReorderableLayer = memo(function ReorderableLayer({
+  id,
+  position,
+  handleStartReordering,
+}) {
+  const element = useStory((v) =>
```
BrookeGraham: nit: value here instead of v for consistency

BrookeGraham (approved):

packages/story-editor/src/components/reorderable/reorderableItem.js
miina: > Anyway, that's my rationale. At the end of the day it's just a name, so I'm fine with whatever. Just thought I'd pitch it as I propose a pattern where we might be updating the use of it.

@littlemilkstudio Makes sense replacing `state.state.foo` with `value.state.foo` but wondering why is `useReorderable((value)) => value.actions...` better than `useReorderable(({ actions })) => actions...`?

packages/story-editor/src/app/story/useStory.js
miina: > The idea behind this is the same rationale as the rest of the changes here which is that useContextSelector limits re-renders by doing strict equality checks amongst return values of the selector. If that equality check is false, it forces a re-render with the updated state:

An alternative could also be using custom equality functions (here and elsewhere where applicable), especially if `useCurrentPageElementIds` is used in one place only, perhaps would make it more simple and avoid adding helpers just for getting past the default equality check. Thoughts?

swissspidy: Got it, so just returning a string to make the equality check work. Just happening to use a custom delimiter to make it more robust.

Apologies, I was on mobile when I wrote this comment, but now with the additional context when seeing the whole code + the explanation things are more clear.

In that case, I'd have suggested the same thing as Miina, using a custom equality function, which can be passed as the second argument to `useStory`.

Also, if this used more often, we could consider storing the current page element list in the StoryProvider itself so we could do something like `useStory((value) => value.state.currentPageElementIds)`.

Either method would help avoiding adding helpers like that.

packages/story-editor/src/components/panels/design/layer/layerList.js
swissspidy: We're returning objects from the selectors **a lot**, basically everywhere, and I have never really seen issues with unnecessary re-renders because of it. Do you perhaps have a small repro case to demonstrate that it actually doesn't work as intended?

If returning objects like this does not work as intended, we would need to do a massive overhaul of the code base to enforce consistent usage in places with improper use.

packages/story-editor/src/components/panels/panel/shared/title.js
```diff
@@ -111,17 +110,16 @@ const Collapse = styled.button`
   cursor: pointer;
   margin-left: -12px;
   transition: ${BUTTON_TRANSITION_TIMING};
-
   &:hover,
-  &.${ThemeGlobals.FOCUS_VISIBLE_SELECTOR}, &[${ThemeGlobals.FOCUS_VISIBLE_DATA_ATTRIBUTE}] {
+   &.${ThemeGlobals.FOCUS_VISIBLE_SELECTOR}, &[${ThemeGlobals.FOCUS_VISIBLE_DATA_ATTRIBUTE}] {
```
swissspidy: How did this indentation happen here?

packages/story-editor/src/app/story/useStory.js
maxyinger: yea, basically anything that doesn't make page level components rerender when the property of a single element updates is fine by me.

As far as custom equality functions, I'm not seeing where `useContextSelector` exposes that as part of its API. Is it somewhere in there that I'm missing? or where you thinking of something else @miina ?
https://github.com/dai-shi/use-context-selector#usecontextselector

As for storing `currentPageElementIds` in state, it feels like duplicate state that just exposes our surface area for things to get out of sync. I feel like deriving this is more robust. thoughts?

this hook would primarily get used for page previews, or any of the element layers on the canvas as far as usage goes.

packages/story-editor/src/components/panels/design/layer/layerList.js
maxyinger: yea we are returning objects `a lot`. luckily this doesn't break anything and doesn't change user facing functionality. This new pattern is a performance enhancement and should primarily be used for documented areas of new performance gains (and possibly newly added code if we want).

There's nothing that suffers from our old pattern except for performance. this is purely an enhancement

maxyinger: can absolutely spin up a couple quick examples. Will also ping the slack channel cus this question seems to come up a lot in my recent perf updates

maxyinger: Here's a very stripped down example with lots of comments explaining the thought process. You can see that `SomeOtherComponent` uses the existing pattern of creating a new object and destructuring it. You can also see that `SomeOtherComponent` re-render when `Count` updates. However, `Count` doesn't re-render when `SomeOtherComponent` updates because it's using this newer pattern of getting more fine grained parts of the context value with it's selector: 

https://codesandbox.io/s/nostalgic-elgamal-5etzb?file=/src/App.js

maxyinger: Here's a more in depth codesandbox exploring this new useContextSelector pattern even further, but there's no comments really tho, so I advise checking out the first codesandbox to gain better context around what's goin on:

https://codesandbox.io/s/competent-agnesi-k2ldj?file=/src/App.js

packages/story-editor/src/app/story/useStory.js
miina: @littlemilkstudio We are currently using a custom equality function in one place, by just accepting it in our hook wrapper:
https://github.com/google/web-stories-wp/blob/2d0a090015349a786e676864c5652df5a134893a/packages/story-editor/src/components/dropTargets/useDropTargets.js#L26-L28

And then adding the custom equality function here (also for performance reasons):
https://github.com/google/web-stories-wp/blob/2d0a090015349a786e676864c5652df5a134893a/packages/story-editor/src/components/library/panes/media/common/innerElement.js#L123-L134

Perhaps we could just accept the equality function for [`useStory`](https://github.com/google/web-stories-wp/blob/main/packages/story-editor/src/app/story/useStory.js#L26) as well?

packages/story-editor/src/components/panels/design/layer/layerList.js
swissspidy: @littlemilkstudio Thanks a lot for the sandboxes! That's very helpful for reasoning about.

Now, your examples demonstrate the effect when using plain `use-context-selector`, but we're using this "patched" version which uses `shallowEqual` by default:

https://github.com/google/web-stories-wp/blob/2d0a090015349a786e676864c5652df5a134893a/packages/react/src/useContextSelector.js

See https://codesandbox.io/s/nice-alex-k5m9l and how there are no re-renders anymore with this patched version.

packages/story-editor/src/app/story/useStory.js
maxyinger: @miina thanks for the reference. didn't realize we added a wrapper around `useContextSelector` to enable other equality functions.

Since shallow equals is the default comparison method, just doing something like:
```
const currentPageElementIds = useStory(({state}) => state.currentPage.elements.map(el => el.id))
```
should achieve the same result we're getting here with joining and splitting with a delimiter. No need for this custom hook with a delimiter. Thanks for walking through this with me

maxyinger: Ahh never mind. the version of `shallowEqual` we're using only applies to objects, it's not actually shallow equaling arrays:

https://github.com/gaearon/react-pure-render/blob/master/src/shallowEqual.js

maxyinger: guess I do need to write a different equality function that basically does shallow equal for arrays and use it here

packages/story-editor/src/components/panels/design/layer/layerList.js
maxyinger: Ah thanks for pointing out that we're using a patched version. Knowing that, we don't need to separate out all the pieces of context value we're grabbing.

I think you may have forgot to save your sandbox? I'm still seeing the same one I sent you

packages/story-editor/src/app/story/useStory.js
maxyinger: updated our shallow-equal to work with arrays as well using this package:
https://www.npmjs.com/package/shallow-equal

packages/story-editor/src/components/panels/panel/shared/title.js
maxyinger: our pre-commit hook style lint step has been misformatting some files. typically it's on random files you don't touch and it just completely breaks them. To fix I normally go back and undo the change and commit the reverts with a `--no-verify` so that the style lint step doesn't break anything again.

This indent happened as a result of that.

The misformatting is still intermittent, but it typically happens on files that are not part of your PR so I don't worry too much about running the pre-commit hook on those files

packages/story-editor/src/components/panels/design/layer/useLayerSelection.js
swissspidy: @miina @barklund curious to hear your thoughts on this part too

miina: @littlemilkstudio Could we perhaps instead use a custom equality function here as well with `useStory`? E.g. only consider the value has changed for the currently selected elements -- `selectedElementIds.includes(elementId)`  (if I understood correctly that's the goal).

maxyinger: yea I that was my first approach, just pulling the relevant data needed for this action off of `useStory` with good selectors.

The issue I ran into is that this hook is rendered in every `Layer` element, and the action this hook is supplying needs the indices of the selected elements to perform the action.

If we reference any of the selected element indices, this hook then needs to re-render every time that data updates (no matter how specific the derived data is). This would then cause every layer element to re-render whenever those indices changed (every selection change & most layer position changes).

The conclusion I came to is that all of the selected element index data needs to be derived in the scope of the reducer to meet the required perf constraints (renders in O(effected elements) vs O(all elements)).

I'm open to changing any implementation details, as long as we're getting the relevant state data in the closure of the reducer to prevent the original re-render issue.

Another way we could achieve this, which I'm open to if everyones opposed to this implementation, is to create a `SELECT_LAYER` action in the `useStory` reducer like: 
```
dispatch({
  type: 'SELECT_LAYER',
  payload: {
    metaKey,
    shiftKey,
    elementId
  }
})
```

And basically just make that the reducer I've created here. Thoughts?

miina: > If we reference any of the selected element indices, this hook then needs to re-render every time that data updates (no matter how specific the derived data is). This would then cause every layer element to re-render whenever those indices changed (every selection change & most layer position changes).

What I was thinking of is that maybe we could create an `equalityFn` for the `useStory` hook here so that it would only re-render for the selected elements (since we only need indices of the selected elements), wouldn't that work?

E.g. something like
```
const { ... } = useStory(
    ({ state, actions }) => ({
      ...
    }),
    (...) => {
      // If the element of this layer is not selected, do not re-render.
      if (hasRenderedAlready && !selectedElementIds.includes(elementId)) {
        return true;
      }
      // Some other conditions.
      ...
    };
  );
```

The example above is very simplified of course and I didn't look into it thoroughly. Perhaps you already looked into this and came to the conclusion that it wouldn't work.

maxyinger: yea I guess where my thinking is diverging is that every layer element needs to know the selected elements indices for this action, not just the selected elements. This is kind of regardless of re-renders, just essential to the behavior of the action we're creating here.

If we're pulling the selected element indices through our useStory hook (however we're deriving it or altering the comparison fn), we'd need every layer to re-render if we want the functionality of the action to work.

maxyinger: I guess another point I want to make is that we really shouldn't be deriving actions based off state they're affecting. Kind of feels like it goes against this core principle and defeats the benefits of a reducer:
```
const [count, setCount] = useState(0);
const incrementCount = useCallback(() => setCount(count + 1), [count]);
```
vs
```
const [count, setCount] = useState(0);
const incrementCount = useCallback(() => setCount(state => state + 1), []);
```

miina: Alright.
@barklund Would be great to hear your thoughts here, too.

barklund: I'm not sure why this whole thing isn't just another action inside the reducer? Why are you creating a new generic action to allow any function to be invoked and only using it once, rather than create a specific action for this specific purpose?

I hear you on the "update state based on state itself", but there are many actions already doing that - they're just handling specific situations and updating state based on the arguments given.

For this entire purpose, why not just make a new action called `toggleLayer` that takes the required arguments (probably just `isMetaPressed` and `elementId`) and move all the logic inside this action?

It would also be a lot easier to write unit tests for. This current thing here is completely untested and probably rather hard to properly test exhaustively.

Perhaps I'm missing something, that invalidates this approach?

swissspidy: Quote from above that answers this question:

> As far as other approaches we could use to achieve the same benefits, we could create a new reducer that basically takes a key as it's payload and share code between the composed reducers here, however I went with the approach above cus it was the quickest way to share code and achieve the benefits mentioned here.
>
> I imagine we'll need to compose reducers more throughout the codebase as we audit their performance, and this setup seems like a quicker win for all of those possible cases instead of writing new reducers every time we need to compose existing ones.

I also think a new dedicated action for this would be cleaner.

maxyinger: works for me

swissspidy (approved):

PR #9518, opened 2021-10-25
Video Captions: Enable by default

swissspidy:
## Context

It's time to improve the video captions appearance.

## Summary

This PR enables the new video captions appearance by default ahead of the planned 1.14.0 release.

## Relevant Technical Choices

N/A

## To-do

N/A

## User-facing changes

See #9043 for context

## Testing Instructions

- [ ] This is a non-user-facing change and requires no QA

See #9043 for context

This PR can be tested by following these steps:

1. Enable the video captions experiment
1. Create a story with 1 or more videos
1. Upload captions for the video(s)
1. Add some other elements on layers above the video(s)
1. Preview story and verify that captions are displayed correctly and on top of all other elements on the page and within the safe zone

### Does this PR have a security-related impact?

No

### Does this PR change what data or activity we track or use?

No

### Does this PR have a legal-related impact?

No

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [x] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [x] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #9216

PR #10056, opened 2021-12-27
Editor: Enable improved video captions by default

### LINKED ISSUE #9938: Video Captions: Enable new captions in editor by default
## Task Description

#9699 added new video captions in the editor, but behind a feature flag.

We want to enable this by default eventually.

amovar18:
## Context
Since work on VideoCaptions has been completed. it would be the correct time to enable it by default.

## Testing Instructions

1. Open the editor.
2. Add a new video to a story and add captions to it
3. Move it around and maybe add some other elements on top of it so it's only partially visible
4. Play the video in the editor
5. See that the captions are displayed on top of all other elements and always at the same position on the canvas
6. The captions appearance should be very close to what you see when previewing the story on the frontend

## Checklist

---

Fixes #9938

### DIFF
--- a/includes/Experiments.php
+++ b/includes/Experiments.php
@@ -376,6 +376,7 @@ class Experiments extends Service_Base implements HasRequirements {
 				'label'       => __( 'Video Captions', 'web-stories' ),
 				'description' => __( 'Improve video captions appearance in the editor', 'web-stories' ),
 				'group'       => 'editor',
+				'default'     => true,
 			],
 		];
 	}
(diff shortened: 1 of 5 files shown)


swissspidy:
Thanks for the PR!

Please note that we usually first enable features by default by setting `'default' => true` in PHP (see #9518 for an example).

Only in a second PR do we remove all of the code like this.

We're still testing this feature on our own site though, so I'm not sure it's 100% ready to be enabled by default yet.

amovar18:
I will update the PR accordingly

swissspidy:
Awesome, thanks!

This looks good, but I'll gather some feedback internally first to see whether this is ready to be merged.

swissspidy (approved):

spacedmonkey (approved):

PR #10057, opened 2021-12-27
Codequality: Fix console warnings

### LINKED ISSUE #10075: Incorrect `PropType` declarations clutter console during development.
## Bug Description

Incorrect `PropType`  declarations make it difficult to debug or even read the console logs in the browser. These warnings occur due to `PropTypes` being incorrectly defined or sometimes type of props being passed are wrong.

## Expected Behaviour

PropTypes warning should not be logged in the console.

## Additional Context

- WordPress Version: 5.8.2
- Operating System: MacOs monterrey (m1)
- Browser: Chrome

amovar18:
## Context
This PR aims to fix some of the browser and warnings that come up in karma tests.

## Summary

A lot of karma fixtures were providing wrong insufficient input to the components such as `ImageDisplay`, `VideoTrim` etc. This PR fixes them and also fixes some of the warnings that come up in browser during development mode.
Some of the proptype warnings:
Screenshot 2021-12-30 at 1 05 36 PM

Screenshot 2021-12-30 at 1 02 10 PM

Screenshot 2021-12-30 at 1 04 25 PM

Screenshot 2021-12-30 at 1 08 01 PM

## Testing Instructions

## Checklist

---

Fixes #10075 
Partially addresses #8809


amovar18:
I guess this will also partially solve #8809 since it removes most of the propType warnings

swissspidy:
Could you double check whether #9944 already covers some of this?

amovar18:
While checking diff across both PR I found out some things:
1. https://github.com/google/web-stories-wp/blob/dcea48efcbaf1af4fcd0314517508206f1be19d9/packages/design-system/src/components/snackbar/snackbarMessage.js#L274 wouldn't doing this be bad for accessibility as in not providing `aria-label` to snackbar?
2. Some things are covered by that PR and some things are covered by this PR(70% of the warnings are covered there but some left out are covered in this, if it makes it easier you can take the changes from this PR and use it in #9944)

packages/media/src/types.js
```diff
@@ -68,7 +68,7 @@ ResourcePropTypes.videoResource = PropTypes.shape({
   width: PropTypes.number.isRequired,
   height: PropTypes.number.isRequired,
   poster: PropTypes.string,
-  posterId: PropTypes.number,
+  posterId: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
```
swissspidy, line 71: Where does it happen that the `posterId` is a string?

packages/story-editor/src/components/form/dateTime/datePicker.js
```diff
@@ -163,7 +163,7 @@ function DatePicker({ currentDate, onChange, onViewChange }) {
 DatePicker.propTypes = {
   onChange: PropTypes.func.isRequired,
   onViewChange: PropTypes.func,
-  currentDate: PropTypes.string,
+  currentDate: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
```
swissspidy, line 166: Why is the currentDate sometimes a number? 🤔  Sounds like bad fixture data

packages/media/src/types.js
amovar18: I think it’s for 3rd party media since they are provided by coverr. Will update this comment with a screenshot.

packages/design-system/src/components/snackbar/snackbarContainer.js
```diff
@@ -170,6 +171,7 @@ export const SnackbarContainer = ({
               <ChildContainer ref={nodeRef} placement={placement}>
                 <Component
                   {...notificationProps}
+                  aria-label={ariaLabel}
```
swissspidy, line 174: That does not seem right. `aria-label` should already be part of `notificationProps`.

This is the aria label for the individual message, not for the container.

So adding a label like "Snackbar for dashboard" would be incorrect.

packages/dashboard/src/components/interfaceSkeleton/index.js
```diff
@@ -162,6 +162,7 @@ const InterfaceSkeleton = ({ additionalRoutes }) => {
         </PageContent>
       </AppFrame>
       <Snackbar.Container
+        aria-label="Snackbar for dashboard"
```
swissspidy, line 165: As per my other comment, I don't think this is right.

Also, keep in mind to use the available i18n functions.

swissspidy:
> While checking diff across both PR I found out some things:
> 
> 1. https://github.com/google/web-stories-wp/blob/dcea48efcbaf1af4fcd0314517508206f1be19d9/packages/design-system/src/components/snackbar/snackbarMessage.js#L274
>     wouldn't doing this be bad for accessibility as in not providing `aria-label` to snackbar?

`aria-label` is only really needed if the snackbar message itself doesn't suffice already.
If none is provided, the snackbar message is read.

> 2. Some things are covered by that PR and some things are covered by this PR(70% of the warnings are covered there but some left out are covered in this, if it makes it easier you can take the changes from this PR and use it in [Karma: Improve output to reduce noise #9944](https://github.com/google/web-stories-wp/pull/9944))

If you could target your PR against #9944 that would be great!

amovar18:
Will change the base from main to #9944

packages/story-editor/src/components/form/dateTime/datePicker.js
amovar18: @swissspidy the reason is the `value` prop sent down from `DateTime` is actually number (confirmed this be logging typeof date) this is the case if the webstory is either in the following states [`draft`, `pending`, `auto-draft`] or if 
```
(date === modified || date === null);
```
which leads to passing of `displayDate` as value whose value type is number

packages/design-system/src/components/snackbar/snackbarContainer.js
amovar18: Oh my bad but after further research i found out that `showSnackbar` only sends down only these 3 things in object
```
dismissible: true
id: "7e9ce02c-8d17-46c5-9d05-7f9cf81346aa"
message: "Message to be shown in snackbar"
```
so would it be nice if we send `aria-label` for the snackbar something like this `'aria-label': 'Notification for '+message,` or `'aria-label': 'Snackbar for '+message,`

PR #10063, opened 2021-12-27
Checklist: Offer page reload option when toggling Video Optimization setting

### LINKED ISSUE #7473: Video Optimization PPC: Inform user they need to reload after enabling auto-optimization
When auto-optimization is enabled from the PPC, a reload is required for the auto-optimization to start working. We should either provide the user a note saying they must refresh the page, or refresh for them

BrookeGraham:
## Context

If a user has opted out of the video optimization, then in the pre-publish checklist, there is an option to re-enable it. However, once you toggle it 'on' you must save and reload your story before it will apply the change. 

## Summary

We are adding a button to the pre-publish checklist video optimization item that allows the user to save their progress and reload the editor. The reload is necessary for the optimization to take affect. 

## Relevant Technical Choices

Added a `hasOptedIn` state that gets updated to `true` if the user toggles the video optimization on. This is used to determine which `footer` to render in the checklist's card for video optimization. 
When they have opted in,  a `Save & Reload` button will now appear as the footer. 

## User-facing changes

save

## Testing Instructions

1. Go to settings in the editor, and turn off video optimization. 
2. Back in the story editor, add a new video that has not had the optimization. 
3. Open the checklist
4. You should see a checklist item under `Accessibility` that mentions the video optimization.
5. Click the checkbox to re-enable video optimization.
6. You should now see a `Save & Reload` button. 
7. Your progress should now be saved, then the page will reload.  

no

no

no

## Checklist

---

Fixes #7473

### DIFF
--- a/packages/wp-story-editor/src/components/checklist/accessibility/videoOptimizationCheckbox/index.js
+++ b/packages/wp-story-editor/src/components/checklist/accessibility/videoOptimizationCheckbox/index.js
@@ -49,6 +49,11 @@ const ButtonContainer = styled.div`
   justify-content: center;
 `;
 
+const BoldText = styled.span`
+  color: ${({ theme }) => theme.colors.standard.white};
+  font-weight: ${({ theme }) => theme.typography.weight.bold};
+`;
+
 const DescriptionText = styled(Text).attrs({
   size: THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL,
 })`
@@ -101,17 +106,31 @@ function VideoOptimizationCheckbox() {
       )}
       footer={
         hasOptedIn ? (
-          <ButtonContainer>
-            <Button
-              variant={BUTTON_VARIANTS.RECTANGLE}
-              type={BUTTON_TYPES.PRIMARY}
-              size={BUTTON_SIZES.SMALL}
-              onClick={handleSave}
-              disabled={isSaving}
-            >
-              {__('Save and Reload', 'web-stories')}
-            </Button>
-          </ButtonContainer>
+          <>
+            <p>
+              <TranslateWithMarkup
+                mapping={{
+                  b: <BoldText />,
+                }}
+              >
+                {__(
+                  'Automatic video optimization enabled. <b>Reload the page</b> to apply changes.',
+                  'web-stories'
+                )}
+              </TranslateWithMarkup>
+            </p>
+            <ButtonContainer>
+              <Button
+                variant={BUTTON_VARIANTS.RECTANGLE}
+                type={BUTTON_TYPES.PRIMARY}
+                size={BUTTON_SIZES.SMALL}
+                onClick={handleSave}
+                disabled={isSaving}
+              >
+                {__('Save and Reload', 'web-stories')}
+              </Button>
+            </ButtonContainer>
+          </>
         ) : (
           <CheckboxContainer>
             <Checkbox
--- a/packages/wp-story-editor/src/components/checklist/accessibility/videoOptimizationCheckbox/index.js
+++ b/packages/wp-story-editor/src/components/checklist/accessibility/videoOptimizationCheckbox/index.js
@@ -34,6 +34,7 @@ import {
   useConfig,
   useCurrentUser,
   useIsChecklistMounted,
+  useStory,
   ChecklistCard,
 } from '@web-stories-wp/story-editor';
 
@@ -83,6 +84,15 @@ function VideoOptimizationCheckbox() {
     toggleWebStoriesMediaOptimization();
   }, [toggleWebStoriesMediaOptimization]);
 
+  const { isSaving, saveStory } = useStory(({ state, actions }) => ({
+    isSaving: state.meta.isSaving,
+    saveStory: actions.saveStory,
+  }));
+  const handleSave = async () => {
+    await saveStory();
+    window.location.reload(true);
+  };
+
   return (showCheckbox || hasOptedIn) && isChecklistMounted ? (
     <ChecklistCard
       title={__(
@@ -96,8 +106,8 @@ function VideoOptimizationCheckbox() {
               variant={BUTTON_VARIANTS.RECTANGLE}
               type={BUTTON_TYPES.PRIMARY}
               size={BUTTON_SIZES.SMALL}
-              // onClick={handleSaveButton}
-              // disabled={isSaving}
+              onClick={handleSave}
+              disabled={isSaving}
             >
               {__('Save and Reload', 'web-stories')}
             </Button>
--- a/packages/wp-story-editor/src/components/checklist/accessibility/videoOptimizationCheckbox/index.js
+++ b/packages/wp-story-editor/src/components/checklist/accessibility/videoOptimizationCheckbox/index.js
@@ -38,6 +38,7 @@ import {
   useConfig,
   useCurrentUser,
   useIsChecklistMounted,
+  useRegisterCheck,
   useStory,
   ChecklistCard,
   DefaultFooterText,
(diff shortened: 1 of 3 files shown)


maxyinger (approved):
Nice. this looks great!

packages/wp-story-editor/src/components/checklist/accessibility/videoOptimizationCheckbox/index.js
```diff
@@ -37,6 +43,17 @@ const CheckboxContainer = styled.div`
...
 
+const ButtonContainer = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+`;
+
+const BoldText = styled.span`
+  color: ${({ theme }) => theme.colors.standard.white};
+  font-weight: ${({ theme }) => theme.typography.weight.bold};
```
BrittanyIRL: nit, the `Text` component is adaptable to be nested, it's kinda 6 or half dozen, just mentioning it. 

```suggestion
const BoldText = styled(Text).attrs({ as: 'span', isBold: true })`
  color: ${({ theme }) => theme.colors.standard.white};`
```

packages/wp-story-editor/src/components/checklist/accessibility/videoOptimizationCheckbox/index.js
```diff
@@ -66,44 +84,86 @@ function VideoOptimizationCheckbox() {
...
+              >
+                {__(
+                  'Automatic video optimization enabled. <b>Reload the page</b> to apply changes.',
+                  'web-stories'
+                )}
+              </TranslateWithMarkup>
+            </p>
+            <ButtonContainer>
+              <Button
+                variant={BUTTON_VARIANTS.RECTANGLE}
```
BrittanyIRL: rectangle is default, if you don't want to give it that it'll still work

packages/wp-story-editor/src/components/checklist/accessibility/videoOptimizationCheckbox/index.js
```diff
@@ -66,44 +84,86 @@ function VideoOptimizationCheckbox() {
...
-                      trackClick(evt, 'click_video_optimization_settings')
-                    }
-                    href={dashboardSettingsLink}
-                    isBold
-                  />
-                ),
-              }}
+        hasOptedIn ? (
+          <>
+            <p>
```
BrittanyIRL: Let's use the same `DefaultFooterText` (`Text size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.X_SMALL}`) for the paragraph here so that spacing is consistent.

BrittanyIRL (changes requested):
Nice!! One little a11y thing: When I check the box to allow optimization via my keyboard (spacebar), the focus gets lost. It should be brought to the 'save and reload' button.

packages/wp-story-editor/src/components/checklist/accessibility/videoOptimizationCheckbox/index.js
BrookeGraham: Ah sweet, thanks!

BrookeGraham:
> Nice!! One little a11y thing: When I check the box to allow optimization via my keyboard (spacebar), the focus gets lost. It should be brought to the 'save and reload' button.

Great call out! I embarrassingly forgot to check. Fixed.

samwhale (approved):
Nice this is great. Would be super dope if we could add a karma test here 🤔 

Or maybe we just add a unit test that shows that when we click the checkbox it shows the new `Save and Reload` button

kkalarickal:
Verified via [stories-qa](https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/post.php?post=15905&action=edit#page=%2522b66294f7-5080-4354-99b3-6f062df53f23%2522) running #10063 

Status: **Needs minor improvement**   CC: @swissspidy @BrookeGraham 

Reason: As a user, I would expect the Checklist > Accessibility count to reflect the number of "issues" that are reported.  In the case of video optimization is turned off in stories > settings, this means that there should be +1 added to the accessibility section. However, this is not the case as can be seen here

image

Notes:
- confirmed that unchecking video optimizations generates the extra modal in the checklist > accessibility that mentions adding it back
- checking the box brings up the save & reload button
- clicking the save & reload, results in the unsaved changes to the story being saved in draft
- if the user chooses NOT to save & reload, then that notice disappears when  manually reloading

Call out:
1. After checking the box under checklist > accessibility to change the global video optimization setting, if user chooses NOT to save & reload, then that CTA is maintained in the checklist. However, if the story is published or changed back to draft etc. the CTA remains in the checklist (even though the story itself was saved via independent user action).

swissspidy:
> Reason: As a user, I would expect the Checklist > Accessibility count to reflect the number of "issues" that are reported. In the case of video optimization is turned off in stories > settings, this means that there should be +1 added to the accessibility section. However, this is not the case as can be seen here

I _think_ this is by design, because the second card is not actionable yet. After the reload the count will be correct.

That said, this is a separate feature request that warrants its own ticket.

> After checking the box under checklist > accessibility to change the global video optimization setting, if user chooses NOT to save & reload, then that CTA is maintained in the checklist. However, if the story is published or changed back to draft etc. the CTA remains in the checklist (even though the story itself was saved via independent user action).

You mean if the story was just saved by the user, there's no need to call it "Save & Reload", as we can just call it "Reload"?

Sounds like a potential small enhancement on top of this. Maybe easiest in a separate ticket.

BrookeGraham:
> I would expect the Checklist > Accessibility count to reflect the number of "issues" that are reported. In the case of video optimization is turned off in stories > settings, this means that there should be +1 added to the accessibility section.

@swissspidy I don't think that count issue has to do with the reload. It seems the Video Optimization isn't included in the count at all. This was likely sense we are showing this checkbox if any video is apart of the story and the settings has been changed to _not_ optimize. I have fixed it. Updating now. 

Personally, I find it odd to have this Video Optimization checkbox under Accessibility. But that is just me.

BrittanyIRL:
> I don't think that count issue has to do with the reload. It seems the Video Optimization isn't included in the count at all. This was likely sense we are showing this checkbox if any video is apart of the story and the settings has been changed to _not_ optimize. I have fixed it. Updating now.

so the count was intentionally leaving off this checkbox since it's a setting to update not an issue in the story yet.

BrookeGraham:
>so the count was intentionally leaving off this checkbox since it's a setting to update not an issue in the story yet.

Yes as far as I could tell. It was never included in the count. I am assuming the reason for that was because it was a settings issue and not a story issue. Which is why I think it can be misleading to have this check here under `Accessibility`. 

Also, there is another `Video Optimization` card that we have. It is the same as the other cards where we get thumbnails for each page that has the issue. I couldn't figure out a way to add video that would trigger that card. But it is different from the `Video Optimization Checkbox` in that  the `Video Optimization Checkbox` is where we update the overall setting. As opposed to the regular `Video Optimization` where we show the issue as a thumbnail.

BrittanyIRL:
> Yes as far as I could tell. It was never included in the count. I am assuming the reason for that was because it was a settings issue and not a story issue. 

Yup, you're right on this, that's exactly it. 

> Which is why I think it can be misleading to have this check here under `Accessibility`.

There was a bit of back and forth here on this last summer. Ultimately, I think this is a design call. Technically, optimizing video is an accessibility thing because it makes the story more performant and viewable for a variety of connections. 

> Also, there is another Video Optimization card that we have. It is the same as the other cards where we get thumbnails for each page that has the issue. I couldn't figure out a way to add video that would trigger that card. But it is different from the Video Optimization Checkbox in that the Video Optimization Checkbox is where we update the overall setting. As opposed to the regular Video Optimization where we show the issue as a thumbnail.

This is from a design last summer here: https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=8259%3A261122
Screen Shot 2022-01-10 at 4 07 36 PM
so it's fine as is unless i misunderstood.

BrookeGraham:
So should the count be updated or no?

BrittanyIRL:
> So should the count be updated or no?

i say no, let's run that past design and if they want to update it we can do a new ticket.

BrittanyIRL:
looks like one of the js unit tests failing is from this pr: 
Screen Shot 2022-01-12 at 10 48 38 AM

BrookeGraham:
>looks like one of the js unit tests failing is from this pr:

Ah yep. I moved the checked outside the component to match the way we implement the other checks. I'll revert that.

BrittanyIRL (approved):
keyboard focus works great now! thanks!

PR #10068, opened 2021-12-28
Add catch safeguard for `preloadImage`

### LINKED ISSUE #9326: Uncaught promise rejections when using `preloadImage`
## Bug Description
There's a console error displayed when saving a new page template with images.
Let's figure out if there's actually anything broken and add handle the error properly.

## Steps to Reproduce
1. Create a new page with some images
2. Go to the Page Templates tab in the library and add click to save a new template
3. See the console error

## Screenshots
Screenshot 2021-10-07 at 12 31 20
Screenshot 2021-10-07 at 12 30 50

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

amovar18:
## Context

This PR aims to add safeguard for `preLoadImage` to catch uncaught promise rejections.

## Testing Instructions

1.

## Checklist

---

Fixes #9326

### DIFF
--- a/packages/media/src/getImageDimensions.js
+++ b/packages/media/src/getImageDimensions.js
@@ -26,7 +26,7 @@ import preloadImage from './preloadImage';
  * @return {Promise} Image dimensions object.
  */
 const getImageDimensions = async (src) => {
-  const img = await preloadImage(src);
+  const img = await preloadImage(src).catch(() => {});
   return {
     width: img.naturalWidth,
     height: img.naturalHeight,
--- a/packages/media/src/getImageDimensions.js
+++ b/packages/media/src/getImageDimensions.js
@@ -26,12 +26,7 @@ import preloadImage from './preloadImage';
  * @return {Promise} Image dimensions object.
  */
 const getImageDimensions = async (src) => {
-  let img;
-  try {
-    img = await preloadImage(src);
-  } catch (error) {
-    //ignore
-  }
+  const img = await preloadImage(src);
   return {
     width: img.naturalWidth,
     height: img.naturalHeight,
--- a/packages/media/src/getImageDimensions.js
+++ b/packages/media/src/getImageDimensions.js
@@ -26,7 +26,12 @@ import preloadImage from './preloadImage';
  * @return {Promise} Image dimensions object.
  */
 const getImageDimensions = async (src) => {
-  const img = await preloadImage(src).catch(() => {});
+  let img;
+  try {
+    img = await preloadImage(src);
+  } catch (error) {
+    //ignore
+  }
   return {
     width: img.naturalWidth,
     height: img.naturalHeight,
--- a/packages/story-editor/src/elements/media/imageDisplay.js
+++ b/packages/story-editor/src/elements/media/imageDisplay.js
@@ -83,7 +83,12 @@ function ImageDisplay({ element, box, previewMode }) {
     if (resourceList.get(resource.id)?.type !== 'fullsize' && resource.src) {
       timeout = setTimeout(async () => {
         const url = getProxiedUrl(resource, resource.src);
-        const preloadedImg = await preloadImage(url, srcSet).catch(() => {});
+        let preloadedImg;
+        try {
+          preloadedImg = await preloadImage(url, srcSet).catch(() => {});
+        } catch (error) {
+          //ignore
+        }
         if (mounted) {
           resourceList.set(resource.id, {
             type: 'fullsize',
--- a/packages/story-editor/src/elements/media/imageDisplay.js
+++ b/packages/story-editor/src/elements/media/imageDisplay.js
@@ -83,7 +83,7 @@ function ImageDisplay({ element, box, previewMode }) {
     if (resourceList.get(resource.id)?.type !== 'fullsize' && resource.src) {
       timeout = setTimeout(async () => {
         const url = getProxiedUrl(resource, resource.src);
-        const preloadedImg = await preloadImage(url, srcSet);
+        const preloadedImg = await preloadImage(url, srcSet).catch(() => {});
         if (mounted) {
           resourceList.set(resource.id, {
             type: 'fullsize',
--- a/packages/story-editor/src/elements/media/imageDisplay.js
+++ b/packages/story-editor/src/elements/media/imageDisplay.js
@@ -83,9 +83,8 @@ function ImageDisplay({ element, box, previewMode }) {
     if (resourceList.get(resource.id)?.type !== 'fullsize' && resource.src) {
       timeout = setTimeout(async () => {
         const url = getProxiedUrl(resource, resource.src);
-        let preloadedImg;
         try {
-          preloadedImg = await preloadImage(url, srcSet);
+          const preloadedImg = await preloadImage(url, srcSet);
           if (mounted) {
             resourceList.set(resource.id, {
               type: 'fullsize',
--- a/packages/story-editor/src/elements/media/imageDisplay.js
+++ b/packages/story-editor/src/elements/media/imageDisplay.js
@@ -86,15 +86,15 @@ function ImageDisplay({ element, box, previewMode }) {
         let preloadedImg;
         try {
           preloadedImg = await preloadImage(url, srcSet);
-        } catch (error) {
-          //ignore
-        }
-        if (mounted) {
-          resourceList.set(resource.id, {
-            type: 'fullsize',
-          });
-          setSrc(preloadedImg.currentSrc);
-          setSrcType('fullsize');
+          if (mounted) {
+            resourceList.set(resource.id, {
+              type: 'fullsize',
+            });
+            setSrc(preloadedImg.currentSrc);
+            setSrcType('fullsize');
+          }
+        } catch {
+          // Ignore
         }
       });
     }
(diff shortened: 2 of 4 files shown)


packages/media/src/getImageDimensions.js
```diff
@@ -26,7 +26,7 @@ import preloadImage from './preloadImage';
  * @return {Promise} Image dimensions object.
  */
 const getImageDimensions = async (src) => {
-  const img = await preloadImage(src);
+  const img = await preloadImage(src).catch(() => {});
```
miina: `preloadImage` is used with `await` here everywhere, would adding `catch` like this do anything?
Probably should use `try` and `catch` instead.

packages/media/src/getImageDimensions.js
```diff
@@ -26,7 +26,12 @@ import preloadImage from './preloadImage';
...
 const getImageDimensions = async (src) => {
-  const img = await preloadImage(src).catch(() => {});
+  let img;
+  try {
+    img = await preloadImage(src);
+  } catch (error) {
+    //ignore
+  }
   return {
     width: img.naturalWidth,
```
miina: Now this would give an error if the `img` is `null` / `undefined`, no?

packages/media/src/getImageDimensions.js
```diff
@@ -26,7 +26,12 @@ import preloadImage from './preloadImage';
  * @return {Promise} Image dimensions object.
  */
 const getImageDimensions = async (src) => {
-  const img = await preloadImage(src).catch(() => {});
+  let img;
+  try {
+    img = await preloadImage(src);
+  } catch (error) {
+    //ignore
```
miina: Here and elsewhere:
```suggestion
    // Ignore.
```

packages/story-editor/src/elements/media/imageDisplay.js
```diff
@@ -83,7 +83,12 @@ function ImageDisplay({ element, box, previewMode }) {
     if (resourceList.get(resource.id)?.type !== 'fullsize' && resource.src) {
       timeout = setTimeout(async () => {
         const url = getProxiedUrl(resource, resource.src);
-        const preloadedImg = await preloadImage(url, srcSet).catch(() => {});
+        let preloadedImg;
+        try {
+          preloadedImg = await preloadImage(url, srcSet).catch(() => {});
```
miina: This still has a `catch` here.

Also, if `preloadedImage` is `null` / `undefined`, it will still error below with `preloadedImg.currentSrc`.

packages/story-editor/src/utils/getBlurHashFromImage.js
```diff
@@ -37,7 +37,12 @@ const getImageData = (image) => {
...
 
 const getBlurHashFromImage = async (src) => {
-  const image = await preloadImage(src).catch(() => {});
+  let image;
+  try {
+    image = await preloadImage(src);
+  } catch (error) {
+    //ignore
+  }
   const imageData = getImageData(image);
```
miina, line 46: Wouldn't this also throw an error when passing `null` / `undefined` as `image`?

Would be great if you could actually test out if these cases work as expected when you purposefully trigger an error in `preloadImage`.

miina (changes requested):

packages/media/src/getImageDimensions.js
```diff
@@ -26,7 +26,12 @@ import preloadImage from './preloadImage';
...
  */
 const getImageDimensions = async (src) => {
-  const img = await preloadImage(src);
+  let img;
+  try {
+    img = await preloadImage(src);
+  } catch (error) {
+    //ignore
+  }
   return {
```
swissspidy: If running into the `catch` above, `img` will be `undefined` still, and trying to access `img.naturalWidth` or `img.naturalHeight` will break.

Second, no need to do `catch (errror) {` if `error` is unused. Can just do ` catch {`

I'd probably write the whole thing like this:

```js
try {
    const img = await preloadImage(src);
    return {
      width: img.naturalWidth,
      height: img.naturalHeight,
    };
  } catch {
    return { width: 0, height: 0 };
  }
```

But I am skeptical that we actually need the `catch()` here.

Everywhere `getImageDimensions` is being used there already seem to be some `catch` in place.

Adding one here and returning `width:0, height: 0` would unexpectedly break things in other ways.

**So in short:** I don't think we should add a `catch` here at all.

packages/story-editor/src/elements/media/imageDisplay.js
```diff
@@ -83,7 +83,12 @@ function ImageDisplay({ element, box, previewMode }) {
     if (resourceList.get(resource.id)?.type !== 'fullsize' && resource.src) {
       timeout = setTimeout(async () => {
         const url = getProxiedUrl(resource, resource.src);
-        const preloadedImg = await preloadImage(url, srcSet);
+        let preloadedImg;
+        try {
+          preloadedImg = await preloadImage(url, srcSet);
+        } catch (error) {
+          //ignore
```
swissspidy: There should be a `return` here.

If you just `catch`, accessing `preloadedImg.currentSrc` below will cause an error because `preloadedImg` is `undefined`.

Also, can just do `catch {`, without `erorr`.

packages/story-editor/src/utils/getBlurHashFromImage.js
amovar18: @miina this wont cause a problem because it is being used at only one place that it in `useDetectBlurHash` which already wraps `getPreloadImage` in a try catch thinking of it now I think should this even be wrapped in try catch since useDetectBlurHash wraps it as a whole?

packages/story-editor/src/elements/media/imageDisplay.js
amovar18: moved the whole if condition in try catch

packages/media/src/getImageDimensions.js
amovar18: https://github.com/google/web-stories-wp/pull/10068#discussion_r778719305

packages/story-editor/src/utils/getBlurHashFromImage.js
miina: Since this is an util that could in theory be used elsewhere as well, then to be on the safe side, I think it's still better to use `try` here as well, not assuming the util will only ever be used in one place. Since the try-catch around using this util inside `useDetectBlurHash` is used for `getBlurHashFromImage` only (as `saveBlurHash` has its own error handling), might make more sense to remove it from there instead but still keep it in the util.

amovar18: Leaving getBlurhashFromImage with a try/catch

miina: Ok, but let's make sure `getImageData` doesn't trigger error here then as well if `preloadImage` fails.

packages/story-editor/src/utils/getBlurHashFromImage.js
```diff
@@ -37,7 +37,12 @@ const getImageData = (image) => {
 };
 
 const getBlurHashFromImage = async (src) => {
-  const image = await preloadImage(src);
+  let image;
+  try {
+    image = await preloadImage(src);
+  } catch {
+    // Ignore
```
spacedmonkey: We should not ignore this. If the image fails to load here, we should error out or something. 

```suggestion
    return Promise.reject(new Error('image failed to load'));
```

spacedmonkey:
What about this function? 

https://github.com/google/web-stories-wp/blob/636aa1e447226936c058d40d7983c647fe4aead8/packages/media/src/getImageDimensions.js#L28-L34

amovar18:
> What about this function?
> 
> https://github.com/google/web-stories-wp/blob/636aa1e447226936c058d40d7983c647fe4aead8/packages/media/src/getImageDimensions.js#L28-L34
@spacedmonkey 
https://github.com/google/web-stories-wp/pull/10068#discussion_r778719305

packages/story-editor/src/utils/getBlurHashFromImage.js
miina: (Marking it back to unresolved since currently, `getImageData` would still error if we pass `undefined` image there.)

amovar18: It will now return a `rejected` Promise if getPreloadImage fails to load the image

packages/story-editor/src/elements/media/imageDisplay.js
```diff
@@ -83,13 +83,18 @@ function ImageDisplay({ element, box, previewMode }) {
...
-        const preloadedImg = await preloadImage(url, srcSet);
-        if (mounted) {
-          resourceList.set(resource.id, {
-            type: 'fullsize',
-          });
-          setSrc(preloadedImg.currentSrc);
-          setSrcType('fullsize');
+        let preloadedImg;
+        try {
+          preloadedImg = await preloadImage(url, srcSet);
```
swissspidy:
```suggestion
        try {
          const preloadedImg = await preloadImage(url, srcSet);
```

miina (approved):
LGTM but would be good to get an approval from @spacedmonkey / @swissspidy as well as your team has mostly worked on the media part.

swissspidy (approved):

Issue #10064, opened 2021-12-27
Performance: SingleSelectionMoveable slow render time Discovery


maxyinger:
When performing an action like dragging a selected element to a new position, the `SingleSelectionMoveable` component seems to have the largest render time with a render of ~45ms:
![Screen Shot 2021-12-27 at 3.02.53 PM.png](https://images.zenhubusercontent.com/5ea0a3e2f739259b2b0b092a/4c641a7c-f843-4cc1-a829-054abaef127a)

Completion of this ticket involves finding out what's making this component take so long to render and documenting a possible solution

maxyinger:
Seems to be a giant layout recalc causing this component to take so long to render:
![Screen Shot 2021-12-27 at 3.04.55 PM.png](https://images.zenhubusercontent.com/5ea0a3e2f739259b2b0b092a/28fdcfc2-3778-42b5-90fd-9274bf4ef0a4)

The layout recalc thrashing is caused by `useSnapping` getting the bounding box off of `canvasContainer` and `pageContainer` every time it's re-rendered.

Tested a quick poc solution to this which included holding bounding box data inside the `CanvasProvider` and using a resize observer to update the bounding box instead of `getBoundingClientRect`. unlike `getBoundingClientRect` , `resizeObserver` shouldn't cause a style recalc and should avoid this layout thrashing.

The quick POC seems fruitful from performing the same interaction as before. Got `SingleSelectionMoveable` render time down to ~1.5ms from ~45ms using this approach:

**before**
![Screen Shot 2021-12-27 at 3.02.53 PM.png](https://images.zenhubusercontent.com/5ea0a3e2f739259b2b0b092a/e4a089fb-a6c5-4aa2-904f-b0cfcb466325)

**after**
![Screen Shot 2021-12-27 at 3.15.59 PM.png](https://images.zenhubusercontent.com/5ea0a3e2f739259b2b0b092a/ca4de7a7-e381-4977-9a28-20d5c1878ac6)


Here's the follow up ticket to figure out an elegant approach using the method described above:
https://app.zenhub.com/workspaces/web-stories-5e94d3aced449034e1e9b226/issues/google/web-stories-wp/10065

PR #10069, opened 2021-12-28
Performance: SingleSelectionMoveable Improve Render Speed

### LINKED ISSUE #10065: Performance: update bounding box usage with useCanvas to obtain from resizeObserver
Reference discovery ticket and POC findings here:
https://github.com/google/web-stories-wp/issues/10064#issuecomment-1001765281

This ticket is to update `useCanvas` to obtain bounding boxes for `canvasContainer` & `pageContainer` through a resize observer instead of using `getBoundingClientRect` on them. This should avoid some key layout re-calcs in this action that are causing layout thrashing and heavily effecting the performance of `SingleSelectionMoveable`

maxyinger:
## Context
Part of looking into `SingleSelectionMoveable` which had the single largest render cost that contributed to ~40ms of the largest commit in updating an elements position.

Recorded interaction was to drag an already selected element to update its position.

| Test Page (50 elements) | Render Cost |
| ---- | ---- |
| 147141603-77df3d34-db93-4ac2-bea7-bcc38644f492 | Screen Shot 2021-12-27 at 3 02 53 PM |

Look up findings and summary more in depth here:
https://github.com/google/web-stories-wp/issues/10064#issuecomment-1001765281

## Summary
Fixes render performance of `SingleSelectionMoveable` by updating `useSnapping` & `CanvasProvider`.

| Before | After |
| ---- | ---- |
| ~40ms | ~1ms |
| Screen Shot 2021-12-28 at 12 41 13 PM |  Screen Shot 2021-12-28 at 12 34 42 PM |

## Relevant Technical Choices
Uses a centralized resize observer to update bounding boxes instead of `getBoundingClientRect()` in the hook which attributed to the large style recalc causing `SingleSelectionMoveable` to be such a costly render.

Considered using our `useResizeEffect` but it didn't quite fit for what we needed here:
- useResizeEffect is set up to track a single node, this is extensible to multiple nodes
- useResizeEffect is only used for height and width, this is used for the whole content rect
- useResizeEffect's api is set up to conform to to the effect api, this uses the callback ref pattern @choumx linked in web-stories slack channel

In general, I thought it best to use a custom solution with `ResizeObserver` instead of try to augment our api for `useResizeEffect` to fit this use case. This approach may be useful in other areas, and we can always pull it out and abstract on it, but want to wait till more use cases present themselves instead of pre-emptively doing that

## To-do
If everyone likes the approach with the centralized resize observer, I think there might be a lot of nodes held in state inside `CanvasProvider` that are just held to get their bounding box later. This could replace all those use cases. 

Will create a separate ticket tho if we decide to go with that refactor since it's a separate issue from render perf of SingleSelectionMoveable.

## User-facing changes
NA

## Testing Instructions

1. See that snapping while moving an element is not effected by the update

## Checklist

---

Fixes #10065


packages/story-editor/src/components/canvas/frameElement.js
```diff
@@ -79,42 +79,31 @@ const EmptyFrame = styled.div`
...
-    (state) => ({
-      selectedElementIds: state.state.selectedElementIds,
-      currentPage: state.state.currentPage,
-    })
+  const setNodeForElement = useCanvas((v) => v.actions.setNodeForElement);
+  const handleSelectElement = useCanvas((v) => v.actions.handleSelectElement);
+  const isEditing = useCanvas((v) => v.state.isEditing);
+  const isSelected = useStory((v) => v.state.selectedElementIds.includes(id));
+  const isSingleElement = useStory(
+    (v) => v.state.selectedElementIds.length === 1
```
samwhale: Nice reducing this to one line

samwhale (approved):
Nice dude ✨ Thanks for jumping on a call with me about this

packages/story-editor/src/app/canvas/useCanvasBoundingBox.js
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Returns the bounding box associated with the relative CANVAS_BOUNDING_BOX_IDS
+ *
+ * @param {string} boundingBoxId a value from CANVAS_BOUNDING_BOX_IDS
+ * @return {Object} bounding box associated with `boundingBoxId`
+ */
+export function useCanvasBoundingBox(boundingBoxId) {
+  return useContextSelector(
+    Context,
+    (v) => v.state.boundingBoxes[boundingBoxId]
```
BrittanyIRL: looks like this is throwing off some unit tests

https://github.com/google/web-stories-wp/runs/4652759966?check_suite_focus=true#step:6:6874
Screen Shot 2022-01-03 at 12 57 51 PM

packages/story-editor/src/components/canvas/frameElement.js
```diff
@@ -79,42 +79,31 @@ const EmptyFrame = styled.div`
...
+  const setEditingElement = useCanvas((v) => v.actions.setEditingElement);
   const { id, type, flip } = element;
   const { Frame, isMaskable, Controls } = getDefinitionForType(type);
   const elementRef = useRef();
   const [hovering, setHovering] = useState(false);
-  const {
-    state: { isAnythingTransforming },
-  } = useTransform();
+  const isAnythingTransforming = useTransform(
+    (v) => v.state.isAnythingTransforming
```
BrittanyIRL: i know in another PR there's a discussion about `v` vs `value` - whatever is decided there, this should be consistent in updates.

packages/story-editor/src/components/canvas/selection.js
```diff
@@ -27,8 +27,12 @@ import SingleSelectionMoveable from './singleSelectionMoveable';
 import MultiSelectionMoveable from './multiSelectionMoveable';
 
 function Selection(props) {
-  const { currentPage, selectedElements, isAnimating } = useStory((state) => ({
-    currentPage: state.state.currentPage,
+  const isBackground = useStory((v) => {
+    return (
+      v.state.selectedElements[0]?.id === v.state.currentPage.elements?.[0]?.id
+    );
+  });
```
BrittanyIRL: nit
```suggestion
  const isBackground = useStory(({ state }) =>  state.selectedElements[0]?.id === state.currentPage.elements?.[0]?.id);
```

BrittanyIRL (commented):
Awesome find, think the write up makes sense too. 

Only thing I see is that a few unit tests are now failing because of this - looks like they just need their structure updates so they're aware of the updates. I'll approve once those are fixed.

packages/story-editor/src/components/canvas/utils/useSnapping.js
```diff
@@ -71,12 +77,12 @@ function useSnapping({
...
   );
 
-  if (!canvasContainer || !pageContainer) {
+  if (!canvasContainerBB || !pageContainerBB) {
     return {};
   }
 
-  const canvasRect = canvasContainer.getBoundingClientRect();
-  const pageRect = pageContainer.getBoundingClientRect();
+  const canvasRect = canvasContainerBB;
```
miina: Maybe just use `canvasRect` above instead of `canvasContainerBB` and `pageRect` instead of `pageContainerBB`?

packages/story-editor/src/components/canvas/utils/useSnapping.js
```diff
@@ -24,21 +24,27 @@ import { FULLBLEED_RATIO } from '@web-stories-wp/units';
...
-      canvasContainer,
-      designSpaceGuideline,
-    })
+  const canvasContainerBB = useCanvasBoundingBox(
+    CANVAS_BOUNDING_BOX_IDS.CANVAS_CONTAINER
+  );
+  const pageContainerBB = useCanvasBoundingBox(
+    CANVAS_BOUNDING_BOX_IDS.PAGE_CONTAINER
   );
+  const designSpaceGuideline = useCanvas((v) => v.state.designSpaceGuideline);
```
miina:
```suggestion
  const designSpaceGuideline = useCanvas(({ state }) => state.designSpaceGuideline);
```

packages/story-editor/src/components/canvas/singleSelectionMoveable/index.js
```diff
@@ -53,30 +53,20 @@ function SingleSelectionMoveable({
...
 
   const actionsEnabled = !selectedElement.isBackground;
 
   const latestEvent = useRef();
 
-  const { backgroundElement } = useStory(({ state: { currentPage } }) => ({
-    backgroundElement: currentPage.elements[0] ?? {},
-  }));
+  const backgroundElementId = useStory(
+    (v) => v.state.currentPage?.elements[0]?.id
```
miina: Hm, why replace using `useStory(({ state }))` with `useStory((v))` (here and elsewhere)?

packages/story-editor/src/components/canvas/singleSelectionMoveable/index.js
```diff
@@ -53,30 +53,20 @@ function SingleSelectionMoveable({
...
-      scrollLeft,
-      scrollTop,
-    })
-  );
+  const nodesById = useCanvas((v) => v.state.nodesById);
+  const getBox = useUnits((v) => v.actions.getBox);
+  const pushTransform = useTransform((v) => v.actions.pushTransform);
+  const zoomSetting = useLayout((v) => v.state.zoomSetting);
+  const scrollLeft = useLayout((v) => v.state.scrollLeft);
+  const scrollTop = useLayout((v) => v.state.scrollTop);
```
miina: Maybe leave a comment to explain why it's better that `zoomSetting`, `scrollLeft`, `scrollTop` are all using `useLayout` separately instead of just using `useLayout(({ state })` for all.

packages/story-editor/src/components/canvas/frameElement.js
```diff
@@ -79,42 +79,31 @@ const EmptyFrame = styled.div`
...
-    })
-  );
-  const { selectedElementIds, currentPage, isAnimating } = useStory(
-    (state) => ({
-      selectedElementIds: state.state.selectedElementIds,
-      currentPage: state.state.currentPage,
-    })
+  const setNodeForElement = useCanvas((v) => v.actions.setNodeForElement);
+  const handleSelectElement = useCanvas((v) => v.actions.handleSelectElement);
+  const isEditing = useCanvas((v) => v.state.isEditing);
```
miina: Here, too, perhaps leave a comment to explain why these are using `useCanvas` separately here to ensure someone in the future won't just move these back together.

packages/story-editor/src/components/canvas/canvasLayout.js
```diff
@@ -18,14 +18,20 @@
...
-import { memo, useRef, useCallback } from '@web-stories-wp/react';
+import { memo, useRef } from '@web-stories-wp/react';
 import { __ } from '@web-stories-wp/i18n';
 import PropTypes from 'prop-types';
 
 /**
  * Internal dependencies
  */
-import { useCanvas, useLayout } from '../../app';
+import { useComposeRefs } from '@web-stories-wp/design-system';
```
miina: Not related to this PR perhaps but wondering what's the difference between `useCombinedRefs` and `useComposeRefs`? First time seeing `useComposeRefs` in the `story-editor` package, so started wondering. Thoughts?

packages/story-editor/src/app/canvas/useCanvasBoundingBox.js
```diff
@@ -0,0 +1,70 @@
...
+import { useCallback, useRef, useContextSelector } from '@web-stories-wp/react';
+/**
+ * Internal dependencies
+ */
+import Context from './context';
+import { RESIZE_OBSERVATION_KEY } from './constants';
+
+/**
+ * Returns a ref that when applied to an element, stores the elements
+ * bounding box info. can retieve the relevant bounding box info like so:
```
miina:
```suggestion
 * bounding box info. Can retrieve the relevant bounding box info like so:
```

packages/story-editor/src/app/canvas/canvasProvider.js
```diff
@@ -56,6 +58,23 @@ function CanvasProvider({ children }) {
   const [pageCanvasData, setPageCanvasData] = useState(null);
   const [pageCanvasPromise, setPageCanvasPromise] = useState(null);
 
+  const resizeObserver = useMemo(
```
miina: Might be useful to add some comments for this as well, looks like it's nicely described in the description of the PR why `useResizeEffect` wouldn't work here, maybe add a line here, too for clarity?

packages/story-editor/src/components/canvas/singleSelectionMoveable/index.js
swissspidy: Same question asked here: https://github.com/google/web-stories-wp/pull/10051/files#r774172314

I don't know whether this makes any difference at all and would like some proof.

packages/story-editor/src/components/canvas/frameElement.js
maxyinger: ah yup. will update to `value` . have done it on all the other ones, just forgot to update it on this PR

packages/story-editor/src/components/canvas/singleSelectionMoveable/index.js
maxyinger: Yup. put an answer there too. Will post something in the web-stories slack channel about the update too cus I've been getting a lot of questions about it:
https://github.com/google/web-stories-wp/pull/10051/files#r778230330

packages/story-editor/src/components/canvas/frameElement.js
maxyinger: I'll post in the slack channel going over this. I don't think we need specific comments everywhere we're doing this, but maybe a group sync and like a little performance README. What are your thoughts?

packages/story-editor/src/components/canvas/canvasLayout.js
maxyinger: they look like duplicates. Thanks for calling out. I'll create a little ticket to consolidate them

maxyinger: Here's that ticket:

#10093

packages/story-editor/src/app/canvas/canvasProvider.js
maxyinger: Updated this to use an IntersectionObserver so we can get the `clientRect` here which I learned is a different set of coordinates from the `contentRect` you have access too in a resizeObserver. General approach is still the same tho.

miina:
@littlemilkstudio Looks like the JS unit tests are still failing.

Would be great to be able to run the Karma tests here as well before merging. Currently it looks like most of the Karma tests that run, fail with this error:
https://github.com/google/web-stories-wp/runs/4718249940?check_suite_focus=true#step:6:5719

Usually (with the current Karma tests issues) the tests still run until it disconnects and generally fails but looks like in this PR something else might be going on here as well. Thoughts?

maxyinger:
> Usually (with the current Karma tests issues) the tests still run until it disconnects and generally fails but looks like in this PR something else might be going on here as well. Thoughts?

@miina yea it was from doing this:
```
useEffect(() => observer.disconnect, []);
```
vs this:
```
useEffect(() => () => observer.disconnect(), []);
```
cus of the top one not maintaining context

Thanks for pointing out!

miina (approved):

PR #10073, opened 2021-12-29
Mock blur hash in story editor karma tests

### LINKED ISSUE #10052: Story Editor Karma Timeout Failure
## Bug Description

All of the Karma tests are failing due to a timeout issue! 

This started last week. I believe the issue stems from commit `c6798df96` to main - PR #9952. I've documented my findings here: https://github.com/google/web-stories-wp/pull/10045 as @BrookeGraham  and I attempted to debug while most of the team's out on holiday PTO. At this point, it needs fresh perspective. 

## Expected Behaviour

Karma tests should run and ideally pass, but at least get them running again :) 

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

samwhale:
## Context

Karma tests are failing due to long timeouts. @spacedmonkey identified #9953 as the possible culprit. This is an attempt to mock the blur hash functionality.

## Summary

Mocks the `useDetectBlurHash` hook. This hook now returns a function that is a `no-op`.

Mocking this seems to make the tests pass.

## Relevant Technical Choices

A naive first approach that may need some more work. Not sure if we need to be implementing new blur hash behavior in the karma tests as well.

## To-do

n/a

no

no

no

## Checklist

---

Fixes #10052

### DIFF
--- a/packages/story-editor/src/karma/fixture/fixture.js
+++ b/packages/story-editor/src/karma/fixture/fixture.js
@@ -42,6 +42,7 @@ import { createPage } from '../../elements';
 import { TEXT_ELEMENT_DEFAULT_FONT } from '../../app/font/defaultFonts';
 import formattedTemplatesArray from '../../dataUtils/formattedTemplatesArray';
 import { PRESET_TYPES } from '../../constants';
+import * as BlurHash from '../../app/media/utils/useDetectBlurhash.js';
 import getMediaResponse from './db/getMediaResponse';
 import { Editor as EditorContainer } from './containers';
 import taxonomiesResponse from './db/getTaxonomiesResponse';
@@ -182,6 +183,11 @@ export class Fixture {
       }
       return origCreateElement(type, props, ...children);
     });
+
+    //eslint-disable-next-line jasmine/no-unsafe-spy
+    spyOn(BlurHash, 'default').and.callFake(() => ({
+      updateBlurHash: Promise.resolve,
+    }));
 
     this.apiProviderFixture_ = new APIProviderFixture({ mocks });
     this.stubComponent(APIProvider).callFake(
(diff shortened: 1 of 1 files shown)


packages/story-editor/src/karma/fixture/fixture.js
```diff
@@ -183,6 +184,11 @@ export class Fixture {
       return origCreateElement(type, props, ...children);
     });
 
+    //eslint-disable-next-line jasmine/no-unsafe-spy
+    spyOn(BlurHash, 'default').and.callFake(() => ({
```
BrittanyIRL, line 188: Good lookin out, everything is passing now! I got too deep in on the wrong path and thought the wrong thing was timing out.

BrittanyIRL (approved):
Sure this might need some ironing out, i'm honestly not super sure what the best protocol here is for mocking blur hash - but if this can unblock our pipeline then I'm all for iterating from here.

BrookeGraham (approved):
🕵🏻 Amazing!

maxyinger (approved):
Awesome! very nice

spacedmonkey (approved):

swissspidy:
Have you seen #10016?
If the blurhash is added everywhere for fixtures this function shouldn‘t even be called during tests in the first place.

miina:
@samwhale Just want to mention that even though the tests pass here, there is still the timeout issue present and not all tests run, see https://github.com/google/web-stories-wp/runs/4659065178?check_suite_focus=true#step:6:3121

Perhaps you saw this already but since it's currently marked as a fix for #10052 but still has the issue then bringing up anyway.

samwhale:
> Have you seen #10016? If the blurhash is added everywhere for fixtures this function shouldn‘t even be called during tests in the first place.

Ah nice! If that fix is added then that would make this karma update useless. If you prefer that then I think we should close this PR @swissspidy 

---

> @samwhale Just want to mention that even though the tests pass here, there is still the timeout issue present and not all tests run, see https://github.com/google/web-stories-wp/runs/4659065178?check_suite_focus=true#step:6:3121
> 
> Perhaps you saw this already but since it's currently marked as a fix for #10052 but still has the issue then bringing up anyway.

☹️ Aw good catch, I didn't realize. Maybe this fix doesn't work at all then.

PR #10074, opened 2021-12-29
Design Panel: Add contrast warning

### LINKED ISSUE #7230: Editor: Add a contrast check to the "style" panel
## Feature Description

If a user were to pick a style that would make the text hard to read, announce `This color combination may be hard for people to read`.

The `useLiveRegion` hook can be used to announce the message to the user. However, we will need to implement some sort of contrast check in the panel.

## Alternatives Considered

n/a

## Additional Context

Split off from #6059 due to scope creep

---

BrookeGraham:
## Context

We want to notify users when they choose a text color and background color combination that doesn't meet accessibility. 

## Summary

When a user chooses a font color or background color that doesn't have sufficient contrast, we want to show some warning text where they pick the colors. Now inside the Design Panel -> Text we are showing a warning message. We also are making use of our `useLiveRegion` hook to announce the message as it appears. 
I also added the warning to the `background` panel. 

## Relevant Technical Choices

Created a `Warning` component that can be reused with custom messages as we find other areas where we want to show feedback sooner than in the checklist. 
Added a constant for the warning messages under `ACCESSIBILITY_COPY`
Moved `getFailingPages` into the `checklist/checks/pageBackgroundLowTextContrast/check.js` file so that we can use it in both in the checklist check, and here in the panel. Also, renamed it to `getPagesWithFailedContrast`
\`pageBackgroundTextLowContrast(page)` (which is called inside `getPagesWithFailedContrast` ) now returns the offending elements' id. This way we know exactly which elements should be flagged.

The reason I am checking the whole page for contrast issues is because we could have more than one text selection. If either of them doesn't have enough contrast, we should show the message. Plus, we already need to grab all elements from the page to determine which ones are considered as part of the background. 

## To-do
Some background images may flag for low contrast. I am not sure how we want to accommodate for this. 

## User-facing changes
Screen Shot 2021-12-31 at 12 50 11 PM
Screen Shot 2021-12-31 at 12 50 22 PM
Screen Shot 2021-12-31 at 12 50 42 PM
Screen Shot 2021-12-31 at 12 53 21 PM

## Testing Instructions

1. In editor add some text. 
2. Change the font color to match the background. 
3. See the warning appear. 
4. Change it back and the warning should disappear. 
5. Check that the new message is announced. 
6. Turn on voice over, or other reader. 
7. Follow above steps, you should hear the message announced. 
8. In addition, the warning should appear on the background panel. 
9. Click the background on the canvas. The background panel should open. 
10. Change the color to low contrast and you should see the warning message. 
11. The check should still work as normal in the Pre-publish Checklist as well. 

no

no

no

## Checklist

---

Fixes #7230


packages/story-editor/src/components/panels/design/pageBackground/pageBackground.js
```diff
@@ -181,6 +204,7 @@ function PageBackgroundPanel({ selectedElements, pushUpdate }) {
           />
         </Row>
       )}
+      {failedContrast && <Warning message={COLOR_COMBINATION} />}
```
BrookeGraham: Added the warning message here as well cause I figured why not 🤷🏻‍♀️

packages/story-editor/src/components/panels/design/warning/constants.js
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * External dependencies
+ */
+import { __ } from '@web-stories-wp/i18n';
+
+export const COLOR_COMBINATION = __(
+  'This color combination or font size may be hard for people to read.',
```
BrookeGraham: I added the bit on `font size` because that can affect whether or not the element passes accessibility.

packages/design-system/src/icons/warning_low.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="32" height="32" viewBox="0 0 32 32" fill="none" xmlns="http://www.w3.org/2000/svg">
```
BrittanyIRL: icon exists in design system already as `ExclamationTriangle`
Screen Shot 2022-01-03 at 10 53 48 AM

You can run `npm run storybook` and go to design system -> icons to see what we've got.

packages/story-editor/src/components/checklist/checks/pageBackgroundLowTextContrast/check.js
```diff
@@ -344,6 +344,42 @@ function getBackgroundColorByType(element) {
       return () => undefined;
   }
 }
+/**
+ * Returns all failing pages as well as the offending elements' id.
```
BrittanyIRL: nit: 
```suggestion
 * Returns all failing pages as well as the offending elements' id(s).
```

packages/story-editor/src/components/checklist/checks/pageBackgroundLowTextContrast/component.js
```diff
@@ -34,7 +34,7 @@ import { getVisibleThumbnails, ThumbnailPagePreview } from '../../utils';
 import { ACCESSIBILITY_COPY } from '../../constants';
 import { useRegisterCheck } from '../../countContext';
 import { useIsChecklistMounted } from '../../popupMountedContext';
-import { pageBackgroundTextLowContrast } from './check';
+import { getFailingPages } from './check';
```
BrittanyIRL: nit: could this be renamed as like `getPagesWithFailedContrast` or something that's clearer since it's now getting used outside of `/pageBackgroundLowTextContrast` which previously provided more context.

packages/story-editor/src/components/panels/design/pageBackground/pageBackground.js
```diff
@@ -35,11 +35,14 @@ import {
...
 import { Color, Row as DefaultRow } from '../../../form';
-import { useStory } from '../../../../app';
+import { useStory, useLayout } from '../../../../app';
+import { getFailingPages } from '../../../checklist/checks/pageBackgroundLowTextContrast/check';
 import { SimplePanel } from '../../panel';
 import { FlipControls } from '../../shared';
 import { getDefinitionForType } from '../../../../elements';
 import { states, styles, useHighlights } from '../../../../app/highlights';
+import Warning from '../warning/warning';
+import { COLOR_COMBINATION } from '../warning/constants';
```
BrittanyIRL: thoughts on `import { default as Warning, COLOR_COMBINATION } from '../warning';` ?

packages/story-editor/src/components/panels/design/pageBackground/pageBackground.js
```diff
@@ -114,6 +128,15 @@ function PageBackgroundPanel({ selectedElements, pushUpdate }) {
     })
   );
 
+  useEffect(() => {
+    getFailingPages(pages, pageSize, currentPage).then((failedPages) => {
+      // getFailingPages returns an array of pages, since we only care
+      // about currentPage, we can grab the single page result.
+      const result = failedPages[0]?.result;
```
BrittanyIRL: currentPage will always be index 0?

packages/story-editor/src/components/panels/design/pageBackground/pageBackground.js
```diff
@@ -114,6 +128,15 @@ function PageBackgroundPanel({ selectedElements, pushUpdate }) {
...
+      setFailedContrast(result?.length ? true : false);
```
BrittanyIRL: This should work the same, i believe. 

```suggestion
      setFailedContrast(Boolean(result));
```

packages/story-editor/src/components/panels/design/textStyle/karma/contrast.karma.js
```diff
@@ -0,0 +1,65 @@
...
+    beforeEach(async () => {
+      fixture = new Fixture();
+      await fixture.render();
+      await fixture.collapseHelpCenter();
+    });
+
+    afterEach(() => {
+      fixture.restore();
+    });
+
```
BrittanyIRL, line 38: thanks for writing more tests for this! 

Would you mind adding a second check for any aXe violations on the warning panel here? 
It'll look something like: 

````
const contrastWarning = fixture.editor.inspector.designPanel.queryByTitle('Low Warning');
await expectAsync(contrastWarning).toHaveNoViolations();
```

packages/story-editor/src/components/panels/design/textStyle/textStyle.js
```diff
@@ -28,9 +28,13 @@ import PropTypes from 'prop-types';
  */
 import getUpdatedSizeAndPosition from '../../../../utils/getUpdatedSizeAndPosition';
 import { styles, useHighlights, states } from '../../../../app/highlights';
+import { useStory, useLayout } from '../../../../app';
+import { getFailingPages } from '../../../checklist/checks/pageBackgroundLowTextContrast/check';
```
BrittanyIRL: You added `getFailingPages` to exports on `pageBackgroundLowTextContrast` so you could probably import from there too.
```suggestion
import { getFailingPages } from '../../../checklist/checks/pageBackgroundLowTextContrast';
```

packages/story-editor/src/components/panels/design/textStyle/textStyle.js
```diff
@@ -28,9 +28,13 @@ import PropTypes from 'prop-types';
...
 import { usePresubmitHandler } from '../../../form';
 import PanelContent from '../../panel/shared/content';
 import Panel from '../../panel/panel';
+import Warning from '../warning/warning';
+import { COLOR_COMBINATION } from '../warning/constants';
```
BrittanyIRL: Samesies here as earlier: 
`import { default as Warning, COLOR_COMBINATION } from '../warning';`

packages/story-editor/src/components/panels/design/warning/constants.js
BrittanyIRL: 🤔 I'm wondering if this new set of constants belongs closer to the contrast check that it's giving readable user context too - we could put it in `checklist/constants` under `ACCESSIBILITY_COPY.lowContrast` as `panel` or something else that indicates that it's getting used as a corresponding notification of the check in a panel.

packages/story-editor/src/components/panels/design/warning/warning.js
```diff
@@ -0,0 +1,66 @@
...
+const WarningContainer = styled.div`
+  display: flex;
+  gap: 8px;
+  width: 100%;
+  margin: 14px auto;
+  border: 1px solid ${({ theme }) => theme.colors.border.defaultHover};
+  border-radius: ${({ theme }) => theme.borders.radius.small};
+  padding: 8px;
+  color: ${({ theme }) => theme.colors.fg.primary};
+  font-weight: ${({ theme }) => theme.typography.weight.regular};
```
BrittanyIRL: Since you've got `Message` as `Text` these shouldn't be necessary

packages/story-editor/src/components/panels/design/warning/warning.js
```diff
@@ -0,0 +1,66 @@
...
+`;
+const WarningIcon = styled(Icons.WarningLow)`
+  width: 32px;
+  height: 100%;
+  color: #f4b844;
```
BrittanyIRL: If this is a new color, can you add it to the design system? Or if it's similar to another color we've got in there, let's use that instead to keep the color palette tight(ish).

packages/story-editor/src/components/panels/design/warning/warning.js
```diff
@@ -0,0 +1,66 @@
...
+  width: 100%;
+  margin: 14px auto;
+  border: 1px solid ${({ theme }) => theme.colors.border.defaultHover};
+  border-radius: ${({ theme }) => theme.borders.radius.small};
+  padding: 8px;
+  color: ${({ theme }) => theme.colors.fg.primary};
+  font-weight: ${({ theme }) => theme.typography.weight.regular};
+`;
+const WarningIcon = styled(Icons.WarningLow)`
+  width: 32px;
```
BrittanyIRL, line 44: if you swap to the `ExclamationTriangle` default width is 32px which should work the same.

packages/story-editor/src/components/panels/design/warning/warning.js
```diff
@@ -0,0 +1,66 @@
...
+function Warning({ message }) {
+  const speak = useLiveRegion('assertive');
+
+  useEffect(() => {
+    speak(message);
+  }, [message, speak]);
+
+  return (
+    <WarningContainer>
+      <WarningIcon title={__('Low Warning', 'web-stories')} />
```
BrittanyIRL: 🤔  should this be stored with the copy that is `message` for a single source of truth on copy?

BrittanyIRL (changes requested):
Stoked you got to this last week! Love the added notes and how you extended the existing functionality. Couple small things.

packages/design-system/src/icons/warning_low.svg
BrookeGraham: Ah nice. Thanks.

packages/story-editor/src/components/panels/design/pageBackground/pageBackground.js
BrookeGraham: Yea. If we send in the `currentPage`, it will only check that single page and thus the returned array will have a max length of one. If the `currentPage`  passes, it returns an empty array.

packages/story-editor/src/components/panels/design/warning/warning.js
BrookeGraham: I couldn't find any thing of the sort, happy to add.

packages/story-editor/src/components/panels/design/warning/warning.js
BrookeGraham: Good looking out. I may have added this before I was using `Text` I'll double check it isn't needed.

packages/story-editor/src/components/checklist/checks/pageBackgroundLowTextContrast/check.js
```diff
@@ -344,6 +344,42 @@ function getBackgroundColorByType(element) {
...
+ * elements' id(s).
+ */
+export async function getFailingPages(storyPages, pageSize, currentPage) {
+  const promises = [];
+  (storyPages || []).forEach((page) => {
+    // If we only care about checking currentPage bail early
+    // In this case, getFailingPages will return a max of 1 page.
+    if (currentPage && currentPage.id !== page.id) {
+      return;
+    }
```
samwhale: 💭 I think rather than updating to add a new argument, we could pass in the current page in its own array using the same function as before:

```js
const failedPages = getFailingPages([currentPage], pageSize);
const didCurrentPageFail = failedPages[0];
```

This reduces the surface area we'd need to test and keeps the function a little more simple. Thoughts?

packages/story-editor/src/components/panels/design/pageBackground/pageBackground.js
```diff
@@ -114,6 +128,15 @@ function PageBackgroundPanel({ selectedElements, pushUpdate }) {
...
+    });
+  }, [currentPage, pageSize, pages]);
```
samwhale: 🤔 I think we don't need this effect to run when `pages` changes. However, if we use the `getFailingPages` function like this:

```getFailingPages([currentPage], pageSize)```

then this effect won't be run when `pages` changes

File: packages/story-editor/src/components/panels/design/pageBackground/pageBackground.js (hunk 1 above)
Comment (samwhale) [reply packages/story-editor/src/components/panels/design/pageBackground/pageBackground.js]: I think it makes sense to put here. Maybe we should slightly alter the copy to make it specific to background color vs text color. May be a good question for @agingoldseco though:

```
Background: Some text on the page may be hard for people to read against this background. Update the background color to increase the contrast.

Text: This text may be hard for people to read. Update the color or the font size to increase the contrast.
```

File: packages/story-editor/src/components/panels/design/textStyle/karma/contrast.karma.js [hunk 18]
```diff
@@ -0,0 +1,65 @@
...
+    it('should show contrast warning', async () => {
```
samwhale: very nice  👍

packages/story-editor/src/components/panels/design/textStyle/textStyle.js
```diff
@@ -65,10 +69,38 @@ function StylePanel(props) {
...
+
+    getFailingPages(pages, pageSize, currentPage).then((failedPages) => {
+      // getFailingPages returns an array of pages, since we only care
+      // about currentPage, we can grab the single page result.
+      const result = failedPages[0]?.result;
+      setFailedContrast(
+        result?.some((elementId) => selectedElementIds.includes(elementId))
+      );
+    });
+  }, [currentPage, pageSize, pages, selectedElements]);
```
samwhale: 💭 It would be ideal if this effect only ran when `currentPage` changed (like before).

I think we could derive whether the selected text element has a contrast warning rather than setting it here.

```js
  useEffect(() => {
    const selectedElementIds = selectedElements.map(
      (selectedElement) => selectedElement.id
    );

    getFailingPages(pages, pageSize, currentPage).then((failedPages) => {
      // getFailingPages returns an array of pages, since we only care
      // about currentPage, we can grab the single page result.
      const result = failedPages[0]?.result;
      setFailedPages(result); // set failed ids in state
    });
  }, [currentPage, pageSize, pages, selectedElements]);
...

const showContrastWarning = failedPages?.some((elementId) => selectedElementIds.includes(elementId));

return (
    ...
    {showContrastWarning && <Warning message={COLOR_COMBINATION} />}
    ...
)
```

packages/story-editor/src/components/panels/design/warning/warning.js
```diff
@@ -0,0 +1,66 @@
...
+  height: 100%;
+  color: #f4b844;
+`;
+
+const Message = styled(Text)`
+  width: 100%;
+`;
+
+function Warning({ message }) {
+  const speak = useLiveRegion('assertive');
```
samwhale, line 56: nice!

packages/story-editor/src/components/panels/design/warning/warning.js
```diff
@@ -0,0 +1,66 @@
...
+`;
+const WarningIcon = styled(Icons.WarningLow)`
+  width: 32px;
+  height: 100%;
+  color: #f4b844;
+`;
+
+const Message = styled(Text)`
+  width: 100%;
+`;
```
samwhale: 💭 I feel like this font size is really big compared to the rest of the panel font sizes. Should we make it a little smaller? @agingoldseco

samwhale (commented):
This works great. I left a few ideas about how we may not need to update `getFailingPages` but rather change the way we use it.

packages/story-editor/src/components/checklist/checks/pageBackgroundLowTextContrast/check.js
BrookeGraham: That makes total sense, I don't know why I made it harder than it should be.

packages/story-editor/src/components/panels/design/warning/warning.js
BrookeGraham: Agreed

BrookeGraham: Ooops this was my bad. Updated to match figma ie 14px.

packages/story-editor/src/components/panels/design/warning/warning.js
BrookeGraham: I added `warning: {yellow: #f4b844}` Don't know if that is how we want to add this. We didn't have anything similar so I am unsure. Do I need to add something to Figma as well?

packages/story-editor/src/components/checklist/checks/pageBackgroundLowTextContrast/check.js
BrookeGraham: This was so obvious I seriously don't know why I did the other thing 🤦🏻‍♀️ Thanks so much @samwhale

packages/story-editor/src/components/panels/design/pageBackground/pageBackground.js
BrookeGraham: Happy to update copy to whatever y'all think is best. I am not great at copy so I have no preference.

BrookeGraham:
@agingoldseco What do you think about the copy on the background panel vs the text panel? Also how do you feel about the copy including a bit about font size? ie text panel: `This color combination or font size may be hard for people to read.` and background panel: `This background color may make text hard to read.` or something.

packages/story-editor/src/components/panels/design/warning/warning.js
BrittanyIRL: you don't need to add to figma - the theme colors populate in storybook: https://google.github.io/web-stories-wp/storybook/?path=/story/designsystem-colors--default 

I think that this yellow goes with the `status` section of colors, keeping colors organized by purpose can allow for easier updates. If we call it yellow and then maybe someday (just an example) the designs change the warning color to be an orange, now that reads wrong because the yellow is actually orange and we'd have to update the instances of that instead of just the theme color.

packages/design-system/src/theme/colors.js
```diff
@@ -100,6 +100,10 @@ const standard = {
   white: '#FFF',
 };
 
+const warning = {
+  yellow: '#f4b844',
```
BrittanyIRL: let's move this up to `status`  (line 94) - `warning: '#f4b844'`

packages/story-editor/src/components/panels/design/textStyle/karma/contrast.karma.js
```diff
@@ -0,0 +1,82 @@
...
+        fixture.editor.inspector.designPanel.textStyle.fontColor.button
+      );
+      await waitFor(() =>
+        fixture.events.click(
+          fixture.screen.getByRole('option', { name: '#fff' })
+        )
+      );
+
+      const contrastWarning = fixture.screen.getByTestId('warningContainer');
+      await expectAsync(contrastWarning).toHaveNoViolations();
```
BrittanyIRL: 😻

BrittanyIRL (commented):
Que bonita. Let's wait for design confirmation on copy. Otherwise, just the theme warning color getting put into `status` instead of a new object.

packages/story-editor/src/components/colorPicker/karma/colorPicker.karma.js
```diff
@@ -203,18 +203,13 @@ describe('ColorPicker', () => {
           '<span style="color: #c4c4c4">Fill in some text</span>'
         );
       });
+    });
 
-      it('should allow saving text background color', async () => {
-        // Add text element
+    describe('CUJ: Creator can Apply or Save a Color from/to Their Preset Library: Manage Color Presets', () => {
+      it('should allow deleting local and global color presets', async () => {
+        // Add text element and a color preset.
```
BrookeGraham: Oops bad rebase. Fixing.

packages/story-editor/src/components/colorPicker/karma/colorPicker.karma.js
```diff
@@ -158,7 +158,11 @@ describe('ColorPicker', () => {
 
         // Add text and apply the previously saved color.
         await fixture.events.click(fixture.editor.library.textAdd);
-        await waitFor(() => fixture.editor.canvas.framesLayer.frames[1].node);
+        await waitFor(() => {
+          if (!fixture.editor.canvas.framesLayer.frames[1].node) {
+            throw new Error('node not ready');
+          }
+        });
```
BrookeGraham: I had to fix a bad rebase, so while I was in here I updated the `waitFor`s too be a little more robust and actually work the way we want them too.

agingoldseco:
>What do you think about the copy on the background panel vs the text panel? Also how do you feel about the copy including a bit about font size? ie text panel: `This color combination or font size may be hard for people to read.` and background panel: `This background color may make text hard to read.` or something.

I like the bit about font size and separating out the warning types! But would a warning, more often than not, show up in both the background and text panels at once? 

I think this copy works, would just add the word "some" to soften it a bit? `This color combination or font size may be hard for some people to read.` and `This background color may make some text hard to read.` —does that make sense?

BrookeGraham:
> But would a warning, more often than not, show up in both the background and text panels at once?

Yep, @agingoldseco  it would be displayed in both places, but those two panels are not ever open at the same time. Depending on what element is select, the corresponding panel would show. ie. if you have the background selected, you wouldn't see the text panel info / warning.

packages/story-editor/src/components/panels/design/pageBackground/pageBackground.js
```diff
@@ -239,6 +261,9 @@ function PageBackgroundPanel({ selectedElements, pushUpdate }) {
           />
         </Row>
       )}
+      {failedContrast && (
+        <Warning message={ACCESSIBILITY_COPY.lowContrast.backgroundPanel} />
+      )}
```
samwhale, line 270: Thanks for all the updates! Was giving it one more sweep and seeing this:

I added some text that's the same color as an image (black). The contrast check shows correctly on the text but the `page background` also shows the warning even though the text is fine against the background.

May be more of a UX question but I think we should probably show the warning on the element that the text is against rather than the background in this case: (text doesn't show because it's the same color as the black part of the image)

image

BrookeGraham: I see what you are saying. In this case the "background" doesn't really matter because the text is on top of the image. Making the image the _actual_ background. Therefore the warning shouldn't appear on the background panel. 

There are going to be a ton of edge cases like this. I am not sure how we are going to handle those. 🤔

BrookeGraham: I'll look into only showing the message here if the failure is cause by background color as opposed to images. Also, the message being here wasn't a requirement so happy to remove it if we decide it's too confusing.

samwhale: Totally - I don't want any scope creep here.

I think either of these would be good:
1. Always the contrast warning from the page background panel
2. Only removing the contrast warning from the page background panel when the text contrast problem isn't from the background

BrookeGraham: Such a great catch @samwhale I have updated to only show on the background panel if the contrast problem stems from the actual background element.

samwhale (approved):
Thanks for making those fixes, works great!

packages/story-editor/src/components/panels/design/pageBackground/pageBackground.js
maxyinger: Hmph thinking about it now, I'm wondering if the background is a good candidate for this check or not. The primary scenario I'm thinking about is:
- you have 2 text elements, one has high enough contrast with the background, the other does not
- you select the background and see that it has a warning that says it doesn't have enough contrast, and the user should change the color
- the user changes the color to better contrast with the original conflicting text, but now the background is conflicting with the other text element (that didn't throw an error before)
- to the user, they changed the color, but the warning just never went away

maxyinger: ^^ not sure if this is a likely scenario, but curious if it's one we should consider.

maxyinger: anyway, I don't want to throw a wrench in this PR which has already gotten design approval. I think if we find ourselves running into the scenario above more frequently, we can always do a little followup PR. Just some behavior to keep our eyes out for in the future

packages/story-editor/src/components/colorPicker/karma/colorPicker.karma.js
```diff
@@ -158,7 +158,12 @@ describe('ColorPicker', () => {
 
         // Add text and apply the previously saved color.
         await fixture.events.click(fixture.editor.library.textAdd);
-        await waitFor(() => fixture.editor.canvas.framesLayer.frames[1].node);
+        await waitFor(() => {
+          if (!fixture.editor.canvas.framesLayer.frames[1].node) {
+            throw new Error('node not ready');
+          }
+          expect(fixture.editor.canvas.framesLayer.frames[1].node).toBeTruthy();
```
maxyinger, line 165: if this expect was only ever in here for the `waitFor` statement, I wonder if we even need it anymore?

packages/story-editor/src/components/checklist/checks/pageBackgroundLowTextContrast/component.js
```diff
@@ -45,30 +45,14 @@ const PageBackgroundTextLowContrast = () => {
...
-
   useEffect(() => {
-    getFailingPages().then((failures) => {
-      setFailingPages(failures);
-    });
-  }, [getFailingPages]);
+    getPagesWithFailedContrast(storyPages, pageSize)
+      .then((failures) => {
+        const failedPages = failures.map(({ page }) => page);
+        failedPages.length && setFailingPages(failedPages);
```
maxyinger: Ahh I think this may be a bug.

Lets run through the scenario:
- a story with one page that has a white background, a black square, and some black text, with the black text positioned over the background
- you move the text to be on top of the black square
- this triggers a failing page
- you open the help center and see this check
- with the help center open, you move the text to be positioned back over the background
- this check should go away, but doesn't because it only sets state if `failedPages.length`

maxyinger: Also small (and probably very rare) race case to consider is if a user is able to update the story multiple times before this check completes.

We may want some sort of stale flag to ensure that we're always deriving failed pages off of the freshest version of `storyPages`. Something like:
```js
useEffect(() => {
  // can name this whatever you want
  let isInputStale = false;
  getPagesWithFailedContrast(storyPages, pageSize)
      .then((failures) => {
        if (isInputStale) {
          return;
        }
        const failedPages = failures.map(({ page }) => page);
        setFailingPages(failedPages);
      })
      .catch(() => {
        // maybe add a TODO comment in here or something :shrug:
      });
     
  return () => {
    isInputStale = true;
  }
}, [storyPages, pageSize]);
```

maxyinger (changes requested):
Other than the small bug I pointed out, this looks great and works as expected. Happy to slap an approval on this after that's fixed.

**Small note on the implementation chosen here:**
The only thing I'd watch out for is that the contrast check was the most expensive check in the PPC. It was a big part of why we had to refactor the PPC.

The solution proposed here, basically doubles the expensive computation by performing it both in the PPC and in the design panel. We should keep our eyes out for this, and we may need to figure out a solution in the future that performs this expensive calculation once and stores it some place we can reference it vs performing this calculation everywhere we need it referenced.

Anyway, we still haven't performed an audit of the design panels yet. If this does make a noticeable impact we can always update the approach then

packages/story-editor/src/components/checklist/checks/pageBackgroundLowTextContrast/component.js
BrookeGraham: Great catch. Not sure how I overlooked that.

maxyinger (approved):

packages/story-editor/src/components/panels/design/pageBackground/pageBackground.js
BrookeGraham: Yeah I had thought about that too. My thinking was that regardless of which text element is causing the error, there is still an error and they should find a background that is suitable for all text if they don't want to see the message. The copy here was updated to `This background color may make some text hard to read.` Which I thinks helps convey that _some_  not necessarily all text may be hard to read.

packages/story-editor/src/components/colorPicker/karma/colorPicker.karma.js
BrookeGraham: Above we are clicking the `addText` button. Since these test can run kinda fast in ci, the node hasn't always been mounted to the dom by the time we need to interact with it. So we added the `waitFor` to force karma to repoll until the node appears on the dom. At least that is my understanding. Otherwise we end up with a flakey test that doesn't always find the node it is looking for.

BrookeGraham:
>The solution proposed here, basically doubles the expensive computation by performing it both in the PPC and in the design panel. We should keep our eyes out for this, and we may need to figure out a solution in the future that performs this expensive calculation once and stores it some place we can reference it vs performing this calculation everywhere we need it referenced.

Totally agree! It is a lot. 

>The only thing I'd watch out for is that the contrast check was the most expensive check in the PPC. It was a big part of why we had to refactor the PPC.

While I was testing, it appears that the checklist mounts and calculates this function regardless of whether or not the checklist 'panel' is open. It then reruns the `getFailingPages` (well, the whole `AccessibilityChecks`) check every time it is opened and closed. 

@littlemilkstudio ☝️

BrittanyIRL:
this test failure isn't because of this pr - looks like the default templates got updated in two PRs and one didn't get tests re run after the other so the failure didn't show up: https://github.com/google/web-stories-wp/pull/10078 and https://github.com/google/web-stories-wp/pull/10217   

(new issue to fix it: https://github.com/google/web-stories-wp/issues/10263)

https://github.com/google/web-stories-wp/runs/4871081344?check_suite_focus=true#step:6:386
Screen Shot 2022-01-19 at 10 18 52 AM

BrittanyIRL (approved):

PR #10076, opened 2021-12-30
Decoupling: Hide 'copy story url' from dashboard menu when url is `undefined`

mohdsayed:
## Context

Hide “Copy story URL” and "Open in new Tab" links when URL is `undefined`. Also because they look redundant in [storybook](https://google.github.io/web-stories-wp/storybook/?path=/story/playground-dashboard--default)

## Testing Instructions

1.

## Checklist

---


swissspidy:
The one thing that bugs me here is that it's very unintuitive, because of the way `menuItems` and `menuItemActions` are constructed in the dashboard.

I don't see why these are two separate things.

As a dev I'd have expected `action` to be part of the menuItem, e.g. something more like `menuItems = [ { label: '...', value: '...', capability: '', action: '...' } ] `, that would make the filtering easier: if there's no action, just remove it.

We could even turn `capability` into a more general purpose callback that is called in the `menuItemsFiltered` filter.

Maybe something like this:

```js
export const STORY_CONTEXT_MENU_ITEMS = [
  {
    label: __('Open in editor', 'web-stories'),
    value: STORY_CONTEXT_MENU_ACTIONS.OPEN_IN_EDITOR,
    isEnabled: (story) => Boolean(story?.capabilities.hasEditAction),
  },
  {
    label: __('Open in new tab', 'web-stories'),
    value: STORY_CONTEXT_MENU_ACTIONS.OPEN_STORY_LINK,
    isEnabled: (story) => Boolean(story.previewLink),
  },
  {
    label: __('Copy Story URL', 'web-stories'),
    value: STORY_CONTEXT_MENU_ACTIONS.COPY_STORY_LINK,
    isEnabled: (story) => Boolean(story.previewLink),
  },
  {
    label: __('Rename', 'web-stories'),
    value: STORY_CONTEXT_MENU_ACTIONS.RENAME,
    separator: 'top',
    isEnabled: (story) => Boolean(story?.capabilities.hasEditAction),
  },
  {
    label: __('Duplicate', 'web-stories'),
    value: STORY_CONTEXT_MENU_ACTIONS.DUPLICATE,
    isEnabled: (story) => Boolean(story?.capabilities.hasEditAction),
  },
  {
    label: __('Delete Story', 'web-stories'),
    value: STORY_CONTEXT_MENU_ACTIONS.DELETE,
    separator: 'top',
    isEnabled: (story) => Boolean(story?.capabilities.hasDeleteAction),
  },
];

// And then in generateStoryMenu():

const menuItemsFiltered = menuItems.filter(
  (item) => typeof item.isEnabled === 'undefined' || item.isEnabled(story)
);
```

BrittanyIRL:
@samwhale refactored the context menu, let's pause on this until that can get merged: https://github.com/google/web-stories-wp/pull/9960/files

BrittanyIRL:
> @samwhale refactored the context menu, let's pause on this until that can get merged: https://github.com/google/web-stories-wp/pull/9960/files

That PR got merged yesterday into main.

BrittanyIRL (approved):
Am I correct in this being to resolve storybook showing these redundant options like so? 
Screen Shot 2022-01-18 at 4 40 33 PM

Tested the dashboard for any regressions and it looks good. Dig the update to use a switch in storiesView.  This menu has historically been a really hard to read - it just grew beyond its initial scope and functionality kept getting added on top. Appreciate the eyes on it.

mohdsayed:
> Am I correct in this being to resolve storybook showing these redundant options like so?

@BrittanyIRL Yes, thanks for testing.

packages/dashboard/src/app/views/myStories/content/storyGridView/index.js
```diff
@@ -136,31 +136,42 @@ const StoryGridView = ({
...
+      STORY_CONTEXT_MENU_ACTIONS.DELETE,
+      STORY_CONTEXT_MENU_ACTIONS.RENAME,
+      STORY_CONTEXT_MENU_ACTIONS.DUPLICATE,
+    ];
+
+    const menuItems = storyMenu.menuItems.map((item) => {
+      return actions.includes(item?.value)
+        ? getModifiedMenuItem(item?.value)
+        : item;
+    });
```
swissspidy:
```suggestion
    const menuItems = storyMenu.menuItems.map((item) => {
      if (actions.includes(item?.value)) {
        item.action = (story) => {
          manuallySetFocusOut();
          item.action(story);
        };
      }

      return item;
    });
```

If we do this then we don't need `getModifiedMenuItem`

mohdsayed: Maybe it's creating some infinite loop somewhere 🤔 , checking...

swissspidy (approved):

swissspidy:
Some failures all over the place now 👀

packages/dashboard/src/app/views/myStories/content/storyGridView/index.js
mohdsayed: When using a keyboard shortcut to delete the story

image

mohdsayed: In fact, even when clicking the delete button it creates an infinite loop.

mohdsayed: Fixed now, waiting for tests to pass.

samwhale:
Seeing this error in the console when visiting the dashboard

image

samwhale (approved):
All current functionality works great other than that error in the console. Thanks for the refactor here - didn't love the way it used to be.

But couldn't test the main idea: How do you set a story to have an `undefined` url?

swissspidy:
> But couldn't test the main idea: How do you set a story to have an undefined url?

`npm run storybook` and open the dashboard there

samwhale:
Tested in storybook and looks good. Thanks @swissspidy

PR #10078, opened 2021-12-30
Page Templates: Fix dropdown select in library pane

### LINKED ISSUE #10077: Page Templates: selecting already selected dropdown option breaks storybook
## Bug Description

In the dropdown menu to select what template list ( default or custom ) needs to be shown, selecting the already selected option toggles the selection.

This bug can also be seen in the current storybook build, but the callback getCustomPageTemplates not being defined here breaks the editor. Subsequent loads of storybook will also break on opening page templates pane until localStorage is cleared.

## Expected Behaviour

Not change the selection in the dropdown when an already selected option is selected.

## Screenshots

https://user-images.githubusercontent.com/53055971/147749201-4ff32958-0430-4d3c-b4e7-64fb4b5a831e.mp4

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

ayushnirwal:
## Context

Fixes dropdown select function in pageTemplatesPane in the library.

## Summary

In the dropdown menu to select what template list ( default or custom ) needs to be shown, selecting the already selected option toggles the selection.

## Testing Instructions

1.

## Checklist

---

Fixes #10077


packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
```diff
@@ -157,9 +157,9 @@ function PageTemplatesPane(props) {
     setNextTemplatesToFetch,
   ]);
 
-  const handleToggle = () => {
-    setShowDefaultTemplates(!showDefaultTemplates);
-    localStore.setItemByKey(LOCAL_STORAGE_KEY, !showDefaultTemplates);
+  const handleSelect = (event, menuItem) => {
+    setShowDefaultTemplates(DEFAULT === menuItem);
```
BrookeGraham: Excellent find. Only thing I would suggest would be to not reset `showDefaultTemplates` or `localStore` if the new selection is the same as what is already selected. No need to reset them if they are just gonna be the same. Something like
```
if (showDefaultTemplates !== (DEFAULT === menuItem)) {
...}
```

BrookeGraham (approved):

packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
```diff
@@ -157,9 +157,9 @@ function PageTemplatesPane(props) {
     setNextTemplatesToFetch,
   ]);
 
-  const handleToggle = () => {
-    setShowDefaultTemplates(!showDefaultTemplates);
-    localStore.setItemByKey(LOCAL_STORAGE_KEY, !showDefaultTemplates);
+  const handleSelect = (event, menuItem) => {
```
BrittanyIRL: looks like `event` isn't getting used. When targeting the second value passed through we typically do `(_, menuItem)`

BrittanyIRL (changes requested):
Thanks so much for this contribution. I agree with @BrookeGraham 's suggestion to check the values against each other before firing functions.  After you update just tag for re-review.

BrittanyIRL (commented):
Awesome! thanks for the updates and so sorry for the delayed response. If you would be so kind as to rebase this branch off of `/main` so we can double check that all of the integration tests are still good (they should be), then this will be good to go.

BrittanyIRL (approved):
thank you so much for the extra hoops to jump through for this. appreciate the fix a lot.  We will send this on to QA and merge once they are through with it.

swissspidy:
Note that external contributions can't easily be QAed on our env right now due to some limitations. If you have verified that this fixes an actual issue feel free to merge this without another person looking at it.

PR #10082, opened 2022-01-03
Decoupling: Fix `updateCurrentUser` error when browsing help center tip

mohdsayed:
## Context

When browsing help center tips in storybook, the editor breaks because `updateCurrentUser` is not defined.

https://user-images.githubusercontent.com/6297436/148059732-e0588b74-52c5-4d64-b41d-569765aff71d.mp4

## Testing Instructions

1.

## Checklist

---

See https://github.com/google/web-stories-wp/pull/9949


swissspidy:
Sounds like there is some infinite loop or something?

packages/story-editor/src/app/helpCenter/provider.js
```diff
@@ -213,6 +214,10 @@ const createBooleanMapFromKey = (key) =>
 
 function HelpCenterProvider({ children }) {
   const { additionalTips = {} } = useConfig();
+  const {
+    actions: { updateCurrentUser: _updateCurrentUser },
+  } = useAPI();
+  const canUpdateCurrentUser = Boolean(_updateCurrentUser);
```
swissspidy, line 220: Could probably move this to `CurrentUserProvider`

mohdsayed:
> Sounds like there is some infinite loop or something?

Did not see an infinite loop, but I see this error when clicking on "Next" button in tips because `updateCurrentUser` is not provided.

image

swissspidy:
Hmm, so theoretically we only need the change in `packages/story-editor/src/app/currentUser/currentUserProvider.js` here, no?

mohdsayed:
> Hmm, so theoretically we only need the change in `packages/story-editor/src/app/currentUser/currentUserProvider.js` here, no?

The `.catch` part is also a problem in `provider.js`. 

https://github.com/google/web-stories-wp/blob/2d0a090015349a786e676864c5652df5a134893a/packages/story-editor/src/app/helpCenter/provider.js#L278-L282

If we only want to update `currentUserProvider.js`, we may return an empty promise from `updateCurrentUser` if is not defined here

https://github.com/google/web-stories-wp/blob/2d0a090015349a786e676864c5652df5a134893a/packages/story-editor/src/app/currentUser/currentUserProvider.js#L41-L44

packages/story-editor/src/app/helpCenter/provider.js
mohdsayed: You mean pass `canUpdateCurrentUser` as state from `CurrentUserProvider` ? We could but, that doesn't look very intuitive because `canUpdateCurrentUser` sounds more like a capability which it is not.

swissspidy (approved):

PR #10083, opened 2022-01-03
Decoupling: Make `getFonts` API callback optional

mohdsayed:
## Context

We have tried to make most API callbacks optional to reduce the effort to spin up a new story-editor and dashboard app ( unless the app cannot work without those callbacks ) in https://github.com/google/web-stories-wp/pull/9949 and https://github.com/google/web-stories-wp/pull/9752. We recently added a new required API callback `getFonts` in https://github.com/google/web-stories-wp/pull/9952 but I think if we keep it optional, creating a standalone editor would be a bit simpler.

## Testing Instructions

1.

## Checklist

---

See https://github.com/google/web-stories-wp/pull/9949,  https://github.com/google/web-stories-wp/pull/9952


spacedmonkey:
Can you give some more detail on this PR?

mohdsayed:
> Can you give some more detail on this PR?

@spacedmonkey Added to the PR description.

mohdsayed:
@swissspidy Not fully satisfied with how I am keeping `getFonts` function in `packages/story-editor/src/defaultConfig.js`. Wanted to know your thoughts on any alternative or if you think `getFonts` should not be optional at all.

swissspidy:
So I am not 100% sure about this either.

It probably depends on what other consumers/platforms might want here.

Would they all want _all_ Google Fonts or perhaps just a subset of them? Or maybe even only system fonts (Arial, Verdana, etc.)?
Would they want custom fonts too and thus customize this anyway?

For a default config that might not be used most of the times, the bundle size increase of 55KB is certainly not great, so I'd lean towards not making this optional at all — or at least use a different default config.

WDYT?

mohdsayed:
@swissspidy All I wanted was that it should not be a hard dependency to spin up a standalone editor. Earlier the editor was breaking without `getFonts` callback but I see that `getFonts` calls were wrapped in `try/catch` recently, so it's no longer breaking and the text section is also looking okay without this callback. So we are good, closing the PR.

image

swissspidy:
`getFontsBySearch` still uses `getFonts` without any `try / catch`. So perhaps still an issue?

mohdsayed:
Without a font list, the dropdown doesn't appear and therefore the fonts can not be searched. So I think `getFontsBySearch` wouldn't be called. But will raise a quick PR for it if that's an issue for integration tests @ayushnirwal is writing.

PR #10095, opened 2022-01-05
Fix Karma tests / disconnect issues

### LINKED ISSUE #10052: Story Editor Karma Timeout Failure

BrittanyIRL:
## Context

Back around #9953 and #9952 karma started to disconnect super early - running around 0 to 100 tests but since the tests were just "skipped" they [weren't showing up as errors](https://github.com/google/web-stories-wp/runs/4761414889?check_suite_focus=true#step:6:3953). This allowed some updates to sneak past that carried with them things that needed to facilitate change that weren't caught. 

I can't pin this issue on any one thing that was updated. I really do believe it was a combination of updates between blurhash, fonts, puppeteer, and chromium - that these just kinda all stacked up and break our config. 

I think this fixes it for now. I'm including a list of other warnings and errors in these latest karma runs that we should figure out a solution for independently sooner rather than later. Max is also working on solutions for the memory limit issue here #10143 . 

WIP PRs from the last week that have gotten us to this point: 
https://github.com/google/web-stories-wp/pull/10103
https://github.com/google/web-stories-wp/pull/10073
https://github.com/google/web-stories-wp/pull/10120
https://github.com/google/web-stories-wp/pull/10102
https://github.com/google/web-stories-wp/pull/10132

Helpful threads: 
https://github.com/karma-runner/karma-chrome-launcher/issues/154#issuecomment-712233562
https://bbs.archlinux.org/viewtopic.php?pid=1967969#p1967969

## Summary

Restore Karma connection: 
```
Chrome Headless 93.0.4577.0 (Linux x86_64): Executed 407 of 461 (skipped 54) SUCCESS (40 mins 23.278 secs / 40 mins 4.308 secs)
TOTAL: 407 SUCCESS
```

## Relevant Technical Choices

Handful of fixes here. 

The way fonts are getting retrieved changed and Karma was throwing a fit about the external links because they weren't listed in karma-story-editor.config.cjs. We've added fonts.googleapis.com to the list of files that are OK for karma to go grab (regex was Pascal's idea). We're still seeing some errors about not finding the font css inline but not nearly as much as before - this also fixes some of the issues I was seeing for text elements where nodes weren't getting found (combined with the updates made there for waiting). Reference: [Karma Runner Docs]( https://karma-runner.github.io/6.3/config/files.html#loading-files-from-another-server)

`waitForInitialPage` is now set to true by default in karma-puppeteer-launcher, this might not need to stick around depending on what @littlemilkstudio comes up with. For now, I found it doesn't hurt and anything we can do to make sure these tests know to hold up until things are loaded is helpful. [This also defaults to true when it's not included.](https://github.com/puppeteer/puppeteer/blob/main/docs/api.md)  I'll list this as a 'todo' to explore removing it here after we get unblocked. Seems like a silly reason to run everything again. 

`--force-gpu-mem-discardable-limit-mb` is added to puppeteer's args for launch.  This sets the maximum GPU memory to use for discardable caches. ✨  this is part of the magic!!! This prevents disconnect timeouts!!! Reference: [Chromium Command Line Switches](https://peter.sh/experiments/chromium-command-line-switches/#force-gpu-mem-discardable-limit-mb)

`--disable-software-rasterizer` - so the force gpu arg was enough to get remote headless working but when you run headless locally (which is important to maintain), it wasn't cutting it. This argument combined with the force gpu seems to be our winning combo. See this [Stack Overflow Thread](https://stackoverflow.com/questions/67501093/passthrough-is-not-supported-gl-is-disabled) (credit to @BrookeGraham and @samwhale for this one) 

Passing all the puppeteerOptions to puppeteer on launch. This just seemed consistent since the options all match up anyways and we want the args to end up in there on launch to get seen. 

Adding `elementId` to `loadStylesheet` to prevent `id=undefined` which was making allllll of the roboto font styles stack up in the head since the browser's not actually refreshing for each set of tests. Prevents a bunch of lag. Stylesheets were duplicated around 700 times if the tests all run. 

Add 2 missing `blurHash` mocks for the media items in `mediaFetching` (again, thanks Brooke!) 

Various test updates (mostly related to text nodes) to wait so that fonts have time to catch up. 

## To-do

Potential areas to look into to make our integration tests more robust: 

- [overlapping act() calls]( https://github.com/google/web-stories-wp/runs/4766928054?check_suite_focus=true#step:6:1160)  -- incomplete usage of async/await
-[ error inlining remote css ](https://github.com/google/web-stories-wp/runs/4767418885?check_suite_focus=true#step:6:3402)- tried to fix this by adding regex pattern for google fonts to config file, but still getting this error sometimes. 
- [Failed to connect to the bus:](https://github.com/google/web-stories-wp/runs/4767418885?check_suite_focus=true#step:6:1156) Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix") 
- [ALSA lib confmisc.c:767:](https://github.com/google/web-stories-wp/runs/4767418885?check_suite_focus=true#step:6:1170)(parse_card) cannot find card '0'; ALSA lib conf.c:4732:(_snd_config_evaluate) function snd_func_card_driver returned error: No such file or directory; ALSA lib confmisc.c:392:(snd_func_concat) error evaluating strings (feels related to checklist?) 
- Tests I disabled to give us a clean success: 
  - [should have a delay before autoplaying videos](https://github.com/google/web-stories-wp/issues/10144)
  - [should use "Esc" to exit the dialog](https://github.com/google/web-stories-wp/issues/10146)
  - [should correctly handle image dragged from library straight to edge (no cached base color)](https://github.com/google/web-stories-wp/issues/10145)
  - [should not show cards that require the \`hasUploadMediaAction\` permission](https://github.com/google/web-stories-wp/issues/10147) 

## User-facing changes

None. 

## Testing Instructions

Verify that karma's working again and nothing new broke. A quick regression test on the user side would be good for sanity.

1.

## Checklist

---

Fixes #10052

### DIFF
--- a/karma-story-editor.config.cjs
+++ b/karma-story-editor.config.cjs
@@ -68,53 +68,53 @@ module.exports = function (config) {
         served: true,
         nocache: false,
       },
+      // {
+      //   // eslint-disable-next-line prettier/prettier
+      //   pattern: '^https://fonts.googleapis.com/css*',
+      //   type: 'css',
+      //   included: false,
+      //   served: true,
+      //   nocache: false,
+      // },
       {
-        // eslint-disable-next-line prettier/prettier
-        pattern: '^https://fonts.googleapis.com/css*',
+        pattern:
+          'https://fonts.googleapis.com/css2?display=auto&family=Roboto%3Aital%2Cwght%400%2C100%3B0%2C300%3B0%2C400%3B0%2C500%3B0%2C700%3B0%2C900%3B1%2C100%3B1%2C300%3B1%2C400%3B1%2C500%3B1%2C700%3B1%2C900',
         type: 'css',
-        included: false,
+        included: true,
         served: true,
         nocache: false,
       },
-      // {
-      //   pattern:
-      //     'https://fonts.googleapis.com/css2?display=auto&family=Roboto%3Aital%2Cwght%400%2C100%3B0%2C300%3B0%2C400%3B0%2C500%3B0%2C700%3B0%2C900%3B1%2C100%3B1%2C300%3B1%2C400%3B1%2C500%3B1%2C700%3B1%2C900',
-      //   type: 'css',
-      //   included: true,
-      //   served: true,
-      //   nocache: false,
-      // },
-      // {
-      //   pattern: 'https://fonts.googleapis.com/css2?display=auto&family=Bungee',
-      //   type: 'css',
-      //   included: true,
-      //   served: true,
-      //   nocache: false,
-      // },
-      // {
-      //   pattern:
-      //     'https://fonts.googleapis.com/css2?display=auto&family=Bungee+Shade',
-      //   type: 'css',
-      //   included: true,
-      //   served: true,
-      //   nocache: false,
-      // },
-      // {
-      //   pattern:
-      //     'https://fonts.googleapis.com/css2?display=auto&family=Chivo%3Aital%2Cwght%400%2C300%3B0%2C400%3B0%2C700%3B0%2C900%3B1%2C300%3B1%2C400%3B1%2C700%3B1%2C900',
-      //   type: 'css',
-      //   included: true,
-      //   served: true,
-      //   nocache: false,
-      // },
-      // {
-      //   pattern:
-      //     'https://fonts.googleapis.com/css?family=Google+Sans%7CGoogle+Sans:b%7CGoogle+Sans:500&display=swap',
-      //   type: 'css',
-      //   included: true,
-      //   served: true,
-      //   nocache: false,
-      // },
+      {
+        pattern: 'https://fonts.googleapis.com/css2?display=auto&family=Bungee',
+        type: 'css',
+        included: true,
+        served: true,
+        nocache: false,
+      },
+      {
+        pattern:
+          'https://fonts.googleapis.com/css2?display=auto&family=Bungee+Shade',
+        type: 'css',
+        included: true,
+        served: true,
+        nocache: false,
+      },
+      {
+        pattern:
+          'https://fonts.googleapis.com/css2?display=auto&family=Chivo%3Aital%2Cwght%400%2C300%3B0%2C400%3B0%2C700%3B0%2C900%3B1%2C300%3B1%2C400%3B1%2C700%3B1%2C900',
+        type: 'css',
+        included: true,
+        served: true,
+        nocache: false,
+      },
+      {
+        pattern:
+          'https://fonts.googleapis.com/css?family=Google+Sans%7CGoogle+Sans:b%7CGoogle+Sans:500&display=swap',
+        type: 'css',
+        included: true,
+        served: true,
+        nocache: false,
+      },
       'node_modules/axe-core/axe.js',
     ],
 
--- a/karma-story-editor.config.cjs
+++ b/karma-story-editor.config.cjs
@@ -69,52 +69,52 @@ module.exports = function (config) {
         nocache: false,
       },
       {
-        pattern:
-          'https://fonts.googleapis.com/css?family=Google+Sans|Google+Sans:b|Google+Sans:500&display=swap',
+        // eslint-disable-next-line prettier/prettier
+        pattern: '^https://fonts.googleapis.com/css2*',
         type: 'css',
-        included: true,
+        included: false,
         served: true,
         nocache: false,
       },
-      {
-        pattern:
-          'https://fonts.googleapis.com/css2?display=auto&family=Roboto%3Aital%2Cwght%400%2C100%3B0%2C300%3B0%2C400%3B0%2C500%3B0%2C700%3B0%2C900%3B1%2C100%3B1%2C300%3B1%2C400%3B1%2C500%3B1%2C700%3B1%2C900',
-        type: 'css',
-        included: true,
-        served: true,
-        nocache: false,
-      },
-      {
-        pattern: 'https://fonts.googleapis.com/css2?display=auto&family=Bungee',
-        type: 'css',
-        included: true,
-        served: true,
-        nocache: false,
-      },
-      {
-        pattern:
-          'https://fonts.googleapis.com/css2?display=auto&family=Bungee+Shade',
-        type: 'css',
-        included: true,
-        served: true,
-        nocache: false,
-      },
-      {
-        pattern:
-          'https://fonts.googleapis.com/css2?display=auto&family=Chivo%3Aital%2Cwght%400%2C300%3B0%2C400%3B0%2C700%3B0%2C900%3B1%2C300%3B1%2C400%3B1%2C700%3B1%2C900',
-        type: 'css',
-        included: true,
-        served: true,
-        nocache: false,
-      },
-      {
-        pattern:
-          'https://fonts.googleapis.com/css?family=Google+Sans%7CGoogle+Sans:b%7CGoogle+Sans:500&display=swap',
-        type: 'css',
-        included: true,
-        served: true,
-        nocache: false,
-      },
+      // {
+      //   pattern:
+      //     'https://fonts.googleapis.com/css2?display=auto&family=Roboto%3Aital%2Cwght%400%2C100%3B0%2C300%3B0%2C400%3B0%2C500%3B0%2C700%3B0%2C900%3B1%2C100%3B1%2C300%3B1%2C400%3B1%2C500%3B1%2C700%3B1%2C900',
+      //   type: 'css',
+      //   included: true,
+      //   served: true,
+      //   nocache: false,
+      // },
+      // {
+      //   pattern: 'https://fonts.googleapis.com/css2?display=auto&family=Bungee',
+      //   type: 'css',
+      //   included: true,
+      //   served: true,
+      //   nocache: false,
+      // },
+      // {
+      //   pattern:
+      //     'https://fonts.googleapis.com/css2?display=auto&family=Bungee+Shade',
+      //   type: 'css',
+      //   included: true,
+      //   served: true,
+      //   nocache: false,
+      // },
+      // {
+      //   pattern:
+      //     'https://fonts.googleapis.com/css2?display=auto&family=Chivo%3Aital%2Cwght%400%2C300%3B0%2C400%3B0%2C700%3B0%2C900%3B1%2C300%3B1%2C400%3B1%2C700%3B1%2C900',
+      //   type: 'css',
+      //   included: true,
+      //   served: true,
+      //   nocache: false,
+      // },
+      // {
+      //   pattern:
+      //     'https://fonts.googleapis.com/css?family=Google+Sans%7CGoogle+Sans:b%7CGoogle+Sans:500&display=swap',
+      //   type: 'css',
+      //   included: true,
+      //   served: true,
+      //   nocache: false,
+      // },
       'node_modules/axe-core/axe.js',
     ],
 
--- a/karma-story-editor.config.cjs
+++ b/karma-story-editor.config.cjs
@@ -68,54 +68,7 @@ module.exports = function (config) {
         served: true,
         nocache: false,
       },
-      {
-        pattern:
-          'https://fonts.googleapis.com/css?family=Google+Sans|Google+Sans:b|Google+Sans:500&display=swap',
-        type: 'css',
-        included: true,
-        served: true,
-        nocache: false,
-      },
-      {
-        pattern:
-          'https://fonts.googleapis.com/css2?display=auto&family=Roboto%3Aital%2Cwght%400%2C100%3B0%2C300%3B0%2C400%3B0%2C500%3B0%2C700%3B0%2C900%3B1%2C100%3B1%2C300%3B1%2C400%3B1%2C500%3B1%2C700%3B1%2C900',
-        type: 'css',
-        included: true,
-        served: true,
-        nocache: false,
-      },
-      {
-        pattern: 'https://fonts.googleapis.com/css2?display=auto&family=Bungee',
-        type: 'css',
-        included: true,
-        served: true,
-        nocache: false,
-      },
-      {
-        pattern:
-          'https://fonts.googleapis.com/css2?display=auto&family=Bungee+Shade',
-        type: 'css',
-        included: true,
-        served: true,
-        nocache: false,
-      },
-      {
-        pattern:
-          'https://fonts.googleapis.com/css2?display=auto&family=Chivo%3Aital%2Cwght%400%2C300%3B0%2C400%3B0%2C700%3B0%2C900%3B1%2C300%3B1%2C400%3B1%2C700%3B1%2C900',
-        type: 'css',
-        included: true,
-        served: true,
-        nocache: false,
-      },
-      {
-        pattern:
-          'https://fonts.googleapis.com/css?family=Google+Sans%7CGoogle+Sans:b%7CGoogle+Sans:500&display=swap',
-        type: 'css',
-        included: true,
-        served: true,
-        nocache: false,
-      },
-      // eslint-disable-next-line prettier/prettier
+      // eslint-disable-next-line prettier/prettier, no-useless-escape
       '^https:\/\/fonts.googleapis.com\/css*',
       'node_modules/axe-core/axe.js',
     ],
--- a/karma-story-editor.config.cjs
+++ b/karma-story-editor.config.cjs
@@ -68,6 +68,14 @@ module.exports = function (config) {
         served: true,
         nocache: false,
       },
+      {
+        pattern:
+          'https://fonts.googleapis.com/css2?display=auto&family=Roboto%3Aital%2Cwght%400%2C100%3B0%2C300%3B0%2C400%3B0%2C500%3B0%2C700%3B0%2C900%3B1%2C100%3B1%2C300%3B1%2C400%3B1%2C500%3B1%2C700%3B1%2C900',
+        type: 'css',
+        included: true,
+        served: true,
+        nocache: false,
+      },
       'node_modules/axe-core/axe.js',
     ],
 
--- a/karma-story-editor.config.cjs
+++ b/karma-story-editor.config.cjs
@@ -115,11 +115,10 @@ module.exports = function (config) {
         served: true,
         nocache: false,
       },
+      // eslint-disable-next-line prettier/prettier
+      '^https:\/\/fonts.googleapis.com\/css*',
       'node_modules/axe-core/axe.js',
     ],
-
-    // ^https:\/\/fonts.googleapis.com\/css*
-
     // list of files / patterns to exclude
     exclude: ['**/test/**/*.js', '**/*.test.js'],
 
@@ -179,6 +178,7 @@ module.exports = function (config) {
         snapshots: config.snapshots || false,
         snapshotsDir: '.test_artifacts/karma_snapshots',
         defaultViewport: getViewport(config.viewport),
+        waitForInitialPage: true,
       },
     },
 
--- a/karma-story-editor.config.cjs
+++ b/karma-story-editor.config.cjs
@@ -70,7 +70,46 @@ module.exports = function (config) {
       },
       {
         pattern:
+          'https://fonts.googleapis.com/css?family=Google+Sans|Google+Sans:b|Google+Sans:500&display=swap',
+        type: 'css',
+        included: true,
+        served: true,
+        nocache: false,
+      },
+      {
+        pattern:
           'https://fonts.googleapis.com/css2?display=auto&family=Roboto%3Aital%2Cwght%400%2C100%3B0%2C300%3B0%2C400%3B0%2C500%3B0%2C700%3B0%2C900%3B1%2C100%3B1%2C300%3B1%2C400%3B1%2C500%3B1%2C700%3B1%2C900',
+        type: 'css',
+        included: true,
+        served: true,
+        nocache: false,
+      },
+      {
+        pattern: 'https://fonts.googleapis.com/css2?display=auto&family=Bungee',
+        type: 'css',
+        included: true,
+        served: true,
+        nocache: false,
+      },
+      {
+        pattern:
+          'https://fonts.googleapis.com/css2?display=auto&family=Bungee+Shade',
+        type: 'css',
+        included: true,
+        served: true,
+        nocache: false,
+      },
+      {
+        pattern:
+          'https://fonts.googleapis.com/css2?display=auto&family=Chivo%3Aital%2Cwght%400%2C300%3B0%2C400%3B0%2C700%3B0%2C900%3B1%2C300%3B1%2C400%3B1%2C700%3B1%2C900',
+        type: 'css',
+        included: true,
+        served: true,
+        nocache: false,
+      },
+      {
+        pattern:
+          'https://fonts.googleapis.com/css?family=Google+Sans%7CGoogle+Sans:b%7CGoogle+Sans:500&display=swap',
         type: 'css',
         included: true,
         served: true,
--- a/karma-story-editor.config.cjs
+++ b/karma-story-editor.config.cjs
@@ -131,7 +131,6 @@ module.exports = function (config) {
         snapshots: config.snapshots || false,
         snapshotsDir: '.test_artifacts/karma_snapshots',
         defaultViewport: getViewport(config.viewport),
-        waitForInitialPage: true,
       },
     },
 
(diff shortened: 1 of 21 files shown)


packages/story-editor/src/utils/loadStylesheet.js
```diff
@@ -19,6 +19,7 @@ function loadStylesheet(url, id) {
     const link = document.createElement('link');
     link.id = id;
     link.rel = 'stylesheet';
+    link.type = 'text/css';
```
swissspidy: This should not be mandatory. It's the default.

packages/story-editor/src/karma/fixture/fixture.js
```diff
@@ -230,7 +230,23 @@ export class Fixture {
...
+        return true;
+      }
+      for (let i = 2; i < styleSheets.length; i++) {
+        styleSheets[i].parentNode.removeChild(styleSheets[i]);
+      }
+      return true;
+    }
+
+    await clearSheets();
+
```
swissspidy: Good idea to clean up the stylesheets!

But let's target stylesheets with specific IDs instead of relying on some magic numbers.

Font CSS elements have an ID with the format `web-stories-${handle}-font-css`, so we could do a small regex check on the ID here.

We also need to target `<style>` inline styles for custom fonts.

We also need to add `elementId` as the second param here:

https://github.com/google/web-stories-wp/blob/2d0a090015349a786e676864c5652df5a134893a/packages/story-editor/src/app/font/actions/useLoadFontFiles.js#L45

So that the ID is actually properly added.

packages/story-editor/src/karma/fixture/fixture.js
```diff
@@ -230,7 +230,23 @@ export class Fixture {
     });
   }
 
-  restore() {
+  async restore() {
+    const styleSheets = window.document.querySelectorAll(
+      'link[rel=stylesheet]'
+    );
+
+    function clearSheets() {
```
swissspidy: `removeEditorFontStyles` or something.

packages/story-editor/src/components/eyedropper/index.js
```diff
@@ -91,7 +91,8 @@ export default ({ onChange }) => {
                     setEyedropperPixelData(pixelData);
                     setEyedropperImg(canvas.toDataURL());
                     resolve();
-                  });
+                  })
+                  .catch(() => undefined);
```
swissspidy: `resolve()` too?

packages/story-editor/src/app/font/actions/useLoadFontFiles.js
```diff
@@ -42,7 +42,12 @@ function useLoadFontFiles() {
 
     switch (service) {
       case 'fonts.google.com':
-        await loadStylesheet(getGoogleFontURL([{ family, variants }], 'auto'));
+        await loadStylesheet(
+          getGoogleFontURL([{ family, variants }], 'auto')
+        ).catch(() => {
```
swissspidy:
```suggestion
        await loadStylesheet(
          getGoogleFontURL([{ family, variants }], 'auto'),
          elementId
        ).catch(() => {
```

karma-story-editor.config.cjs
```diff
@@ -68,6 +68,14 @@ module.exports = function (config) {
...
         nocache: false,
       },
+      {
+        pattern:
+          'https://fonts.googleapis.com/css2?display=auto&family=Roboto%3Aital%2Cwght%400%2C100%3B0%2C300%3B0%2C400%3B0%2C500%3B0%2C700%3B0%2C900%3B1%2C100%3B1%2C300%3B1%2C400%3B1%2C500%3B1%2C700%3B1%2C900',
+        type: 'css',
+        included: true,
+        served: true,
+        nocache: false,
+      },
```
swissspidy: Interesting 👀 

Could you elaborate on this?

swissspidy:
Converting to draft just so we don't unnecessarily run some things like creating Percy snapshots

BrittanyIRL:
thanks for the early eyes, @swissspidy ! 

something's still wrong though - it just silently failed and skipped everything 🤔 
https://github.com/google/web-stories-wp/runs/4711058346?check_suite_focus=true#step:6:5419 

i'll keep this draft open and try a few things narrowed down. I thought I'd see where this landed running it last night before heading to bed since it was working locally, thus the 0 context 🤪

karma-story-editor.config.cjs
BrittanyIRL: absolutely, if this ends up sticking i'll add to the description. For now: 

I did this based on searching for other people having issues with running karma when adding in external fonts as `<Link />` - I know that we've been using Google fonts for a while but the way we're doing it is new and how they are getting sent out for has been updated. 
Adding this in fixed this error: `"Error loading remote stylesheet SecurityError: Failed to read the 'cssRules' property from 'CSSStyleSheet': Cannot access rules"` which pops up a lot (here's one as an example: https://github.com/google/web-stories-wp/runs/4704301224?check_suite_focus=true#step:6:498)

packages/story-editor/src/karma/fixture/fixture.js
BrittanyIRL: absolutely, 100% just a draft with minimal effort spent on clean up to focus on finding the root of the issues <3 <3

packages/story-editor/src/utils/loadStylesheet.js
BrittanyIRL: Fair! Added in with the config update

karma-story-editor.config.cjs
swissspidy: Ah, I see! Does something like `pattern: 'https://fonts.googleapis.com/css2*'` work as well?

https://karma-runner.github.io/6.3/config/files.html

packages/story-editor/src/karma/fixture/fixture.js
BrittanyIRL: so turns out once the links have proper ids they stop stacking. thanks for the add. I had just assumed it was intentionally left off, that's on me.

packages/story-editor/src/karma/fixture/fixture.js
BrittanyIRL: (same as other comments but incase this is read out of that context - once the ID is current the new roboto link is replacing the old so they don't append anymore, removing this whole chunk!)

packages/story-editor/src/app/font/actions/useLoadFontFiles.js
```diff
@@ -42,7 +42,12 @@ function useLoadFontFiles() {
 
     switch (service) {
       case 'fonts.google.com':
-        await loadStylesheet(getGoogleFontURL([{ family, variants }], 'auto'));
+        await loadStylesheet(
+          getGoogleFontURL([{ family, variants }], 'auto'),
+          elementId
+        ).catch(() => {
+          return;
+        });
```
swissspidy: Since we're using async/await here, we'd need to wrap this in a `try ... catch`. It's not a Promise where one could use `.catch()`.

```suggestion
				try {
          await loadStylesheet(
         	  getGoogleFontURL([{ family, variants }], 'auto'),
         	  elementId
          );
        } catch {
        	// Do nothing.
      	}
```

karma-story-editor.config.cjs
BrittanyIRL: looks like it needs to be a specific string. https://karma-runner.github.io/6.3/config/files.html#loading-files-from-another-server
> Globing is not support, so each URL must be specified as a separate pattern.

swissspidy: Hmm but the first sentence there says "Patterns _can_ be an absolute URL".

I understand it as:

* Use a string with an absolute URL. No glob supported.
* Use a `pattern`, glob supported

All the examples there indicate the same to me there.

Could you perhaps give it a try if you haven't already?

BrittanyIRL: i did try but i'll do it again and post the logs to double check

BrittanyIRL: hi hi, yup - forgot to escape the backslashes and also prettier was sneaking in what it thinks is better. we're good on this front now. thanks for pushing it.  ([logs](https://github.com/google/web-stories-wp/runs/4729597042?check_suite_focus=true#step:6:10234))

packages/story-editor/src/karma/fixture/fixture.js
```diff
@@ -230,9 +230,9 @@ export class Fixture {
     });
   }
 
-  restore() {
+  async restore() {
     window.location.hash = '#';
-    localStorage.clear();
+    await localStorage.clear();
```
swissspidy: This is not an async function: https://developer.mozilla.org/en-US/docs/Web/API/Storage/clear

packages/karma-puppeteer-launcher/src/index.cjs
```diff
@@ -50,22 +51,22 @@ function puppeteerBrowser(baseBrowserDecorator, config) {
         // See http://crbug.com/715363
         // We use this flag to work-around this issue.
         '--disable-dev-shm-usage',
+        '--source-map=false',
+        '--force-gpu-mem-discardable-limit-mb',
+        //  set memory allocated for browse https://github.com/karma-runner/karma-chrome-launcher/issues/154#issuecomment-569698577
+        //  no super sure this is current anymore
+        '--max-old-space-size=8192',
```
maxyinger: hmph what's the difference between this and what they have in their comment:
```
"--js-flags=--max-old-space-size=<size>"
```
maybe we should have the `--js-flags=` arg prefixing the `--max-old-space-size` flag.

after looking at chrome flags, it doesn't seem to be an argument listed [here](https://peter.sh/experiments/chromium-command-line-switches/) However it is found in the possible js-flags [here](https://chromium.googlesource.com/v8/v8/+/refs/heads/main/src/flags/flag-definitions.h#188) just gotta `ctrl + f` the term `max_old` to find it

maxyinger: also curious about how this flag affects things:
```
"--remote-debugging-port=9222", // Without a remote debugging port, Google Chrome exits immediately.
```

maxyinger: as far as max space vs old space, here's an [article](https://deepu.tech/memory-management-in-v8/) I just read a little bit of, mentions the `max-old-space-size` flag. makes sense since everything moves there that's survived 2 GC cycles, which I imagine is a lot of stuff by the later tests.

Also Addy Osmani has a great video on it memory in general. Starts going into the difference between new and old memory here: https://youtu.be/LaxbdIyBkL0?t=2341

packages/karma-puppeteer-launcher/src/index.cjs
```diff
@@ -40,6 +40,7 @@ function puppeteerBrowser(baseBrowserDecorator, config) {
       devtools: false,
       defaultViewport: null,
       snapshots: false,
+      waitForInitialPage: true,
```
maxyinger: I know we have a todo in our karma fixture `render` method that's supposed to await for the entire page to render:
https://github.com/google/web-stories-wp/blob/cc2fd32f90bc46b071192be43c93952c059a98de/packages/story-editor/src/karma/fixture/fixture.js#L401-L402

We can probably leverage a tactic I used in the saved template image generation here that just wraps a promise around a callback ref. will have to alter it a bit, but should work because react testing libraries `render` is just a wrapper around `ReactDOM.render` under the hood:
https://github.com/google/web-stories-wp/pull/9584/files#r781294665

maxyinger: also might be good to add the `cleanup()` method from react testing library to `fixture.restore()` just to cleanup all the nodes from the previous `fixture.render` call. I tried this locally and it doesn't fix our memory leak issue, but it may slow the bleeding a little bit

maxyinger (commented):
Looking great so far. Awesome findings! 

The `loadStyleSheet` update with the added `karma-story-editor.config.cjs` patterns for google fonts api are the two main things that I would think would have caused the break. Not seeing how a failed test would cause the whole process to exit.

Maybe prior to the `loadStyleSheet` update, the repetitive font loading was a big leak in memory. Not sure how large each of those would be to store, and if they were ever eligible for  garbage collection in the lifetime of our tests running, but could've been a big culprit to running out of memory 🤷 

Anyway, just posted most of my findings from looking into memory stuff last week as comments on here. If you want me to implement them in a separate PR, happy to do so as well.

packages/story-editor/src/app/font/actions/useLoadFontFiles.js
```diff
@@ -42,7 +42,14 @@ function useLoadFontFiles() {
 
     switch (service) {
       case 'fonts.google.com':
-        await loadStylesheet(getGoogleFontURL([{ family, variants }], 'auto'));
+        try {
```
BrittanyIRL: not convinced this is necessary. will clear this out and try again eventually.

packages/story-editor/src/components/dropTargets/karma/dropTarget.karma.js
```diff
@@ -107,7 +107,7 @@ describe('Drop-Target integration', () => {
       expect(bgElement.resource.baseColor).toEqual('#734727');
     });
 
-    it('should correctly handle image dragged from library straight to edge (no cached base color)', async () => {
+    xit('should correctly handle image dragged from library straight to edge (no cached base color)', async () => {
```
BrittanyIRL, line 112: todo: add issue to fix test

packages/story-editor/src/components/footer/gridview/karma/gridView.karma.js
```diff
@@ -201,7 +201,7 @@ describe('GridView integration', () => {
     expect(fixture.editor.gridView.close).toHaveFocus();
   });
 
-  it('should use "Esc" to exit the dialog', async () => {
+  xit('should use "Esc" to exit the dialog', async () => {
```
BrittanyIRL, line 206: todo: add issue to fix test

packages/karma-puppeteer-launcher/src/index.cjs
swissspidy: > also curious about how this flag affects things:

I think `karma-chrome-launcher` adds this by default if it's missing.

packages/story-editor/src/app/font/actions/useLoadFontFiles.js
```diff
@@ -42,7 +42,10 @@ function useLoadFontFiles() {
 
     switch (service) {
       case 'fonts.google.com':
-        await loadStylesheet(getGoogleFontURL([{ family, variants }], 'auto'));
+        await loadStylesheet(
+          getGoogleFontURL([{ family, variants }], 'auto'),
+          elementId
```
BrittanyIRL, line 47: Previously, without `elementId` getting passed here the id was getting set as 'undefined' which was making the stylesheets stack up with every new test set in `head` on karma runs which was contributing to early disconnect and slowing down run time.

packages/design-system/src/components/input/index.js
```diff
@@ -245,7 +245,7 @@ export const InputPropTypes = {
   onFocus: PropTypes.func,
   suffix: PropTypes.node,
   unit: PropTypes.string,
-  value: PropTypes.string.isRequired,
+  value: PropTypes.string,
```
BrittanyIRL, line 248: just silencing some noise, value can actually be null so this shouldn't be required.

BrittanyIRL:
@BrookeGraham @samwhale @littlemilkstudio  please add any notes or context that I missed from the last week that helped lead us here.

BrookeGraham:
One thing to note is the silent failures or false positives we had been getting. When the test suite would disconnect, the second pass only retries test that have already ran _and_ failed. This lead to many tests never being executed. Once the second pass finished we would get a green light, but in reality not every test was attempted. @littlemilkstudio has a pr #10143  up to remove the second pass on fail.

BrittanyIRL:
> One thing to note is the silent failures or false positives we had been getting. When the test suite would disconnect, the second pass only retries test that have already ran _and_ failed. This lead to many tests never being executed. Once the second pass finished we would get a green light, but in reality not every test was attempted. @littlemilkstudio has a pr #10143 up to remove the second pass on fail.

Great call out, an example of this would be: https://github.com/google/web-stories-wp/runs/4761414889?check_suite_focus=true#step:6:3953 

I'll add to the criteria of what we were fixing.

BrookeGraham (approved):
🐢 🎉 🌮

packages/story-editor/src/elements/media/videoImage.js
```diff
@@ -28,23 +28,25 @@ const Video = styled.video`
   object-fit: cover;
 `;
 
-function VideoImage({ alt, ...attrs }) {
+function VideoImage({ attrs }) {
```
swissspidy: Can you elaborate on the changes to this component?

miina (approved):
Thanks so much for looking into this! ⭐ 

(Don't know a lot about these Puppeteer settings but otherwise LGTM)

packages/karma-puppeteer-launcher/src/index.cjs
```diff
@@ -58,14 +63,7 @@ function puppeteerBrowser(baseBrowserDecorator, config) {
...
     // See https://github.com/puppeteer/puppeteer/blob/v3.0.4/docs/api.md#puppeteerlaunchoptions.
-    browser = await puppeteer.launch({
-      product: puppeteerOptions.product,
-      slowMo: puppeteerOptions.slowMo,
-      dumpio: puppeteerOptions.dumpio,
-      headless: puppeteerOptions.headless,
-      devtools: puppeteerOptions.devtools,
-      defaultViewport: puppeteerOptions.defaultViewport,
-    });
+    browser = await puppeteer.launch(puppeteerOptions);
```
swissspidy, line 65: I presume this is just for readability to prevent repetition?

packages/karma-puppeteer-launcher/src/index.cjs
swissspidy: FWIW this option is `true` by default.

https://github.com/puppeteer/puppeteer/blob/51e61696c1c20cc09bd4fc068ae1dfa259c41745/src/node/Launcher.ts#L85

packages/karma-puppeteer-launcher/src/index.cjs
```diff
@@ -50,6 +51,10 @@ function puppeteerBrowser(baseBrowserDecorator, config) {
         // See http://crbug.com/715363
         // We use this flag to work-around this issue.
         '--disable-dev-shm-usage',
+        // Added to prevent disconnect timeouts: sets the maximum GPU memory to use for discardable caches.
+        '--force-gpu-mem-discardable-limit-mb',
+        // https://stackoverflow.com/questions/67501093/passthrough-is-not-supported-gl-is-disabled
+        '--disable-software-rasterizer',
```
swissspidy, line 56: So this is just to get rid of some error message? Does the error crash the tests or just pollute the logs?

BrittanyIRL: just pollute the logs remotely, this seems to also maintain headless working when it is run locally. 

(edit) 
Actually, no this is the combo that gives us a success. It still feels like it's maybe not the entire issue so it feels vague and I don't know enough about the inner workings of chromium to properly grok this. But these two flags of disable software rasterizer and force-gpu-mem give us the ability to run through all of the tests.  Sometimes it seemed that just the force gpu was needed, but the disable software flag makes it consistent across local and remote which feels more stable - especially given we're def getting that gl passthrough error i've added a comment for above the arg.

packages/karma-puppeteer-launcher/src/index.cjs
BrittanyIRL: and to pass the list of `args` through on launch.

packages/story-editor/src/elements/media/videoImage.js
BrittanyIRL: by separating `alt` from `attrs` it's not getting the `alt` set, just the `title` which was giving us some proptype noise. By doing it this way we're still giving the title the `alt` value per usual but allowing `alt` to also be set through spreading the rest of the attrs.

packages/karma-puppeteer-launcher/src/index.cjs
maxyinger: ended up just adding the updates `fixture.render()` & `fixture.restore()` here:
https://github.com/google/web-stories-wp/pull/10143

packages/story-editor/src/elements/media/videoImage.js
swissspidy: But there is no `attrs` prop.

This is how the component is used:

https://github.com/google/web-stories-wp/blob/8464d30ce8ad02d85ffcd5759217920f4c90a5b6/packages/story-editor/src/elements/video/icon.js#L40

The only reason the prop types warning went away with this change is because the newly-added `attrs` prop isn't marked as required. But `attrs` does not exist, so `src` and `alt` are now always `undefined` in that component.

In short, this _creates_ a bug instead of _fixing_ one.

The correct fix would be to update fixture data so resources always contain an `alt`.

I believe #9944 fixes this, so let's defer prop types cleanup to that PR and undo this bug here.

maxyinger (approved):

packages/karma-puppeteer-launcher/src/index.cjs
swissspidy: Right, but there aren't really any new args to pass through that were previously missed, right?

And the new `waitForInitialPage` arg is redundant AFAICT because `true` is the default there anyway.

packages/story-editor/src/elements/media/videoImage.js
BrittanyIRL: sure, will undo

packages/karma-puppeteer-launcher/src/index.cjs
BrittanyIRL: the list of `args` was getting ignored on launch, we set args in `defaultPuppeteerOptions` and then `puppeteerOptions` throws those together with `config.puppeteer` overrides and then uses that new object to launch the browser.  

i'll remove waitForInitialPage cuz it is redundant (i hadn't wanted to run the tests again just for that 1 line of code to dry up)

swissspidy (approved):
Let's get this in!

Thanks everyone for getting this to work again!

Especially the newly added Puppeteer flags are quite the hotfix, so it'd be good to follow up with improvements to Karma / the app to prevent memory issues.

packages/karma-puppeteer-launcher/src/index.cjs
merapi: > hmph what's the difference between this and what they have in their comment:

CMIIW but `--max-old-space-size` is a V8 (node) flag, and something like `--disable-dev-shm-usage` is a Chromium flag, that's why you need to prefix the first one with `--js-flags=` so it will be passed down to V8.

packages/karma-puppeteer-launcher/src/index.cjs
merapi: I think mainly because it was not forwarding the `args` defined in `puppeteerOptions`.

karma-story-editor.config.cjs
```diff
@@ -68,9 +68,10 @@ module.exports = function (config) {
         served: true,
         nocache: false,
       },
+      // eslint-disable-next-line prettier/prettier, no-useless-escape
+      '^https:\/\/fonts.googleapis.com\/css*',
```
swissspidy, line 72: Seeing `Pattern "/home/runner/work/web-stories-wp/web-stories-wp/^https:/fonts.googleapis.com/css*" does not match any file.` in the Karma logs now

packages/design-system/src/components/input/index.js
swissspidy: This seems to _cause_ noise:

```
Warning: `value` prop on `%s` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.
```

Sagor2214: Reply

PR #10096, opened 2022-01-05
Disable Update button if there are no changes to be saved

### LINKED ISSUE #3548: Update button not disabled despite no changes
## Bug Description

In `assets/src/edit-story/components/header/buttons.js` there is some logic to disable the "Update" button if there are no new changes.

However, that does not seem to be working. The Update button is always visible & enabled, allowing me to update the story despite there being no new changes.

If there are no changes, the button should be disabled.

## Expected Behaviour

1. Publish story
1. Don't make any changes
1. The Update button is disabled

## Steps to Reproduce

1. Publish story
1. The Update button is not disabled
1. Click on Update
1. The Update button is still not disabled

## Screenshots

Screenshot 2020-07-30 at 10 29 30

https://github.com/google/web-stories-wp/blob/9f6927743e52c9dd37254e1b7c504a6144f13116/assets/src/edit-story/components/header/buttons.js#L248-L271

## Additional Context

- Plugin Version: `master`
- Operating System:
- Browser:

---

amovar18:
## Context

Before this PR the update button would stay enabled even though there werent new changes. It was because the update button was meant to be disabled only if there was any `uploading` or any `saving` action happening.

## Summary

This PR changes the condition and adds a unit test to check the update button works properly.

## Testing Instructions

1. Open a published story or any draft story in story-editor
2. Dont make any changes you should see that the Update button is disabled.

## Checklist

---

Fixes #3548


packages/story-editor/src/components/header/buttons/test/update.js
```diff
@@ -84,7 +84,7 @@ function arrange({
 }
 
 describe('UpdateButton', () => {
-  const FUTURE_DATE = '2022-01-01T20:20:20Z';
+  const FUTURE_DATE = '2022-01-10T20:20:20Z';
```
miina: Maybe change the year here instead? Seem like this might not be accurate quite soon.

amovar18: I dont think it would matter because `Mockdate` actually manipulates the `now` date so anytime test is run the system is at `2020-07-15T12:00:00+00:00`

miina: Ah, okay, why the change at all then?

amovar18: sorry forgot to remove thought it was the reason the test failed will revert the change

miina:
@amovar18 Saw you requested a review but the PR is still a draft. Is this final now and can be marked ready for review?
Could you also merge `main`?

swissspidy (approved):

PR #10100, opened 2022-01-05
Code Quality: Refactor `setOrCreateImage` function

### LINKED ISSUE #10025: Code Quality: Refactor `setOrCreateImage` function
## Feature Description

Currently the `setOrCreateImage` is a little messy and the code is unclear. It has the following issues. 

- It creates a hidden dom element when it doesn't need to. 
- It is not async code. 
- It passing functions again. 
- Needs better jsdocs.

spacedmonkey:
## Summary

Refactor logic of setOrCreateImage. Instead of creating an image tag and append it to body tag, just pass a image tag into the color thief. Remove lots of unneeded logic and generally make it easier to read. 

## User-facing changes

Screenshot 2022-01-10 at 18 02 00

## Testing Instructions

1. Create a story. 
2. Set the background of first page to a image of a solid color, like white. 
3. Create text element.
4. Add some text
5. Change text color the same color as background.
6. Publish story. 
7. Go to check. See accessibility warning. 

## Checklist

---

Fixes #10025


packages/story-editor/src/utils/getMediaBaseColor.js
```diff
@@ -20,41 +20,14 @@
...
-
 export async function getMediaBaseColor(src) {
   if (!src) {
     return Promise.reject(new Error('No source to image'));
   }
   let color;
   const trackTiming = getTimeTracker('load_get_base_color');
   try {
-    color = await setOrCreateImage({
+    color = await createImage({
```
miina: Would be nice to rename this to `createImageWithCallback` or something like that for making it more readable.

miina (approved):
LGTM!

Requesting review from Brittany as well though for the checklist part.

packages/story-editor/src/utils/getMediaBaseColor.js
spacedmonkey: I was going to originally change to that, but I thought others wouldn't like it. I will change it now.

packages/story-editor/src/utils/getMediaBaseColor.js
```diff
@@ -89,34 +62,32 @@ function getDefaultOnloadCallback(nodeKey, resolve, reject) {
...
-    if (!imgNode) {
-      imgNode = document.createElement('div');
-      imgNode.id = getImgNodeId(id);
-      Object.assign(imgNode.style, STYLES);
-      document.body.appendChild(imgNode);
-      document.body[NODE_KEY] = imgNode;
-    }
-    imgNode.innerHTML = '';
-
     const img = new Image();
```
spacedmonkey: Worth noting here, I would just try and use the `preloadImage` function here, but couldn't work out a way to do this cleanly.

packages/story-editor/src/utils/getMediaBaseColor.js
spacedmonkey: Fixed.

packages/story-editor/src/utils/getMediaBaseColor.js
swissspidy: What was the issue?

swissspidy:
Why can't we use promises instead of passing callbacks? 🤔

swissspidy:
Here's an idea for making this better readable: c908605d38626984151fb8dc13b1ebbca5897ae5

spacedmonkey:
@swissspidy Can you review again.

spacedmonkey:
In my testing, this made base color generation x2-x5 faster. Maybe because it comes on another thread now.

packages/story-editor/src/utils/getMediaBaseColor.js
```diff
@@ -19,104 +19,60 @@
...
-  return `__web-stories-bg-${elementId}`;
-}
+/**
+ * Extracts the base color from an image element.
+ *
+ * @param {Image} img Image.
+ * @return {Promise<string>} Hex color.
+ */
+async function extractColorFromImage(img) {
+  const { default: ColorThief } = await import(
```
miina, line 31: Is it a good idea to load a module with `await`? Why not load it async? It should also handle failing loading, too.

miina:
> In my testing, this made base color generation x2-x5 faster.

Nice! Did you record any before-after testing results as well?

packages/story-editor/src/utils/getMediaBaseColor.js
swissspidy: Async/await is just syntactic sugar really

spacedmonkey: I personally just find this more readable.

spacedmonkey:
> Nice! Did you record any before-after testing results as well?

Testing 

Image 1 - 170KB

Before - 512.7649999856949
After - 303.0049999952316

Image 2 -  10mb
Before - 668.6350000202656
Adter - 556.5399999916553

miina (approved):
Would still be good to have a review from someone from Pea pod as well for the checklist part.

packages/story-editor/src/components/checklist/checks/pageBackgroundLowTextContrast/check.js
```diff
@@ -181,35 +178,27 @@ function getTextImageBackgroundColor({ background, text, page }) {
...
  * @param {Element} arguments.bgImage The background element
  * @param {{ width: number, height: number }} arguments.bgBox The containing box of the background image - needed for calculating canvas translations for rotated elements
  * @param {{ x: number, y: number, width: number, height: number }} arguments.overlapBox The position and size of the text element relative to the scaled and roated background image
- * @return {Promise<RGB>} Resolves to the dominant color of the background image in the overlap box area
+ * @return {Promise<Object>} Resolves to the dominant color of the background image in the overlap box area
  */
-function getOverlapBgColor({ elementId, pageId, bgImage, bgBox, overlapBox }) {
-  function getOnloadCallback(nodeKey, resolve, reject) {
-    return () => {
+async function getOverlapBgColor({ bgImage, bgBox, overlapBox }) {
```
BrittanyIRL, line 198: @BrookeGraham  heads up on this: i think it might impact updates on your PR here https://github.com/google/web-stories-wp/pull/10074

packages/media/src/preloadImage.js
```diff
@@ -19,11 +19,13 @@
  *
  * @param {string} src Image source.
  * @param {string} [srcset] Image source set.
+ * @param {number|string} [width] Image width.
+ * @param {number|string} [height] Image height.
  * @return {Promise} Image object.
  */
-const preloadImage = (src, srcset = undefined) => {
+const preloadImage = (src, srcset = undefined, width, height) => {
```
BrittanyIRL: now that `preloadImage` has 4 params getting passed this would be more readable if it's an object: 
`const preloadImage = ({src, srcset = undefined, width, height}) => {` this helps me know why you're passing `undefined` on the background contrast checklist.

BrittanyIRL (changes requested):
So, it looks like this update is impacting the check on whether the background + text contrast is ok.  Did you test this and see it not regressing? 
Screen Shot 2022-01-10 at 9 18 10 AM

packages/media/src/preloadImage.js
spacedmonkey: Fixed,

spacedmonkey:
@BrittanyIRL How do I use this functionality? Are we sure this is working in main?

BrittanyIRL (approved):
Discussed with @spacedmonkey who confirmed that contrast check is working as expected for him, so this looks good to me.

packages/story-editor/src/components/checklist/checks/pageBackgroundLowTextContrast/check.js
BrookeGraham: Thank you! I'll check it out.

BrittanyIRL:
> 

Just to keep this thread updated vs a slack DM: 
Right now in staging there is a check for contrast between background color and foreground text color. You can verify this by having a page with a background color and then creating a text element that is that same shade or a nearby shade. It shows up in the 'accessibility' section, so you'll need to have > 3 pages present for that check to show up in the checklist.

BrookeGraham (approved):
📸 LGTM 😀

PR #10101, opened 2022-01-05
E2E Tests: Run tests against WP 5.9 RC1

spacedmonkey:
## Testing Instructions

1.

## Checklist

---

Fixes #


swissspidy:
@spacedmonkey Tests against RC 1 are failing now because of this change in core: https://core.trac.wordpress.org/ticket/54181

In `trashAllTerms()`, `#bulk-action-selector-top` is used, which in WP 5.9+ won't be displayed anymore if there are no terms.

So we need to make it a bit more robust.

packages/e2e-test-utils/src/visitBlockWidgetScreen.js
```diff
@@ -27,17 +27,23 @@ async function visitBlockWidgetScreen() {
   await visitAdminPage('widgets.php');
 
   // Disable welcome guide if it is enabled.
-  const isWelcomeGuideActive = await page.evaluate(() =>
+  const isWelcomeGuideActive = await page.evaluate(() => {
+    if (!wp.data.dispatch('core/edit-widgets').__unstableIsFeatureActive) {
```
swissspidy: I think the reason this is not available in 5.9 is because the Twenty Twenty-Two theme doesn't have any widgets support because of Full Site Editing. You'll get this error page:

Screenshot 2022-01-06 at 15 28 00

It would be good to test for this accordingly or at least add an inline comment with this explanation.

packages/e2e-tests/src/config/bootstrap.js
```diff
@@ -89,6 +89,13 @@ const ALLOWED_ERROR_MESSAGES = [
 
   // Another Firefox warning.
   'Layout was forced before the page was fully loaded',
+
+  // Expected css error.
+  'Stylesheet twentytwenty-block-editor-styles-css was not properly added.',
```
swissspidy: Sounds like a bug that should reported in core.

swissspidy:
The `Web Stories Widget Block` e2e test fails because the Twenty-Twenty Two theme does not support widgets.

We might want to consider installing/activating Twenty Twenty-One by default in `bin/local-env/install-wordpress.sh` and then have dedicated E2E tests for block-based themes.

packages/e2e-tests/src/config/bootstrap.js
spacedmonkey: It seems to be a known issue. https://github.com/WordPress/gutenberg/pull/28165/files#r667483229

packages/e2e-test-utils/src/visitBlockWidgetScreen.js
spacedmonkey: But we don't use 2022 theme in e2e tests. [See](https://github.com/google/web-stories-wp/blob/58f8648bc2df6a53c91db89d8a434cf25965a39b/bin/local-env/install-wordpress.sh#L152).

spacedmonkey: These unstable data entries were added here https://github.com/WordPress/gutenberg/pull/31925.
Removed in https://github.com/WordPress/gutenberg/pull/33774/

spacedmonkey:
> The Web Stories Widget Block e2e test fails because the Twenty-Twenty Two theme does not support widgets.

How is that possible, we are using 2020 for e2e test ain't we?

packages/e2e-tests/src/config/bootstrap.js
swissspidy: Nice find! That explanation should be added as an inline comment.

I wouldn't consider an old comment on a merged PR to be a well-known issue though.

Do you know during which test this error is happening?

Given that `wp-admin/widgets.php` and `wp-admin/customize.php` don't show widgets for that theme, so it's not clear to me how this would happen.

If we know where this happens, this can be reported in core.

swissspidy:
Ah, you're right. Then I suppose some selectors have simple changed.

`.block-editor-inserter__search-input` no longer exists.

Maybe `.block-editor-inserter__search input` works.

packages/e2e-test-utils/src/visitBlockWidgetScreen.js
swissspidy: Nice find!

swissspidy:
I am converting this to a draft simply to avoid generating unnecessary Percy snapshots while this is WIP.

packages/e2e-tests/src/config/bootstrap.js
spacedmonkey: Sorry, the comment isn't very good. It was just there for debugging. I will add a better comment.

packages/e2e-tests/src/config/bootstrap.js
```diff
@@ -89,6 +89,13 @@ const ALLOWED_ERROR_MESSAGES = [
 
   // Another Firefox warning.
   'Layout was forced before the page was fully loaded',
+
+  // Expected css error.
```
spacedmonkey:
```suggestion
  // Upsteam issue in gutenberg and twentytwenty theme. 
```

swissspidy:
There are lint errors

spacedmonkey:
> There are lint errors

Fixed.

packages/e2e-tests/src/config/bootstrap.js
```diff
@@ -89,6 +89,13 @@ const ALLOWED_ERROR_MESSAGES = [
 
   // Another Firefox warning.
   'Layout was forced before the page was fully loaded',
+
+  // Upsteam issue in gutenberg and twentytwenty theme.
+  'Stylesheet twentytwenty-block-editor-styles-css was not properly added.',
+
+  // deprecated notice in 5.9.
+  // @todo fix issue.
```
swissspidy:
```suggestion
  // TODO(#9240): Fix usage in the web stories block.
```

packages/e2e-tests/src/specs/wordpress/blockWidget.js
```diff
@@ -40,7 +40,7 @@ describe('Web Stories Widget Block', () => {
   it('should insert a new web stories block', async () => {
     await visitBlockWidgetScreen();
     await expect(page).toClick('button[aria-label="Add block"]');
-    await page.type('.block-editor-inserter__search-input', 'Web Stories');
+    await page.type('input[placeholder="Search"]', 'Web Stories');
```
swissspidy, line 43: 👍

packages/e2e-test-utils/src/visitBlockWidgetScreen.js
```diff
@@ -24,20 +24,34 @@ import visitAdminPage from './visitAdminPage';
...
+    // TODO Change after 5.9 release.
+    if ('latest' === version) {
+      return wp.data
+        .select('core/edit-widgets')
+        .__unstableIsFeatureActive('welcomeGuide');
+    }
+    return wp.data
+      .select('core/interface')
+      .isFeatureActive('core/edit-widgets', 'welcomeGuide');
+  }, WPVersion);
```
swissspidy: Is the version check really needed? Didn't it work the way you had it before, by checking whether the functions exist?

spacedmonkey: Sadly it didn't work the way I had it before. See https://github.com/google/web-stories-wp/runs/4728644759?check_suite_focus=true#step:11:11

spacedmonkey: I would love another way around doing this check.

swissspidy: Hmm sounds like all that is needed is some conditional check, perhaps easiest with optional chaining (`?.`) ,to avoid the `isFeatureActive is not a function` error.

```js
// Disable welcome guide if it is enabled.

// The former selector is for WP < 5.9
const isWelcomeGuideActive = await page.evaluate(() => {
  return wp.data.select('core/edit-widgets')?.__unstableIsFeatureActive?.('welcomeGuide') ||
  wp.data.select( 'core/interface' )?.isFeatureActive?.( 'core/edit-widgets', 'welcomeGuide' );
});

if (isWelcomeGuideActive) {
  // The former action is for WP < 5.9
  await page.evaluate(() => {
    wp.data.dispatch('core/edit-widgets')?.__unstableToggleFeature?.('welcomeGuide');
    wp.data.dispatch( 'core/interface' )?.toggleFeature?.( 'core/edit-widgets', 'welcomeGuide' );
  });
}
```

swissspidy (approved):

PR #10102, opened 2022-01-05
Try: karma font fix

BrookeGraham:
## Testing Instructions

1.

## Checklist

---

Fixes #

### DIFF
--- a/karma-story-editor.config.cjs
+++ b/karma-story-editor.config.cjs
@@ -48,6 +48,7 @@ module.exports = function (config) {
       'karma-sourcemap-loader',
       'karma-webpack',
       'karma-coverage-istanbul-reporter',
+      'karma-verbose-reporter',
       require('@web-stories-wp/karma-puppeteer-launcher'),
       require('@web-stories-wp/karma-puppeteer-client'),
       require('@web-stories-wp/karma-failed-tests-reporter'),
@@ -100,7 +101,8 @@ module.exports = function (config) {
     // possible values: 'dots', 'progress'
     // available reporters: https://npmjs.org/browse/keyword/karma-reporter
     reporters: [
-      'dots',
+      'progress',
+      // 'verbose',
       '@web-stories-wp/karma-failed-tests-reporter',
       config.coverage && 'coverage-istanbul',
     ].filter(Boolean),
@@ -167,7 +169,7 @@ module.exports = function (config) {
 
     // Prevent duplicate logging to console
     browserConsoleLogOptions: {
-      terminal: false,
+      terminal: true,
     },
 
     // When a browser crashes,try to relaunch more than just 2 times (which is the default)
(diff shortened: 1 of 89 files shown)


packages/story-editor/src/storyEditor.js
```diff
@@ -64,16 +64,16 @@ function StoryEditor({ config, initialEdits, children }) {
...
-                  <HistoryProvider size={50}>
-                    <SnackbarProvider>
-                      <StoryProvider
-                        storyId={storyId}
-                        initialEdits={initialEdits}
-                      >
-                        <TaxonomyProvider>
-                          <CurrentUserProvider>
-                            <FontProvider>
+                <FontProvider>
```
BrookeGraham, line 67: Moved the font provider up to where the old `FileProvider` was.

karma-story-editor.config.cjs
```diff
@@ -145,6 +174,8 @@ module.exports = function (config) {
       runInParent: true,
     },
 
+    crossOriginAttribute: false,
```
BrookeGraham: > Description: When true, this will append the crossorigin attribute to generated script tags, which enables better error reporting for JavaScript files served from a different origin. Disable this when you need to load external scripts that are served without the necessary Access-Control-Allow-Origin header.

packages/story-editor/src/karma/fixture/fixture.js
```diff
@@ -233,6 +233,15 @@ export class Fixture {
...
     localStorage.clear();
+    console.log('🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥🎥');
+    console.log(Array.from(document.getElementsByTagName('video')).length);
+
+    const videos = Array.from(document.getElementsByTagName('video'));
+    if (videos.length) {
+      videos.map((video) => video.parentNode.removeChild(video));
+    }
+    console.log('🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻🕵🏻');
+    console.log(Array.from(document.getElementsByTagName('video')).length);
```
swissspidy: Sounds like you're trying to fix those video player messages?

Check out #9944 for context, especially the discussion at https://github.com/google/web-stories-wp/pull/9944#discussion_r777486850


It should not cause test failures; it's just noisy.

BrookeGraham: Yeah at this point we are unsure where the karma test suite disconnect is occurring. There was a slim hope that the memory leak from these would cause the thread to run out of memory and thus diconnect. But this isn't the case. Seeing as the webMediaPlayer not being garbage collected is an issue with Chrome, I am not sure we can really solve this one. Max is looking into it though.

PR #10119, opened 2022-01-06
Decoupling: Move WordPress-specific style constants to integration packages.

mohdsayed:
## Context

Constants defined in `design-system/src/theme/constants/wpAdmin.js` are WordPress specific which would add additional spacing from the left and top of the popup background in the standalone editor and also have an effect on other components including the dashboard UI.

image

image

## Summary

This PR moves those constants in `wp-story-editor` and `wp-dasboard` and passes it back using the `config` prop.

## Testing Instructions

1.

## Checklist

---


swissspidy:
That's quite a lot of changes/complexity just for positioning some items.

I'd love some insights on how this originally came to be. cc @BrittanyIRL @littlemilkstudio 

I wonder whether this couldn't be simplified by detecting offsets automatically or perhaps reduce the number of places where these numbers are used.

BrittanyIRL:
> 

The history of those constants is that it was in the dashboard and editor separately for modal positioning, so when the modal got moved to the design system the constants were moved as well to prevent duplication. More historical than anything at this point. It's fine to change, there just wasn't a need for it earlier.

swissspidy:
Any ideas on how we could perhaps get rid of these constants altogether then?

BrittanyIRL:
> Any ideas on how we could perhaps get rid of these constants altogether then?

happy to take a closer look next week, right now i'm pretty zeroed in on karma which is blocking everything.

BrittanyIRL:
> > Any ideas on how we could perhaps get rid of these constants altogether then?
> 
> happy to take a closer look next week, right now i'm pretty zeroed in on karma which is blocking everything.

Ok, sorry for the delay. What about something like this? https://github.com/google/web-stories-wp/pull/10173

swissspidy:
Interesting approach!

My worry there is the tight coupling. This PR just passes offsets to the editor/dashboard and lets it do with it what's needed, whereas #10173 targets specific elements within the app to style them as needed. If the DOM structure in the app ever changes (or if we implement some new thing like `AwsesomePopup` that needs to know about offsets), we need to update the `wp-*` packages as well. And if we forget that, these two could easily get out of sync.

Somehow my thinking here was to use something more like this PR, but without the need for `TOOLBAR_HEIGHT, MENU_WIDTH, MENU_FOLDED_WIDTH ` etc. Perhaps by using a `useEffect` to get the offsets once.
But then again, the menu can be folded, so that would require an observer/listener.

So I suppose this current PR is cleaner in that regard 🤔 

I'll wait for Sayed to chime in when he's back next week

BrittanyIRL:
> Interesting approach!
> 
> My worry there is the tight coupling. This PR just passes offsets to the editor/dashboard and lets it do with it what's needed, whereas #10173 targets specific elements within the app to style them as needed. If the DOM structure in the app ever changes (or if we implement some new thing like `AwsesomePopup` that needs to know about offsets), we need to update the `wp-*` packages as well. And if we forget that, these two could easily get out of sync.
> 
> Somehow my thinking here was to use something more like this PR, but without the need for `TOOLBAR_HEIGHT, MENU_WIDTH, MENU_FOLDED_WIDTH ` etc. Perhaps by using a `useEffect` to get the offsets once. But then again, the menu can be folded, so that would require an observer/listener.
> 
> So I suppose this current PR is cleaner in that regard 🤔
> 
> I'll wait for Sayed to chime in when he's back next week

yeah, it's kinda a tough between two rocks type thing because there's really only 3 things that need to have this - sticky headers and popups/modals - so how much of the editor/dashboard (void of WordPress) do we want to cater to WordPress. I know in themes and other plugins it's not unheard of to have to target specifics like this (not saying it's good), I'm just wondering what the trade off is. 

That said, this PR that Sayed has does give the most communication and update to what's happening and why so it's probably the best option. I just wish we didn't need it at all. 

This solution also gives a path forward for other tweaks we might need when working in other CMS as well. So I think I'm in favor of what's already here.

mohdsayed:
> Somehow my thinking here was to use something more like this PR, but without the need for TOOLBAR_HEIGHT, MENU_WIDTH, MENU_FOLDED_WIDTH etc. Perhaps by using a useEffect to get the offsets once.
But then again, the menu can be folded, so that would require an observer/listener.

@swissspidy That would probably be the only other way and it was okay to get the offset once in some `useEffect` but I think using an observer would perhaps be an overkill ( like you rightly said initially that it's quite a lot of complexity just for positioning some items. ).

mohdsayed:
My other argument for not using an observer would be that if we think of the editor/dashboard in a decoupled context, collapsable sidebar/menu is generally a Wordpress thing, the core packages should just provide a way to add support for it 
from the integration packages instead of handling it itself.

swissspidy:
> I just wish we didn't need it at all.

I think we all agree on this :-)

Let's go with the current solution then for now 👍 

@sayedtaqui FYI there are some failing unit tests now.

packages/dashboard/src/components/pageStructure/leftRail.js
```diff
@@ -86,7 +86,12 @@ function getNewTemplatesMetaData(metaDataEntries, days) {
 
 function LeftRail() {
   const { state } = useRouteHistory();
-  const { newStoryURL, version, leftRailSecondaryNavigation } = useConfig();
+  const {
+    newStoryURL,
+    version,
+    leftRailSecondaryNavigation,
+    styleConstants: { topOffset },
```
BrittanyIRL, line 93: this'll need to get updated for the unit test set up of leftRail (or give a default, that might also work). 
Screen Shot 2022-01-18 at 4 25 04 PM

mohdsayed: Thanks

swissspidy (approved):

PR #10122, opened 2022-01-07
Media: Round width and height of images to ints.

spacedmonkey:
## Summary

In the media library, the value of height and width can be floats. This is invalid and can result in large height / width attributes, with lots of decimals. 

#### Before 
Screenshot 2022-01-07 at 08 51 29

#### After
Screenshot 2022-01-07 at 08 50 41

 

## Testing Instructions

1.

## Checklist

---

Fixes #

### DIFF
--- a/packages/story-editor/src/components/library/panes/media/common/mediaElement.js
+++ b/packages/story-editor/src/components/library/panes/media/common/mediaElement.js
@@ -103,8 +103,8 @@ function Element({
   const rawWidth = requestedWidth || requestedHeight / oRatio;
   const rawHeight = requestedHeight || rawWidth / oRatio;
 
-  const width = parseInt(rawWidth);
-  const height = parseInt(rawHeight);
+  const width = Math.round(rawWidth);
+  const height = Math.round(rawHeight);
 
   const mediaElement = useRef();
   const [showVideoDetail, setShowVideoDetail] = useState(true);
--- a/packages/story-editor/src/components/library/panes/media/common/mediaElement.js
+++ b/packages/story-editor/src/components/library/panes/media/common/mediaElement.js
@@ -100,8 +100,11 @@ function Element({
 
   const oRatio =
     originalWidth && originalHeight ? originalWidth / originalHeight : 1;
-  const width = requestedWidth || requestedHeight / oRatio;
-  const height = requestedHeight || width / oRatio;
+  const rawWidth = requestedWidth || requestedHeight / oRatio;
+  const rawHeight = requestedHeight || rawWidth / oRatio;
+
+  const width = parseInt(rawWidth);
+  const height = parseInt(rawHeight);
 
   const mediaElement = useRef();
   const [showVideoDetail, setShowVideoDetail] = useState(true);


packages/story-editor/src/components/library/panes/media/common/mediaElement.js
```diff
@@ -100,8 +100,11 @@ function Element({
 
   const oRatio =
     originalWidth && originalHeight ? originalWidth / originalHeight : 1;
-  const width = requestedWidth || requestedHeight / oRatio;
-  const height = requestedHeight || width / oRatio;
+  const rawWidth = requestedWidth || requestedHeight / oRatio;
+  const rawHeight = requestedHeight || rawWidth / oRatio;
+
+  const width = parseInt(rawWidth);
+  const height = parseInt(rawHeight);
```
spacedmonkey:
```suggestion
  const width = Math.round(rawWidth);
  const height = Math.round(rawHeight);
```

swissspidy:
Is there any actual bug this addresses?

AFAIK the spec technically says dimensions should be integers, but browsers don‘t actually bother. So it shouldn‘t make a difference. If it ain‘t broke …

spacedmonkey:
Rounding this value has number of benefits. 

1. Rounded values, means less markup to render. 
2. Width is used to get nearest sized image. The rounded version, is easier to compare and more likely to find a matching width. 


It is just something that bugs me every time as well. Willing to close it, but it is a small tweak.

spacedmonkey:
I also considered using sourcesets here as well..

swissspidy:
That‘s just the displayed width, no? Not related to fetching the nearest size.

Don‘t think we need to worry about saving a few bytes by not using decimals.

spacedmonkey:
> That‘s just the displayed width, no? Not related to fetching the nearest size.

What isn't true. 

https://github.com/google/web-stories-wp/blob/cc2fd32f90bc46b071192be43c93952c059a98de/packages/story-editor/src/components/library/panes/media/common/innerElement.js#L158

https://github.com/google/web-stories-wp/blob/cc2fd32f90bc46b071192be43c93952c059a98de/packages/story-editor/src/components/library/panes/media/common/innerElement.js#L103

https://github.com/google/web-stories-wp/blob/cc2fd32f90bc46b071192be43c93952c059a98de/packages/story-editor/src/components/library/panes/media/common/mediaElement.js#L232

https://github.com/google/web-stories-wp/blob/cc2fd32f90bc46b071192be43c93952c059a98de/packages/story-editor/src/components/library/panes/media/common/mediaElement.js#L105

swissspidy:
Hmm then I'd be more inclined to add `Math.round()` within `getSmallestUrlForWidth()`

Then we'd pick the 150px version if the `width` is 150.1111 or something, which is acceptable, but don't unnecessarily round the width for display.

swissspidy:
Closing for now as I don't think this is worth it.

Willing to discuss a change related to `getSmallestUrlForWidth` as mentioned above in a dedicated ticket

PR #10125, opened 2022-01-07
Media: Fix video drag an drop upload in Safari

### LINKED ISSUE #10098: Media: Drag and drop video in safari not working
## Bug Description

Dragging and dropping a video onto canvas in Safari does not upload the video. However dragging image into canvas does work. 

## Expected Behaviour

On dragging video into canvas, it should upload. 

## Screenshots

https://user-images.githubusercontent.com/237508/148226381-9693c062-a934-4bb1-97e7-30a6271df492.mov

## Additional Context

- Plugin Version: 1.15.1
- WordPress Version: 5.9-RC1
- Operating System: Mac OS 11.6
- Browser: 15.1

spacedmonkey:
## Summary

This change fixes the issue with video upload from safari. Change logic in seekVideo, use seeking and seeked events. canPlay is not fired in safari. 

## Testing Instructions

1. In safari. 
2. Open editor. 
3. Drag video into canvas. 
4. See it upload with preview. 

## Checklist

---

Fixes #10098


swissspidy:
There's `preloadVideoMetadata` that should be used instead

packages/story-editor/src/app/media/utils/getResourceFromLocalFile.js
```diff
@@ -76,17 +75,25 @@ const getVideoResource = async (file) => {
...
 
-  const videoEl = await preloadVideo(src);
-  const canPlayVideo = '' !== videoEl.canPlayType(mimeType);
+  const videoEl = document.createElement('video');
+  videoEl.src = src;
+  videoEl.addEventListener('loadedmetadata', () => {
+    const lengthData = getVideoLength(videoEl);
+    length = lengthData.length;
+    lengthFormatted = lengthData.lengthFormatted;
+  });
```
swissspidy: This won't work because it's not in a promise. There's gonna be race conditions.

Hence I'd highly recommend using `preloadVideoMetaData` instead.

That said, I think this uses `preloadVideo` for a reason.

Need to ensure that the length information and the `canPlayVideo` value, as well as the `hasAudio` stuff below still works as expected.

spacedmonkey: But it was work.. Why does it have to be a promise here? We are loading a resource here, it is a source url. Why did it work before ( before it was a promise ) and why does it work?

swissspidy: Then it was wrong before. It won't work **_reliably_** due to a possible race condition. If the `loadedmetadata` event fires only after 10 seconds the whole code below will have been executed already. Hence the need for `preloadVideoMetadata`

swissspidy:
Need to ensure that the length information and the `canPlayVideo` value, as well as the `hasAudio` stuff below still works as expected.

Have you tested that in multiple browsers?

I do recall that this uses `preloadVideo` on purpose, to ensure as much information as possible (length, audio, can play, poster) is available in all browsers for a variety of video types. I just want to prevent regressions.

spacedmonkey:
> Have you tested that in multiple browsers?

I did some testing on Friday. It seemed to work. However, doing more testing today it doesn't seem to work. 

> Then it was wrong before. 

I was unaware this. Can you explain exactly why this NEEDS to asyncly load?

swissspidy:
> I was unaware this. Can you explain exactly why this NEEDS to asyncly load?

What I mean is that this code here does not work reliably:

```js
   let length;
  let lengthFormatted;

  const videoEl = document.createElement('video');
  videoEl.src = src;
  videoEl.addEventListener('loadedmetadata', () => {
    const lengthData = getVideoLength(videoEl);
    length = lengthData.length;
    lengthFormatted = lengthData.lengthFormatted;
  });

  return length; // This will still be undefined because there is no guarantee `loadedmetadata` has fired yet.
```

That's why this needs to be a Promise, which is why I asked you to change it.

swissspidy:
Just making this a draft to avoid creating Percy snapshots unnecessarily while this is being worked on

packages/media/src/seekVideo.js
```diff
@@ -29,19 +29,15 @@ function seekVideo(video, offset = 0.99) {
...
+    // If the seek takes longer 3 seconds, guess it timed out and error out.
+    video.addEventListener('seeking', (evt) => {
       const wait = setTimeout(() => {
         clearTimeout(wait);
         reject(evt);
       }, THREE_SECONDS);
     });
     video.addEventListener('error', reject);
-    video.addEventListener('canplay', () => resolve(video), { once: true });
+    video.addEventListener('seeked', () => resolve(video), { once: true });
```
spacedmonkey, line 40: For reference https://github.com/google/web-stories-wp/blob/b051c770f668a88bd636787c19337e059349d44a/packages/media/src/preloadVideo.js#L32

This is how it used to be.

swissspidy:
`seekVideo()` is used in lots of places, e.g. when taking frames from a video in the video trimming screen, and when doing audio detection. Do these still work as expected, even in Safari?

spacedmonkey:
> seekVideo() is used in lots of places, e.g. when taking frames from a video in the video trimming screen, and when doing audio detection. Do these still work as expected, even in Safari?

Worked well in my testing, I tested, chroumium, safari, firefox and chrome. This also fixes issues with chroumium that can not play mp4s.

packages/media/src/seekVideo.js
```diff
@@ -29,19 +29,15 @@ function seekVideo(video, offset = 0.99) {
...
-    // desired to reject on "suspend" straight away.
-    // Instead, delay rejection in the hopes that the "canplay" event fires
-    // in the meantime.
-    video.addEventListener('suspend', (evt) => {
+    // If the seek takes longer 3 seconds, guess it timed out and error out.
+    video.addEventListener('seeking', (evt) => {
       const wait = setTimeout(() => {
         clearTimeout(wait);
         reject(evt);
       }, THREE_SECONDS);
```
spacedmonkey, line 41: @swissspidy why was this 3 seconds in the first place. Now it based on seeking, I would if this could fail quickly.

swissspidy: The inline comment explained it.

The suspend event is potentially fired multiple times, even if the browser eventually loads the video.

It rejects the error if the video was „properly“ suspended and the browser really never loaded it. In my testing 3s was a reasonable time to wait.

What you have now is listening to the seeking event, which is immediately fired when the browser starts seeking, to start the timeout.
That seems odd to me. With this, I don‘t think it makes sense to set the timeout in there. Have you tried moving that out of there? No need to listen for seeking at all.

spacedmonkey: Wasn't the question I asked here, why was 3 seconds? Just a random number?

Should I just remove this?

swissspidy: > In my testing 3s was a reasonable time to wait.

swissspidy: The `suspend` listener was added because `canplay` potentially never fired. Are you certain that `seeked` is always fired? I am a bit skeptical.

Need to ensure there's proper handling for when seeking doesn't happen.

You mentioned `stalled` and `aborted` on Slack as events to potential listen for.

Maybe you need to give that a try.

spacedmonkey: Seeked event would always fire on success. See [docs](https://html.spec.whatwg.org/multipage/media.html#event-media-seeked). 

Hooking into seeking, makes sense as once the seek start, there is then 3 seconds for it to be fulfilled or it will be rejected.

spacedmonkey:
After some more testing, here are my findings. 

1. Adding a promise reject on abort, does not make sense. This event is only fired when a user abort a video loading. 
2. Adding a promise reject on shalled might be problem, as I was seeing shalled events safari, even when video worked. 
3. Seek events, always fire when the currentTime is changed and success.
4. Keeping a timeout to seekVideo, to be overly careful and ensure that seekVideo will always timeout and error out. 

The PR works well in it's current form, let's leave it as it is.

swissspidy (approved):
If this doesn't cause any regressions then I guess it's fine

PR #10131, opened 2022-01-10
Karma tests debugging

merapi:
## Testing Instructions

1.

## Checklist

---

Fixes #

### DIFF
--- a/packages/karma-puppeteer-launcher/src/index.cjs
+++ b/packages/karma-puppeteer-launcher/src/index.cjs
@@ -50,9 +50,7 @@ function puppeteerBrowser(baseBrowserDecorator, config) {
         // See http://crbug.com/715363
         // We use this flag to work-around this issue.
         '--disable-dev-shm-usage',
-        // Remove: Blocked attempt to create a WebMediaPlayer as there are too many WebMediaPlayers already in existence. See crbug.com/1144736#c27
-        '--max-web-media-player-count=20',
-        '--devtools-flags=disable',
+        '--force-gpu-mem-discardable-limit-mb',
       ],
     };
     const puppeteerOptions = {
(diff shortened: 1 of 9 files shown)


swissspidy:
Just converting to draft to not unnecessarily create Percy snapshots and the like

packages/karma-puppeteer-launcher/src/index.cjs
```diff
@@ -50,6 +50,7 @@ function puppeteerBrowser(baseBrowserDecorator, config) {
         // See http://crbug.com/715363
         // We use this flag to work-around this issue.
         '--disable-dev-shm-usage',
+        '--force-gpu-mem-discardable-limit-mb',
```
swissspidy: This also seems to help in Brittany's PR, #10095

merapi: I wanted to check what makes the difference, flags or disabled tests/other fixes.
Looks like this flag alone doesn't do much: https://github.com/google/web-stories-wp/runs/4774876792?check_suite_focus=true
I'll try the second.

merapi: Locally I removed both flags from #10095 and it's still ✅ 
image
I will just check it on CI and then look into disabled tests/other fixes.

swissspidy: Running tests locally hasn't been an issue AFAIK, only running on CI

merapi: I've had a few buffer/memory-related messages locally, like this one from the latest CI run (with those 2 flags):

image

So far it looks like the flags are not an instant fix, but they might lower the overall amount of used memory/buffer size produced.

That `268435456` is https://source.chromium.org/chromium/chromium/src/+/main:content/browser/devtools/devtools_http_handler.cc;l=83

PR #10143, opened 2022-01-10
Integration Tests: Remove Retry-Failed 

maxyinger:
## Context
Part of looking into/fixing stability of karma tests. tThis PR doesn't fix them, but helps with some brittleness.

## Summary
(Thanks to @samwhale for pointing this out)Turns out retrying failed tests can lead to a lot of false positives in our integration tests. This issue happens when we've exceeded memory limits and the process gets killed. Basically looks like this:
- run karma tests
- say were 100 tests into 400 tests and we've had 1 failure so far
- we then exceed memory limits and the process gets killed
- retry-failed runs and only tries to run the one failed test
- the one failed test passes
- Integration tests say they passed, however 300 integration tests were skipped

Other things I've done in this PR:
- Added `cleanup()` method to `fixture.restore`. Doesn't fix memory leaks, but should hopefully prevent a possible source
- Removed a `TODO` in the render method and added a small implementation there to make sure the story is fully rendered before `fixture.render` resolves. (I don't think this was much of an issue prior, but shouldn't hurt anything)
- Just added the increased old memory limit for v8. pulled off of this PR which @BrittanyIRL is currently working on and should hopefully unbreak our integration tests: https://github.com/google/web-stories-wp/pull/10095

## User-facing changes
NA

## Testing Instructions
NA

1.

## Checklist

---

Fixes #

### DIFF
--- a/packages/story-editor/src/karma/fixture/fixture.js
+++ b/packages/story-editor/src/karma/fixture/fixture.js
@@ -26,6 +26,7 @@ import {
   act,
   screen,
   waitFor,
+  cleanup,
 } from '@testing-library/react';
 import { setAppElement } from '@web-stories-wp/design-system';
 import { FixtureEvents } from '@web-stories-wp/karma-fixture';
@@ -233,6 +234,10 @@ export class Fixture {
   restore() {
     window.location.hash = '#';
     localStorage.clear();
+    cleanup();
+    this._container = null;
+    this._screen = null;
+    this._editor = null;
   }
 
   get container() {
@@ -330,7 +335,7 @@ export class Fixture {
     setAppElement(root);
 
     const { container, getByRole } = render(
-      <StoryEditor key={Math.random()} config={this._config}>
+      <StoryEditor config={this._config}>
         <Layout
           header={<HeaderLayout />}
           footer={{
@@ -354,6 +359,7 @@ export class Fixture {
         container: root,
       }
     );
+
     // The editor should always be given 100%:100% size. The testing-library
     // renders an extra container so it should be given the same size.
     container.style.width = '100%';
@@ -398,8 +404,16 @@ export class Fixture {
       });
     });
 
-    // @todo: find a stable way to wait for the story to fully render. Can be
-    // implemented via `waitFor`.
+    await waitFor(() => {
+      if (
+        this._editor?.canvas?.displayLayer?.node &&
+        this._editor?.canvas?.framesLayer?.node
+      ) {
+        return;
+      }
+
+      throw new Error('Not Ready: Story not fully rendered');
+    });
   }
 
   /**
--- a/.github/workflows/tests-karma-dashboard.yml
+++ b/.github/workflows/tests-karma-dashboard.yml
@@ -62,7 +62,7 @@ jobs:
         run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
 
       - name: Run integration tests
-        run: npm run test:karma:dashboard -- --headless --viewport=1600:1000 --coverage || npm run test:karma:dashboard:retry-failed -- --headless --viewport=1600:1000
+        run: npm run test:karma:dashboard -- --headless --viewport=1600:1000 --coverage
         env:
           DISABLE_ERROR_BOUNDARIES: true
 
--- a/.github/workflows/tests-karma-editor.yml
+++ b/.github/workflows/tests-karma-editor.yml
@@ -67,7 +67,7 @@ jobs:
         run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
 
       - name: Run integration tests
-        run: npm run test:karma:story-editor -- --headless --snapshots --viewport=1600:1000 --coverage || npm run test:karma:story-editor:retry-failed -- --headless --snapshots --viewport=1600:1000
+        run: npm run test:karma:story-editor -- --headless --snapshots --viewport=1600:1000 --coverage
         env:
           DISABLE_ERROR_BOUNDARIES: true
 
--- a/packages/karma-puppeteer-launcher/src/index.cjs
+++ b/packages/karma-puppeteer-launcher/src/index.cjs
@@ -50,6 +50,7 @@ function puppeteerBrowser(baseBrowserDecorator, config) {
         // See http://crbug.com/715363
         // We use this flag to work-around this issue.
         '--disable-dev-shm-usage',
+        '--js-flags=--max-old-space-size=8192',
       ],
     };
     const puppeteerOptions = {


swissspidy:
Once we‘ve fixed the memory issues being able to rerun failed tests is crucial. Why remove now?

I'd rather look into whether we can check the exit code of the first run to see whether we need to re-run or not

BrookeGraham:
> Once we‘ve fixed the memory issues being able to rerun failed tests is crucial. Why remove now?
I'd rather look into whether we can check the exit code of the first run to see whether we need to re-run or not

🤔 I am not sure if the disconnection issue actually outputs an exit code before the retry step.

packages/story-editor/src/karma/fixture/fixture.js
```diff
@@ -233,6 +234,10 @@ export class Fixture {
   restore() {
     window.location.hash = '#';
     localStorage.clear();
+    cleanup();
+    this._container = null;
+    this._screen = null;
+    this._editor = null;
   }
```
BrookeGraham, line 241: Nice.

miina:
> Remove Retry-Failed

IIRC, the retry was originally added due to tests that most of the time pass, sometimes failing (flakiness). Some of those tests pass basically almost always but fail very rarely, however, among the 400+ tests there quite often happen to be some of those.

This meant previously that we had to rerun the tests much more often than we do now, wasting much more time waiting for the tests to pass.

Perhaps removing the retry for the failed tests is something that can be added together with running the tests in chunks? Otherwise we'll likely be spending a lot of time again (more than now) waiting for the tests to pass...
The issue of false positives + disconnection has happened twice during the whole project history (now and approx 1.5 years ago) whereas re-running the tests influences everyone every day, IMHO it's worthwhile to have it and discuss alternatives.

swissspidy:
> > Once we‘ve fixed the memory issues being able to rerun failed tests is crucial. Why remove now?
> > I'd rather look into whether we can check the exit code of the first run to see whether we need to re-run or not
> 
> 🤔 I am not sure if the disconnection issue actually outputs an exit code before the retry step.

There's _always_ an exit code when running a command. But it looks like it's always `1`, no matter the reason.

In that case, we can work around this in the `onRunComplete` listener:

https://github.com/google/web-stories-wp/blob/cb3d537e0c3144d6a20a25a936158ff7bce1a40c/packages/karma-failed-tests-reporter/src/index.cjs#L38-L43

First of all, we should not write down the list of failed tests if the browser disconnected afterwards. Like so:

```js
this.onRunComplete = function (browsers, results) {
  if (!outputFile || browsers.length === 0) {
    return;
  }

  if(results.disconnected) {
    // If the browser disconnected, we don't want to report any failed tests.
    return;
  }

  fs.mkdirSync(path.dirname(outputFile), { recursive: true });
  fs.writeFileSync(outputFile, failedTests.join('\n'));
};
```

But that would mean the whole test suite would be retried.

To avoid this, we can update the GitHub Action workflows to only retry tests if the file was written (and the previous exit code was non-zero).

Maybe like so:

```yaml
- name: Run integration tests
  run: npm run test:karma:dashboard -- --headless --viewport=1600:1000 --coverage
  env:
    DISABLE_ERROR_BOUNDARIES: true

- name: Re-run failed tests
  run: |
    if [[ -f "build/karma-dashboard-failed-tests.txt" ]]; then
      npm run test:karma:dashboard:retry-failed -- --headless --viewport=1600:1000
    fi
  if: failure()
  env:
    DISABLE_ERROR_BOUNDARIES: true

- name: Upload code coverage report
  uses: codecov/codecov-action@v2.1.0
  if: always()
  with:
    file: build/logs/karma-coverage/dashboard/lcov.info
    flags: karmatests
```

swissspidy (changes requested):
We should not retry failed tests in case of disconnections instead of removing this entirely.

maxyinger:
> Once we‘ve fixed the memory issues being able to rerun failed tests is crucial. Why remove now?

I wouldn't consider the memory issues fixed, but instead, stable enough to be able to run the entire test suite again. I think the issue is that it's not easy for someone to diagnose that we've surpassed the memory limit, and if we ever do surpass the memory limit, retry has a high probability of getting us into the same situation again.

> I'd rather look into whether we can check the exit code of the first run to see whether we need to re-run or not

Yea this is a good idea. If we can check for the ~exit code~ disconnect I see no harm in the retry step.

Long term, I think it's probably best to split up the the story editor integration tests into multiple actions with less tests in each action. I think it would decrease brittleness and increase DX. The smaller feedback loop in CI would be real nice imo

**update**
Just read through @swissspidy 's proposal for the updated reporter and gh actions. looks great

maxyinger:
>  IMHO it's worthwhile to have it and discuss alternatives.

Yea absolutely agree. Just trying to get the conversation rolling and really appreciate all the input here so far

samwhale:
> > > Once we‘ve fixed the memory issues being able to rerun failed tests is crucial. Why remove now?
> > > I'd rather look into whether we can check the exit code of the first run to see whether we need to re-run or not
> > 
> > 
> > 🤔 I am not sure if the disconnection issue actually outputs an exit code before the retry step.
> 
> There's _always_ an exit code when running a command. But it looks like it's always `1`, no matter the reason.

I've seen exit codes on most other logs, but with the problems lately in the karma tests it seems like sometimes they're not there. What happened here - https://github.com/google/web-stories-wp/runs/4696123861?check_suite_focus=true?

Was the exit code swallowed up somewhere?

swissspidy:
> Was the exit code swallowed up somewhere?

No. This is just Bash.

The command being run is

```bash
npm run test:karma:story-editor -- --headless --snapshots --viewport=1600:1000 --coverage || npm run test:karma:story-editor:retry-failed -- --headless --snapshots --viewport=1600:1000
```

The `||` means the second command is run if the first one has a non-zero exit code.

swissspidy:
> Long term, I think it's probably best to split up the the story editor integration tests into multiple actions with less tests in each action. I think it would decrease brittleness and increase DX. The smaller feedback loop in CI would be real nice imo

I think we're all in agreement there. It has come up before but never found time to invest in this as there wasn't a super straightforward way to do this with Karma.

If someone wants to look into it more closely in a dedicated ticket, I'm definitely supportive of that.

samwhale:
> > Long term, I think it's probably best to split up the the story editor integration tests into multiple actions with less tests in each action. I think it would decrease brittleness and increase DX. The smaller feedback loop in CI would be real nice imo
> 
> I think we're all in agreement there. It has come up before but never found time to invest in this as there wasn't a super straightforward way to do this with Karma.
> 
> If someone wants to look into it more closely in a dedicated ticket, I'm definitely supportive of that.
 
I'm looking into this right now! Would be great to have a dedicated ticket associated with it.

BrittanyIRL:
Can probably close this since the PR to run tests in parallel got merged last week and this seems to no longer be a thing. 

Worth noting are these other things done here: 

- Added cleanup() method to fixture.restore. Doesn't fix memory leaks, but should hopefully prevent a possible source
- Removed a TODO in the render method and added a small implementation there to make sure the story is fully rendered before fixture.render resolves. (I don't think this was much of an issue prior, but shouldn't hurt anything) 

Might be good to just do small PRs for each of these since those are nice cleanups in and of themselves. 

Thoughts?

BrittanyIRL:
Closing in favor of smaller PRs addressing the clean up (noted in working doc, will make issues once it's decided on)

swissspidy:
WDYT about the proposal above? https://github.com/google/web-stories-wp/pull/10143#issuecomment-1009923019

BrittanyIRL:
> WDYT about the proposal above? #10143 (comment)

I say we try it but I think it should be it's own issue

swissspidy:
Of course. Just felt like it was forgotten about :)

BrittanyIRL:
> Of course. Just felt like it was forgotten about :)

Added it into the working doc for karma updates and linked back to your comment. thank you for calling it out, i would have lost it in the shuffle.

PR #10159, opened 2022-01-11
Editor: Added icon to rotation handle

### LINKED ISSUE #10105: UI Simplification: Add rotate handle
## Feature Description
Add an icon for the rotate handler instead of just a filled circle.

See [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=11122%3A445797)

## Additional Context
Screenshot 2022-01-06 at 11 36 59

barklund:
## Context

This makes the rotation handle larger and adds a 28x28px icon to it as per conversation with @aaskedall (and not a 19x19px icon as per Figma).

## Relevant Technical Choices

I needed the icon as either a path or as a raw SVG, so added a new "section" to the exported icons in the design-system, that also exports a raw SVG. This is probably not the best way to do it, so maybe a subfolder instead - so the resource would be `Icons.svg.arrowCircle` (lowercase letter, because it is not a component)?

I will add a comment inline in the PR for discussion.

(currently I think Storybook is broken, because it assumes all exports from icons are react components, and this one is not)

## Todo

- [ ] Fix the export/import of raw SVG's.

## User-facing changes

| Before | After |
|-|-|
| Screen Shot 2022-01-11 at 16 25 25 | Screen Shot 2022-01-11 at 16 25 03 |

## Testing Instructions

1. Select an element on a page
2. Observe the rotation handle

## Checklist

---

Fixes #10105


packages/design-system/src/icons/index.js
```diff
@@ -146,3 +146,6 @@ export { ReactComponent as TurningLine } from './turning_line.svg';
 export { ReactComponent as Union } from './union.svg';
 export { ReactComponent as Video } from './video.svg';
 export { ReactComponent as Launch } from './launch.svg';
+
+// This icon is also needed as a raw SVG for CSS
+export { default as ArrowCircleSVG } from './arrow_circle.svg';
```
barklund: This is not optimal. As mentioned in the PR description, I think it makes more sense to put these in a new object as `Icons.svg.arrowCircle` or something similar (and make sure that the Storybook story for icons ignores the `svg` object).

Thoughts?

miina: Some non-standard icons we've also added directly to the folder it's used in, I think. If it's used only in one place and is non-standard, might also be an option to just import it from the same folder to avoid design system icons issues.

Alternatively sgtm to add a new folder for svg-s and ignore those in the Storybook.

miina (approved):
LGTM pending the SVG import topic

packages/design-system/src/icons/index.js
barklund: I decided to go for a slightly different and cleaner approach. The reason why I kept it here (in the design system) is that it's the exact same icon that we use elsewhere as a regular component, so didn't want to make a duplicate.

barklund:
> | Filename | Size | Change |  |
> | :--- | :---: | :---: | :---: |
> | `assets/js/vendors-wp-dashboard.js` | 151 kB | +55.6 kB (+58%) | 🆘 |

@swissspidy, this doesn't look good. Could you please take a look and see how this can be avoided?

packages/design-system/src/icons/svg.js
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Internal dependencies
+ */
+export { default as arrowCircle } from './arrow_circle.svg';
```
swissspidy: Alternatively, maybe just update `icons/index.js` to do `export { ReactComponent as ArrowCircle, default as ArrowCircleRaw } from './arrow_circle.svg';`?

Feels a bit cleaner than this separation here.

barklund: Then the Storybook stories for the icons would have to be amended to filter this value out, because it just imports `* as icons` from here. See [this conversation](https://github.com/google/web-stories-wp/pull/10159#discussion_r782541788).

But that's of course doable too. I went a bit back and forth on what to do here and have no strong opinions one way or the other.

swissspidy: That shouldn't be too difficult to do I think.

cc @BrittanyIRL for thoughts

swissspidy:
> > Filename
> > Size
> > Change
> > 
> > 
> > 
> > 
> > 
> > `assets/js/vendors-wp-dashboard.js`
> > 151 kB
> > +55.6 kB (+58%)
> > 🆘
> 
> @swissspidy, this doesn't look good. Could you please take a look and see how this can be avoided?

Merging in `main` helped fix the numbers :-)

The additional ~1kb sounds reasonable due to the additional SVG export and some more CSS

packages/design-system/src/icons/svg.js
BrittanyIRL: I'm honestly fine with either approach personally. Updating storybook is doable, I don't really see the benefit from swapping approaches. Think what @barklund did here makes sense and is readable but @swissspidy if you feel strongly about it then I'd be cool with it.

swissspidy: Don't feel strongly about it, but I do think we should at least update the name.

The original icon is called `ArrowCircle`, not `arrowCircle` like here.

swissspidy:
Whoops, accidentally pushed to this branch. Reverted now.

PR #10162, opened 2022-01-12
Media Library: Fix video drag & drop and crash upon deletion

### LINKED ISSUE #9913: Media Library: Reset state when losing focus
## Bug Description

If we open the media element dropdown menu from keyboard and then close it by pressing `Esc` key, the focus moves to the canvas but the media dropdown (3 dots) button keeps visible. ( See attached video )

Addition: pressing `Esc` doesn't reset the element `tabindex`, so when someone tries to refocus the media element, it doesn't start with the first element. ( See 2nd video )

## Expected Behaviour
Pressing on`Esc` key either should refocus the media element, or it shouldn't show the menu button when the focus gets moved to the canvas; plus it should reset the `tabindex`.

## Steps to Reproduce

1. Focus any media element
2. Press `Space` key
3. Press `Esc` key

## Screenshots

https://user-images.githubusercontent.com/17360543/144904883-59c130f3-0a40-4b66-9cd4-ca0768a5145e.mov

https://user-images.githubusercontent.com/17360543/144904924-c9190aa2-9e3e-444f-9fec-57841ab6c490.mov

## Additional Context

- WordPress Version: 5.8.2
- Browser: Chrome

### LINKED ISSUE #10161: Media: Deleting media element crashes the editor
## Bug Description
While deleting a video element the story-editor breaks with a `console.log` printing that error was in the `mediaElement.js`. Further exploration revealed that the `mediaElement` ref changes from `video` element to `image` element thus leading to `pause()` throwing an error.

## Expected Behaviour

The video should delete properly without breaking the editor.

## Steps to Reproduce

1. Go to any local media hover on it.
2. Click on dropdown 
3. Then click on delete from dropdown menu
4. Notice the editor breaking.

## Screenshots

https://user-images.githubusercontent.com/59614577/149081887-33a987f8-1491-4b10-8ccd-ba5ec85ea108.mov

## Additional Context

- Plugin Version: 1.16.0-alpha-0
- WordPress Version: 5.8.3
- Operating System: MacOS monterrey
- Browser: Chrome

### LINKED ISSUE #10166: Drop-targeting video directly from the library is not working
## Bug Description
It's currently not possible to drag and drop a video from the library directly onto the page as a background or into a shape.
Works for images.

## Steps to Reproduce
1. Open the library
2. Try to drag a video and drop it as the background of the page.

## Screenshots
drop-target-video

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

amovar18:
## Context
- This PR fixes mediaElement breaking the story-editor
- This PR also Fixes #10166 
## Summary

Due to recent changes the deletion of a local `video` would lead to breaking the story editor since the node in ref changed from `video` tag to `img` tag.

## User-facing changes
1. You should be able to delete a local video without breaking the editor.
2. You should be able to reset the focus after closing the menu

## Testing Instructions

For #10161:
1. Go to local media tab
2. Focus on any local video
3. Click on dropdown and click on delete
4. You should be now able to delete without breaking the story-editor.

For #10166 :
1. Pick a shape from the shape tab
2. Use a local video and take it over the shape in canvas.
3. You should be able to see the video in the shape.
4. Use another video and drag it to the edge of canvas
6. You should be able to see the video as a background. 

## Checklist

---

Fixes #10161 
Fixes #10166

### DIFF
--- a/packages/story-editor/src/components/library/panes/media/common/mediaElement.js
+++ b/packages/story-editor/src/components/library/panes/media/common/mediaElement.js
@@ -134,12 +134,7 @@ function Element({
         setHoverTimer(null);
       }
     };
-    if (isMenuOpen) {
-      if (mediaElement.current && !mediaElement.current.paused) {
-        // If it's a video, pause the preview while the dropdown menu is open.
-        mediaElement.current.pause();
-      }
-    } else {
+    if (!isMenuOpen) {
       if (active) {
         setShowVideoDetail(false);
         if (mediaElement.current && hoverTimer == null) {
--- a/packages/story-editor/src/components/library/panes/media/common/innerElement.js
+++ b/packages/story-editor/src/components/library/panes/media/common/innerElement.js
@@ -240,12 +240,6 @@ function InnerElement({
   }
 
   const dragHandler = (event) => {
-    if (
-      [ContentType.VIDEO, ContentType.GIF].includes(type) &&
-      !mediaElement.current?.paused
-    ) {
-      mediaElement.current?.pause();
-    }
     if (!hasSetResourceTracker.current) {
       // Drop-targets handling.
       resourceList.set(resource.id, {


packages/story-editor/src/components/library/panes/media/common/mediaElement.js
```diff
@@ -179,9 +175,15 @@ function Element({
         url={resource.attribution.author.url}
       />
     );
-
   const ref = useRef();
-
+  const onMenuCancelled = useCallback(() => {
+    if (firstElementId) {
+      document.querySelector(`[data-id='${firstElementId}']`).tabIndex = 0;
```
swissspidy: Hmm doesn't feel very clean to use a data attribute to get the first element 🤔 

Is there no other way, e.g. by going via `ref.current.parentElement` or something?

amovar18: @swissspidy I think I can get the first child of the parent and set it’s tabIndex to 0 need to try. Would that look clean like either ref.current.parentElement.firstChild or either get parentRef and modify its firstChild

spacedmonkey:
Should we add a e2e test for this?

swissspidy:
It'd be more suitable for a Karma test, but yeah tests would be nice

amovar18:
> It'd be more suitable for a Karma test, but yeah tests would be nice

Don think so there is already a test which checks if the `edit/delete menu` is open:
https://github.com/google/web-stories-wp/blob/2406aa45b9baa78ef66fa41508ab8eb1201be456/packages/story-editor/src/components/library/karma/mediaTab.karma.js#L59-L81

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
```diff
@@ -179,9 +174,13 @@ function Element({
         url={resource.attribution.author.url}
       />
     );
-
   const ref = useRef();
-
+  const onMenuCancelled = useCallback(() => {
+    ref.current.parentElement.firstChild.tabIndex = 0;
```
swissspidy: That looks much cleaner!

spacedmonkey (approved):

swissspidy:
Deletion works great again, but now I also see this issue:

When I go through the media library with the arrow keys and then try to open the menu, the button disappears:

https://user-images.githubusercontent.com/841956/149503820-c9c9b58e-9eaa-49fe-992b-5290de211375.mov

But this also happens on `main`, so could be fixed in its own ticket.

It would be great to ideally have Karma tests for this as well as for the tabindex reset.

Is this something you'd be able to help with?

We could also consider splitting this PR up so that we fan fix the crash ASAP and deal with the focus change later.

amovar18:
Let me just revert the focus issue and push the updates, I will create a separate PR with focusChange later Sure would love to add karma test to check for `tabIndex reset`

amovar18:
Reverted to commit which led to closing of PR

swissspidy (approved):
Works like a charm!

PR #10169, opened 2022-01-12
Replace `useComposeRefs` with `useCombinedRefs`

### LINKED ISSUE #10093: Code Quality: Consolidate useCombinedRefs & useComposeRefs usage
These two hooks seem to duplicate the same functionality. This ticket is to just use one of them and replace the other everywhere else.

https://github.com/google/web-stories-wp/blob/main/packages/react/src/useCombinedRefs.js

https://github.com/google/web-stories-wp/blob/main/packages/design-system/src/utils/useComposeRefs.js

Probably best to keep whichever hook we consolidate too inside the `/react` package because of the current dependency setup.

spacedmonkey:
## Summary

These two hooks seem to duplicate the same functionality. Remove useComposeRefs

## Testing Instructions

1.

## Checklist

---

Fixes #10093


packages/react/src/test/useComposeRefs.js
```diff
@@ -16,14 +16,14 @@
...
 import { renderHook } from '@testing-library/react-hooks';
 /**
  * Internal dependencies
  */
-import useComposeRefs from '../useComposeRefs';
+import { useRef } from '..';
+import useCombinedRefs from '../useCombinedRefs';
 
-describe('useComposeRefs', () => {
+describe('useCombinedRefs', () => {
```
swissspidy: Please also rename the file to `useCombinedRefs.js`

spacedmonkey: I thought I did.

swissspidy (approved):

maxyinger:
@spacedmonkey just noticed this with `useCombineRefs` implementation, but it will create a callback ref on every render because a rest parameter returns a new array instance every time it is called. Made a little codesandbox to visualize:
https://codesandbox.io/s/patient-grass-b66b8?file=/src/index.js

I believe the `useComposeRefs` implementation avoided this while maintaining the same functionality. Sorry for not catching this earlier.

Probably doesn't have too many implications, but probably good just for code quality sake

swissspidy:
@littlemilkstudio could you perhaps create quick follow-up PR?

maxyinger:
follow-up here:
https://github.com/google/web-stories-wp/pull/10183

Issue #10157, opened 2022-01-11
Karma: Explore running tests in parallel


swissspidy:
## Task Description

As was mentioned multiple times over the course of this project and again this week, it might be beneficial to split up Karma tests across multiple CI jobs in parallel to speed things up and shorten the feedback loop.

[`karma-parallel`](https://www.npmjs.com/package/karma-parallel) looks like a promising candidate for that.

We might be able to use it in combination with a [test matrix](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) to pass a custom `shardIndex`, but not sure if that works.

That package _should_ work well with code coverage in theory, but not sure if it also does in practice.

swissspidy:
Assigning @samwhale because he expressed interest in this.

/cc @littlemilkstudio @BrittanyIRL @merapi for visibility

PR #10182, opened 2022-01-12
Draft: Split up karma tests in CI

samwhale:
## Context

Draft pr for exploring ideas.

## Summary

Current approach:

Thinking that we shouldn't use [`karma-parallel`](https://www.npmjs.com/package/karma-parallel) or [`karma-sharding`](https://github.com/rschuft/karma-sharding) if we're wanting a different github action for each chunk of tests. From my understanding, those two packages would run the test chunks on different browser instances in the same action.

Instead I'm attempting to create a script that runs a set of tests that we could then pair with a [test matrix](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) to run all the tests in separate actions at the same time.

## Testing Instructions

1.

## Checklist

---

See #10157

### DIFF
--- a/packages/story-editor/src/karma-tests.cjs
+++ b/packages/story-editor/src/karma-tests.cjs
@@ -14,6 +14,12 @@
  * limitations under the License.
  */
 
+const start = process.env.KARMA_START;
+const count = process.env.KARMA_COUNT;
+
 // Collect all Karma tests together for the test webpack config.
 const testsContext = require.context('.', true, /\.karma\.js$/);
-testsContext.keys().forEach(testsContext);
+testsContext
+  .keys()
+  .slice(start, start + count)
+  .forEach(testsContext);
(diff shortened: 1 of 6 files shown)


packages/story-editor/src/karma-tests.cjs
```diff
@@ -14,6 +14,12 @@
...
+const start = process.env.KARMA_START;
+const count = process.env.KARMA_COUNT;
+
 // Collect all Karma tests together for the test webpack config.
 const testsContext = require.context('.', true, /\.karma\.js$/);
-testsContext.keys().forEach(testsContext);
+testsContext
+  .keys()
+  .slice(start, start + count)
+  .forEach(testsContext);
```
swissspidy, line 25: Heads up: This won't work anymore once we move to webpack v5, see #5792 

This file will be removed as tests will be loaded differently.

swissspidy:
> Thinking that we shouldn't use `karma-parallel` or `karma-sharding` if we're wanting a different github action for each chunk of tests. From my understanding, those two packages would run the test chunks on different browser instances in the same action.

I was hoping that we could make it work by providing a custom shard config to `karma-parallel` together with our matrix in order to force it to run a specific shard but only on 1 browser.

swissspidy:
See #10186 for what I had in mind.

PR #10186, opened 2022-01-13
Use karma-parallel for running tests on CI

### LINKED ISSUE #10157: Karma: Explore running tests in parallel

swissspidy:
## Summary

This change splits up the Karma tests on CI into multiple shards using a matrix in the GitHub Actions workflow and [`karma-parallel`](https://github.com/joeljeske/karma-parallel).

Might not be worth it for the dashboard tests because there are just so few tests there. The overhead of running these in parallel might make things actually slower than just running all of them in a single job.

## Relevant Technical Choices

Needed to patch karma-parallel to ensure our top-level `beforeAll` of our karma fixture gets executed properly for every shard.

## To-do

* [x] Maybe remove again for dashboard tests

## User-facing changes

N/A

## Testing Instructions

1.

No

No

No

## Checklist

---

Fixes #10157

### DIFF
--- a/karma-dashboard.config.cjs
+++ b/karma-dashboard.config.cjs
@@ -41,6 +41,13 @@ module.exports = function (config) {
       .join('|');
   }
 
+  let totalShards = 1;
+  let shardNumber = 1;
+  if (config.shard) {
+    [shardNumber, totalShards] = config.shard.split('/');
+  }
+  const shardIndex = shardNumber - 1;
+
   config.set({
     plugins: [
       'karma-chrome-launcher',
@@ -48,7 +55,7 @@ module.exports = function (config) {
       'karma-sourcemap-loader',
       'karma-webpack',
       'karma-coverage-istanbul-reporter',
-      require('karma-parallel'),
+      'karma-parallel',
       require('@web-stories-wp/karma-puppeteer-launcher'),
       require('@web-stories-wp/karma-puppeteer-client'),
       require('@web-stories-wp/karma-failed-tests-reporter'),
@@ -162,21 +169,23 @@ module.exports = function (config) {
     // how many browsers should be started simultaneously
     concurrency: Infinity,
 
-    // Sharding.
+    // Sharding configuration for CI.
+    // We trick karma-parallel into using only 1 browser (parallelOptions.executors === 1)
+    // while using a custom strategy that splits up the tests into multiple shards (config.executors > 1).
+    // This allows us to run ony a subset of the tests like so:
+    // npm run test:karma:dashboard -- --headless --shard=1/3 # Run the first of 3 desired shards.
     parallelOptions: {
-      shardStrategy: typeof config.shardIndex !== 'undefined' && typeof config.executors !== 'undefined' ? 'custom' : 'round-robin',
+      shardStrategy: config.shard ? 'custom' : 'round-robin',
       // If we're using custom sharding, just spin up 1 browser,
       // but do the splitting in the custom strategy below.
-      executors: typeof config.shardIndex !== 'undefined' ? 1 : undefined, // Default is cpu cores - 1.
-      shardIndex: typeof config.shardIndex !== 'undefined' ? Number(config.shardIndex) : undefined,
+      executors: config.shard ? 1 : undefined, // Default is cpu cores - 1.
       // Re-implements a round-robin strategy, but with a custom shardIndex.
-      // Need to use the Function constructor here so we have access to config.shardIndex,
+      // Need to use the Function constructor here so we have access to the outer shardIndex and totalShards vars,
       // because karma-parallel serializes this function.
       customShardStrategy: new Function('parallelOptions', `
         window.parallelDescribeCount = window.parallelDescribeCount || 0;
         window.parallelDescribeCount++;
-        const shouldRunThisTest = (window.parallelDescribeCount % ${Number(config.executors)} === ${Number(config.shardIndex)});
-        console.log( "[SHARDING DEBUG]", "Count:", window.parallelDescribeCount, "Executors:", ${Number(config.executors)}, "Current shardIndex:", ${Number(config.shardIndex)}, "Should run test?", shouldRunThisTest);
+        const shouldRunThisTest = (window.parallelDescribeCount % ${totalShards} === ${shardIndex});
         return shouldRunThisTest;
       `),
     },
--- a/karma-dashboard.config.cjs
+++ b/karma-dashboard.config.cjs
@@ -186,6 +186,7 @@ module.exports = function (config) {
       // Re-implements a round-robin strategy, but with a custom shardIndex.
       // Need to use the Function constructor here so we have access to the outer shardIndex and totalShards vars,
       // because karma-parallel serializes this function.
+      // eslint-disable-next-line no-new-func
       customShardStrategy: new Function(
         'parallelOptions',
         `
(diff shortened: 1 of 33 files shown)


swissspidy:
Unfortunately all but the first shard currently fail with some unexpected errors like this:

```
	TypeError: expectAsync(...).toHaveNoViolations is not a function
```

Also surprised to suddenly see these warnings:

```
Warning: react-modal: App element is not defined. Please use `Modal.setAppElement(el)` or set `appElement={el}`. This is needed so screen readers don't see main content when modal is opened. It is not recommended, but you can opt-out by setting `ariaHideApp={false}
```

I think the reason is that with `karma-parallel`, the `beforeAll()` `packages/karma-fixture/src/init.js` somehow is lost.

Maybe because it's a top-level `beforeAll()` without a `describe()`, and `karma-parallel` shards tests by `describe`

Need to do some debugging in `initKarmaParallelizer`

karma-dashboard.config.cjs
```diff
@@ -161,6 +162,25 @@ module.exports = function (config) {
...
+      // Need to use the Function constructor here so we have access to config.shardIndex,
+      // because karma-parallel serializes this function.
+      customShardStrategy: new Function('parallelOptions', `
+        window.parallelDescribeCount = window.parallelDescribeCount || 0;
+        window.parallelDescribeCount++;
+        const shouldRunThisTest = (window.parallelDescribeCount % ${Number(config.executors)} === ${Number(config.shardIndex)});
+        console.log( "[SHARDING DEBUG]", "Count:", window.parallelDescribeCount, "Executors:", ${Number(config.executors)}, "Current shardIndex:", ${Number(config.shardIndex)}, "Should run test?", shouldRunThisTest);
+        return shouldRunThisTest;
+      `),
+    },
```
samwhale: will we be doing this with both the dashboard and the story editor? I was working in the story editor first

karma-dashboard.config.cjs
```diff
@@ -161,6 +162,25 @@ module.exports = function (config) {
...
+    // Sharding.
+    parallelOptions: {
+      shardStrategy: typeof config.shardIndex !== 'undefined' && typeof config.executors !== 'undefined' ? 'custom' : 'round-robin',
+      // If we're using custom sharding, just spin up 1 browser,
+      // but do the splitting in the custom strategy below.
+      executors: typeof config.shardIndex !== 'undefined' ? 1 : undefined, // Default is cpu cores - 1.
+      shardIndex: typeof config.shardIndex !== 'undefined' ? Number(config.shardIndex) : undefined,
+      // Re-implements a round-robin strategy, but with a custom shardIndex.
+      // Need to use the Function constructor here so we have access to config.shardIndex,
+      // because karma-parallel serializes this function.
```
samwhale: I'm a little confused here. So we wouldn't be able to access the `shardIndex` from the config that's injected into the `customShardStrategy`?

This worked for me:
```js
    customShardStrategy: (config2) => {
        window.parallelDescribeCount = window.parallelDescribeCount || 0;
        window.parallelDescribeCount++;
        const shouldRunThisTest =
          window.parallelDescribeCount % Number(config2.executors) ===
          Number(config2.shardIndex);
        console.log(
          '[SHARDING DEBUG]',
          'Count:',
          window.parallelDescribeCount,
          'Executors:',
          Number(config2.executors),
          Number(config2.shardIndex),
          'Should run test?',
          shouldRunThisTest
        );
        return shouldRunThisTest;
      },```

File: karma-dashboard.config.cjs (hunk 1 above)
Comment (swissspidy) [reply karma-dashboard.config.cjs]: If it works, we‘d want to do it for both, yes.

I was just experimentig with the dashboard first because there are fewer tests to run.

File: karma-dashboard.config.cjs (hunk 2 above)
Comment (swissspidy) [reply karma-dashboard.config.cjs]: Hmm I initially had some issues with the one from config2 not having the correct values, so I wanted to access the one I passed via CLI. Will take another look.
Of course it would be great if it‘s not needed.

Comment (swissspidy) [reply karma-dashboard.config.cjs]: Oh, we need access to the outside executors variable which is 4, whereas config2.executors is 1 in this PR. So I don‘t see how yours works

**Comment by merapi** [comment]
Oh, we are using matrix already, this looks good :)

**Comment by samwhale** [comment]
This is awesome! I was worried initially that doing it with karma-parallel would prevent us from running tests in different vms but this looks perfect. Really really nice how individual sets of tests can fail and we can get quicker actionable feedback.

File: karma-dashboard.config.cjs (hunk 1 above)
Comment (samwhale) [reply karma-dashboard.config.cjs]: Ah yup that makes sense. Thanks!

File: karma-dashboard.config.cjs (hunk 2 above)
Comment (samwhale) [reply karma-dashboard.config.cjs]: I see what you're saying now. The outside `executors` and the inside `executors` are not the same value. We need to keep the `executors` in the parallel config at 1 while the outside one needs to be dynamic 👍

Comment (swissspidy) [reply karma-dashboard.config.cjs]: Yes exactly!

I‘ll try to document this better.

File: packages/dashboard/src/dataUtils/formattedTemplatesArray.js [hunk 3]
```diff
@@ -1489,7 +1489,7 @@ const formattedTemplatesArray = [
       { label: 'Black', color: '#28292b', family: 'Black' },
       { label: 'White', color: '#fff', family: 'White' },
       { label: 'Brown', color: '#eadfd6', family: 'Brown' },
-      { label: 'Test Color', color: '#eadfd6', family: 'Test Color' },
+      { label: 'Test Color', color: '#fff000', family: 'Test Color' },
```
swissspidy, line 1492: Already fixed in #9944 but it bugged me here during testing

patches/karma-parallel+0.3.1.patch
```diff
@@ -0,0 +1,17 @@
...
+@@ -120,9 +120,11 @@ function createFakeTestContext(ctx, shouldUseDescription) {
+       hasFocusedWithoutFaking =
+         hasFocusedWithoutFaking || (isFocus && !isFaking);
+ 
++      const isTopLevelSetupOrTeardown = typeof name === 'function' && !isFocus && !isDescription && !isSpec && depth === 0;
++
+       if (isDescription) def = wrapDescription(def);
+ 
+-      if (!isFaking || forceDescribe) {
++      if (!isFaking || forceDescribe || isTopLevelSetupOrTeardown) {
```
swissspidy, line 14: This makes our fixture run properly for each shard

merapi (approved):
For a small number of tests in Dashboard (50), it makes no difference (long setup):

image

image

but for the Editor - it's a massive QoL improvement 🥇  

 
image

image

samwhale (approved):
Huge wins! Thanks for helping me out with this 😄

swissspidy:
Thanks for the reviews!

Yeah I think I‘m going to remove it again from the dashboard so that there‘s one less GHA job in the queue

PR #10197, opened 2022-01-14
Code Quality: De-duplicate popup

### LINKED ISSUE #10089: Deduplicate Popup
Noticed in review of #9944 that we still have 2 `Popup` components (one in the design system and one in the story-editor), these are nearly identical and combining them would be a nice clean up. 

## Acceptance 
- no regressions in used popups
- Popup component lives in Design System

amovar18:
## Context
This PR removes the `Popup` from `story-editor` and brings in `Popup` from `design-system`.

## Summary

Since there were 2 `Popups` defined at separate locations in codebase it would be best to keep the `Popup` at a single place and reduce redundancy.

## Testing Instructions

1. All the places where `Popup` has been used should work without breaking the story-editor.

## Checklist

---

Fixes #10089


BrittanyIRL (commented):
This looks good! I think I tested a range of use cases and everything is playing nicely. 

@BrookeGraham  could you take a look at this and make sure there's no regressions on the tooltip edge cases you fixed like 6 months back? 

Otherwise, just rebase this off main so we can ensure all tests are good to go. Didn't realize you'd put this up last week, apologies for the delay on that front.

packages/design-system/src/components/popup/utils/getOffset.js
```diff
@@ -64,61 +79,45 @@ export function getYOffset(placement, spacing = 0, anchorRect) {
...
- *
- * @param {string} placement Placement.
- * @param {Spacing} spacing Spacing.
- * @param {MutableRefObject<HTMLElement>} anchor Anchor element
- * @param {MutableRefObject<HTMLElement>} dock Dock element.
- * @param {MutableRefObject<HTMLElement>} popup Popup element.
- * @return {Offset} Popup offset.
- */
-export function getOffset(placement, spacing, anchor, dock, popup) {
+export function getOffset(placement, spacing, anchor, dock, popup, isRTL) {
```
swissspidy: Why remove the JSDoc?

amovar18: Removed it by mistake added in https://github.com/google/web-stories-wp/pull/10197/commits/249932807171078b645d44e02514c784dc44a594

packages/design-system/src/components/popup/index.js
```diff
@@ -35,76 +36,84 @@ import { themeHelpers } from '../../theme';
...
   placement = PLACEMENT.BOTTOM,
-  zIndex = DEFAULT_POPUP_Z_INDEX,
   spacing,
   isOpen,
+  invisible,
   fillWidth = false,
   fillHeight = false,
   onPositionUpdate = noop,
   refCallback = noop,
+  topOffset = DEFAULT_TOPOFFSET,
```
swissspidy, line 80: Where is `DEFAULT_TOPOFFSET  = 32` coming from? The admin bar?

A sane default here is `0`. Because in a non-WordPress standalone fullscreen editor there won't be any offset.

The `32` should be passed from WordPress only

packages/design-system/src/components/popup/utils/getOffset.js
```diff
@@ -86,39 +99,54 @@ export function getYOffset(placement, spacing = 0, anchorRect) {
  * @param {MutableRefObject<HTMLElement>} anchor Anchor element
  * @param {MutableRefObject<HTMLElement>} dock Dock element.
  * @param {MutableRefObject<HTMLElement>} popup Popup element.
+ * @param {boolean} isRTL isRTL.
  * @return {Offset} Popup offset.
  */
-export function getOffset(placement, spacing, anchor, dock, popup) {
+
```
swissspidy: errant empty line here

packages/design-system/src/components/popup/utils/getOffset.js
```diff
@@ -86,39 +99,54 @@ export function getYOffset(placement, spacing = 0, anchorRect) {
  * @param {MutableRefObject<HTMLElement>} anchor Anchor element
  * @param {MutableRefObject<HTMLElement>} dock Dock element.
  * @param {MutableRefObject<HTMLElement>} popup Popup element.
+ * @param {boolean} isRTL isRTL.
```
swissspidy, line 102: thanks 👍

packages/design-system/src/components/popup/index.js
amovar18: oh ok will change it to 0

packages/story-editor/src/components/canvas/pageAttachment/index.js
```diff
@@ -144,7 +144,10 @@ function PageAttachment({ pageAttachment = {} }) {
     icon,
     theme,
   } = pageAttachment;
-
+  const {
+    isRTL,
+    styleConstants: { topOffset },
```
miina: If I understand correctly, this is a value that could potentially come from the backend (but is not set now)? So currently, it's `0` (the default) and would still need defining `32` in the backend?

Wondering if we could make the name more explanatory, e.g. `styleConstants: { editorTopOffset },` or something similar? it's currently not clear what kind of top offset is meant.

An alternative would be to actually calculate the real top offset once the editor has loaded instead of taking a static value from the backend (e.g. we're calculating the real page size once it has loaded as well), thoughts? 

(In case I understand correctly that's what this `topOffset` means here)

amovar18: Yes the `topOffset` value comes from the `wp-story-editor` where it has been defined as a constant so whoever wants to integrate may pass it the same way we are doing. I am fine doing it either way but if it were to be calculated once page loads then i guess it would require another PR

miina: Ah, okay, thought that this was supposed to come from the backend but coming from `wp-story-editor` makes sense.

miina:
From testing (note: using 13'' screen):

The color picker popup is not scrollable currently:
color-picker

amovar18:
@miina I checked on the latest main branch it is not scrollable there too, could you check if its the same for you or I will try to fix if it came from my changes.

miina:
> I checked on the latest main branch it is not scrollable there too, could you check if its the same for you or I will try to fix if it came from my changes.

I can see it working on `main` (scrollable when the color picker doesn't fit on the screen as it happens on my 13'')
main

packages/design-system/src/components/tooltip/index.js
```diff
@@ -267,6 +267,7 @@ function Tooltip({
         spacing={spacing}
         isOpen={Boolean(shown && (shortcut || title))}
         onPositionUpdate={positionArrow}
+        noOverFlow
```
amovar18, line 285: Added this because of `overflow-y:auto` used to display scrollbars on tooltip and made the UI look bad, therefore if `noOverFlow` is passed then `overflow-y` will not be attached to the popup.

Edit: had to do the same to solve the problem regarding https://github.com/GoogleForCreators/web-stories-wp/pull/10197#issuecomment-1018714806 since `Popup` for `ColorPicker` has different offset calculations and `Popup` for `Tooltip` has different offset Thats why I have passed `shouldHaveDifferentoffset` from `ColorPicker`. Could not find a better way to calculate offset

packages/story-editor/src/components/canvas/pageAttachment/index.js
mohdsayed: @miina I agree, `styleConstants: { editorTopOffset }` would be more explanatory 👍  . We may do it now or I will raise a PR for it. 
Calculating the real top offset looks like a natural choice here however we also have a collapsable sidebar menu involved, so we would need some observer/listener for it. See the discussion in https://github.com/google/web-stories-wp/pull/10119

BrookeGraham:
There are some regressions involving the `tooltip` which uses the `popup`. When there isn't sufficient space within the viewport `placement` should be updated. This should work in both rtl and ltr with placement being either `right` or `left`, and it should work for `top` / `bottom`

||RTL|Top / Bottom|
|--|--|--|
|Main|main rtl|main top to bottom|
|This PR|regression rtl|regression top to bottom|

samwhale:
Heads up! We saw a bug where clicking and dragging a video into the canvas was crashing the app. This was happening in the story editor's `Popover` component #10383

Let's make sure to test that interaction here too!

packages/design-system/src/components/tooltip/index.js
BrookeGraham: Yea it seems it is now forcing the scroll bar on `ColorPicker` to outside of component. 
|Main|This pr|
|--|--|
|main scroll|this pr scroll bar|

BrookeGraham (commented):
The ColorPicker (and likely other Popups that use `ToolTip` from the Story Editor end up getting the wrong placement in RTL. Most likely because we invert placement before moving on to popup. Since popup is now bringing in `isRTL` I think it get reverted for a second time basically undoing the new placement values. 
|Main|This pr|
|--|--|
|rtl main|rtl this pr|

packages/design-system/src/components/popup/index.js
```diff
@@ -173,6 +171,7 @@ function Popup({
           invisible={invisible}
           topOffset={topOffset}
           noOverFlow={noOverFlow}
+          isRTL={isRTL}
```
BrookeGraham, line 158: 🎉

amovar18: Does it work properly now?

BrookeGraham: Looks great. I will have time to finish a review later today. But at first glance it looks beautiful. Thanks for all the work on this. I know the popup stuff can be tricky!

packages/wp-story-editor/src/components/documentPane/publish/publishTime.js
```diff
@@ -128,6 +132,7 @@ function PublishTime() {
         />
       </Row>
       <Popup
+        topOffset={TOOLBAR_HEIGHT}
```
BrookeGraham, line 136: Should also have `isRTL`

packages/design-system/src/components/popup/index.js
```diff
@@ -20,63 +20,64 @@
...
   spacing,
   isOpen,
+  invisible,
   fillWidth = false,
   fillHeight = false,
   onPositionUpdate = noop,
   refCallback = noop,
+  topOffset = DEFAULT_TOPOFFSET,
+  noOverFlow = false,
+  shouldHaveDifferentoffset = false,
```
BrookeGraham: nit: `shouldHaveDifferentOffset`

packages/design-system/src/components/popup/index.js
```diff
@@ -91,27 +92,53 @@ function Popup({
...
       // If scrolling within the popup, ignore.
       if (evt?.target?.nodeType && popup.current?.contains(evt.target)) {
         return;
       }
-
-      await setPopupState({
-        offset:
-          anchor?.current && getOffset(placement, spacing, anchor, dock, popup),
+      setPopupState({
+        offset: anchor?.current
```
BrookeGraham: If there is no `anchor` ref let's not render at all. See [comment](https://github.com/GoogleForCreators/web-stories-wp/pull/10383#issuecomment-1023119291) for discussion.

packages/design-system/src/components/popup/index.js
```diff
@@ -163,6 +192,7 @@ Popup.propTypes = {
   zIndex: PropTypes.number,
   spacing: PropTypes.object,
   isOpen: PropTypes.bool,
+  noOverFlow: PropTypes.bool,
   fillWidth: PropTypes.oneOfType([PropTypes.number, PropTypes.bool]),
   fillHeight: PropTypes.oneOfType([PropTypes.number, PropTypes.bool]),
   onPositionUpdate: PropTypes.func,
```
BrookeGraham, line 183: Looks like we are missing a few props. `isRTL`, `invisible`, `shouldHaveDifferentOffset` and `topOffset`

packages/design-system/src/components/tooltip/index.js
```diff
@@ -282,6 +280,7 @@ function Tooltip({
         spacing={spacing}
```
BrookeGraham, line 282: 🤔 Should we be passing through  `isRTL`, `invisible`, `shouldHaveDifferentOffset` and `topOffset`? Granted when using Tooltip from Design system it does do it's own RTL calculating.

packages/story-editor/src/components/library/panes/media/local/dropDownMenu.js
```diff
@@ -134,13 +134,15 @@ function DropDownMenu({
     () => setShowDeleteDialog(false),
     [setShowDeleteDialog]
   );
-
+  const { isRTL } = useConfig();
```
BrookeGraham: Move this to be with lines 143-145 where topOffset is defined, or vis versa.

packages/design-system/src/components/popup/index.js
```diff
@@ -20,63 +20,64 @@
...
+  isRTL = false,
   anchor,
   dock,
   children,
   renderContents,
   placement = PLACEMENT.BOTTOM,
-  zIndex = DEFAULT_POPUP_Z_INDEX,
   spacing,
   isOpen,
+  invisible,
```
BrookeGraham: Let's give this a fallback.

packages/design-system/src/components/popup/index.js
BrookeGraham: What do y'all think about making `isRTL` and `topOffset` required? There are a couple instances where it's possible that `isRTL` wouldn't effect placement. eg. `Dropdown` uses `Popup` but _usually_ has a `placement` of `bottom` though, it can get updated to `top` if space is an issue. In theory, however, we could make a `Dropdown` placed `right` or `left`. But I don't think we do.

packages/story-editor/src/components/panels/design/textStyle/stylePresets/stylePresets.js
```diff
@@ -156,6 +159,8 @@ function PresetPanel({ pushUpdate }) {
             <Icons.ChevronDownSmall />
           </StyledMoreButton>
           <Popup
+            topOffset={topOffset}
+            isRTL={isRTL}
```
BrookeGraham, line 160: This is now getting cut off. 

Screen Shot 2022-01-28 at 1 48 21 PM

BrookeGraham (changes requested):
MediaInput` is another place where we aren't yet sending in `isRTL` to `Popup`.

Also, maybe for another ticket,  `Tooltip` uses `Popup`, but handles it's own RTL switching. Maybe we can clean those up and not have it in two places.

packages/design-system/src/components/popup/index.js
amovar18: I think `topOffset` should remain optional because from what I understand `Popup` that lives in `design-system` is not aware which CMS(no CMS) it is in. If not passed then it sets to 0. 
Edit: As for isRTL i think we should make it required.

packages/design-system/src/components/tooltip/index.js
amovar18: I dont think we need to pass it since the calculation wont be affected

BrookeGraham:
Thanks for those fixes. Looks like there are some merge conflicts. Mind resolving those and I will take another look at this.

packages/design-system/src/components/popup/index.js
```diff
@@ -20,98 +20,124 @@
...
       // If scrolling within the popup, ignore.
       if (evt?.target?.nodeType && popup.current?.contains(evt.target)) {
         return;
       }
-
-      await setPopupState({
-        offset:
-          anchor?.current && getOffset(placement, spacing, anchor, dock, popup),
+      setPopupState({
+        offset: anchor?.current
```
BrookeGraham: nit: We don't need the optional chaining here since we already check for it above. `anchor.current` is fine.

BrookeGraham (approved):
I think this looks great. Only one small nit. Can we get another set of 👀 before moving this to QA? @BrittanyIRL @samwhale @miina

packages/design-system/src/components/popup/utils/getOffset.js
```diff
@@ -86,38 +99,59 @@ export function getYOffset(placement, spacing = 0, anchorRect) {
...
+ * @param {number} topOffset Header Offset.
+ * @param {boolean} shouldHaveDifferentOffset Boolean to set different offset for colorInput popup.
```
miina: Could we make this comment a bit more general? The `design-system` components shouldn't be aware of where or why they are used, however, the comment says here that it's specifically for `colorInput` prop.

Maybe there's a better name and/or better comment for this prop that would explain what it does exactly instead of where it's used?

E.g. if `shouldHaveDifferentOffset` is `true`, what happens then (regardless of where it's used)? It's not very clear from `shouldHaveDifferentOffset`/comment what will happen with the popup when it's set. Thoughts?

amovar18: would this work `shouldHaveDifferentOffset Boolean to change calculation of y coordinate for popup.`?

packages/design-system/src/components/popup/index.js
```diff
@@ -20,98 +20,124 @@
...
   })
 )`
   /*! @noflip */
   left: 0px;
   top: 0px;
   position: fixed;
-
-  ${themeHelpers.scrollbarCSS};
+  z-index: 2;
+  ${({ noOverFlow }) => (noOverFlow ? `` : `overflow-y: auto;`)};
```
miina: Nit for consistency: 
```suggestion
  ${({ noOverFlow }) => (noOverFlow ? '' : 'overflow-y: auto;')};
```

packages/design-system/src/components/popup/index.js
```diff
@@ -20,98 +20,124 @@
...
-  zIndex = DEFAULT_POPUP_Z_INDEX,
   spacing,
   isOpen,
+  invisible = false,
   fillWidth = false,
   fillHeight = false,
   onPositionUpdate = noop,
   refCallback = noop,
+  topOffset = DEFAULT_TOPOFFSET,
+  noOverFlow = false,
```
miina, line 82: Nit: generally, it's easier to read positive props, e.g. `hasOverflow` instead of `noOverflow`

packages/design-system/src/components/popup/index.js
```diff
@@ -20,98 +20,124 @@
...
+    isRTL,
+    invisible,
   }) => ({
     style: {
-      transform: `translate(${x}px, ${y}px) ${getTransforms(placement)}`,
+      transform: `translate(${x}px, ${y}px) ${getTransforms(placement, isRTL)}`,
       ...(fillWidth ? { width: `${width}px` } : {}),
       ...(fillHeight ? { height: `${height}px` } : {}),
-      zIndex,
+      ...(invisible ? { visibility: `hidden` } : {}),
```
miina: Nit for consistency:
```suggestion
      ...(invisible ? { visibility: 'hidden' } : {}),
```

miina (dismissed):
LGTM, nice work!
(I only had very tiny nitpicking comments which are related to readability and consistency and can be just ignored really.)

packages/design-system/src/components/popup/utils/getOffset.js
miina: Hm, I'm still not understanding what it does exactly though :) How does it calculate the y coordinate differently? E.g. is this needed for popups that don't fit to the screen? Or in which case would this need to be used?

miina: For example, when I resize my screen height to very small and open the Document Panel -> Publish time -> Open Calendar, the following happens:
onPR

The calendar is not scrollable until the end and is positioned a bit differently than on `main` where the whole calendar is accessible and positioned so that as much as possible of it would fit to the screen:
onMain

Would `shouldHaveDifferentOffset` help here with the Calendar bug so that it wouldn't be based on the anchor but take the available room so that the calendar would fit as much as possible?

miina: ^ If that's the case and that's what `shouldHaveDifferentOffset` means then I wonder if `shouldHaveDifferentOffset` should be the default behaviour? If the popup fits to the screen, it's positioned next to the anchor, if it doesn't, then it's positioned so that it would fit as much as possible on the screen (as it currently seems to be on `main`).

amovar18: AFAIK if `shouldHaveDifferentOffset` is passed it will also include the `topOffset` while calculation otherwise it will calculate the y coordinate using the bodyrect prop only this is used for popups which are bigger than tooltip (in story-editor there are only 2 cases `colorInput` and `textSets` and `publishTime`)
Yes this also does solve the problem of the calendar placement

miina: > Yes this also does solve the problem of the calendar placement

Hm, okay, is there a reason why not always use the `shouldHaveDifferentOffset` logic? So that whenever a popup doesn't fit the screen in its default position, it would be adjusted accordingly? 

This way, no matter the popup, it would always support any screen size. It seems to do so on `main`. Also, it's a bit tricky for the engineers to have to determine and track in code if the popup fits on screen or not (depends on the screen / window resizing / also the popup content might change and take more room than originally). Thoughts?

miina: Tested it out and it looks like it should work as expected with using just this:
```
...
y: Math.max(topOffset, Math.min(offsetY, maxOffsetY)),
...
```
instead of adding additional `shouldHaveDifferentOffset` boolean... With just this the popups render next to the anchor _or_ adjust if there's too few room on the screen. Tooltips seem to show ok as well.

In any case, deciding how to calculate could also be done based on the rect values instead of having to pass the prop `shouldHaveDifferentOffset` manually. E.g. -- if the room between bottom edge and the anchor is smaller than the popup height, use the adjusted `y` calculation, otherwise just the `topOffset`.

But using `Math.max(topOffset, Math.min(offsetY, maxOffsetY))` already seems to be covering it, so let's try to get rid of the `shouldHaveDifferentOffset` bool.

BrittanyIRL:
Hi there @amovar18! 

First off, thank you for your diligence here. 

I'm sorry this has been such a back and forth. The `Popup` (as you can surely surmise) is a bit slippery and has been stretched for various uses. When this dedupe ticket was created, it was really just to bring the functionality that the story editor needed for the few one off cases (primarily the color picker) and make sure that the popup in the design system could handle this. There's 3 values that the design system (ds) wasn't aware of previously: `isRTL, topOffset, invisible` - you did capture these and you have worked back through many of the fixed regressions (thanks @miina and @BrookeGraham for lending your memory and time). 

That said, this PR changes some of the structures that make me a little hesitant because of the spread of what all the popup touches. Pretty much the only thing that was really needed here was to pull in the things from the editor that didn't exist in the ds popup and bring RTL awareness to the offset and transform utils. 

Would you take a look at this draft PR I did today and let me know what you think? https://github.com/GoogleForCreators/web-stories-wp/pull/10460

packages/design-system/src/components/popup/index.js
```diff
@@ -20,98 +20,106 @@
...
 `;
-
 function Popup({
+  isRTL = false,
   anchor,
   dock,
   children,
   renderContents,
   placement = PLACEMENT.BOTTOM,
-  zIndex = DEFAULT_POPUP_Z_INDEX,
```
BrittanyIRL, line 73: being able to override the zIndex still doesn't hurt. Let's leave this be.

BrittanyIRL (commented):
This looks great!! 

I unresolved 2 of Miina's nits because they are still relevant and I would prefer to keep the zIndex override available. That has cropped up as a need in some designs. Keeping all of the available options for this as the two get combined will prevent any weird regression that we don't notice for like 2 months (😅  or at least hopefully). 

Great catches on my draft for the overflow. I really appreciate you combining the ideas. 

I'll approve this when the zIndex is back. Just trigger a 'rereview' from me - i'm usually really quick on that because I see the emails come through.

packages/design-system/src/components/popup/index.js
swissspidy: I think even ` ${({ noOverFlow }) => noOverFlow && 'overflow-y: auto;'};` works

BrittanyIRL (approved):
Looks great, thank you! Sending to QA

swissspidy:
@BrittanyIRL @felipebochehin87 Please note that external PRs are not automatically available on the QA environment because of security reasons. So I just manually created a ZIP for you here that you can manually upload to the  QA site (or wherever):
[web-stories-1.18.0-alpha.0.zip](https://github.com/GoogleForCreators/web-stories-wp/files/8025263/web-stories-1.18.0-alpha.0.zip)

PR #10198, opened 2022-01-14
Run JS unit tests and e2e tests in parallel on CI

swissspidy:
## Context

While working on parallelizing the Karma tests (see #10186) I wondered whether we could do the same for our Jest tests.

## Summary

Implements sharding for the JS unit tests and the e2e tests by leveraging a custom test sequencer class.

## Relevant Technical Choices

Adds a new private package for the shared test sequencer.

## User-facing changes

N/A

## Testing Instructions

1.

No

No

No

## Checklist

---

See #10157 and #10186

### DIFF
--- a/.github/workflows/tests-e2e.yml
+++ b/.github/workflows/tests-e2e.yml
@@ -63,13 +63,15 @@ concurrency:
 
 jobs:
   e2e:
-    name: '${{ matrix.browser }} - WP ${{ matrix.wp }}'
+    name: '${{ matrix.browser }} - WP ${{ matrix.wp }} (${{ matrix.shard }})'
     runs-on: ubuntu-latest
     timeout-minutes: 30
     continue-on-error: ${{ matrix.experimental == true }}
     strategy:
       matrix:
         # TODO: add back Firefox once support is more mature.
+        # We want to split up the tests into 2 parts running in parallel.
+        shard: ['1/2', '2/2']
         include:
           - browser: 'chrome'
             wp: 'latest'
@@ -140,6 +142,7 @@ jobs:
         run: npm run test:e2e
         env:
           WP_VERSION: ${{ matrix.wp }}
+          SHARD: ${{ matrix.shard }}
         if: ( ! matrix.snapshots ) || ( github.event.pull_request.draft == true )
 
       - name: Stop Docker environment
--- a/.github/workflows/tests-e2e.yml
+++ b/.github/workflows/tests-e2e.yml
@@ -71,7 +71,7 @@ jobs:
       matrix:
         # TODO: add back Firefox once support is more mature.
         browser: ['chrome']
-        wp: ['latest', '5.5']
+        wp: ['5.5']
         snapshots: [false]
         # We want to split up the tests into 2 parts running in parallel.
         shard: ['1/2', '2/2']
@@ -92,15 +92,6 @@ jobs:
           - browser: 'chrome'
             wp: '5.9-RC2'
             experimental: true
-            snapshots: false
-            shard: '2/2'
-        exclude:
-          - browser: 'chrome'
-            wp: 'latest'
-            snapshots: false
-            shard: '1/2'
-          - browser: 'chrome'
-            wp: 'latest'
             snapshots: false
             shard: '2/2'
 
--- a/.github/workflows/tests-e2e.yml
+++ b/.github/workflows/tests-e2e.yml
@@ -70,19 +70,39 @@ jobs:
     strategy:
       matrix:
         # TODO: add back Firefox once support is more mature.
+        browser: ['chrome']
+        wp: ['latest', '5.5']
+        snapshots: [false]
         # We want to split up the tests into 2 parts running in parallel.
         shard: ['1/2', '2/2']
         include:
           - browser: 'chrome'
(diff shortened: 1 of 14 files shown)


spacedmonkey:
I am only seeing WP 5.5 E2E testing running. 
Screenshot 2022-01-14 at 12 50 36

Is this is the desired behaviour?

swissspidy:
Hmm right, because we're using `include:` for the matrix there I probably need to add all shards manually... Ugh.

swissspidy:
Now they're all there:

Screenshot 2022-01-14 at 14 33 12

swissspidy:
Strangely the `Stories Dashboard › should be able to open the dashboard` e2e test seems to be failing here 🤔

spacedmonkey:
This looks good, but there e2e tests are now failing. I will approve once they are working again.

swissspidy:
All tests are passing now.

.github/workflows/tests-e2e.yml
```diff
@@ -63,24 +63,46 @@ concurrency:
...
+            shard: '2/2'
+        exclude:
+          - browser: 'chrome'
+            wp: 'latest'
+            snapshots: false
+            shard: '1/2'
+          - browser: 'chrome'
+            wp: 'latest'
             snapshots: false
+            shard: '2/2'
```
samwhale: Just for my own education: why do we have to build each of these out separately?

packages/e2e-tests/src/specs/dashboard/adminMenu.js
```diff
@@ -27,9 +27,9 @@ describe('Admin Menu', () => {
 
     // Initial visit to `/` makes `Dashboard` link current in WP
     await page.hover('#menu-posts-web-story');
-    await await expect(
-      page.$('#menu-posts-web-story .current a')
-    ).resolves.toMatch('Dashboard');
+    await expect(page.$('#menu-posts-web-story .current a')).resolves.toMatch(
```
samwhale, line 30: 😂 this test file is getting flagged by codecov. I can write a ticket to look into this (unless it's a known problem already)
image

packages/jest-parallel-sequencer/README.md
```diff
@@ -0,0 +1,5 @@
+# Jest Parallel Sequencer
+
+Simple custom [test sequencer](https://jestjs.io/docs/configuration#testsequencer-string) for Jest to allow for sharding on CI since Jest [does not support this out of the box](https://github.com/facebook/jest/issues/11252).
+
+Requires the `SHARD` environment variable to be set.
```
samwhale, line 5: 🥳 love documentation

packages/jest-parallel-sequencer/src/index.js
```diff
@@ -0,0 +1,49 @@
...
+   */
+  sort(tests) {
+    const chunkSize = Math.ceil(tests.length / totalShards);
+    const minIndex = shardIndex * chunkSize;
+    const maxIndex = minIndex + chunkSize;
+
+    return Array.from(tests)
+      .sort((testA, testB) => (testA.path > testB.path ? 1 : -1))
+      .slice(minIndex, maxIndex);
+  }
```
samwhale, line 46: 💭 should we run tests in random orders each time? Could be nice to keep things moving rathe than always running the same tests on the same shards. (Not sure if this already happens)

But if we want known randoms we could randomize the order based off of the current commit on main 🤔

tests/js/jest.config.js
```diff
@@ -41,6 +41,11 @@ export default {
     '**/test/**/*.[jt]s',
     '**/?(*.)test.[jt]s',
   ],
+  // @jest/test-sequencer is the default.
+  testSequencer:
+    'true' === process.env.CI
+      ? '@web-stories-wp/jest-parallel-sequencer'
+      : undefined,
```
samwhale, line 48: 🥳

samwhale (approved):
This stuff is so cool man. Learning a lot from seeing this stuff!

packages/jest-parallel-sequencer/src/index.js
swissspidy: Good idea!

Running e2e tests in random order could help catch cases like aa46d477aac4461cf4029df4a6f6a7099a37f453 where tests inadvertently depend on the specific order of tests being run. It's why we also run PHP unit tests in random order  in a separate job allowed to fail.

I do think though that this is best handled in a future enhancement, ideally when Jest supports this out of the box (see https://github.com/facebook/jest/issues/4386).

packages/e2e-tests/src/specs/dashboard/adminMenu.js
swissspidy: Updated the codecov config,  that should fix this. Thanks for flagging!

.github/workflows/tests-e2e.yml
swissspidy: Hmm I can probably simplify this.

Basically we want to run these tests:

* WP 5.5, no snapshots
* WP latest, with snapshots
* WP 5.9 RC, no snapshots, experimental

So this involves using `include` here to add the additional tests.

Thanks to your comment I realized that I could simplify this matrix a bit though.

PR #10199, opened 2022-01-14
UI Simplification: Allow selecting Layer from the element context menu

### LINKED ISSUE #10107: UI Simplification: Allow selecting Layer from the right-click menu
## Feature Description
See [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=11122%3A448445)

- [ ] Add a new menu item to the right-click menu for selecting a layer
- [ ] It displays all the Layers that are under the cursor at the time of the click
- [ ] Make sure it's possible to use both menus with keyboard (needs reworking the current navigation logic)

NOTE: We will not add the icon to the Layers currently and will only use text (and the tick for the selected layer). See [Slack message](https://xwp.slack.com/archives/C01ESBB8ZN1/p1642088410023000?thread_ts=1641470719.003600&cid=C01ESBB8ZN1).

## Additional Context
Screenshot 2022-01-06 at 12 07 46

miina:
## Summary
Adds a new item to the Context Menu (right-click menu of the element): Select Layer

## Relevant Technical Choices

- a11y: Since moving between parent-submenu happens with arrow keys, using left and right arrow keys does not work anymore for moving vertically to the next item in the menu,
- Using the same `Menu` component for rendering the submenu as well.

## To-do

- [x] Don't display Select Layer if empty results for intersecting elements
- [x] Accessibility -- Left-Right arrow keys will be only for moving between submenu and the parent menu now.
- [x] When opening the context menu from keyboard, what to display then?
- ~Refactor the right-click menu file into smaller files.~ Will be done by Pea Pod in https://github.com/google/web-stories-wp/issues/10212 instead.
- [x] Style details to make it match with Figma more precisely (adjust paddings + add Select Layer icon)
- [x] Tests

## User-facing changes
Screenshot 2022-01-17 at 16 01 55

## Testing Instructions

1. Add some elements.
2. Now add an image on top of the other elements.
3. Open the right-click menu for the image.
4. Verify it has "Select Layer" option, click on it.
5. Verify you can see the top layer (image) selected and other layers listed **that were under the place where you clicked with cursor when opening the right-click menu.** 
6. Verify that selecting a layer in the submenu will select it and close the menu.
7. Test the same for all other element types as well (single-selection only!)
8. Verify that opening the right-click menu from the Layer panel will not display the Select Layer item in the menu (since there are no elements under the cursor while doing it!)
9. Verify that opening the right-click menu on top of one element only will not display the Select Layer item in the menu (since the only element is already selected, so no point opening it!)
10. Verify that layers that are not under the cursor while opening the right-click menu will not be displayed in the submenu either.
11. Accessibility: verify you can move between the parent menu and the submenu using arrow keys.

No

Yes, tracking `context_menu_action` + `layer_selected` now.

## Checklist

---

Fixes #10107


packages/story-editor/src/app/rightClickMenu/useLayerSelect.js
```diff
@@ -0,0 +1,111 @@
...
+        key: id,
+        supportsIcon: true,
+        icon: selectedElements[0].id === id ? <Icons.CheckmarkSmall /> : null,
+        label: isBackground ? (
+          <span>{__('Background', 'web-stories')}</span>
+        ) : (
+          <LayerContent element={element} />
+        ),
+        onClick: () => setSelectedElementsById({ elementIds: [id] }),
+        ...menuItemProps,
```
miina, line 118: @BrittanyIRL This PR is still WIP but wanted to tag you here since your pod has mainly worked on the element context menu.
Let me know if you see anything off with adding the sub-menu in this way. I tried to keep similar logic as it currently is: defining the props in the `app` folder and rendering these as needed elsewhere.

Also, the `rightClickMenu` provider file is quite massive currently with 1200+ lines, I'm planning to refactor it into smaller files within this PR, let me know if you have any concerns with that.

BrittanyIRL: I'll look when i get back on Tuesday, thank you for the heads up. We would have been happy to help with the menu adds and can take the refactor to make smaller files if you would like.

miina: Thanks, Brittany. Sure, I'll add an issue to refactor the provider and assign it to you.

Edit: Added https://github.com/google/web-stories-wp/issues/10212

merapi (approved):
Works great 👍🏻 

Some random thoughts after using it for the first time:
1. Should the submenu open on hover? I was waiting for it to open, I guess because all other submenus (OS, Chrome, Figma) do that.

2. Should we change the order of layers to match the Layers panel (background at the bottom)?

image

3. Tested it with VoiceOver and it's not obvious that you can expand that Select Layer option, I've looked at [this](https://www.w3.org/TR/2019/WD-wai-aria-practices-1.2-20191218/examples/toolbar/toolbar.html) example (font selection or just right-click menu in browser) and I think it should say "menu pop-up" or "submenu" so you know that you can expand it (and then e.g. "expanded"), thoughts?

miina:
@merapi 

> Should the submenu open on hover? I was waiting for it to open, I guess because all other submenus (OS, Chrome, Figma) do that.

Hm, perhaps. I was also wondering about that. I'll confirm today with UX.

> Should we change the order of layers to match the Layers panel (background at the bottom)?

Good point, yes.

> Tested it with VoiceOver and it's not obvious that you can expand that Select Layer option, I've looked at this example (font selection or just right-click menu in browser) and I think it should say "menu pop-up" or "submenu" so you know that you can expand it (and then e.g. "expanded"), thoughts?

Adding "expanded" vs "collapsed". For some reason I thought that `aria-haspopup="menu"` should announce that it has menu popup, but it doesn't. Trying to figure out what adds that.

packages/story-editor/src/components/canvas/rightClickMenu.js
```diff
@@ -89,19 +157,30 @@ const RightClickMenu = () => {
...
+                        isRTL={isRTL}
+                        aria-expanded
+                        isOpen
+                        isSubMenu
+                      >
+                        {subMenuItems.map((item) => menuButtonRenderer(item))}
+                      </SubMenuWrapper>
                     )}
-                  </ContextMenuComponents.MenuButton>
+                  </ButtonWrapper>
```
samwhale: 💭 I think we don't need to do this.

The goal of the refactor was to be able to create context menus with more flexibility. Included in this was the idea that we could _nest_ context menus together.

```js

<RightClickMenuContainer position={menuPosition} ref={ref}>
  <DirectionAware>
    <ContextMenu
      data-testid="right-click-context-menu"
      isOpen={isMenuOpen}
      onDismiss={onCloseMenu}
      aria-label={__('Context Menu for the selected element', 'web-stories')}
      maskRef={maskRef}
      onMouseDown={(evt) => evt.stopPropagation()}
      isRTL={isRTL}
    >
      {rightClickMenuItems.map(
        ({ label, shortcut, separator, subMenuItems, ...buttonProps }) => (
          <Fragment key={label}>
            {separator === 'top' && <ContextMenuComponents.MenuSeparator />}
            <ContextMenuComponents.MenuButton {...buttonProps}>
              {label}
              {shortcut && (
                <ContextMenuComponents.MenuShortcut>
                  {shortcut.display}
                </ContextMenuComponents.MenuShortcut>
              )}
            </ContextMenuComponents.MenuButton>
            {Boolean(subMenuItems?.length) && (
               {/* The position of the new menu is defined here */}
              <RightClickMenuContainer position={{ y: 0, x: 430 }}>
                <ContextMenu
                  data-testid="right-click-context-menu"
                  isOpen={isMenuOpen}
                  // No `onDismiss` so clicking the layer buttons won't close the menu
                  // if that is the desired behavior
                  aria-label={__(
                    'Select a layer',
                    'web-stories'
                  )}
                  onMouseDown={(evt) => evt.stopPropagation()}
                  isRTL={isRTL}
                >
                  {subMenuItems.map(
                    ({
                      label: subLabel,
                      shortcut: subShortcut,
                      icon,
                      supportsIcon,
                      SuffixIcon,
                      ...nestedButtonProps
                    }) => (
                      <ContextMenuComponents.MenuButton
                        key={subLabel}
                        {...nestedButtonProps}
                      >
                        <ContextMenuComponents.MenuLabel>
                          {icon}
                          {subLabel}
                        </ContextMenuComponents.MenuLabel>
                        {shortcut && (
                          <ContextMenuComponents.MenuShortcut>
                            {subShortcut.display}
                          </ContextMenuComponents.MenuShortcut>
                        )}
                        {SuffixIcon && (
                          <SuffixIconWrapper>
                            <SuffixIcon />
                          </SuffixIconWrapper>
                        )}
                      </ContextMenuComponents.MenuButton>
                    )
                  )}
                </ContextMenu>
              </RightClickMenuContainer>
            )}
            {separator === 'bottom' && <ContextMenuComponents.MenuSeparator />}
          </Fragment>
        )
      )}
    </ContextMenu>
  </DirectionAware>
</RightClickMenuContainer>;
```

The above code should render the menu without needing the extra styles:
image

Some massaging will be needed (like adding an icon to the button that expands the sub menu, and accessibility work).

Was there a reason that you didn't want to do it this way that I'm missing?

packages/story-editor/src/app/rightClickMenu/useLayerSelect.js
```diff
@@ -0,0 +1,132 @@
...
+  // Only display if submenu has any items.
+  return subMenuItems.length > 0
+    ? {
+        label: __('Select Layer', 'web-stories'),
+        separator: 'bottom',
+        onClick: () => setIsSubMenuOpen(!isSubMenuOpen),
+        subMenuItems: isSubMenuOpen ? subMenuItems : [],
+        dismissOnClick: false,
+        SuffixIcon: Icons.ChevronRightSmall,
+        'aria-haspopup': 'menu',
```
miina: @barklund Perhaps you know why this is not announced by the VoiceOver as "menu popup"? I assumed using `aria-haspopup` would do that, however, it's not happening. Any thoughts on what might be wrong here?

packages/story-editor/src/components/canvas/rightClickMenu.js
```diff
@@ -42,6 +42,45 @@ const RightClickMenuContainer = styled.div`
...
+  }
+`;
+
+const SuffixIconWrapper = styled.div`
+  svg {
+    width: 32px;
+    color: ${({ theme }) => theme.colors.fg.secondary};
+    margin: -12px 0 !important;
+  }
+`;
```
samwhale: 🤔 If we're to follow nesting context menus like the comment below then all of these styles shouldn't be necessary

There's a component defined in the `contextMenu` folder called `Label` (exported as `MenuLabel`). Styling this may help position and size icons in the sub menu. Otherwise you could style the `Button` directly to deal with icons as children.

packages/story-editor/src/components/canvas/rightClickMenu.js
```diff
@@ -89,19 +157,30 @@ const RightClickMenu = () => {
...
-                  <ContextMenuComponents.MenuButton {...buttonProps}>
-                    {label}
-                    {shortcut && (
-                      <ContextMenuComponents.MenuShortcut>
-                        {shortcut.display}
-                      </ContextMenuComponents.MenuShortcut>
+                  <ButtonWrapper>
+                    {menuButtonRenderer({ label, shortcut, ...buttonProps })}
+                    {Boolean(subMenuItems?.length) && (
+                      <SubMenuWrapper
```
samwhale: 💭 this needs an `aria-pressed` attribute. With using a nested context menu I believe this would still be true.

image

packages/design-system/src/components/contextMenu/menu.js
```diff
@@ -113,33 +126,64 @@ const Menu = ({
...
+      if (
+        !isSubMenu &&
+        (KEYS.RIGHT === key || (isRTL && KEYS.LEFT === key)) &&
+        prevItemHasSubMenuOpen
+      ) {
+        const subMenu = menuRef.current.querySelector('[role="menu"]');
+        if (subMenu) {
+          const subMenuItems = getFocusableChildren(subMenu, true);
+          subMenuItems[0]?.focus();
+        }
```
samwhale: It's great that you can move between menu and sub menu with the arrow keys. However, for keyboard accessibility I believe it would be best if a user could open the sub menu using the left/right arrow keys as well.

Radix ui has an excellent context menu that I use for reference on keyboard accessibility and component design. https://www.radix-ui.com/docs/primitives/components/context-menu

packages/story-editor/src/app/rightClickMenu/useLayerSelect.js
```diff
@@ -0,0 +1,132 @@
...
+      setSelectedElementsById,
+    })
+  );
+
+  useEffect(() => {
+    // Close submenu if the menu itself also closes.
+    if (!isMenuOpen) {
+      setIsSubMenuOpen(false);
+    }
+  }, [isMenuOpen]);
```
samwhale, line 65: 👍

packages/story-editor/src/app/rightClickMenu/useLayerSelect.js
```diff
@@ -0,0 +1,132 @@
...
+        const node = nodesById[id];
+        if (!node) {
+          return null;
+        }
+        const { x: elX, y: elY, width, height } = node.getBoundingClientRect();
+        const elementP = createPolygon(rotationAngle, elX, elY, width, height);
+        return SAT.pointInPolygon(clickedPoint, elementP) ? element : null;
+      })
+      .filter((el) => el);
+  }, [currentPage, x, y, nodesById]);
```
samwhale: This is cool. @littlemilkstudio do you think this should use a `ResizeObserver`? May be hard to tell if it's necessary without doing some performance auditing.

May not be necessary for now and we can always follow up later.

packages/story-editor/src/components/canvas/rightClickMenu.js
```diff
@@ -72,6 +111,35 @@ const RightClickMenu = () => {
...
+      {!supportsIcon && label}
+      {shortcut && (
+        <ContextMenuComponents.MenuShortcut>
+          {shortcut.display}
+        </ContextMenuComponents.MenuShortcut>
+      )}
+      {SuffixIcon && (
+        <SuffixIconWrapper>
+          <SuffixIcon />
+        </SuffixIconWrapper>
```
samwhale: 🤔 The chevron should to be flush with the shortcuts

|Designs|Right Click Menu|
|--|--|
|image|image|

packages/story-editor/src/components/canvas/rightClickMenu.js
```diff
@@ -72,6 +111,35 @@ const RightClickMenu = () => {
...
+        </LabelWrapper>
+      )}
+      {!supportsIcon && label}
+      {shortcut && (
+        <ContextMenuComponents.MenuShortcut>
+          {shortcut.display}
+        </ContextMenuComponents.MenuShortcut>
+      )}
+      {SuffixIcon && (
+        <SuffixIconWrapper>
```
samwhale: If the `SuffixWrapper` is necessary then this should exist in `packages/design-system/src/components/contextMenu/components/` to follow the existing pattern.

All context menu components that aren't one-off things should be defined there for re-usability. I believe the `SuffixWrapper` will be one of these things (similar to shortcut) 😄

samwhale (changes requested):
Left some suggestions for improvements following the new patterns implemented by #9960. If you disagree with any let me know!

Also, I saw one strange error that I couldn't reproduce. When opening the submenu with the keyboard and then pressing `Esc` the app crashed. I haven't been able to reproduce it, but just a heads up in case you see it too!

samwhale:
>Should the submenu open on hover? I was waiting for it to open, I guess because all other submenus (OS, Chrome, Figma) do that.

I'd use the [radix-ui context menu](https://www.radix-ui.com/docs/primitives/components/context-menu) as a reference for behavior. Checking with design about it is a good idea too, but hover should open the sub menu imo.

packages/story-editor/src/components/canvas/rightClickMenu.js
miina: Hm, I tried originally with using a second `ContextMenu` but it didn't work as expected -- wasn't aware of the `RightClickMenuContainer` though! Probably that's why.

I'd still create a separate menuItemRenderer though that's shared between the menus since they're using exactly the same style. Any reason why you think it's better not to?

packages/story-editor/src/components/canvas/rightClickMenu.js
miina: The message says that `aria-pressed` is **not allowed though** not that it's needed. Will take a look.

packages/story-editor/src/components/canvas/rightClickMenu.js
miina: > The chevron should to be flush with the shortcuts

Not sure what you mean with this, could you elaborate? Do you mean that we should combine the shortcuts and chevron into one suffix instead of using shortcuts + suffix?

packages/story-editor/src/components/canvas/rightClickMenu.js
samwhale: Yeah that's the magic touch! Just setting the position using the containing div should work.

`menuItemRenderer`: Sure that's fine! Maybe we could even make this its own component?

packages/story-editor/src/components/canvas/rightClickMenu.js
samwhale: Oh sorry, what I mean by that is that the right edge of both the chevron and the shortcuts should be 15px from the edge of the context menu:
image

(really 16px since we round to 4s in the design system)

miina: Ah, okay, now I understand. Yep, I think there are inconsistencies currently in Figma, it has always been an even number at least. Will adjust, thanks!

packages/story-editor/src/components/canvas/rightClickMenu.js
samwhale: Hmm weird. Radix seems to match this implementation too:
image

you're right, `aria-pressed` should not be needed

packages/design-system/src/components/contextMenu/menu.js
miina: Yep, confirmed with UX that left/right arrow keys should open the menu as well, also the moving out of the menu with left-right arrow key will close the menu.

And should open on hover, clicking on `Select Layer` will not do anything with mouse.

Thanks for the suggestions 👍

packages/story-editor/src/components/canvas/rightClickMenu.js
```diff
@@ -42,32 +42,23 @@ const RightClickMenuContainer = styled.div`
...
+      margin-left: -12px;
+    }
+    span {
+      padding-left: 18px;
+      font-size: 12px;
+      white-space: nowrap;
+      overflow: hidden;
+      text-overflow: ellipsis;
+    }
+  `}
```
samwhale: 💭 Maybe this should be `MenuItemPrefix` in case we want to use this again later?

miina: I guess it could be `MenuItemWithPrefix` since it would influence all the buttons of that menu. Or maybe `MenuItemPrefix` could be a prefix which behaves as a space placeholder as well, have to figure out the styling though. I'll check that as the last thing, will try to sort out the hovering part first since that might need a few changes.

packages/design-system/src/components/contextMenu/components/suffix.js
```diff
@@ -0,0 +1,39 @@
...
+  svg {
+    width: 32px;
+    color: ${({ theme }) => theme.colors.fg.secondary};
+    margin: -12px 0 !important;
+  }
+`;
+
+function Suffix({ children }) {
+  return <Wrapper>{children}</Wrapper>;
+}
```
barklund: What's the difference between `Wrapper` and `Suffix` here? Why not just `export default Wrapper` and be done with it?

packages/story-editor/src/components/canvas/rightClickMenu.js
samwhale: It's a lot of new stuff for the context menu but I think it's all good stuff. I tend to make a PR to the design system first and then implement it later in the editor

Faster testing and feedback in storybook too

packages/design-system/src/components/contextMenu/menu.js
```diff
@@ -55,22 +55,34 @@ MenuWrapper.propTypes = {
...
+  const allButtons = Array.from(
+    parent.querySelectorAll(FOCUSABLE_SELECTORS.join(', '))
+  );
+  if (isSubMenu) {
+    return allButtons;
+  }
+  // Skip considering the submenu, and the submenu items.
+  return allButtons.filter(
+    (elem) => !elem.matches('[role="menu"], [role="menu"] [role="menu"] *')
+  );
```
barklund, line 74: Could this not be added permanently to the `FOCUSABLE_SELECTORS` - that they should be something like `button:not(:disabled):not([role=menu])`? Or maybe it's the last entry, the tab index one, that also includes the menu?

If not, alternatively make two different arrays of selectors, and just use the proper one here:

```
function getFocusableChildren(parent) {
  const selectors = isSubMenu ? FOCUSABLE_SUBMENU_SELECTORS : FOCUSABLE_SELECTORS;
  return Array.from(parent.querySelectorAll(selectors.join(', ')));
```

packages/design-system/src/components/contextMenu/menu.js
barklund: Just typed this same suggestion in another comment, now deleted it. I agree with all the parts mentioned.

One thing though, when displaying the submenu on hover, make sure to close the menu again only after a short delay on mouseout (e.g. 300ms).

If you're dragging the mouse directly from the parent item to an item further down the submenu list, you might inadvertently pass through other items in the parent list, but you don't actually intend to close the menu because of this.

(I also think the submenu should only open after a short delay)

packages/story-editor/src/app/rightClickMenu/useLayerSelect.js
barklund: Went quite down the rabbit hole on this one.

VoiceOver should inform users of this property, and it does so _if_ the role is button, but _not_ if the role is menuitem. At least that's what I can see by playing around with VO.

Please see [this table](https://a11ysupport.io/tech/aria/aria-haspopup_attribute#support-table-4) and [this reference test](https://a11ysupport.io/tests/html/aria/aria-haspopup_examples.html). If you add a `role="menuitem"` on the test page to the buttons, VO stops reporting that it's a menu popup button and just says that it's a menu item.

So I think that we are in the clear here - we have done as much as we can, it's just VO that aren't good enough at its job.

packages/story-editor/src/app/rightClickMenu/useLayerSelect.js
```diff
@@ -0,0 +1,132 @@
...
+  return subMenuItems.length > 0
+    ? {
+        label: __('Select Layer', 'web-stories'),
+        separator: 'bottom',
+        onClick: () => setIsSubMenuOpen(!isSubMenuOpen),
+        subMenuItems: isSubMenuOpen ? subMenuItems : [],
+        dismissOnClick: false,
+        SuffixIcon: Icons.ChevronRightSmall,
+        'aria-haspopup': 'menu',
+        'aria-pressed': isSubMenuOpen,
```
barklund: I'm not convinced that we need `aria-pressed` here. `aria-expanded` should do the right job.

packages/design-system/src/components/contextMenu/menu.js
```diff
@@ -113,33 +126,64 @@ const Menu = ({
...
 
       if (prevIndex === -1 && focusableChildren.length) {
         setFocusedId(focusableChildren[0].id);
-        return;
+        prevIndex = 0;
       }
 
-      const isAscending = [KEYS.UP, KEYS.LEFT].includes(key);
-      let newIndex = prevIndex + (isAscending ? -1 : 1);
+      const prevItemHasSubMenuOpen =
```
barklund: Not sure where exactly to add this comment, so adding it here because why not.

When you focus the parent item with a submenu and press enter to open the submenu, the first item in the submenu should receive focus, the focus should not remain on the parent item as it does now.

packages/story-editor/src/app/rightClickMenu/useLayerSelect.js
```diff
@@ -0,0 +1,132 @@
...
+
+  // Only display if submenu has any items.
+  return subMenuItems.length > 0
+    ? {
+        label: __('Select Layer', 'web-stories'),
+        separator: 'bottom',
+        onClick: () => setIsSubMenuOpen(!isSubMenuOpen),
+        subMenuItems: isSubMenuOpen ? subMenuItems : [],
+        dismissOnClick: false,
+        SuffixIcon: Icons.ChevronRightSmall,
```
barklund: This icon should change direction in RTL:
Screen Shot 2022-01-19 at 12 09 29

barklund (commented):
This is awesome stuff! You made a lot happen here. There are just a few things regarding the a11y, but it's very close! 🎉

packages/story-editor/src/app/rightClickMenu/useLayerSelect.js
```diff
@@ -0,0 +1,132 @@
...
+        const { x: elX, y: elY, width, height } = node.getBoundingClientRect();
+        const elementP = createPolygon(rotationAngle, elX, elY, width, height);
+        return SAT.pointInPolygon(clickedPoint, elementP) ? element : null;
+      })
+      .filter((el) => el);
+  }, [currentPage, x, y, nodesById]);
+
+  const subMenuItems = useMemo(() => {
+    const intersectingElements = getIntersectingElements();
+    intersectingElements.reverse();
```
BrittanyIRL: 🤔  why not just have them returned from `getIntersectingElements` already reversed?

BrittanyIRL (commented):
Well, I'm late to the review party but I echo Morten and Samuel and will happily pull this down once their updates are made and run through this. Anything I add right now is redundant.

packages/story-editor/src/app/rightClickMenu/useLayerSelect.js
miina: It just felt like `getIntersectingElements` was for just getting the intersecting elements and using those elements doesn't concern the function itself. I could of course change it to `getReversedIntersectingElements`.

packages/story-editor/src/app/rightClickMenu/useLayerSelect.js
miina: Thank you for taking a look!

packages/design-system/src/components/contextMenu/menu.js
miina: I tried out different selectors for being able to ignore items that are children of menu but didn't find a selector, thus opted for filtering out the submenu items if it's not the submenu.

The problem is that `button:not(:disabled)` inside the parent will also select all the non-disabled buttons of the submenu inside that menu. Even if ignoring the submenu itself as an item is doable, couldn't find a simple way to ignore the items inside the submenu. Anyway, I can try again.

miina: OK, so I tried a bunch of different options again creating a separate selector array for the parent menu, e.g. `:not([role="menu"]) button...` and `...:not([role="menu"])` etc. but these didn't work, still the submenu items are getting included. Let me know if you know a selector which would exclude children of a determined parent ([role="menu"] in this case).

packages/design-system/src/components/contextMenu/menu.js
miina: @samwhale Currently, when the submenu is opened, the first item is receiving focus, however, it doesn't have visible focus. Any thoughts on how to make sure that it also has visible focus? Tried [using `setFocusedId`](https://github.com/google/web-stories-wp/pull/10199/files#diff-4b4f2ec27b2ca00e1e22c7773f5269fc5ff36c65681344d1af0fe60bbf3a56d8R103) but without luck:
```
      subMenuItems[0].focus();
      setFocusedId(subMenuItems[0].id);
```

packages/story-editor/src/components/canvas/rightClickMenu.js
miina: @samwhale I've added new Story now for the menu with submenu as well, PTAL (WithSubMenu).

There are probably still things to improve, e.g. currently `firstChild` is used in quite a few places (to get the actual menu, not just the container where the ref is), not sure if there's a better alternative.

packages/design-system/src/components/contextMenu/components/item.js
```diff
@@ -0,0 +1,79 @@
...
+          <SuffixIcon />
+        </Suffix>
+      )}
+    </StyledButton>
+  );
+}
+
+const MenuItem = forwardRef(MenuItemWithRef);
+
+MenuItemWithRef.propTypes = MenuItem.propTypes;
```
barklund: But this hasn't been assigned yet? Should this swap with the next statement?

packages/design-system/src/components/contextMenu/components/subMenuTrigger.js
```diff
@@ -0,0 +1,170 @@
...
+    pointerTracker.current.x = e.clientX;
+    pointerTracker.current.y = e.clientY;
+  }, []);
+
+  useEffect(() => {
+    document.addEventListener('pointermove', trackPointer);
+    return () => {
+      document.removeEventListener('pointermove', trackPointer);
+    };
+  }, [trackPointer]);
```
barklund: Do we really need to track every movement of the pointer anywhere while a context menu is displayed? Could we limit it to a specific element and add it as a react listener (`onPointerMove`)?

packages/design-system/src/components/contextMenu/components/item.js
```diff
@@ -0,0 +1,79 @@
...
+
+MenuItem.propTypes = {
+  label: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
+  shortcut: PropTypes.object,
```
BrittanyIRL: Adding shape to this and `SuffixIcon` since they are in the design system and should be able to be reused would be really helpful for our future selves. 
```suggestion
  shortcut: PropTypes.shapeOf({display: PropTypes.string, whatever else is needed}),
```

packages/design-system/src/components/contextMenu/components/subMenuTrigger.js
```diff
@@ -0,0 +1,170 @@
...
+  };
+
+  const maybeCloseSubMenu = useDebouncedCallback(() => {
+    if (!ref.current || !subMenuRef.current?.firstChild) {
+      return;
+    }
+    // If after 200ms the cursor is not in the submenu and not in itself, leave.
+    if (
+      pointerIsOutside(ref.current) &&
+      pointerIsOutside(subMenuRef.current.firstChild)
```
barklund, line 84: Something here isn't right - or this function isn't correctly invoked at the right times. When you move the pointer out of the submenu quickly, it doesn't close. It only closes if you move it a little bit out. Not exactly sure what's going on here though, it seems a bit inconsistent.

maybe-close-bug

packages/design-system/src/components/contextMenu/menu.js
barklund: > Let me know if you know a selector which would exclude children of a determined parent (`[role="menu"]` in this case).

Ah no, you can't do that. Maybe with `:has()`, but that isn't supported yet.

It wasn't a big deal, just looked like something that could potentially be optimized a bit. But if not, fine with me!

packages/design-system/src/components/contextMenu/menu.js
```diff
@@ -113,33 +128,59 @@ const Menu = ({
...
+        return;
+      }
+      // Maybe move from submenu to parent menu.
+      if (
+        isSubMenu &&
+        ((!isRTL && KEYS.LEFT === key) || (isRTL && KEYS.RIGHT === key))
+      ) {
+        // Get the button with expanded popup.
+        const parentButton = parentMenuRef.current.querySelector(
+          'button[aria-expanded="true"][aria-haspopup]'
```
barklund: Hm, to be nitpicky, we don't just want the element to have the `aria-haspopup` property, as it (despite its `has` prefix) isn't a boolean property but an enum, so it should have the right property as well. But on the other hand, only something with a popup would have the expanded property, so we could actually just remove the last condition?

```js
          'button[aria-expanded="true"]'
```

packages/design-system/src/components/contextMenu/stories/index.js
```diff
@@ -559,6 +560,76 @@ const RightClickContextMenuContainer = styled.div`
...
+        <MenuItems.MenuSeparator />
+        <MenuItems.MenuButton onClick={action('Clicked on `i am a button`')}>
+          {'i am a button!'}
+          <MenuItems.MenuShortcut>{'⌥ ⌘ A'}</MenuItems.MenuShortcut>
+        </MenuItems.MenuButton>
+      </ContextMenu>
+    </Container>
+  );
+};
+
```
BrittanyIRL, line 689: Nit: it would be nice to keep this view current with the editor as well: 
Screen Shot 2022-01-21 at 3 47 03 PM

barklund (approved):
This works really well! 🚀

packages/story-editor/src/app/rightClickMenu/useLayerSelect.js
```diff
@@ -0,0 +1,138 @@
...
+const ReversedIcon = styled(Icons.ChevronRightSmall)`
+  transform: rotate(180deg);
+`;
+
+function useLayerSelect({ menuItemProps, menuPosition, isMenuOpen }) {
+  const { isRTL } = useConfig();
+  const [isSubMenuOpen, setIsSubMenuOpen] = useState(false);
+  const { nodesById } = useCanvas(({ state: { nodesById } }) => ({
+    nodesById,
+  }));
```
BrittanyIRL: nit: you shouldn't need to deconstruct here since it's just `nodesById` 🤷‍♀️  
```suggestion
  const nodesById  = useCanvas(({ state: { nodesById } }) => nodesById); 
```

packages/story-editor/src/components/canvas/karma/rightClickMenu.karma.js
```diff
@@ -436,6 +453,52 @@ describe('Right Click Menu integration', () => {
     });
   });
 
+  describe('Right click menu: Select Layer', () => {
```
BrittanyIRL, line 451: 🎉

BrittanyIRL (commented):
Tested in Chrome and Firefox with both a cursor and a keyboard and it works as described! 

Two questions:

Should the 'select layer' show when there's only 1 layer to select? 
Screen Shot 2022-01-21 at 3 28 02 PM

Should the checkmark on the selected layer flip when in RTL vs LTR? Right now it stays in the same direction.

miina:
> Should the 'select layer' show when there's only 1 layer to select?

I was also wondering about that. Perhaps not, it doesn't really give anything to the user, will change.

> Should the checkmark on the selected layer flip when in RTL vs LTR? Right now it stays in the same direction.

Saw it stays the same in other places (e.g. font picker) so thought I'd keep it here the same, too.

packages/design-system/src/components/contextMenu/components/subMenuTrigger.js
miina: Made some changes, now tracking `pointermove` of the parent menu. Still with adding event listener since this is handled inside the trigger item and only in case the submenu is open -- the parent menu doesn't have to be aware of that.

packages/design-system/src/components/contextMenu/components/subMenuTrigger.js
miina: Should be fixed now, PTAL.

packages/design-system/src/components/contextMenu/stories/index.js
miina: @BrittanyIRL I've added a new static menu to there: "Right click menu on top of shape and background"

Screenshot 2022-01-24 at 14 29 07

Since Select Layer is not displayed unless it's on top of the selected element + one more then perhaps that's more accurate.

packages/design-system/src/components/contextMenu/components/item.js
```diff
@@ -0,0 +1,82 @@
...
+const MenuItem = forwardRef(MenuItemWithRef);
+
+MenuItem.propTypes = {
+  label: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
+  shortcut: PropTypes.shape({
+    display: PropTypes.object.isRequired,
+    title: PropTypes.string,
+  }),
+  icon: PropTypes.node,
+  SuffixIcon: PropTypes.object,
```
BrittanyIRL, line 77: nit: here too - `PropTypes.shape`

BrittanyIRL (approved):
Thanks for the updates! One extremely minor nit, but otherwise 👍 👍

packages/design-system/src/components/contextMenu/components/subMenuTrigger.js
```diff
@@ -0,0 +1,170 @@
...
+      ref={ref}
+      onPointerEnter={openSubMenu}
+      onPointerLeave={() => {
+        // Reset tracker in case we moved out of the menu fully.
+        pointerTracker.current = {};
+        maybeCloseSubMenu();
+      }}
+      onClick={(e) => e.preventDefault()}
+      aria-haspopup
+      aria-expanded={isSubMenuOpen}
```
samwhale, line 155: 👍

packages/design-system/src/components/contextMenu/components/suffix.js
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * External dependencies
+ */
+import styled from 'styled-components';
+
+const Suffix = styled.div`
+  svg {
+    width: 32px;
+    color: ${({ theme }) => theme.colors.fg.secondary};
+    margin: -12px -14px -12px 0 !important;
```
samwhale: Nit: would be good to not have to use `!important` here. maybe we could absolutely position the svg inside of the `Suffix` component?

samwhale (approved):
Accessibility looks good, thanks for making those fixes.

packages/design-system/src/components/contextMenu/components/item.js
miina: This is actually the icon directly (e.g. `Icons.Chevron`), so no shape option to add here.

packages/design-system/src/components/contextMenu/components/suffix.js
miina: Good idea! Changed.

packages/design-system/src/components/contextMenu/components/subMenuTrigger.js
barklund: Awesome, works perfect! 🎉

PR #10205, opened 2022-01-14
Context Menu: Clear text styles

### LINKED ISSUE #10018: Context Menu: Add functionality to clear text styles
## Feature Description

Text styles cannot be cleared because the menu item doesn't exist! The code has already been written to reset text styles to default but it's just not being applied to text elements.

1. Add the menu item `Clear text styles` for text styles and make sure that text styles can be cleared
2. Make sure text styles can be cleared when multiple elements are selected

## Additional Context

This came up in #9885 QA: https://github.com/google/web-stories-wp/issues/9885#issuecomment-995321400

samwhale:
## Context

Clearing styles from text elements wasn't enabled.

## Summary

Enable clearing styles from text elements.

## Testing Instructions

1. Insert a text element into the canvas
2. Right click the text element
3. Select `Clear Text Styles`. Text should be reset to 'default styles' (not specific to preset. All text elements will be reset to this configuration)
4. Add text or revert to original style
5. Select multiple elements of different types
6. Right click on the frame and select `Clear Styles`
7. Text element (and other elements) should be reverted to their style defaults.

## Checklist

---

Fixes #10018


packages/story-editor/src/app/rightClickMenu/provider.js
```diff
@@ -613,20 +611,20 @@ function RightClickMenuProvider({ children }) {
...
 
     // only show snackbar if any elements had styles reset
     if (stylesReset) {
       showSnackbar({
         actionLabel: __('Undo', 'web-stories'),
         dismissible: false,
-        message: __('Cleared style.', 'web-stories'),
+        message: _n(
+          'Cleared style.',
+          'Cleared styles.',
```
samwhale: 💭 I think we don't need this to be plural. What do y'all think?

packages/story-editor/src/components/canvas/karma/rightClickMenu.karma.js
```diff
@@ -1283,159 +1375,158 @@ describe('Right Click Menu integration', () => {
...
-      lineHeight: 1.4,
-      textAlign: 'center',
-      border: {
-        left: 1,
-        right: 1,
-        top: 1,
-        bottom: 1,
-        lockedWidth: true,
-        color: {
+    it('should clear styles for all foreground elements', async () => {
```
samwhale, line 1379: moved this test up into the `multiple elements selected` describe block

packages/story-editor/src/app/rightClickMenu/provider.js
maxyinger: yea I think either makes sense

maxyinger (commented):
I think the code and tests look great! 

Only issue I'm running into behaviorally is that any style that is both resetable and effects the size of the elements selection box, causes the selection box to be out of sync with the element when reset.

Noticing this with attributes like `fill` or `font-size`

packages/story-editor/src/app/rightClickMenu/provider.js
BrittanyIRL: In english it makes sense but we need both so it can be translated properly

BrittanyIRL (commented):
Can we take this opportunity to adjust what's getting cleared for text elements? 

I think that font weight should get reset to 'regular' (if available). 
In our design sync we also discussed disabling the "clear" option if nothing is available to clear. Happy to make that it's own ticket since it's not initially part of this but wanted to mention it here. 

Should the 'default' that text gets set back to be the text element that shows up if you click the plus sign to the right of the 't' icon? 

Thanks for working through this update, it'll be nice to get this all sorted out.

packages/story-editor/src/app/rightClickMenu/provider.js
samwhale: I think in english we'd still say 'cleared styles' if one singular style was cleared. However, this is based on the number of selected elements

So clearing styles from one element would show `Cleared style.` even if multiple styles were cleared.

I think this should always be `styles` in english. However in other languages it may be necessary to make that differentiation.

BrittanyIRL: oh, i thought you were asking if we needed the singular/plural structure for translation not asking what the proper english phrasing would be here. 

with your additional explanation I agree with you it should be styles.

BrookeGraham (approved):
💅 It is beautiful.

packages/story-editor/src/app/rightClickMenu/test/utils.js
```diff
@@ -57,12 +58,14 @@ const expectedTextStyles = objectPick(ALL_PROPERTIES, [
...
-  'padding',
-  'textAlign',
   'font',
   'fontSize',
+  'fontWeight',
   'lineHeight',
+  'opacity',
+  'padding',
+  'textAlign',
+  'width',
```
samwhale, line 68: Sizing box still out of sync - need to fix that separately

samwhale:
After a discussion with @agingoldseco and @BrittanyIRL it was decided that the 'clearing styles' actions are not necessary. This ticket will be closed, and another one will be added to remove that action from the right click menu.

PR #10206, opened 2022-01-14
Editor: Fix image src desync

### LINKED ISSUE #10192: Bug: Image element src not in sync when replaced with other media
Found in this thread here once we fixed some bugs with the animation provider:
https://github.com/google/web-stories-wp/pull/10024#issuecomment-1012362521

When replacing an image element with new media, the src remains on the old resource while the srcset updates to the new resource.

Visually there are no issues because the browser pulls the currentSrc from the srcset instead of the src, but src resource shouldn't get out of sync with the active resource. 

Completion of this ticket involves updating `story-editor/src/elements/media/imageDisplay.js` to maintain the active resource as its src and to re-enable these karma tests to make sure they're passing:
- should correctly copy image to page with existing image
- should correctly handle image dragged from library straight to edge replacing old image
- should correctly handle image dropped on edge replacing old image
- should replace top image when bg image is set and another one is on top
- should replace the top image when two images are in the same place on canvas

This behavior can't be seen on the canvas until this PR is merged, but you can view it on main right now in the carousel page preview after replacing media of an element:
https://github.com/google/web-stories-wp/pull/10024

merapi:
## To-do

- [ ] Enable tests

## Testing Instructions

1.

## Checklist

---

Fixes #10192

### DIFF
--- a/packages/story-editor/src/elements/media/imageDisplay.js
+++ b/packages/story-editor/src/elements/media/imageDisplay.js
@@ -108,6 +108,8 @@ function ImageDisplay({ element, box, previewMode }) {
           // Ignore
         }
       });
+    } else {
+      setSrc(resource.src);
     }
     return () => {
       mounted = false;
--- a/packages/story-editor/src/components/canvas/karma/backgroundCopyPaste.karma.js
+++ b/packages/story-editor/src/components/canvas/karma/backgroundCopyPaste.karma.js
@@ -168,9 +168,7 @@ describe('Background Copy Paste integration', () => {
     expect(await getCanvasBackgroundElement()).toBeEmpty();
   });
 
-  // TODO https://github.com/google/web-stories-wp/issues/10192
-  // eslint-disable-next-line jasmine/no-disabled-tests
-  xit('should correctly copy image to page with existing image', async () => {
+  it('should correctly copy image to page with existing image', async () => {
     // Arrange the backgrounds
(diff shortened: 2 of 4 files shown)


packages/story-editor/src/elements/media/imageDisplay.js
```diff
@@ -108,6 +108,8 @@ function ImageDisplay({ element, box, previewMode }) {
           // Ignore
         }
       });
+    } else {
+      setSrc(resource.src);
     }
```
merapi, line 113: Previous code was only settings the `src` when the full-size image was not loaded (so it was preloading the image and then settings the `src`) and did nothing when we already had full-size in resourseList (and component was updated, the newly mounted component was ok), now, if we have the full-size image we just set the `src` (it's there because we first see the preview of it when dropping?).

maxyinger: > it's there because we first see the preview of it when dropping?

✅  Yea this is the behavior I was seeing when I was logging it out yesterday

maxyinger (approved):
Awesome. Thanks for the quick fix 🙌

barklund (approved):
Seems to do what it claims to. It's a weird issue given that it doesn't directly impact us, because we have duplicate information, but I guess that's just how it is.

miina (approved):

PR #10209, opened 2022-01-14
Telemetry: Remove trackEvent call on mount

### LINKED ISSUE #7738: Telemetry: Excessive number of events fired
## Bug Description

Looks like usage of `useEffect()` for tracking events causes lots of tracking events being fired

## Expected Behaviour

Tracking events should only be fired upon user interaction.

## Steps to Reproduce

1. Open editor
2. See a bunch of events being fired

## Screenshots

Just on example:

Screenshot 2021-05-26 at 11 21 05

## Additional Context

- Plugin Version: 1.7.2
- WordPress Version: 5.7
- Operating System: macOS
- Browser: Chrome

---

BrookeGraham:
## Context

In some instances we were calling `trackEvent` in an `useEffect`. Most of these cases included a check as to whether the event should be track due to some triggering event. However, in a couple places there wasn't a preventative check which was leading to the `trackEvent` getting called on mount, or during each re-render. We only want to call trackEvent when those events are truly triggered. 

## Summary

Add a check before calling `trackEvent` when trackEvent is called in a `useEffect` to avoid calling them when the component mounts. 

## Relevant Technical Choices

Check for searchTerm length before calling trackEvent. 

## Testing Instructions

1. You can turn on tracking via the tracking package shared config file.  (There might be another way to change this setting, but this worked for me)
2. Go to package -> tracking -> src -> shared 
3. change `trackingAllowed: true, trackingEnabled: true`
4. Tracking changed in 3 place. `TextSets` `Media3p` and the `Dashboard`
5. Go to text set tab and open console. 
6. In the console type `window.webStoriesTrackingDataLayer` and hit enter. 
7. You will see a list of all the tracking events
8. Filter text set by either `Match fonts from story` or selecting one of the pills ie. `COVER` , `STEPS` etc.
9.  In the console type `window.webStoriesTrackingDataLayer` and hit enter again. You should see the new `search` event in `window.webStoriesTrackingDataLayer` TWICE. Each one is sent to two different places. Not sure why. 
10. Repeat for Media3p search and Dashboard search

no

Yes!

no

## Checklist

---

Fixes #7738

### DIFF
--- a/packages/story-editor/src/components/library/panes/text/textSets/textSetsPane.js
+++ b/packages/story-editor/src/components/library/panes/text/textSets/textSetsPane.js
@@ -22,6 +22,7 @@ import {
   useMemo,
   useCallback,
   useEffect,
+  useRef,
 } from '@web-stories-wp/react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
@@ -89,6 +90,7 @@ function TextSetsPane({ paneRef }) {
     })
   );
   const [showInUse, setShowInUse] = useState(false);
+  const trackChange = useRef(false);
 
   const allTextSets = useMemo(() => Object.values(textSets).flat(), [textSets]);
   const storyPages = useStory(({ state: { pages } }) => pages);
@@ -172,22 +174,25 @@ function TextSetsPane({ paneRef }) {
       localStore.setItemByKey(`${LOCAL_STORAGE_PREFIX.TEXT_SET_SETTINGS}`, {
         selectedCategory,
       });
+      trackChange.current = true;
     },
     [speak]
   );
 
-  const onChangeShowInUse = useCallback(
-    () => requestAnimationFrame(() => setShowInUse((prevVal) => !prevVal)),
-    [setShowInUse]
-  );
+  const onChangeShowInUse = useCallback(() => {
+    requestAnimationFrame(() => setShowInUse((prevVal) => !prevVal));
+    trackChange.current = true;
+  }, [setShowInUse]);
 
   useEffect(() => {
-    trackEvent('search', {
-      search_type: 'textsets',
-      search_term: '',
-      search_category: selectedCat,
-      search_filter: showInUse ? 'show_in_use' : undefined,
-    });
+    if (selectedCat || showInUse || trackChange.current) {
+      trackEvent('search', {
+        search_type: 'textsets',
+        search_term: '',
+        search_category: selectedCat,
+        search_filter: showInUse ? 'show_in_use' : undefined,
+      });
+    }
   }, [selectedCat, showInUse]);
 
   const sectionId = useMemo(() => `section-${uuidv4()}`, []);
--- a/packages/story-editor/src/components/library/panes/text/textSets/textSetsPane.js
+++ b/packages/story-editor/src/components/library/panes/text/textSets/textSetsPane.js
@@ -185,7 +185,7 @@ function TextSetsPane({ paneRef }) {
   }, [setShowInUse]);
 
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


packages/story-editor/src/components/library/panes/text/textSets/textSetsPane.js
```diff
@@ -172,22 +174,25 @@ function TextSetsPane({ paneRef }) {
...
-      search_filter: showInUse ? 'show_in_use' : undefined,
-    });
+    if (selectedCat || showInUse || trackChange.current) {
+      trackEvent('search', {
+        search_type: 'textsets',
+        search_term: '',
+        search_category: selectedCat,
+        search_filter: showInUse ? 'show_in_use' : undefined,
+      });
+    }
```
BrookeGraham: @swissspidy The way I have made this work is that when you open the `textSet` tab, if you had previously changed the textSet filtering this event will still fire (we are storing those filtering setting in localStorage). Is that the desired behavior? Or do we only want to track when the user actively changes the filtering setting?

Also, not sure if we want to rename this event sense we are no longer using the search input on textSet, only filtering with text that match fonts from story and via categories. ie. `cover` `steps` etc.

swissspidy: > if you had previously changed the textSet filtering this event will still fire (we are storing those filtering setting in localStorage). Is that the desired behavior? Or do we only want to track when the user actively changes the filtering setting?

Ideally only on user interaction, but if that's too big of a change it should be fine to keep I think.

> Also, not sure if we want to rename this event sense we are no longer using the search input on textSet

We're good; `search` has a special meaning in analytics that we wanna keep.

maxyinger (approved):
I guess I'm not too familiar with how we want the `trackEvent` data to look after its been reported. A lot of these changes from the react side make me think we should just be putting these calls explicitly in actions. Just kind of feels like we are inferring actions within effects vs just putting stuff directly in an action to track an action.

Anyway, these changes definitely fix the reported issue. Thanks for the fix!

swissspidy:
> A lot of these changes from the react side make me think we should just be putting these calls explicitly in actions. Just kind of feels like we are inferring actions within effects vs just putting stuff directly in an action to track an action.

Yeah that would be better, though it might cause repetition if an action can be triggered in multiple ways in the UI. Maybe something for a follow-up ticket.

samwhale (approved):
Agree with max - seems better if we place this in an `onChange` event handler rather than in an effect, but happy to add that to a later ticket.

I don't mind some added repetition if that were to be the case.

BrittanyIRL (approved):
Appreciate the detailed testing directions!

packages/story-editor/src/components/library/panes/text/textSets/textSetsPane.js
```diff
@@ -172,22 +174,25 @@ function TextSetsPane({ paneRef }) {
...
     [speak]
   );
 
-  const onChangeShowInUse = useCallback(
-    () => requestAnimationFrame(() => setShowInUse((prevVal) => !prevVal)),
-    [setShowInUse]
-  );
+  const onChangeShowInUse = useCallback(() => {
+    requestAnimationFrame(() => setShowInUse((prevVal) => !prevVal));
+    trackChange.current = true;
```
swissspidy, line 184: When is this ever set to `false` again?

BrookeGraham: Currently, only when the `textSet` pane it gets remounted. So anytime the pane closes and then gets reopened.  Which means tracking will be sent every time the user changes the filtering via the pills or the toggle.

PR #10217, opened 2022-01-17
Allow completely disabling/hiding default templates

### LINKED ISSUE #10124: Allow completely disabling/hiding default templates
## Feature Description
A user has requested a way to customize the default view for the Templates dashboard. Currently, the Templates dashboard loads all templates, but this feature request would allow a customized / filtered view of the templates on load.

## Additional Context

related support topic: https://wordpress.org/support/topic/remove-default-story-and-page-templates/

-----

## Acceptance Criteria

* Pass`'canViewDefaultTemplates' => true` to app in `\Google\Web_Stories\Admin\Editor::get_editor_settings()` and `\Google\Web_Stories\Admin\Dashboard::get_dashboard_settings()`
* In dashboard, hide "Explore Templates" from menu if user does not have capability
* In dashboard, prevent direct access to templates if user does not have capability (feasible?)
* In editor, hide "Default Templates" from the page templates tab, hide the selection dropdown, and always show the custom templates.

This will allow devs to disable access to default templates via the `web_stories_editor_settings` and `web_stories_dashboard_settings` filters.

**Alternative:**

Use something like `current_user_can( 'web_stories_view_default_templates' )` instead, which allows devs to filter this via `map_meta_cap`.

It's a bit more verbose which helps a lot with comprehension.

We could even do the same for the `showMedia3p` config.

spacedmonkey:
## Summary

Give developers the ability to disable default templates, by adding a field to the editor / dashboard settings object. 

## User-facing changes

|Before |After|
--- | --- | 
|Screenshot 2022-01-17 at 13 30 26|Screenshot 2022-01-17 at 13 30 05|
|Screenshot 2022-01-17 at 13 27 35|Screenshot 2022-01-17 at 15 36 26|

## Testing Instructions

1.

## Checklist

---

Fixes #10124


packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
```diff
@@ -178,12 +185,14 @@ function PageTemplatesPane(props) {
     return () => clearTimeout(timeout);
   }, [highlightedTemplate]);
 
-  const options = [
-    {
+  const options = [];
```
swissspidy, line 193: Let's hide `DropDownWrapper` below if `options.length <= 1`

The dropdown is not useful when there is only one option or no option at all.

spacedmonkey: So just don't show the dropdown at all? Or show a title?

spacedmonkey: Screenshot 2022-01-17 at 15 12 09
Spacing on this looks off.

swissspidy: I see.

Let's do it this way then:

1. Only hide the `<Select>`, but not the `DropDownWrapper`
1. Remove height from `DropDownWrapper`. It serves no purpose anyway. Can actually also remove the `text-align`

So essentially:

```js
const DropDownWrapper = styled.div`
  margin: 28px 16px 17px;
`;

{/* Still rendering DropDownWrapper to keep spacing nice and tidy */}
<DropDownWrapper>
  {options.length > 1 && (
    <Select
      options={options}
      selectedValue={showDefaultTemplates ? DEFAULT : SAVED}
      onMenuItemClick={handleToggle}
      aria-label={__('Select templates type', 'web-stories')}
    />
  )}
</DropDownWrapper>
```

With the result looking like this:

Screenshot 2022-01-17 at 23 18 14

That'll do for now

I'll ping UX if needed in the future.

packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
```diff
@@ -157,9 +163,10 @@ function PageTemplatesPane(props) {
     setNextTemplatesToFetch,
   ]);
 
-  const handleToggle = () => {
-    setShowDefaultTemplates(!showDefaultTemplates);
-    localStore.setItemByKey(LOCAL_STORAGE_KEY, !showDefaultTemplates);
+  const handleToggle = (event, menuItem) => {
+    const value = menuItem === 'default';
```
miina:
```suggestion
    const value = menuItem === DEFAULT;
```

miina: Since it's not really a toggle, might be good to rename it just for readability, e.g. `handleSelect`. WDYT?

packages/dashboard/src/components/pageStructure/leftRail.js
```diff
@@ -193,6 +198,13 @@ function LeftRail() {
                       numNewTemplates
                     )
                   : label;
+
+              const isTemplatesEnabled =
```
miina: Should this be renamed to `isTemplatesDisabled`? Also, could this be moved to the line right after `{PRIMARY_PATHS.map(({ Icon, ...path }) => {` to return `null` as soon as possible?

miina (commented):
Would be nice to have some docs about this as well (the option to disable the default templates).

swissspidy (approved):

packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
spacedmonkey: Fixed.

BrittanyIRL (approved):
Echo Miina's comment about some documentation being nice, otherwise looks good!

swissspidy:
I think we can put some notes in the `docs/` folder in a follow-up PR 👍 

Would also be a good opportunity to mention the similar flag to disable the media3p library.

swissspidy:
The `should not render default templates` e2e test now needs updating it seems

PR #10219, opened 2022-01-17
Embeds: Use native `<img>` instead of `<amp-img>`

### LINKED ISSUE #10213: Renderer: Use native `<img>` instead of `<amp-img`
## Feature Description

With `amp-img` being deprecated in favor of native `img` (https://github.com/ampproject/amphtml/issues/30442), we can now finally stop using `amp-img` by default.

The AMP plugin v2.3 (coming out soon) will make the switch as well. See https://github.com/ampproject/amp-wp/pull/6805

We're currently using `amp-img` here:

https://github.com/google/web-stories-wp/blob/e7cb5fe4dee961dbfc318dbc7b3c75b3373b6a02/includes/Renderer/Stories/Renderer.php#L594-L613

We should:

1. Remove that condition and use `img` only
2. Use `loading="lazy"` and `decoding="async"` for `img`
2. update any CSS targeting `amp-img` 
3. Verify that there are no regressions when viewing the AMP version  of a post

## Additional Context

**Note:** This does not apply to stories, i.e. `<amp-story>` remains unaffected.

RahiDroid:
## Context
- With the introduction of sandboxing levels (check https://github.com/ampproject/amp-wp/pull/6546) in [AMP 2.2.0 release](https://github.com/ampproject/amp-wp/releases/tag/2.2.0), it was noticed that the Web Stories' poster images weren't rendering properly.

## Summary
- The PR removes the `<amp-img/>` element from the markup so the native `` HTML element will be used on both AMP & Non-AMP pages.

## Relevant Technical Choices
- The issue of Poster images not rendering properly happened in the first place because the `amp-img` tag had some missing styles which were only given for the `img` tag (non-AMP page), and since there's no CSS processing happening on the `Loose` sandboxing level, the AMP plugin wasn't adding in the styles for `amp-img`.
- adding the missing styles could've also fixed the issue but since `amp-img` is going to be deprecated soon, it makes sense to remove the markup entirely.

## Testing Instructions
- Install latest (2.2.0) [AMP Plugin](https://wordpress.org/plugins/amp/) from the Plugins directory.
- Install latest (1.15.1) [Web Stories plugin](https://wordpress.org/plugins/web-stories/) from the Plugins directory.
- Enable the `Loose` experimental sandboxing level. You'll have to have the AMP running on `Standard` mode for this option to be visible.
Screenshot 2022-01-17 at 7 42 12 PM
- Add the `Web Stories` block on a post or a page. By default, if there's no AMP-invalid markup on a page, the AMP plugin will switch to higher/stricter sandboxing levels automatically, if that's the case, you can add a `Custom HTML` core block with some invalid AMP content (`<script></script>` will do the job).
- Visit the page/post on front-end and check the Poster images of the Web Stories block aren't being rendered properly.
- Make sure on the admin-bar, you see the sandboxing level is set to `Loose/1`.
Screenshot 2022-01-17 at 7 46 38 PM

1. Checkout the branch and you can follow instructions mentioned above.

No

No

## Checklist

---

## Screenshots
- Before
Screenshot 2022-01-14 at 11 21 26 PM

- After
Screenshot 2022-01-14 at 11 23 53 PM

Fixes #10213

### DIFF
--- a/includes/Renderer/Stories/Renderer.php
+++ b/includes/Renderer/Stories/Renderer.php
@@ -591,26 +591,14 @@ abstract class Renderer implements RenderingInterface, Iterator {
 			?>
 			<div class="web-stories-list__story-poster">
 				<a href="<?php echo esc_url( $story->get_url() ); ?>">
-					<?php
-					if ( $this->context->is_amp() ) {
-						// Set the dimensions to '0' so that we can handle image ratio/size by CSS per view type.
-						?>
-						<amp-img
-							src="<?php echo esc_url( $poster_url ); ?>"
-							layout="responsive"
-							width="0"
-							height="0"
-							alt="<?php echo esc_attr( $story->get_title() ); ?>"
-						>
-						</amp-img>
-	<?php } else { ?>
-						<img
-							src="<?php echo esc_url( $poster_url ); ?>"
-							alt="<?php echo esc_attr( $story->get_title() ); ?>"
-							width="<?php echo absint( $this->width ); ?>"
-							height="<?php echo absint( $this->height ); ?>"
-						>
-					<?php } ?>
+					<img
+						src="<?php echo esc_url( $poster_url ); ?>"
+						alt="<?php echo esc_attr( $story->get_title() ); ?>"
+						width="<?php echo absint( $this->width ); ?>"
+						height="<?php echo absint( $this->height ); ?>"
+						loading="lazy"
+						decoding="async"
+					>
 				</a>
 			</div>
 			<?php
--- a/packages/stories-block/src/css/common.css
+++ b/packages/stories-block/src/css/common.css
@@ -83,11 +83,6 @@
   box-sizing: border-box;
 }
 
-.web-stories-list__story-poster amp-img img {
-  object-fit: cover;
-  box-sizing: border-box;
-}
-
 .web-stories-list__story-poster::before {
   content: '';
   display: block;
--- a/packages/stories-block/src/css/views/circles.css
+++ b/packages/stories-block/src/css/views/circles.css
@@ -30,11 +30,7 @@
   .web-stories-list__story-poster
   .web-stories-list__story-poster-placeholder,
 .web-stories-list.is-view-type-circles .web-stories-list__story-poster > img,
-.web-stories-list.is-view-type-circles .web-stories-list__story-poster a > img,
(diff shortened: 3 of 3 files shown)


includes/Renderer/Stories/Renderer.php
```diff
@@ -591,26 +591,14 @@ protected function render_story_with_poster() {
...
-							height="<?php echo absint( $this->height ); ?>"
-						>
-					<?php } ?>
+					<img
+						src="<?php echo esc_url( $poster_url ); ?>"
+						alt="<?php echo esc_attr( $story->get_title() ); ?>"
+						width="<?php echo absint( $this->width ); ?>"
+						height="<?php echo absint( $this->height ); ?>"
+						loading="lazy"
+						decoding="async"
```
spacedmonkey, line 600: These changes to reasonable to me, but just want to confirm adding these attributes is the right thing. CC @swissspidy

spacedmonkey:
Should the following be removed as well. 
https://github.com/google/web-stories-wp/blob/663c75726e4baf410b9b4f52f737ba26abbea3b9/includes/KSES.php#L578-L596 

and references in [Sanitization_Utils.php](https://github.com/google/web-stories-wp/blob/663c75726e4baf410b9b4f52f737ba26abbea3b9/includes/AMP/Traits/Sanitization_Utils.php).

includes/Renderer/Stories/Renderer.php
swissspidy: Yes. Without these it would be invalid AMP. I explicitly mentioned these in #10213 as well.

swissspidy:
Thanks for bringing this up.


We still need the `amp-img` references in KSES. That should be left untouched.

They are for single stories, not embeds.

Single stories remain unaffected by this, as they still require `amp-img`

includes/Renderer/Stories/Renderer.php
spacedmonkey: Just confirming, no need to thumbs down 😢  . Should, implies some wiggle room, if it needs it, maybe you that wording.

swissspidy: I swear this was meant to be a thumbs up

spacedmonkey (approved):

swissspidy (approved):
Works like a charm!

PR #10240, opened 2022-01-18
History: reduce polluting history

### LINKED ISSUE #10009: History: Prevent polluting history
## Feature Description
Currently, history is polluted for example by props that are added async in the background. Such props are for example:
```
  'resource.baseColor',
  'resource.isMuted',
  'resource.poster',
  'resource.posterId',
```

There might be more. In #9822, logic is put in place to allow excluding properties from creating separate history entries.
Once we have both that and #9577 merged, we should have everything in place to not pollute history anymore.

miina:
## Summary
Ignores various media properties that change in the background, polluting history.

Note: There are still things that pollute history but it takes quite a bit of debugging to figure those out so we should probably create follow-up investigation issues per case.

## To-do

- [x] More testing
- ~Try to track down values still causing extra entries in the history~ This takes quite a bit of time, it might be good to create separate issues for separate cases for further investigation, e.g. "Dragging video onto the page from local files".

## Testing Instructions

This PR is not very straightforward to test but can be tested comparing `main` and the results on this PR in the following way:
1. Reload the page on `main`
2. Do an action (wait until it has fully finished if there's processing happening)
3. Click "Undo" until getting back to the initial state
4. Log how many times you had to click "Undo"
5. Compare with results of the same action for this PR, verify that it takes less steps on this PR than on `main`. Would be great to log the differences in GitHub as well.

The actions to test:
- Removing audio from a video
- Dragging a video from local directly onto the Page
- Trimming a video

No

No

No

## Checklist

---

Fixes #10009


packages/story-editor/src/app/story/effects/useHistoryEntry.js
```diff
@@ -27,8 +27,13 @@ import { useHistory } from '../../history';
 import deleteNestedKeys from '../utils/deleteNestedKeys';
 
 // Changes to these properties of elements do not create a new history entry.
-// Props to be added in
-const ELEMENT_PROPS_TO_IGNORE = [];
+const ELEMENT_PROPS_TO_IGNORE = [
+  'resource.baseColor',
```
spacedmonkey, line 33:
```suggestion
  'resource.baseColor',
  'resource.blurHash',
```

Can we add blurhash as well?

swissspidy: Isn't that only added for media library items? So I don't think it's actually relevant here.

spacedmonkey: It seems like they are added to story data. 
Screenshot 2022-01-19 at 11 40 49

miina: > It seems like they are added to story data.

Since it's never needed inside a story (if I understand correctly?), should we remove it from adding to the story data instead? Or alternatively, should the blurhash be used instead of the `baseColor`?

swissspidy: > It seems like they are added to story data.

Yes, but the blurhash generation is done _before_ an image is inserted to the page.

I don't think this scenario ever happens:

1. Add image to the page
2. Blurhash is added to the image
3. History is polluted

> Since it's never needed inside a story (if I understand correctly?), should we remove it from adding to the story data instead?

It doesn't hurt if it stays there, no?

miina: > It doesn't hurt if it stays there, no?

No, not really. Was just thinking in case it pollutes the history. If it doesn't then it's ok.

spacedmonkey: No, that isn't the case. If you drag an image into canvas, you generate a local resource while while it is uploading. That is then replaces with real src etc. Then once loaded, blurhash and base color are backfilled. Meaning the resource woud be updated, polluting history.

spacedmonkey: Does it hurt anything to ignore it?

swissspidy: I thought the backfilling only updates resources in the library, _not_ the page. But I can now see that it does both:

https://github.com/google/web-stories-wp/blob/37cd92a734635bd1cbb8b878f2b3df7db35d4dd9/packages/story-editor/src/app/media/utils/useDetectBlurhash.js#L58-L62

So, alright, let's add it to the list.

miina: > Does it hurt anything to ignore it?

Nope, I can definitely add it here if it pollutes history.

spacedmonkey (approved):

swissspidy:
> This takes quite a bit of time, it might be good to create separate issues for separate cases for further investigation, e.g. "Dragging video onto the page from local files".

Could we dump the history state in-between to see the difference in the various history entries being added?

miina:
> Could we dump the history state in-between to see the difference in the various history entries being added?

Yes but it doesn't necessarily always show what caused the changed -- e.g. if the object changes but values are visibly the same. Still a good place to start though.
Also, it probably needs to be determined case by case how to fix those issues -- might not be just a case of ignoring the value in history but combining changes in code for example.

I can create a follow-up issue to investigate some of those cases. Not sure if that's something for Infra (in case it's related to all-things-media-processing implemented by Infra) or Prometheus. WDYT?

swissspidy:
A follow-up ticket would be great to track this. Maybe start with assigning Prometheus and if it turns out to be a media processing issue we can re-assign if needed.

miina:
Added: https://github.com/google/web-stories-wp/issues/10289

merapi (changes requested):
We need some additional guards in `deleteNestedKeys`, it's crashing the editor at the moment (add shape or text):

image

(guard against that `undefined`)
image

packages/story-editor/src/app/story/effects/useHistoryEntry.js
```diff
@@ -27,8 +27,14 @@ import { useHistory } from '../../history';
 import deleteNestedKeys from '../utils/deleteNestedKeys';
 
 // Changes to these properties of elements do not create a new history entry.
```
swissspidy: Since you mentioned this only happens if _only_ this property changes and nothing else, it would be good to reflect that here in the docs

miina: Updated the comment.

swissspidy (approved):

PR #10241, opened 2022-01-18
Context Menu: New elements should be selected after duplication

### LINKED ISSUE #10179: Right Click Actions: Duplicate Selection
## Feature Description

Pointed out by @barklund and @agingoldseco during bug bash.

> When duplicating an element (or multiple elements), I would expect the new element(s) to be selected rather than the old one(s). That’s what happens if you do it by keyboard (cmd+d), so the two ways should at least be identical.

samwhale:
## Context

Issue found during Bug Bash 1.16.0.

## Summary

Duplicating one or more elements will select the newly created element(s).

## User-facing changes

|Multiple|Single|From Layer Panel|
|--|--|--|
|duplicateelements|layerduplicate|duplicateelement|

## Testing Instructions

You will need to test that duplicating elements still works as expected:
- duplicates all animations applied to element
- places newly duplicated element in a different spot

New functionality:
- the newly created element will be selected after duplication.
- clicking the duplication button in the layer panel should select the new element on creation.

1. Add element(s) to canvas. Add styles + animations to your heart's content
2. Select element(s)
3. Duplicate
4. Verify that new elements are selected and the old elements are not
5. Verify that elements were duplicated correctly

no

no

no

## Checklist

---

Fixes #10179


BrittanyIRL (approved):
🤝  🎸 🚀  

Works as described

packages/story-editor/src/app/story/useStoryReducer/actions.js
```diff
@@ -119,12 +119,12 @@ const updateElementById =
       payload: { elementIds: [elementId], properties },
     });
 
-const duplicateElementById =
+const duplicateElementsById =
   (dispatch) =>
-  ({ elementId }) =>
+  ({ elementIds }) =>
     dispatch({
       type: types.DUPLICATE_ELEMENT_BY_ID,
```
miina: Let's change this as well then everywhere for consistency:

```suggestion
      type: types.DUPLICATE_ELEMENTS_BY_ID,
```

packages/story-editor/src/app/story/useStoryReducer/reducers/duplicateElementsById.js
```diff
@@ -0,0 +1,91 @@
...
+    newPage.elements = [
+      ...newPage.elements.slice(0, elementIndex),
+      elementToDuplicate,
+      element,
+      ...newPage.elements.slice(elementIndex + 1),
+    ];
+
+    return;
+  });
+
```
miina, line 77: Could we here just `return state` if `!newSelection.length`? Would avoid creating a new history entry if there are no  elements duplicated for some reason.

packages/story-editor/src/app/story/useStoryReducer/reducers/setSelectedElements.js
```diff
@@ -26,7 +26,7 @@ import { intersect } from './utils';
 /**
  * Set selected elements to the given list of ids.
  *
- * If given list is not a list, do nothing.
+ * If given `elementIds` are not a list, do nothing.
```
miina:
```suggestion
 * If given `elementIds` is not a list, do nothing.
```

packages/story-editor/src/app/story/useStoryReducer/test/duplicateElementByIds.js
```diff
@@ -169,15 +176,81 @@ describe('duplicateElementById', () => {
...
       (el) => !initialState.pages[0].elements.includes(el)
     );
     expect(result.pages[0].elements).toHaveLength(4);
-    expect(initalElement.x).not.toBe(newElement.x);
-    expect(initalElement.y).not.toBe(newElement.y);
+    expect(initialElement.x).not.toBe(newElement.x);
+    expect(initialElement.y).not.toBe(newElement.y);
+  });
+
+  it('duplicates more than one element but not the background', () => {
```
miina: Maybe let's add one with incorrect element IDs as well and verify the initial state doesn't change.

miina (commented):
Nice work! Left a few nits.

packages/story-editor/src/app/story/useStoryReducer/reducers/duplicateElementsById.js
samwhale: Ah great catch!

maxyinger (approved):
Code looks and works great!

Just want to make sure it's also intentional for the layers panel to select the new element when you click the duplicate icon on hover?

I also think one of the main reasons why this action was created in the layers panel vs what already exists in the codebase was to not have to select elements when they're duplicated. Now that it is doing that, we may be able to unify some implementations in the app. i.e.

https://github.com/google/web-stories-wp/blob/67d010eec4efe7e9797362f0610ecf55e24b8338/packages/story-editor/src/app/canvas/useCanvasKeys.js#L233-L262

samwhale:
>Just want to make sure it's also intentional for the layers panel to select the new element when you click the duplicate icon on hover?
>
>I also think one of the main reasons why this action was created in the layers panel vs what already exists in the codebase was to not have to select elements when they're duplicated. Now that it is doing that, we may be able to unify some implementations in the app. i.e.

@littlemilkstudio we talked about this earlier but posting here for visibility. I think the layer panel's duplication should match the duplication used elsewhere in the app. Now that the `duplicateElementsByIds` reducer action exists, all duplication should follow that same pattern.

This also makes it easier to maintain in the future since the functionality is all centralized. 2cd774e replaces the old functionality. ~but there's one more spot that the old function is being used:~

~https://github.com/google/web-stories-wp/blob/main/packages/story-editor/src/utils/copyPaste.js#L39-L87~

~I can add a ticket to visit that last spot and see if we can remove `duplciateElement.js` entirely.~

Edit: We don't need to do that.

maxyinger:
> @littlemilkstudio we talked about this earlier but posting here for visibility. I think the layer panel's duplication should match the duplication used elsewhere in the app. Now that the duplicateElementsByIds reducer action exists, all duplication should follow that same pattern.

This looks awesome. Thanks for adding that in!

packages/story-editor/src/app/story/useStoryReducer/test/duplicateElementByIds.js
```diff
@@ -0,0 +1,346 @@
+/*
```
miina: Nit: The filename should probably be `duplicateElementsById`.

packages/story-editor/src/app/story/useStoryReducer/reducers/duplicateElementsById.js
```diff
@@ -0,0 +1,96 @@
...
+    const { element, elementAnimations } = duplicateElement({
+      element: elementToDuplicate,
+      animations: oldPage.animations,
+      existingElements: oldPage.elements,
+    });
+
+    newSelection.push(element.id);
+
+    newPage.animations = [...(newPage.animations || []), ...elementAnimations];
+    newPage.elements = [
```
miina, line 68: One thing that happens like this is that the duplicated layer is added right next to the origin element, e.g. if I duplicate 3 elements:
- IMAGE 1
- IMAGE 2
- IMAGE 3

it becomes
- IMAGE 1
- IMAGE 1 dupe
- IMAGE 2
- IMAGE 2 dupe
- IMAGE 3
- IMAGE 3 dupe

instead of 
- IMAGE 1
- IMAGE 2
- IMAGE 3
- IMAGE 1 dupe
- IMAGE 2 dupe
- IMAGE 3 dupe

Is this intended? I was thinking that all the dupes would go to the end of the list of the layers as new elements, however, not really sure what's the standard behaviour in this case for layers. Thoughts?

samwhale: Good question. I don't have context as to why that happens, but this is what existed before this ticket:

https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/story-editor/src/app/story/useStoryReducer/reducers/duplicateElementById.js#L47-L52

@littlemilkstudio do you have more context here?

BrittanyIRL: This should be a question for UX not eng - @agingoldseco  do you have a quick second to weigh in on this?

agingoldseco: Can't say I have a good gauge of "standard" behavior here, but I think it make sense for copy/paste and duplicate to act differently. I think copy/paste the way it is now makes sense, where the new layers appear in the layers panel op top of everything else. And Duplicate also makes sense as you might want a duplicated element to appear on the same "level" as its origin element. But Morten's comment in bug bash is right—in my experience I want to interact with the duplicated elements first after duplication, so they should be selected rather than the origin elements. Just checked, and that's actually what Figma does too! 

Here's a video—see what happens in the left layers column when I copy the circles (first) and then duplicate them using option+drag:

https://user-images.githubusercontent.com/90645922/150835731-de1f427f-aea4-48b9-b2f4-b4eb9b128475.mov

samwhale: @agingoldseco this pr fixes that! More wondering about placement of the newly created elements.

Seems like in figma that they're all added to the end/beginning of the layers together rather than splitting them up and placing them directly before their original element.

samwhale: @agingoldseco this pr leaves the newly created elements selected 👍 

Seems like figma also places the newly created elements right next to their parent elements. For now I'll leave the behavior as is, but it will be an easy change if in the future we'd like that to act the same as copy/paste.

PR #10242, opened 2022-01-18
Video Trimmer: don't use `railBackgroundImage` if `null`

swissspidy:
## Context

I noticed that the browser tried to load a `null` resource, caused by `railBackgroundImage` initially being `null`

## Summary

Don't use `railBackgroundImage` as the background image if it's `null`.

## User-facing changes

N/A

## Testing Instructions

1.

## Checklist

---

Fixes #

### DIFF
--- a/packages/story-editor/src/components/videoTrim/videoTrimmer.js
+++ b/packages/story-editor/src/components/videoTrim/videoTrimmer.js
@@ -165,7 +165,11 @@ function VideoTrimmer() {
       <RailWrapper>
         <Rail
           width={railWidth}
-          style={{ backgroundImage: `url(${railBackgroundImage})` }}
+          style={{
+            backgroundImage: railBackgroundImage
+              ? `url(${railBackgroundImage})`
+              : undefined,
+          }}
         >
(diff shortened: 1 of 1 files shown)


packages/story-editor/src/components/videoTrim/videoTrimmer.js
```diff
@@ -165,7 +165,11 @@ function VideoTrimmer() {
       <RailWrapper>
         <Rail
           width={railWidth}
-          style={{ backgroundImage: `url(${railBackgroundImage})` }}
+          style={{
+            backgroundImage: railBackgroundImage
+              ? `url(${railBackgroundImage})`
+              : undefined,
```
miina, line 171: Shouldn't it be `none`? 🤔 Or if it's `undefined`, maybe it's not displayed at all in the DOM?

swissspidy: It won't be displayed at all.

Here's the DOM when it's missing:
Screenshot 2022-01-19 at 11 37 11

Here's when it arrives:
Screenshot 2022-01-19 at 11 37 47

miina (approved):

PR #10247, opened 2022-01-18
Media Library: don't use item index as `key` prop

swissspidy:
## Context

Miina observed some strange issues with poster images showing up in the media library after trimming a video.

I tracked this down to wrong usage of the `key` prop in the `imageRenderer` within the media library.

Trimming a video means a new item is added to the media list, which of course messes things up when using the array index as the key.

Using indexes for the `key` prop in React is bad practice, see https://reactjs.org/docs/lists-and-keys.html and https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318 for some explanations why.

We do have a lint rule for this, but it's not applied in this section because it's in a render prop and not in a loop.

## Summary

This PR updates the `key` prop to use the photo's ID instead.

The ID is either a temp ID for resources still being uploaded, or a numeric ID from the WP database.

## Testing Instructions

1. Trim a video in the editor
2. Scroll down the media library list
3. See that there are no stretched poster images suddenly appearing before/after a regular video in the library grid

No

No

No

## Checklist

---

Fixes #

### DIFF
--- a/packages/story-editor/src/components/library/panes/media/common/mediaGallery.js
+++ b/packages/story-editor/src/components/library/panes/media/common/mediaGallery.js
@@ -45,6 +45,7 @@ const PHOTO_MARGIN = 4;
  */
 function MediaGallery({ resources, onInsert, providerType, canEditMedia }) {
   const photos = resources.map((resource) => ({
+    id: resource.id,
     src: resource.src,
     width: resource.width,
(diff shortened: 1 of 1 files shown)


miina (approved):
Thanks for tracking this down!

PR #9995, opened 2021-12-14
Media Library: Fix video posters not showing in Safari

spacedmonkey:
## Summary

Solution found from this [article](https://muffinman.io/blog/hack-for-ios-safari-to-display-html-video-thumbnail/). 

### Before 
Screenshot 2021-12-14 at 18 14 35

### After
Screenshot 2021-12-14 at 18 14 08

## Testing Instructions

- [ ] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1. Open Safari browser
2. Open Stories editor
3. Navigate to the 3P Video panel, and select Video tab
4. Note the video poster images are blacked out and are not rendering
5. Hover over videos one by one and notice the images now display. Note the videos that haven't been hovered over do not display/render.

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [x] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [x] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #8449

PR #10023, opened 2021-12-17
Media: Fix video poster not rendering in Safari

spacedmonkey:
## Summary

Fix poster image render, by using an image tag to render poster instead of video tag. 

For performance reasons, it much better to use an image tag anyway. These are lazily loaded. 

On hover/focus, the image tag is replaced with a video tag and autoplayed. 

#### Before 

Screenshot 2021-12-17 at 14 20 33

#### After

Screenshot 2021-12-17 at 14 18 03

## Testing Instructions

- [ ] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1. Open editor in Safari, Edge, Firefox, Chrome. 
2. See poster images of videos in media library (including videos under third-party media).
3. Video should play when hovering over video

### Does this PR have a security-related impact?

No

### Does this PR change what data or activity we track or use?

No

### Does this PR have a legal-related impact?

No

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [x] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [x] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #8449

PR #10253, opened 2022-01-18
Fix Video Poster Not Rendering on Canvas in Safari

### LINKED ISSUE #10167: Canvas: Video poster not rendering in Safari
## Bug Description

Follow up https://github.com/google/web-stories-wp/issues/8449. Video poster are not being render of video elements on canvas. 

## Steps to Reproduce

1. Create a story. 
2. Insert video into canvas from media library. 
3. Save story. 
4. Reload story. 

## Screenshots

https://user-images.githubusercontent.com/237508/149138574-83b62b6c-b9b9-41d7-afaf-5214f49448f0.mov

## Additional Context

- Plugin Version: trunk
- WordPress Version: 5.9-RC2
- Operating System: Mac OS 12.1
- Browser: Safari 15.2

BrittanyIRL:
## Context

Video posters aren't getting rendered on canvas (have to hit play in order for it to render). 

| Safari | Chrome/Firefox | 
| --- | --- | 
| Screen Shot 2022-01-18 at 3 02 32 PM | Screen Shot 2022-01-18 at 3 02 49 PM |

## Summary

Now when you load a story with video in safari you get to see their poster images. 

## Relevant Technical Choices

The `video` element that's in charge of canvas videos needed to get its `preload` tag updated from `metadata`. Even though specs say Safari can use the preload tag, that browser treats 'metadata' differently from others? We're using `all` as the preload value in the gif `video` implementation so I opted to use that for the video tag here. It's the same as putting down `auto` or `true` or `akldfjlafdjlakdfjlaksf` because it's just the default value. 

## To-do

I remember seeing a ticket for this, I will track it down again, but sometimes the images rendered in the footer on safari will still have the pesky gray rectangle for video. It's a separate issue from this but I wanted to mention it. 

## User-facing changes

Videos will load poster now on canvas in safari. 

## Testing Instructions

Open the editor and add video (local, 3p), see that you don't get a gray rectangle. Now save and refresh and note that the video posters still load (no gray rectangle in canvas). 

1.

## Checklist

---

Fixes #10167

### DIFF
--- a/packages/media/src/types.js
+++ b/packages/media/src/types.js
@@ -68,7 +68,7 @@ ResourcePropTypes.videoResource = PropTypes.shape({
   width: PropTypes.number.isRequired,
   height: PropTypes.number.isRequired,
   poster: PropTypes.string,
-  posterId: PropTypes.number,
+  posterId: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
   tracks: PropTypes.arrayOf(ResourcePropTypes.trackResource),
   alt: PropTypes.string,
   title: PropTypes.string,
--- a/packages/story-editor/src/elements/video/display.js
+++ b/packages/story-editor/src/elements/video/display.js
@@ -112,7 +112,7 @@ function VideoDisplay({ previewMode, box: { width, height }, element }) {
           poster={poster || resource.poster}
           style={style}
           {...videoProps}
-          preload="metadata"
+          preload="all"
           loop={loop}
           muted={muted}
           ref={ref}


packages/media/src/types.js
```diff
@@ -68,7 +68,7 @@ ResourcePropTypes.videoResource = PropTypes.shape({
   width: PropTypes.number.isRequired,
   height: PropTypes.number.isRequired,
   poster: PropTypes.string,
-  posterId: PropTypes.number,
+  posterId: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
```
BrittanyIRL, line 71: Getting rid of this noise

Screen Shot 2022-01-18 at 2 56 56 PM

packages/story-editor/src/elements/video/display.js
```diff
@@ -112,7 +112,7 @@ function VideoDisplay({ previewMode, box: { width, height }, element }) {
           poster={poster || resource.poster}
           style={style}
           {...videoProps}
-          preload="metadata"
+          preload="all"
```
BrittanyIRL, line 115: preload all is also used in elements > gif > display, and the equivalent (true) is ued in components > helpCenter > quickTip > index. 

it might be worth consolidating these for the future to avoid thinking they do something different.

maxyinger (approved):
I'm not familiar with the context around why `preload="metadata"` was set, so not sure what we're trading off here, but your explanation as to why this makes it work in safari looks good to me.

swissspidy:
Are there perhaps any performance concerns with `preload="all"`, e.g. when editing a story with many videos?

For the media library we opted to display the poster images instead and only load the video on hover (to autoplay). See #10023

BrittanyIRL:
> Are there perhaps any performance concerns with `preload="all"`, e.g. when editing a story with many videos?

There are minor ones, yes. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#attr-preload - it felt like an ok trade off though for the content showing up and being playable. I can do a check on a large story with many videos in the AM to give a concrete answer - but you're only looking at 1 page at a time and a page shouldn't really frequently have more than 1 video playing on it, so feels like a non-issue. 

> For the media library we opted to display the poster images instead and only load the video on hover (to autoplay). See #10023

I looked into this at first, admittedly I'm new to this portion of the code base so maybe I'm missing something that is obvious to others about the canvas media implementation but it wasn't very clear. I didn't want to add any additional state to track loading an img vs the video, especially since the issue is the browser handling the preload tag - it just ended up feeling kind of gross and convoluted in the canvas.

spacedmonkey:
I am not seeing the poster image render in safari. 

https://user-images.githubusercontent.com/237508/150125910-69176efa-04dc-4108-91ce-1762fb559ea9.mov

I am seeing the first frame of the video, which is not the poster of the video. Poster image is the frame from the first second of the video or a custom one the users have uploaded. 

Some extra context - https://muffinman.io/blog/hack-for-ios-safari-to-display-html-video-thumbnail/

swissspidy:
> I am not seeing the poster image render in safari.

That is true. While it is not displaying the poster image, at least the first video frame is shown instead of nothing—which is the bug here.

If there are no performance concerns then I'd say that is fine.

> but you're only looking at 1 page at a time and a page shouldn't really frequently have more than 1 video playing on it, so feels like a non-issue.

@BrittanyIRL I think my main worry there was that a lot of videos will be downloaded if you navigate through the pages but never actually play the videos.

>  I didn't want to add any additional state to track loading an img vs the video, especially since the issue is the browser handling the preload tag - it just ended up feeling kind of gross and convoluted in the canvas.

Fair, it definitely adds complexity.

-----

Aside: while the video or poster is being loaded, should we display the baseColor/blurHash during that time? To have a nicer loading experience. Of course something for its own PR.

spacedmonkey:
>While it is not displaying the poster image, at least the first video frame is shown instead of nothing—which is the bug here

The name of the ticket is Video poster not rendering in Safari. Poster being the key word here. Yes, this behaviour is better than the one before, but doesn't fix the bug, which is no poster. By not render the poster image, that makes the look of a poster element different in different browser, which is a little strange to me.

swissspidy:
I would argue the actual bug is that the user sees this:

Screenshot 2022-01-19 at 13 37 34

spacedmonkey:
> Aside: while the video or poster is being loaded, should we display the baseColor/blurHash during that time? To have a nicer loading experience. Of course something for its own PR.

I like this idea. Created a ticket. https://github.com/google/web-stories-wp/issues/10258

spacedmonkey:
Using this [workaround](https://muffinman.io/blog/hack-for-ios-safari-to-display-html-video-thumbnail/), I created this [branch](https://github.com/google/web-stories-wp/tree/fix/seeking-video). @BrittanyIRL can you take a look and see if you like the idea. 

Also in my branch, I also fix poster images in gif elemetns.

swissspidy:
That workaround is specifically for Safari on iOS, so I don't think that necessarily applies here?

What's the difference of using that hack vs. `preload="all"`? Won't the browser preload the video either way?

BrittanyIRL:
Well I definitely feel like I understand the underlaying ask here a little better. Weird bug to be fixing since pea pod hasn't touched much media in the past year 😂  

So, Jonny's idea does actually fix the issue in Safari - not just IOS. And it's also more performant. It was tried here also and then closed: https://github.com/google/web-stories-wp/pull/9995 


Turns out, Pascal, you had the right question about performance with this idea - it's a marginal change in chrome however Safari gets overwhelmed really quickly when we're asking it to preload everything, so I'm going to close this PR because it's not right. That said, Jonny's branch works great. Here's some timelines tracking the requests in both safari and chrome. My branch actually crashed the safari timeline capture (not good!) so it's not here - got tired of waiting for it to finish loading the timeline. 

| Safari - main | Safari - jonny's hack |
| --- | --- |
| control | jonnysfix | 

| Chrome - main | Chrome - both hacks |
| --- | --- |
| Screen Shot 2022-01-19 at 9 52 55 AM | Screen Shot 2022-01-19 at 9 50 03 AM |


I'm still hesitant to add anything to the canvas media element that resembles this fix here https://github.com/google/web-stories-wp/pull/10023/files#diff-cb7f8e6e41df9942887dcb7b105a9ae5012cf6fb3e6699bb0e2dae66cb95c670 which is what it sounds like is getting mentioned in the above discussion. The canvas's media element isn't set up the same way and the 'hover' + play video isn't in the same area as this video tag that's causing the gray box (which I think is the real bug to solve since it's not helpful in the canvas).  I think the solution is whatever gets video elements in safari to load on the canvas with some sort of visual that relates to their media (aka not a gray box unless that's what they uploaded as a video for some reason) and ALSO doesn't complicate the structure and bog down performance for other browsers is what we want to do here. So, maybe that's Jonny's fix - especially if we can just keep an eye on when their preload tag for video gets updates (wishful thinking?). 

What do you guys think?

PR #10256, opened 2022-01-19
Decoupling : Integration layer testing

ayushnirwal:
## Context

This PR adds fallback tests to check if fallbacks implemented in [PR](https://github.com/google/web-stories-wp/pull/9949) works properly.

## Summary

List of API callbacks and their fallbacks which are tested with tests implemented in this PR.
| API Callback             | Fallback description                                                                                                                                                                |
| ------------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `getCurrentUser`         | No UI fallback, but should not break                                                                                                                                                |
| `updateCurrentUser`      | No UI fallback, but should not break                                                                                                                                                |
| `getProxyUrl`            | No UI fallback, but should not break                                                                                                                                                |
| `getPublisherLogos`      | No UI fallback, but should not break                                                                                                                                                |
| `addPublisherLogo`       | No UI fallback, but should not break                                                                                                                                                |
| `getStoryById`           | This is `initialEdits`'s fallback (if `initialEdits` is undefined then pull initialState using `getStoryById`). If both are not provided then the editor should break                                            |
| `getMedia`               | Editor should not render 1p media panel           |
| `getCustomPageTemplates` | Editor should not render saved page templates option in page templates pane's drop-down                                                                                                                                         |                                                                                                                                               

## Testing Instructions

1.

## Checklist

---

Fixes #


packages/story-editor/src/karma/integrationLayerTesting/optionalCallbacks.karma.js
```diff
@@ -0,0 +1,122 @@
...
+  });
+
+  it('if initialEdits is NOT passed, getStoryById should be called to load a story', async () => {
+    const getStoryById = jasmine
+      .createSpy('getStoryById')
+      .and.callFake((storyId) => {
+        return asyncResponse({
+          storyId,
+          ...storyResponse,
+        });
```
swissspidy: Nit: if we just do `Promise.resolve({
          storyId,
          ...storyResponse,
        })` here we don't need to export `asyncResponse`

swissspidy (approved):

Issue #10158, opened 2022-01-11
Performance Report: Jan 2022 baseline perf report


BrittanyIRL:
## Task Description

As we move to focus on performance as a feature we should have a proper baseline to compare and seek improvements from.

maxyinger:
Testing three base interactions on `main` prior to any of the recent performance improvements to establish a baseline (commit hash: `1d5c7228e`):
- Select an element
- Drag a selected element to update it's position
- Change the layer order of a selected element

These tests are being performed on a fully specced out 2018 macbook pro on this sample story:
https://stories-new-wordpress-amp.pantheonsite.io/wp-admin/post.php?post=20239&action=edit#page=%2522593109cf-a227-4f95-af6b-f413e823064a%2522

The story should look like this and have 1 page that contains 50 shape elements (51 elements total including the background):
<img width="725" alt="Screen Shot 2022-01-13 at 12 17 15 PM" src="https://user-images.githubusercontent.com/35983235/149386851-7314f2e2-78f4-4a3e-b4d9-950a2a8bb497.png">

All tests are being performed with the layers panel expanded, the design panel open, and the library pane set to the media tab

maxyinger:
**Select an Element**
| React Specific | Total JS |
| ------------ | ----------|
| Largest Commit: ~150ms | Rough total interaction time: ~350ms |
| <img width="825" alt="Screen Shot 2022-01-13 at 12 27 17 PM" src="https://user-images.githubusercontent.com/35983235/149388080-be576bd9-3eb3-464e-b901-2511457768ad.png"> |  <img width="854" alt="Screen Shot 2022-01-13 at 12 30 50 PM" src="https://user-images.githubusercontent.com/35983235/149388553-b17cc72c-6b1b-47d7-9701-d34a085bc6b5.png"> |

maxyinger:
**Drag a selected element to update it's position** 

**Initiating the drag**
| React Specific | Total JS |
| ------ | ------ |
| Relevant Commit: ~115ms | Rough total interaction time:  ~225ms |
[...]

PR #10262, opened 2022-01-19
Performance: Remove Extraneous Re-Renders in Frames Layer

### LINKED ISSUE #10079: Performance: Frames Layer remove empty frame element re-renders
Reduce empty re-renders of unaffected frame elements. POC done in a previous ticket and here's the rough implementation brief:

- memoize frameElement
- `handleSelectElement` from useCanvas: uses all of currentElements just to get the background element id. just pass the background element id as a dep instead of all current elements
- `isAnythingTransforming` from useTransform: calculate isLinkActive within `useTransform` selector instead
- go through the rest of the hooks and make sure selectors don't have empty referential equality when returning the same thing

Reference this comment here:
https://github.com/google/web-stories-wp/issues/10071#issuecomment-1003111218

Implement changes and document before and after flame charts for moving a selected element on the test page with 50 elements

maxyinger:
## Context
Part of ongoing performance work.

## Summary
Removes empty re-renders of frame elements.

Running against these tests:
https://github.com/google/web-stories-wp/issues/10158

**Element Selection**
| Before | After |
| ----- | -----|
| React | React |
| Longest Commit | Longest Commit |
| ~92ms | ~56ms |
| Frames Layer | Frames Layer |
| ~45ms | ~1.7ms |
| Screen Shot 2022-01-19 at 10 16 24 AM |  Screen Shot 2022-01-19 at 10 16 09 AM |
| Total JS | Total JS |
| ~167ms | ~120ms |
| Screen Shot 2022-01-19 at 10 28 02 AM |  Screen Shot 2022-01-19 at 10 23 37 AM |

**Starting the drag of element to update position**
| Before | After |
| ----- | -----|
| React | React |
| Longest Commit | Longest Commit |
| ~80ms | ~2ms |
| Frames Layer | Frames Layer |
| ~44ms | ~1.6ms |
| Screen Shot 2022-01-19 at 10 07 55 AM |  Screen Shot 2022-01-19 at 10 11 59 AM |
| Total JS | Total JS |
| ~182ms | ~36ms |
| Screen Shot 2022-01-19 at 10 04 24 AM |  Screen Shot 2022-01-19 at 10 05 04 AM |

**Ending the drag of element to update position**
| Before | After |
| ----- | -----|
| React | React |
| Longest Commit | Longest Commit |
| ~92ms | ~60ms |
| Frames Layer | Frames Layer |
| ~44ms | ~7.5ms |
| Screen Shot 2022-01-19 at 9 54 59 AM | Screen Shot 2022-01-19 at 9 51 20 AM |
| Total JS | Total JS |
| ~382ms | ~204ms |
| Screen Shot 2022-01-19 at 10 01 13 AM |  Screen Shot 2022-01-19 at 10 01 40 AM |

**Changing Layer Order**
_Screenshots of flame graphs are similar to above, so going to omit for brevity here_
| Before | After |
| ----- | ----- |
| React | React |
| Longest Commit | Longest Commit |
| ~92ms | ~55ms |
| Frames Layer | Frames Layer |
| ~42ms | ~3.6ms |
| Total JS | Total JS|
| ~182ms | ~127ms |

## Relevant Technical Choices
Better context selector usage and memoization.

## To-do
NA

## User-facing changes
NA

## Testing Instructions
NA

1.

## Checklist

---

Fixes #10079

### DIFF
--- a/packages/story-editor/src/components/canvas/framesLayer.js
+++ b/packages/story-editor/src/components/canvas/framesLayer.js
@@ -58,7 +58,7 @@ const DesignSpaceGuidelineBorder = styled.div`
   visibility: ${({ isVisible }) => (isVisible ? 'visible' : 'hidden')};
 `;
 
-const DesignSpaceGuideline = function DesignSpaceGuidline() {
+function DesignSpaceGuideline() {
   const setDesignSpaceGuideline = useCanvas(
     ({ actions }) => actions.setDesignSpaceGuideline
   );
@@ -72,7 +72,7 @@ const DesignSpaceGuideline = function DesignSpaceGuidline() {
       isVisible={isAnythingTransforming}
     />
   );
-};
+}
 
 function FramesLayer() {
   // were returning this directly because we want the elementIds array to be shallowly
@@ -89,7 +89,8 @@ function FramesLayer() {
   );
 
   const framesLayerRef = useRef(null);
-  // TODO: refactor `useCanvasKeys`. This is the last hook causing extraneous re-renders in this component.
+  // TODO: https://github.com/google/web-stories-wp/issues/10266
+  // refactor `useCanvasKeys`. This is the last hook causing extraneous re-renders in this component.
   // - pulls most of state from useStory and only creates actions and attaches them to hot keys
   // - extraneous re-renders in this component contribute only ~1ms to total re-render time,
   //   so this is a high hanging fruit with little reward.
--- a/packages/story-editor/src/components/canvas/framesLayer.js
+++ b/packages/story-editor/src/components/canvas/framesLayer.js
@@ -75,7 +75,7 @@ function DesignSpaceGuideline() {
 }
 
 function FramesLayer() {
-  // were returning this directly because we want the elementIds array to be shallowly
+  // We are returning this directly because we want the elementIds array to be shallowly
   // compared between re-renders. This allows element properties to update without re-rendering
   // this top level component.
   const elementIds = useStory(
--- a/packages/story-editor/src/components/canvas/framesLayer.js
+++ b/packages/story-editor/src/components/canvas/framesLayer.js
@@ -27,7 +27,7 @@ import { STORY_ANIMATION_STATE } from '@googleforcreators/animation';
  * Internal dependencies
  */
 import { useKeyDownEffect } from '@googleforcreators/design-system';
-import { DESIGN_SPACE_MARGIN } from '../../constants';
+import { DESIGN_SPACE_MARGIN, STABLE_ARRAY } from '../../constants';
 import {
   useStory,
   useCanvas,
@@ -45,7 +45,8 @@ const FramesPageArea = styled(PageArea)`
   pointer-events: initial;
 `;
 const marginRatio = 100 * (DESIGN_SPACE_MARGIN / PAGE_WIDTH);
-const DesignSpaceGuideline = styled.div`
+
+const DesignSpaceGuidelineBorder = styled.div`
   border: 1px solid ${({ theme }) => theme.colors.border.negativePress};
   left: ${marginRatio}%;
   right: ${marginRatio}%;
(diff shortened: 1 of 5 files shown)


packages/story-editor/src/components/canvas/framesLayer.js
```diff
@@ -57,28 +58,44 @@ const DesignSpaceGuideline = styled.div`
...
 `;
 
-function FramesLayer() {
-  const { currentPage, isAnimating } = useStory((state) => ({
-    currentPage: state.state.currentPage,
-    isAnimating: [
-      STORY_ANIMATION_STATE.PLAYING,
-      STORY_ANIMATION_STATE.SCRUBBING,
-    ].includes(state.state.animationState),
+const DesignSpaceGuideline = function DesignSpaceGuidline() {
```
swissspidy: 🤔 

Why not just `function DesignSpaceGuideline() {`?

Not to mention the typo in `Guidline`

packages/story-editor/src/components/canvas/framesLayer.js
```diff
@@ -57,28 +58,44 @@ const DesignSpaceGuideline = styled.div`
...
+      isVisible={isAnythingTransforming}
+    />
   );
+};
 
-  const { isAnythingTransforming } = useTransform((state) => ({
-    isAnythingTransforming: state.state.isAnythingTransforming,
-  }));
+function FramesLayer() {
+  // were returning this directly because we want the elementIds array to be shallowly
```
BrittanyIRL:
```suggestion
  // We're returning this directly because we want the elementIds array to be shallowly
```

packages/story-editor/src/components/canvas/framesLayer.js
```diff
@@ -57,28 +58,44 @@ const DesignSpaceGuideline = styled.div`
...
+      state.currentPage?.elements?.map((el) => el.id) || STABLE_ARRAY
+  );
+  const isAnimating = useStory(({ state }) =>
+    [STORY_ANIMATION_STATE.PLAYING, STORY_ANIMATION_STATE.SCRUBBING].includes(
+      state.animationState
+    )
+  );
 
   const framesLayerRef = useRef(null);
+  // TODO: refactor `useCanvasKeys`. This is the last hook causing extraneous re-renders in this component.
```
BrittanyIRL: Can you associate this todo with an issue?

packages/story-editor/src/components/canvas/framesLayer.js
BrittanyIRL: I'm also a little confused here, I get `const DesignSpaceGuideline` because you're using it as a component below (`<DesignSpaceGuideline />`) and you want to keep these checks internal to it, so setting it as a function means that it's only going to ever get updated when the component is rendered but why not `const DesignSpaceGuideline = () => {}` - wouldn't that do the same, I'm lost on where `DesignSpaceGuid[e]line()` is needing to be named?

BrittanyIRL (changes requested):
Testing this locally and found a lil bug from reducing the amount of rerenders: 
- set media as your background 
- use the quick action menu to replace that background media 
- notice the layers panel icon has updated to the new media but the canvas has not

packages/story-editor/src/components/canvas/framesLayer.js
maxyinger: ah I was trying something where I was wrapping this in a `forwardRef`, then forgot to remove this when I removed the `forwardRef` bit.

Thanks for pointing out. will clean up

maxyinger:
> Testing this locally and found a lil bug from reducing the amount of rerenders:
> 
> * set media as your background
> * use the quick action menu to replace that background media
> * notice the layers panel icon has updated to the new media but the canvas has not

hmph I'm not able to replicate this consistently. Was able to get it to fail once, but was also able to get it to fail once on main as well. I don't think it's related to these issues.

Can you get it to fail consistently on your machine? if so, can you get the same issue to fail consistently on main?

BrittanyIRL:
> > Testing this locally and found a lil bug from reducing the amount of rerenders:
> > 
> > * set media as your background
> > * use the quick action menu to replace that background media
> > * notice the layers panel icon has updated to the new media but the canvas has not
> 
> hmph I'm not able to replicate this consistently. Was able to get it to fail once, but was also able to get it to fail once on main as well. I don't think it's related to these issues.
> 
> Can you get it to fail consistently on your machine? if so, can you get the same issue to fail consistently on main?

So i can get it to fail on this branch and not on staging - it's not every time, try repeating the action a few times.

maxyinger:
@BrittanyIRL just to clarify, you're talking about this button right?

Screen Shot 2022-01-20 at 8 09 12 AM

If so, I'm doing it a lot of times and can't replicate it at all today. In theory, this should be related to the `DisplayLayer` too. Which I haven't touched in this PR. Hmph 🤔 . Can anybody else replicate it and share any more details on how to get it to occur on my machine?

As far as a baseline goes tho, I would check `main` instead of staging to see if you can reproduce the error. Since it's an image, it could be an origin issue or some image loading error that's getting swallowed and not a state/component thing 🤷

maxyinger:
okay interesting. able to get the image to fail loading in the icon and filters panel, but the canvas works fine for me.

Screen Shot 2022-01-20 at 8 41 00 AM


But just checked and I'm also getting this when running main on my machine

BrittanyIRL:
> okay interesting. able to get the image to fail loading in the icon and filters panel, but the canvas works fine for me.
> But just checked and I'm also getting this when running main on my machine

Yes, so I meant that the icon on the layers panel updates but the canvas isn't - the bad image lingers. We chatted a bit off github so i'm just catching up here. You're totally right it's on main too - so false alarm. Let's keep tabs on this though - seems related to video. 

https://user-images.githubusercontent.com/10720454/150401895-afa0b092-4775-4010-8ab0-30e8b7703d65.mp4

packages/story-editor/src/components/canvas/framesLayer.js
```diff
@@ -57,28 +58,45 @@ const DesignSpaceGuideline = styled.div`
...
+      state.currentPage?.elements?.map((el) => el.id) || STABLE_ARRAY
+  );
+  const isAnimating = useStory(({ state }) =>
+    [STORY_ANIMATION_STATE.PLAYING, STORY_ANIMATION_STATE.SCRUBBING].includes(
+      state.animationState
+    )
+  );
 
   const framesLayerRef = useRef(null);
+  // TODO: https://github.com/google/web-stories-wp/issues/10266
```
BrittanyIRL, line 92: 👍  thanks!

packages/story-editor/src/components/canvas/framesLayer.js
```diff
@@ -114,14 +132,10 @@ function FramesLayer() {
...
-          {currentPage &&
-            currentPage.elements.map((element) => {
-              return <FrameElement key={element.id} element={element} />;
-            })}
-          <DesignSpaceGuideline
-            ref={setDesignSpaceGuideline}
-            isVisible={isAnythingTransforming}
-          />
+          {elementIds.map((id) => {
+            return <FrameElement key={id} id={id} />;
```
BrittanyIRL, line 136: nit: typically we try to avoid keys being ids, no?

BrittanyIRL (approved):
This is awesome, it's amazing the difference this makes.

packages/story-editor/src/components/canvas/framesLayer.js
swissspidy: We want to avoid using _array indexes_ for keys, but using the unique element ID here is fine

swissspidy (approved):

Issue #10249, opened 2022-01-18
Rename all packages to use new namespace


swissspidy:
## Task Description

This will be a massive change, renaming all packages to use the new `@googleforcreators/` namespace, matching the GitHub org.

PR #10271, opened 2022-01-19
Change namespace of fonts package

swissspidy:
See #10249


spacedmonkey:
Adding +49.4 kB to wp-story-editor bundle seems like a lot. It likely because of this change. 

```
import getFontDataFromUrl from '@web-stories-wp/fonts/src/utils/getFontDataFromUrl';
```

Do we need to make this change?

packages/fonts/src/utils/index.js
```diff
@@ -16,3 +16,4 @@
 
 export { default as getFontCSS } from './getFontCSS';
 export { default as getGoogleFontURL } from './getGoogleFontURL';
+export { default as getFontDataFromUrl } from './getFontDataFromUrl';
```
mohdsayed: Was it not exported purposely because it was increasing the bundle size?  🤔

swissspidy: It shouldn't be in the fonts package in the first place. It is only used in `wp-dashboard`.

PR #10273, opened 2022-01-19
Template Details: Chips are divs not buttons

### LINKED ISSUE #10175: Template Detail: Chip Regression to "Button"
## Bug Description

Pointed out by @spacedmonkey  during bug bash, the chips here are buttons - they used to be divs as they are not clickable, but that setting in the design system must have been removed. 

## Expected Behaviour

Chips here aren't buttons. We should at the very least make sure these are divs and probably have a conversation with design about a different way to present this list that doesn't look like buttons :) 

Update 1/18/2022 after meeting with @agingoldseco 
- we're going to just fix the regression here so that the chips are divs not buttons. 

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

BrittanyIRL:
## Context

Called out in bug bash, template details chips that hold template tags are rendered as disabled buttons. They aren't clickable with our current implementation so we should not have them be buttons. 

## Summary

Forwards `div` to `Chip` component to override what element it's rendered as. I remember this being there at some earlier point but it got away. 

## User-facing changes

Chip is a div now not a button. No real changes to notice. 

## Testing Instructions

Verify this: 

Yes: 
Screen Shot 2022-01-19 at 2 50 50 PM 

No: 
image

## Checklist

---

Fixes #10175

### DIFF
--- a/packages/dashboard/src/app/views/exploreTemplates/modal/templateDetails/content/index.js
+++ b/packages/dashboard/src/app/views/exploreTemplates/modal/templateDetails/content/index.js
@@ -73,7 +73,7 @@ const PaginationContainer = styled.div`
         `}
 `;
 
-const TemplateTag = styled(Chip)`
+const TemplateTag = styled(Chip).attrs({ forwardedAs: 'div' })`
   margin-right: 12px;
   margin-bottom: 12px;
   > span {


samwhale (approved):
Dope 👍

BrookeGraham (approved):
Nice.

PR #10280, opened 2022-01-20
Change namespace of react package

swissspidy:
See #10249


c0b41:
why webstories using custom react dependencies? and why not peerDependencies?

swissspidy:
@c0b41 great question! we might switch to `peerDependencies` once we publish these packages to npm. Right now they are all private, so it does not matter at all.

PR #10288, opened 2022-01-20
Background Audio: Add Initial Captions Support

### LINKED ISSUE #8821: Background Audio: Captions Support
## Feature Description

As per https://github.com/google/web-stories-wp/issues/48#issuecomment-899622816, we might want to support captions for background audio in case it's used for voiceover.

For that, we'd need to use a `video` for the audio file instead of the `background-audio` attribute so that we could show captions this way. See https://iandevlin.com/blog/2015/12/html5/webvtt-and-audio/ for a good explanation.

## Acceptance Criteria

**Important note:** This is only for **page background audio**, not **story background auido**

* Add new feature flag, e.g. `enhancedPageBackgroundAudio`
    * Purposefully generic name so that we can also use it for #10010
*  In `PageBackgroundAudioPanel`, add a new "Add Captions" button to the page background setting (only if there's already audio of course).
    * Maybe easiest o achieve this by passing a new `supportsCaptions={isFeatureFlagEnabled}` prop to `BackgroundAudioPanelContent`.
    * The button should work similarly as for videos.
    * **Note:** this is behind a feature flag. UI is not finalized. We'll do that later.
* Update `OutputPage`:
    * If the feature flag is enabled and there is a caption file:
        * Do **not** set `background-audio` on `<amp-story-page>`.
        * Render a hidden `<amp-video>` with the audio file and the caption (see link above).
        * Append caption to `videoCaptions`, matching the video ID.
        * For the video ID, something like `page-{$id}-background-audio` should be descriptive enough
* Add unit tests

spacedmonkey:
## Summary

Add support for background audio captions. Add a new button to select / upload vtt files, the same way in which captions are selected for videos. 

This feature is hidden behind a feature flag and has to be enabled. 

Once enabled, instead of adding background-audio attribute to amp-story-page tag, it adds a hidden amp-video tag, so that captions can be supported. 

#### Upstream error

If you create mulitple pages, then go between pages, the video will not stop. This results in video that autoplay and play over top of each other. 

## Relevant Technical Choices

Reuse the existing caption UI for videos. 
Reuse the tracks and track resource prop types, no need to add a new prop type. 

## To-do

- Fix upstream bug. 

## Testing Instructions

#### Test as admin
1. Enable background audio caption experience. 
2. Create a story. 
3. Add background audio to page. 
4. Add caption to background audio. 
5. Repeat process for a couple of pages. 
6. Save story. 
7. Preview story. 

#### Test as contributor
1. Enable background audio caption experience. 
2. Create a story. 
3. Add background audio to page. 
4. Add caption to background audio. 
5. Change author to contributor
6. Save story. 
7. Login as contributor. 
8. Edit story. 
9. Remove background audio captions. 
10. Save story. 
11. Preview story. 

## Checklist

---

Fixes #8821


packages/story-editor/src/output/utils/test/getUsedAmpExtensions.js
```diff
@@ -135,4 +135,26 @@ describe('getUsedAmpExtensions', () => {
       ])
     );
   });
+
+  it('should include the amp-stroy-captions script if there is a background audio with tracks', () => {
```
swissspidy, line 149: Nit: also test that `amp-video` is included

packages/story-editor/src/components/panels/design/captions/captions.js
```diff
@@ -185,67 +151,14 @@ function CaptionsPanel({ selectedElements, pushUpdate }) {
...
-          )}
-        </>
-      )}
+      <CaptionsPanelContent
+        isMixedValue={isMixedValue}
+        tracks={tracks}
+        handleChangeTrack={handleChangeTrack}
+        handleRemoveTrack={handleRemoveTrack}
+        renderUploadButton={renderUploadButton}
+        uploadError={uploadError}
```
swissspidy: This should be all within `CaptionsPanelContent`, not passed as a prop.

But since this is completely unused anyway, let's just remove `uploadError` entirely.

packages/story-editor/src/output/page.js
```diff
@@ -81,15 +83,25 @@ function OutputPage({ page, autoAdvance = true, defaultPageDuration = 7 }) {
...
 
+  const hasBackgroundAudioWithTracks =
+    backgroundAudio?.src && tracks?.length > 0;
+
+  if (hasBackgroundAudioWithTracks) {
+    videoCaptions.push(`el-${backgroundAudio.id}-captions`);
+  }
+
+  const backgroundAudioSrc =
+    backgroundAudio?.src && !tracks?.length ? backgroundAudio.src : undefined;
```
swissspidy: It would be nice to re-use `hasBackgroundAudioWithTracks` here for improved readability

```js
const backgroundAudioSrc = !hasBackgroundAudioWithTracks && backgroundAudio?.src ? backgroundAudio?.src : undefined;
```

packages/story-editor/src/output/page.js
```diff
@@ -41,6 +42,7 @@ function OutputPage({ page, autoAdvance = true, defaultPageDuration = 7 }) {
     elements,
     backgroundColor,
     backgroundAudio,
+    tracks = [],
```
swissspidy: Maybe `tracks: backgroundAudioTracks = []` for improved readability?

So further down it's clearer that we deal with `videoTracks` and `backgroundAudioTracks`

packages/story-editor/src/output/test/page.js
```diff
@@ -1243,6 +1243,41 @@ describe('Page output', () => {
         'background-audio="https://example.com/audio.mp3"'
       );
     });
+    it('should add background audio as amp-video', async () => {
```
swissspidy, line 1249: An AMP validation test for background audio would be good

packages/story-editor/src/output/test/page.js
```diff
@@ -1243,6 +1243,41 @@ describe('Page output', () => {
...
+            },
+          ],
+          id: '123',
+          elements: [],
+        },
+        autoAdvance: false,
+        defaultPageDuration: 7,
+      };
+
+      const { container } = render(<PageOutput {...props} />);
```
swissspidy, line 1278: We'd want an assertion here that the `background-audio` attribute on `amp-story-page` is not set if we have captions.

swissspidy: Could also add an assertion that the ID of `amp-story-captions` matches the video ID.

The snapshot test doesn't make that obvious.

Aside: why the snapshot test? Seems to be the first one in this file. Just curious.

packages/story-editor/src/components/panels/design/captions/captions.js
spacedmonkey: Fixed https://github.com/GoogleForCreators/web-stories-wp/pull/10288/commits/d2bead5d655e2e350b9bc638c3b495e1313bc4f8

packages/story-editor/src/output/page.js
spacedmonkey: Fixed https://github.com/GoogleForCreators/web-stories-wp/pull/10288/commits/c4702cdff4a2b35dc06052d7e3d0c6898bd4696d

packages/story-editor/src/output/page.js
spacedmonkey: Fixed in https://github.com/GoogleForCreators/web-stories-wp/pull/10288/commits/c4702cdff4a2b35dc06052d7e3d0c6898bd4696d

packages/story-editor/src/output/test/page.js
spacedmonkey: > Aside: why the snapshot test? Seems to be the first one in this file. Just curious.

I wanted to avoid generating a snapshot file. But as amp video tag was complex, it could not be avoided. I thought, if we already have a snapshot file for this test, why not use it in both spaces. 

> Could also add an assertion that the ID of amp-story-captions matches the video ID. 

Do we do this for other amp-video / captions? If not, why is it needed here. 

> We'd want an assertion here that the background-audio attribute on amp-story-page is not set if we have captions.

Good idea!

packages/story-editor/src/output/test/page.js
spacedmonkey: Fixed in https://github.com/GoogleForCreators/web-stories-wp/pull/10288/commits/37c26ccee02ec3c2c449ea3e626cc6772d35bd58

packages/story-editor/src/output/test/page.js
spacedmonkey: Added a test

packages/story-editor/src/output/page.js
```diff
@@ -85,11 +87,23 @@ function OutputPage({ page, autoAdvance = true, defaultPageDuration = 7 }) {
     )
     .map(({ id: videoId }) => `el-${videoId}-captions`);
 
+  const hasBackgroundAudioWithTracks =
+    backgroundAudio?.src && backgroundAudioTracks?.length > 0;
+
+  if (hasBackgroundAudioWithTracks) {
+    videoCaptions.push(`el-${backgroundAudio.id}-captions`);
```
swissspidy: This needs to be the same ID that you pass to `HiddenAudio`, which is the page ID. Not `backgroundAudio.id`

```suggestion
    videoCaptions.push(`el-${id}-captions`);
```

packages/story-editor/src/output/test/page.js
swissspidy: > Do we do this for other amp-video / captions? If not, why is it needed here.

I don't know, but we should add one.

Because right now it's incorrect as per my other message https://github.com/GoogleForCreators/web-stories-wp/pull/10288#discussion_r790741798

So a test would catch this bug.

packages/story-editor/src/output/utils/HiddenAudio.js
```diff
@@ -0,0 +1,75 @@
...
+    src,
+  };
+
+  const videoProps = {
+    autoPlay: 'autoplay',
+    layout: 'responsive',
+    width: '0',
+    height: '0',
+    loop: 'loop',
+  };
```
swissspidy: Since these are static we could just add them directly inline to `<amp-video` below

swissspidy: Also, looks like we can just provide `layout="nodisplay"`. No width or height necessary

spacedmonkey: > Also, looks like we can just provide layout="nodisplay". No width or height necessary

This based on your recommendations.

swissspidy: I know. Because I first thought `layout="nodisplay"` does not work, but it does appear to work. Now this is my new recommendation.

packages/story-editor/src/output/test/page.js
spacedmonkey: Fixed.

packages/story-editor/src/output/utils/HiddenAudio.js
spacedmonkey: Fixed.

packages/story-editor/src/output/test/page.js
```diff
@@ -1450,6 +1495,38 @@ describe('Page output', () => {
...
+            {
+              track: 'https://example.com/track.vtt',
+              trackId: 123,
+              trackName: 'track.vtt',
+              id: 'rersd-fdfd-fdfd-fdfd',
+              srcLang: '',
+              label: '',
+              kind: 'captions',
+            },
+          ],
```
swissspidy: With #10350 in mind, what do you think about slightly changing the data format like so:

```js
backgroundAudio: {
  resource: { src, id, mimeType }
  tracks: [ ... ],
  loop: true/false
}
```

This feels a bit cleaner

spacedmonkey: I do like this idea, how I do think we should do this in a follow up ticket. It would require a migration of data. I think it opens up a larger conversion. For example, I think we should have a audio resource type considering the conversation around https://github.com/GoogleForCreators/web-stories-wp/issues/10002.

swissspidy: It's easier to do a migration _before_ introducing new data like `tracks`, not _afterwards_

A dedicated/full audio resource type is not something I want us to invest time in right now. That can be looked into once there is an actual need.

For this use case here, something simple like `resource: { src, id, mimeType }` should suffice

packages/story-editor/src/output/test/page.js
```diff
@@ -1450,6 +1495,38 @@ describe('Page output', () => {
...
+          animations: [],
+          elements: [],
+        },
+        autoAdvance: true,
+        defaultPageDuration: 11,
+      };
+
+      await expect(<PageOutput {...props} />).toBeValidAMPStoryPage();
```
swissspidy: Interesting that this passes despite the missing video poster 🤔

packages/story-editor/src/output/test/page.js
spacedmonkey: Okay, but the resource will need to be `{ src, id, mimeType, length }` for the looping ticket. There is an augment for title and lengthFormatted, as those are useful.

swissspidy: I haven't listed all properties for brevity reasons. We should use whatever fields we already have and move them within `resource`. Including `length` (in the separate PR you are already working on) and title and length sounds reasonable.

spacedmonkey: Should story also be updated?

swissspidy: What do you mean? Adding a data migration?

swissspidy (approved):

PR #10313, opened 2022-01-21
Dashboard: Template Details Accessibility Improvements

### LINKED ISSUE #10176: Template Details: Clean up A11Y
## Feature Description

@spacedmonkey brought up a good question about screen readers announcing the proper info on the templates detail page. We should make sure everything's working properly. 

Right now, screen readers announce the template title and the description. The chips holding the keywords and the color tags are not, these elements can't be interacted with. Should they be turned into lists instead so that they could be given proper structures by screen readers?

Additionally, announcing the template title when the modal opens might be a nice improvement. 

Update should include some tests to make sure functionality doesn't regress. 

## Additional Context

Template Details has never actually been finished! It was a prototype that never got priority, so let's use this as an opportunity to scale what we need to.

BrittanyIRL:
## Context

Now that the template details view is a modal, making sure everything is accessible. 

## Summary

Few tweaks to improve screen reader experience and make sure content is viewable. 

## Relevant Technical Choices

Noticed that `setAppElement` is no longer in play on either the dashboard or the editor except for in tests and storybook, I'm sure this was a 1 line regression that was super easy to miss in review of separating into separate packages and honestly, focus on our modals was behaving just fine with tabbing so this isn't really that dire - it's something that was flagged when I ran this modal through aXe though and _should_ be brought back around. I targeted it like I did (`document.body.setAttribute('id', 'dashboardWrapper');  setAppElement('#dashboardWrapper > div:first-of-type');`) because I don't want it to be an id that's dependent on the CMS - so much work has been done to separate the dashboard from WordPress that I don't want to introduce anything new. I figure by setting a new id of dashboardWrapper on the body and grabbing whatever the first div is, we can be agnostic to where ever the dashboard is embedded.  Alternatively, we could add a config value and pass it in but that just seems extra.
 

## To-do

Follow up ticket to setAppElement in editor as well for modals there if everyone's ok with this implementation. 

## User-facing changes

Shouldn't be any visual changes. 

If you use a screen reader (chrome vox or voiceover on mac) you should get the following: 
1. Helpful messages on all focusable content 
2. Announcing new template title when you use the arrows (works on chrome vox, I think I've got user error on my voice over set up) 
3. Announcing when the dialog opens or closes 
4. If you select text, it'll get read. 

If you tab, you'll hit: 1) close button 2) use template button 3) (if available) prev template button 4) group of buttons to change preview image 5) next template button. 

## Testing Instructions

1.

No

No

No

## Checklist

---

Fixes #10176

### DIFF
--- a/packages/dashboard/src/app/views/exploreTemplates/index.js
+++ b/packages/dashboard/src/app/views/exploreTemplates/index.js
@@ -17,6 +17,7 @@
 /**
  * External dependencies
  */
+import { sprintf, __ } from '@googleforcreators/i18n';
 import {
   useMemo,
   useEffect,
@@ -25,7 +26,10 @@ import {
   useRef,
 } from '@googleforcreators/react';
 import { trackEvent, trackScreenView } from '@googleforcreators/tracking';
-import { uniqueEntriesByKey } from '@googleforcreators/design-system';
+import {
+  uniqueEntriesByKey,
+  useLiveRegion,
+} from '@googleforcreators/design-system';
 
 /**
  * Internal dependencies
@@ -41,6 +45,7 @@ import Header from './header';
 import TemplateDetailsModal from './modal';
 
 function ExploreTemplates() {
+  const speak = useLiveRegion();
   const [isDetailsViewOpen, setIsDetailsViewOpen] = useState(false);
   const [activeTemplate, setActiveTemplate] = useState(null);
   const [activeTemplateIndex, setActiveTemplateIndex] = useState(0);
@@ -160,6 +165,7 @@ function ExploreTemplates() {
       if (idRef.current) {
         idRef.current = undefined;
       }
+
       replace(`?id=${currentTemplate.id}&isLocal=${currentTemplate.isLocal}`);
     },
     [replace, orderedTemplates]
@@ -172,17 +178,19 @@ function ExploreTemplates() {
         title && trackScreenView(title);
 
         if (newIsOpen && id) {
+          speak(__('Enter detail templates view', 'web-stories'));
           updateTemplateView(id);
         }
 
         if (!newIsOpen) {
           replace('');
+          speak(__('Exit detail templates view', 'web-stories'));
         }
 
         return newIsOpen;
       });
     },
-    [replace, updateTemplateView]
+    [replace, speak, updateTemplateView]
   );
 
   const switchToTemplateByOffset = useCallback(
@@ -191,8 +199,15 @@ function ExploreTemplates() {
       setActiveTemplate(newTemplate);
       setActiveTemplateIndex(offset);
       replace(`?id=${newTemplate.id}&isLocal=${newTemplate.isLocal}`);
+      speak(
+        sprintf(
+          /* translators: %s: template title */
+          __('Viewing %s', 'web-stories'),
+          newTemplate?.title
+        )
+      );
     },
-    [orderedTemplates, replace]
+    [orderedTemplates, replace, speak]
   );
 
   const templateActions = useMemo(
--- a/packages/dashboard/src/app/views/exploreTemplates/index.js
+++ b/packages/dashboard/src/app/views/exploreTemplates/index.js
@@ -178,7 +178,6 @@ function ExploreTemplates() {
         title && trackScreenView(title);
 
         if (newIsOpen && id) {
-          speak(__('Enter detail templates view', 'web-stories'));
           updateTemplateView(id);
         }
 
@@ -203,7 +202,7 @@ function ExploreTemplates() {
         sprintf(
           /* translators: %s: template title */
           __('Viewing %s', 'web-stories'),
-          newTemplate?.title
+          newTemplate.title
         )
       );
     },
--- a/packages/dashboard/src/app/views/exploreTemplates/modal/templateDetails/header/index.js
+++ b/packages/dashboard/src/app/views/exploreTemplates/modal/templateDetails/header/index.js
@@ -61,7 +61,7 @@ function Header({
         type={BUTTON_TYPES.TERTIARY}
         variant={BUTTON_VARIANTS.SQUARE}
         size={BUTTON_SIZES.SMALL}
-        aria-label={__('Close detail view', 'web-stories')}
+        aria-label={__('Close', 'web-stories')}
         onClick={handleDetailsToggle}
       >
         <Icons.CrossLarge />
--- a/packages/dashboard/src/app/views/exploreTemplates/modal/templateDetails/header/index.js
+++ b/packages/dashboard/src/app/views/exploreTemplates/modal/templateDetails/header/index.js
@@ -25,7 +25,6 @@ import {
   BUTTON_SIZES,
   BUTTON_VARIANTS,
   Icons,
-  themeHelpers,
 } from '@googleforcreators/design-system';
 /**
  * Internal dependencies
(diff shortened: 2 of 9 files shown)


packages/dashboard/src/app/views/exploreTemplates/modal/templateDetails/header/index.js
```diff
@@ -62,13 +57,6 @@ function Header({
     templateActions || {};
   return (
     <Nav>
-      <HiddenHeading>
```
BrittanyIRL, line 65: Redundant, no need since this is a modal now

packages/dashboard/src/dashboard.js
```diff
@@ -48,6 +49,11 @@ function Dashboard({ config, children }) {
     ...externalDesignSystemTheme,
     colors: lightMode,
   };
+  // Set up modal focus trap on app's mount
+
+  document.body.setAttribute('id', 'dashboardWrapper');
+  setAppElement('#dashboardWrapper > div:first-of-type');
```
maxyinger, line 55: is this equivalent to: `setAppElement('body > div:first-of-type')` ?

Also should this be in an effect so it only runs once?

maxyinger: Just want to clarify here, the assumption you're making is that the first div in the body will be the root element we're mounting the dashboard inside of?

BrittanyIRL: > should this be in an effect so it only runs once? 

It only runs on initial mount of the `Dashboard` so it's not getting called again as you navigate the dashboard so putting this in an effect doesn't do anything. 

> is this equivalent to: setAppElement('body > div:first-of-type') ?

Yes, but I wanted to add the id so that it reads easier or can be updated more readily if my assumption about the first div in the body being the root element to mount goes sour. Honestly, this should hold true - the way that the Dashboard is getting exported and used will maintain this structure. It feels icky but this was the best way I could come up with grabbing the right element to set for hiding the page when we're in a modal without also hiding the modal _and_ being agnostic of WordPress. I'll play with it a little more today just to triple check.

maxyinger: Cool sounds good. I'm not as familiar with how we're actually mounting the dashboard, or how this looks porting to another CMS, so just wanted to check.

BrittanyIRL: i'm glad you did, this part feels weird but since it's essentially embedded in WordPress i think it's gonna have to stay a little weird - at least for now. 😿

packages/dashboard/src/app/views/exploreTemplates/index.js
```diff
@@ -191,8 +199,15 @@ function ExploreTemplates() {
       setActiveTemplate(newTemplate);
       setActiveTemplateIndex(offset);
       replace(`?id=${newTemplate.id}&isLocal=${newTemplate.isLocal}`);
+      speak(
+        sprintf(
+          /* translators: %s: template title */
+          __('Viewing %s', 'web-stories'),
```
samwhale, line 204: 💭 Haven't found a way to read the template description using voice box. Maybe it should be announced when the template is opened?

samwhale (approved):
I like this.

However, seems like there's a lot of text being said when we open a template detail page. This is what is announced:

```
"Enter Dialog"
"Exited main viewing of 'avail' templates with 59 items" // is this necessary?
"Template details for '<template name>'
```

Would be cool if the middle line could be removed, but also I want to add a way for the user to read the description with a screen reader. Maybe it could be added to the `speak` that's announced on open.

packages/dashboard/src/app/views/exploreTemplates/index.js
BrookeGraham: This doesn't seem to be reading the title for me when I switch to a new template. Does it for y'all?

packages/dashboard/src/app/views/exploreTemplates/index.js
```diff
@@ -191,8 +199,15 @@ function ExploreTemplates() {
...
+          newTemplate?.title
```
BrookeGraham: `newTemplate` should always be defined since we are declaring it up above, the optional chaining shouldn't be necessary.

packages/dashboard/src/app/views/exploreTemplates/modal/templateDetails/karma/templateDetails.karma.js
```diff
@@ -105,6 +105,12 @@ describe('See template details modal', () => {
   }
 
   describe('Action: Navigate template details modal', () => {
+    it('should pass accessibility tests', async () => {
+      const templateDetails = fixture.screen.queryByRole('dialog');
+      expect(templateDetails).toBeTruthy();
+      await expectAsync(templateDetails).toHaveNoViolations();
```
BrookeGraham, line 111: Just curious, why the `expectAsync`?

BrittanyIRL:
> ```
> "Enter Dialog"
> "Exited main viewing of 'avail' templates with 59 items" // is this necessary?
> "Template details for '<template name>'
> ```
> 
> 
>     
> 
>   
> 
> Would be cool if the middle line could be removed, but also I want to add a way for the user to read the description with a screen reader. Maybe it could be added to the `speak` that's announced on open.

It is quite a bit of words! So i can't do anything about the 'exited main view...' because that's tied to what is going out of focus and is really important to knowing where you are on the page. HOWEVER, the 'enter dialog' getting spoken, is overkill. I'm going to remove that because it's not giving a user any info they don't already have.

packages/dashboard/src/app/views/exploreTemplates/modal/templateDetails/karma/templateDetails.karma.js
BrittanyIRL: yeah! so basically just so the expect knows it needs to wait on checking for violations. It's set up like this in the docs: https://github.com/nickcolley/jest-axe and it'll fail without it because the check isn't done yet.

BrittanyIRL:
@BrookeGraham re 
> This doesn't seem to be reading the title for me when I switch to a new template. Does it for y'all? 

It works for me using chrome vox, are you using mac's voice over because there's some extra settings there to update

packages/dashboard/src/app/views/exploreTemplates/index.js
BrittanyIRL: So, you need to enable reading buttons for the title to get read when you're changing templates in the modal pagination like this: 
Screen Shot 2022-01-27 at 1 33 41 PM


There's a similar setting for reading static content. You can hear it getting read out if you highlight the text you want read. It's hard because like 0.01% of people who use a screen reader are using Mac's Voiceover and chrome vox gets closer but not by much. So if you're actually using this as a feature you've got paid for software that's combing through the page well.

samwhale:
I'm also using chromevox @BrookeGraham

packages/dashboard/src/app/views/exploreTemplates/modal/templateDetails/karma/templateDetails.karma.js
BrookeGraham: Awesome Thanks!

BrookeGraham:
>It works for me using chrome vox, are you using mac's voice over because there's some extra settings there to update

Yep I was using native macosx voiceover.

BrookeGraham (approved):
🧝‍♀️ Love it. Thanks for the clarifying info.

samwhale (approved):
🚢

maxyinger (approved):

PR #10343, opened 2022-01-24
Improve save-blocking media uploading checks

### LINKED ISSUE #10341: Media: Only prevent saving if actually uploading to story
## Feature Description

Right now we disable saving a story if _any_ upload is in progress, even if just uploading to the library and not the canvas

Instead, we only want to disable saving a story if any media item within the story itself is currently uploading, as only those are concerning due to blob URLs.

swissspidy:
## Context

Right now we disable saving a story if any upload is in progress, even if just uploading to the library and not the canvas

Instead, we only want to disable saving a story if any media item within the story itself is currently uploading, as only those are concerning due to blob URLs.

## Summary

This PR improves the media uploading checks we have in multiple places to only disable save buttons when any story elements are currently being uploaded.

So if I upload a video only to the media library, I can still save the story even when the upload is still in progress.

## Relevant Technical Choices

Uses a new hook for easier reusability

## To-do

* [x] Unit tests
* [x] Test that it works as expected with video poster generation

## Testing Instructions

1. Drag & drop a video or image to the media library
2. See that the story can be saved/published
3. Drag & drop a video or image to the canvas
4. See that the update/publish button is disabled while the upload is in progress

No

No

No

## Checklist

---

Fixes #10341


packages/wp-story-editor/src/components/documentPane/status/status.js
```diff
@@ -101,6 +104,7 @@ function StatusPanel() {
       value: 'private',
       label: __('Private', 'web-stories'),
       helper: __('Visible to site admins & editors only', 'web-stories'),
+      disabled: isUploading && visibility !== 'private',
```
miina, line 107: Nit: Might be nice to move all these status strings to constants.

packages/story-editor/src/utils/useIsUploadingToStory.js
```diff
@@ -0,0 +1,64 @@
...
+    useLocalMedia(
+      ({
+        state: { isCurrentResourceUploading, isCurrentResourceProcessing },
+      }) => ({
+        isCurrentResourceUploading,
+        isCurrentResourceProcessing,
+      })
+    );
+
+  // TODO: What about video posters?
```
miina, line 52: Are background audios, publisher logos, Story poster images also relevant?

packages/wp-story-editor/src/components/documentPane/status/status.js
miina: Also wondering if all these should be disabled when uploading -- for consistency, even though it doesn't practically make a difference, thoughts?

packages/story-editor/src/utils/useIsUploadingToStory.js
swissspidy: Great question! Since they cannot be uploaded via drag and drop or be optimized/trimmed/muted they will never be blob URLs that we need to watch out for. So, no, not relevant.

packages/wp-story-editor/src/components/documentPane/status/status.js
swissspidy: Hmm good question. I think I lean towards not disabling if not really needed.

packages/story-editor/src/utils/useIsUploadingToStory.js
miina: Alright, will become relevant with https://github.com/GoogleForCreators/web-stories-wp/issues/6532 then, I suppose.

miina (approved):

PR #10361, opened 2022-01-24
Context Menu: Refactor right click menu provider

### LINKED ISSUE #10212: Code quality: refactor the right-click menu provider into smaller files
## Feature Description
Currently, the [right-click menu provider file is 1200+ lines](https://github.com/google/web-stories-wp/blob/e7cb5fe4dee961dbfc318dbc7b3c75b3373b6a02/packages/story-editor/src/app/rightClickMenu/provider.js). It would be good to refactor it into smaller files for readability and for simpler management of the file, e.g. by menu items per element + common items, or some other way.

samwhale:
## Summary

Pulled out code from the massive [right click menu provider](https://github.com/GoogleForCreators/web-stories-wp/blob/e7cb5fe4dee961dbfc318dbc7b3c75b3373b6a02/packages/story-editor/src/app/rightClickMenu/provider.js) into smaller hooks.

This was one way to approach it, but happy to hear other ideas.

## Relevant Technical Choices

Note: All of these actions are tested in the karma tests already. May be redundant to test them at the unit level.

Paying special attention to the values that are pulled out of context. Not every change should cause updates in each hook 😉.

@merapi Has a different approach here: https://github.com/GoogleForCreators/web-stories-wp/blob/7a4914adabb6cc691f2d0eaccad5d9cc056cf6af/packages/story-editor/src/app/rightClickMenu/menus/emptyStateMenu.js. If y'all like setting up the objects in these hooks we can do that here instead. For now I've left that step in the provider to keep the hooks from getting too bloated.

## To-do

- [x] finish moving actions
- [x] add unit tests now that we can do that
- [ ] ~identify performance improvements (if any) using flame charts~ will be done in a later issue #10459 
    - Note: since these hooks all pour into the same provider, it may be hard to see any specific performance improvements. Especially with `useApplyStyle` (`useRichTextFormatting` makes the browser do a lot of work).

## Testing Instructions

Perform these steps for each element type and on the background:
1. Open the right click menu 
2. Test each action individually
3. All should act the same as before

no

no

no

## Checklist

---

Fixes #10212


packages/story-editor/src/app/rightClickMenu/hooks/useElementActions.js
```diff
@@ -0,0 +1,201 @@
...
+    duplicateElementById,
+    selectedElements,
+    setBackgroundElement,
+    updateElementsById,
+  } = useStory(({ state, actions }) => ({
+    clearBackgroundElement: actions.clearBackgroundElement,
+    duplicateElementById: actions.duplicateElementById,
+    selectedElements: state.selectedElements,
+    setBackgroundElement: actions.setBackgroundElement,
+    updateElementsById: actions.updateElementsById,
```
samwhale: Only story state that should re-render this hook is `selectedElements`

packages/story-editor/src/app/rightClickMenu/hooks/useLayerActions.js
```diff
@@ -0,0 +1,151 @@
...
+import { useStory } from '../..';
+
+const useLayerActions = () => {
+  const { arrangeElement, elements, selectedElement } = useStory(
+    ({ state, actions }) => ({
+      arrangeElement: actions.arrangeElement,
+      elements: state.currentPage?.elements || [],
+      selectedElements: state.selectedElements?.[0] || null,
+    })
+  );
```
samwhale: Not returning `currentPage`, but rather the `elements` on current page.

packages/story-editor/src/app/rightClickMenu/hooks/usePageActions.js
```diff
@@ -0,0 +1,95 @@
...
+    selectedElementType,
+  } = useStory(({ state, actions }) => ({
+    addPageAt: actions.addPageAt,
+    currentPage: state.currentPage,
+    deleteCurrentPage: actions.deleteCurrentPage,
+    isSelectedElementBackground:
+      state.selectedElements?.[0]?.background || false,
+    pageLength: state.pages?.length || 1,
+    selectedElementType: state.selectedElements?.[0]?.type || null,
+  }));
```
samwhale, line 49: Page actions need the `currentPage` for duplication purposes.

packages/story-editor/src/app/rightClickMenu/hooks/useCopyPasteActions.js
```diff
@@ -0,0 +1,250 @@
...
+import { UNDO_HELP_TEXT } from './constants';
+
+/**
+ * Creates the right click menu copy/paste actions.
+ *
+ * @param {Function} dispatch The right click reducer dispatch function
+ * @param {Object} copiedElement Stored element styles
+ * @return {Object} Right click menu copy/paste actions
+ */
+const useCopyPasteActions = (dispatch = noop, copiedElement) => {
```
samwhale: 💭 Don't love having to pass in the dispatch function and our stored state from the reducer. However, the reducer is initialized in the provider. Do y'all have any ideas of better things to do here?

Making an entire new context for one function and this state seems like overkill.

packages/story-editor/src/app/rightClickMenu/provider.js
```diff
@@ -61,14 +39,14 @@ import rightClickMenuReducer, {
...
-  __('Press %s to undo the last change', 'web-stories'),
-  prettifyShortcut('mod+z')
-);
+import {
+  useCopyPasteActions,
+  useElementActions,
+  useLayerActions,
+  usePageActions,
+  usePresetActions,
+} from './hooks';
```
miina: I'll have more time to review it next week but generally makes sense to me, dividing into hooks like this.

We could additionally even create a separate file for each element as well so that the relevant hooks would be called in those files and the provider would just use the element hooks.
E.g. for text
```
// In provider.
const textItems = useTextItems(...);
const pageItems = usePageItems(...);

// In `useTextItems`
import { useCopyPasteActions, useElementActions, useLayerActions } from './hooks`;
...
// In `usePageItems`
import { useCopyPasteActions, usePageActions, useLayerActions } from './hooks';

... etc.
```

In this case, if there are items inside the `useElementActions` that are for text only, those could be added to `useTextItems` directly and `useElementActions` could include actions that are for at least 2 element types (unique items would be in the hook per element instead).

Just an idea without having gone through all the hook files to see how many unique/common items there are. Thoughts?

samwhale: Thanks for looking at this! Sorry it's so big. I prefer to make smaller prs, but since this mostly is moving code around I figured it'd be better to do it all at once

I do like the idea of having hooks for each element type and grabbing those options. However, I've had experience with nesting custom hooks inside other custom hooks. IMO it's better not to nest custom hooks inside other custom hooks because the logic becomes very spread out and dependencies can get muddy. One change in a hook can affect `useEffects` in other hooks and it's hard to tell until a bug comes up.

I prefer to keep hooks one level deep. What do you think?

I debated making just hooks for each menu but there would be a lot of repeated code in the actions for each. However, we could make pure functions for the actions and pass in the necessary scoped variables and functions as arguments.

packages/story-editor/src/app/rightClickMenu/hooks/useCopyPasteActions.js
maxyinger: looking at `ACTION_TYPES.COPY_ELEMENT_STYLES` & `copiedElement`, I'm not seeing why they're needed to be coupled with the reducer state.

Would it be cleaner just to `const [copiedElement, setCopiedElement] = useState(null)` internally here? and return `copiedElement` along with `onCopyStyles` & `onPasteStyles` from this hook?

maxyinger (commented):
Still reading over, but looks good so far!

packages/story-editor/src/app/rightClickMenu/hooks/useCopyPasteActions.js
samwhale: Yeah, there's no need for that state to be included in the reducer state. Just how it was set up initially

Good catch - it would be better to move that state into this hook.

packages/story-editor/src/app/rightClickMenu/provider.js
maxyinger: ➕ Yea I'm in favor of minimizing nesting of hooks. Adds complexity and scope to tracking down causes of a re-render

packages/story-editor/src/app/rightClickMenu/hooks/useCopyPasteActions.js
```diff
@@ -0,0 +1,238 @@
...
+ */
+import { useSnackbar } from '@googleforcreators/design-system';
+import { __ } from '@googleforcreators/i18n';
+import { useCallback, useRef, useState } from '@googleforcreators/react';
+import { trackEvent } from '@googleforcreators/tracking';
+import { v4 as uuidv4 } from 'uuid';
+
+/**
+ * Internal dependencies
+ */
```
maxyinger: remove?

packages/story-editor/src/app/rightClickMenu/hooks/useCopyPasteActions.js
```diff
@@ -0,0 +1,238 @@
...
+            return {
+              ...updates,
+              ...sizeUpdates,
+            };
+          },
+        });
+      }
+    },
+    [selectedElementId, updateElementsById]
+  );
```
maxyinger, line 118: since both `selectedElementId` & `updateElementsById` are coming from the story provider, and every other dependency in the scope of this function is passed in as args or a module dep, we can make this a story reducer action to limit re-renders. something like `updateSelectedElements`. Probably useful in other areas too.

Can probably look as simple as:
```
  const pushUpdate = useCallback(
    (update, commitValues) => updateSelectedElements({update, commitValues}), 
    [updateSelectedElements]
  );
```

Should cut down on the size of this file too 🙌 .

packages/story-editor/src/app/rightClickMenu/hooks/useCopyPasteActions.js
```diff
@@ -0,0 +1,238 @@
...
+
+  const handleApplyStyle = useApplyStyle({ pushUpdate });
```
maxyinger, line 120: Okay, turns out one of the main deps of `useApplyStyle` that causes it to update is the argument `pushUpdate` here. Since this is currently recomputing on every element selection, this causes `useApplyStyle` to have some expensive computations every time `selectedElementId` updates:
https://github.com/GoogleForCreators/web-stories-wp/blob/cd6886589628f0e2813d9b533db7b40850f78684/packages/story-editor/src/components/panels/design/textStyle/stylePresets/useApplyStyle.js#L29
https://github.com/GoogleForCreators/web-stories-wp/blob/cd6886589628f0e2813d9b533db7b40850f78684/packages/story-editor/src/components/panels/design/textStyle/stylePresets/useApplyStyle.js#L50
https://github.com/GoogleForCreators/web-stories-wp/blob/cd6886589628f0e2813d9b533db7b40850f78684/packages/story-editor/src/components/panels/design/textStyle/stylePresets/useApplyStyle.js#L40
https://github.com/GoogleForCreators/web-stories-wp/blob/cd6886589628f0e2813d9b533db7b40850f78684/packages/story-editor/src/components/panels/design/textStyle/stylePresets/useApplyStyle.js#L61
https://github.com/GoogleForCreators/web-stories-wp/blob/cd6886589628f0e2813d9b533db7b40850f78684/packages/story-editor/src/components/panels/design/textStyle/useRichTextFormatting.js#L86
https://github.com/GoogleForCreators/web-stories-wp/blob/cd6886589628f0e2813d9b533db7b40850f78684/packages/story-editor/src/components/panels/design/textStyle/useRichTextFormatting.js#L115
https://github.com/GoogleForCreators/web-stories-wp/blob/cd6886589628f0e2813d9b533db7b40850f78684/packages/story-editor/src/components/panels/design/textStyle/useRichTextFormatting.js#L109
https://github.com/GoogleForCreators/web-stories-wp/blob/cd6886589628f0e2813d9b533db7b40850f78684/packages/story-editor/src/components/panels/design/textStyle/useRichTextFormatting.js#L176
https://github.com/GoogleForCreators/web-stories-wp/blob/cd6886589628f0e2813d9b533db7b40850f78684/packages/story-editor/src/components/panels/design/textStyle/useRichTextFormatting.js#L136-L175

Results in all this work that contributes to the long render time of the right click menu provider when the selected element updates:
Screen Shot 2022-02-01 at 11 25 13 AM

By just making that ` updateSelectedElements` that's stable between selection updates, then updating this to use a proper selector:
https://github.com/GoogleForCreators/web-stories-wp/blob/cd6886589628f0e2813d9b533db7b40850f78684/packages/story-editor/src/components/panels/design/textStyle/useRichTextFormatting.js#L87-L90

And adding this filter step here:
https://github.com/GoogleForCreators/web-stories-wp/blob/cd6886589628f0e2813d9b533db7b40850f78684/packages/story-editor/src/components/panels/design/textStyle/useRichTextFormatting.js#L101-L106
```
    // loop over all elements, find info for content and reduce to common value
    // (setting MULTIPLE_VALUE appropriately)
    return selectedElements
      .map(({ content }) => content)
      .filter((v) => v !== undefined)
      .map(getHTMLInfo)
      .reduce(reduceWithMultiple, {});
```

You should be able to cut down the render time here significantly. I was able to knock down this step down from ~3ms -> 0.1ms doing so:
Screen Shot 2022-02-01 at 11 42 35 AM

maxyinger: @miina here's a great example of how nesting hooks leads to complexity in update logic and can negatively impact transparency and performance

packages/story-editor/src/app/rightClickMenu/hooks/useCopyPasteActions.js
```diff
@@ -0,0 +1,238 @@
...
+    selectedElementAnimations,
+    updateElementsById,
+  } = useStory(({ state, actions }) => ({
+    addAnimations: actions.addAnimations,
+    selectedElement: state.selectedElements?.[0],
+    selectedElementAnimations: state.selectedElementAnimations,
+    updateElementsById: actions.updateElementsById,
+  }));
+
+  const { showSnackbar } = useSnackbar();
```
maxyinger: this should make this stable for the life of the snackbar provider and not cause unnecessary updates
```
  const showSnackbar  = useSnackbar(value => value.showSnackbar);
```

packages/story-editor/src/app/rightClickMenu/hooks/useCopyPasteActions.js
```diff
@@ -0,0 +1,238 @@
...
+      );
+      const { content, ...copiedElementStyles } = copiedElement.styles;
+      const updatedElementStyles = {
+        ...copiedElementStyles,
+        ...textStyles[0],
+        ...colors[0].color,
+        animation: oldAnimationToDelete,
+        border: copiedElementStyles.border || null,
+      };
+      handleApplyStyle(updatedElementStyles);
```
maxyinger: I wonder if we could extract a lot of this out into a pure function. some util that takes a text element and some animations and returns `updatedElementStyles`.

packages/story-editor/src/app/rightClickMenu/hooks/useCopyPasteActions.js
```diff
@@ -0,0 +1,238 @@
...
+        properties: (currentProperties) =>
+          updateProperties(
+            currentProperties,
+            {
+              ...copiedElement.styles,
+              animation: oldAnimationToDelete,
+            },
+            /* commitValues */ true
+          ),
+      });
```
maxyinger: if we were to create an `updateSelectedElements(...)` action, we could use that here

packages/story-editor/src/app/rightClickMenu/hooks/useCopyPasteActions.js
```diff
@@ -0,0 +1,238 @@
...
+        setCopiedElement(oldStyles);
+
+        trackEvent('context_menu_action', {
+          name: 'undo_copy_styles',
+          element: selectedElement?.type,
+          isBackground: selectedElement?.isBackground,
+        });
+      },
+      actionHelpText: UNDO_HELP_TEXT,
+    });
```
maxyinger: What are your thoughts on actually moving `copiedElement` to the story provider?

The benefits I see here:
- can create a new story action which doesn't need payload `copySelectedElementStyles()`
- can just leverage our history provider undo action for the snackbar `onAction` here
- `cntrl+z` which is normally associated with undo would now undo copying of element styles and have parity with the possible snackbar action

packages/story-editor/src/app/rightClickMenu/hooks/useCopyPasteActions.js
```diff
@@ -0,0 +1,238 @@
...
+    selectedElement,
+    selectedElementAnimations,
+    showSnackbar,
+    updateElementsById,
+  ]);
+
+  return {
+    copiedElementType: copiedElement?.type,
+    onCopyStyles: handleCopyStyles,
+    onPasteStyles: handlePasteStyles,
```
maxyinger: I hate critiquing naming conventions, but just gonna drop a small not on my mental model of `on` vs `handle`

I'd be fine with these just being `handle...` vs `on...`

My reasoning for this is that `on..` is normally associated as an event that you can tap into, `handle...` is normally associated with the action that taps into that event. 

This hook is creating actions that handle how the copy or paste function, not exposing when copy and paste fire

packages/story-editor/src/app/rightClickMenu/hooks/useElementActions.js
```diff
@@ -0,0 +1,213 @@
...
+    setBackgroundElement,
+    updateElementsById,
+  } = useStory(({ state, actions }) => ({
+    clearBackgroundElement: actions.clearBackgroundElement,
+    duplicateElementsById: actions.duplicateElementsById,
+    selectedElements: state.selectedElements,
+    setBackgroundElement: actions.setBackgroundElement,
+    updateElementsById: actions.updateElementsById,
+  }));
+  const { showSnackbar } = useSnackbar();
```
maxyinger:
```
const showSnackbar = useSnackbar(value => value.showSnackbar);
```

maxyinger (approved):
I went a little more in depth in the `useCopyAndPaste` actions. The only thing I'd recommend is fixing the perf like I outlined there. Beyond that, there's some tactics you can apply to the other hooks as you please, but also don't need to go over board.

At the end of the day everything already works here as is.

The last area I'd have you look at, even tho it's not part of your diff is right here:
https://github.com/GoogleForCreators/web-stories-wp/blob/cd6886589628f0e2813d9b533db7b40850f78684/packages/story-editor/src/app/rightClickMenu/useLayerSelect.js#L89-L97
I'd just update it to this:
```
  const subMenuItems = useMemo(() => {
    if (!isMenuOpen) {
      return [];
    }
    const intersectingElements = getIntersectingElements();
    // If the only intersecting element is the selected element, don't display the menu.
    if (
      intersectingElements.length === 1 &&
      intersectingElements[0].id === selectedElements[0].id
    ) {
      return [];
    }
    ...
```
This just prevents the expensive `getIntersectingElements()` action from running when we don't need it. Doing that along with what I outlined in `useCopyAndPaste` actions got the render time from around 10ms to <1ms when updating selection on the 50 element tester story without opening the right click menu:
Screen Shot 2022-02-01 at 2 46 21 PM

packages/story-editor/src/app/rightClickMenu/provider.js
maxyinger: 💭 A lot of these `menuItems` mirror composition in jsx...

Instead of using switch statements and all this object composition in the provider, could we just achieve all this composition in the markup? so in `components > canvas > RightClickMenu.js` it might just look something like this:
```
const menuType = deriveMenuType(selectedElements);
const menu = useMemo(() => {
  switch(menuType) {
    case MENU_TYPES.MULTI_ELEMENT:
      return <MultiElementMenu />;
    case MENU_TYPES.SHAPE:
      return <ShapeMenu />
    ...
  }
}, [menuType]);
...
render createPortal(
  ...
  {menu}
  ...
);
```
then we can compose the menus through jsx in the individual menus:
```
// ShapeMenu.js
function ShapeMenu() {
  const presetActions = usePresetActions();
  ...
  return (
    <>
      <ContextMenuComponents.MenuItem 
        label={RIGHT_CLICK_MENU_LABELS.DUPLICATE_ELEMENTS(
          selectedElements.length
        )}
        onClick={elementActions.onDuplicateSelectedElements},
        separator="bottom"
      />
      ...
      <LayerSubMenu />
    </>
  );
}
```

I think this would have the benefit of:
- not needing to create every action and every menu item on every render
- drastically simplifying the right click menu provider
- remove some needless abstraction
- leaving the composition for the part of react that's really good at composition, the jsx

samwhale:
@littlemilkstudio I think your critique of the copy/paste hook are spot on. I know this ticket was meant to refactor all of this at once, but I think we should split that out into a new ticket.

This ticket can focus on making sure the structure is there. Then in a smaller pr we can tackle that.

packages/story-editor/src/app/rightClickMenu/provider.js
samwhale: I like this idea a lot, Max

packages/story-editor/src/app/rightClickMenu/hooks/useCopyPasteActions.js
samwhale: Love this idea. Going to add it to the follow up ticket

packages/story-editor/src/app/rightClickMenu/hooks/useCopyPasteActions.js
samwhale: This comment + exploration is amazing. Thank you for such a detailed dive. I'm going to focus on this in the following ticket. This will help avoid scope creep and make it easier to review those changes.

packages/story-editor/src/app/story/useStoryReducer/reducers/copyElementById.js
```diff
@@ -0,0 +1,86 @@
...
+  'font',
+  'fontSize',
+  'lineHeight',
+  'lockAspectRatio',
+  'opacity',
+  'overlay',
+  'padding',
+  'rotationAngle',
+  'textAlign',
+];
```
samwhale, line 38: 💭  The old copy/paste functionality depended on the `copyableAttributes` objects that were in each element folder. That was confusing, and different for each element type because each element has different properties that can be updated

I removed the old confusing code and centralized all properties that may be copy and pasted here. If an element has any of these properties, they may be copied and pasted to a different element of the same type.

packages/story-editor/src/components/canvas/karma/rightClickMenu.karma.js
```diff
@@ -1031,20 +1072,22 @@ describe('Right Click Menu integration', () => {
...
-        ...copyableTextProperties,
+        ...ATTRIBUTES_TO_COPY,
         'content',
       ]);
       expect(content).toBe(
         '<span style="color: #ff0110">Another Text Element</span>'
       );
       expect(copiedProperties).toEqual(pastedProperties);
+
+      // TODO: Make sure bounding box height changes to correct size
```
samwhale: Reminder to self: this isn't fixed and needs to be fixed in a later ticket

maxyinger:
> @littlemilkstudio I think your critique of the copy/paste hook are spot on. I know this ticket was meant to refactor all of this at once, but I think we should split that out into a new ticket.
> 
> This ticket can focus on making sure the structure is there. Then in a smaller pr we can tackle that.

Completely agree. yea these most recent changes definitely set us up for some fun stuff. good call

packages/story-editor/src/app/rightClickMenu/hooks/useCopyPasteActions.js
samwhale: Added this to the story provider state! There's more to cleanup for sure but for now I think it's good enough. We can look at it again in the future copy/paste perf + cleanup ticket.

packages/story-editor/src/app/rightClickMenu/hooks/useCopyPasteActions.js
samwhale: Yeah this would be good to do in the copy/paste update ticket 👍

samwhale:
Added the copy/paste update ticket here: #10459

packages/story-editor/src/app/rightClickMenu/hooks/useCopyPasteActions.js
```diff
@@ -0,0 +1,236 @@
...
+import { trackEvent } from '@googleforcreators/tracking';
+import { v4 as uuidv4 } from 'uuid';
+/**
+ * Internal dependencies
+ */
+import { PRESET_TYPES } from '../../../constants';
+import updateProperties from '../../../components/inspector/design/updateProperties';
+import useApplyStyle from '../../../components/panels/design/textStyle/stylePresets/useApplyStyle';
+import getUpdatedSizeAndPosition from '../../../utils/getUpdatedSizeAndPosition';
+import { getTextPresets } from '../../../utils/presetUtils';
```
BrittanyIRL, line 31: could do?
```suggestion
import { getTextPresets, getUpdatedSizeAndPosition } from '../../../utils';
```

packages/story-editor/src/app/rightClickMenu/menus/foregroundMediaMenu.js
```diff
@@ -0,0 +1,214 @@
...
+            parentMenuRef={parentMenuRef}
+            subMenuRef={subMenuRef}
+            isSubMenuOpen={isSubMenuOpen}
+            {...subMenuTriggerProps}
+          />
+          <SubMenuContainer
+            ref={subMenuRef}
+            position={{
+              x: 210,
+              y: 0,
```
BrittanyIRL: Could these be constants somewhere to name them so that the position purpose is easier to understand when you read through?

packages/story-editor/src/app/rightClickMenu/menus/shared.js
```diff
@@ -0,0 +1,38 @@
...
+ * External dependencies
+ */
+import { __ } from '@googleforcreators/i18n';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+export const SubMenuContainer = styled.div`
+  position: absolute;
+  top: ${({ position }) => position?.y ?? 0}px;
+  left: ${({ position }) => position?.x ?? 0}px;
```
BrittanyIRL, line 27: nit: Don't forget propTypes here 

SubMenuContainer.propTypes = {
position: PropTypes.shape({ x: PropTypes.number, y: PropTypes.number })
}

packages/story-editor/src/app/rightClickMenu/reducer.js
```diff
@@ -15,18 +15,12 @@
...
 
 export const ACTION_TYPES = {
-  COPY_ELEMENT_STYLES: 'COPY_ELEMENT_STYLES',
   OPEN_MENU: 'OPEN_MENU',
   CLOSE_MENU: 'CLOSE_MENU',
   RESET: 'RESET',
 };
 
 export const DEFAULT_RIGHT_CLICK_MENU_STATE = {
-  copiedElement: {
```
BrittanyIRL, line 25: i love that this reducer is now just for the menu's state.

packages/story-editor/src/app/story/storyProvider.js
```diff
@@ -121,7 +128,7 @@ function StoryProvider({ storyId, initialEdits, children }) {
     };
   }, [currentPage, selection]);
 
-  // This effect loads and initialises the story on first load (when there's no pages).
+  // This effect loads and initializes the story on first load (when there's no pages).
```
BrittanyIRL: 😂  classic British vs American english 🇬🇧  🇺🇸

packages/story-editor/src/app/story/useStoryReducer/actions.js
```diff
@@ -219,9 +219,18 @@ const toggleLayer =
       payload: { metaKey, shiftKey, elementId },
     });
 
+const copyElementById =
```
BrittanyIRL, line 222: makes sense here 🤝

packages/story-editor/src/components/canvas/karma/rightClickMenu.karma.js
BrittanyIRL: Will you make a follow up ticket and put the issue number here with the todo?

BrittanyIRL (approved):
Despite being a chonky PR this reads so well. It really feels like this will allow the menus to be further updated much easier. The hooks were a great idea and I love having different menus already set up. I think, given how complex these got and that we're continuing to add complexity to them, having a 1 stop shop for creating all the context menus doesn't stand the weight of it all. 

Big fan of separating the updates for performance gains too. 

Testing this, the only thing I found that would be good to handle is when I've selected multiple elements and right click I'm getting `Duplicate element` as the menu item rather than `elements`. 
Screen Shot 2022-02-03 at 2 01 30 PM

Otherwise just all nits, so i'm going to move this to QA since it's a lot to review for them.

packages/story-editor/src/app/rightClickMenu/hooks/useCopyPasteActions.js
samwhale: There's no index file there 😕 . Maybe we just leave as-is for now and come back later?

packages/story-editor/src/app/rightClickMenu/menus/foregroundMediaMenu.js
samwhale: Good catch - these shouldn't be hard coded

packages/story-editor/src/app/story/storyProvider.js
samwhale: OOps! My grammar checker just flagged it, but looks like it was correct before. i'll revert hehe

packages/story-editor/src/components/canvas/karma/rightClickMenu.karma.js
samwhale: Turns out this works 👍  removed it.

swissspidy:
FYI there was a failing JS unit test due to some circular imports that tripped up Jest.

Fixed in 3aaa2a5fd05586200518b43674fdaa20188dc558.

samwhale:
Oh shoot, thanks for fixing that!

PR #10369, opened 2022-01-25
E2E Tests: Optimize workflow for parallel runs

### LINKED ISSUE #10333: E2E Tests: Move percy uploads to separate job
## Task Description

Now that our e2e tests are running in a parallel job, it means that we're uploading Percy snapshots 2 times instead of once
Screenshot 2022-01-22 at 17 15 20

To fix this, we need to somehow move the snapshot upload step to its own job that runs after the matrix job, passing around the snapshot files using `actions/upload-artifact`

That means moving away from using `@percy/cli-exec` as-is and probably starting the snapshot server before the test run and stopping it & uploading the snapshots afterwards

```
percy.start()

// run tests in matrix, ensuring that snapshots are not uploaded directly

percy.stop()
```

swissspidy:
## Context

Now that our e2e tests are running in a parallel job, it means that we're uploading Percy snapshots 2 times instead of once

To fix this, we need to somehow move the snapshot upload step to its own job that runs after the matrix job

## Summary

Luckily, Percy has documented best practices for this, so it was easy to configure Percy correctly for our setup.

In addition to optimizing the Percy step, this change also moves the build out of the test step so that we only build the plugin once.

## Relevant Technical Choices

Followed recommendations from https://docs.percy.io/docs/parallel-test-suites#parallel-suites-split-across-different-machines

## User-facing changes

N/A

## Testing Instructions

1.

## Checklist

---

Fixes #10333

### DIFF
--- a/.github/workflows/tests-e2e.yml
+++ b/.github/workflows/tests-e2e.yml
@@ -215,6 +215,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - name: Setup Node
+        uses: actions/setup-node@v2.5.1
+        with:
+          node-version-file: '.nvmrc'
+          cache: npm
+
       - name: Install dependencies
         run: npm install @percy/cli
 
--- a/.github/workflows/tests-e2e.yml
+++ b/.github/workflows/tests-e2e.yml
@@ -62,11 +62,59 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  build:
+    name: Build plugin
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Setup Node
+        uses: actions/setup-node@v2.5.1
+        with:
+          node-version-file: '.nvmrc'
+          cache: npm
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.0'
+          coverage: none
+          tools: composer
+
+      - name: Install dependencies
+        run: npm ci
+        env:
+          PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: true
+
+      - name: Install PHP dependencies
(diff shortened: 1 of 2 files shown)


swissspidy:
This now works as expected, see Percy result here: https://percy.io/google/web-stories-wp-e2e/builds/15266857

Only wondering how to best cache the Percy browser so that the tests don't take 20 minutes to run with snapshots 🤔 

Apart from that this PR is ready

swissspidy:
Now the run times are around ~10min again:

Screenshot 2022-01-25 at 18 46 12

.github/workflows/tests-e2e.yml
```diff
@@ -161,3 +204,22 @@ jobs:
...
           COMPOSE_INTERACTIVE_NO_CLI: true
+
+  percy:
+    name: Percy
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    if: always()
+    needs: [e2e, nonce]
+    steps:
+      - name: Checkout
```
swissspidy, line 205: TODO: Run `actions/setup-node` afterwards

.github/workflows/tests-e2e.yml
```diff
@@ -62,11 +62,68 @@ concurrency:
...
+          path: build/web-stories
+
+  nonce:
+    name: Percy Nonce
+    runs-on: ubuntu-latest
+    outputs:
+      result: ${{ steps.nonce.outputs.result }}
+    steps:
+      - id: nonce
+        run: echo "::set-output name=result::${{ github.run_id }}-$(date +%s)"
```
swissspidy, line 119: See https://docs.percy.io/docs/parallel-test-suites#setting-percy_parallel_nonce for context

swissspidy:
Some snapshots look super broken, as if some CSS is missing.

But that appears to be an issue on `main` and other PRs as well, see e.g. https://percy.io/google/web-stories-wp-e2e/builds/15288296 / #10390

Maybe related to WP 5.9? 🤔

PR #10370, opened 2022-01-25
UI Simplification: Consider aspect ratio lock when resizing

### LINKED ISSUE #10106: UI Simplification: Unlock corners of rectangles and images when resizing
## Feature Description
Currently, the ratio of rectangle shapes and images are locked when resizing from the corners. 

Note: this is for single selection only.
- [ ] For single selection: use the aspect ratio lock based on what's set in the design panel. E.g. if the ratio is locked there, also lock it when resizing the element (no matter from where)
- [ ] Unlock the ratio by default for images and rectangles
- [ ] For text element, continue to only allow resizing from the corners and use the "Auto" height for unlocked state.
- [ ] When "Shift" is held, the ratio is locked, no matter the setting in the design panel

miina:
## Summary
Previously, the aspect ratio lock of the element (configured in the design panel) and resizing the element on canvas were not related.

With this PR, the configured lock will also influence resizing the element. If the ratio is locked, it's also locked when resizing, when unlocked, it's unlocked (even from the corners).

New images and rectangles will now have unlocked aspect ratio by default.
Holding down `Shift` key will force lock, no matter the design panel setting.

## To-do

- [x] Update tests

## User-facing changes
Text resize: unlocked and then locked (with `Shift`)

text-resize

Image resize: unlocked and then locked (with `Shift`)
image-resize

## Testing Instructions

1. Verify that a newly added rectangle or image has unlocked aspect ratio
2. Verify that resizing any element with unlocked aspect ratio (from corners as well) indeed resizes it without keeping ratio (note: stickers can't have unlocked ratio).
3. Verify that resizing from any side with locked ratio keeps the ratio locked.
4. Verify that when text ratio is unlocked, it can only be resized horizontally.
5. Verify that holding down "Shift" forces resizing with keeping ratio, independent of the design panel setting.

No

No

No

## Checklist

---

Fixes #10106


packages/story-editor/src/components/library/panes/shapes/shapePreview.js
```diff
@@ -138,6 +139,8 @@ function ShapePreview({ mask, isPreview, index }) {
       mask: {
         type: mask.type,
       },
+      // Rectangles have unlocked aspect ratio by default.
+      lockAspectRatio: MaskTypes.RECTANGLE !== mask.type,
```
barklund, line 143: Any particular reason why only rectangles have this?

barklund (approved):
This works great. Ship it! 🚢 

I feel like <kbd>shift</kbd> should toggle the lock to do the opposite (so if it's locked, holding <kbd>shift</kbd> would unlock). But that's a UX decision, and if they don't want that, fine with me. It just seems natural, somehow?

packages/story-editor/src/components/library/panes/shapes/shapePreview.js
miina: Specific UX request. I suppose it's because some other shapes might start looking odd when not locked but the rectangle will likely be changed anyway to a non-square. Just assuming here of course.

PR #10383, opened 2022-01-25
Editor: Stop crashing when pulling in video element

### LINKED ISSUE #10373: Dragging video into canvas crashes editor
## Bug Description

Clicking and dragging a video into the canvas crashes the editor. This happens with videos in the media panel and the 3p media panel.

## Expected Behaviour

Clicking and dragging videos into the canvas should not crash the editor

## Steps to Reproduce

1. Open the media or 3p media tab in the library
2. Click and drag a video into the canvas
3. Editor crashes

## Screenshots

image

https://user-images.githubusercontent.com/22185279/151026704-2fc231a6-c9e6-426a-a30a-b43b37ba2ca3.mp4

## Additional Context

Stack trace seems to point to the popover component:
image

- Plugin Version: 1.17.0-alpha.0+e1a8832
- WordPress Version: 5.8.3
- Operating System: MacOSX Big Sur 11.6.2
- Browser: Chrome

samwhale:
## Context

Discovered during bug bash 1.17

## Summary

Prevents `offset` from being set to `undefined`. Editor won't crash when [destructuring off of `offset`](https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/story-editor/src/components/popup/index.js#L42) now.

## Relevant Technical Choices

Possibly related to changes in #9944 with the `isMounted` changes. Putting up a quick fix to unblock everybody but digging deeper to figure out the intended behavior here.

The popup has a `Container` element that [destructures some values off of the `$offset` property](https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/story-editor/src/components/popup/index.js#L42). By making sure `$offset` is always an object, the editor won't crash.

Another approach would be to avoid destructuring and instead use optional chaining like `$offset?.x`.

Relevant code:

https://github.com/GoogleForCreators/web-stories-wp/blob/f9530da9116b559494381dbc28abda834ea0dce8/packages/story-editor/src/components/popup/index.js#L92-L110

If y'all have any context about this, let me know. Trying to figure out if this function should be setting the height at all if `anchor` is not defined.

## To-do

n/a

## User-facing changes

Doesn't crash

## Testing Instructions

1. Click and drag a video from the media or 3p media sections in the library
2. Editor should not crash

## Checklist

---

Fixes #10373

### DIFF
--- a/packages/story-editor/src/components/popup/index.js
+++ b/packages/story-editor/src/components/popup/index.js
@@ -99,10 +99,11 @@ function Popup({
       if (evt?.target?.nodeType && popup.current?.contains(evt.target)) {
         return;
       }
+
       setPopupState({
-        offset:
-          anchor?.current &&
-          getOffset(placement, spacing, anchor, dock, popup, isRTL, topOffset),
+        offset: anchor?.current
+          ? getOffset(placement, spacing, anchor, dock, popup, isRTL, topOffset)
+          : {},
         height: popup.current?.getBoundingClientRect()?.height,
       });
     },


packages/story-editor/src/components/popup/index.js
```diff
@@ -99,10 +99,11 @@ function Popup({
...
         return;
       }
+
       setPopupState({
-        offset:
-          anchor?.current &&
-          getOffset(placement, spacing, anchor, dock, popup, isRTL, topOffset),
+        offset: anchor?.current
+          ? getOffset(placement, spacing, anchor, dock, popup, isRTL, topOffset)
+          : {},
```
samwhale, line 106: https://github.com/GoogleForCreators/web-stories-wp/blob/f9530da9116b559494381dbc28abda834ea0dce8/packages/story-editor/src/components/popup/index.js#L92-L110

If `anchor.current` is `undefined` - should this function even set the popup state?

maxyinger (approved):

BrittanyIRL (approved):
not seeing the crash happen

BrookeGraham (approved):
>The popup has a Container element that destructures some values off of the $offset property. By making sure $offset is always an object, the editor won't crash.

I see. So Popup first sets a [fallback {} for offset](https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/story-editor/src/components/popup/index.js#L82) `const { isRTL, styleConstants: { topOffset } = {} } = useConfig();` but then if there isn't an anchor, `offset` gets set to `undefined`. So the destructuring in `Container` blows up.  Great find! 

🤔 Perhaps a reason anchor _could_ be undefined would be like if a popup wasn't tied to a specific element. Like if a popup's placement was tied to just the top right corner or something. Not sure. Just rambling here.

packages/story-editor/src/components/popup/index.js
BrittanyIRL: Yes I think we do want it to set back to an empty object to clear the popupState.

swissspidy:
Where did this regress?

Can we add tests so it doesn't happen again?

packages/story-editor/src/components/popup/index.js
miina: Looks like the same is used in the popup of the `design-system`, maybe change it there as well?

Once #10089 gets merged, only the `design-system` popup will be used.

samwhale:
> Where did this regress?

Tested commits individually - looks like #9944 introduced this bug. e00e50d doesn't seem to have the bug but 293a3ad9cc569b601c5cbfbf2d4c88bde31b7666 does.

> Can we add tests so it doesn't happen again?

#10197 is de-duplicating the popup so this one won't exist soon. While I don't think that adding unit testing will be beneficial, I think adding a simple karma test of clicking and dragging a video into the canvas may work. Not sure if one exists already.

samwhale:
It seems like the '`anchor.current -> undefined` when the video is placed' has existed for a while, but now the styled component is rendering when `offset` is `undefined`.

When letting go of the mouse after clicking and dragging a video element, the video is added to the canvas. The `anchor` ref is `undefined` on the first render of the `Popover`.

After that the ref gets registered to the wrapping element:
clickndrag

swissspidy (approved):

packages/story-editor/src/components/popup/index.js
miina: Hm, okay, looks like the PR got merged without this already. Another PR? 😛 Not sure if the `design-system` Popup is causing any crashes currently, I have seen the right-click menu crashing once but haven't been able to repro.

swissspidy: Sorry about that, I missed this thread. A new PR would be good, or just push changes to #10197 to bring it over the finish line

samwhale:
Looks like the fix was merged (p0) - but I still think it's worth talking about the intention of what's happening here so we can make sure it doesn't regress again and potentially find a way to test it reliably since it doesn't happen consistently

@miina @barklund wanted to get your input here. I think y'all have the most context

When the `anchor` ref is not set - do we want the `offset` in the popover state to be set?

miina:
> When the anchor ref is not set - do we want the offset in the popover state to be set?

I'd actually think that if anchor ref is not set, it shouldn't render, and maybe `setPopupState(null)` should be called instead? As far as I understand, the popup should never be rendered without `anchor.ref`. 🤔 

Also wondering about the `isMounted` ref (for future purposes) -- @swissspidy did you try there if checking for `popup.current` also worked, or do you think it's generally better to use `isMounted` everywhere for consistency regardless of other existing refs?

swissspidy:
I haven't, but it sounds reasonable. 

----

Admittedly, `isMounted` is [not the best solution](https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html), so if we can avoid it that's great. I suppose ideally we'd use AbortController wherever we currently have `isMounted` checks around aysnc callbacks...

Issue #10251, opened 2022-01-18
Add workflow for releasing npm packages


swissspidy:
## Task Description

Add GitHub workflow for publishing packages regularly

PR #10386, opened 2022-01-25
Build packages using Rollup

swissspidy:
## Context

Before we can publish our packages on npm we need a build step for them, ideally to produce both CommonJS and ESM versions.

## Summary

This is an attempt at this by leveraging Rollup and several Rollup plugins that fit our needs.

Building packages creates `dist` and `dist-module` folders that are then used for distribution.

The `src` folder is entirely ignored for publishing on npm.

Basic usage:

```bash
# Build all packages using Rollup
$ npm run bundle

# Build only the migration package
$ npm run bundle -- --configPackages=migration

# Build only the module version of all packages
$ npm run bundle -- --configEntries=es

# Build only the CommonJS version of the migration and text-sets packages
$ npm run bundle -- --configEntries=cjs --configPackages=migration,text-sets
```

## Relevant Technical Choices

Decided to try Rollup because we're already using it in the project and it was relatively easy to get off the ground.

Changed some imports to avoid circular dependencies, though this still needs more work.

Changed some imports to avoid mixing default imports and named exports. Quote:

> It is bad practice to mix default and named exports in the same module, though it is allowed by the specification.

Used custom resolvers for webpack and Jest so they continue to look for packages in the `src` folders instead of `dist`.

## To-do

* [x] Testing
* [x] Fix inline web worker usage in `story-editor`
* [x] Fix `import/no-unresolved` false positives
* [ ] Add sourcemaps?
* [ ] Fix duplicate React version in `react-photo-gallery`
* [ ] Fix some circular dependencies (see below)

```
(!) Circular dependencies
packages/story-editor/src/masks/index.js -> packages/story-editor/src/elements/index.js -> packages/story-editor/src/elements/elementTypes.js -> packages/story-editor/src/elements/text/index.js -> packages/story-editor/src/elements/text/display.js -> packages/story-editor/src/elements/shared/index.js -> packages/story-editor/src/utils/elementBorder.js -> packages/story-editor/src/masks/index.js

packages/story-editor/src/app/story/index.js -> packages/story-editor/src/app/story/useStoryReducer/reducers/index.js -> packages/story-editor/src/app/story/useStoryReducer/reducers/combineElements.js -> packages/story-editor/src/masks/index.js -> packages/story-editor/src/elements/index.js -> packages/story-editor/src/elements/elementTypes.js -> packages/story-editor/src/elements/text/index.js -> packages/story-editor/src/elements/text/edit.js -> packages/story-editor/src/app/story/index.js

packages/story-editor/src/app/story/index.js -> packages/story-editor/src/app/story/useStoryReducer/reducers/index.js -> packages/story-editor/src/app/story/useStoryReducer/reducers/combineElements.js -> packages/story-editor/src/masks/index.js -> packages/story-editor/src/elements/index.js -> packages/story-editor/src/elements/elementTypes.js -> packages/story-editor/src/elements/text/index.js -> packages/story-editor/src/elements/text/frame.js -> packages/story-editor/src/app/story/index.js

...and 48 more

```

## User-facing changes

None

## Testing Instructions

1. Verify that all status checks pass
2. Run `npm run bundle` to build the packages yourself

Not that I could think of

No

No

## Checklist

---

See #10251


packages/dashboard/src/index.js
```diff
@@ -27,5 +23,3 @@ export * from './app/views/shared';
 export { GlobalStyle as DashboardGlobalStyle } from './theme';
 export { default as DashboardKeyboardOnlyOutline } from './utils/keyboardOnlyOutline';
 export { default as InterfaceSkeleton } from './components/interfaceSkeleton';
-
-export default Dashboard;
```
mohdsayed, line 31: Curious to know, what issue we had with the default export?

mohdsayed: Okay I see

https://rollupjs.org/guide/en/#exporting

> It is bad practice to mix default and named exports in the same module, though it is allowed by the specification.

rollup.config.js
```diff
@@ -0,0 +1,126 @@
...
+ * Internal dependencies
+ */
+
+const plugins = [
+  resolve({
+    preferBuiltins: true,
+    dedupe: [],
+  }),
+  babel({
+    babelHelpers: 'inline',
```
mohdsayed, line 46: Trying to learn why `inline` is used as an option for [babelHelpers](https://github.com/rollup/plugins/tree/master/packages/babel#babelhelpers) here

swissspidy: I used `runtime` initially as it was the recommendation, but I got some warnings when building.

mohdsayed:
```bash
(!) Circular dependencies
packages/story-editor/src/masks/index.js -> packages/story-editor/src/elements/index.js -> packages/story-editor/src/elements/elementTypes.js -> packages/story-editor/src/elements/text/index.js -> packages/story-editor/src/elements/text/display.js -> packages/story-editor/src/elements/shared/index.js -> packages/story-editor/src/utils/elementBorder.js -> packages/story-editor/src/masks/index.js
packages/story-editor/src/app/story/index.js -> packages/story-editor/src/app/story/useStoryReducer/reducers/index.js -> packages/story-editor/src/app/story/useStoryReducer/reducers/combineElements.js -> packages/story-editor/src/masks/index.js -> packages/story-editor/src/elements/index.js -> packages/story-editor/src/elements/elementTypes.js -> packages/story-editor/src/elements/text/index.js -> packages/story-editor/src/elements/text/edit.js -> packages/story-editor/src/app/story/index.js
packages/story-editor/src/app/story/index.js -> packages/story-editor/src/app/story/useStoryReducer/reducers/index.js -> packages/story-editor/src/app/story/useStoryReducer/reducers/combineElements.js -> packages/story-editor/src/masks/index.js -> packages/story-editor/src/elements/index.js -> packages/story-editor/src/elements/elementTypes.js -> packages/story-editor/src/elements/text/index.js -> packages/story-editor/src/elements/text/frame.js -> packages/story-editor/src/app/story/index.js
...and 48 more
```

51 indirect circular dependencies are a lot to resolve. 😵‍💫

mohdsayed:
I think we may ignore these circular dependency warning for now by using `onwarn` and deal with them later separately.

```js
onwarn(warning, warn) {
  if (warning.code !== 'CIRCULAR_DEPENDENCY') {
    warn(warning);
  }
}
```

swissspidy:
They don't cause the build to error though, so not a big deal to keep them visible and remind os of this technical debt :)

mohdsayed:
> They don't cause the build to error though, so not a big deal to keep them visible and remind os of this technical debt :)

Yes, maybe a lot of them will be resolved automatically when we break down the `story-editor` in more packages in the upcoming package organization PRs.

mohdsayed:
Rollup doesn't seem to have support for Web Workers, see https://github.com/rollup/rollup/issues/3842 

Tried [rollup-plugin-web-worker-loader](https://www.npmjs.com/package/rollup-plugin-web-worker-loader) npm package however it breaks with the following error

```bash
[!] (plugin rollup-plugin-web-worker-loader) TypeError: Could not load rollup-plugin-worker-loader::module:/Users/sayedtaqui/Local Sites/ws/app/public/wp-content/plugins/web-stories-wp/packages/story-editor/src/utils/generateBlurhash.worker.js (imported by packages/story-editor/src/utils/getBlurHashFromImage.js): The "path" argument must be of type string. Received undefined
TypeError [PLUGIN_ERROR]: Could not load rollup-plugin-worker-loader::module:/Users/sayedtaqui/Local Sites/ws/app/public/wp-content/plugins/web-stories-wp/packages/story-editor/src/utils/generateBlurhash.worker.js (imported by packages/story-editor/src/utils/getBlurHashFromImage.js): The "path" argument must be of type string. Received undefined
    at new NodeError (node:internal/errors:371:5)
    at validateString (node:internal/validators:119:11)
    at Object.dirname (node:path:1276:5)
    at Object.write (/Users/sayedtaqui/Local Sites/ws/app/public/wp-content/plugins/web-stories-wp/node_modules/@rollup/plugin-url/dist/index.js:82:85)
    at /Users/sayedtaqui/Local Sites/ws/app/public/wp-content/plugins/web-stories-wp/node_modules/rollup/dist/shared/rollup.js:22830:37
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)

[!] Error: unfinished hook action(s) on exit:
````

Will try again tomorrow.

**Comment by mohdsayed** [comment]
@swissspidy Can you explain this one a bit from the to-do list?

> Fix import/no-unresolved false positives

**Comment by swissspidy** [comment]
> Rollup doesn't seem to have support for Web Workers, see [rollup/rollup#3842](https://github.com/rollup/rollup/issues/3842)
> 
> Tried [rollup-plugin-web-worker-loader](https://www.npmjs.com/package/rollup-plugin-web-worker-loader) npm package however it breaks with the following error

I'm inclined to just ditch our current web worker logic completely. Not a huge fan of the webpack `worker-loader` anyway.

We could take a very simple approach by manually turning it into an inline/embedded web worker (see examples at https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers#embedded_workers) using a blob — which is basically what `worker-loader` is doing anyway — and what `rollup-plugin-web-worker-loader` should be supporting as well.

So something like this:

```js

function generateBlurHash() {
// ...
}

const blob = new Blob(['('+generateBlurHash.toString()+')()'], {type: 'text/javascript'})
const workerUrl = URL.createObjectURL(blob);

new Worker(workerUrl)
```

At least for the short term this should do the trick.

Suggestions welcome!

> @swissspidy Can you explain this one a bit from the to-do list?
> 
> > Fix import/no-unresolved false positives

Sure.

Basically right now we have tons of ESLint warnings here on this PR.

They are caused by all the changes I did to the `package.json` files for all the packages (specifically, changing the `main` field).

They come from [eslint-plugin-import](https://github.com/import-js/eslint-plugin-import), specifically the [no-unresolved](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-unresolved.md) rule

It uses [Node-style module resolution](https://www.npmjs.com/package/eslint-import-resolver-node) by default, which has some open issues about this

https://github.com/import-js/eslint-plugin-import/issues/1868
https://github.com/import-js/eslint-plugin-import/issues/1810

Perhaps using eslint-import-resolver-webpack fixes this?

Otherwise I suppose we can just disable this rule until the above issues get fixed.

mohdsayed:
Great idea of turning into an inline/embedded web worker using blob 👍 

For Eslint warnings, I tried eslint-import-resolver-webpack, but that didn't work for me.

Gutenberg also had a similar issue here https://github.com/WordPress/gutenberg/pull/31792 for which they used [eslint-import-resolver-node](https://www.npmjs.com/package/eslint-import-resolver-node), see [/.eslintrc.js#L53](https://github.com/WordPress/gutenberg/blob/71635329224c3dcc148258b7b0e855acf3b5d733/.eslintrc.js#L53)

swissspidy:
> Gutenberg also had a similar issue here [WordPress/gutenberg#31792](https://github.com/WordPress/gutenberg/pull/31792) for which they used [eslint-import-resolver-node](https://www.npmjs.com/package/eslint-import-resolver-node), see [/.eslintrc.js#L53](https://github.com/WordPress/gutenberg/blob/71635329224c3dcc148258b7b0e855acf3b5d733/.eslintrc.js#L53)

My understanding is that the node resolver is the default 🤔 

It looks like Gutenberg actually wrote their own resolver, which is what this config adds. See https://github.com/WordPress/gutenberg/blob/71635329224c3dcc148258b7b0e855acf3b5d733/tools/eslint/import-resolver.js

Just resolving `@googleforcreators/` packages to `packages/<package>/src` like they do would indeed solve this issue.

We can definitely give that a try and create our own resolver as well 👍

mohdsayed:
@swissspidy Due to the path issue of the resolver file, it's not working with `.eslintrc`, but it does work when we change the file type to `eslintrc.js` and use `require.resolve` to resolve the file path.

swissspidy:
We can rename the file, no problem.

Alternatively, what might work is moving this custom resolver into a dedicated `@googleforcreators/eslint-import-resolver` package and setting `import/resolver: '@googleforcreators/eslint-import-resolver`

This we we could easily add documentation, tests, etc. for it.

WDYT?

mohdsayed:
> Alternatively, what might work is moving this custom resolver into a dedicated @googleforcreators/eslint-import-resolver package and setting import/resolver: '@googleforcreators/eslint-import-resolver

So the only use of this package would be to contain this resolver file ( and its related tests etc ) ? Doesn't that look more suitable for a file which we can be kept in some folder ( named like tools or something )? Or do you think something else may also be added to this package in the future?

swissspidy:
Yeah it's probably a bit overkill :-)

Was thinking that it might be nice to have things a bit more organized this way. It could even be made configurable so that for example Gutenberg could use this package too.


But let's just us a file in some folder for now, we can always change it in the future.

packages/i18n/package.json
```diff
@@ -26,7 +26,9 @@
   "exports": {
     "default": "./src/index.js"
   },
-  "main": "./src/index.js",
+  "main": "./dist/index.js",
+  "module": "./dist-module/index.js",
+  "source": "./src/index.js",
```
mohdsayed: Was [source](https://parcel2-docs.vercel.app/configuration/package-json/#source) added for parcel? I wonder why Gutenberg doesn't have source in their `package.json`

swissspidy: Yes. It's not an [official field](https://docs.npmjs.com/cli/v8/configuring-npm/package-json) so it's only relevant for Parcel, hence GB not using it.

Maybe we don't need it either, haven't tested

mohdsayed: Okay, I see it's being used in `rollup.config.js`

rollup.config.js
```diff
@@ -0,0 +1,126 @@
...
+import url from '@rollup/plugin-url';
+import json from 'rollup-plugin-json';
+import license from 'rollup-plugin-license';
+import workspacesRun from 'workspaces-run';
+
+/**
+ * Internal dependencies
+ */
+
+const plugins = [
```
mohdsayed, line 39: We also need a rollup plugin to clean up the dist folder files when bundling. 

After a few builds I started getting the following error

```bash
npm notice === Tarball Details ===
npm notice name:          @googleforcreators/story-editor
npm notice version:       0.1.202201271454
npm notice filename:      @googleforcreators/story-editor-0.1.202201271454.tgz
npm notice package size:  14.9 MB
npm notice unpacked size: 46.9 MB
npm notice shasum:        fb74cfc732ad86c4b018be5e1224cd46a9b7ae8e
npm notice integrity:     sha512-+0SiTYz3Q2+9R[...]bP2o7eCD5PfYQ==
npm notice total files:   1108
npm notice
npm ERR! code E413
npm ERR! 413 Payload Too Large - PUT http://localhost:4873/@googleforcreators%2fstory-editor - request entity too large

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/sayedtaqui/.npm/_logs/2022-01-27T14_54_19_044Z-debug.log
```


It's because it keeps adding new files without removing the old ones. We may use [rollup-plugin-delete](https://www.npmjs.com/package/rollup-plugin-delete) and clear `dist` and `dist-module` folders

```js
del({
    targets: ['dist/*', 'dist-module/*']
  }),
```

packages/animation/package.json
```diff
@@ -26,7 +26,9 @@
   "exports": {
     "default": "./src/index.js"
   },
-  "main": "./src/index.js",
+  "main": "./dist/index.js",
+  "module": "./dist-module/index.js",
```
mohdsayed: Maybe we need to remove `./` or else the imports would not work in another library. That's what we have in the rollup example repo
https://github.com/rollup/rollup-starter-lib/blob/88a2d55d449e2cb12191199d993c08bc07c90be2/package.json#L4-L5

and in Gutenberg 
https://github.com/WordPress/gutenberg/blob/262d4675532ead489a37bfb7957a84f9b3bd8817/packages/block-library/package.json#L24-L25

mohdsayed:
Tested it after manually merging https://github.com/GoogleForCreators/web-stories-wp/pull/10297 and

In `package.json`
1. Updated ` "module": "./dist-module/index.js"` to ` "module": "dist-module/index.js"`
2. Removed `"type": "module"`
3. Removed `"exports"` ( don't know why but it imports from `./src` directory when leaving this in `package.json` )

I think all import issues were resolved but now we have some node errors.

```bash
ERROR in ./node_modules/@googleforcreators/story-editor/dist-module/index-276f5876.js
Module not found: Error: Can't resolve './raw' in '/Users/sayedtaqui/playground/react-app/node_modules/@googleforcreators/story-editor/dist-module'
 @ ./node_modules/@googleforcreators/story-editor/dist-module/index-276f5876.js 404:624305-624330
 @ ./node_modules/@googleforcreators/story-editor/dist-module/index.js
 @ ./src/index.js

ERROR in ./node_modules/@googleforcreators/story-editor/dist-module/index-276f5876.js
Module not found: Error: Can't resolve './raw' in '/Users/sayedtaqui/playground/react-app/node_modules/@googleforcreators/story-editor/dist-module'
 @ ./node_modules/@googleforcreators/story-editor/dist-module/index-276f5876.js 136:46199-46219
 @ ./node_modules/@googleforcreators/story-editor/dist-module/index.js
 @ ./src/index.js

ERROR in ./node_modules/@googleforcreators/story-editor/dist-module/index-c3e659ac.js
Module not found: Error: Can't resolve 'fs' in '/Users/sayedtaqui/playground/react-app/node_modules/@googleforcreators/story-editor/dist-module'
 @ ./node_modules/@googleforcreators/story-editor/dist-module/index-c3e659ac.js 17:99-117 17:24780-24781
 @ ./node_modules/@googleforcreators/story-editor/dist-module/index-276f5876.js
 @ ./node_modules/@googleforcreators/story-editor/dist-module/index.js
 @ ./src/index.js
```

rollup.config.js
swissspidy: Good idea 👍

packages/animation/package.json
swissspidy: Will remove, seems to work fine

rollup.config.js
swissspidy: Will add

swissspidy:
> In `package.json`
> 
> 1. Updated ` "module": "./dist-module/index.js"` to ` "module": "dist-module/index.js"`
> 2. Removed `"type": "module"`
> 3. Removed `"exports"` ( don't know why but it imports from `./src` directory when leaving this in `package.json` )

I'm getting lots of `Unresolved dependencies` warnings/errors during the Rollup build when I do that. Are you seeing those too?

mohdsayed:
> I'm getting lots of Unresolved dependencies warnings/errors during the Rollup build when I do that. Are you seeing those too?

Not getting any `Unresolved dependencies` errors/warnings. What are those?

swissspidy:
For example

```
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
@googleforcreators/react (imported by packages/i18n/src/translateWithMarkup.js)
```

Gonna push my latest local changes in a bit, that might help

.eslintrc
```diff
@@ -450,6 +458,7 @@
         "packages/dashboard/src/karma-tests.cjs",
         "packages/story-editor/src/karma-tests.cjs",
         "packages/migration/src/**/*.js",
+        "packages/eslint-import-resolver/src/*.js",
```
swissspidy: Not sure why this doesn't have any effect 🤔

packages/eslint-import-resolver/src/index.js
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * External dependencies
+ */
+const path = require('path');
+const nodeResolver = require('eslint-import-resolver-node');
+
+exports.interfaceVersion = 2;
+
+exports.resolve = function (source, file, config) {
```
swissspidy: @sayedtaqui tried the custom package approach anyway, just to see how well it works

mohdsayed:
> > In `package.json`
> > 
> > 1. Updated ` "module": "./dist-module/index.js"` to ` "module": "dist-module/index.js"`
> > 2. Removed `"type": "module"`
> > 3. Removed `"exports"` ( don't know why but it imports from `./src` directory when leaving this in `package.json` )
> 
> I'm getting lots of `Unresolved dependencies` warnings/errors during the Rollup build when I do that. Are you seeing those too?

Its happening after we removed 

```js
 "exports": {
    "default": "./src/index.js"
  },
```

mohdsayed:
For my info, what was the use of adding `exports` in `package.json` ?

swissspidy:
IIRC it was to ensure that the imports work for as many consumers (jest, eslint, node, webpack, etc.) as possible. Because `main` and `module` were still very new, so many use `exports` instead

Used articles like https://2ality.com/2019/10/hybrid-npm-packages.html as orientation

.eslintrc
mohdsayed: Just guessing that maybe because it's `*.js` instead of `*.cjs` and `"type": "module"` is defined in `package.json` so it's considering it es module using `require` ?

package.json
```diff
@@ -156,9 +171,11 @@
     "webpack-bundle-analyzer": "^4.5.0",
     "webpack-cli": "^4.9.2",
     "webpackbar": "^5.0.2",
-    "worker-loader": "^3.0.8"
+    "worker-loader": "^3.0.8",
+    "workspaces-run": "^1.0.1"
   },
   "scripts": {
+    "bundle": "rollup",
```
mohdsayed: Shouldn't this be `rollup -c` ?

swissspidy: Haven‘t tested it again recently, but I thought Rollup automatically looks for the default config so there‘s no need to specify it here.

mohdsayed: Okay, looks like it doesn't work without it.
image

swissspidy: Oh, looks like we need to use `rollup -c`. Then it will default to `rollup.config.js`. Unexpected.

packages/jest-resolver/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@web-stories-wp/jest-resolver",
+  "description": "Custom Jest resolver for monorepos",
+  "private": true,
```
mohdsayed, line 4: Should this be `"private": true` ?

mohdsayed (approved):

packages/jest-resolver/package.json
swissspidy: Yes this does not need to published; it can stay private

PR #10388, opened 2022-01-26
UI Simplification: Library insertion icon/menu

### LINKED ISSUE #9884: Saved Template Wrong Keyboard Shortcut

### LINKED ISSUE #10109: UI Simplification: Add menu / icon for adding elements
## Feature Description
See [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=11122%3A446483) (NB -- as of 7th of Feb, the file has previous design, needs updating to the new with 2 buttons)

It was decided to use the two buttons/menus option for the element (where applicable). The changes per element:

**Local image / video:**
- We will have two buttons visible when hovering / focusing on the media element in the library: the three dotted "More" menu and also Insertion menu with `+` icon.
- When clicking on the Insertion menu button, the following options appear:
  - Insert Image (or Video, depending)
  - Add as background
- When clicking on the "More" menu button, the "Edit meta data" and "Delete from library" options appear (as currently but slightly changed names)
- Accessibility:
  - The Media element itself will not be focusable anymore.
  - When tabbing into a Media element, the Insertion menu button will be in focus first (Enter/Space will open that menu)
  - When tabbing while being focused on the Insertion button, the "More" button will be in focus (Enter/Space will open that menu)
  - When tabbing while focused on "More" button, the next focusable element after the Media Library will be in focus (as previously when tabbing while focused on the media item in the library
  - When using arrow keys while focused on either of the menu buttons, the focus will move to the Insertion menu of the next / previous etc. Media item Insertion menu button (as currently, but the Insertion menu button will get the focus)

**3rd party image / video:**
- We will have the Insertion menu visible when hovering focusing on the 3p media element in the library.
- When clicking on the Insertion menu, the following options appear:
  - Insert Image (or Video, depending)
  - Add as background
- Accessibility:
  - As in case of local media, the focus outline will now (visibly) be on the Insertion menu button when focusing instead of the whole media element. The rest of the things will stay the same since we have one button here only.
 
**Shape, stickers, Text elements, Default Page Templates:**
- We will have the Insertion button visible when hovering focusing on the 3p media element in the library.
- When clicking on the insertion button, the element will be inserted right away (or template applied) -- that's the only option for these elements so no need for opening a menu.
- Accessibility:
  - The focus outline will now (visibly) be on the insertion button when focusing instead of the whole media element. The rest of the things will stay the same as we have one button here only.

**Saved Page Templates:**
- We will have two buttons visible when hovering / focusing on the media element in the library: the Delete button in the corner (as currently), and also Insertion button with `+` icon.
- When clicking on the insertion button, the page template will be applied.
- When clicking on the Delete button, the confirmation dialogue will open (as currently)
- Accessibility:
  - The page template itself will not be focusable anymore.
  - When tabbing into a Saved Page Template, the Insertion button will be in focus first (Enter/Space will use the template)
  - When tabbing while being focused on the Insertion button, the "Delete" button will be in focus (Enter/Space will the Delete confirmation dialogue)
  - When tabbing while focused on "Delete" button, the next focusable element after the Library will be in focus (as currently when tabbing while focused on a Saved Page templates in the library)
  - When using arrow keys while focused on either of the buttons, the focus will move to the Insertion button of the next / previous etc. (depending on the arrow direction) saved Page Template Insertion button (as currently, but the Insertion button will get the focus instead of the whole template)

Todo:
- [ ] Add additional `+` icon when hovering / focusing on element in the library
- [ ] When clicking anywhere else but the icon on the element, the element is directly inserted.
- [ ] When clicking on the icon, it either opens the menu or adds the element (if inserting is the only option available).
- [ ] Rework the accessibility to display focus on the insertion button by default (see details above)
- [ ] Rework accessibility to allow tabbing between the insertion button and the More menu (where applicable, see above)
- [ ] Enter / Space on the focused button is the same as clicking
- [ ] Ensure drag-drop still works as expected as well
- [ ] Remove the text quick-add button from the tabs

OLD INFORMATION:
~NB! Please see https://github.com/google/web-stories-wp/issues/10109#issuecomment-1016614451 -- the UX changed a bit meanwhile but Figma might not be updated!~

~We will have just one button (and get rid of the three-dotted button / delete buttons) with the following options:
Media~
- ~Insert Image (or Video, depending)~
- ~Add as background~
- ~Edit meta data~
- ~Delete from library~

~Shape, Text, default Page Templates: clicking will insert directly~

~Custom Page Templates: Menu will open:~
- ~Use template~
- ~Delete ~

~(Remove the Delete icon from the corner)~

## Additional Context
Screenshot 2022-01-06 at 12 36 22

Screenshot 2022-02-07 at 11 36 37

miina:
## Summary
- Adds an insertion overlay to each library element which has a scrim and a non-functional plus icon, or plus icon that opens an action menu
- Drag-drop still works
- Clicking anywhere other than the plus icon still inserts the element

## To-do

- [x] Refactor the new nested roving tabIndex util to combine it with the default one
- [x] Review all style
- [x] Rework Saved Page Templates to have two buttons
- [x] Adjust tests
- [x] Add new tests for the new menu items

## User-facing changes
Screenshot 2022-02-01 at 19 12 18
Screenshot 2022-02-01 at 19 12 11
Screenshot 2022-02-18 at 12 14 20
Screenshot 2022-02-18 at 12 14 05
Screenshot 2022-02-18 at 12 13 58

## Testing Instructions

1. Open the Local Media tab
2. Verify that hovering over media will reveal a two menu buttons (previously it was just one -- the More button in the right top corner). Verify that clicking on the Insertion menu button will open a menu with two options: Inserting / Using as background. Verify these work as expected.
3. Open 3rd party Media tab.
4. Verify that hovering over media will reveal a menu button -- clicking on it opens dropdown with two options: Inserting / Using as background. Verify these work as expected.
5. Verify that for Stickers, Shapes, Default Page Templates, Text Presets and Text Sets, the plus icon appears when hovering but it will insert the element right away instead of opening a menu.
6. For Saved Page Templates -- verify the plus icon uses the template and Delete icon will open the Delete confirmation dialogue.
7. For all element types: Verify that dragging the media from around the menu button is still possible, as is clicking outside of the menu button for inserting right away.
8. For all elements: Verify that keyboard accessibility works as expected (please see details per element in the [original issue description](https://github.com/GoogleForCreators/web-stories-wp/issues/10109) -- it's written out per each element under the "Accessibility" point)

## Checklist

---

Fixes #10109 
Fixes #9884


miina:
@barklund @merapi This still needs some test adjustments but would be great if you could already review/test otherwise.

There are a lot of file changes, however, most of the file changes are due to all the existing Karma tests needing updates since:
- There is no Text quick-add action anymore and that was used a lot.
- Clicking in the center of Media doesn't add it directly anymore, it opens the menu instead, this was also used quite a lot.

packages/e2e-test-utils/src/addTextElement.js
```diff
@@ -14,8 +14,12 @@
  * limitations under the License.
  */
 export default async function addTextElement() {
-  await expect(page).toClick('button[aria-label="Add new text element"]');
```
miina: @BrittanyIRL Tagging you here just to share that this PR will remove the text quick-add button and updates all the tests using it as well -- the button will soon not be available anymore for tests.

I'll be using the `Paragraph` instead when adding text since that's neutral styling-wise.

BrittanyIRL: Awesome, this also means that this line (which I neglected to make a proper TODO when I updated all the karma tests to check for accessibility issues) can run: https://github.com/GoogleForCreators/web-stories-wp/pull/9955/files#diff-7dc084a68522b10b39061c684b76c52b8c7e54e281adc120bbac8b5b4e891941R80

merapi:
I have one UX consideration (maybe follow-up?) because for me, the additional actions are difficult to discover (because the button is a +/add/insert sign?): should we add right-click menu with all of those options repeated?

Right now if you have the intention to delete the image it goes like this:
You just see the + icon so maybe you right-click the image to see if the delete option is there (common pattern).
It opens the browser menu, no delete option and now if you click on the + it will do nothing (+ button is broken now), just a second click will insert the image, check here:

https://user-images.githubusercontent.com/21036224/152330034-97a494bd-3c41-4d22-a4db-91cbe9e339f5.mov

I've found a strange bug while testing the keyboard navigation, checked it with an older version - it's also there, so filed a bug: https://github.com/GoogleForCreators/web-stories-wp/issues/10445

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
```diff
@@ -181,24 +181,13 @@ function Element({
...
-
   useKeyDownEffect(
     ref,
     {
       key: ['enter', 'space'],
     },
-    handleKeyDown,
-    [handleKeyDown]
+    () => setIsMenuOpen(true),
+    [setIsMenuOpen]
```
barklund: Because it's a state setter and we know it to be stable, we can effectively remove it, but I guess it's better to be safe here if someone refactors it to not be stable.

barklund (dismissed):
This works really well. My heart bleeds for all the tests though... I did not look through them at all!

miina:
> I have one UX consideration (maybe follow-up?) because for me, the additional actions are difficult to discover (because the button is a +/add/insert sign?): should we add right-click menu with all of those options repeated?

Yep, might be a good idea, knowing there are are options when clicking the `+` icon is not straightforward at all indeed. I'll follow up with UX (as a follow-up issue).

miina:
@merapi Let me know if you found anything else -- would like to move this to QA but would be great to get a second review as well since it touches quite a lot of files (even though most of these are very similar test file changes).

merapi (dismissed):
> @merapi Let me know if you found anything else -- would like to move this to QA but would be great to get a second review as well since it touches quite a lot of files (even though most of these are very similar test file changes).

I'm still checking the test files 😂 

LGTM for the most part, I would just look more deeply into `useCallbacks` because some of them (most of them?) don't do anything, **but don't worry too much about it here** because it's a project-wide issue for me (we don't have good practices around them).

For example:
`packages/story-editor/src/components/library/panes/text/textSets/textSet.js`
```js
  const [active, setActive] = useState(false);
  const makeActive = useCallback(() => setActive(true), []);
  const makeInactive = useCallback(() => setActive(false), []);
  ...
  return (
    <TextSetItem
      translateX={translateX}
      translateY={translateY}
      ref={ref}
      onKeyUp={handleKeyboardPageClick}
      onPointerEnter={makeActive}
      onFocus={makeActive}
      onPointerLeave={makeInactive}
      onBlur={makeInactive}
      {...rest}
    >
      ...
      {active && <InsertionOverlay />}
```

`TextSetItem` is just a `styled.button` so is there really a penalty for re-rendering it?
Let's say there is - but there is no `memo` on `TextSetItem` so it will re-render regardless.
Even if there would be `memo`, if `active &&` is used in the contents/children - it will re-render every time it changes.

The same thing happens in `fontPreview.js`, `shapePreview.js`.
In `pageTemplate.js`, `handleKeyboardPageClick` is `useCallbacked` but then `onKeyUp={(event) => handleKeyboardPageClick(event, page)}`.

That's why I'm a little bit against:
image

in our current approach, so **without matching `memo`.**

Here is an interesting post about an alternative approach (as opposed to using only when needed): https://attardi.org/why-we-memo-all-the-things/
So for me, it's either: use them with matching `memo` or use all of them all the time, just not like we currently do it.

miina:
> LGTM for the most part, I would just look more deeply into useCallbacks because some of them (most of them?) don't do anything, but don't worry too much about it here because it's a project-wide issue for me (we don't have good practices around them).

I was wondering about it as well but since these were generally used elsewhere as well, decided to opt for consistency 🤷 
Perhaps we should create a code quality issue to review these generally.

miina:
> Perhaps we should create a code quality issue to review these generally.

Or well, I can at least remove these from where it's assigned to just styled components in this PR.

Edit: Actually, better to create a new issue to look into that 😛

merapi:
> > Perhaps we should create a code quality issue to review these generally.
> 
> Or well, I can at least remove these from where it's assigned to just styled components in this PR.

I would say don't bother for now, maybe in the all-in approach, there are some benefits for using all three of them even on styled-components (they don't have an explicit render method) - that needs to be researched.

andrew-askedall:
Hi all, I think there was some miscommunication from the accessibility meeting we had about this. There should still be an overflow and a user can tab between the plus and overflow to access the items. Adding screenshots. I hope that resolves this issue.
Screen Shot 2022-02-04 at 10 59 08 AM
Screen Shot 2022-02-04 at 10 59 02 AM
Screen Shot 2022-02-04 at 10 58 57 AM

miina:
@andrew-askedall Hm, I'm quite certain the decision was made about having just one button with 4 menu options with people agreeing that it's better accessibility-wise to have just one menu button instead of two 🤔 .
[Figma:](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=11394%3A454653) (the one I have access to and work based off)
Screenshot 2022-02-04 at 18 03 38

The idea was also that clicking either Enter or Space will open the menu, making it straightforward for the keyboard users.

If we're now going back to the 2 buttons then we're again in the same place as before that discussion -- how would this work with keyboard accessibility with having two buttons and two menus like this? How would the user get to the buttons using keyboard? What would be highlighted when focusing on the whole element? What would happen with Enter/Space?

Let me know if we should schedule a new call for discussing this.

cc @swissspidy for visibility

andrew-askedall:
tapping space or enter do two different actions on the current implementation. Let's keep having space open the menu for the overflow and if a user hits enter, instead of automatically adding the image, the add menu is opened. I think that would solve the issue? Thoughts?

miina:
> tapping space or enter do two different actions on the current implementation. Let's keep having space open the menu for the overflow and if a user hits enter, instead of automatically adding the image, the add menu is opened. I think that would solve the issue? Thoughts?

This would mean that the focus highlight would always stay on the element itself, correct? Yes, this would technically definitely work with the Enter vs Space. I'm not sure how accessible it would be, e.g. when using screen readers -- how would the user know that these two options are available. Thoughts?
(cc @barklund since this is about accessibility, maybe you'd know if it's possible to make this work with screen readers)

miina:
@andrew-askedall I chatted with Morten quickly about this, copying his thoughts on having two buttons + accessibility:
> I think the proper solution is to allow the user to tab between the two buttons when focusing an item. So tab order would be "item before media list" > "add menu button on current media item" > "more menu button on current media item" > "item after media list". If you have keyboard focus anywhere within the media item, arrow keys change focused item. When changing focus to a new item, the "add menu button" would be the focused button. Then both menu buttons would activate using both enter and space as normal buttons do.

This would make it accessible in terms of being able to display focus around the currently focused menu button and also understandable for screen readers. It would mean that each media item would include two buttons to focus on, and not be focusable itself via keyboard. However, the whole media item as itself still needs to be clickable (to insert an element right away).

Let me know if this sounds reasonable to you.

andrew-askedall:
That sounds great. Lets do it.

packages/story-editor/src/components/library/panes/shared/hooks/useNestedRovingTabIndex.js
```diff
@@ -0,0 +1,265 @@
...
+}
+
+/**
+ * Gets the parent element of the active element based on given depth.
+ *
+ * @param e
+ * @param depth
+ * @returns {*}
+ */
+function getParentByDepth(e, depth) {
```
miina: This gets the parent element that is on the same level with the parent elements of other buttons. Perhaps `getAncestorByDepth` would be more accurate name here 🤔

packages/story-editor/src/components/library/panes/shared/hooks/useNestedRovingTabIndex.js
```diff
@@ -0,0 +1,265 @@
...
+
+/**
+ * Internal dependencies
+ */
+import { useConfig } from '../../../../../app';
+
+const SELECTOR = '[tabIndex="-1"]:not(:disabled)';
+
+// @todo Combine this with `useRovingTabIndex` so that if depth is set, this file is used, otherwise the default one is used.
+// @todo Move common helpers into a shared file between the two.
```
miina: @barklund @merapi WDYT of this approach for the nested buttons roving tab index? 

This is an adjustment of `useNestedRovingTabIndex` so a lot of similar stuff but with variations.

This still needs some polishing but should work with local Media library (used with the Insertion buttons only currently, the "More" button not yet). I'm also thinking of combining `useRovingTabIndex` and `useNestedRovingTabIndex` so that in the code `useRovingTabIndex` could always be used but if the `depth` is set, it would just use the nested version in the background.
Some assumptions it has:
- The current button (or any focusable element) has `tabIndex` `0`,  all the others have `-1` (kind of as it works currently)
- The selector for looking into nested elements to shift focus includes the `tabIndex=-1` meaning that the first element with `tabIndex=-1` will be selected. This means that multiple nested elements with `tabIndex=-1` wouldn't potentially work as expected. Unlikely use case anyway though.
- If nesting level changes but the `depth` is not updated, the code breaks (true now as well -- currently it works only with focusable siblings -- it has sometimes broken because those focusable elements have been wrapped into another element)

Thoughts?

packages/story-editor/src/components/library/panes/media/local/dropDownMenu.js
```diff
@@ -53,6 +54,7 @@ const MoreButton = styled(Button).attrs({ variant: BUTTON_VARIANTS.ICON })`
   border-radius: 100%;
   width: 28px;
   height: 28px;
+  opacity: ${({ display }) => (display ? '1' : '0')};
```
miina: Since this is now tabbable, it needs to be available already before, thus hiding it visually only.

packages/story-editor/src/components/library/panes/shapes/shapePreview.js
```diff
@@ -54,16 +53,14 @@ const Aspect = styled.button`
...
   background-color: ${({ theme }) => theme.colors.interactiveBg.previewOverlay};
 
-  transition: background-color ${BUTTON_TRANSITION_TIMING};
-
-  &:hover,
-  &:focus,
-  &.${ThemeGlobals.FOCUS_VISIBLE_SELECTOR} {
-    background-color: ${({ theme }) =>
-      theme.colors.interactiveBg.secondaryHover};
+  &.${ThemeGlobals.FOCUS_VISIBLE_SELECTOR} [role='presentation'],
```
miina, line 56: Since the [Insertion button](https://github.com/GoogleForCreators/web-stories-wp/pull/10388/files#diff-228b60b16d62b088a0a823741704a209b4858c13f3e2d24872ad26b9d50edc1aR45) is not a real button (unless it opens a menu -- in case of media and Saved page templates) then it seemed simpler to just assign a presentational role to it and add the focus outline inside the element that's actually focusable and interactive if the parent is focused.

Alternative could be reworking all the element library to have the `+` icon as a "real" button instead and focus on that only. Then we would need to use the new nested roving tab index hook. The parent element still needs to be interactive as well since it needs to allow clicking anywhere and register hovering, so not sure if it's worthwhile but might be more "correct" perhaps. Not sure.

Thoughts?

cc @barklund / @merapi For a sanity check

packages/story-editor/src/components/canvas/karma/backgroundCopyPaste.karma.js
```diff
@@ -55,16 +55,14 @@ describe('Background Copy Paste integration', () => {
 
     // Verify setup - 1 element on each page in the right color
     await gotoPage(1);
-    expect(await getPageArea()).toHaveStyle(
-      'background-color',
-      'rgb(255, 0, 0)'
-    );
+    await waitFor(() => {
+      expect(getPageArea()).toHaveStyle('background-color', 'rgb(255, 0, 0)');
```
BrookeGraham, line 63: Here, since `getPageArea` is a `querySelector` which could return `null`, the `waitFor` won't repoll if the first return happens to be `null` since that is a valid option. If we want to wait until it isn't `null` we would have to manually throw an error within this `waitFor` ie.
```
 await waitFor(() => {
  const pageArea = getPageArea();
  if(!pageArea) {
    throw new Error('not ready!')
  }
    expect(getPageArea()).toHaveStyle('background-color', 'rgb(0, 255, 0)');
});
```

packages/story-editor/src/components/library/panes/media/common/insertionMenu.js
```diff
@@ -0,0 +1,200 @@
...
+  right: calc(50% - 16px);
+  color: ${({ theme }) => theme.colors.fg.primary};
+  border-radius: 100%;
+  width: 28px;
+  height: 28px;
+  opacity: ${({ display }) => (display ? '1' : '0')};
+`;
+
+const IconContainer = styled.div`
+  height: 32px;
```
BrittanyIRL: any value in this being a const and then using it divided by 2 to get your 16px on line 51/52 so that the sizing and positioning stay connected easily?

packages/story-editor/src/components/library/karma/mediaTab.karma.js
```diff
@@ -73,11 +74,43 @@ describe('Library Media Tab', () => {
...
+      ).toBeDefined();
+    });
+
+    it('should allow setting media as background from the dropdown menu', async () => {
+      const mediaItem = fixture.editor.library.media.item(0);
+      // Hover the media
+      await fixture.events.mouse.moveRel(mediaItem, 20, 20, { steps: 2 });
+      await waitFor(() =>
+        expect(
+          fixture.screen.getByRole('button', { name: 'More' })
```
BrookeGraham: Since you want to await this, you can use `findByRole` which combines `getByRole` and `waitFor`. 
ie. 
```
expect( await fixture.screen.findByRole('button', { name: 'More' })).toBeDefined();
```

packages/story-editor/src/components/library/panes/media/common/insertionMenu.js
```diff
@@ -0,0 +1,200 @@
...
+          value: 'addBackground',
+        },
+      ],
+    },
+  ];
+
+  const handleCurrentValue = (evt, value) => {
+    const thumbnailUrl = poster
+      ? poster
+      : getSmallestUrlForWidth(width, resource);
```
BrittanyIRL: nit: This'd work the same.
```suggestion
    const thumbnailUrl = poster || getSmallestUrlForWidth(width, resource);
```

packages/story-editor/src/components/library/panes/media/common/insertionMenu.js
```diff
@@ -0,0 +1,200 @@
...
+    const newElement = getElementProperties(resource.type, {
+      resource,
+    });
+    switch (value) {
+      case 'insert':
```
BrittanyIRL: nit: can you pull these in as constants instead?

packages/story-editor/src/components/library/panes/media/common/insertionMenu.js
```diff
@@ -0,0 +1,200 @@
...
+          firstElement: newElement,
+          secondId: currentBackgroundId,
+        });
+        break;
+      default:
+        break;
+    }
+  };
+
+  useNestedRovingTabIndex({ ref: insertButtonRef }, [], 3);
```
BrittanyIRL: What is `3` here?  I'm a broken record here, but giving this a name would help when reading through the code to know what it does without having to trace it back to the hook.

packages/story-editor/src/components/library/panes/media/common/insertionMenu.js
```diff
@@ -0,0 +1,200 @@
...
+      >
+        <IconContainer>
+          <Icons.PlusFilled />
+        </IconContainer>
+      </AddButton>
+      {(display || isMenuOpen) && (
+        <Popup
+          anchor={insertButtonRef}
+          placement={PLACEMENT.BOTTOM_START}
+          isOpen={isMenuOpen}
```
BrittanyIRL, line 176: should menu be open on `display` true too?

packages/story-editor/src/components/library/panes/media/common/insertionMenu.js
```diff
@@ -0,0 +1,200 @@
...
+            />
+          </DropDownContainer>
+        </Popup>
+      )}
+    </MenuContainer>
+  );
+}
+
+InsertionMenu.propTypes = {
+  resource: PropTypes.object.isRequired,
```
BrittanyIRL: could you use this `ResourcePropTypes` (found it referenced in story-editor types) so avoid the blanket `object` shape?

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
```diff
@@ -216,7 +191,7 @@ function Element({
       onFocus={makeActive}
       onPointerLeave={makeInactive}
       onBlur={makeInactive}
-      tabIndex={index === 0 ? 0 : -1}
+      tabIndex="-1"
```
BrittanyIRL, line 194: 🤔  this shouldn't be needed at all, `Container` is a `div` which isn't focusable by default.  That said, if we ever wanted to do aXe testing here, this set up might make it angry because it has `onFocus`, etc. So, should this be a button?

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
```diff
@@ -247,6 +222,13 @@ function Element({
           isCurrentResourceUploading(resourceId)) && (
           <LoadingBar loadingMessage={__('Uploading media…', 'web-stories')} />
         )}
+        <InsertionMenu
```
BrittanyIRL, line 225: 💭  When we had to set up buttons that appeared over an image that had it's own functionality in various iterations of the dashboard what we ended up doing was just visually overlaying them and letting the html hierarchy fall like this: 
- container that is also a button that can get clicked on and used 
  - image shown for that container 
- another button 
- another button

```
<myLayeredDesign>
  <firstInteractiveElementContainer>
    <whateverINeedNestedToGetCSSWorking /> 
  </firstInteractiveElementContainer> 
  <WrapperToEaseOverlayFunctionality>
    <visuallyOnTopOfFirstInteractiveElement /> 
    <visuallyOnTopOfFirstInteractiveElement />
  </WrapperToEaseOverlayFunctionality>
</myLayeredDesign>
```

This makes the tab order way easier to preserve and allows screenreaders and keyboard users to have easier HTML outline from the get go.

packages/story-editor/src/components/library/panes/media/local/dropDownMenu.js
```diff
@@ -206,6 +212,7 @@ DropDownMenu.propTypes = {
   onMenuOpen: PropTypes.func.isRequired,
   onMenuCancelled: PropTypes.func.isRequired,
   onMenuSelected: PropTypes.func.isRequired,
+  index: PropTypes.number.isRequired,
```
BrittanyIRL: i don't see where `index` is getting used here?

packages/story-editor/src/components/library/panes/pageTemplates/dropDownMenu.js
```diff
@@ -0,0 +1,193 @@
...
+import {
+  Button,
+  BUTTON_VARIANTS,
+  Icons,
+  Menu,
+  PLACEMENT,
+  Popup,
+} from '@googleforcreators/design-system';
+
+const AddButton = styled(Button).attrs({ variant: BUTTON_VARIANTS.ICON })`
```
BrittanyIRL: Same as other `AddButton` - could this be pulled into 1 place?

BrittanyIRL (commented):
I only had an hour before a meeting, so i'm reviewing this in batches!

packages/story-editor/src/components/library/panes/pageTemplates/templateList.js
```diff
@@ -127,19 +128,6 @@ function TemplateList({
     gridItemIds: pageIds,
   });
 
-  const handleKeyboardPageClick = useCallback(
```
BrittanyIRL, line 130: Does this solve this? https://github.com/GoogleForCreators/web-stories-wp/issues/9884

packages/story-editor/src/components/library/panes/shapes/shapePreview.js
```diff
@@ -123,6 +120,10 @@ function ShapePreview({ mask, isPreview, index }) {
     dataToEditorY: state.actions.dataToEditorY,
   }));
 
+  const [active, setActive] = useState(false);
+  const makeActive = useCallback(() => setActive(true), []);
+  const makeInactive = useCallback(() => setActive(false), []);
```
BrittanyIRL: nit: Callbacks here don't save any resources. https://kentcdodds.com/blog/usememo-and-usecallback

packages/story-editor/src/components/library/panes/shared/hooks/useNestedRovingTabIndex.js
BrittanyIRL: perhaps, parent siblings is the most descriptive of the same level parents?

packages/story-editor/src/components/library/panes/shared/insertionOverlay.js
```diff
@@ -0,0 +1,62 @@
...
+  left: 0;
+  background-color: ${({ theme }) => rgba(theme.colors.standard.black, 0.5)};
+`;
+
+function InsertionOverlay({ showIcon = true }) {
+  // The icon looks like a button but is just representational.
+  // The real interactive element is the containing element.
+  return (
+    <Wrapper>
+      {showIcon && (
```
BrittanyIRL, line 53: 🤔  is it intentional to still render the wrapper when showIcon is false?

packages/story-editor/src/components/library/panes/text/fontPreview.js
```diff
@@ -35,10 +41,11 @@ import { useUnits } from '@googleforcreators/units';
 import { useFont, useHistory } from '../../../../app';
 import StoryPropTypes from '../../../../types';
 import stripHTML from '../../../../utils/stripHTML';
-import { focusStyle } from '../../../panels/shared';
 import usePageAsCanvas from '../../../../utils/usePageAsCanvas';
 import useLibrary from '../../useLibrary';
 import LibraryMoveable from '../shared/libraryMoveable';
+import InsertionOverlay from '../shared/insertionOverlay';
+import useRovingTabIndex from '../../../../utils/useRovingTabIndex';
```
BrittanyIRL, line 48: nit: can you import `useRovingTabIndex` and `usePageAsCanvas` just from `../../../../utils` together?

packages/story-editor/src/components/library/test/mediaElement.js
```diff
@@ -193,66 +194,4 @@ describe('MediaElement', () => {
 
     expect(getByAriaLabel('More')).toBeInTheDocument();
   });
-
-  it("should render dropdown menu's more icon users without permission", () => {
```
BrittanyIRL, line 197: could these tests still be used, just updated for new menu?

BrittanyIRL (commented):
Ok, went through the rest of it! I think this is looking good. I don't have any issue actually with how you did the buttons/popup stuff by element, on revisit. 

I also ran through with aXe for any issues I wasn't seeing and no complaints. I do think adding an aXeNoViolations karma test for the elements when an element is focused would be a good thing to add. 
 
_classic I'm not a designer but_ it surprised me that clicking some of these plus signs did an action automatically and some opened a menu. I'm all for consistency, so i get the same overlay but it felt unexpected at first. Maybe that's because I'm used to how it is now? 

There's quite a bit of repeated styles occurring. I'm wondering how best to dry that up so it's easier to maintain or update? In an ideal world, this is the first step towards a reusable panel element list that can take some of this and make the behavior consistent! The panels all have like slightttt differences. 😅 

General observations: 
- Testing with a screen reader, the icon menus that have a popup don't give me any context what the menu is for - having the labeling include the element's title or description would be really helpful. You can contrast it with the icons that are just presentational which announce what they are (shapes -> "rectangle button")
- `shift + tab` when in a list of elements should untrap the focus back to the previous element outside of the list.  I think this isn't always the case in staging, but if a new roving hook is getting introduced it should be taken care of to not persist the issue.  (What I'm seeing is: local media works great, 3p media focus gets lost on shift tab, text sets shift tab goes up a row, shapes and stickers work great, templates work well too)
- I suppose this is out of scope as it's current with staging, but keyboard focus does get lost if you go to a dialog and then cancel or complete an action, it'd be nice to have focus return to the element in at least cancel cases. 
- This doesn't super matter since we don't have the ability to time travel (and if we did this would be a really lame use of it) - it might have been easier to process all of this had you just done 1 panel as a POC first as a PR. ❤️  ❤️  That said, I appreciate the comments and notes you left.

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
BrittanyIRL: meh, on revisit i think what you have makes sense, it just took me some time to parse through it.

packages/story-editor/src/components/canvas/karma/backgroundCopyPaste.karma.js
miina: Thanks for pointing that out, Brooke! Since there are still changes coming in this PR, I'll leave the tests for the last but I'll definitely go through everything making sure `waitFor` is used correctly and double check the usage of `getByRole` => `findByRole`.

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
miina: Thanks so much for reviewing the PR, Brittany! It might need a bit too much refactoring for this PR to change it to a button since it currently already contains a button (the dropdown menu) + has to also work with the Moveable for dragging etc. I will create a follow-up issue for looking into changing these to buttons (here and elsewhere where these are currently not buttons).

packages/story-editor/src/components/library/panes/shared/insertionOverlay.js
miina: Yep, since in some cases the icon is just presentational (when clicking anywhere will insert the element right away) where in other cases, it's an actual button opening a menu. But the dark scrim will always be displayed regardless of the icon. I'll leave a comment about it.

packages/story-editor/src/components/library/test/mediaElement.js
miina: Yep, I haven't added these back yet. In the previous iteration of this issue, the More menu was removed and everything combined to one menu, later UX made a change to have 2 menus instead -- haven't gotten to updating the tests yet after that change.

packages/story-editor/src/components/library/panes/pageTemplates/templateList.js
miina: Yes, it should fix that one. Enter and Space will now have default behaviour while focused on the button (clicking it).

packages/story-editor/src/components/library/panes/shared/insertionOverlay.js
BrittanyIRL: awesome, thanks for the clarification.

packages/story-editor/src/components/library/panes/media/common/mediaElement.js
BrittanyIRL: totally agree it's a good follow up - not anything for right now.

packages/story-editor/src/components/library/panes/shapes/shapePreview.js
miina: I added it mainly because saw the same thing done elsewhere, for consistency. We have an open issue for generally establishing best practices for these: #10467. I can remove from here though.

packages/story-editor/src/components/library/panes/text/fontPreview.js
miina: There is no index file currently in that folder so all the utils are imported from the specific files.

packages/story-editor/src/components/library/panes/shapes/shapePreview.js
miina: Keeping as it is for now for simplicity, we can create a follow-up issue to rework this if necessary. Changing it to a real button will require quite a bit of refactoring.

miina:
@barklund @merapi The tests still need updating after the UX was changed but please review the rest already when you can.

merapi:
Should we close the menu after inserting/adding as BG?

https://user-images.githubusercontent.com/21036224/154256445-e6975433-82c8-480a-bccf-8bc4d2eedac9.mov



image

Should we use nowrap here?
image

merapi:
> _classic I'm not a designer but_ it surprised me that clicking some of these plus signs did an action automatically and some opened a menu. I'm all for consistency, so i get the same overlay but it felt unexpected at first. Maybe that's because I'm used to how it is now?

Same for me.

miina:
>> classic I'm not a designer but it surprised me that clicking some of these plus signs did an action automatically and some opened a menu. I'm all for consistency, so i get the same overlay but it felt unexpected at first. Maybe that's because I'm used to how it is now?

> Same for me.

Same here but I might also be just used to it due to seeing everything as is from the very beginning?
Reasoning for adding the + icons was that it makes it more intuitive for the users to understand that clicking on an element will insert it. As far as I understand the new UX has received positive feedback based on the mocks. We'll see the user feedback after the release :)

packages/story-editor/src/app/helpCenter/useHelpCenter/effects.js
```diff
@@ -94,10 +94,7 @@ export function deriveUnreadTipsCount(previous, next) {
 }
 
 export function deriveAutoOpen(previous, next) {
```
miina: Ignore changes here, this is reverting a PR that causes multiple test failures, see https://github.com/GoogleForCreators/web-stories-wp/pull/10589#issuecomment-1042875858

packages/story-editor/src/components/errorBoundary/index.js
```diff
@@ -53,7 +53,6 @@ class ErrorBoundary extends Component {
 }
 
 const shouldDisableErrorBoundaries =
-  typeof WEB_STORIES_DISABLE_ERROR_BOUNDARIES !== 'undefined' &&
```
miina: Ignore changes here, this is reverting a PR that causes multiple test failures, see https://github.com/GoogleForCreators/web-stories-wp/pull/10589#issuecomment-1042875858

packages/story-editor/src/components/library/panes/text/textSets/textSet.js
```diff
@@ -113,9 +115,12 @@ function TextSet({ id, elements, translateY, translateX, ...rest }, ref) {
     [onClick]
   );
 
-  const renderImages =
-    typeof WEB_STORIES_DISABLE_OPTIMIZED_RENDERING !== 'undefined' &&
-    WEB_STORIES_DISABLE_OPTIMIZED_RENDERING !== 'true';
```
miina, line 118: Ignore this change, this is reverting a PR that causes multiple test failures, see https://github.com/GoogleForCreators/web-stories-wp/pull/10589#issuecomment-1042875858

packages/design-system/src/icons/index.js
```diff
@@ -78,6 +78,7 @@ export { ReactComponent as DistributeHorizontalchevron } from './distribute_hori
 export { ReactComponent as DistributeVertical } from './distribute_vertical.svg';
 export { ReactComponent as Dots } from './dots.svg';
 export { ReactComponent as DotsFill } from './dots_fill.svg';
+export { ReactComponent as DotsFillSmall } from './dots_fill_small.svg';
```
swissspidy, line 81: Could these be combined?

merapi: Looks like scaled down version:
image

Similar case with other icons e.g. 3*cross - maybe those can be combined? Large and small are used once. (not for this PR):
image

merapi (approved):
I still see text wrapping, but otherwise ✅ 

image

packages/design-system/src/icons/index.js
miina: I'll create a follow-up issue for revisiting the way how SVG-s are handled.

Edit: added https://github.com/GoogleForCreators/web-stories-wp/issues/10608

packages/story-editor/src/components/library/panes/media/common/insertionMenu.js
```diff
@@ -0,0 +1,197 @@
...
+ * @param {Object} props Component props.
+ * @param {Object} props.resource Selected media element's resource object.
+ * @param {boolean} props.display Whether the more icon should be displayed.
+ * @param {Function} props.onInsert Callback for inserting media.
+ * @param {number} props.width Media width.
+ * @param {number} props.index Element index in the gallery.
+ * @param {boolean} props.isLocal If the menu is for local or 3p media.
+ * @return {null|*} Element or null if should not display the More icon.
+ */
+function InsertionMenu({ resource, display, onInsert, width, index, isLocal }) {
```
barklund: Nit:

```suggestion
function InsertionMenu({ resource, display, onInsert, width, index, isLocal = false }) {
```

packages/story-editor/src/components/library/panes/media/common/insertionMenu.js
```diff
@@ -0,0 +1,197 @@
...
+        break;
+      default:
+        break;
+    }
+  };
+
+  useRovingTabIndex({ ref: insertButtonRef }, [], BUTTON_NESTING_DEPTH);
+  // In case of 3p media, we tab out of the panel since it's the only button there.
+  const focusCanvas = useFocusCanvas();
+  useKeyDownEffect(isLocal ? null : insertButtonRef, 'tab', focusCanvas, [
```
barklund, line 148: Not a big fan of optionally adding the listener like this, but because it's a hook, I guess it's the only really good way to do it.

Alternatively we could add a new argument to the function, that would indicate whether or not to actually add the listener or just do nothing (because hooks can't be optional).

packages/story-editor/src/components/library/panes/text/fontPreview.js
```diff
@@ -79,7 +89,7 @@ const DragContainer = styled.div`
   line-height: ${({ lineHeight }) => lineHeight}px;
 `;
 
-function FontPreview({ title, element, insertPreset, getPosition }) {
+function FontPreview({ title, element, insertPreset, getPosition, index }) {
```
barklund, line 92: So this component only ever renders on mount and never again right - so we don't need memo here?

packages/story-editor/src/components/library/panes/text/fontPreview.js
```diff
@@ -191,9 +201,23 @@ function FontPreview({ title, element, insertPreset, getPosition }) {
     [onClick]
   );
 
+  useRovingTabIndex({ ref: buttonRef });
+
+  const [active, setActive] = useState(false);
+  const makeActive = useCallback(() => setActive(true), []);
+  const makeInactive = useCallback(() => setActive(false), []);
```
barklund: No need to memoize these - preview is just an S-C object and those aren't memoized (and this component never re-renders anyway either).

packages/story-editor/src/components/library/panes/text/textSets/textSet.js
```diff
@@ -117,6 +119,11 @@ function TextSet({ id, elements, translateY, translateX, ...rest }, ref) {
     typeof WEB_STORIES_DISABLE_OPTIMIZED_RENDERING === 'undefined' ||
     WEB_STORIES_DISABLE_OPTIMIZED_RENDERING !== 'true';
 
+  useRovingTabIndex({ ref: ref.current });
+  const [isHovering, setIsHovering] = useState(false);
+  const setHovering = useCallback(() => setIsHovering(true), []);
+  const unsetHovering = useCallback(() => setIsHovering(false), []);
```
barklund: Again, no need for memoization.

packages/story-editor/src/components/canvas/karma/backgroundCopyPaste.karma.js
```diff
@@ -256,7 +265,11 @@ describe('Background Copy Paste integration', () => {
   async function addBackgroundImage(index) {
     // Add image and click "set as background"
     const image = fixture.editor.library.media.item(index);
-    await fixture.events.click(image);
+    await fixture.events.mouse.seq(({ moveRel, down, up }) => [
+      moveRel(image, 20, 20),
+      down(),
+      up(),
+    ]);
```
barklund: You can do `clickOn(image, 20, 20)` - at least I think that would work. It takes optional relative position in x and y as positional arguments at least.

So just:

```suggestion
    await fixture.events.mouse.clickOn(image, 20, 20);
```

packages/story-editor/src/utils/useRovingTabIndex/flatNavigation.js
```diff
@@ -0,0 +1,31 @@
...
+ * Media elements are nested 1 level deep.
+ *
+ * @param {Element} e The element.
+ * @param {string} siblingDirection The sibling direction (previousSibling or nextSibling).
+ * @return {Element} The sibling.
+ */
+export function getNextEnabledSibling(e, siblingDirection) {
+  let sibling = e[siblingDirection];
+  while (sibling && sibling.disabled) {
+    sibling = sibling[siblingDirection];
```
barklund, line 28: This is reported as not covered by test. Could we make sure to add a test that this does in fact skip disabled elements?

packages/story-editor/src/utils/useRovingTabIndex/index.js
```diff
@@ -64,76 +73,73 @@ function getDistanceSq(p1, p2) {
  *
  * @param {boolean} isRTL Whether the document is RTL.
  * @param {string} key The key pressed.
- * @return {string} Either `previousSibling` or `nextSibling`.
+ * @return {string} Either `prev` or `next`.
```
barklund: But that's not what it returns though?

packages/story-editor/src/components/library/panes/media/common/insertionMenu.js
```diff
@@ -0,0 +1,197 @@
...
+          secondId: currentBackgroundId,
+        });
+        break;
+      default:
+        break;
+    }
+  };
+
+  useRovingTabIndex({ ref: insertButtonRef }, [], BUTTON_NESTING_DEPTH);
+  // In case of 3p media, we tab out of the panel since it's the only button there.
```
barklund, line 146: Actually, I think that focus should move to the attribution link to the 3rd party service provider. That's what happens in reverse tab order anyway.

Currently:
1. Focus an unsplash media item with keyboard
2. Press tab and focus moves to canvas
3. Press shift-tab and focus moves to "unsplash" attribution link
4. Press shift-tab and focus moves to unsplash media item

What should happen:
1. Focus an unsplash media item with keyboard
2. Press tab and moves to "unsplash" attribution link
2. Press tab and focus moves to canvas
3. Press shift-tab and focus moves to "unsplash" attribution link
4. Press shift-tab and focus moves to unsplash media item

WDYT?

packages/story-editor/src/components/library/panes/media/common/insertionMenu.js
```diff
@@ -0,0 +1,197 @@
...
+        onClick={onMenuOpen}
+        aria-label={__('Open insertion menu', 'web-stories')}
+        aria-pressed={isMenuOpen}
+        aria-haspopup
+        aria-expanded={isMenuOpen}
+        aria-owns={isMenuOpen ? listId : null}
+        id={buttonId}
+        $display={display}
+        tabIndex={index === 0 ? 0 : -1}
+      >
```
barklund, line 169: If you tab out of the media item, while the insertion menu is open, this action button stays visible. This also means, that if you shift-tab back to the media item, this action button receives keyboard focus, but you cannot use arrow keys to navigate to neighboring media items.

How to reproduce:

* Move keyboard focus to local media item
* Press enter to open insertion menu
* Press tab to move focus to canvas
* Press shift-tab to move focus back to media item, and observe focus moves to insertion menu button
* Try to navigate to other media items using arrow keys and observe it **does not** work
* Press shift-tab again to move focus back to the entire media item
* Try to navigate to other media items using arrow keys and observe it **does** work

I think the solution is to always hide this button once the media item loses `focus-within`, but not sure.

insertion-tab-bug

packages/story-editor/src/components/library/panes/media/common/insertionMenu.js
```diff
@@ -0,0 +1,197 @@
...
+ * @return {null|*} Element or null if should not display the More icon.
+ */
+function InsertionMenu({ resource, display, onInsert, width, index, isLocal }) {
+  const insertButtonRef = useRef();
+  const [isMenuOpen, setIsMenuOpen] = useState(false);
+  const onMenuOpen = useCallback((e) => {
+    e.stopPropagation();
+    setIsMenuOpen(true);
+  }, []);
+  const onMenuCancelled = useCallback(() => setIsMenuOpen(false), []);
```
barklund, line 101: When you press esc to close the insertion menu, focus is moved to the canvas rather than back to the media item.

How to reproduce:
* Use tab to focus media item
* Press <kbd>enter</kbd> to open insertion menu
* Press <kbd>esc</kbd> to close insertion menu
* Observe that focus is now on canvas, not media item

The solution here is potentially just to force focus to the media item in this `onMenuCancelled` callback?

barklund (approved):
This works really well. Just a bunch of nits on keyboard focus flow mostly.

packages/story-editor/src/components/library/panes/media/common/insertionMenu.js
miina: There's an issue to follow-up in this: #10607 
(This was an issue before this PR already, too)

packages/story-editor/src/components/library/panes/media/common/insertionMenu.js
miina: Hm, I'll create a follow-up issue for it. Looks like the same thing is happening on `main`, too.

Follow-up: https://github.com/GoogleForCreators/web-stories-wp/issues/10642

packages/story-editor/src/components/library/panes/media/common/insertionMenu.js
miina: We could perhaps create a follow-up issue for it, seems like a general code quality question since that's currently indeed the way to do it.

packages/story-editor/src/components/library/panes/text/fontPreview.js
miina: It might start needing memo in #10609 -- I've added a note already into a todo list there to check Performance to ensure it will only re-renders when actually needed. Let's revisit there.

felipebochehin87:
Tested against https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/, using PR #10388.

1. Tested a11y
2. Tested add button, drag and drop, clicking on the item itself
3. Tested delete
4. Tested saved templates (add, delete)
5. No issues found except for the existing ones, already mentioned and that will be fixed in a new PR


https://user-images.githubusercontent.com/96055551/154937450-8e99f578-f800-4edd-8f2c-e6a81af15f80.mp4


https://user-images.githubusercontent.com/96055551/154937458-822c5e56-4906-4577-b4f1-411d1413a51b.mp4


https://user-images.githubusercontent.com/96055551/154937460-8e8b9d13-0d8a-42ca-8a46-abcd8575aacd.mp4

packages/story-editor/src/components/canvas/karma/backgroundCopyPaste.karma.js
miina: Eh, good point!

packages/story-editor/src/components/library/panes/media/common/insertionMenu.js
miina: Looks like the same behaviour is happening on main as well. Included this into https://github.com/GoogleForCreators/web-stories-wp/issues/10607

packages/story-editor/src/components/library/panes/media/common/insertionMenu.js
barklund: Yeah, I know. It just looks a bit ugly, and it's only because I know that you want to do a conditional hook, that I know what the code is supposed to do. If I didn't know that, this would confuse the hell out of me! 😂 

Follow-up is probably a good idea. I think I've done similar tricks elsewhere and felt bad about it.

miina: Added a follow-up issue: https://github.com/GoogleForCreators/web-stories-wp/issues/10670

PR #10389, opened 2022-01-26
Support `tel:` and `mailto:` links

### LINKED ISSUE #3221: Allow non-http links in URL
## Feature Description

Currently, links entered in the URL field are automatically appended `http://`. I am trying to add a `tel:`  link so that mobile users can initiate a phone call.

## Alternatives Considered

There are no other alternatives at the moment. 

## Additional Context

image

---

swissspidy:
## Context

Currently anything that isn't `http` or `https` is denied

## Summary

Adds support for `tel:` and `mailto:` URLs for links and page attachments

## User-facing changes

Screenshot 2022-01-26 at 16 04 52
Screenshot 2022-01-26 at 16 04 38

## Testing Instructions

1. Ensure that `tel:` and `mailto:` links can be used when adding links and page attachments

## Checklist

---

Fixes #3221

### DIFF
--- a/packages/design-system/src/utils/test/url.js
+++ b/packages/design-system/src/utils/test/url.js
@@ -28,6 +28,8 @@ describe('isValidUrl', () => {
     expect(isValidUrl('https://foo-bar.com/test.jpg')).toBeTrue();
     expect(isValidUrl('https://foo-bar.com/example.html')).toBeTrue();
     expect(isValidUrl('https://foo-bar.com/font.ttf')).toBeTrue();
+    expect(isValidUrl('tel:1234567')).toBeTrue();
+    expect(isValidUrl('mailto:example@example.com')).toBeTrue();
   });
 });
 
--- a/packages/story-editor/src/components/panels/design/pageAttachment/pageAttachment.js
+++ b/packages/story-editor/src/components/panels/design/pageAttachment/pageAttachment.js
@@ -132,7 +132,7 @@ function PageAttachmentPanel() {
 
   const populateUrlData = useDebouncedCallback(async (value) => {
     // Nothing to fetch for tel: or mailto: links.
-    if (!value.startsWith('http://') || !value.startsWith('https://')) {
+    if (!value.startsWith('http://') && !value.startsWith('https://')) {
       return;
     }
 
--- a/packages/story-editor/src/components/panels/design/pageAttachment/pageAttachment.js
+++ b/packages/story-editor/src/components/panels/design/pageAttachment/pageAttachment.js
@@ -130,31 +130,39 @@ function PageAttachmentPanel() {
   const debouncedUpdate = useDebouncedCallback(updatePageAttachment, 300);
   const { getProxiedUrl, checkResourceAccess } = useCORSProxy();
 
-  const populateUrlData = useDebouncedCallback(async (value) => {
-    // Nothing to fetch for tel: or mailto: links.
-    if (!value.startsWith('http://') && !value.startsWith('https://')) {
-      return;
-    }
+  const populateUrlData = useDebouncedCallback(
+    useCallback(
+      async (value) => {
+        // Nothing to fetch for tel: or mailto: links.
+        if (!value.startsWith('http://') && !value.startsWith('https://')) {
+          return;
+        }
 
-    setFetchingMetadata(true);
-    try {
-      const { image } = getLinkMetadata ? await getLinkMetadata(value) : {};
-      const iconUrl = image ? toAbsoluteUrl(value, image) : '';
-      const needsProxy = iconUrl ? await checkResourceAccess(iconUrl) : false;
+        setFetchingMetadata(true);
+        try {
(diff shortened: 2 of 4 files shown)


packages/story-editor/src/components/panels/design/pageAttachment/pageAttachment.js
```diff
@@ -131,6 +131,11 @@ function PageAttachmentPanel() {
   const { getProxiedUrl, checkResourceAccess } = useCORSProxy();
 
   const populateUrlData = useDebouncedCallback(async (value) => {
+    // Nothing to fetch for tel: or mailto: links.
+    if (!value.startsWith('http://') || !value.startsWith('https://')) {
```
miina: Would it be better to just check for `tel:` and `mailto:` here instead and return only if the value starts with those?

Even though currently, only `withProtocol` URLs are used with `populateUrlData` then still seems more correct -- for `populateUrlData` not assuming it's used correctly. Thoughts?

swissspidy: We could add an `isValidURL` check here if we're worried about the URL being incorrect.

Opted for checking `http`/`https` in case we want to support more protocols in the future. So an _allowlist_ rather than a _denylist_.

miina: Ok, it's just that if there's a URL that doesn't start with `http(s)://` and also not with `tel:` nor `mailto:` that ends up here,  it would change the behaviour. So I'd prefer a denylist here :) But if you think it's ok then let's keep it, it's unlikely to happen.

swissspidy: > Ok, it's just that if there's a URL that doesn't start with `http(s)://` and also not with `tel:` nor `mailto:` that ends up here, it would change the behaviour.

How so? We cannot populate URL data for anything that isn't an http or https link. Hence returning early here.

packages/design-system/src/utils/test/url.js
```diff
@@ -28,6 +28,8 @@ describe('isValidUrl', () => {
     expect(isValidUrl('https://foo-bar.com/test.jpg')).toBeTrue();
     expect(isValidUrl('https://foo-bar.com/example.html')).toBeTrue();
     expect(isValidUrl('https://foo-bar.com/font.ttf')).toBeTrue();
+    expect(isValidUrl('tel:1234567')).toBeTrue();
+    expect(isValidUrl('mailto:example@example.com')).toBeTrue();
```
miina, line 32: Q: Should we also validate the email in case it's `mailto:`?

packages/story-editor/src/components/panels/design/pageAttachment/pageAttachment.js
miina: > How so? We cannot populate URL data for anything that isn't an http or https link. Hence returning early here.

No, we can't, but we're setting URL as invalid and remove the icon if that's the case. Now we would just return without doing it.

packages/design-system/src/utils/test/url.js
swissspidy: Seems overlay precautious.

packages/story-editor/src/components/panels/design/pageAttachment/pageAttachment.js
swissspidy: Maybe I'm not following you here 🤔 

`handleChangeUrl` only calls `populateUrlData` for valid (aka legit-looking) URLs, and then if `populateUrlData` can't actually reach it, it's marked as invalid again.

The behavior I'm expecting:

1. Type `foo:bar` into field -> marked as invalid
2. Type `blahblah` into field -> gets turned into `https://blahblah` -> looks valid -> try to fetch data -> marked as invalid
3. Type `example.com` into field -> gets turned into `https://example.com` -> looks valid -> try to fetch data -> valid
4. Type `mailto:foobar` into field -> is valid

Note:

I just realized the `||` in this condition needs to be an `&&`. Fixing that right now.

miina: > Maybe I'm not following you here 

The thought was that _if_ an URL ends up inside `populateUrlData` that doesn't have the protocol nor `tel` nor `mailto` _then_ the behaviour would change (because then we would just return instead of setting the URL invalid + removing the icon). Meaning that currently, `populateUrlData` assumes that the url that gets to it is already validated. Which is ok since currently it's true -- `populateUrlData` is only called with valid URLs. The idea was mainly that I'd prefer not assuming that a function is always called with valid URLs but if you think it's OK here, we can keep it as is.

swissspidy: I'd prefer making this assumption and having the validation in one place instead of multiple places.

swissspidy:
While testing I noticed that `populateMetadata`/`populateUrlData` are still running even when the URL has already changed.

Added some cancellation there so that this doesn't happen anymore.

miina (approved):

PR #10390, opened 2022-01-26
Update tests after WordPress 5.9 release

### LINKED ISSUE #10257: Update tests after WordPress 5.9 release
## Task Description

After WP 5.9 is released we can remove the e2e test against WP 5.9 RC.

We can also update the `php-stubs/wordpress-stubs` dependency accordingly.

Note that this dependency will be removed by #4817, so all that's needed afterwards is a `composer update`

swissspidy:
Fixes #10257


spacedmonkey (commented):
👍

PR #10396, opened 2022-01-27
Performance: Reduce Layout Forced Reflow Exploration

### LINKED ISSUE #10291: Performance: Look into reducing layout thrashing in canvas/layout.js
Seems to be some layout thrashing caused by this effect when you drag to move an element:
https://github.com/google/web-stories-wp/blob/eebdc23de672bedea0bcdbd5e35d4a4f4d66400b/packages/story-editor/src/components/canvas/layout.js#L411-L417

Goal of this ticket is primarily around discovery here. See if there's a possible approach that avoids this and document solution.

maxyinger:
## Context
Part of performance audit. Was exploring reducing the DOM rendering speed for updating an element position:

Screen Shot 2022-01-27 at 12 46 26 PM

## Summary
Ultimately this ended up unfruitful. I tried a couple techniques:

**Reducing code causing the forced reflow**
This was the first attempt to just remove the forced reflow all together. Initially I removed the unnecessary effect run by not creating a new ref on every render of `PageArea` . All though this removed this part of our codebase invoking a forced reflow, the way forced reflows work, if any of these properties get called on an element after a style recalculation is scheduled, the browser forces the synchronous reflow, so you see no results unless you remove *all* reads of these properties:
https://gist.github.com/paulirish/5d52fb081b3570c81e3a

This ultimately doesn't work because the internal moveable library calls `target.offsetWidth` & `target.offsetHeight` internally on dragEnd regardless of if we make one of these reads in our internal code or not. This lead me to the next approach.

**Reducing effected elements from the reflow**
Since the forced reflow seemed unavoidable, the next approach was to reduce the scope of its impact. Currently it effects all elements in the DOM:
Screen Shot 2022-01-27 at 1 39 12 PM

I attempted using the [css contain property](https://developer.mozilla.org/en-US/docs/Web/CSS/contain) on a common ancestor around where I imagined most of the DOM mutations and reads were coming from to indicate that the areas on the canvas were isolated from the rest of the DOM and the browser could limit the scope of the style and layout calculations to those elements. Primarily focused on adding the property to this element here:
https://github.com/GoogleForCreators/web-stories-wp/blob/6a6a1f072d0266c03476c9bb3587720e2f26057c/packages/story-editor/src/components/canvas/canvasLayout.js#L46-L55

but it didn't seem to limit the scope of the layout calculation at all. Sprinkled it in a couple other areas just to see if I could cull the effected scope at all, but had no success.

**Current Status**
Ultimately it's hard to pin point exactly which areas are causing the browser to opt out of the `contain` element and make the scope of the reflows effect every element on the page. The tools the browser provides us to identify this issue are relatively opaque to the best of my knowledge. 

For now, I'm going to put a pin in this and go back to focusing on reducing unnecessary re-renders. Right now the two lower hanging fruits are:
- Stopping the design panel from re-rendering all it's children every time theres an update to the currentPage
- limiting expensive calls in `useApplyStyles` if possible.

After that, it's probably wise to start looking into performance of the other library panels as I know different tabs can add performance impacts to these basic actions. There really shouldn't be variation of perf of updating an element property depending on which library tab you have open.

Hopefully after reducing re-renders some more, it clears up the waters a little but to circle back to this issue and gives us some more transparency as to the root cause.

## Relevant Technical Choices
Just lots of reading. relevant articles:
- https://developers.google.com/web/fundamentals/performance/rendering/avoid-large-complex-layouts-and-layout-thrashing
- https://developers.google.com/web/updates/2016/06/css-containment#layout_contain_layout
- https://css-tricks.com/almanac/properties/c/contain/
- https://mattandre.ws/2014/05/really-fixing-layout-thrashing/
- https://gist.github.com/paulirish/5d52fb081b3570c81e3a

## User-facing changes
NA

## Testing Instructions

1.

## Checklist

---

Fixes #10291

### DIFF
--- a/packages/story-editor/src/components/canvas/canvasLayout.js
+++ b/packages/story-editor/src/components/canvas/canvasLayout.js
@@ -52,7 +52,6 @@ const Background = styled.section.attrs({
   height: 100%;
   position: relative;
   user-select: none;
-  contain: strict;
 `;
 
 function CanvasLayout({ header, footer }) {
--- a/packages/story-editor/src/components/canvas/canvasLayout.js
+++ b/packages/story-editor/src/components/canvas/canvasLayout.js
@@ -52,6 +52,7 @@ const Background = styled.section.attrs({
   height: 100%;
   position: relative;
   user-select: none;
+  contain: strict;
 `;
 
 function CanvasLayout({ header, footer }) {
--- a/packages/story-editor/src/components/canvas/layout.js
+++ b/packages/story-editor/src/components/canvas/layout.js
@@ -367,7 +367,7 @@ function useLayoutParamsCssVars() {
 const PageArea = forwardRef(function PageArea(
   {
     children,
-    fullbleedRef = createRef(),
+    fullbleedRef: _fullbleedRef = null,
     fullBleedContainerLabel = __('Fullbleed area', 'web-stories'),
     overlay = [],
     background,
@@ -381,6 +381,8 @@ const PageArea = forwardRef(function PageArea(
   },
   ref
 ) {
+  const internalFullblledRef = useRef();
+  const fullbleedRef = _fullbleedRef || internalFullblledRef;
   const {
     hasVerticalOverflow,
     hasHorizontalOverflow,


samwhale (approved):
Bummer when you diagnose a problem but you can't fix it due to the code being in a 3rd party lib (moveable)

Very interesting read man. I'm sure we'll see some good performance gains when we deal with the design panel and the `useRichTextFormatting` hook. Very thorough write up 👍

BrookeGraham (approved):
Thanks for the additional reading content. It is super helpful.

swissspidy:
> This ultimately doesn't work because the internal moveable library calls `target.offsetWidth` & `target.offsetHeight` internally on dragEnd regardless of if we make one of these reads in our internal code or not.

Do you think it's possible to make some upstream contribution to that library to help address this?

BrittanyIRL (approved):
Thank you for walking us all through your process of exploration on this, that was really well put and makes sense. Like Samuel said, bummer when you can diagnose but not fix.  

Pascal brings up an interesting question about looking into upstream contribution of movable, wonder if anyone else has run into this with that.

swissspidy:
> Pascal brings up an interesting question about looking into upstream contribution of movable, wonder if anyone else has run into this with that.

The maintainer seems pretty welcoming towards new contributors.

If there is something worth contributing to that would result in some performance wins, that'd be awesome. Might wanna consider it :-)

maxyinger:
> Do you think it's possible to make some upstream contribution to that library to help address this?

Yea I can create a ticket for it. Not sure if there'll be a fruitful solution, but worth some exploration

https://app.zenhub.com/workspaces/web-stories-5e94d3aced449034e1e9b226/issues/googleforcreators/web-stories-wp/10448

PR #10398, opened 2022-01-28
Packages: Create new transform package

mohdsayed:
## Context

See https://github.com/GoogleForCreators/web-stories-wp/issues/4737

## Summary

Create a new `transform` package from `story-editor/src/components/transform`.

## Testing Instructions

- [x] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1.

## Checklist

- [ ] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [ ] I have verified accessibility to the best of my abilities ([docs](https://github.com/googleforcreators/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [ ] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [x] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/googleforcreators/web-stories-wp/tree/main/docs#testing))
- [x] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Partially addresses [#456](https://github.com/GoogleForCreators/web-stories-wp/issues/4737)

PR #10399, opened 2022-01-28
Packages: Create new moveable package

mohdsayed:
## Context

See https://github.com/GoogleForCreators/web-stories-wp/issues/4737

## Summary

Create a new `moveable` package from `story-editor/src/components/moveable`.

Also move `story-editor/src/components/overlay` to `moveable` package.

## Testing Instructions

1.

## Checklist

---

Partially addresses [#456](https://github.com/GoogleForCreators/web-stories-wp/issues/4737)


packages/moveable/package.json
```diff
@@ -0,0 +1,39 @@
+{
```
swissspidy, line 1: Same issue here with tabs vs spaces as in #10398

packages/moveable/package.json
```diff
@@ -0,0 +1,39 @@
...
+	"author": {
+		"name": "Google",
+		"url": "https://creators.google/"
+	},
+	"license": "Apache-2.0",
+	"keywords": [
+		"web stories"
+	],
+	"homepage": "https://github.com/googleforcreators/web-stories-wp/",
+	"repository": "git+https://github.com/GoogleForCreators/web-stories-wp.git",
```
swissspidy:
```suggestion
	"repository": {
    "type": "git",
    "url": "https://github.com/GoogleForCreators/web-stories-wp.git",
    "directory": "packages/moveable"
  },
```

packages/moveable/package.json
```diff
@@ -0,0 +1,39 @@
...
+	},
+	"dependencies": {
+		"@googleforcreators/design-system": "*",
+		"@googleforcreators/react": "*",
+		"prop-types": "^15.7.2",
+		"react-moveable": "^0.30.3",
+		"styled-components": "^5.3.3"
+	},
+	"devDependencies": {
+		"prop-types": "^15.7.2"
```
swissspidy: Already in dependencies above

packages/moveable/src/overlay/index.js
```diff
@@ -19,11 +19,11 @@
  */
 import styled from 'styled-components';
 import { forwardRef, useContext, createPortal } from '@googleforcreators/react';
+import { pointerEventsCss } from '@googleforcreators/design-system';
```
swissspidy: At first glance it looks like this import and the SVG import in`moveStyle.js` are the only places where this now depends on the design system.

I wonder if we could/should inline those usages so we could avoid this dependency? And if that is any useful?

swissspidy (approved):
At first glance looks good; just one question regarding tweaking dependencies

packages/moveable/src/overlay/index.js
mohdsayed: That's a good idea to remove design-system dependency and move one util back to story-editor

swissspidy: Looks like that made a difference of another 700b, so the bundle size is now reduced by 1.4kB in total 👍

packages/transform/README.md
```diff
@@ -0,0 +1,3 @@
+# Transform
+
+Utility package for element transformations.
```
miina: Are two PR-s combined here? The transformation package was in a separate PR 🤔

swissspidy: It was, yes. But because I merged that one, GitHub changed the base branch automatically, causing the changes from that branch to still be included here.

Simply updating the base branch / rebasing should fix this.

packages/story-editor/src/elements/media/scalePanel.js
```diff
@@ -22,10 +22,11 @@ import styled from 'styled-components';
 import { _x } from '@googleforcreators/i18n';
 import { Slider } from '@googleforcreators/design-system';
 import { BG_MIN_SCALE, BG_MAX_SCALE } from '@googleforcreators/animation';
+import { InOverlay } from '@googleforcreators/moveable';
+
 /**
  * Internal dependencies
  */
-import InOverlay from '../../components/overlay';
```
miina, line 28: This is one place where the overlay doesn't seem to be related to Moveable -- is it still ok to move everything related to Overlay to the moveable packages, even though it's used in other cases, too?

mohdsayed: `InOverlay` is being used in three places in `story-editor`. Did you mean we should move the entire `ScalePanel`, `LibraryMoveable` and `SelectionCanvas` components to `movable` package?

`LibraryMoveable` btw, has a lot of internal dependencies to deal with.

https://github.com/GoogleForCreators/web-stories-wp/blob/a354b43c28b49b5b5cb83fd465c618d6278ab052/packages/story-editor/src/components/library/panes/shared/libraryMoveable.js#L26-L41

miina: > Did you mean we should move the entire ScalePanel, LibraryMoveable and SelectionCanvas components to movable package?

I was just wondering if it makes sense to move the whole overlay part to Moveable package at all even though it's not directly for Moveable only (as opposed to keeping it in the editor package). But I guess that's the best option here.

miina (approved):

PR #10400, opened 2022-01-28
CI: Update PHP unit tests config

### LINKED ISSUE #10149: CI: Run tests against PHP 7.2+
## Task Description

We want to stop running tests against 7.0 and instead run tests against 7.2+

That means changing these two lines:

https://github.com/google/web-stories-wp/blob/709a2a87ab4c5787d0869e8b29f8ca0e394687cc/.github/workflows/tests-unit-php.yml#L55

https://github.com/google/web-stories-wp/blob/709a2a87ab4c5787d0869e8b29f8ca0e394687cc/.github/workflows/tests-unit-php.yml#L68

We can then also use PHPUnit 8 instead of PHPUnit 7.5 when running tests against 7.2:

https://github.com/google/web-stories-wp/blob/709a2a87ab4c5787d0869e8b29f8ca0e394687cc/.github/workflows/tests-unit-php.yml#L119-L131

swissspidy:
## Summary

- Test against WP 5.9
- Use PHP 7.2+ only
- Update logic for choosing PHPUnit version

## User-facing changes

N/A

## Testing Instructions

1.

## Checklist

---

Fixes #10149

### DIFF
--- a/.github/workflows/tests-unit-php.yml
+++ b/.github/workflows/tests-unit-php.yml
@@ -128,10 +128,15 @@ jobs:
       - name: Update PHPUnit
         run: |
           if [[ $WP_VERSION == "5.5" && $PHP_VERSION == "7.2" ]]; then
+            echo "Installing PHPUnit 7.5.x"
             composer require --ignore-platform-reqs --no-interaction --no-scripts phpunit/phpunit:^7.5 --with-dependencies
           elif [[ $PHP_VERSION == "7.2" ]]; then
+            echo "Installing PHPUnit 8.5.x"
             composer require --ignore-platform-reqs --no-interaction --no-scripts phpunit/phpunit:^8.5 --with-dependencies
+            echo "Downgrading phpunit/php-token-stream because the latest version requires PHP 7.3"
+            composer require --ignore-platform-reqs --no-interaction --no-scripts phpunit/php-token-stream:^3.1.3 --with-dependencies
           elif [[ $WP_VERSION == "latest" || $WP_VERSION == "trunk" || $PHP_VERSION == "7.4" ]]; then
+            echo "Installing latest version of PHPUnit"
             composer update --ignore-platform-reqs --no-interaction --no-scripts yoast/phpunit-polyfills --with-dependencies
           fi
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests-unit-php.yml
```diff
@@ -116,15 +115,29 @@ jobs:
...
+            echo "Installing PHPUnit 7.5.x"
             composer require --ignore-platform-reqs --no-interaction --no-scripts phpunit/phpunit:^7.5 --with-dependencies
+          elif [[ $PHP_VERSION == "7.2" ]]; then
+            echo "Installing PHPUnit 8.5.x"
+            composer require --ignore-platform-reqs --no-interaction --no-scripts phpunit/phpunit:^8.5 --with-dependencies
+            echo "Downgrading phpunit/php-token-stream because the latest version requires PHP 7.3"
+            composer require --ignore-platform-reqs --no-interaction --no-scripts phpunit/php-token-stream:^3.1.3 --with-dependencies
+          elif [[ $WP_VERSION == "latest" || $WP_VERSION == "trunk" || $PHP_VERSION == "7.4" ]]; then
+            echo "Installing latest version of PHPUnit"
+            composer update --ignore-platform-reqs --no-interaction --no-scripts yoast/phpunit-polyfills --with-dependencies
```
spacedmonkey, line 140: Can you explain what is happening here?

swissspidy: The goal is to install the latest possible PHPUnit version for a given WP / PHP version (see table + link above).

Since `yoast/phpunit-polyfills` has [PHPUnit as its dependency](https://github.com/Yoast/PHPUnit-Polyfills/blob/5cd00434959285ac65b90071d88345d514c6b06a/composer.json#L27), this line here simply means "Update to whatever is the latest PHPUnit supported by yoast/phpunit-polyfills"

This is currently PHPUnit 9, the latest version.

Since this line is only called for WP latest / WP trunk / PHP 7.4, which all support PHPUnit 9, this means these tests are now running with PHPUnit 9.

You can check this by going through the GitHub action log to see which version is being used.

Here's an example: https://github.com/GoogleForCreators/web-stories-wp/runs/4980454198?check_suite_focus=true#step:15:10

Note that this line and this logic already existed before, I just moved it around.

spacedmonkey (approved):

swissspidy:
Waiting to merge until after the 1.17.0 release

Issue #10115, opened 2022-01-06
UI Simplification: Publish Story Modal


miina:
## Feature Description
See [Figma](https://www.figma.com/file/ZrO763yvMWJjYeAPsRnNfW/_Editor---Sprint-Feature-Level-Designs_2022?node-id=57%3A14792)

- [x] When clicking on "Publish", display a modal with various relevant inputs

Breaking this up into chunks to make review cleaner, it might actually sustain more than 5 points depending on implementation and shared code. 

- [x] New modal component for publish story (new header and different spacing patterns so let's not reuse the standard dialog in the design system) 
- [x] Publish button to be disabled if fields in second column are empty (new) 
- [x] Unit tests 
- [x] Karma tests
- [x] New experiment 
- [x] Make inputs in second column (story title and description, password protected requires a password) mandatory 
- [x] Can we reuse the inspector panel contents? How is the best way to share this across this implementation as well as the tab structure? 
- [x] Pull in functionality for saving and updating story title 
- [x] Pull in functionality for saving and updating story description  _separate_ from document panel (where it lives now)
- [x] Pull in functionality for saving and updating status and visibility of story _separate_ from document panel (where it lives now)
- [x] Close story details modal and open checklist when checklist button is clicked 
- [x] Cover preview is brand new functionality - we're not showing a cover preview at all right now. Can we pull this in from AMP, are we doing it ourselves? + Update story cover functionality (when user clicks pencil icon does the story details modal close and open media picker? Does Story Details modal reopen after new image is selected or cancelled?)
- [x] Replace existing modal that opens when first publish occurs with this new one. No more 'are you sure you want to publish' with checklist flags, this new one will suffice.
- [x] e2e tests for functionality at different roles

## Additional Context
<img width="500" alt="Screenshot 2022-01-06 at 13 59 25" src="https://user-images.githubusercontent.com/3294597/148379890-feac3eea-f518-4133-a981-10022a67a60f.png">

miina:
To be defined:
- Will this modal be always opened or only if there are checklist issues?
- Details of the fields that are displayed in the modal. Will only fields with issues be displayed? Or the full Document panel?
- What happens when clicking on "Checklist", will the modal close and checklist open? Or will this replace the checklist?
- How does this influence any currently existing confirmation dialogues we have? For example when changing a Draft Story status to Private, this appears:
[...]

PR #10403, opened 2022-01-28
New Story Details Modal: Mandatory Field Functionality + UI Updates

BrittanyIRL:
## Context

Part of bigger feature for new story details modal, splitting up into chunks for faster review and QA where needed. 

## Summary

Wires up `useStory` to update story title and description, following same logic as existing excerpt and title updates in existing panels/inputs. 
Adds a 'required' label to the modal
More general structure. The modal still isn't connected to the editor yet. I'll do that next.  

## Relevant Technical Choices

- New label shared component for modal to distinguish between required and optional fields. Amy's good with the asterisk and tooltip. 
- Some constants.
- Trying to limit data getting passed around.
- New util for shared story updates - made one file for them for now, if this gets long or needs to be more specific i'll do that later. No use in over engineering from the start. We're going to have some overlapping functionality with these story updates and form components. There's _slight_ differences that made me want to just pull duplicative logic into utils to share (like the slug update) rather than try to force 2 different displays with extra logic in places (take the 'highlight' structure used in the existing panels). I think this keeps code separate where it makes sense and shared where the logic kicks in. I used the story title + slug on blur as an initial proof of concept for this patterning. 
- Initial unit tests set up - will do Karma set up with experiment set up next. 

## To-do

(See #10115 for full list) 

## User-facing changes

- Pulls the slug update functionality into a util. This impacts the story title on blur, which uses the same logic so it needs to be tested for regression. Testing instructions listed below are specifically for this. 

## Testing Instructions

**Slug Regression Testing**

The permalink in the document panel updates based on the story title IF there is no custom permalink set before updating the story title. So, with that in mind we need to just make sure that permalink updates are working as expected (it _should_, all I did was pull it out into a util and update the currentSlug equality check to turn it into a number to check properly against the current story id as a fallback). 
1. If a story is published without a story title or custom setting a permalink the permalink will be the story id 
2. If a story title is given, when user clicks or tabs out of the story title input the permalink is updated with a slug version of the title ('hold on to your butts!' turns into 'hold-on-to-your-butts') 
3. If the permalink is updated on it's own ('jurassic-park-rules') and then the story title is updated, the permalink remains as it was 
4. If a permalink is removed and then the story title is updated or the story is published, the permalink is a slug based on the title or if no title (shame on you for not titling your story) it'll be the story id. 

Otherwise, you can run storybook locally. Now the story details modal looks like this: 

Screen Shot 2022-01-28 at 2 56 47 PM

1.

## Checklist

---

Partially addresses #10115


packages/story-editor/src/utils/storyUpdates.js
```diff
@@ -0,0 +1,47 @@
...
+ * @param {Function} args.updateStory The callback to useStory's action for updating the story with new slug.
+ * @return {void}
+ */
+export const updateSlug = ({
+  currentSlug,
+  currentTitle,
+  storyId,
+  updateStory,
+}) => {
+  if (!currentSlug || parseInt(currentSlug) === storyId) {
```
BrittanyIRL: This line is different from the original check to update the slug that was in the `title` onBlur that I pulled out into a util. When I was testing it I noticed that `currentSlug === storyId` is never going to be true without `parseInt` because the slug's a string and storyId is a number. This is why I want to specifically do regression testing on the slug updates. 

@miina is there any history here with this check that I'm missing or context that makes this a bad update?

miina: Perhaps the check without `parseInt` was for when the story was still new and hadn't been saved at all, and then changing the slug, could it be? Or perhaps some non-standard permalink setting?
I'll double check tomorrow if that's not the case (finished the day already).

BrittanyIRL: If the story is new and hasn't been saved at all then the slug's empty. 

The story id is used as a fallback if there's no story title. So, for instance, if I focus the story title and then tab out of it, there's still no viable story title in which to make a slug out of so the story id is used as a fallback. 

Where the parseInt comes into play is if the current slug is equal to the story id on a story title update then we should check and see if there's an update, otherwise the slug isn't updated every time the story title changes (which is consistent with staging). I honestly think this shoulda been a parseInt all along but it's such an edge case it wasn't really noticed. I did set up unit tests now so we can be confident in when this is updating and how.

BrookeGraham (approved):
Curious if Mina has any more insight on the slug. It all sounds (and looks) good to me. Tests look great too!

packages/story-editor/src/utils/storyUpdates.js
miina: Ah, okay, got it. The only place where I can see this could now cause issues is if the slug starts with the story ID.
E.g. the story ID is 3 and the slug is `3-things-you-can-do`, then blurring the title would overwrite that slug, if I understand correctly.

swissspidy: `Number()` seems preferable in that case over `parseInt`

packages/design-system/src/components/typography/headline/index.js
```diff
@@ -46,7 +46,7 @@ export const Headline = styled.h1`
 `;
 
 Headline.propTypes = {
-  as: PropTypes.oneOf(['h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'a']),
+  as: PropTypes.oneOf(['h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'a', 'label']),
```
samwhale, line 49: 💭  Doesn't the `label` already exist as an option of `<Text>`?

https://github.com/GoogleForCreators/web-stories-wp/blob/d16f28fb45d808bf5abef2a5ae2c69feb5a63f66/packages/design-system/src/components/typography/text/index.js#L88-L94

Does the headline need this option?

packages/story-editor/src/components/publishModal/header/index.js
```diff
@@ -48,31 +49,41 @@ const PublishButton = styled(Button)`
...
-      </PublishButton>
+      </Button>
       <Headline
         as="h2"
         size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.XXX_SMALL}
       >
         {__('Story Details', 'web-stories')}
       </Headline>
-      <Button
+      <PublishButton
```
samwhale, line 70: 👍

packages/story-editor/src/components/publishModal/mainContent/formLabel.js
```diff
@@ -0,0 +1,62 @@
...
+    </Tooltip>
+  ) : (
+    <Headline
+      as="label"
+      size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.XX_SMALL}
+      htmlFor={htmlFor}
+    >
+      {copy}
+    </Headline>
+  );
```
samwhale, line 54: 💭  what are your thoughts on adding this directly to the `TextArea` in the design system? The `TextArea` component has an option to render a label that matches the designs but no option to render the tooltip on the label

Wonder if that would be good to add 🤔

packages/story-editor/src/components/publishModal/test/mainContent.js
```diff
@@ -0,0 +1,96 @@
...
+
+  const mockInputValues = {
+    [INPUT_KEYS.EXCERPT]:
+      'A voyage to the great beyond, where none have gone and going back is impossible.',
+    [INPUT_KEYS.TITLE]: "David Bowman's Odyssey",
+    [INPUT_KEYS.SLUG]: '2001-space-odyssey',
+  };
+  afterEach(() => {
+    mockHandleUpdateStoryInfo.mockReset();
+    mockHandleUpdateSlug.mockReset();
```
samwhale: nit: could write `jest.resetAllMocks()` here if you want. Would also be easier if more mocks are added in the future

samwhale: Also - doesn't seem like these mocks are returning anything here. We could use `clearAllMocks` instead: 

https://jestjs.io/docs/mock-function-api#mockfnmockreset
image

packages/story-editor/src/utils/test/storyUpdates.js
```diff
@@ -0,0 +1,79 @@
...
+ */
+import { updateSlug } from '../storyUpdates';
+
+describe('storyUpdates', () => {
+  describe('updateSlug', () => {
+    const updateStory = jest.fn();
+
+    afterEach(() => {
+      updateStory.mockReset();
+    });
```
samwhale, line 27: same here!

samwhale (approved):
Great tests, very straight forward. Modal is looking clean too.

packages/story-editor/src/components/publishModal/mainContent/formLabel.js
BrittanyIRL: Since this is a 1 off pattern (as far as I'm aware), i think keeping it out of the design system for now is better.

packages/design-system/src/components/typography/headline/index.js
BrittanyIRL: 😢 design has the line spacing matching the `Headline` instead of the `Text` so i added a label option there.  I should check with Amy though, maybe it was an accident or not important.

felipebochehin87:
Tested against https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/, using PR #10403.

1. Published story with both title and permalink empty - story id is set as permalink
2. Added a title and pressed tab - permalink is set to the same as the title
3. Changed permalink and changed title - permalink remains as it was
4. Removed permalink and updated story title - permalink is the story title
5. Removed permalink and story title - permalink is story id


https://user-images.githubusercontent.com/96055551/152569606-9b6a2fd3-1331-46bc-9524-4ad3d9873f52.mp4

PR #10404, opened 2022-01-29
Editor: First stage of floating menu bar

### LINKED ISSUE #10112: UI Simplification: Base structure/architecture for Floating element menu bar
## Feature Description
See [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=11394%3A457049)

Full Todo list:

- [x] Add this behind a Feature flag to release together with #10111 
- [x] Add a floating menu bar that is opened by default.
- [x] When closing the menu bar, it can be opened again by re-selecting the element, it will always be open for the element that gets selected.
- [ ] Add items to the menu bar based on Figma -- these will not be links as in quick-action menu but can be directly edited from the bar.
- [ ] Note: Border and Corner inputs are only displayed when locked. If unlocked, don't display these in the menu.
- [ ] When clicking on "More" button, it will open and highlight the Design Panel
- [ ] Use flyouts as in Figma
- [ ] Replace button for media edit bar -- replaces the media, keeping other props the same as possible.
- [ ] Remove duplicates from the quick-action menu.

Duplicated to remove:
- From text element, remove: Change Color; Edit Text
- From shape element, remove: Change Color
- From video, remove: Trim

The remaining work has now been moved to a number of child issues:

- #10624
- #10625
- #10626
- #10627
- #10628
- #10629
- #10630
- #10631
- #10632
- #10633
- #10634
- #10635
- #10636
- #10638
- #10639

barklund:
## Context

This PR is the first partial PR to implement the floating menu bar.

Included in this PR:

* New feature flag
* Floating menu that aligns below the current selection whatever that may be

## Relevant Technical Choices

* I decided to listen directly to the mutation events of the moveable object rather than all the ways an object can be moved because they are plenty. I actually started out by listening to some of those events but realized that there were way more than expected. This seemed a lot faster! The current selection can update in many different ways:
  * Selecting and deselecting elements
  * Moving selected elements using the keyboard
  * Resizing, moving, or rotating the current selection using the mouse
  * Changing properties directly in the design panel
  * Resizing the browser window
  * Zooming the canvas using the zoom dropdown or pinch-zoom
  * Scrolling a zoomed canvas
* Also, the clamping of the menu inside the workspace is done in CSS! That's an incredibly efficient way to do it, so very happy with that!
* To get the full workspace height, I had to do some hacks in the layout provider, but that wasn't a big deal.

## To-do

* This does not work in text-edit mode and it probably should. It's most likely a small fix to also call the onMoveableMount callback in `editElement`, if the element is a text element (and not if media!), but left that for future optimization.
* [x] ~This has not been tested in RTL and the semantics for the menu are not proper either.~ It does work in RTL!

## User-facing changes

_Note: These aren't actually user-facing, because it's behind a feature flag._

The menu isn't pretty (or is it?), but it's correctly placed!

| Manipulation | View |
|-|-|
| Single element | menu-single-element |
| Multi-selection | menu-multi-select |
| Window resize | menu-window-resize |
| Workspace clamping | menu-workspace-clamp |
| Zoom and scroll | menu-zoom |

Note this weirdness when it comes to the multi-selection. When rotating, the frame will reset on completed rotation, which causes the menu to also jump. But the menu perfectly follows the selection rectangle, so I feel it's okay:
menu-multi-weird

## Testing Instructions

## Checklist

---

Partially addresses #10112

### DIFF
--- a/includes/Experiments.php
+++ b/includes/Experiments.php
@@ -385,9 +385,9 @@ class Experiments extends Service_Base implements HasRequirements {
 			 * Creation date: 2022-01-27
 			 */
 			[
-				'name'        => 'newDesignPanel',
-				'label'       => __( 'New Design Panel', 'web-stories' ),
-				'description' => __( 'Enable the new design panel integrated in the left-side menu including floating design bar', 'web-stories' ),
+				'name'        => 'floatingMenu',
+				'label'       => __( 'Floating Menu', 'web-stories' ),
+				'description' => __( 'Enable the new floating design menu', 'web-stories' ),
 				'group'       => 'editor',
 			],
 		];
--- a/includes/Experiments.php
+++ b/includes/Experiments.php
@@ -378,6 +378,18 @@ class Experiments extends Service_Base implements HasRequirements {
 				'description' => __( 'Enable adding captions to background audio', 'web-stories' ),
 				'group'       => 'editor',
 			],
+
+			/**
+			 * Author: @barklund
+			 * Issue: #10112
+			 * Creation date: 2022-01-27
+			 */
+			[
+				'name'        => 'newDesignPanel',
+				'label'       => __( 'New Design Panel', 'web-stories' ),
+				'description' => __( 'Enable the new design panel integrated in the left-side menu including floating design bar', 'web-stories' ),
+				'group'       => 'editor',
+			],
 		];
 	}
 
--- a/packages/story-editor/src/components/floatingMenu/layer.js
+++ b/packages/story-editor/src/components/floatingMenu/layer.js
@@ -23,6 +23,7 @@ import { useEffect, useRef, useState } from '@googleforcreators/react';
  * Internal dependencies
  */
 import { useCanvas, useLayout } from '../../app';
+import { FLOATING_MENU_DISTANCE } from '../../constants';
 import FloatingMenu from './menu';
 
 // Note that this has to work for a lot of different types of moveable
@@ -55,6 +56,9 @@ function FloatingMenuLayer() {
   const workspaceSize = useRef();
 
   // Whenever the selection frame (un)mounts, update the reference to moveable
+  // This happens when selection changes between the three possible states: None,
+  // single, multiple. If the selection merely changes inside one of those (so
+  // from single to another single), this function is not invoked.
   useEffect(() => {
     setMoveableMount(() => setMoveable);
     return () => setMoveableMount(null);
@@ -81,7 +85,7 @@ function FloatingMenuLayer() {
       menu.style.display = 'flex';
       const centerX = frameRect.left + frameRect.width / 2;
       menu.style.left = `clamp(0px, ${centerX}px - (var(--width) / 2), ${width}px - var(--width))`;
-      const bottomX = frameRect.top + frameRect.height + 10;
+      const bottomX = frameRect.top + frameRect.height + FLOATING_MENU_DISTANCE;
       menu.style.top = `clamp(0px, ${bottomX}px, ${height}px - var(--height))`;
     };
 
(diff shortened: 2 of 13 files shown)


packages/story-editor/src/components/floatingMenu/layer.js
```diff
@@ -0,0 +1,103 @@
...
+      return null;
+    }
+
+    const updatePosition = () => {
+      const frameRect = moveable.getRect();
+      const { width, height } = workspaceSize.current;
+      menu.style.display = 'flex';
+      const centerX = frameRect.left + frameRect.width / 2;
+      menu.style.left = `clamp(0px, ${centerX}px - (var(--width) / 2), ${width}px - var(--width))`;
+      const bottomX = frameRect.top + frameRect.height + 10;
```
miina: Nit: in finalized version later would be good to use a constant for `10`.

packages/story-editor/src/components/floatingMenu/layer.js
```diff
@@ -0,0 +1,103 @@
...
+    // And update when certain elements' properties update
+    const observer = new MutationObserver(updatePosition);
+    MUTATE_SELECTORS.map((selector) => document.querySelector(selector))
+      .filter(Boolean)
+      .forEach((node) => observer.observe(node, { attributes: true }));
+
+    return () => observer.disconnect();
+  }, [moveable]);
+
+  return <FloatingMenu ref={menuRef} />;
```
miina, line 104: Should the layer be displayed at all if there's no moveable ref?

includes/Experiments.php
```diff
@@ -378,6 +378,18 @@ public function get_experiments(): array {
...
+
+			/**
+			 * Author: @barklund
+			 * Issue: #10112
+			 * Creation date: 2022-01-27
+			 */
+			[
+				'name'        => 'newDesignPanel',
+				'label'       => __( 'New Design Panel', 'web-stories' ),
+				'description' => __( 'Enable the new design panel integrated in the left-side menu including floating design bar', 'web-stories' ),
```
miina: Maybe let's have a flag just for the Floating menu? The main reason for adding this behind a feature flag and merging together with the right side panel change was that UX thought it might not fit otherwise. But if it fits, we could merge it earlier.

The main goal is not to get rid of the right panel before the floating menu. Having floating menu alone is ok though!

miina (approved):
Works great!

> Note this weirdness when it comes to the multi-selection. When rotating, the frame will reset on completed rotation, which causes the menu to also jump. But the menu perfectly follows the selection rectangle, so I feel it's okay:

Yep, seems expected!

Q: Not sure if you checked but could using `ContextMenu` (or `ContextMenuProvider`) from the design system be useful here? I _think_ it might have some useful logic implemented already, such as dismissing menu, some accessibility stuff, and possibly also positioning (not sure if as needed here though). All the other popup menus currently us it (right-click menu, quick-action menu, page menu), might not be applicable here though. Thoughts?

includes/Experiments.php
barklund: Sure, done in 626c3738c.

packages/story-editor/src/components/floatingMenu/layer.js
barklund: Yeah, I'm not happy about hiding/showing the menu rather than simply not rendering it at all, but if not, the menu would show up briefly in the wrong position (and I would need to do a lot of checks on whether `menuRef` was still live).

I will try to rework this as I bring in the actual menu. The idea of using the `ContextMenu` is probably a good choice - I will look into that, thanks!

packages/story-editor/src/components/floatingMenu/layer.js
barklund: Done in af5cb37c6.

PR #10405, opened 2022-01-31
Packages: Create new rich-text package

### LINKED ISSUE #8126: Packages: Move `rich-text` to new structure
## Feature Description

This proposed `rich-text` package is intended to be a small wrapper around `draft-js`.

The dedicated `elements` package provides various APIs to register & retrieve element types.

**References:**

* [Package organization doc](https://docs.google.com/document/d/1ImRe9wYKWGCAZiq3jro5KLpSJK_NMKXHC6FWqs3nwl0/edit#)
* Proof-of-concept implementation: https://github.com/google/web-stories-wp/tree/try/next-level-packages/packages/rich-text/src

**To-do:**

* [ ] Move some of the code from `/assets/src/edit-story/components/richText` to `/packages/rich-text/src/`
  Definitely **not** move `context.js` and `provider.js`
* [ ] Update imports where necessary.  
  Use `@web-stories-wp/...` imports for anything that is already a package, relative imports for everything else.
* [ ] Create `/packages/rich-text/package.json` referencing its dependencies
* [ ] Create `/packages/rich-text/README.md` with short documentation.

### Blockers ⚠️

* The proof-of-concept contains the `MULTIPLE_VALUE` constant, but it doesn't really make sense to live there. Needs a better home.

## Additional Context

---

mohdsayed:
## Context

See https://github.com/GoogleForCreators/web-stories-wp/issues/8126

## Summary

Create a new `rich-text` packages from `story-editor/src/components/richText`.
Keep `karma` folder in `story-editor`.
Move `getValidHTML.js` from `story-editor/src/utils` to the new `rich-text` package utility.

## Testing Instructions

1.

## Checklist

---

Fixes https://github.com/GoogleForCreators/web-stories-wp/issues/8126
Partially addresses [#456](https://github.com/GoogleForCreators/web-stories-wp/issues/4737)


packages/rich-text/package.json
```diff
@@ -0,0 +1,41 @@
+{
+	"name": "@googleforcreators/rich-text",
```
swissspidy: Note the tabs vs spaces here

packages/rich-text/package.json
```diff
@@ -0,0 +1,41 @@
...
+	"author": {
+		"name": "Google",
+		"url": "https://creators.google/"
+	},
+	"license": "Apache-2.0",
+	"keywords": [
+		"web stories"
+	],
+	"homepage": "https://github.com/googleforcreators/web-stories-wp/",
+	"repository": "git+https://github.com/GoogleForCreators/web-stories-wp.git",
```
swissspidy:
```suggestion
	"repository": {
    "type": "git",
    "url": "https://github.com/GoogleForCreators/web-stories-wp.git",
    "directory": "packages/rich-text"
  },

**Review by swissspidy — APPROVED** [review]

File: packages/rich-text/src/customConstants.js [hunk 3]
```diff
@@ -21,3 +21,6 @@ export const WEIGHT = 'CUSTOM-WEIGHT';
 export const COLOR = 'CUSTOM-COLOR';
 export const LETTERSPACING = 'CUSTOM-LETTERSPACING';
 export const UPPERCASE = 'CUSTOM-UPPERCASE';
+
+// @todo May be move it to new elements package?
+export const MULTIPLE_VALUE = '((MULTIPLE))';
```
miina, line 26: Yep, this seems like specific to the usage and not to the rich text itself, not sure where it should live though but perhaps there's a better place in the future.

miina (approved):

packages/rich-text/src/customConstants.js
mohdsayed: I am thinking about its placement in the [next PR ](https://github.com/GoogleForCreators/web-stories-wp/pull/10452) . Do you have any suggestions for where it should live?

miina: It seems like something that's related to the design panel and how the values are displayed/updated from there -- doesn't seem to be something that the elements themselves need to be aware of. Multiple value means that more than one element is selected and the values of a specific property of the element object in the inputs differ -- should the elements package "be aware" of that? 🤔 

But I'm not sure what's exactly the system you're following when pulling code into different packages really and how strictly the separation is being made. If we have to move it out of the `story-editor` package then perhaps `elements` could be a good one, yes, since it will include the attributes per element type and multiple value could be remotely related.

mohdsayed: Yes, `elements` looks like a good place at first glance, will think over it more in the other PR. Thank you!

Issue #4817, opened 2020-10-06
Directly use phpstan-wordpress extension


szepeviktor:
I've contributed static analysis to project when the WordPress extension was not published.
Now it is. And more advanced then ever.

Please consider using it.
https://packagist.org/packages/szepeviktor/phpstan-wordpress

Plus one: depends on php-stubs/wordpress-stubs so you don't have to depend on it

----

## Acceptance Criteria

Ping szepeviktor on the final PR for this

swissspidy:
Pretty sure it was already published back then. We‘ved used the extension in the past on CI. But that was on CI only, with ignoring platform reqs option.

Due to different PHP version requirements we‘re using phpstan via phar only, which I don‘t think is possible with the extension.

Hence the current approach with copying some of the files, as inspired by the amp-wp repo.

szepeviktor:
Would you accept a PR that installs these tools in CI?
As that job of the CI run on PHP 7.4.

swissspidy:
That's what we had before, see #4540.

It basically did `composer require --no-interaction --dev szepeviktor/phpstan-wordpress --ignore-platform-reqs`

Ideally we he have a setup that works identically both locally and on CI.

szepeviktor:
I see!

How do you mean "both locally and on CI"? PHPStan does not start on PHP < 7.2

szepeviktor:
Ah! You mean locally with adequate version of PHP.

szepeviktor:
These conditions are very tough!
I come back with a solution...

swissspidy:
Correct :) That's why we have the current solution, and I think it works fine at the moment.

Once we bump the PHP requirement we can easily remove this workaround.

szepeviktor:
I am thinking about a Composer script - where all commands should live.

```json
"script: {
  "analyze": [
    "composer info --path szepeviktor/phpstan-wordpress || composer require --dev szepeviktor/phpstan-wordpress",
    "vendor/bin/phpstan analyze"
  ]
}
```

PR #10411, opened 2022-01-31
Composer: bump PHP requirement to 7.2+

### LINKED ISSUE #10150: Composer: bump PHP requirement to 7.2+
## Task Description

Change PHP requirement to `^7.2 || 8.0`:

https://github.com/google/web-stories-wp/blob/709a2a87ab4c5787d0869e8b29f8ca0e394687cc/composer.json#L8

Change `platform` config to `7.2.5` (seems reasonable and the same as Composer). But if we can use newer versions of some dependencies by doing so, we can also try `7.2.24` (which seems to be the lowest 7.2 version our users seem to be running):

https://github.com/google/web-stories-wp/blob/709a2a87ab4c5787d0869e8b29f8ca0e394687cc/composer.json#L72-L74

After that, update dependencies using `composer update`.

Update `Requires PHP` in the main plugin file:

https://github.com/google/web-stories-wp/blob/709a2a87ab4c5787d0869e8b29f8ca0e394687cc/web-stories.php#L17

Update `WEBSTORIES_MINIMUM_PHP_VERSION` in the main plugin file to `7.0.5` or whatever we decided to use in `composer.json`:

https://github.com/google/web-stories-wp/blob/709a2a87ab4c5787d0869e8b29f8ca0e394687cc/web-stories.php#L49

spacedmonkey:
## Summary

Bump PHP requirement to 7.2+

## Testing Instructions

1.

## Checklist

---

Fixes #10150


composer.json
```diff
@@ -27,6 +27,7 @@
     "php-stubs/wordpress-stubs": "^v5.9.0",
     "phpcompatibility/phpcompatibility-wp": "^2.1",
     "phpmd/phpmd": "^2.9",
+    "phpstan/phpstan": "^1.4",
```
swissspidy, line 30: Note that here's a dedicated ticket #4817 to change the PHPStan dependency.

There, we don't need to include `phpstan/phpstan` as we only need to require `php-stubs/wordpress-stubs`

I would like to keep this change separate.

spacedmonkey: I don't see the harm in doing this here. If anything it will speed composer install up and making upgrading easier.

swissspidy (approved):

Issue #5326, opened 2020-11-17
Semantic HTML tags for headings


swissspidy:
## Summary

After reading through https://wordpress.org/support/topic/add-the-code-of-the-hn-tags/ I realized that we don't seem to have any ticket or documentation about the expectations around semantic HTML output for stories. There's only #1610 about semantic tags for inline formatting.

For several reasons (SEO, accessibility, to name the most important ones) it is advisable to use semantic heading tags like `<h1>` or `<h2>`.

At the moment it is not possible to use any of those.

We need to define if and how this should be possible, and document this accordingly.

If we _do_ want to support this

Should we allow multiple `<h1>` tags in a story (but only 1 per page)? Demos like https://amp.dev/documentation/guides-and-tutorials/learn/webstory_technical_details/ seem to indicate so.

How is this reflected in the pre-publish checklist?

## Additional Context

In the legacy story editor we tried to be smart, detecting whether an element was for example positioned high up on the page and with large font size, and automatically turning it into an `<h1>` on the frontend. The user did not have to worry about it (but was able to override it)

<img width="316" alt="67967022-df536800-fbd2-11e9-9bdb-a1f32dc3c6ff" src="https://user-images.githubusercontent.com/841956/151340514-bff7f3b1-bd10-4138-afb1-c83572e5307b.png">

## Acceptance Criteria

I am imagining this to be similar to what is shown in the screenshot above (though adding UI is P2).

1. Multiple H1 in a story are OK
2. Only one H1 allowed on a story _page_
3. Devise algorithm for automatically setting the heading level based on the element's appearance
3. When using one of the existing text sets or text presets, these should automatically have the right heading level
  e.g. when I insert the "Heading 1" preset it should be an H1 ideally, due to its appearance (because the font size is so large)
4. If the above is not feasible or some text sets are too complex, we could force the element type for some text elements, e.g. always make an element H1, regardless of its appearance

### Does this epic have any performance impact?

There shouldn't be, as the algorithm for automatically setting the heading level (p, h1, h2) is fairly trivial

### Does this epic have telemetry?

Not needed

swissspidy:
Pinging @o-fernandez @choumx for consideration

dreamofabear:
Seems reasonable to use h1, h2, h3 etc. for text presets and text sets. I don't think we need to bother with quantity restrictions.

IMO this is a P3/P4 though.

LuckynaSan:
Updating this request with related support topics:
[...]

PR #10413, opened 2022-01-31
Add initial semantic headings logic

### LINKED ISSUE #10394: Semantic Headings: Implement initial algorithm for choosing heading level
## Feature Description

See #5326 for background and acceptance criteria.

The initial (feature-flagged) algorithm should devise the heading level for text elements on a page based on its appearance (dimensions, font size, boldness, etc.) and change the output accordingly.

swissspidy:
## Context

See #5326 for full context.

We want to automatically change the semantic HTML tag being used for text elements, depending on an element's appearance.

## Summary

This change implements the initial logic for using semantic heading tags, based on font size as the sole criterium.

Large text elements will be wrapped in an H1, H2, H3 or paragraph depending on the font size.

There is only ever one H1 on a single story page.

That means multiple H1 tags within the whole story are possible and OK.

## User-facing changes

No visual changes

## Testing Instructions

1. Create a new story
2. In the left-hand panel, go to the "Text" tab and insert the text presets (Title 1, Title 2, etc.)
3. Preview the story
4. The HTML tags for the inserted text elements should correspond to their title, i.e. Title 1 should be wrapped in `<h1>`
5. Repeat the same for the _Cover_ text sets, where the title should also be wrapped in `<h1>`
6. Confirm that there is only ever a single `<h1>` element on an individual story page, even after inserting multiple presets or text sets

No

No

No

## Checklist

---

Fixes #10394


packages/story-editor/src/app/story/actions/useAutoSave.js
```diff
@@ -39,7 +39,7 @@ function useAutoSave({ storyId, pages, story }) {
   const {
     actions: { autoSaveById },
   } = useAPI();
-  const { metadata } = useConfig();
+  const { metadata, flags } = useConfig();
```
spacedmonkey, line 42: Why use flags like this? Why can't we use the flagged provider?

spacedmonkey (approved):

packages/story-editor/src/app/story/actions/useAutoSave.js
swissspidy: Because `getStoryMarkup()` (which generates the output) is just a simple function, not a hook or React component

PR #10414, opened 2022-01-31
Settings: Improve UX for Stories Archives setting

### LINKED ISSUE #10203: UX improvements for Archive setting
## Feature Description

We regularly receive requests on the support forum for the ability to adjust the default Web Stories archive page (`web-stories`). In #4862, this was discussed and the alternative of allowing users to build out their own custom archive page using Gutenberg blocks was determined to be the best next step.

However, we find in the forums that users don't understand that the Custom option in Web Stories > Settings > Stories Archives will allow them to do this and have the control that they're looking for. I think there might be some confusion on the user side, as our approach is pretty uncommon in my experience with WP plugins. I've seen settings like this before, but they typically generate an automated archive on the selected page that is *not* modifiable; see https://wordpress.org/support/topic/custom-archive-page-is-empty/. Our approach is different.

To make this option clearer to users, I think there are some adjustments that we can make here. The suggestions below should definitely be vetted by our UX team.

## Disabled option
- Change helper text from "**This disables the default archive page.**" to "**Turn off the default archive page at [site URL]/web-stories/". Users will receive your 404 page when they go to the default archive page.**"

## Custom option
- Change "**Custom**" label to "**Create your own**"
- Change helper text from "**Choose a custom page as your stories archive.**" to "**Choose a page to serve as your Web Stories archive. Use the Web Stories block to add stories to this page.**"
- Change helper text that appears under page select from "**Visit archive page at [custom page URL]**" to "**Visit archive page at [custom page URL]. [site URL]/web-stories/ will automatically redirect to this page.**"

## Additional Context

Previous archive customization requests:
- https://wordpress.org/support/topic/editing-gws-archives-page/
- https://wordpress.org/support/topic/web-stories-page-edit/
- https://wordpress.org/support/topic/how-to-customize-web-stories/
- https://wordpress.org/support/topic/archives-stories/
- https://wordpress.org/support/topic/how-do-i-add-h1-heading-to-the-web-stories-archive-page/

spacedmonkey:
## Summary

Update wording of custom archive page. 

 
## User-facing changes

Screenshot 2022-01-31 at 17 13 26
Screenshot 2022-01-31 at 17 13 19
Screenshot 2022-01-31 at 17 13 10

## Testing Instructions

1.

## Checklist

---

Fixes #10203


includes/Admin/Dashboard.php
```diff
@@ -432,6 +432,7 @@ public function get_dashboard_settings(): array {
 			'locale'                  => $this->locale->get_locale_settings(),
 			'newStoryURL'             => $new_story_url,
 			'archiveURL'              => $this->story_post_type->get_archive_link(),
+			'archiveURLRaw'           => $this->story_post_type->get_archive_link( true ),
```
swissspidy: `raw` implies this being an unsanitized/unescaped variable.

Let's rename this to something less ambiguous everywhere, e.g. `defaultArchiveURL`

includes/Post_Type_Base.php
```diff
@@ -294,12 +294,15 @@ public function get_parent_controller(): WP_REST_Controller {
 	 *
 	 * @since 1.14.0
 	 *
+	 * @SuppressWarnings(PHPMD.BooleanArgumentFlag)
```
swissspidy, line 297: This is a legit warning for a legit code small

> A boolean flag argument is a reliable indicator for a violation of the Single Responsibility Principle (SRP). You can fix this problem by extracting the logic in the boolean flag into its own class or method.

A separate `get_default_archive_link` method would be less ambiguous and easier to comprehend.

packages/dashboard/src/karma/fixture.js
```diff
@@ -82,6 +82,7 @@ const defaultConfig = {
   cdnURL: 'https://cdn.example.com/',
   version: '1.0.0-alpha.9',
   archiveURL: 'https://example.com/',
+  archiveURLRaw: 'https://example.com/',
```
swissspidy: Nit: Using different URLs here would make it easier to understand the difference

packages/wp-dashboard/src/components/editorSettings/archive/index.js
```diff
@@ -217,11 +221,21 @@ export default function ArchiveSettings({
                 ),
               }}
             >
-              {sprintf(
-                /* translators: %s: archive url. */
-                __('Visit archive page at <a>%s</a>.', 'web-stories'),
-                archiveURL
-              )}
+              {archivePageId
```
swissspidy:
```suggestion
              {archivePageId && archiveURL !== archiveURLRaw
```

In case the custom page archive URL is the same as the default archive URL.

includes/Post_Type_Base.php
spacedmonkey: But that would mean having to method with basically the same logic.

includes/Admin/Dashboard.php
spacedmonkey: Another idea, was to break this up into object. So 

```
archiveURL: {
     current: '...',
     'default': '...',
}
```

Thoughts?

includes/Post_Type_Base.php
swissspidy: Shared logic could live in a shared private method if needed.

includes/Admin/Dashboard.php
swissspidy: I think two separate properties are good enough, we don't need extra nesting. But I don't feel strongly about it.

includes/Post_Type_Base.php
spacedmonkey: I am not sure how that would look and what the point of it would be. I have updated the PR with some cleaner logic. https://github.com/GoogleForCreators/web-stories-wp/pull/10414/commits/3d17e603c8083d1f074a78b3249f6f63f99472bf

packages/dashboard/src/karma/fixture.js
spacedmonkey: Updated https://github.com/GoogleForCreators/web-stories-wp/pull/10414/commits/2636431014094256655ac3f9c46f1518176932b9

includes/Admin/Dashboard.php
spacedmonkey: Changed

includes/Post_Type_Base.php
swissspidy: Just making the default value `null` instead of `false` (for something that still is a boolean flag) is just way to trick the linter, but the issue still exists.

spacedmonkey: I just don't see the point of having duplicated code here.

spacedmonkey: I made a bash and breaking into two methods. See https://github.com/GoogleForCreators/web-stories-wp/pull/10414/commits/a44023e92f6cca1666e47c1100ce4544933e87de

swissspidy: You could have copied over the tests and the improved PHPDoc as well.

includes/Post_Type_Base.php
```diff
@@ -286,20 +286,52 @@ public function get_parent_controller(): WP_REST_Controller {
...
+	 * @since 1.17.0
+	 *
+	 * @return string|false The post type archive permalink. False if the post type
+	 *                      does not exist or does not have an archive.
+	 */
+	public function get_default_archive_link() {
+		return $this->get_archive_link_data( true );
+	}
+
+
```
swissspidy: Duplicate new lines.

```suggestion

```

includes/Post_Type_Base.php
```diff
@@ -286,20 +286,52 @@ public function get_parent_controller(): WP_REST_Controller {
...
+			return false;
+		}
+
+		return $this->get_archive_link_data( $post_type_obj->has_archive );
+	}
+
+	/**
+	 * Calls get_archive_link_data method to get the value of archive by forcing has_archive to true.
+	 *
+	 * @since 1.17.0
```
swissspidy:
```suggestion
	 * @since 1.18.0
```

Please check out the generally improved PHPDoc I added in #10442

The return description is not accurate and the short description above is not useful either.

includes/Post_Type_Base.php
swissspidy: This code is still not ideal, it's just moving the `true === $post_type_obj->has_archive || $ignore_has_archive` check you had originally to a separate method.

Given that, when comparing this change with your original method, the original one would have been preferable over this.

spacedmonkey: Have lots of lines of duplicated code was my issue with your PR.

What should I do here, revert back to the way it was?

swissspidy: First of all let's add the tests that I've written so that I didn't write them for nothing.

Regarding the implementation, I would have preferred at least discussing my PR to avoid all this back and forth.

I _know_ my suggested code adds _some_ code duplication but it's an acceptable amount of duplicated code in favor of clearer separation of responsibilities.

That said, I don't have a super strong opinion on it though and it's not really worth arguing over, so please just revert your later changes

swissspidy: Can you either revert your `get_archive_link` implementation to what you had originally (with suppressing the boolean flag warning) or implement my suggested changes?
I won't approve the current version.

spacedmonkey: Reverted. But I kept the tests, as they as still valid.

swissspidy (approved):

PR #10427, opened 2022-01-31
Editor: Added floating menu to text while in edit-mode

barklund:
## Context

This is another partial PR for #10112. This time adding the floating menu to the text-edit element.

## Relevant Technical Choices

* Adding the floating menu was easy, but because the coordinate system was shifted in the edit element, I had to reorganize some things. Edit mode still works exactly as before though.

## User-facing changes

You can see that the element is in edit mode, when the squiggly red line is visible below the non-word (because of spellchecker).
text-edit-menu

## Testing Instructions

## Checklist

---

Partially addresses #10112

### DIFF
--- a/packages/story-editor/src/components/canvas/editElement.js
+++ b/packages/story-editor/src/components/canvas/editElement.js
@@ -17,7 +17,7 @@
 /**
  * External dependencies
  */
-import { useState, forwardRef } from '@googleforcreators/react';
+import { memo, useState, forwardRef } from '@googleforcreators/react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import { useUnits } from '@googleforcreators/units';
@@ -39,35 +39,35 @@ const Wrapper = styled.div`
   pointer-events: initial;
 `;
 
-const EditElement = forwardRef(function EditElement(
-  { element, editWrapper, onResize },
-  ref
-) {
-  const { id, type } = element;
-  const { getBox } = useUnits((state) => ({
-    getBox: state.actions.getBox,
-  }));
+const EditElement = memo(
+  forwardRef(function EditElement({ element, editWrapper, onResize }, ref) {
+    const { id, type } = element;
+    const { getBox } = useUnits((state) => ({
+      getBox: state.actions.getBox,
+    }));
 
-  // Needed for elements that can scale in edit mode.
-  const [localProperties, setLocalProperties] = useState(null);
+    // Needed for elements that can scale in edit mode.
+    const [localProperties, setLocalProperties] = useState(null);
 
-  const { Edit } = getDefinitionForType(type);
-  const box = getBox(
-    localProperties ? { ...element, ...localProperties } : element
-  );
+    const { Edit } = getDefinitionForType(type);
+    const elementWithLocal = localProperties
+      ? { ...element, ...localProperties }
+      : element;
+    const box = getBox(elementWithLocal);
 
-  return (
-    <Wrapper aria-labelledby={`layer-${id}`} {...box} ref={ref}>
-      <Edit
-        element={localProperties ? { ...element, ...localProperties } : element}
-        box={box}
-        editWrapper={editWrapper}
-        onResize={onResize}
-        setLocalProperties={setLocalProperties}
-      />
-    </Wrapper>
-  );
-});
+    return (
+      <Wrapper aria-labelledby={`layer-${id}`} {...box} ref={ref}>
+        <Edit
+          element={elementWithLocal}
+          box={box}
+          editWrapper={editWrapper}
+          onResize={onResize}
+          setLocalProperties={setLocalProperties}
+        />
+      </Wrapper>
+    );
+  })
+);
 
 EditElement.propTypes = {
   element: PropTypes.object.isRequired,
--- a/packages/story-editor/src/components/canvas/editElement.js
+++ b/packages/story-editor/src/components/canvas/editElement.js
@@ -17,6 +17,7 @@
 /**
  * External dependencies
  */
+import { useState, forwardRef } from '@googleforcreators/react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import { useUnits } from '@googleforcreators/units';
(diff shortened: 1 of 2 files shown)


packages/story-editor/src/components/canvas/editElement.js
```diff
@@ -17,80 +17,62 @@
...
 } from '../../elements/shared';
-import SingleSelectionMoveable from './singleSelectionMoveable';
 
 const Wrapper = styled.div`
   ${elementWithPosition}
   ${elementWithSize}
-	${elementWithRotation}
-	pointer-events: initial;
+  ${elementWithRotation}
+  pointer-events: initial;
```
barklund, line 39: Not sure what this is about?

packages/story-editor/src/components/canvas/editLayer.js
```diff
@@ -33,11 +39,12 @@ import withOverlay from '../overlay/withOverlay';
...
+import SingleSelectionMoveable from './singleSelectionMoveable';
 
-const LayerWithGrayout = styled(Layer)`
+const LayerWithGrayout = withOverlay(styled(Layer)`
   background-color: ${({ grayout, theme }) =>
     grayout ? theme.colors.opacity.overlayDark : 'transparent'};
-`;
+`);
 
 const EditPageArea = withOverlay(PageArea);
```
barklund, line 49: I'm not sure if both elements need the overlay thingy here. @miina, do you know? It seems like it is required, so I will keep it for now.

packages/story-editor/src/components/canvas/editLayer.js
```diff
@@ -113,8 +141,21 @@ function EditLayerForElement({ element, showOverflow }) {
...
+        />
       </EditPageArea>
+      {hasEditModeMoveable && editWrapper && (
+        <SingleSelectionMoveable
+          selectedElement={element}
+          targetEl={editWrapper}
+          isEditMode
+          ref={setRef}
+        />
+      )}
```
barklund, line 158: I needed the moveable to exist outside of the page area (which is the full-bleed boundary) and have coordinates relative to the entire workspace, so had to move it out here.

packages/story-editor/src/components/canvas/editElement.js
merapi: It's just tabs vs spaces holy war 😉 

image

packages/story-editor/src/components/canvas/editElement.js
```diff
@@ -17,80 +17,62 @@
...
 
-  const [editWrapper, setEditWrapper] = useState(null);
   // Needed for elements that can scale in edit mode.
   const [localProperties, setLocalProperties] = useState(null);
 
-  const { Edit, hasEditModeMoveable } = getDefinitionForType(type);
+  const { Edit } = getDefinitionForType(type);
   const box = getBox(
     localProperties ? { ...element, ...localProperties } : element
   );
```
merapi: Just wondering why not:
```js
  const [localProperties, setLocalProperties] = useState({});
  const box = getBox({ ...element, ...localProperties }); // in both places
```

packages/story-editor/src/components/canvas/editLayer.js
```diff
@@ -90,6 +100,24 @@ function EditLayerForElement({ element, showOverflow }) {
...
+  const moveable = useRef(null);
+  const setRef = useCallback(
+    (moveableRef) => {
+      moveable.current = moveableRef;
+      onMoveableMount?.(moveableRef);
+    },
+    [onMoveableMount]
+  );
+
+  const onResize = () => {
```
miina: If we don't wrap this into `useCallback`, it will make the `EditElement` always re-render, maybe add `useCallback` here as well?

packages/story-editor/src/components/canvas/editElement.js
```diff
@@ -17,80 +17,62 @@
...
-      {hasEditModeMoveable && editWrapper && (
-        <SingleSelectionMoveable
-          selectedElement={element}
-          targetEl={editWrapper}
-          isEditMode
-          ref={moveable}
-        />
-      )}
-    </>
+    <Wrapper aria-labelledby={`layer-${id}`} {...box} ref={ref}>
```
miina: The naming here is a bit confusing now, maybe still rename the ref to `setEditWrapper` above if it's possible?
```
const EditElement = forwardRef(function EditElement(
  { element, editWrapper, onResize },
  ref: setEditWrapper
) {
```

packages/story-editor/src/components/canvas/editLayer.js
miina: Not sure either, if it was previously for `Moveable` only, then it shouldn't be needed anymore -- needs testing.

miina (commented):
(Will test in a bit as well and then approve if all ok)

packages/story-editor/src/components/canvas/editLayer.js
barklund: I think it's also used for some parts of the media edit, so I will keep it for now.

packages/story-editor/src/components/canvas/editLayer.js
barklund: It's not memoized so it will always re-render regardless. We could memoize it of course.

packages/story-editor/src/components/canvas/editElement.js
barklund: Yeah, you can even spread `null`, that doesn't break. I'll change this.

barklund: Hmm, spreading and re-structuring the (property-rich) element without cause is pretty wasteful actually. But I will move it to a separate variable.

barklund: Reworked it a bit in 4a44d28bb.

packages/story-editor/src/components/canvas/editLayer.js
barklund: I memoized both the component and the callback in 4a44d28bb. But I actually don't think it matters. I cannot think of an instance, where the `editLayer` renders without the `editElement` having to re-render too anyway. So either both have to render or neither, making memoization irrelevant. But memoizing is good practice anyway, so added.

miina (approved):

packages/story-editor/src/components/canvas/editElement.js
merapi: Yes and no :smile: 1000000 element spread operations took ~54ms, while simple assignment took ~0.35ms, quite a big difference - yes, does 0.000054 vs 0.00000035 do any real-world difference - no, given that it is executed only in edit mode, on pre/post resize text element and such, because on every media edit (scale, move) you will have those localProperties and it will spread.

Either way, 👍🏻   for moving that to a variable.

packages/story-editor/src/components/canvas/editElement.js
barklund: But it's just using the forwarded `ref`, regardless of what that `ref` is. I could rename to `forwardedRef`, but I think it's fine as is, no?

packages/story-editor/src/components/canvas/editElement.js
barklund: Well, that depends if spreads are re-written in babel to pre-es5 support or not. I actually do not know. I kind of assumed it was, and if so, it is definitely more expensive. But if it's transpiled, we should be good.

packages/story-editor/src/components/canvas/editElement.js
miina: It's just not easy to read the file like this and understand that the editWrapper and the `ref` are the same thing. Maybe just add a comment in a next PR?

PR #10395, opened 2022-01-27
Karma: waitfor implementation improvements part one

BrookeGraham:
## Context

This is to clean up some karma tests that were not using `waitFor` correctly. 

## Relevant Technical Choices

In places where we were using `waitFor` in addition to `getBy`, converted to using  await `findBy` since ```findBy methods are a combination of getBy queries and waitFor. They accept the waitFor options as the last argument```.  When that approach doesn't suffice, we throw an error if what we are waiting for isn't ready.

## Todo
Broke up the pr into two in an attempt to make reviewing easier. Additional pr #10429 😉

Found a test that is failing due to poor implementation. The test was grabbing the wrong element, which was leading to the appearance of it being flakey. Created a bug ticket #10428 

- [x] getting a false positive for canvas keys, fix dialog issue.  

## Testing Instructions

- [x] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1.

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/googleforcreators/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [x] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/googleforcreators/web-stories-wp/tree/main/docs#testing))
- [x] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Partially addresses #10194

Issue #10194, opened 2022-01-13
Karma: Improve `waitFor` implementation


BrookeGraham:
## Feature Description

Several Karma test seem to be "flaky" where nodes will be missing, or something we are awaiting doesn't alway get waited on when using `waitFor`. This is due to a slight misuse of the `waitFor` function. 

Let's go through the karma test where we are misusing `waitFor` and add an additional check to ensure successful polling / callback usage. 

## Additional Context

The `await waitFor(...)` function only polls to retry the callback _if_ the callback throws an error. We should be checking that the node we’re trying to wait on actually exists (something like `expect(node).tobeTruthy()`) inside the `waitFor`.  We can manually throw errors if needed as well. Without the re-polling it can contribute to the nodes being missing on some tests, seemingly at random.

PR #10429, opened 2022-02-01
Karma: `waitfor` implementation  improvements part two

### LINKED ISSUE #9930: Flaky Karma tests for `TextEdit integration`
## Bug Description
The following TextEdit integration tests are flaky and need fixing:
- `should mount editor` [Example run](https://github.com/google/web-stories-wp/runs/4437212276?check_suite_focus=true#step:6:1471)
- `should handle a command, exit and save` [Example run](https://github.com/google/web-stories-wp/runs/4437212276?check_suite_focus=true#step:6:1457)

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

BrookeGraham:
## Context
This is to clean up some karma tests that were not using waitFor correctly.

## Relevant Technical Choices
In places where we were using waitFor in addition to getBy, converted to using await findBy since findBy methods are a combination of getBy queries and waitFor. They accept the waitFor options as the last argument. When that approach doesn't suffice, we throw an error if what we are waiting for isn't ready.

## Todo
Broke up the pr into two in an attempt to make reviewing easier. Additional pr #10395  😉

## Testing Instructions

1.

## Checklist

---

Partially addresses #10194 
Fixes #9930


BrookeGraham:
Side note: I accidentally branched off of the other karma test pr #10395 which is why this one is pointing to it. Sorry.

packages/story-editor/src/elements/text/karma/edit.karma.js
```diff
@@ -176,8 +177,14 @@ describe('TextEdit integration', () => {
...
       it('should not change height when entering and exiting edit mode', async () => {
         await fixture.events.click(fixture.editor.library.textAdd);
-        await waitFor(() => fixture.editor.canvas.framesLayer.frames[1].node);
-        frame = fixture.editor.canvas.framesLayer.frames[1].node;
+        await waitFor(() => {
+          const node = fixture.editor.canvas.framesLayer.frames[1].node;
+          if (!node) {
+            throw new Error('node not ready');
+          }
+          expect(node).toBeTruthy();
```
miina, line 185: Is there still a purpose for adding `expect(node).toBeTruthy();` if it would throw an error above in case of `!node`? Seeing this in some other places as well. Thoughts?

packages/story-editor/src/components/richText/karma/singleSelection.karma.js
```diff
@@ -169,7 +164,6 @@ describe('Styling single text field', () => {
       await data.fixture.events.click(await fontWeight.option('Bold'));
       await data.fixture.events.sleep(300);
       await data.fixture.events.click(fontColor.button);
-      waitFor(() => fontColor.picker);
```
miina, line 172: This was for waiting for one tick for the Color Picker to open, curious if anything changed that it's not needed anymore, or if it wasn't needed before either?

packages/story-editor/src/elements/text/karma/edit.karma.js
BrookeGraham: No we could just throw and that would suffice.

packages/story-editor/src/components/richText/karma/singleSelection.karma.js
BrookeGraham: The `waitFor` here wasn't actually doing anything. Since `font.picker` doesn't throw an error, the `waitFor` never repolls. `waitFor` doesn't care if `font.picker` is null or not. If we think we do need to wait, we can add the same format as some of the other changes here. ie 
```
await waitFor(() => {
  if (!something) {
   throw new Error('this will make it so that the waitFor polls until no error is thrown')
  }
  expect(font.picker).toBeTruthy();
}
```

miina: > The waitFor here wasn't actually doing anything.

Ah, okay, I thought that `waitFor` would first do the initial interval wait and then run the callback, but it looks like it's running the callback first.

BrookeGraham: We can use `waitFor` without manually throwing errors _if_ the `expect` (or anything) inside throws an error for us. 
https://testing-library.com/docs/dom-testing-library/api-async/

packages/story-editor/src/elements/text/karma/edit.karma.js
```diff
@@ -77,22 +82,18 @@ describe('TextEdit integration', () => {
...
-        editLayer = fixture.querySelector('[data-testid="editLayer"]');
+        editor = await fixture.screen.findByTestId('textEditor');
+        editLayer = await fixture.screen.findByTestId('editLayer');
         boldToggle = fixture.editor.inspector.designPanel.textStyle.bold;
       });
 
-      // TODO https://github.com/googleforcreators/web-stories-wp/issues/9930
-      // eslint-disable-next-line jasmine/no-disabled-tests
-      xit('should mount editor', async () => {
+      it('should mount editor', async () => {
```
BrittanyIRL, line 90: 🎉

packages/story-editor/src/elements/text/karma/edit.karma.js
BrittanyIRL: i don't mind this despite the redundancy in some places, both patterns (like you said, Brooke) work just fine, i think this is a good reminder of how to actually use `waitFor` and having it so easy to read sometimes is helpful for future versions of ourselves.

BrittanyIRL (approved):
I love how many of these `waitFor`s you were able to just remove completely as noise. I've learned a lot from these two PRs 🙇  

One idea - I think it would be valuable to update the `integration-tests.md` in `/docs`  with how to properly use `waitFor` and when to rely on `findBy`. 
Screen Shot 2022-02-02 at 9 21 07 AM

packages/story-editor/src/components/panels/design/textStyle/karma/stylePresets.karma.js
```diff
@@ -109,11 +124,10 @@ describe('Panel: Style Presets', () => {
...
       // Confirm in the dialog since it's a global preset.
-      await waitFor(() => {
-        // We have two dialogues open now.
-        const dialogs = fixture.screen.getAllByRole('dialog');
-        expect(dialogs.length).toBe(2);
-      });
+
+      // We have two dialogues open now.
+      const dialogs = await fixture.screen.findAllByRole('dialog');
+      expect(dialogs.length).toBe(2);
```
BrookeGraham, line 130: Interestingly I found that the checklist companion dialog was always (at least when I was debugging here) present. So when we would check for role=‘dialog’ we were often getting the checklist dialog and not whatever one we were actually trying to target.

BrookeGraham:
>One idea - I think it would be valuable to update the integration-tests.md in /docs with how to properly use waitFor and when to rely on findBy.

You are right, it should be documented. I will add this. 🎉

packages/story-editor/src/components/panels/design/textStyle/karma/stylePresets.karma.js
samwhale: Yeah, the checklist is always rendered to the screen, although hidden, so that the checks inside the checklist may run on every update (since they run in the components).

samwhale (approved):
great!

PR #10432, opened 2022-02-01
Editor: Fix broken links

### LINKED ISSUE #10428: Canvas: Links no longer show Tooltip on hover.
## Bug Description

While cleaning up karma test in #10194, found the design link karma test was disabled citing it being flakey. However I found that the test wasn't grabbing the correct "tooltip" element. Either while this test was disabled or due to it testing the wrong thing a regression occurred causing links to no longer show a `tooltip` when hovering on the link element on the canvas. 

## Expected Behaviour

Tooltip should appear when hovering an element with a link. Karma test `CUJ: Creator Can Add A Link: Apply a link to any element should display the link tooltip correctly'` should be updated to grab the tooltip, and it should pass. 

## Steps to Reproduce

Add text element. 
Click `Link` in design panel
Add url
Deselect the text element on the canvas
Hover the text element
`Tooltip` should appear but does not. 

## Additional Context

- Plugin Version: 1.16.0
- WordPress Version: 7.4
- Operating System: OSX
- Browser: Chrome Version 97.0.4692.99 (Official Build) (x86_64)

merapi:
## Summary

I will enable the `xit` test after the first pass.

## Testing Instructions

In addition to fixing https://github.com/GoogleForCreators/web-stories-wp/issues/10428 (Steps to reproduce), it also fixes the element so it's fully functional - you can select and move it.

Add an element, add a link, deselect the element, and then verify the tooltip on hover and select/move actions.

## Checklist

---

Fixes #10428

### DIFF
--- a/packages/story-editor/src/components/elementLink/frame.js
+++ b/packages/story-editor/src/components/elementLink/frame.js
@@ -79,10 +79,6 @@ function WithLink({ element, active, children, anchorRef }) {
       </>
     ) : null;
 
-  if (!tooltipContent) {
-    return children;
-  }
-
   return (
     <StyledTooltip
       forceAnchorRef={anchorRef}
--- a/packages/design-system/src/components/tooltip/index.js
+++ b/packages/design-system/src/components/tooltip/index.js
@@ -87,7 +87,7 @@ let lastVisibleDelayedTooltip = null;
  * @param {Function} props.onPointerLeave Pointer leave event callback function
  * @param {string} props.placement Where to place the tooltip {@link: PLACEMENT}
  * @param {string} props.shortcut Shortcut text to display in tooltip
- * @param {string} props.title Text to display in tooltip
+ * @param {import('react').ReactNode|string|null} props.title Text to display in tooltip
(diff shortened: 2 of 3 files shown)


packages/story-editor/src/components/elementLink/frame.js
```diff
@@ -79,10 +79,6 @@ function WithLink({ element, active, children, anchorRef }) {
       </>
     ) : null;
 
-  if (!tooltipContent) {
-    return children;
-  }
```
barklund, line 84: Any idea what this bit of code was supposed to do? Is something else not working as it should now?

barklund (approved):
This seems to fix the issue.

packages/story-editor/src/components/elementLink/frame.js
merapi: https://xwp.slack.com/archives/CCCAUAH6F/p1643723188141189?thread_ts=1643716364.949999&cid=CCCAUAH6F

swissspidy (approved):

PR #10449, opened 2022-02-03
Fix Video Trimmer re-rendering when `currentTime` changes

### LINKED ISSUE #10243: Video Trimmer constantly re-renders due to changing `currentTime`
## Bug Description

The `useVideoNode` hook in the video trimmer tracks the video's `currentTime` in state, which is then passed on by `VideoTrimProvider` and consumed by `VideoTrimmer`.

There, it is only used to update the slider as the video is playing:

https://github.com/google/web-stories-wp/blob/083ade640608cab6e532b24716a51a88f34ece9a/packages/story-editor/src/components/videoTrim/videoTrimmer.js#L172-L178

This means the whole `VideoTrimmer` component is constantly re-rendering as long as the video is playing.

## Expected Behaviour

I feel like this could be improved by getting the `currentTime` from the actual video node instead, or at least do a simple

```js
const { currentTime } = useVideoTrim(({ state: { currentTime } }) => ({
    currentTime,
  }));
```

call within the `CurrentTime` component itself so that the re-render is scoped to that component.

## Steps to Reproduce

1. Enter trim mode
2. Observe re-renders in React dev tools or with the `useWhyDidYouUpdate` hook

## Additional Context

- Plugin Version: 1.16.0
- WordPress Version: 5.9
- Operating System: maCOS
- Browser: Chrome

miina:
## Summary
Moves getting `currentTime` to the new `CurrentTime` component to avoid re-rendering the whole `VideoTrimmer` constantly every second while the video is playing.

## To-do

- [x] Test that it actually works

## User-facing changes
N/A

## Testing Instructions

1.

## Checklist

---

Fixes #10243

### DIFF
--- a/packages/story-editor/src/components/videoTrim/currentTime.js
+++ b/packages/story-editor/src/components/videoTrim/currentTime.js
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


merapi (approved):
Does the job ✅  (verified with "highlight updates").

PR #10450, opened 2022-02-03
PHPCS: Improve linting for PHPDoc and imports

### LINKED ISSUE #7754: Add a sniff / lint to remove unused use statements
## Feature Description

Add a lint to detect and remove unused use statements in PHP codebase. See [UnusedUseStatementSniff](https://github.com/klausi/coder/blob/8.x-2.x/coder_sniffer/Drupal/Sniffs/Classes/UnusedUseStatementSniff.php) for inspiration. 

## Additional Context

---

spacedmonkey:
## Summary

Improving phpdocs, use statements and namespace functions calls, with these new linting rules. 

## Testing Instructions

1.

## Checklist

---

Fixes #7754


swissspidy:
Took a quick stab at this and slightly tweaked the PHPCS config you had here originally. That included removing rules that only really work on PHP 8+.
The result is stricter enforcement around order of PHPDoc annotations and `use` statements (especially for global functions).

So far this looks pretty decent.

The only remaining thing, which is currently disabled, is enforcing class constant visibility. But that should be trivial.

swissspidy:
The current config is something I can get behind.

includes/AMP/Canonical_Sanitizer.php
```diff
@@ -2,10 +2,10 @@
 /**
  * Class Canonical_Sanitizer.
  *
- * @package   Google\Web_Stories
```
spacedmonkey, line 5: @swissspidy Do we want package removed here?

includes/AMP/Output_Buffer.php
```diff
@@ -227,7 +225,7 @@ public function prepare_response( string $response ): string {
 		$dom = Document::fromHtml( $response );
 
 		if ( ! $dom instanceof Document ) {
-			return $this->render_error_page( SanitizationException::from_document_parse_error() );
+			return $this->render_error_page( \Google\Web_Stories\Exception\SanitizationException::from_document_parse_error() );
```
spacedmonkey, line 228: 😕

includes/AMP/Canonical_Sanitizer.php
swissspidy: Yes, I deliberately made that change in 6ce0f3a6c1b5785709229f8669ba7157c26e012c

It's really redundant as we have namespaces to indicate "packages".

includes/Settings.php
```diff
@@ -34,99 +34,71 @@
 class Settings extends Service_Base {
 	/**
 	 * Settings group.
-	 *
-	 * @var string
```
spacedmonkey, line 38: Do we want these removed?

swissspidy: Yes. It's redundant as the type can be inferred from the value below.

includes/AMP/Output_Buffer.php
swissspidy: Hmm this comes from when the `SlevomatCodingStandard.Namespaces.FullyQualifiedExceptions`  rule was initially enabled.

I have since removed it though, but the (automated) change was made before that though.

I don't mind it though.

phpcs.xml.dist
```diff
@@ -1,121 +1,257 @@
...
+
+  <!-- Check for cross-version support for PHP 7.2 and higher. -->
+  <config name="testVersion" value="7.2-" />
+  <rule ref="PHPCompatibilityWP" />
+
+  <!-- The main plugin file should be parsable by PHP < 7.0. -->
+  <rule ref="Generic.Arrays.DisallowLongArraySyntax.Found">
+    <exclude-pattern>web-stories.php</exclude-pattern>
+    <exclude-pattern
+    >includes/compat/Web_Stories_Compatibility.php</exclude-pattern>
```
spacedmonkey, line 90: ?

includes/AMP/Output_Buffer.php
spacedmonkey: It is only one place, so let's leave it.

spacedmonkey:
> The only remaining thing, which is currently disabled, is enforcing class constant visibility. But that should be trivial.

Should this be part of this PR?

phpcs.xml.dist
swissspidy: I think this comes from Prettier. I ran everything through `npx prettier --write phpcs.xml.dist`

swissspidy:
Whatever you prefer. I think this PR is already big enough and adding visibility to every class constant needs some consideration for each class. Some need to be public because they're used elsewhere, some can be private/protected.

spacedmonkey:
> Whatever you prefer. I think this PR is already big enough and adding visibility to every class constant needs some consideration for each class. Some need to be public because they're used elsewhere, some can be private/protected.

You are right, this PR is already very big. Let's follow up with this later.

swissspidy (approved):

PR #10452, opened 2022-02-03
Packages: Create `masks`, `elements` and `element-library` packages

### LINKED ISSUE #8125: Packages: Move `elements` to new structure
## Feature Description

The dedicated `elements` package provides various APIs to register & retrieve element types.

**References:**

* [Package organization doc](https://docs.google.com/document/d/1ImRe9wYKWGCAZiq3jro5KLpSJK_NMKXHC6FWqs3nwl0/edit#)
* Proof-of-concept implementation: https://github.com/google/web-stories-wp/tree/try/next-level-packages/packages/elements/src

**To-do:**

* [ ] Move some of the code from `/assets/src/edit-story/elements` to `/packages/elements/src/`
* [ ] Update imports where necessary.  
  Use `@web-stories-wp/...` imports for anything that is already a package, relative imports for everything else.
* [ ] Create `/packages/elements/package.json` referencing its dependencies
* [ ] Create `/packages/elements/README.md` with short documentation.

### Blockers ⚠️

* [ ] Remove any references to specific element types. For example, `createPage` references the `shape element` type. Move this util to story editor instead.
* [ ] Need a good place to reference rich-text related utils. (new `rich-text` package?)
* [ ] Need a good place to reference masks. (new `masks` package?)

## Additional Context

---

### LINKED ISSUE #8777: Packages: Move `element-library` to its own package
## Task Description

The idea for this dedicated `element-library` package would be to contain the individual elements as supported by the editor. Provides a function to “register” them in the app. So every consumer needs to register these "core" element types.

See https://docs.google.com/document/d/1ImRe9wYKWGCAZiq3jro5KLpSJK_NMKXHC6FWqs3nwl0/edit for context.
See also `@wordpress/block-library` for context.

**To-do**

* [ ] The individual elements like the text element currently depend on Canvas, Story and Rich Text providers from `story-editor`. So this package would need to be used from within `wp-story-editor` to avoid circular dependencies.

mohdsayed:
## Context

As part of package organisation ( https://github.com/GoogleForCreators/web-stories-wp/issues/4737 ), this PR creates three more packages `masks`, `elements` and `element-library` from `story-editor` package.

Also see https://github.com/GoogleForCreators/web-stories-wp/issues/8125 and https://github.com/GoogleForCreators/web-stories-wp/issues/8777

## Testing Instructions

1.

## Checklist

---

Fixes https://github.com/GoogleForCreators/web-stories-wp/issues/8125
Fixes https://github.com/GoogleForCreators/web-stories-wp/issues/8777


packages/elements/src/elementType.js
```diff
@@ -14,11 +14,11 @@
  * limitations under the License.
  */
 
-/**
- * External dependencies
- */
-import { __ } from '@googleforcreators/i18n';
+// @todo Create a custom hook to manage state.
+const elementTypes = [];
```
mohdsayed: @swissspidy Need your opinion here for handling `elementTypes` registration. Though a simple array works here however it doesn't feel right to keep it like this. I would expect it to become some state perhaps by keeping it in a custom hook, something like this

```js
/**
 * External dependencies.
 */
import { useState } from '@googleforcreators/react';

function useElementTypes() {
  const [ elementTypes, setElementTypes ] = useState( [] );

  const registerElementType = (elementType) => {
    setElementTypes( [ ...elementTypes, elementType ] );
  }

  return { elementTypes, registerElementType };
}

function useGetDefinitionForType( type ) {
  const { elementTypes } = useElementTypes();

  return () => elementTypes.find((el) => el.type === type);
}

export { useElementTypes, useGetDefinitionForType };
````

however, the problem is that if `elementTypes` becomes a state, it will have a cascading effect on other parts of the code using it for example `getDefinitionForType` , `createNewElement` would have to become custom hooks too. Then a lot of areas where these utility functions are being used in `story-editor` will have to be refactored to handle hooks.

Ideally, I would wish to keep them as static functions ( for example `registerblocktype` in Gutenberg ) but unfortunately, we do not have a central store that can do that ( like data package in Gutenberg ).

Comment (swissspidy) [reply packages/elements/src/elementType.js]: I see the problem and I am on the same page as you. Using a hook here would indeed cause quite a cascading effect. So to me a central/global store also sounds more appealing right now.

This is a good reason/opportunity to revisit this in a dedicated ticket explaining the use case.

Comment (mohdsayed) [reply packages/elements/src/elementType.js]: Yes and like you previously asked, that central store may also help us in making data from the story editor available to plugins like Yoast SEO.

Comment (mohdsayed) [reply packages/elements/src/elementType.js]: @swissspidy I suppose #10614 discussion and the actual work might take some time and `story-editor` is constantly getting updated. I am wondering what to do with this PR as it would be a little difficult to keep it in sync and my other package organization tickets may also depend on this PR.

Comment (swissspidy) [reply packages/elements/src/elementType.js]: As discussed on Slack, using something like Redux or zustand can be done in follow-up tickets. This here works as a temporary solution in the meantime.

Comment (maxyinger) [reply packages/elements/src/elementType.js]: Curious about the proposed use case here. So the idea is that once the app is running a plugin might add a new element type that necessitates a re-render in the application?

Or would this be data we just need at build time?

Comment (mohdsayed) [reply packages/elements/src/elementType.js]: > So the idea is that once the app is running a plugin might add a new element type that necessitates a re-render in the application?

That's the wish when we set up a proper data package.

Comment (mohdsayed) [reply packages/elements/src/elementType.js]: I think if use `useSelector` to get state in components and `store.dispatch` to update/register using `react-toolkit`, we would have a static function to register elements and the app will re-render. Will try out a POC to see how it works.

Comment (maxyinger) [reply packages/elements/src/elementType.js]: Cool. I'm excited to see what you whip up! should definitely be easier to discuss/visualize once we have something a little more tangible to look at.

Comment (mohdsayed) [reply packages/elements/src/elementType.js]: Made an attempt to covert related functions like `getDefinitionForType`, `getElementProperties`, `createElementForCanvas` to hooks so that `useSelector` can be used to get state from the store and app re-render on state change, however it doesn't look like a quick win. They are also being used conditionally and inside loops at some places and it would require a lot of refactoring. We need to work on it as a separate ticket. 

So for now, even if we use a store, the element registration will be required at build time.

File: packages/story-editor/src/components/canvas/utils/test/getElementProperties.js [hunk 2]
```diff
@@ -17,6 +17,11 @@
 /**
  * Internal dependencies
  */
+/**
+ * External dependencies
+ */
```
spacedmonkey: What is happening here

packages/story-editor/src/components/panels/design/filter/filter.js
```diff
@@ -19,29 +19,49 @@
...
 import { Row, Color, FilterToggle } from '../../../form';
 import { SimplePanel } from '../../panel';
-import { getDefinitionForType } from '../../../../elements';
 import { useCommonColorValue } from '../../shared';
-import { MULTIPLE_VALUE } from '../../../../constants';
+import useCORSProxy from '../../../../utils/useCORSProxy';
+import { useStory } from '../../../../app';
 import convertOverlay from './convertOverlay';
 
+export const OverlayPreset = {
```
swissspidy: This seems to be only used in this file, so no need to export.

swissspidy:
Superb work on untangling the code base to create these new packages!

We're definitely on the right track here. This PR highlights very well the areas where there is still some work to do, because there are still many pieces we haven't found the right home for yet.

Some pieces we possibly need to find a better solution for, either here or in follow-up PRs:

* `BACKGROUND_TEXT_MODE`
* `TEXT_ELEMENT_DEFAULT_FONT`
* `MULTIPLE_VALUE` and `MULTIPLE_DISPLAY_VALUE`
* `ELEMENT_TYPES` usage in quick actions and right click menu
  This usage means the editor still has some hardcoded knowledge of what element types exist and are supported.
  These could probably be moved to the element definition, so that the element defines what menu items it has. Similar to how it defines its list of supported panels.
* `PanelTypes`
* `getTransformFlip`

spacedmonkey:
This PR seems good to me, but not sure if I am best person to review this.

packages/design-system/src/utils/constants.js
```diff
@@ -32,3 +32,62 @@ export const FOCUSABLE_SELECTORS = [
   'textarea:not(:disabled)',
   '[tabindex]:not([tabindex="-1"])',
 ];
+
```
BrittanyIRL, line 35: 🤔 none of these feel like they are part of the design system. These are specific to the editor, no? 

The idea of a design system was that it's just shared components/theme.

packages/element-library/src/media/scalePanel.js
```diff
@@ -24,11 +24,10 @@ import { Slider } from '@googleforcreators/design-system';
...
 
-/**
- * Internal dependencies
- */
-import { Z_INDEX_CANVAS } from '../../constants';
-
+const Z_INDEX_CANVAS = {
+  MOVABLE: 10,
+  FLOAT_PANEL: 11,
+};
```
BrittanyIRL: This feels a little slippery - z index is specific to the editor, maybe I'm missing something about how things are getting bucketed.  We've got work coming up next sprint on addressed the z-index issues pertaining to the floating menus and I'm a little apprehensive about this move until I understand the thought process better about how the boundaries here are getting drawn. Apologies if I missed a comment or thread somewhere.

packages/element-library/src/test/index.js
```diff
@@ -13,19 +13,28 @@
...
-import { TEXT_ELEMENT_DEFAULT_FONT } from '../../app/font/defaultFonts';
 import {
   createNewElement,
   createPage,
   duplicatePage,
   duplicateElement,
-} from '..';
+  registerElementType,
+} from '@googleforcreators/elements';
+import { TEXT_ELEMENT_DEFAULT_FONT } from '@googleforcreators/design-system';
```
BrittanyIRL: To me this feels like this constant belongs in `elements` not `design-system` since it's the default for an element that the design system should be agnostic about.

packages/element-library/src/media/scalePanel.js
mohdsayed: The only reason why it was moved here is that the only place `Z_INDEX_CANVAS` is being used is in this file. When it is required in more than one place, we can think of its proper home.

packages/element-library/src/test/index.js
mohdsayed: Thank you! Moved in https://github.com/GoogleForCreators/web-stories-wp/pull/10452/commits/fad8ae498113d54de0b225d524d06c78404c9f6b

mohdsayed:
Based on your valuable feedback, I created a list of objectionable items and their possible solutions. Can think about these together?

- `BACKGROUND_TEXT_MODE`
  - Previously in: `story-editor/src/constants/index.js`
  - Now in: `design-system/src/utils/constants.js`
  - Used in : `element-library`, `elements`, `story-editor`
  - Possible Solution: Move to `elements` package.
- `TEXT_ELEMENT_DEFAULT_FONT` ( DONE )
  - Previously in: `story-editor/src/app/fonts/defaultFonts.js`
  - Now in: `design-system/src/utils/constants.js`
  - Used in: `story-editor`, `element-library`
  - Possible Solution: Move to `elements` or `element-library` packages.
  - Brittany >  Feels like this constant belongs in `elements`, not `design-system` since it's the default for an element that the design system should be agnostic about.
- `MULTIPLE_VALUE`  ( DONE )
  - Previously in: `story-editor/src/constants/multipleValue.js`
  - Now in: `elements/src/constants.js`
  - Used in: `story-editor`, `rich-text`, `elements`
  - Possible Solution: Move to `rich-text` or copy in all three packages as it's just a small value `'((MULTIPLE))'` and less likely to update.
- `MULTIPLE_DISPLAY_VALUE` ( DONE )
  - Previously in: `story-editor/src/constants/multipleValue.js`
  - Now in: `elements/src/constants.js`
  - Used in: `story-editor` 
  - Possible solution: Was moved along with `MULTIPLE_VALUE`, but we can just keep it in `story-editor` as it's not being used elsewhere.
- `PanelTypes`
  - Previously in: `story-editor/src/components/panels/design/types.js`
  - Now in: `design-system/src/utils/panelTypes.js`
  - Used in: `element-library`, `story-editor`
  - Possible solution: Move to `element-library`
- `getTransformFlip`
  - Previously in: `story-editor/src/elements/shared/getTransformFlip.js`
  - Now in: `packages/design-system/src/utils/getTransformFlip.js`
  - Used in: `masks`, `element-library`
  - Possible Solution: Move to `element-library` .
- `OverlayType`
  - Previously in: `story-editor/src/utils/overlay.js`
  - Now in: `design-system/src/utils/constants.js`
  - Used in: `story-editor`, `elements`
  - Possible Solution: Move to `elements`.
- `DEFAULT_ATTRIBUTES_FOR_MEDIA`
  - Previously in : `story-editor/src/constants/index.js`
  - Now in: `design-system/src/utils/constants.js`
  - Used in: `story-editor`, `element-library`
  - Possible Solution: Move to `element-library`

---

> ELEMENT_TYPES usage in quick actions and right click menu
This usage means the editor still has some hardcoded knowledge of what element types exist and are supported.
These could probably be moved to the element definition, so that the element defines what menu items it has. Similar to how it defines its list of supported panels.

I agree, I think we can deal with this one, in a follow-up PR.

packages/design-system/src/utils/constants.js
mohdsayed: Agree, I included all of them in the list here to discuss https://github.com/GoogleForCreators/web-stories-wp/pull/10452#issuecomment-1059702590

packages/design-system/src/utils/index.js
```diff
@@ -21,5 +21,13 @@ export { default as useLiveRegion } from './useLiveRegion';
 export { default as uniqueEntriesByKey } from './uniqueEntriesByKey';
 export { default as deepMerge } from './deepMerge';
 export { default as useMouseDownOutsideRef } from './useMouseDownOutsideRef';
+export { default as pointerEventsCss } from './pointerEventsCss';
+export { default as objectWithout } from './objectWithout';
+export { default as getKeyboardMovement } from './getKeyboardMovement';
+export { default as areEventsDragging } from './areEventsDragging';
+export { default as usePerformanceTracking } from './usePerformanceTracking';
```
miina, line 27: Wouldn't this be more suitable in the `tracking` package?

packages/design-system/src/utils/index.js
```diff
@@ -21,5 +21,13 @@ export { default as useLiveRegion } from './useLiveRegion';
 export { default as uniqueEntriesByKey } from './uniqueEntriesByKey';
 export { default as deepMerge } from './deepMerge';
 export { default as useMouseDownOutsideRef } from './useMouseDownOutsideRef';
+export { default as pointerEventsCss } from './pointerEventsCss';
+export { default as objectWithout } from './objectWithout';
+export { default as getKeyboardMovement } from './getKeyboardMovement';
+export { default as areEventsDragging } from './areEventsDragging';
```
miina: This also feels a bit out of place here 🤔 This util was mostly created for Moveable -- since a click for moveable triggers dragging but we want to be able to handle click separately. So in this sense, it might make sense for the Moveable package. However, it's used in one more place but maybe that's ok.

packages/element-library/src/media/scalePanel.js
miina: I second Brittany here that this feels out of place, even if it's used only in this file. There are some other Z_INDEX constants out there in the editor, e.g.
```
export const Z_INDEX = {
  NAV: 2,
  EDIT: 3,
};
```

Maybe there's a place where these constants should be combined then in the editor? These z-indexes are editor-specific. Otherwise we could also set the values to 1 and 2 here -- independently it should work ok, it just wouldn't work within the editor anymore. If we'll keep them as it is, would be great to add a comment about why these are 10 and 11.
Thoughts?

packages/story-editor/src/output/element.js
```diff
@@ -20,21 +20,21 @@
...
 import { getBox } from '@googleforcreators/units';
 import { StoryAnimation } from '@googleforcreators/animation';
+import { getDefinitionForType } from '@googleforcreators/elements';
+import { BACKGROUND_TEXT_MODE } from '@googleforcreators/design-system';
+import {
+  OutputWithMask as WithMask,
+  getBorderPositionCSS,
+  getBorderRadius,
+  getBorderStyle,
+  shouldDisplayBorder,
```
miina, line 30: If we're moving some other element props related utils to `element-library` (EDIT: or `elements`), maybe most of the border utils would also be better there? Most of these are not directly related to masks, as far as I know, and are mainly just creating border CSS. It's only the ones detecting if the element supports border at all that are related to masks. Thoughts?

packages/story-editor/src/components/canvas/frameElement.js
```diff
@@ -80,41 +90,72 @@ const EmptyFrame = styled.div`
 const NOOP = () => {};
 
 function FrameElement({ id }) {
-  const { isSelected, isBackground, element } = useStory(({ state }) => ({
-    isSelected: state.selectedElementIds.includes(id),
-    isBackground: state.currentPage?.elements[0].id === id,
-    element: state.currentPage?.elements.find((el) => el.id === id),
+  const [isTransforming, setIsTransforming] = useState(false);
```
miina, line 93: cc @littlemilkstudio Tagging you here just in case to review this file from performance perspective.
If I'm not mistaken you looked into performance of rendering the elements and frames and some point, and looks like there are a bunch of new dependencies added here. (Feel free to ignore if I remember incorrectly)

packages/story-editor/src/components/canvas/editLayer.js
```diff
@@ -156,9 +161,9 @@ function EditLayerForElement({ element, showOverflow }) {
           ref={setRef}
         />
       )}
-      {EditMenu && (
```
miina, line 160: The idea of this was that `EditMenu` can be any `EditMenu` and not only trimming edit menu, even though currently we have only trimming.
Is there any reason to change this?

cc @barklund as you originally created this architecture.

packages/story-editor/src/components/canvas/editLayer.js
```diff
@@ -79,7 +81,10 @@ function EditLayer() {
 function EditLayerForElement({ element, showOverflow }) {
   const ref = useRef(null);
   const pageAreaRef = useRef(null);
-  const { editModeGrayout, EditMenu } = getDefinitionForType(element.type);
+  const { editModeGrayout } = getDefinitionForType(element.type);
+  const { isTrimMode } = useVideoTrim(({ state: { isTrimMode } }) => ({
```
miina, line 86: We could here return the trim mode directly, e.g.
`const isTrimMode = useVideoTrim(({ state }) => state.isTrimMode)`

(Even though I'm not sure why this file is changed)

packages/story-editor/src/components/canvas/displayElement.js
```diff
@@ -106,6 +108,15 @@ function DisplayElement({ element, previewMode, isAnimatable = false }) {
     getBox: state.actions.getBox,
     dataToEditorX: state.actions.dataToEditorX,
   }));
+  const { getProxiedUrl } = useCORSProxy();
```
miina, line 114: Would be interesting to know if/how these changes influence re-rendering / performance. Since the dependencies are now higher up in the hierarchy, wondering if this can also influence performance.

miina:
> Based on your valuable feedback, I created a list of objectionable items and their possible solutions. Can think about these together?

`MULTIPLE_VALUE` - since this is used by all elements then `rich-text` doesn't seem to be the best place, your suggestion of having it in all packages separately makes more sense (to me).

`MULTIPLE_DISPLAY_VALUE` - this feels like completely editor-specific, if possible, let's keep it in the editor only.

`PanelTypes` - also feels like it's editor-specific 🤔 Since ideally it could be chosen by implementation which panels the element supports, and which panels even exist. Maybe it should be possible to define in the editor which panels are registered for which elements instead of determining it statically per element? A new helper?
Perhaps I'm not understanding the purpose of the `element-library` though, maybe it's expected that the panels are pre-defined by the packages and not the implementation.

packages/story-editor/src/output/element.js
mohdsayed: yes, it looks suitable for `element-library` or `elements`. The only problem is that it will create circular dependencies.

1. If we move to `elements`:
- `elements` will have `masks` dependency, because `shouldDisplayBorder` is using utils from `masks` package.
- `masks` will have to add `elements` dependency because `shouldDisplayBorder` is being used in `masks` package.

2. if we move to `element-library`:
- `element-library` is already using `masks` as a dependency.
- `masks` will have to add `element-library` as a dependency because `shouldDisplayBorder` is using utils from `masks` package.

miina: Hm, it looks like `shouldDisplayBorder` might need adjusting then. It currently checks for the following:
`hasBorder`, `canMaskHaveBorder`, `canSupportMultiBorder`. Maybe it needs to be split up better then so that all the relevant parts would be in the relevant packages? Or maybe this util doesn't make sense as it is anymore at all.

Could be for a follow-up issue, this PR is already huge, thoughts?

packages/design-system/src/utils/index.js
swissspidy: Technically yes. Though the tracking package is very lightweight and does not have any external dependencies currently.

Moving this hook would mean adding a dependency on React.

Given that this package is also used outside the story editor (i.e. in the Gutenberg block and the plugin activation notice), we'd need to be careful to not inadvertently increase bundle size for those places.

packages/design-system/src/utils/index.js
swissspidy: Both `element-library` and `story-editor` depend on `moveable` already, so it should be fine to move it to that package.

packages/story-editor/src/components/canvas/editLayer.js
swissspidy: > Is there any reason to change this?

The reason is that the `VideoEdit` component currently has a dependency on `isTrimMode`/ `useVideoTrim`

With the elements being moved to `element-library` it cannot use that hook from the editor.

This change moves this out of `VideoEdit` to this component here.

Any suggestions to make this more generic are welcome.

packages/story-editor/src/components/canvas/editLayer.js
swissspidy: > (Even though I'm not sure why this file is changed)

See https://github.com/GoogleForCreators/web-stories-wp/pull/10452/files/f816101a70ebefbbef9e482ed695d0bf96361dea#r820758283

This line was copied 1:1 from `VideoEdit`

packages/story-editor/src/components/canvas/editLayer.js
mohdsayed: Good question! This is a tricky one.

`EditMenu` is eventually returning only `<VideoTrimmer />; `  if `isTrimMode` true, and its only being used for video.

https://github.com/GoogleForCreators/web-stories-wp/blob/8f1a9d71d66d85dc8f02a000847d78272d115966/packages/story-editor/src/elements/video/editMenu.js#L28-L32

Therefore I just used that condition here instead of passing the component as prop.

> The idea of this was that EditMenu can be any EditMenu and not only trimming edit menu

If we want to use `EditMenu` component, we may have to do something like this.

```js
  const menu = (
    <EditMenu
        videoTrimmer={ <VideoTrimmer /> }
        isTrimMode={ isTrimMode }
    />
  )

  return menu && (
    <FooterArea showOverflow>
      { menu }
    </FooterArea>
  )
```

because `<VideoTrimmer>` and `useVideoTrim` live in `story-editor`.

Currently, it's just video, but as you said, let's say that in the future we want to add more options for the `EditMenu` for something else, we again will have to update this part of the code anyways.

Thoughts?

swissspidy:
> > `PanelTypes` - also feels like it's editor-specific 🤔 Since ideally it could be chosen by implementation which panels the element supports, and which panels even exist. Maybe it should be possible to define in the editor which panels are registered for which elements instead of determining it statically per element? A new helper?

FWIW I've always found the current implementation a bit confusing. Each element currently defines a list of panels it supports (the `panels` const), and then `getDesignPanelsForSelection()` in the editor makes the actual choice.
But I suppose it's just confusing because there are also background elements with some special handling there.

> Perhaps I'm not understanding the purpose of the `element-library` though, maybe it's expected that the panels are pre-defined by the packages and not the implementation.

The idea behind `element-library` is to decouple the available element types from the editor, so that someone building a custom editor could add and remove element types.

----

For context, the way Gutenberg has solved the elements <-> panels relationship is basically by making `element-library` a dependency of `wp-story-editor` (instead of `story-editor`) and then actually add `story-editor` as a dependency to `element-library`. Then each element can import the panel types it wants from `story-editor`.

packages/story-editor/src/output/element.js
mohdsayed: 👍 Yes, I will try in a follow-up PR.

packages/design-system/src/utils/index.js
mohdsayed: Yes, that was the thought behind not moving it in the tracking package, the tracking package currently doesn't have any dependency.

packages/design-system/src/utils/index.js
mohdsayed: Moved to `moveable` package, thanks!

packages/element-library/src/media/scalePanel.js
mohdsayed: Would be happy to move, but I cannot think of any suitable home for `Z_INDEX_CANVAS`.

> If we'll keep them as it is, would be great to add a comment about why these are 10 and 11.

Not sure how it works, can you suggest a `@todo` I can add?

packages/design-system/src/utils/index.js
miina: Hm, okay, makes sense. And it needs to be removed from the `story-editor` since now it's used in masks as well, correct?

To me, it would make the most sense to keep it in the editor only and rework the usage of it so that it would be used in the editor only. E.g. when looking at the `performanceTrackingEvents.js`, that's very editor-specific and shouldn't really be in the `tracking` package nor in the `design-system`. So it feels like it's a custom hook made for the editor specifically based on the tracking package.

(It would require a follow-up issue to remove the usage of the hook from separated packages to use it inside the editor only, too much to tackle it in this PR).

Thoughts?

packages/element-library/src/media/scalePanel.js
miina: > Would be happy to move, but I cannot think of any suitable home for `Z_INDEX_CANVAS`.

Since all these z-indexes are in theory editor-specific and determine how different layers are positioned on top of each other in the editor, then ideally, this should somehow come from the implementation (the editor).

That said, perhaps we could verify first that it actually does influence how things are displayed in the editor -- since it's currently used in one place only, maybe it's not related to all the other z-indexes in the editor. In this case just renaming the constant to something else to avoid confusion could be ok.

But if it indeed influences how the layers are displayed in the editor, then probably a follow-up issue to investigate it / figure out a solution would be necessary.

Thoughts?

packages/design-system/src/utils/index.js
swissspidy: If we can keep it solely in the editor and not in other packages, that would be great for sure.

packages/story-editor/src/components/canvas/frameElement.js
maxyinger: @miina Thanks for tagging. Hmph, yea it looks like this introduces some regressions there.

For reference, most of the perf work I did in here was just around not causing a render for every frame element when only one needed updating. Tested this by doing some simple actions like select, or moving an element.

Looks like this PR introduces a regression here. Here's a picture of the flame graph when selecting a new element:
Screen Shot 2022-03-07 at 10 23 38 AM

packages/story-editor/src/components/canvas/frameElement.js
```diff
@@ -80,41 +90,72 @@ const EmptyFrame = styled.div`
...
   }));
-  const setEditingElement = useCanvas(
-    ({ actions }) => actions.setEditingElement
-  );
+  const { isSelected, selectedElementIds, isActive, isBackground, element } =
+    useStory(({ state }) => ({
+      isSelected: state.selectedElementIds.includes(id),
+      isBackground: state.currentPage?.elements[0].id === id,
+      element: state.currentPage?.elements.find((el) => el.id === id),
+      selectedElementIds: state.selectedElementIds,
```
maxyinger: If I had to guess, this is introducing the performance regression I described earlier. Is there a way we could get the intended functionality without re-rendering every instance of this component every time the selection updates?

maxyinger (commented):
Still need to audit some more, but looks like this introduces some performance regressions

packages/story-editor/src/components/canvas/displayElement.js
maxyinger: Yea I believe it would. In general, any action associated with these dependencies would now cause all display elements to render vs just the element types which actually need the data returned from the hook.

packages/story-editor/src/components/canvas/frameElement.js
miina: From what I can see, the `selectedElementIds` is used only in the text frame for example. So we should be able to at least reduce rendering by having only the text element IDs returned.

Also, looking at the `text/frame.js`, it looks like the relevant thing is to know if a text element is the only element in selection. We probably don't need to get the whole `selectedElementIds` for that -- looks like improvements could be made there as well.

swissspidy: Looks like this was added because the text element's `TextFrame` component was using `selectedElementIds` before:

https://github.com/GoogleForCreators/web-stories-wp/blob/e6dc2e95961688d3935a9849587ae50426b0c0f2/packages/story-editor/src/elements/text/frame.js#L61-L70

With elements being moved to their own package, it can't use `useStory()` any longer.

swissspidy: This would already be a big improvement I think:

```js
  const { isSelected, isOnlySelectedElement, isActive, isBackground, element } =
    useStory(({ state }) => {
      const isOnlySelectedElement = state.selectedElementIds.length === 1 &&
        state.selectedElementIds.includes(id);
      return {
        isSelected: state.selectedElementIds.includes(id),
        isBackground: state.currentPage?.elements[0].id === id,
        element: state.currentPage?.elements.find((el) => el.id === id),
        selectedElementIds: state.selectedElementIds,
        isOnlySelectedElement,
        isActive:
          isOnlySelectedElement
          !isTransforming &&
          !isEditing,
      }
    });
```

Then `TextFrame` just needs to check `isOnlySelectedElement`

packages/element-library/src/media/scalePanel.js
mohdsayed: Oh okay, I think now I understand the concern a little bit. We can keep it in `story-editor` and maybe pass it down to `ScalePanel` via props.

packages/design-system/src/utils/index.js
mohdsayed: hmm, `usePerformanceTracking` is custom hook, if we keep it only in `story-editor`, we cannot even pass it down as a prop to `mask` package 🤔 


https://github.com/GoogleForCreators/web-stories-wp/blob/ac8bd1373a286d36c8d62cf2e2057a19653791a6/packages/masks/src/frame.js#L190-L197

packages/story-editor/src/components/canvas/displayElement.js
swissspidy: Worth noting that `getProxiedUrl` never really changes

packages/design-system/src/utils/index.js
swissspidy: I think what Miina is saying is that its usage in the `mask` package needs to be refactored, so that it will only be used in the `story-editor` package in the future; but that this can be done in a follow-up ticket since this PR is already quite big.

packages/story-editor/src/components/canvas/frameElement.js
mohdsayed: @littlemilkstudio Not able to see much difference in my profiler test. Can you lay down the steps you followed to get to this result?

packages/design-system/src/utils/index.js
miina: ^ Yep, that's what I was trying to say.

miina: @sayedtaqui Could you also perhaps link the follow-up issues to the resolved threads for visibility here and elsewhere? Unless there's a different system for tracking those in case of decoupling. Thoughts?

Edit: Then the original thread will be linked to the issue and it's easier to get the context.

packages/story-editor/src/components/canvas/displayElement.js
miina: > Worth noting that getProxiedUrl never really changes

Yep, I was thinking generally of all the new dependencies there, not just that :)

Since the Page rendering performance has been an issue and a blocker for displaying the prev and next page in the canvas (which is now prioritized by UX), would be ideal if we wouldn't cause performance regressions here, or at least we could log those and fix as a priority.

packages/design-system/src/utils/index.js
mohdsayed: Sure, I am noting down the to-do items, and will create issues at the end and will link in comments.

packages/element-library/src/media/scalePanel.js
mohdsayed: Does this look good? https://github.com/GoogleForCreators/web-stories-wp/pull/10452/commits/1c2577106d1f4fd84b6fe7b0842c42c5337196a7

packages/story-editor/src/components/canvas/editLayer.js
miina: Hm, I suppose the question is also -- is the Video Trimmer editor-specific or element specific? If it's element-specific and should be supported by default for the video element, then the trimmer itself should also live in the `elements` package perhaps. Thoughts?

packages/element-library/src/media/scalePanel.js
miina: Looking at it again, I'm seeing that the `zIndex.MOVABLE` is not even used at all anywhere? 🤔 
In this case, we could just pass one z-index directly. WDYT?

```
function ScalePanel({
  setProperties,
  width,
  height,
  x,
  y,
  scale,
  zIndex,
  ...rest
}) {
  return (
    <InOverlay zIndex={zIndex} pointerEvents="initial">
  ...
```

mohdsayed: Since `MOVABLE` is currently not being used we can remove it from the object but still pass the same object so that more keys can be passed in the future.

packages/story-editor/src/components/canvas/displayElement.js
mohdsayed: Tried one solution which prevents re-rendering when tested, can you take a look https://github.com/GoogleForCreators/web-stories-wp/pull/10452/commits/1767cf0c066c560c4148881260ab8c26cf8d1ace

packages/story-editor/src/components/canvas/editLayer.js
mohdsayed: To me, it looks element-specific, but since its using a lot of hooks and components from `story-editor`, it wouldn't be easy to move it to the element package.

https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/story-editor/src/components/videoTrim/videoTrimmer.js

miina: I also feel like ideally it would be element-specific, however, this would need a lot of reworking indeed.
And allowing different `EditMenu` options also makes sense only if those `EditMenu` items actually live inside the elements' library, if the `EditMenu` content (e.g. video trimmer) lives in the editor for now anyway, then I think what you have currently makes sense.

Might be good to create a potential follow-up issue for future to look into options of moving the video trimmer into the elements package, just to have it there somewhere.

Thoughts?

BrittanyIRL:
> I created a list of objectionable items and their possible solutions. Can think about these together?

I really love this, thank you for taking the time. Happy with it being isolated as a follow up. Can you track it in a ticket so it doesn't get lost?

packages/design-system/src/utils/index.js
mohdsayed: Issue https://github.com/GoogleForCreators/web-stories-wp/issues/10862

packages/story-editor/src/components/canvas/frameElement.js
mohdsayed: Refactored in https://github.com/GoogleForCreators/web-stories-wp/pull/10452/commits/be6a5baed59440a757780009a52b1fbc0c4ba63b

packages/story-editor/src/components/canvas/editLayer.js
mohdsayed: Yes, created a follow-up issue https://github.com/GoogleForCreators/web-stories-wp/issues/10867 . Let's deal with it separately.

packages/story-editor/src/components/canvas/frameElement.js
mohdsayed: Speaking in general, since we are breaking down the code into multiple packages, there would again be instances where we would want to use some states from `story-editor` ( for example we couldn't use `useStory` here ). The only option we have is that we use some workaround by passing down props to avoid their usage in another package ( Because other packages cannot import anything from `story-editor` to avoid circular dependencies ). This works for now, but it doesn't look very scalable and is probably not a proper solution.

💭  Maybe these are some more examples of why some of our states should live in some central package that can be shared among all packages ( https://github.com/GoogleForCreators/web-stories-wp/issues/10614 ), but not sure how these should be organized.

packages/story-editor/src/components/canvas/frameElement.js
mohdsayed: This seems to have improved now

**Before:**

before_be6a5ba


**After:**

after_be6a5ba


Tested by
- adding multiple kinds of elements.
- selected one of them
- started Profiling
- selected any unselected element
- stoped profiling

mohdsayed: Results from the `main` branch

main

swissspidy:
There are some errors in Karma tests:

```
Warning: Failed %s type: %s%s prop The prop `isCurrentResourceProcessing` is marked as required in `ImageDisplay`, but its value is `undefined`.

...

Uncaught TypeError: isCurrentResourceProcessing is not a function thrown
```

swissspidy (approved):
Took a while to go through all files, but of course not as long as it took you to get this up and running. Kudos! 👏 🚀 

As mentioned, there are some failing tests, but I think the rest is looking good so far. And we do have some follow-up tickets already file thanks to you, which is great.

swissspidy:
`element-library` does not have a dependency on `design-system` and I don't think it should necessarily have one. Let's define a `const noop = () => {}` within that package. Really not a util we should need the design system for.

mohdsayed:
> `element-library` does not have a dependency on `design-system` and I don't think it should necessarily have one. Let's define a `const noop = () => {}` within that package. Really not a util we should need the design system for.

It actually does have a dependency on `design-system`. The error was because I had mistakenly used `@web-stories-wp` namespace. 

We are using a mix of local and `design-system` import of `noop` in `story-editor`. We should choose one, what do you prefer?

swissspidy:
> We are using a mix of local and `design-system` import of `noop` in `story-editor`. We should choose one, what do you prefer?

Each package can easily have its own local `noop`.

packages/story-editor/src/output/element.js
miina: @sayedtaqui Looks like this still needs a follow-up issue to be created.

miina (approved):

packages/story-editor/src/output/element.js
mohdsayed: I missed this one. Created here https://github.com/GoogleForCreators/web-stories-wp/issues/10917 thank you!

packages/story-editor/src/components/dropTargets/provider.js
```diff
@@ -19,16 +19,23 @@
...
+  MEDIA_ELEMENT_TYPES,
+} from '@googleforcreators/elements';
+/**
+ * Internal dependencies
+ */
+import { noop } from '../../utils/noop';
 
 /**
  * Internal dependencies
  */
```
swissspidy: Nit: duplicate doc blocks

swissspidy:
Looks like `packages/story-editor/src/components/footer/carousel/karma/carouselNavigation.karma.js` needs to be updated

```
Carousel Navigation should duplicate the first page FAILED
Error: Unknown element type: shape
```

mohdsayed:
@swissspidy I realized that we were blinding pushing element types to `elementTypes` using `registerElementType` which was temporarily fine for story editor because we register element types when component mounts ( in `StoryEditor` component ), however since `StoryEditor` initializes multiple times in integration tests, the `elementTypes` array gets filled with too many duplicate element types. Therefore I have changed the `elementTypes` variable from `array` to `object`, so that we never have duplicate element types and it can also be overridden easily later.

maxyinger (approved):
Only gave this a quick perf check since last time, but looks like all the perf regressions have been taken care of  🎉 

Thanks for putting in all this work. Nice lift 💪

PR #10458, opened 2022-02-03
Editor: Autoset media as bg if user scales it to fit the page

### LINKED ISSUE #5356: Autoset image/video as bg if user scales it to fit the page
## Feature Description

Autoset image/video as bg if image/video is stretched to fill the entire page. Impetus for doing this:

1. Prevent users from applying link to an asset that covers the entire page
2. Allow user to set only page color or image and not both. When a full bleed image is place on top of a page and a user attempts to set a page color, there has been confusion as to why the page color has not been set. This is caused by the fact that the full bleed image is obscuring the page.
3. Page can always be selected. When a user sets a page attachment and has a an image that is fullbleed, it is not possible to select the page if they want to edit/remove the page attachment.
4. Performance issues: Avoid situation where a full bleed bg image is set over a bg image

Todo:

- [x] If the page does not have background media, detect if a video/image is resized so that it covers all edges. If yes, set it as a background automatically.
- [x] In terms of keeping properties, it will behave the same way as currently setting media as background.
- [x] Make sure the history works as expected: if the user clicks Undo, it should undo only setting as background.
- [x] Add Snackbar message: "Full bleed images and videos are automatically set as background. Double click to scale and position at any time."

---

merapi:
## To-do

- [x] Snackbar message
- [x] Tests

## Testing Instructions

Every time the media element (image, video, gif) will be resized (or moved) so that it covers the whole canvas - it will be set as background media.
If there is already background media on a page - it will not be set.

1. Add image
2. Resize it to cover the canvas
3. Should be auto-set as the background image

## Checklist

---

Fixes #5356


packages/story-editor/src/utils/isTargetCoveringContainer.js
```diff
@@ -0,0 +1,47 @@
...
+  // We can also use unionRatio, the difference is: when the element is bigger
+  // than the container then coverRatio = 1, while unionRatio will be < 1 and only
+  // be 1 if they both are equal in size/position.
+  // const targetArea = targetBox.width * targetBox.height;
+  // const unionArea = targetArea + containerArea - intersectionArea;
+  // const unionRatio = intersectionArea / unionArea;
+  // return unionArea > 0.99;
+
+  return coverRatio > 0.995;
+}
```
merapi: You can uncomment this code and check the alternative solution, right now it will be set as BG even if the image is bigger than canvas, alternatively, it can be set as BG only if all edges ~match (0.99+ is equal to resizing the image with the current edge snapping).

miina: > right now it will be set as BG even if the image is bigger than canvas

That's the intention here -- whenever it's bigger than canvas, we should set it.

The reason for this is that (some) users are not aware that they can set background and instead of setting a background and scaling, they resize media to cover the page and then resize it even more to scale it 😛

packages/story-editor/src/components/canvas/singleSelectionMoveable/useResize.js
```diff
@@ -146,6 +150,9 @@ function useSingleSelectionResize({
     if (handleElementOutOfCanvas(target)) {
       return;
     }
+    if (handleFullbleedMediaAsBackground(target)) {
```
miina: We might need to add this after updating the selected element: otherwise, if the user clicks Undo, the pre-resizing state will be restored, however, only setting as background should be reverted.

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Internal dependencies
+ */
+import isTargetCoveringContainer from '../../../utils/isTargetCoveringContainer';
+import { useStory, useCanvas } from '../../../app';
+
+function useFullbleedMediaAsBackground({ selectedElement }) {
+  const { setBackgroundElement, currentPage } = useStory((state) => ({
+    setBackgroundElement: state.actions.setBackgroundElement,
+    currentPage: state.state.currentPage,
```
miina: Since we're only using the `isDefaultBackground`, we could get that directly instead of getting the whole `currentPage`, e.g. 
```
const { setBackgroundElement, isBackgroundElement } = useStory((state) => ({
    setBackgroundElement: state.actions.setBackgroundElement,
    isBackgroundElement: state.state.currentPage?.elements[0]?.isDefaultBackground,
...
```

If I'm not mistaken then @littlemilkstudio has done quite a bit of performance work to improve how we're using the selectors and reduce re-rendering, and that's one part of it -- whenever possible, to select precisely what we need instead of the whole object (including having some logic inside the selectors, example):
https://github.com/GoogleForCreators/web-stories-wp/blob/597789a61e47ee739038e62d3234de7485ee09cb/packages/story-editor/src/components/canvas/selection.js#L30-L34

 (@littlemilkstudio to clarify / correct me if I'm lying about something here 😛 )

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
```diff
@@ -0,0 +1,52 @@
...
+    ({ state: { fullbleedContainer } }) => ({
+      fullbleedContainer,
+    })
+  );
+
+  const handleFullbleedMediaAsBackground = (target) => {
+    if (!currentPage?.elements[0]?.isDefaultBackground) {
+      return false;
+    }
+    if (selectedElement.type !== 'image' && selectedElement.type !== 'video') {
```
miina: Could also be `gif`.

packages/story-editor/src/components/canvas/singleSelectionMoveable/useResize.js
merapi: I was wondering to what state it should undo because when you drop media as BG it will undo to a different state (post-move/pre-bg, and not to the initial state which was pre-move [state the video started with]):

https://user-images.githubusercontent.com/21036224/152511777-fdbc618d-b6f4-4343-af8e-153d93045147.mov

Is that drop as BG undo state correct or not?

packages/story-editor/src/utils/isTargetCoveringContainer.js
```diff
@@ -0,0 +1,47 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+function isTargetCoveringContainer(target, container) {
+  const targetBox = target.getBoundingClientRect();
+  const containerBox = container.getBoundingClientRect();
+
+  const intersectionArea =
```
miina: Hm, wondering about the rotated media, does this consider rotation as well?

We're also using [`sat`](https://www.npmjs.com/package/sat) library in the editor (search `import SAT from 'sat';` for use cases) which might have something that could calculate this for us, e.g. there seems to be `aInB` which says if polygon A is fully inside polygon B -- might be worthwhile checking.

packages/story-editor/src/components/canvas/singleSelectionMoveable/useResize.js
miina: > Is that drop as BG undo state correct or not?

Yep, that should be correct (if I understand correctly) -- the move / resizing stays but just setting the background is undone -- that's the information I have from UX at least.

The only oddity with this will be that every time the user moves it again, it will be set as background again as long as it covers the whole page. 🤔

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
merapi: Is it worth tieing to `const DROP_SOURCE_ALLOWED_TYPES = ['image', 'gif', 'video'];` (`DROP_SOURCE_ALLOWED_TYPES.includes(selectedElement.type)`)?

miina: Yep, maybe even renaming it to `MEDIA_ELEMENT_TYPES` or something similar, and moving into the general constants file (we have `constants/media.js` for example)? Looks like it's used only in one place currently so would be a small change.

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
merapi: I was wondering if there will be any gain here, you know, re-rendering parents would wipe out those gains, e.g. singleSelectionMoveable is re-rendering, it looks like there is no gain, but I will do that in case I'm wrong. Maybe Max will shed some light here.

packages/story-editor/src/utils/isTargetCoveringContainer.js
merapi: The question is, how should it work? because there is no rotation in the page background element, so maybe rotated images shouldn't be auto-set 🤔

merapi: The more I'm thinking about it the more I'm convinced that it shouldn't autoset rotated images. Imagine someone carefully resizing the rotated image to cover the whole canvas and then realizing that the image is straightened up.
@aaskedall @agingoldseco WDYT?

packages/story-editor/src/components/canvas/singleSelectionMoveable/useResize.js
merapi: Will this be ok? 86407ba5ab8b34b04e53f628068868e3c3e72a47

https://user-images.githubusercontent.com/21036224/152532863-0f4c3703-ede1-4817-b08b-b60d0505e537.mov

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
merapi: Should be fine https://github.com/GoogleForCreators/web-stories-wp/pull/10458/commits/8601f296e9df69e05a92d96b0809119a5010203a ?
I will do the local change next once you are ok with this^.

miina: Yep, lgtm!

packages/story-editor/src/components/canvas/singleSelectionMoveable/useResize.js
miina: Can't see how many times you click on the Undo button but if I understand correctly then yes -- this should be correct -- when clicking "Undo" the first time, the media will stay larger than the Page but not be background anymore.

packages/story-editor/src/utils/isTargetCoveringContainer.js
andrew-askedall: My ux gut is that we probably should have a way for users to rotate a bg image once it's set as bg. This would then allow for a user to set a rotated image as bg. What do you think about adding this feature?

merapi: Agree, we will probably also want to add BG color when media is set (I saw the discussion about it somewhere?). So it will take a while, until then, we can:
1) leave it as is 
2) prevent auto-set for rotated media

You can check how it works now on [QA env](https://stories-qa-wordpress-amp.pantheonsite.io/).

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
merapi: 👍🏻  https://github.com/GoogleForCreators/web-stories-wp/pull/10458/commits/75112b51f4b56c16021396fbc06e57c3e74fa7c3

packages/story-editor/src/components/canvas/singleSelectionMoveable/useResize.js
merapi: > when clicking "Undo" the first time, the media will stay larger than the Page but not be background anymore.

✅

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
miina: Even if there's no gain specifically here, it might still be better to use this as a new best practice everywhere -- probably no point getting the whole object if we only need one property from it, especially if in some places it will be beneficial for preventing re-rendering as well. Thoughts?

packages/story-editor/src/utils/isTargetCoveringContainer.js
miina: > The more I'm thinking about it the more I'm convinced that it shouldn't autoset rotated images. Imagine someone carefully resizing the rotated image to cover the whole canvas and then realizing that the image is straightened up.

Makes sense! Especially if the plan is to support rotated background as well later.

swissspidy: IIRC in the past we deliberately decided to not support rotated background elements due to potential issues that could arise from that. Just to keep that in mind when creating a ticket for adding support for this. Personally I've never seen any user requests for this, so perhaps we could even save us this hassle :-)

andrew-askedall: With this change we could leave as is and when it's sized to the background it will "unrotate" to and see if we get any user feedback or insight if this is a real problem or a hypothetical.

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
```diff
@@ -0,0 +1,54 @@
...
+  const { fullbleedContainer } = useCanvas(
+    ({ state: { fullbleedContainer } }) => ({
+      fullbleedContainer,
+    })
+  );
+
+  const handleFullbleedMediaAsBackground = (target) => {
+    if (!isDefaultBackground) {
+      return false;
+    }
```
barklund: This confused me a lot for several minutes. I thought it checked it the selected element or the target was a default background, which didn't make a lot of sense. But it reflects whether the actual background on the current page is a default background and thus not a media element. Can we rename `isDefaultBackground` a bit to reflect this? Maybe just `pageHasDefaultBackground` or something like that?

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
```diff
@@ -0,0 +1,54 @@
...
+    if (!MEDIA_ELEMENT_TYPES.includes(selectedElement.type)) {
+      return false;
+    }
+    if (isTargetCoveringContainer(target, fullbleedContainer)) {
+      setBackgroundElement({ elementId: selectedElement.id });
+      return true;
+    }
+    return false;
```
barklund: Return value doesn't seem to be used anywhere. Can we just skip it?

packages/story-editor/src/utils/isTargetCoveringContainer.js
barklund: To go back to @miina's original comment, I think using `sat` here makes sense just to avoid stupid math mistakes - we already have the lib loaded anyway.

barklund (approved):
Looks good pending the discussion on rotated media.

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
miina: @barklund `isDefaultBackground` is something that's used across the whole codebase currently to check if background is media or the default shape. This prop name came already from https://github.com/GoogleForCreators/web-stories-wp/pull/2387 if I remember correctly.

Probably just leaving a comment for now is sufficient, e.g. `// Return false if background media`.

If you feel like changing the prop generally, we should create a code quality issue for it instead (it would influence a looot of files 😛).

miina: I just realized that maybe you meant just changing the name in the selector above and not generally in the code, ignore my previous comment in this case! 🐶

packages/story-editor/src/utils/isTargetCoveringContainer.js
merapi: @barklund @miina 
I really wanted to use this `Separating Axis Theorem` 😎  
I did the implementation, but for me, it's too easy to create a state where it doesn't detect [`aInB`](https://github.com/jriecken/sat-js#satresponse), that's probably because our snapping is not pixel perfect (and that's why I have `coverRatio > 0.995`). `overlap...` props don't provide anything helpful in this case (it's just one [shortest colliding] axis info).

You can pull and check it (scale the canvas while testing).

image

image

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
merapi: > This confused me a lot for several minutes.

Not only you 😉  https://github.com/GoogleForCreators/web-stories-wp/pull/10261#discussion_r789205005
I feel like we should change it globally (`isDefaultBackgroundType` is more clear to me because it's just about type: shape|media), but you probably meant to change it just here?

packages/story-editor/src/utils/isTargetCoveringContainer.js
miina: @merapi Could we solve this issue by leaving a buffer when creating a polygon to compare?
E.g. `const buffer = 2;` and then when creating the page area polygon, deduct/add `buffer` to the points? Or would rotation mess this up?

barklund: Hm, yeah, I see the point. I'm also okay with keeping this as-is. It is not the hardest math ever. Probably won't win you a Fields Medal anytime soon.

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
merapi: 👍🏻   https://github.com/GoogleForCreators/web-stories-wp/pull/10458/commits/7593fb1c382b0e7b49fd7e0bae5990b64b5b6169

packages/story-editor/src/utils/isTargetCoveringContainer.js
merapi: That buffer/padding is an ingenious/hacky idea 💡  it will solve our snapping precision issue.

Rotation: looks like something is off with our rotation calculations (or this lib?):

https://user-images.githubusercontent.com/21036224/153027484-22845bb2-12d1-4689-80a1-dc3f1b1f105a.mov

miina: Hm, odd. We're also using it for the lasso selection, is that also not working correctly then? 🤔 
Probably need a bug issue for investigating. Perhaps `pointInPolygon` (used for layer selection) is not working as expected with rotation but `testPolygonPolygon` (used for lasso) is. Or maybe the library is somehow incorrectly used for layer selection.

merapi: I've quickly checked the lasso - seems fine, but layer selection is indeed broken when dealing with rotated objects.
Small debug tool (transparent area - used in SAT calculations, dashed box - regular bb):

https://user-images.githubusercontent.com/21036224/153068858-0159d4a8-b1c2-40d1-94e1-3305ad501c5c.mov

Let's handle it here: https://github.com/GoogleForCreators/web-stories-wp/issues/10519
After fixing that^ I can also add rotation support here (right now it's 0 because it's hard to say what's wrong).

package.json
```diff
@@ -149,7 +149,7 @@
     "postcss-jsx": "^0.36.4",
     "postcss-syntax": "^0.36.2",
     "prettier": "^2.5.1",
-    "puppeteer": "^13.2.0",
+    "puppeteer": "13.3.1",
```
swissspidy: `^13.3.1` should work just as well, no?

What about the lock file?

merapi: Yep, I just did a normal install now 👍🏻

PR #10460, opened 2022-02-03
Design System: Popup Updates to Eliminate Redundant Custom Popup

### LINKED ISSUE #10089: Deduplicate Popup

BrittanyIRL:
## Context

There were a few places in the editor where a different `Popup` (internal to the editor) was getting used. This updates the shared `Popup` in the design system to consider these needs so we can eliminate the extra implementation. 

## Summary

Pretty much the editor's implementation of Popup existed because it needs to know if the editor's in RTL mode and find the topOffset of its CMS. 

## Relevant Technical Choices

New props for the design system's Popup are: `isRTL, topOffset, invisible` 
- invisible is only ever used in the color input.
- topOffset (which is a bummer of a name because we have an internal `$offset` that's getting used to decide positioning but I didn't change that) is for when a popup is maybe going to reach the top of the page so it needs an extra barrier because the found y min is actually wrong (app starts, say, 32px below the top of the browser window. 
- isRTL flips the y transforms as needed 

The style updates to the Popup's container mean that the scrollbar css isn't needed anymore.

## User-facing changes

| Component | Staging - LTR | Staging - RTL | New - LTR | New - RTL |
| --- | --- | --- | --- | --- | 
| Page Attachment |  link LTR | link RTL | new link LTR | new link RTL |
| Color Input | color picker LTR | color picker RTL | new color picker LTR | new color picker RTL |
| Play Pause | play LTR | play RTL | new play LTR | new play RTL | 
| Style Presets | more styles LTR | more styles RTL | new more styles LTR | new more styles RTL | 

So, hopefully, if i did it right - no changes. 

## Testing Instructions

Unfortunately, the `Popup` in the design system touches a bunch of stuff. Regression testing on the following: 
- Tooltip behavior (be mindful of tooltips that change location when a button is interacted with while they are open - like locked vs unlocked), scroll, when expected space isn't available) 
- Drop downs 
- Editor: style presets, color input, play pause button, links in canvas
- Dashboard: stories view tooltip toggle 

1.

## Checklist

---

Fixes #10089


packages/design-system/src/components/popup/index.js
```diff
@@ -91,7 +115,7 @@ function Popup({
   }, []);
 
   const positionPopup = useCallback(
-    async (evt) => {
+    (evt) => {
       if (!isMounted.current) {
```
BrookeGraham, line 119: Let's add a check here for `!anchor?.current` so we can return early and not blowup when `anchor`s first pass is still null

BrookeGraham (commented):
PageAttachement needs {isRTL} but other than that I think this is very clean and didn't see any other regressions.

swissspidy:
Uhm what about #10197?

BrittanyIRL:
> 

I know. I was getting a little confused about why such a big diff when this was supposed to just de duplicate a Popup that has 3 additional values passed to it so i wanted to do a proof of concept. Brooke gave this a look really quick before I was going to share it in that PR. The popups have caused us a handful of one-off bugs (in both implementations) that have been worked out. My PR here has about half the diff of #10197, which isn't to say that it wouldn't end up working fine - it's just that the popups are one of the trickiest bits and they touch SO MANY components. Wanted to propose a different solution.

amovar18:
Because `overflow-y` is not controlled the popup comes with a scrollbar other than that it looks cleaner than my PR
Screenshot 2022-02-04 at 9 55 39 AM
Edit: I think the `MediaInput` of design-system also need the `isRTL` prop since it uses `Popup`

BrookeGraham:
> Because `overflow-y` is not controlled the popup comes with a scrollbar other than that it looks cleaner than my PR Screenshot 2022-02-04 at 9 55 39 AM

Interesting. It is cutting off the `tail`  Nice catch. Though I only see the scroll bar when I scroll down to see the tail. I am just curious what browser you are using @amovar18

amovar18:
> > Because `overflow-y` is not controlled the popup comes with a scrollbar other than that it looks cleaner than my PR Screenshot 2022-02-04 at 9 55 39 AM
> 
> Interesting. It is cutting off the `tail` Nice catch. Though I only see the scroll bar when I scroll down to see the tail. I am just curious what browser you are using @amovar18

I am using chrome 
Edit: version 98.0.4758.80

BrittanyIRL:
@amovar18 updated his version of this to sync with my idea here and his solves overflow better so I am closing this.

Issue #10375, opened 2022-01-25
Update Storybook actions to come from essentials add-on


BrookeGraham:
## Feature Description

Now that we have upgrades Storybook and are using `controls` from the `Essentials` add-on package, we can remove packages that are included in the `essentials` add-on. Let's update all uses of `actions` in Storybook's stories to come from the `essentials` add-on package. This will allow us to remove the separate actions add-on package entirely.

PR #10461, opened 2022-02-03
Storybook: Dashboard `actions` update

BrookeGraham:
## Context

We want to use Storybook's actions from the `essentials add-on` package. Removing the need for the `actions add-on` package. 

## Summary

New actions (and controls) for Dashboard stories. 

We are updating all Stories that use `action` via the action add-on and are now using action from the `essentials add-on`. We now define `action` within the `argTypes` when declaring the story. 

## Relevant Technical Choices
Removed the `actions add-on` devDep for the Dashboard package since all stories in that package are now using `actions` from the `essentials add-on`

Updated `formattedTemplatesArray.[i]modified` date to use `/` instead of `-` the `toUTCDate` didn't like the dash was breaking the story because it wasn't formatting the date properly. 

## To-do

- [x] WP-Dashboard see #10462 
- [x] WP-Story-Editor see #10462 
- [x] Story-Editor see #10499 
- [x] Design-System see #10520 
- [x] Remove `"@storybook/addon-actions": "^6.4.14"` from package.json see #10550 

## Testing Instructions

1. Fire up Storybook and check out the actions tab under `Dashboard`
Stories to look at:
Dashboard/Views/ExploreTemplates/Content
Dashboard/Views/ExploreTemplates/Header
Dashboard/Views/TemplateDetails/Content/DetailsContent
Dashboard/Views/TemplateDetails/Header
Dashboard/Views/MyStories/Content
Dashboard/Views/MyStories/Header

no
no

no

## Checklist

---

Partially addresses #10375

### DIFF
--- a/packages/dashboard/src/app/views/exploreTemplates/content/stories/index.js
+++ b/packages/dashboard/src/app/views/exploreTemplates/content/stories/index.js
@@ -17,7 +17,6 @@
 /**
  * External dependencies
  */
-import { action } from '@storybook/addon-actions';
 import styled from 'styled-components';
 
 /**
@@ -25,43 +24,9 @@ import styled from 'styled-components';
  */
 import { Layout } from '../../../../../components';
 import { VIEW_STYLE } from '../../../../../constants';
-import {
-  STORYBOOK_PAGE_SIZE,
-  formattedTemplatesArray,
-} from '../../../../../storybookUtils';
+import { formattedTemplatesArray } from '../../../../../storybookUtils';
 import { usePagePreviewSize } from '../../../../../utils';
 import Content from '..';
-
-const search = {
-  keyword: '',
-  setKeyword: action('set search'),
-};
-const view = {
-  style: VIEW_STYLE.GRID,
-  pageSize: STORYBOOK_PAGE_SIZE,
-};
-const page = {
-  value: 1,
-  set: action('set page number'),
-  requestNextPage: action('request next page clicked'),
-};
-
-const templateActions = {
-  createStoryFromTemplate: action('create story from template clicked'),
-  handleDetailsToggle: action('modal was toggled'),
-  switchToTemplateByOffset: action('switched to prev/next template'),
-};
-
-const defaultProps = {
-  allPagesFetched: false,
-  isLoading: false,
-  page: page,
-  search: search,
-  templates: formattedTemplatesArray,
-  view: view,
-  totalTemplates: 3,
-  templateActions,
-};
 
 const StorybookLayoutContainer = styled.div`
   margin-top: 40px;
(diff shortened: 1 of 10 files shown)


packages/dashboard/src/app/views/myStories/content/stories/index.js
```diff
@@ -89,136 +47,152 @@ const longerListOfStories = forceUniqueIds(
...
-            allPagesFetched
-            view={{ ...view, pageSize }}
-          />
-        </StorybookLayoutContainer>
-      </Layout.Provider>
-    </SnackbarProvider>
-  );
+export const NoStories = _default.bind({});
+NoStories.args = {
+  stories: [],
```
BrookeGraham, line 176: I am perfectly happy to make these extra stories for these specific details, but I feel like the idea behind `controls` is to allow the user to switch between these kind of option in order to get the desired view without having to have another story to look at. Do y'all think we should scrap these additional stories and just use `control` to show the various uis?

packages/dashboard/src/components/storyMenu/stories/index.js
```diff
@@ -51,8 +50,7 @@ export const _default = () => {
       <StoryMenu
         onMoreButtonSelected={setContextMenuId}
         contextMenuId={contextMenuId}
-        onMenuItemSelected={(item, story) => {
-          actions('onClick ', item.label, story.id);
+        onMenuItemSelected={() => {
```
BrookeGraham, line 53: I just removed this action as a whole because I don't think we can trigger `onMenuItemSelected` from this story.

BrookeGraham:
There are many more actions defined than what we can trigger via storybook. I was torn between removing them altogether when we can't trigger them. Or leave them as it so we can at least the props. Thoughts?

packages/dashboard/src/app/views/exploreTemplates/content/stories/index.js
```diff
@@ -71,59 +36,88 @@ const StorybookLayoutContainer = styled.div`
...
+        'search',
+        'searchKeyword',
+        'view',
+        'templateActions',
+      ],
+    },
+  },
 };
-export const _default = () => {
+export const Default = (args) => {
```
BrookeGraham: Change back to `_default` for consistency.

packages/dashboard/src/app/views/myStories/content/stories/index.js
BrittanyIRL: i think scrapping the additional stories to use controls is a great plan!

BrittanyIRL (approved):
Updates all run in storybook well. Love the idea of leaning more into controls to remove duplicative code.

swissspidy (approved):

PR #10463, opened 2022-02-04
Move to `react-photo-album`

### LINKED ISSUE #10444: Replace `react-photo-gallery`
## Feature Description

The `react-photo-gallery` package we're using for the media library hasn't been updated in over 3 years.

We already had to patch it once in #4438 because of the outstanding bugfix in https://github.com/neptunian/react-photo-gallery/pull/184 that hasn't been merged to this day.

Another issue is that the library doesn't officially declare support for React 17, causing some issues with duplicated React dependencies (https://github.com/neptunian/react-photo-gallery/issues/205).

I think it's time we look for alternatives.

One often named alternative is https://github.com/igordanchenko/react-photo-album which has a very similar API. Docs: https://react-photo-album.com/

My hope is that it's gonna be almost a drop-in replacement.

ayushnirwal:
## Context

Previously [react-photo-gallery](https://www.npmjs.com/package/react-photo-gallery) was used to display media elements in the library pane, it has not seen any change since 1 July 2019. Therefore it was stuck at React version `^16.8.0` which will cause 2 versions of React to be installed if are packages published.

## Summary

[Documentation](https://react-photo-album.com/documentation) lacked any explanation of related to the props passed to `renderPhoto` ( analogous to `renderImage` in `react-photo-gallery` ). So below is an example of its props.

```js
{
    "photo": {          // photo obj required to have key, source, width and height
        "index": 0,     // additional param added get index
        "key": 85,
        "src": "http://webstories.local/wp-content/uploads/2022/01/small-video.webm",
        "width": 560,
        "height": 320
    },
    "layoutOptions": {
        "layout": "rows",
        "viewportWidth": 1174,
        "containerWidth": 292,
        "columns": 2,
        "spacing": 4,
        "padding": 0,
        "targetRowHeight": 110
    },
    "layout": {
        "height": 81.63779527559055,  //  calculated height which can be passed down to MediaElement
        "width": 142.86614173228347,  //  calculated width which can be passed down to MediaElement
        "photoIndex": 0,              // this represents index in the current row
        "photosCount": 2            // no of photos in current row
    },
    "imageProps": {
        "src": "http://webstories.local/wp-content/uploads/2022/01/small-video.webm",
        "alt": "",
        "style": {
            "display": "block",
            "boxSizing": "content-box",
            "width": "calc((100% - 4px) / 2.01587)",
            "height": "auto",
            "aspectRatio": "560 / 320"
        },
        "className": "react-photo-album--photo",
        "sizes": "13vw"
    }
}
```
`react-photo-album`
1p-react-photo-album

`react-photo-gallery ( previously used )
1p-react-photo-gallery

## Relevant Technical Choices

- index param was introduced to `photo` obj since there was no info about that in `renderPhotos`'s props 
- `targetRowHeight` was set to `150` to make the end result similar.
- remove mock for default export from `react-photo-album` in `Media3pPane` unit test suite.

## Testing Instructions

1.

## Checklist

---

Fixes #10444


swissspidy:
Oh neat, thanks for picking this up!

So this is really mostly a drop-in replacement, which is great!

Curious, would the columns layout perhaps work too and cause the same results? Then we might not need to specify `targetRowHeight`. But I suppose we were using rows layout for a good reason... So maybe not worth testing.

Also, is there any noticeable performance improvement by any chance?

ayushnirwal:
> Also, is there any noticeable performance improvement by any chance?

Sadly no.
> So this is really mostly a drop-in replacement, which is great!

Using it seems similar but it does break unit tests in `packages/story-editor/src/components/library/panes/media/common/test` with error `TypeError: isCurrentResourceProcessing is not a function`.

swissspidy:
> Using it seems similar but it does break unit tests in `packages/story-editor/src/components/library/panes/media/common/test` with error `TypeError: isCurrentResourceProcessing is not a function`.

Since you changed/removed the mock in the unit test we now have to mock `isCurrentResourceProcessing` in the tests as this is used by the single element being rendered.
I can take a quick look if you want.

ayushnirwal:
> I can take a quick look if you want.

Sure. Let me just push my changes to fix `editor/src/components/library/panes/media/common/test/paginatedMediaGallery.js`

swissspidy:
Looks like you're on the right track. Just need to make similar change to `packages/story-editor/src/components/library/panes/media/media3p/test/accesibility.js`

(note: while at it, can also rename the file from `accesibility.js` to `accessibility.js` to fix the typo.

spacedmonkey:
I am seeing very different results from original. 

### react-photo-gallery

Screenshot 2022-02-04 at 11 25 28

Screenshot 2022-02-04 at 11 25 47

### react-photo-album

Screenshot 2022-02-04 at 11 24 00

Screenshot 2022-02-04 at 11 24 11


I don't think we should every have 3 items in one row.

swissspidy:
3 items in a row isn't a big deal, but those large ones on a single line don't look great.

Maybe we should indeed try the columns option to see if that works better.

Let's play a little bit with the config options.

packages/story-editor/src/components/library/panes/media/common/mediaGallery.js
```diff
@@ -44,39 +44,42 @@ const PHOTO_MARGIN = 4;
...
   );
 
   return (
     <div>
       <Gallery
-        targetRowHeight={110}
-        direction={'row'}
-        // This should match the actual margin the element is styled with.
-        margin={PHOTO_MARGIN}
+        layout="rows"
```
spacedmonkey, line 86:
```suggestion
        layout="columns"
```

packages/story-editor/src/components/library/panes/media/common/mediaGallery.js
```diff
@@ -44,39 +44,42 @@ const PHOTO_MARGIN = 4;
...
         photos={photos}
-        renderImage={imageRenderer}
+        renderPhoto={imageRenderer}
+        targetRowHeight={150}
```
spacedmonkey:
```suggestion
        columns={2}
```

spacedmonkey:
Screenshot 2022-02-04 at 12 13 53

This is what columns looks like. I kind of like it.

swissspidy:
Let's not decide too quickly.

> This is what columns looks like. I kind of like it.

Doesn't it have accessibility considerations because items are now laid out differently? (in columns instead of rows)

Let's please test combinations of various settings first before changing from rows (which we have always used) to columns

spacedmonkey:
I tried a number of different layouts to see what looks best. It is true, that the row layout is as similar the currently layout. But  it doesn't  mean we can't use column / masonry. We may want to do this in a follow up PR. 

## Layouts

#### Row layout, row height 150

Screenshot 2022-02-07 at 11 14 31
Screenshot 2022-02-07 at 11 14 03


#### masonry layout, columns 2

Screenshot 2022-02-07 at 11 18 45
Screenshot 2022-02-07 at 11 18 27

#### column layout, columns 2

Screenshot 2022-02-07 at 11 24 23
Screenshot 2022-02-07 at 11 20 28

spacedmonkey:
It is also worth noting that accessibility / keyboard navigation seems to be an issue in both column and row based layouts. 

I tried using keyboard navigation in both layouts, pressing arrows left and right, resulted in just going up and down in column layout and opposite in row layout. 

#### Row layout
https://user-images.githubusercontent.com/237508/152782754-acb983b6-f86e-4f57-ad22-af2622564b5b.mov

#### Column layout
https://user-images.githubusercontent.com/237508/152782827-8ce6d067-6fc4-4a95-baac-a6d52fd850f9.mov

ayushnirwal:
> It is also worth noting that accessibility / keyboard navigation seems to be an issue in both column and row based layouts.

This can be fixed by making custom row and container renderers ( see [commit](https://github.com/GoogleForCreators/web-stories-wp/pull/10463/commits/40dc8672f6a20b82ba70660d0fb23991ff0ca32e) ).

swissspidy:
> > It is also worth noting that accessibility / keyboard navigation seems to be an issue in both column and row based layouts.
> 
> This can be fixed by making custom row and container renderers ( see [commit](https://github.com/GoogleForCreators/web-stories-wp/pull/10463/commits/40dc8672f6a20b82ba70660d0fb23991ff0ca32e) ).

I think it helps for the rows layout but not for columns.

----

Here's my current summary of things:

1. Keyboard navigation is an issue for columns/masonry layout due to different DOM structure. So I don't think these work for us.
2. For this reason, the rows layout is much much preferable.
3. In the rows layout, we want to avoid 3 items per row. However, I don't see an easy way to achieve that. @ayushnirwal Do you perhaps see a solution for that? We could patch `react-photo-album` if needed
4. In the rows layout, I often see images being stretched. Does anyone else see this? Can we avoid this?

packages/story-editor/package.json
```diff
@@ -66,7 +66,8 @@
     "react-blurhash": "^0.1.3",
     "react-calendar": "^3.6.0",
     "react-color": "^2.19.3",
-    "react-photo-gallery": "^8.0.0",
+    "react-moveable": "^0.30.3",
```
spacedmonkey: ?

packages/story-editor/src/components/library/panes/media/common/mediaGallery.js
```diff
@@ -44,39 +44,64 @@ const PHOTO_MARGIN = 4;
...
+      <div
+        ref={ref}
+        {...containerProps}
+        style={{
+          ...containerProps.style,
+          flexDirection: 'row',
+          flexWrap: 'wrap',
+          justifyContent: 'space-evenly',
+        }}
+      >
```
spacedmonkey: Should this be a styled component?

spacedmonkey:
Keyboard navigation is working well!

packages/story-editor/src/components/library/panes/media/common/mediaGallery.js
```diff
@@ -52,21 +53,8 @@ function MediaGallery({ resources, onInsert, providerType, canEditMedia }) {
...
-          flexDirection: 'row',
-          flexWrap: 'wrap',
-          justifyContent: 'space-evenly',
-        }}
-      >
-        {children}
-      </div>
-    );
+  const containerRenderer = ({ children }, ref) => {
+    return <GalleryContainer ref={ref}>{children}</GalleryContainer>;
```
spacedmonkey: Is `containerProps.style` needed here?

ayushnirwal: `containerProps.style` has 4 properties, 3 needed to be overridden ( except `display` ).

packages/story-editor/src/components/library/panes/media/common/mediaGallery.js
```diff
@@ -44,39 +45,51 @@ const PHOTO_MARGIN = 4;
...
-        direction={'row'}
-        // This should match the actual margin the element is styled with.
-        margin={PHOTO_MARGIN}
+        layout="rows"
         photos={photos}
-        renderImage={imageRenderer}
+        renderPhoto={imageRenderer}
+        renderRowContainer={rowRenderer}
+        renderContainer={forwardRef(containerRenderer)}
+        targetRowHeight={150}
```
timarney: Should probably move the  `150` out to a `const` line how PHOTO_MARGIN} is setup

timarney:
This seems to be working quite well including the keyboard nav

As already mentioned the image layout depending on the image sizes seems to be the largest difference

| PR  |      Current      |  
|----------|:-------------:|
| Screen Shot 2022-02-09 at 6 21 32 PM | Screen Shot 2022-02-09 at 6 27 23 PM |

swissspidy:
FYI I am moving this to draft again because there are still [unanswered questions](https://github.com/GoogleForCreators/web-stories-wp/pull/10463#issuecomment-1033620453) that we need to address.

If we can't find solutions for these we need to move this to the backlog again.

@spacedmonkey @timarney No need for you to code review in the meantime.

ayushnirwal:
> In the rows layout, we want to avoid 3 items per row. However, I don't see an easy way to achieve that. @ayushnirwal Do you perhaps see a solution for that? We could patch react-photo-album if needed

[v 1.7.0](https://github.com/igordanchenko/react-photo-album/releases/tag/v1.7.0) of `react-photo-album` now includes a way to constraint the number of elements per row in row layout. This might help to get the desired layout.

packages/story-editor/src/components/library/panes/media/common/mediaGallery.js
```diff
@@ -44,39 +46,54 @@ const PHOTO_MARGIN = 4;
...
-        margin={PHOTO_MARGIN}
+        layout="rows"
         photos={photos}
-        renderImage={imageRenderer}
+        renderPhoto={imageRenderer}
+        renderRowContainer={rowRenderer}
+        renderContainer={forwardRef(containerRenderer)}
+        targetRowHeight={TARGET_ROW_HEIGHT}
+        rowConstraints={{
+          minPhotos: 2,
```
swissspidy: With `minPhotos: 2` I still get 3 items in a row sometimes.

Ideally we only ever have 2. So `maxPhotos: 2`?

swissspidy:
@ayushnirwal This is amazing! At first I was like "this is a very timely update" and then I realized you made the change. Thanks a bunch! 🎉 

At first glance it looks really promising. With `maxPhotos: 2` I see great results. And keyboard navigation works well too.

ayushnirwal:
> In the rows layout, I often see images being stretched. Does anyone else see this? Can we avoid this?

@swissspidy do you still see this issue?
[igordanchenko](https://github.com/igordanchenko) has left a [comment](https://github.com/igordanchenko/react-photo-album/pull/47#issuecomment-1047831800) which might explain it. I did not find such an issue on my end, but I have made some changes according to their suggestion.

swissspidy:
No I don't! No stretched images or videos here.

igordanchenko:
Apologies for jumping in here, but may I suggest you move custom renderers (including the `forwardRef`) to static scope outside the component? Presently `renderContainer`, `renderRowContainer`, and `renderPhoto` are used as functional components internally, and non-stable components refs force React to tear down and rebuild the DOM tree on every render. This is likely to result in all kinds of performance issues or flickering if PhotoAlbum is re-painted frequently.

swissspidy:
Thanks for the suggestion, it's greatly appreciated!

swissspidy (approved):
This looks pretty good now!

spacedmonkey (approved):

miina:
@swissspidy Looks like some new Karma tests are failing since this PR (at least didn't see these before when checking commits on `main`):
For example:
- `should replace top image when bg image is set and another one is on top`
- `Quick Actions integration background image selected clicking the `Add animation` button should select the animation panel and focus the dropdown`
- `Quick Actions integration background image selected should add animations and filters to the background image, click the 'Reset element' button, clear the animations and filters, then click Undo and reapply the animations and filters`

And some others.

Since these are related to images then seem like valid failures.

swissspidy:
I could swear I checked Kama,  not sure how I missed that 🤔 

I'll look into it, but probably not today. If you have a spare moment to  briefly run them locally to see whee they fail, that'd be helpful. Thanks!

miina:
> I'll look into it, but probably not today. If you have a spare moment to briefly run them locally to see whee they fail, that'd be helpful. Thanks!

At least this is failing locally as well:
`should replace top image when bg image is set and another one is on top`

The Quick Actions tests mentioned above passed locally (but fail on the PRs).

I've actually also mostly finished for today so wouldn't have time to check it in more detail until Monday.

swissspidy:
Apparently what all the failing tests have in common is that they try dragging a media library item to the edge of the canvas to set it as the bg. And that doesn't work anymore.

Screenshot 2022-02-28 at 11 27 14

I can confirm that drop-targets are not (reliably?) working in the editor when dragging straight from the media library.

@miina You know that code well, perhaps something comes to mind?

miina:
> I can confirm that drop-targets are not (reliably?) working in the editor when dragging straight from the media library.

Hm, I can actually see it working reliably locally 🤔 Is it sometimes working and sometimes not in your case?

miina:
It wouldn't work when
- `LibraryMoveable` is not displayed for some reason (e.g. getting the active state -- hovering or focusing -- of the item is not detected)
- `LibraryMoveable` is displayed but something else is covering it, interfering with the events.

swissspidy:
Weird, at first it didn't work at all, but now it seems just fine reliably. Still not working in Karma tests though.

This is the relevant bit:

https://github.com/GoogleForCreators/web-stories-wp/blob/928311c162f7d7f0454364755e689ef099badb33/packages/story-editor/src/components/dropTargets/karma/order.karma.js#L47-L56

miina:
> This is the relevant bit:

Would be great to inspect the elements in DOM in Karma tests while the cursor is above a media item to see if the `LibraryMoveable` is displayed as expected. And if anything covers it now that didn't before.

Perhaps it's possible to take a snapshot at that moment and then inspect the source locally?

miina:
Ah, actually -- do you know if it's about the drop-targets not working or about the element not being dragged at all?
I thought it was about the element not being dragged but I now realize you mentioned that for you the drop-targets were not working 🤔

igordanchenko:
@miina,

> Ah, actually -- do you know if it's about the drop-targets not working or about the element not being dragged at all?

In local tests the element is being dragged just fine.

Screen Shot 2022-03-02 at 1 08 07 AM

But it looks like the issue here is that the `handleDrag` callback has no `dropTargets` as there appears to be a race condition between `InnerElement` and `registerDropTarget`. When `registerDropTarget` is called first, everything is working as expected. But when `InnerElement` is rendered before `registerDropTarget` is called, then `dragHandler` callback references stale `dropTargets` which is empty.

https://github.com/GoogleForCreators/web-stories-wp/blob/7fc8dfef15f5fd356907dd6bd4ab607c1b5debdb/packages/story-editor/src/components/library/panes/media/common/innerElement.js#L115-L132

PR #10469, opened 2022-02-04
New Story Details Modal: Add Experiment

BrittanyIRL:
## Context

Initial wiring of new story publish modal via experiment. 

## Summary

So as to not do the rest of this feature in a vacuum - when  `enableUpdatedPublishStoryModal` is true and you click 'publish' you get the new (unfinished) modal instead of the existing modal for the checklist. 

## To-do

(See https://github.com/GoogleForCreators/web-stories-wp/issues/10115 for full list)

## User-facing changes

N/A

## Testing Instructions

- See that publishing works the same on this branch as on staging until you turn on the aforementioned experiment (`enableUpdatedPublishStoryModal`).
- With the experiment enabled see that when you click "publish" in the editor you get the new modal. You can update the story title and description here and then continue with the "publish" button in the modal and it'll publish like normal. 

Screen Shot 2022-02-04 at 12 27 41 PM

1.

## Checklist

---

Partially addresses #10115


includes/Experiments.php
```diff
@@ -313,6 +313,17 @@ public function get_experiments(): array {
...
 			],
+			/**
+			 * Author: @brittanyirl
+			 * Issue: #10115
+			 * Creation date: 2022-02-02
+			 */
+			[
+				'name'        => 'enableUpdatedPublishStoryModal',
+				'label'       => __( 'Updated Publish Story Modal', 'web-stories' ),
+				'description' => __( 'Enable new modal that shows up on Story publish to make sure necessary info is all present before publishing', 'web-stories' ),
```
swissspidy: That's quite a long description that is hard to read and translate.

I'd aim for something simpler. For example `Enable new pre-publish confirmation modal`

packages/story-editor/src/storyEditor.js
```diff
@@ -58,6 +59,9 @@ function StoryEditor({ config, initialEdits, children }) {
   const _config = useMemo(() => deepMerge(defaultConfig, config), [config]);
   const { storyId, isRTL, flags } = _config;
 
+  // Tell modals where to set aria-hidden when modals are open
+  setAppElement('body > div:first-of-type');
```
swissspidy: Why was this added? 🤔 

We call `setAppElement` already within `wp-story-editor` and within each individual test. Not here.

BrittanyIRL: I added this because I wasn't seeing the children of `<body>` getting treated with `aria-hidden` anymore, which to your point, is because it's nested into `#web-stories-editor`, which I missed. 

I am fine to remove this but before I do - this is where I'm at: 
When we have modals open we want everything else to be hidden for sr/keyboards, this includes WordPress admin actions - which would mean bumping the aria-hidden up to where I'm putting it here. 

If we want to leave access to WordPress, that's fine, I just missed that as a rule 🙃  It makes sense that way and I"m fine with it. I just noted it as missing because I wasn't seeing it and modals are added as direct children to `body` so the inheritance of aria-hidden didn't make sense to me. 

|  #web-stories-editor | body > div | 
| --- | --- | 
| Screen Shot 2022-02-07 at 9 00 14 AM |  Screen Shot 2022-02-07 at 9 02 08 AM |

swissspidy: > When we have modals open we want everything else to be hidden for sr/keyboards, this includes WordPress admin actions - which would mean bumping the aria-hidden up to where I'm putting it here.

That would be a change in current behavior though. If we want to change this, this needs to be done in `wp-story-editor`, not `story-editor`. It's a WordPress-specific issue after all.

Here's where we set it in WordPress:

https://github.com/GoogleForCreators/web-stories-wp/blob/7d364a5633daa5d901bce9cefdbcb0452904cdc1/packages/wp-story-editor/src/index.js#L67-L68

Since this would change existing behavior in the WP editor let's discuss this in a separate ticket to have more visibility

BrittanyIRL: yup sounds good!

[Edit]
Issue created: https://github.com/GoogleForCreators/web-stories-wp/issues/10498

packages/story-editor/src/components/header/buttons/publish.js
```diff
@@ -105,13 +109,21 @@ function PublishButton({ forceIsSaving }) {
   }, [refreshPostEditURL, saveStory, hasFutureDate, title, canPublish]);
 
   const handlePublish = useCallback(() => {
-    if (shouldReviewDialogBeSeen && canPublish) {
+    if (
+      (shouldReviewDialogBeSeen && canPublish) ||
+      (isUpdatedPublishModalEnabled && canPublish)
+    ) {
```
samwhale: 💭  could be simplified to
```suggestion
    if (
       canPublish &&
      (shouldReviewDialogBeSeen ||
       isUpdatedPublishModalEnabled)
    ) {
```

packages/story-editor/src/components/publishModal/publishModal.js
```diff
@@ -43,7 +39,7 @@ const Container = styled.div`
   border-radius: ${({ theme }) => theme.borders.radius.medium};
 `;
 
-function PublishModal() {
+function PublishModal({ isOpen, onPublish, onClose }) {
```
samwhale, line 42: 👍

samwhale (approved):
Makes sense to me

packages/story-editor/src/components/header/buttons/publish.js
BrittanyIRL: Sure!

packages/story-editor/src/components/publishModal/karma/publishModal.karma.js
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  describe('Basic structure', () => {
+    it('should have no aXe accessibility violations', async () => {
+      await expectAsync(publishModal).toHaveNoViolations();
+    });
+  });
+
+  describe('Functionality', () => {
+    it('should only allow publish of a Story when both title and description are not empty', async () => {
```
miina, line 66: Currently, the story does allow publishing without a title and without a description (at least when testing in local) -- are we changing that purposefully? Confirming just in case.

packages/story-editor/src/components/publishModal/types.js
```diff
@@ -24,6 +24,6 @@ import { INPUT_KEYS } from './constants';
 
 export const MANDATORY_INPUT_VALUE_TYPES = PropTypes.shape({
   [INPUT_KEYS.EXCERPT]: PropTypes.string,
-  [INPUT_KEYS.SLUG]: PropTypes.string,
+  [INPUT_KEYS.SLUG]: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
miina, line 27: Wouldn't the type still be a string, e.g. `"5"` and not `5`? Thoughts?

miina (approved):
Looks great!

(Had 2 questions that might be unrelated to adding the experiment PR though)

packages/story-editor/src/components/publishModal/karma/publishModal.karma.js
BrittanyIRL: good call out, yup! Design wants title and description here to be mandatory.

packages/story-editor/src/components/publishModal/types.js
BrittanyIRL: I thought so but then i was running through a handful of situations in the editor and it was a number so i updated this here that way we wouldn't get noise in the console.

maxyinger (approved):

packages/story-editor/src/components/publishModal/stories/index.js
```diff
@@ -25,10 +25,16 @@ import PublishModal from '../publishModal';
 
 export default {
   title: 'Stories Editor/Components/Dialog/Publish Modal',
-  component: PublishModal,
+  args: {
+    isOpen: true,
+  },
+  argTypes: {
+    onPublish: { action: 'onPublish clicked' },
+    onClose: { action: 'onClose clicked' },
```
BrookeGraham, line 33: ♥️

packages/story-editor/src/components/publishModal/karma/publishModal.karma.js
```diff
@@ -0,0 +1,108 @@
...
+    await fixture.events.click(publish);
+  }
+
+  beforeEach(async () => {
+    fixture = new Fixture();
+    fixture.setFlags({ enableUpdatedPublishStoryModal: true });
+    await fixture.render();
+
+    await openPublishModal();
+    publishModal = await fixture.screen.findByRole('dialog', {
```
BrookeGraham, line 42: Nice.

BrookeGraham (approved):
Looks great.

packages/story-editor/src/components/publishModal/karma/publishModal.karma.js
```diff
@@ -0,0 +1,108 @@
...
+    fixture.restore();
+  });
+
+  function getPublishModalElement(role, name) {
+    const { getByRole } = within(publishModal);
+
+    return getByRole(role, {
+      name,
+    });
+  }
```
BrookeGraham, line 57: Only thing I would suggest is that if we use `findByRole` here, then we can `await` and get repolling. Since this is getting used for many elements, it may make it a little more robust if something ends up taking a bit longer to appear.

BrittanyIRL: thank you! 

out of pure laziness, i'm going to make this update in the next pr that was branched off of this to avoid waiting for tests to run again.

Issue #6304, opened 2021-02-10
Infrastructure: update to webpack v5


swissspidy:
## Task Description

webpack 5 contains a ton of improvements and bug fixes, and will also help in our endeavour to further optimize our bundles. There are some caveats though. See #5792 for more info.

swissspidy:
@BrittanyIRL This one is at the interesction of Karma and performance improvements, perhaps another thing that Pea pod wants to look into?

#5792 has the WIP for this.

BrittanyIRL:
> @BrittanyIRL This one is at the interesction of Karma and performance improvements, perhaps another thing that Pea pod wants to look into?
> 
> #5792 has the WIP for this.

Sure! This will be a big one but it would be so great to complete.

swissspidy:
The good thing is the majority of the work should done, what's left is fixing the storybook and — surprise — Karma tests. Admittedly, it still might take some time to crack these two.

swissspidy:
cc @merapi as you might have some ideas on how to fix the Karma tests in #5792 - the PR has all the details

swissspidy:
Regarding the open todos on #5792: One of the storybook issues might be fixed by ` v6.5.0-alpha.23`, see https://github.com/storybookjs/storybook/issues/14877#issuecomment-1020290378

felipebochehin87:
Tested against https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/, using PR #5792.

Regressed/smoke tested editor:

1. Created new story
2. Updated existing one
3. Added images/videos/shapes/text
4. Updated colors, layouts
5. Preview
6. Tested checklist
7. Published
8. Added and remove links, background images

Working as expected.

PR #10487, opened 2022-02-07
Don't use node globals in browser context

swissspidy:
## Context

Splitting this out from #5792, see #6304 for full context

## Summary

Replace usage of `global.*` with `window.*`

webpack v5 no longer polyfills Node.js APIs in the browser, including `global`, which is a Node thing, but webpack has happily supported in the past by making it an alias of `window`. Example: `global.navigator` becomes `window.navigator`.

Replace usage of `process.env` with custom variables

webpack v5 no longer polyfills Node.js APIs in the browser, including `process`. Uses variables like `WEB_STORIES_ENV` instead of `process.env.NODE_ENV` in all frontend code.

## Relevant Technical Choices

See above

## User-facing changes

None

## Testing Instructions

1.

## Checklist

---

Partially addresses #6304


packages/wp-story-editor/src/components/mediaUpload/mediaPicker/WordPressImageCropper.js
```diff
@@ -53,7 +53,7 @@ const WordPressImageCropper = window.wp?.media?.controller?.Cropper?.extend?.({
         : control.params.height;
     }
 
-    return wp.ajax.post('crop-image', {
+    return window.wp.ajax.post('crop-image', {
```
spacedmonkey, line 56: 🤔 Is this required or nice to have?

spacedmonkey, line 56: 🤔 Is this required or nice to have?

swissspidy: It is required. I removed `wp` as a recognized global variable in the ESLint config, so that we are forced to explicitly write the full variable with `window.wp`.

swissspidy: Yes. Closing in favor of the [other open thread](https://github.com/GoogleForCreators/web-stories-wp/pull/10487#discussion_r800666513)

spacedmonkey (approved):

packages/e2e-tests/src/config/bootstrap.js
```diff
@@ -116,7 +116,7 @@ const pageEvents = [];
 jest.setTimeout(PUPPETEER_TIMEOUT || 100000);
 
 // Retry flaky tests at most 2 times in CI (off by 1).
-if ('true' === process.env.CI) {
+if ('true' === process.env.NODE_ENV) {
```
samwhale: 💭  should this be `WEB_STORIES_ENV`?

Maybe not since this is for e2e tests

packages/e2e-tests/src/jest.config.js
```diff
@@ -37,7 +37,7 @@ export default {
   testMatch: ['**/specs/**/*.[jt]s', '**/?(*.)spec.[jt]s'],
   // @jest/test-sequencer is the default.
   testSequencer:
-    'true' === process.env.CI
+    'true' === process.env.NODE_ENV
```
samwhale: Same here? I guess this one doesn't matter since process should be available when running jest.

samwhale (approved):

packages/e2e-tests/src/config/bootstrap.js
swissspidy: This code is executed in Node context, not browser context. So using `process.env` is correct.

But I shouldn't have changed this to `NODE_ENV`, it should stay `process.env.CI` 🤦

packages/e2e-tests/src/jest.config.js
swissspidy: Yes, same here. This code is executed in Node context, not browser context. So using `process.env` is correct. 

But it really should stay `process.env.CI`

PR #10490, opened 2022-02-07
Text resizing: Revert changes to Text element from #10106

### LINKED ISSUE #10466: Rework how resizing Text element works
## Feature Description
In https://github.com/GoogleForCreators/web-stories-wp/issues/10106 the design panel ratio lock was related to resizing the elements, however, since it's not clear right away that these are related, there was some feedback about issues with resizing text.

Since the Text element works differently than the other elements (e.g. height is auto-set with unlocked ratio), we should unrelate the design panel lock and resizing on the Page for Text element, and revert the changes to what it was before (for Text only!)

The rest of the elements will still be dependent on the lock in the Design panel and "locked" by default except for Image and Rectangle.

miina:
## Summary
This influences the Text element only:
- Always locks the aspect ratio if resizing a text element from corners
- Always unlocks the ratio when resizing a text element from edges.
- Resizing on the page and the aspect ratio lock in the design panel are not related 

## Testing Instructions

1. Add a text element
2. Verify that resizing from corners is always with locked aspect ratio
3. Verify that resizing from edges is always with unlocked aspect ratio
4. Verify that toggling the aspect ratio lock in the design panel does not influence resizing on the page.

No

No

No

## Checklist

---

Fixes #10466


packages/story-editor/src/components/canvas/singleSelectionMoveable/useResize.js
```diff
@@ -131,11 +129,17 @@ function useSingleSelectionResize({
...
 
-  const onResizeStart = ({ setOrigin, dragStart }) => {
+  const onResizeStart = ({ setOrigin, dragStart, direction }) => {
     setOrigin(['%', '%']);
     if (dragStart) {
       dragStart.set(frame.translate);
     }
+    // Both `direction[]` values for diagonals are either 1 or -1. Non-diagonal
+    // directions have 0s.
+    const newResizingMode = direction[0] !== 0 && direction[1] !== 0;
```
barklund, line 139:
```suggestion
    const newResizingMode = direction.every(n => n !== 0);
```

barklund (approved):
Looks good. Whether it makes sense or not is beyond me. I normally just try stuff and see what happens, so don't really have a lot of expectations as a user here.

packages/story-editor/src/components/canvas/singleSelectionMoveable/useResize.js
barklund: Or now that I think about it again:

```suggestion
    const newResizingMode = !direction.includes(0);
```

miina: > Or now that I think about it again:

I'm not 100% sure about this, e.g. I don't know if `direction` includes some other indexes as well in addition to `0` and `1`. Just checking the specific indexes that we're interested in seems more safe, especially since there are only 2 of those :) Thoughts?

miina:
Merging with failing e2e since that test seems to fail elsewhere as well and is not related to this PR.

PR #10510, opened 2022-02-07
Block: Adjust margin for "Remove" button

### LINKED ISSUE #10164: Block: Remove button styling off
## Bug Description

The styling of the remove thumbnail button seems to be off. Being too close to the select button see screenshot. 

## Screenshots

Screenshot 2022-01-12 at 10 54 08

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

timarney:
## Context

Updates margin for poster image "remove button" 

## Relevant Technical Choices

Adds a 12px margin (left and right) 
This matches the spacing for other WP buttons i.e. preview & publish)

## User-facing changes

| **Before**   |      **After**    
|----------|:-------------:|
| before |  Screen Shot 2022-02-07 at 4 28 08 PM | 
|before-hover  |   hover   | 

**Right to left support**

Screen Shot 2022-02-07 at 5 09 15 PM

## Testing Instructions

1. Insert a Web Stories block, select displaying a single story
2. Insert a story URL, e.g. `https://preview.amp.dev/documentation/examples/introduction/stories_in_amp`
3. View button spacing for poster image

No

No

No

## Checklist

---

Fixes #10164

### DIFF
--- a/packages/stories-block/src/block/block-types/single-story/embedControls.js
+++ b/packages/stories-block/src/block/block-types/single-story/embedControls.js
@@ -151,7 +151,7 @@ const EmbedControls = (props) => {
                       )}
                 </p>
                 {hasPoster && (
-                  <Button onClick={onRemovePoster} isTertiary>
+                  <Button className="remove" onClick={onRemovePoster}>
                     {__('Remove', 'web-stories')}
                   </Button>
                 )}
--- a/packages/stories-block/src/block/block-types/single-story/embedControls.js
+++ b/packages/stories-block/src/block/block-types/single-story/embedControls.js
@@ -151,7 +151,10 @@ const EmbedControls = (props) => {
                       )}
                 </p>
                 {hasPoster && (
-                  <Button className="remove" onClick={onRemovePoster}>
+                  <Button
+                    className="web-stories-embed-poster-remove"
+                    onClick={onRemovePoster}
+                  >
                     {__('Remove', 'web-stories')}
                   </Button>
                 )}
--- a/packages/stories-block/src/css/embed.css
+++ b/packages/stories-block/src/css/embed.css
@@ -65,6 +65,6 @@
   height: 100%;
 }
 
-.block-editor-block-inspector .remove {
+.block-editor-block-inspector .web-stories-embed-poster-remove {
   margin-left: 12px;
 }
--- a/packages/stories-block/src/css/embed.css
+++ b/packages/stories-block/src/css/embed.css
@@ -64,3 +64,7 @@
   width: 100%;
   height: 100%;
 }
+
+.block-editor-block-inspector .remove {
+  margin-left: 12px;
+}


packages/stories-block/src/block/block-types/single-story/embedControls.js
```diff
@@ -151,7 +151,11 @@ const EmbedControls = (props) => {
                       )}
                 </p>
                 {hasPoster && (
-                  <Button onClick={onRemovePoster} isTertiary>
+                  <Button
+                    style={{ margin: '0 12px' }}
```
spacedmonkey, line 155: This button, should should have the same padding not margin as the button next to it. Is there way to make this `isPrimary` but have a white background?

packages/stories-block/src/block/block-types/single-story/embedControls.js
```diff
@@ -151,7 +151,11 @@ const EmbedControls = (props) => {
...
+                    onClick={onRemovePoster}
+                    isTertiary
+                  >
```
spacedmonkey:
```suggestion
                  <Button onClick={onRemovePoster}>
```

Just removing the `isTertiary` seems to do the trick.

spacedmonkey: See https://wordpress.github.io/gutenberg/?path=/docs/components-button--default

timarney: Pushed an update -- removed isTertiary + the hard coded margin on the button.

timarney: With the margin removed --- the buttons outline when focused still "bump" into each other when using the keyboard to navigate.

With "VoiceOver" on + keyboard nav
Screen Shot 2022-02-08 at 5 59 03 AM

Keyboard nav "VoiceOver" off
Screen Shot 2022-02-08 at 6 30 18 AM

swissspidy: In that case, maybe some left margin + removing isTertiary?

timarney: @spacedmonkey understand your concern about hard coding a margin but I'm not finding other examples in the code base for spacing.

The button group https://wordpress.github.io/gutenberg/?path=/docs/components-buttongroup--default 

Still hard codes a value --- but it applies to both buttons.

```
() => {
	const style = { margin: '0 4px' };
	return (
		<ButtonGroup>
			<Button variant="primary" style={ style }>
				Button 1
			</Button>
			<Button variant="primary" style={ style }>
				Button 2
			</Button>
		</ButtonGroup>
	);
}
```

swissspidy: We have a CSS file in our block for adding custom styling if necessary.

timarney: Re-added the margin here:

https://github.com/GoogleForCreators/web-stories-wp/pull/10510/commits/062fb080cbcfdd2ad9bcbe770e8175486aa7b969

I'm not a fan of hard-coding this either but moving it out to a stylesheet will likely make it harder to track down if / when it needs to be updated.

Open to other suggestions as you see fit.

swissspidy: > moving it out to a stylesheet will likely make it harder to track down if / when it needs to be updated.

Fair concern, but now it's an outlier because all our other styles are in a stylesheet.

Also, our stylesheet files are run through RTLCSS.

So you only need to write `margin-left: 12px` and it will be automatically changed to `margin-right: 12px` for RTL.

That can't be done for inline styles.


So I'd just add a className to the Remove button and style that in the CSS file.

packages/stories-block/src/block/block-types/single-story/embedControls.js
```diff
@@ -151,7 +151,7 @@ const EmbedControls = (props) => {
                       )}
                 </p>
                 {hasPoster && (
-                  <Button onClick={onRemovePoster} isTertiary>
+                  <Button className="remove" onClick={onRemovePoster}>
```
swissspidy: Nit: let's use `web-stories-embed-poster-remove` as the class name, in line with the `web-stories-embed-poster-label` class name above.

timarney: Sounds good -- pushed a commit.

spacedmonkey (approved):

swissspidy (approved):
LGTM!

Really appreciate the detailed PR description including RTL screenshots too.

PR #10511, opened 2022-02-07
Update PHP-Scoper to v0.16.2 

### LINKED ISSUE #10407: Update PHP-Scoper to v0.16.2
## Task Description

The current version of PHP-Scoper is 0.16.2.

In [version 0.16.0](https://github.com/humbug/php-scoper/releases/tag/0.16.0), there were lots of improvements to the config, making it a lot simpler for us to configure PHP-Scoper in a WP plugin context.

It means we can get rid of all the manual `str_replace` calls in the scoper config thanks to `expose-global-{constants, classes, functions}` and `expose-{namespaces, constants, classes, functions}`

We could even automate this by setting the config based on the WP stubs, see https://github.com/humbug/php-scoper/issues/303#issuecomment-1021388471

Though we could split that up in its own ticket.

### Implementation Brief

* Install latest version of PHP-Scoper via Composer
* Remove all the `str_replace` calls in the PHP-Scoper config
* Install [`sniccwp/php-scoper-wordpress-excludes`](https://packagist.org/packages/sniccwp/php-scoper-wordpress-excludes) as a dev dependency and configure the `exclude-classes`, `exclude-functions`, and `exclude-constants` options accordingly

timarney:
## Context

Updates PHP-Scoper to 0.16.2 which allows us to clean up the config file

## Summary
Updates the PHP-Scoper version + config

- Removes "str_replace" calls in the config

## Relevant Technical Choices

- Pulls in `sniccwp/php-scoper-wordpress-excludes` which provides .json files for WP functions, classes to exclude

## User-facing changes

N/A

## Testing Instructions

1.

No outside of introduces a new package `sniccwp/php-scoper-wordpress-excludes`

No

No new package is MIT License

## Checklist

---

Fixes #10407

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,6 +30,7 @@
     "roave/security-advisories": "dev-latest",
     "sirbrillig/phpcs-variable-analysis": "^2.8",
     "slevomat/coding-standard": "^7.0.18",
+    "sniccowp/php-scoper-wordpress-excludes": "^5.9",
     "szepeviktor/phpstan-wordpress": "^1.0",
     "wp-coding-standards/wpcs": "^2.3",
     "yoast/wp-test-utils": "^1.0.0"
@@ -87,7 +88,7 @@
       "php-scoper": {
         "path": "vendor/bin/php-scoper",
         "type": "phar",
-        "url": "https://github.com/humbug/php-scoper/releases/download/0.15.0/php-scoper.phar"
+        "url": "https://github.com/humbug/php-scoper/releases/download/0.16.2/php-scoper.phar"
       }
     },
     "enable-patching": true,
--- a/scoper.inc.php
+++ b/scoper.inc.php
@@ -186,9 +186,10 @@ return [
 
 	'exclude-functions' => $wp_functions,
 	
-	// Note the hardcoded WP constants can be removed if a fix gets merged for
-	// currently exclude-wordpress-constants.json is empty []
-	// https://github.com/sniccowp/php-scoper-wordpress-excludes/issues/2
+	// Currently exclude-wordpress-constants.json is empty [].
+	// Issue https://github.com/sniccowp/php-scoper-wordpress-excludes/issues/2 .
+	// If fixed we can remove the hardcoded WP constants.
+	
 	'exclude-constants' => array_merge(
 		$wp_constants,
 		[
--- a/scoper.inc.php
+++ b/scoper.inc.php
@@ -185,7 +185,10 @@ return [
 	'exclude-classes'   => $wp_classes,
 
 	'exclude-functions' => $wp_functions,
-
+	
+	// Note the hardcoded WP constants can be removed if a fix gets merged for
+	// currently exclude-wordpress-constants.json is empty []
+	// https://github.com/sniccowp/php-scoper-wordpress-excludes/issues/2
 	'exclude-constants' => array_merge(
 		$wp_constants,
 		[
--- a/scoper.inc.php
+++ b/scoper.inc.php
@@ -10,11 +10,15 @@
 
 use Isolated\Symfony\Component\Finder\Finder;
 
+$wp_classes   = json_decode( file_get_contents( 'vendor/sniccowp/php-scoper-wordpress-excludes/generated/exclude-wordpress-classes.json' ) );
+$wp_functions = json_decode( file_get_contents( 'vendor/sniccowp/php-scoper-wordpress-excludes/generated/exclude-wordpress-functions.json' ) );
+$wp_constants = json_decode( file_get_contents( 'vendor/sniccowp/php-scoper-wordpress-excludes/generated/exclude-wordpress-constants.json' ) );
+
 return [
-	'prefix'                     => 'Google\\Web_Stories_Dependencies',
+	'prefix'            => 'Google\\Web_Stories_Dependencies',
 
 	// See: https://github.com/humbug/php-scoper#finders-and-paths.
-	'finders'                    => [
+	'finders'           => [
 		// Main AMP PHP Library.
 		Finder::create()
 			->files()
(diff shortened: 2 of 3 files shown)


composer.json
```diff
@@ -24,12 +24,14 @@
     "automattic/vipwpcs": "^2.3",
     "dealerdirect/phpcodesniffer-composer-installer": "^0.7",
     "dg/bypass-finals": "1.1.3",
+    "php-stubs/wordpress-stubs": "^5.9",
```
swissspidy: This is a transitive dependency of both `szepeviktor/phpstan-wordpress` and `sniccwp/php-scoper-wordpress-excludes`, so we shouldn't need to add it here explicitly

scoper.inc.php
```diff
@@ -10,11 +10,14 @@
 
 use Isolated\Symfony\Component\Finder\Finder;
 
+$wp_classes   = json_decode( file_get_contents( 'vendor/sniccwp/php-scoper-wordpress-excludes/generated/exclude-wordpress-classes.json' ) );
+$wp_functions = json_decode( file_get_contents( 'vendor/sniccwp/php-scoper-wordpress-excludes/generated/exclude-wordpress-functions.json' ) );
```
swissspidy: I think we can add the constants here too, then this should be good to go!

timarney: If we're looking to pull from that package -- I'm seeing an empty array at the moment.

Screen Shot 2022-02-08 at 6 52 38 AM

swissspidy: Hmm that's unexpected. Sounds like a bug in that package.

@calvinalkan why is the constants file empty? I'd expect constants like `WP_CONTENT_DIR` and `WP_CONTENT_URL` to be in there.

timarney: Once we have the answer for that and assuming that's a bug I'll set it up to array merge the rest of the constants

scoper.inc.php
```diff
@@ -10,11 +10,27 @@
...
+ * Gets a list of WP globals to ignore.
+ *
+ * @param string $json_file the file name to pull the excludes list from.
+ * @return array[] Array of WP Classes, WP Functions, or WP Constants to exclude.
+ */
+function get_excludes( string $json_file ): array {
+	try {
+		$excludes_vendor_path = 'vendor/sniccwp/php-scoper-wordpress-excludes/generated/';
+		return json_decode( file_get_contents( $excludes_vendor_path . $json_file ) );
+	} catch ( \Exception $exception ) {
```
swissspidy: Was there a lint warning or something about this?

The simple `json_decode( file_get_contents( 'vendor/sniccwp/php-scoper-wordpress-excludes/generated/exclude-wordpress-classes.json' ) );` statements you had originally seemed fine to me

timarney: I can take it back out if you like ... was looking a little repetitive + getting long with the array merge (which is juts pulling in an empty array at the moment).

scoper.inc.php
swissspidy: Filed an issue at https://github.com/sniccowp/php-scoper-wordpress-excludes/issues/2

Once it's fixed we can do the array merge as you say in a follow-up PR.

scoper.inc.php
swissspidy: We can discuss it today but it didn't feel like a big issue at all.

calvinalkan:
Hey, I just noticed today that we misspelled the package name. Since I released it on packagist about 4 days or so I still decided to fix it.

So the correct name is: `sniccowp/php-scoper-wordpress-excludes`. I'm sorry for the inconvenience.

scoper.inc.php
timarney: reverted to the get contents statements + updated to pull from the `sniccowp/php-scoper-wordpress-excludes` package name

swissspidy (approved):

scoper.inc.php
```diff
@@ -10,11 +10,15 @@
 
 use Isolated\Symfony\Component\Finder\Finder;
 
+$wp_classes   = json_decode( file_get_contents( 'vendor/sniccowp/php-scoper-wordpress-excludes/generated/exclude-wordpress-classes.json' ) );
+$wp_functions = json_decode( file_get_contents( 'vendor/sniccowp/php-scoper-wordpress-excludes/generated/exclude-wordpress-functions.json' ) );
+$wp_constants = json_decode( file_get_contents( 'vendor/sniccowp/php-scoper-wordpress-excludes/generated/exclude-wordpress-constants.json' ) );
```
spacedmonkey, line 15: Should we check if these files exists before trying to use them? Maybe create a little helper function. 

```
function read_file( $path ){
      if( !is_readable( $path ) ){
           return [];
      }
      return json_decode( file_get_contents( $path ) );
}
```

scoper.inc.php
```diff
@@ -173,99 +177,34 @@
...
 
-	// See https://github.com/humbug/php-scoper#constants--classes--functions-from-the-global-namespace.
-	'whitelist-global-classes'   => false,
+	'exclude-functions' => $wp_functions,
 
-	// See https://github.com/humbug/php-scoper#constants--classes--functions-from-the-global-namespace.
-	'whitelist-global-functions' => false,
+	'exclude-constants' => array_merge(
+		$wp_constants,
+		[
```
spacedmonkey, line 195: Why are these needed? Don't they come from `$wp_constants`. If they are a temporary, what is the plan to get rid of them? If temporary, a comment / todo to fix these hardcoded values at a later date.

swissspidy: Please see discussion at https://github.com/GoogleForCreators/web-stories-wp/pull/10511#discussion_r801352022

Basically, this package does not yet provide a list of constants, but most likely will in a future version

scoper.inc.php
timarney: There was a small discussion related to a helper function here --- but not related to the existence of the files 
https://github.com/GoogleForCreators/web-stories-wp/pull/10511#discussion_r802115710

swissspidy: We're including a specific version of a package that is known to include these files.

If a file does not exist we absolutely need to be informed. If this fails we need to know.

scoper.inc.php
spacedmonkey: That comment thread was hidden so didnt see this. My comment about adding a todo / comment is still valid.

timarney: I'll add a comment ... there's an issue filed here -- it's currently an empty file.

https://github.com/sniccowp/php-scoper-wordpress-excludes/issues/2

scoper.inc.php
timarney: > If a file does not exist we absolutely need to be informed. If this fails we need to know.

1) To confirm we do want to add a helper?
2) If yes so we want to echo to the console or throw -- (how do we want to be informed?)

spacedmonkey:
@swissspidy Little confused why this was merged, there were still outstanding questions. https://github.com/GoogleForCreators/web-stories-wp/pull/10511#discussion_r803960676

scoper.inc.php
swissspidy: No helper needed.

I don't see a need for action here.

swissspidy:
I simply missed that comment. Resolved now.

I don't see a need for further action here.

PR #10513, opened 2022-02-08
Update engine block for npm packages.

c0b41:
currently you can't install `@googleforcreators` packages because all of them blocked node 16

### DIFF
--- a/packages/activation-notice/package.json
+++ b/packages/activation-notice/package.json
@@ -21,7 +21,7 @@
     "url": "https://github.com/googleforcreators/web-stories-wp/issues"
   },
   "engines": {
-    "node": ">= 12 || >= 14 || >= 16",
+    "node": ">= 16",
     "npm": ">= 7.3"
   },
   "main": "./src/index.js",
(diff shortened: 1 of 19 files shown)


swissspidy:
Thanks for your contribution!

I take it that you're currently using Node 12 or 14?

We currently use Node 16, which is [the current LTS version](https://nodejs.org/en/about/releases/). We have not tested with older versions, so I can't guarantee that it'll work.

c0b41:
okay but why we need to check node version for client side library?

swissspidy:
Legit question. We'll change this.

But AFAIK this should not _block_ you from installing. npm will just _warn_ you.

c0b41:
npm installs but yarn not, only if you use `yarn install --ignore-engines`, either way thnx for pushing fixes.

swissspidy:
That's good to know, thanks.

Curious, what are you looking to build with these packages?
Any other feedback you have for us?

c0b41:
Currently i am using CRA 4 working great but I have some issues and I am gone open issue for that. ex 10536
And CRA 5 is not working, either esm modules broken or CRA 5 have some issues.

swissspidy:
Yeah it's totally possible that the esm version isn't working properly yet. This is still in early stages.

PR #10514, opened 2022-02-08
Background Audio: Display captions in editor on play

### LINKED ISSUE #10439: Background Audio: Display captions in editor on play
## Feature Description

When playing background audio in the editor and the audio has captions, it would be nice to display them in the editor for a better WYSIWYG experience.

The `VideoCaptionsLayer` component needs to be updated to check for captions existence and to find the audio player on the page.

The audio player needs to be updated to include the `track` (with `kind="metadata"`) so it's not otherwise visible)

spacedmonkey:
## Summary

Reuse the `VideoCaptionsLayer` component to render captions for background audio.  

## Relevant Technical Choices

Change Video to media, as captions are used for audio and video captions. 

## User-facing changes

https://user-images.githubusercontent.com/237508/152998936-292f40b5-4ef7-456c-9d14-6aca39b8a2c7.mov

## Testing Instructions

1. Create a story editor. 
2. Create a new page. 
3. Select background.
4. Add background audio. 
5. Add background audio caption. 
6. Play background audio. 
7. See captions. 

Please ensure that enhanced captions are enabled. 

## Checklist

---

Fixes #10439


packages/story-editor/src/components/canvas/mediaCaptions/trackRenderer.js
```diff
@@ -42,11 +41,11 @@ const Track = styled.div`
  * Track renderer component.
  *
  * @param {Object} props Component props.
- * @param {Object} props.videoElement Video element
+ * @param {string} props.elementId Media element id.
  * @param {number} props.trackIndex Video track index in the tracklist.
  * @return {*} Track component.
  */
-function TrackRenderer({ videoElement, trackIndex }) {
+function TrackRenderer({ trackIndex, elementId }) {
```
spacedmonkey:
```suggestion
function TrackRenderer({ elementId, trackIndex }) {
```

packages/story-editor/src/components/audioPlayer/index.js
```diff
@@ -96,8 +97,26 @@ function AudioPlayer({ title, src, mimeType }) {
   return (
     <Wrapper>
       {/* eslint-disable-next-line jsx-a11y/media-has-caption, styled-components-a11y/media-has-caption */}
-      <Audio crossOrigin="anonymous" loop ref={playerRef}>
+      <Audio
+        crossOrigin="anonymous"
+        loop
+        ref={playerRef}
+        id={audioId || undefined}
```
swissspidy: This is a bit pointless. `audioId` above has an empty string as default value but here you set `undefined` if it's falsey.

Instead you could just remove the default value and do `id={audioId}` here.

packages/story-editor/src/components/panels/shared/backgroundAudioPanelContent.js
```diff
@@ -197,6 +198,8 @@ function BackgroundAudioPanelContent({
               )}
               src={resource?.src}
               mimeType={resource?.mimeType}
+              tracks={tracks || []}
```
swissspidy: `tracks` already defaults to an empty array further above on line 72:

```js
const { resource, tracks = [], loop = true } = backgroundAudio || {};
```

So this shouldn't be needed here

swissspidy:
This seems to work well 👍

packages/story-editor/src/components/panels/document/backgroundAudio/backgroundAudio.js
```diff
@@ -81,6 +81,7 @@ function BackgroundAudioPanel() {
       <BackgroundAudioPanelContent
         backgroundAudio={backgroundAudio}
         updateBackgroundAudio={updateBackgroundAudio}
+        audioId="story-background-audio"
```
swissspidy: We don't need to pass an ID here because a) it's an optional prop and b) story background audio does not have captions anyway.

swissspidy (approved):
One nit, otherwise looks good

packages/story-editor/src/components/panels/document/backgroundAudio/backgroundAudio.js
spacedmonkey: I know this one isn't needed, it is just a little tidier this way. I am more than happy to remove it.

swissspidy: I'd remove it. Otherwise it gives a false impression that it's relevant for something

spacedmonkey:
```suggestion
```

packages/story-editor/src/components/audioPlayer/index.js
```diff
@@ -96,8 +97,21 @@ function AudioPlayer({ title, src, mimeType }) {
   return (
     <Wrapper>
       {/* eslint-disable-next-line jsx-a11y/media-has-caption, styled-components-a11y/media-has-caption */}
-      <Audio crossOrigin="anonymous" loop ref={playerRef}>
+      <Audio crossOrigin="anonymous" loop ref={playerRef} id={audioId}>
         <source src={src} type={mimeType} />
+        {tracks &&
+          tracks.map(({ srclang, label, track: trackSrc, id: key }, i) => (
```
timarney, line 103: Minor - I think `srclang` should be `srcLang` to keep it consistent with the other vars i.e. trackSrc , playerRef etc...

packages/story-editor/src/components/canvas/mediaCaptions/mediaCaptionsLayer.js
```diff
@@ -47,55 +47,77 @@ const CaptionsCanvas = styled.div`
...
 
   useEffect(() => {
-    setVideoTrackCount(0);
+    setMediaTrackCount(0);
+    setMediaElementId('');
 
-    if (isEditing || !videoElement) {
+    if (backgroundAudio) {
+      setMediaElementId(`page-${currentPageId}-background-audio`);
+      setMediaTrackCount(backgroundAudio.tracks.length);
```
timarney, line 95: Worth adding null checks here?
```
backgroundAudio?.tracks?.length
```

timarney:
Works well --- one note for testing steps is option needs to be turned on via experiments

packages/story-editor/src/components/audioPlayer/index.js
spacedmonkey: I am going to fix this in an follow up PR.

swissspidy (approved):

packages/story-editor/src/components/audioPlayer/index.js
spacedmonkey: @timarney Created a follow up PR - https://github.com/GoogleForCreators/web-stories-wp/pull/10548

PR #10517, opened 2022-02-08
Editor: Added dismiss option to floating menu

barklund:
## Context

This is stage four of #10112. This adds a (functional) dismiss button to the floating menu, that resets as soon as selection changes (including selecting a different element of the same type or changes the number of selected elements in a multi-selection), but does not reset just because the actively selected element changes (moves, or otherwise updates). The dismissed menu even stays dismissed as you move in and out of edit mode on a text element.

It also passes a sorted list of selected element types to the menu, which will be needed to render the correct menu elements.

And finally, I also memoized the menu component. Most of the time, this doesn't matter as the menu will receive new props when the layer re-renders, but there are exceptions. E.g. if you change selection from one text field to another, the menu wrapper actually doesn't have to re-render because no new props will be passed to it (though some of the elements inside it will). So I think it does do something (which is why I also memoized the dismiss callback).

## Testing Instructions

## Checklist

---

Partially addresses #10112


packages/story-editor/src/components/floatingMenu/menu.js
```diff
@@ -29,82 +29,85 @@ import {
...
+          </ContextMenuComponents.MenuButton>
+        </ContextMenu>
+      </MenuWrapper>
+    );
+  })
+);
 
 FloatingMenu.propTypes = {
-  isVisible: PropTypes.bool,
+  dismiss: PropTypes.func.isRequired,
```
miina, line 110: Nit: maybe rename to `onDismiss` to make it sound more like a function?

packages/story-editor/src/components/floatingMenu/menu.js
```diff
@@ -29,82 +29,85 @@ import {
...
+              title="Pipette!"
+              placement={TOOLTIP_PLACEMENT.BOTTOM}
+            >
+              <Icons.Pipette />
+            </ContextMenuComponents.MenuIcon>
+          </ContextMenuComponents.MenuButton>
+          <ContextMenuComponents.MenuSeparator />
+          <ContextMenuComponents.MenuButton onClick={dismiss}>
+            <ContextMenuComponents.MenuIcon
+              title="Dismiss menu"
```
miina, line 97: `i18n` if that's final?

packages/story-editor/src/components/floatingMenu/layer.js
```diff
@@ -99,9 +128,20 @@ function FloatingMenuLayer() {
...
+  if (!hasMenu) {
+    return false;
+  }
+
+  return (
+    <FloatingMenu
+      ref={menuRef}
+      dismiss={dismiss}
+      hasMultiSelection={hasMultiSelection}
+      selectionTypes={selectionTypes}
```
miina, line 128: `hasMultiSelection` and `selectionTypes` don't seem to be used now. Are these for later?

packages/story-editor/src/components/floatingMenu/layer.js
```diff
@@ -50,11 +55,35 @@ function FloatingMenuLayer() {
...
+  const {
+    hasSelection,
+    hasMultiSelection,
+    selectionTypes,
+    selectionIdentifier,
+  } = useStory(({ state: { selectedElements } }) => ({
+    hasSelection:
+      selectedElements.filter(({ isBackground }) => !isBackground).length > 0,
+    hasMultiSelection: selectedElements.length > 1,
+    selectionTypes: [
```
miina, line 56: Maybe `selectedTypes`? In case this is for storing the types of elements that are selected. Otherwise it sounds like "types of selection" which isn't clear to understand. Maybe even `selectedElementTypes` for making it easier to understand for people in the future 😛

packages/story-editor/src/components/floatingMenu/layer.js
```diff
@@ -50,11 +55,35 @@ function FloatingMenuLayer() {
...
+      ...new Set(selectedElements.map(({ type }) => type)),
+    ].sort(),
+    selectionIdentifier: selectedElements
+      .map(({ id }) => id)
+      .sort()
+      .join('#'),
```
miina: Curious why is `#` needed, for debugging?

packages/story-editor/src/components/floatingMenu/layer.js
```diff
@@ -71,18 +100,18 @@ function FloatingMenuLayer() {
     // frame will already be updating because of the resize, so a DOM mutation is incoming.
   }, [workspaceWidth, workspaceHeight]);
 
+  const hasMenu = hasSelection && !isDismissed;
```
miina: One additional case to consider -- the menu should probably be hidden for edit mode of media, otherwise scaling slider will be on top of the edit menu. Or perhaps you were planning to do this in a media-specific PR?

miina (approved):
Works great!

Main comment/question I had was about media edit mode, the rest are just nits.

packages/story-editor/src/components/floatingMenu/layer.js
merapi: +1
`selectedElements.map(select type) = selectedElementsTypes` makes sense.

packages/story-editor/src/components/floatingMenu/layer.js
merapi: Do we even need to sort this? (You can't just swap elements in place 🤔 )

image

packages/story-editor/src/components/floatingMenu/menu.js
merapi: `onDismiss` would suggest a function that will do something in response to (dismiss) action, but here we are actually doing the action. I'm ok with this, it's like `focusSearch` or `toggle` used elsewere.

merapi (approved):
LG, just check this bug, I had it 2 times during a short testing session.

https://user-images.githubusercontent.com/21036224/153228567-0c46fd66-f69c-42ad-9882-91bd89b03ccc.mov

packages/story-editor/src/components/floatingMenu/menu.js
miina: True, but somehow it also sounds like a boolean to me currently 🤷‍♀️ Anyway, ok by me if you both think it's ok.

packages/story-editor/src/components/floatingMenu/layer.js
barklund: @miina, this is all assuming that id's can be anything, so they could potentially collide if e.g. you come from a selection of '123' and '456' joined without separator to a selection of just '123456'. But if we assume that id's are always same length, then there isn't an issue here.

@merapi, good point. I wasn't sure whether lasso selection always selected elements in the same order, so you might make two lasso selections, that selected the same elements, but in different order. If that's not an issue, than we don't need to sort. Also, does sort of 5 strings really take more than a second? WTF?

packages/story-editor/src/components/floatingMenu/layer.js
barklund: I'll change it to `selectedElementTypes` in the next stage.

packages/story-editor/src/components/floatingMenu/layer.js
barklund: Yes, that was because they were needed for the next stage.

packages/story-editor/src/components/floatingMenu/menu.js
barklund: Changed in the next stage, I believe.

packages/story-editor/src/components/floatingMenu/menu.js
barklund: What else would it be with that name - we have naming rules for booleans to avoid the confusion. But sure, we can do that. I'll fix that in the next stage PR.

packages/story-editor/src/components/floatingMenu/layer.js
barklund: I don't see a menu in image edit mode. Do you? Edit-mode uses a different selection frame, which I haven't added the menu logic to, so it shouldn't be there at all.
Screen Shot 2022-02-09 at 14 44 35

barklund:
> LG, just check this bug, I had it 2 times during a short testing session.

Oh, I get it. When you move the element, we temporarily hide the handles of the moveable, but when they come back, I don't reassign the MutationObserve to them, because I was unaware that they could disappear. So the newly reappeared bottom-right handle is not observed anymore and thus does not cause the menu to re-render.

That's a tough one. I'll create a separate bug for that.

packages/story-editor/src/components/floatingMenu/layer.js
barklund: Wait, there is a menu right there in the screenshot! How did I miss that? I'll take a look.

barklund: Ah, that was a bi-product of this PR yes. I wasn't there before, and I was really confused why it suddenly was. But that was because before I hid the menu if there was no moveable reference, but I hadn't incorporated that into the new work and forgot that we could have a situation with a selection but no moveable reference. So this should work now again after 096ed3409a.

packages/story-editor/src/components/floatingMenu/layer.js
barklund: Changed both parts in 125649ca00.

packages/story-editor/src/components/floatingMenu/layer.js
```diff
@@ -92,16 +104,30 @@ function FloatingMenuLayer() {
...
     const observer = new MutationObserver(updatePosition);
-    MUTATE_SELECTORS.map((selector) => document.querySelector(selector))
-      .filter(Boolean)
-      .forEach((node) => observer.observe(node, { attributes: true }));
+    const node = document.querySelector('.moveable-control-box');
+    observer.observe(node, {
+      attributes: true,
+      subtree: true,
+      attributeFilter: ['style'],
+    });
```
barklund, line 114: @merapi, I fixed the bug with the menu not registering some moveable changes by listening to attribute changes on the entire tree below the control-box element. It sounds expensive, but doesn't actually seem to be. There are up to 18 elements and for e.g. a rotation, all of them change their style attribute, but all the changes are reported in a single observe event with just a list of the 18 change records.

It seems to perform well, but please test it out and see if you can spot any problems. Thanks 🙏

packages/story-editor/src/components/floatingMenu/menu.js
miina: Yep, I might be overthinking it 😛

packages/story-editor/src/components/floatingMenu/layer.js
merapi: Nah, I always do those kinds of measurements in a 1k-1kk loop, but on a potato laptop and many elements, it can make a difference.

Lasso: that second lasso will/should reset the dismiss flag so it's not an issue(?)

packages/story-editor/src/components/floatingMenu/layer.js
merapi: LG, I didn't find any issues.

PR #10518, opened 2022-02-08
New Story Details Modal: Inspector Context and Panel Adjustments

BrittanyIRL:
## Context

We need the document panel in the new publish modal. 

## Summary

Moves the Inspector up in hierarchy to access from header, wires up Document Pane for modal, some style updates.

## Relevant Technical Choices

Inspector Context
- Moving the Inspector Provider up from `Inspector` to [`Layout`](https://github.com/GoogleForCreators/web-stories-wp/pull/10518/files#diff-e07426bc8115232e1efac75b9510d316cfe08e0ec5441725a361446a9badbd32R90) so that we can access the `DocumentPane` in the new Story Details modal (via header publish button).

`Panel` Updates: 
- [Updating the `Title`](https://github.com/GoogleForCreators/web-stories-wp/pull/10518/files#diff-cb4fb18e75e611c59228354d45b53a41ec97a7a885dba78d148a92d8bebe5f58L257) to conditionally render the `IconWrapper` associated with the toggle to collapse because when you can't collapse there's an extra space where the icon should be and when you hover over it there's a gray circle? I'm pretty sure this is the first case of an actually non-collapsing panel but I want to call it out incase this causes a regression I'm unaware of.  
- Passing `props` through on all inspector panels so that we can override `Panel` props - right now I need `name`, `canCollapse` and `isPersistable` - being able to override them is important to preventing weird cross over "persistence" from opened panels in the editor vs modal - they need to persist independently.
- Pulling all of the panel names into constants so that a list of them can be assembled reliably, we need this to conditionally render them in the modal vs all of them in the editor - it feels bad typing loose strings for these so I tied them together. Only exception is the `status` panel which lives in wp-story-editor and i'm not sure sure what to do with it, there's a [note here](https://github.com/GoogleForCreators/web-stories-wp/pull/10518/files#diff-23115102a5db709933b9ff35192fc363c3cbfb70cc9eac1ac7d53a9cc94162d3R72).

wp-story-editor Updates: 
- Added two new exports to `/documentPanel` - `PublishModalDocumentPane` and `IsolatedStatusPanel` 
- Initially I had updated the `DocumentPane` to pass in some props and then decide dynamically what panels to render, but I think the DocumentPanel should be kept as is and used accordingly. Max had a good idea to simplify it. 
- So now in the `inspectorTab` that's getting set up in `wp-story-editor`'s `Layout` and [sent in as a new object to `story-editor` as `publishModal` where those two new components from documentPane are getting added as values](https://github.com/GoogleForCreators/web-stories-wp/pull/10518/files#diff-638798d8686c20fab74b41ca59460253ed4f7e5c796923421e722bb49d06b29aR56). 

## To-do
- [x] Tests updated accordingly 
- [x] See idea through for conditionally rendering 1 pane 
- [x] Fix horizontal styling (scroll) 
- [x] RTL styling fixed 
- [x] New Karma tests 

Separately: 
- Rest of checklist here: https://github.com/GoogleForCreators/web-stories-wp/issues/10115 
- e2e tests to cover functionality of document panel that is WordPress specific
- When publish date moves to the future the copy on the "publish" button should be "Schedule"
- Need to update the visibility logic so that when you private publish immediately it doesn't ask about the checklist and loop you back to the story details modal 😂 
- Also, the `Visibility` panel needs an update

| From this | to this | 
| --- | --- | 
| Screen Shot 2022-02-10 at 2 33 26 PM | Screen Shot 2022-02-10 at 2 33 46 PM | 

## User-facing changes

(Behind feature flag still, when flag is enabled this is what's new) 
| LTR | RTL | 
| --- | --- |
| ltr panel demo | rtl | 

Storybook can demo the panels too - just a very dumb downed version to avoid having to feed a bunch of providers to it.
Screen Shot 2022-02-10 at 2 09 50 PM

## Testing Instructions

- Yup you guessed it! More regression testing! This time it's the document panel that needs it. Everything should work exactly the same as in staging. 
- Now, with `Updated Publish Story Modal` feature flag enabled, hit 'publish' and see the document panel present in the modal. It has some differences compared to the document panel in the editor. I will outline them below, you can see[ figma ](https://www.figma.com/file/ZrO763yvMWJjYeAPsRnNfW/_Editor---Sprint-Feature-Level-Designs_2022?node-id=57%3A14792)for designs.  
  - Modal has: `Publishing`, `Permalink`, `Page Advancement`, `Background Audio`, `Taxonomies` (5) panels 
  - Modal has isolated `Visibility` "panel" below the `Story Description` in the middle section  
  - If you are looking at designs in figma and notice that the visibility panel looks different, yup. I'm just using the existing design. We can update what it looks like separately.
- Open panel persistence should NOT cross between modal and editor. That is to say, if you have all of your document pane panels open in the editor and then open the modal and collapse 'Publishing' and 'Permalink' and then close the modal, the editor's version of those two panels should still be open. 

1.

## Checklist

---

Partially addresses #10115


BrittanyIRL:
@miina if you have a few spare minutes I'd love your eyes on this - to get the Inspector Pane in this new story details modal I needed to reorganize that provider a bit and then add some props to the `DocumentPane` in wp-story-editor to conditionally render bits and pieces. I feel like this is the right track but the pane structure has been well established for a long time and I don't want to cause any regressions. 

Appreciate any and all thoughts and questions!

miina:
@BrittanyIRL Seems reasonable to me, however, would be great if @barklund could check as well -- I believe he originally created the structure of the panels and also had some potential ideas recently for architecture improvements for the panels.

@barklund Could you take a look at the changes in the PR to check if you have any suggestions/thoughts on the approach?

packages/design-system/src/components/datalist/datalist.js
```diff
@@ -202,7 +204,12 @@ const Datalist = forwardRef(function Datalist(
       />
       {isOpen && !disabled && isInline && list}
       {!disabled && !isInline && (
-        <Popup anchor={internalRef} isOpen={isOpen} fillWidth={DEFAULT_WIDTH}>
+        <Popup
+          anchor={internalRef}
+          isOpen={isOpen}
+          fillWidth={DEFAULT_WIDTH}
+          zIndex={zIndex}
```
BrittanyIRL, line 211: z index needs overriding here because by default (2) it's not high enough to be on top of the story details modal

packages/story-editor/src/components/inspector/inspectorProvider.js
```diff
@@ -154,6 +159,13 @@ function InspectorProvider({ inspectorTabs, children }) {
     });
   }
 
+  if (inspectorTabs?.document && isUpdatedPublishModalEnabled) {
```
BrittanyIRL: Once this isn't under a feature flag it can get added to the same check on line 155, for now I did it this way because I didn't want to nest a check for the feature, this felt cleaner.

packages/story-editor/src/components/panels/panel/shared/title.js
```diff
@@ -254,7 +254,11 @@ function Title({
         aria-controls={panelContentId}
         isCollapsed={isCollapsed}
       >
-        <IconWrapper>{canCollapse && <Icons.ChevronDownSmall />}</IconWrapper>
+        {canCollapse && (
+          <IconWrapper>
+            <Icons.ChevronDownSmall />
+          </IconWrapper>
+        )}
```
BrittanyIRL, line 261: if `canCollapse` is `false` then the `IconWrapper` shouldn't render, i don't think we have any cases where this _is_ false - until this fun new feature, so it went unnoticed.  Without this update we get this: 
canCollapse

packages/story-editor/src/components/panels/panel/simplePanel.js
```diff
@@ -31,6 +31,7 @@ function SimplePanel({ children, name, ariaLabel, title, className, ...rest }) {
     <Panel name={name} {...rest}>
       <PanelTitle
         ariaLabel={ariaLabel ?? (typeof title === 'string' ? title : '')}
+        {...rest}
```
BrittanyIRL, line 34: without this we can't know when `canCollapse` is false in [title](https://github.com/GoogleForCreators/web-stories-wp/pull/10518/files#diff-cb4fb18e75e611c59228354d45b53a41ec97a7a885dba78d148a92d8bebe5f58R166-R175) when `SimplePanel` is used

packages/story-editor/src/components/publishModal/constants.js
```diff
@@ -20,3 +20,6 @@ export const INPUT_KEYS = {
   SLUG: 'slug',
   TITLE: 'title',
 };
+
+export const HEADER_BAR_HEIGHT = '44px';
+export const HEADER_BAR_MARGIN = '1px';
```
BrittanyIRL, line 25: Ended up needing these to set height on the `Main` section of the modal to make scroll nice

packages/story-editor/src/components/publishModal/mainContent/index.js
```diff
@@ -21,56 +21,88 @@ import {
...
 `;
 
 const Footer = styled.div`
   grid-area: footer;
+  display: flex;
+  align-items: flex-end;
+  margin: 0 18px 20px 32px;
 `;
 
+const WP_STATUS_PANEL = 'status'; // Unsure what the best way to grab this value is, if any, from wp-story-editor
```
BrittanyIRL: This is my outstanding question for this. It feels wrong to have a hardcoded string that references the status (visibility) panel in WordPress when so much work has been done to separate packages. Should this come from config???

packages/story-editor/src/components/publishModal/stories/index.js
```diff
@@ -23,8 +23,16 @@ import { useCallback, useState } from '@googleforcreators/react';
...
 import { noop } from '../../../utils/noop';
 import { CheckpointContext } from '../../checklist';
+import InspectorContext from '../../inspector/context';
+import { PageAdvancementPanel, SlugPanel } from '../../panels/document';
 import PublishModal from '../publishModal';
 
+const MockDocumentPane = () => (
+  <>
+    <SlugPanel />
+    <PageAdvancementPanel />
```
BrittanyIRL, line 33: Just grabbing these two because they have the least amount of providers to wrap and provide fake data for :)

packages/story-editor/src/karma/fixture/components/documentPane/index.js
```diff
@@ -22,17 +22,23 @@ import {
...
-      <TaxonomiesPanel />
-    </>
+const availablePanels = {
+  [DOCUMENT_PANEL_NAMES.EXCERPT]: ExcerptPanel,
+  [DOCUMENT_PANEL_NAMES.SLUG]: SlugPanel,
+  [DOCUMENT_PANEL_NAMES.PAGE_ADVANCEMENT]: PageAdvancementPanel,
+  [DOCUMENT_PANEL_NAMES.BACKGROUND_AUDIO]: BackgroundAudioPanel,
+  [DOCUMENT_PANEL_NAMES.TAXONOMIES]: TaxonomiesPanel,
+};
+
```
BrittanyIRL: Same as the one in `wp-story-editor`

packages/wp-story-editor/src/components/documentPane/constants.js
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export const WP_DOCUMENT_PANEL_NAMES = {
+  PUBLISH: 'publishing',
+  STATUS: 'status',
```
BrittanyIRL, line 19: part 2 of my outstanding question - so I'm defining `STATUS` here, do we need to pass this through some how to prevent a separate string that maybe doesn't match with another CMS's name for this panel? will it always be status??

maxyinger (approved):
lgtm

packages/wp-story-editor/src/components/documentPane/index.js
```diff
@@ -44,4 +46,32 @@ function DocumentPane() {
   );
 }
 
+export function PublishModalDocumentPane() {
```
BrittanyIRL, line 48: @littlemilkstudio  thanks for this idea, even though it makes this file a bit longer, i think this makes way more sense.

samwhale (commented):
Axe stuff checks out, only thing I'm seeing is that clicking a radio button will publish the story 🤔. I think clicking those radio buttons should not give the user the option to publish the story.

https://user-images.githubusercontent.com/22185279/153910677-d729ec69-9049-4451-aff6-9454ed45db47.mp4

swissspidy:
> Axe stuff checks out, only thing I'm seeing is that clicking a radio button will publish the story

@samwhale This is existing & expected behavior in WordPress unfortunately. WP doesn't support private drafts, so when you select "Private", that means directly publishing the story as private. This matches behavior in the regular WP post editor, because of this limitation in WP.

packages/story-editor/src/components/publishModal/karma/publishModal.karma.js
```diff
@@ -97,6 +97,48 @@ describe('Publish Story Modal', () => {
...
+      const storySlug = await getPublishModalElement('textbox', 'URL slug');
+      // that initial title should give us an initial slug
+      expect(storySlug.getAttribute('value')).toBe('stews-for-long-journeys');
+
+      await fixture.events.focus(storySlug);
+      await fixture.events.keyboard.type(
+        "bilbo's favorite 30 minute rabbit stew"
+      );
+      await fixture.events.keyboard.press('tab');
+      // now we've updated the slug independent of title
```
BrookeGraham, line 117: Thanks for the comments. They are so helpful.

packages/wp-story-editor/src/components/documentPane/index.js
```diff
@@ -44,4 +46,32 @@ function DocumentPane() {
...
+      <SlugPanel name={`story_details_${DOCUMENT_PANEL_NAMES.SLUG}`} />
+      <PageAdvancementPanel
+        name={`story_details_${DOCUMENT_PANEL_NAMES.PAGE_ADVANCEMENT}`}
+      />
+      <BackgroundAudioPanel
+        name={`story_details_${DOCUMENT_PANEL_NAMES.BACKGROUND_AUDIO}`}
+      />
+      <TaxonomiesPanel
+        name={`story_details_${DOCUMENT_PANEL_NAMES.TAXONOMIES}`}
+      />
```
swissspidy: IMO having to override the complete `name` like this is not very intuitive.

It also means something rather internal like `DOCUMENT_PANEL_NAMES` needs to be exported, just because we need to override `name` here.

WDYT about adding a new `context` prop or something (`<TaxonomiesPanel context="publish_modal" />`, `<TaxonomiesPanel context="some_other_place" />`, and then each panel does `name={${context}_${myPanelName}}`?

This way we can also avoid spreading `...rest` in all the panels

samwhale:
> > Axe stuff checks out, only thing I'm seeing is that clicking a radio button will publish the story
> 
> @samwhale This is existing & expected behavior in WordPress unfortunately. WP doesn't support private drafts, so when you select "Private", that means directly publishing the story as private. This matches behavior in the regular WP post editor, because of this limitation in WP.

Ah, interesting. Seems strange to have those rendered as radio buttons if clicking them will close the modal and publish the story. 💭 Maybe in the future we could move those into the toolbar? But that seems out of scope for this ticket.

```
Publish | Publish Privately | Publish Password Protected
```

BrookeGraham (commented):
Tooltips strike again! The z-index on these is gettin placed behind the modal. 
Screen Shot 2022-02-14 at 10 23 02 AM
Screen Shot 2022-02-14 at 10 22 50 AM

packages/wp-story-editor/src/components/documentPane/index.js
BrittanyIRL: 🤔  

I get that overriding the name isn't super intuitive, however it's the one thing that controls the panel collapse state for local storage so it will need an override. This is also the only place that the panels are getting reused, so introducing a new piece of context for it seems more complicated, my opinion. Plus, we'll have to communicate here through each of these that they are part of the modal since it's the only access point that shows the separation. 

The `Panel` component is already doing so much. I get not wanting to spread `...rest`, would you be more ok with this if it's an optional name override in the individual panels, such as `name={overrideName || defaultName}` and a note's included?

BrittanyIRL:
> Ah, interesting. Seems strange to have those rendered as radio buttons if clicking them will close the modal and publish the story. 💭 Maybe in the future we could move those into the toolbar? But that seems out of scope for this ticket.

Left a note about the visibility panel in the PR description - look for `Also, the Visibility panel needs an update`

BrittanyIRL:
> Tooltips strike again! The z-index on these is gettin placed behind the modal. 

Good catch, thank you! Should be all better 

visible tooltips

BrookeGraham (approved):
Sweet.

packages/wp-story-editor/src/components/documentPane/index.js
swissspidy: > so introducing a new piece of context for it seems more complicated

To clarify, I am not suggesting introducing a new piece of (React) context. Just using a separate prop for prefixing the `name` instead of completely overriding it.

Re-read my comment and replace "context" with "prefix" if that makes it more clear.

BrittanyIRL: gotcha, i think we're on the same page then. https://github.com/GoogleForCreators/web-stories-wp/pull/10518/commits/1e7113358135c217423dd8e32f49abe24d76f92a

BrittanyIRL: i had read 'context' and thought you meant in the inspector context and i just didn't want to go there 😂

swissspidy:
> > > Axe stuff checks out, only thing I'm seeing is that clicking a radio button will publish the story
> > 
> > 
> > @samwhale This is existing & expected behavior in WordPress unfortunately. WP doesn't support private drafts, so when you select "Private", that means directly publishing the story as private. This matches behavior in the regular WP post editor, because of this limitation in WP.
> 
> Ah, interesting. Seems strange to have those rendered as radio buttons if clicking them will close the modal and publish the story. 💭 Maybe in the future we could move those into the toolbar? But that seems out of scope for this ticket.
> 
> ```
> Publish | Publish Privately | Publish Password Protected
> ```

Agreed. They are radio buttons for historical reasons and support for "Private" was added later but the UI stayed the same.

BrittanyIRL:
@felipebochehin87 this is the last separate PR that needs regression testing for this new story details modal feature. 

I've outlined what's changed in the 'testing' section of the PR description. For regression testing here I just need a sanity check that the document panel of the editor is functioning the same in this PR as it is in staging. It should be, I only lifted some behind the scenes things, but I want to get your eyes on it before merging. 

Let me know if you have questions!

samwhale (approved):

felipebochehin87:
Tested against https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin, using PR #10518.

1. Regressed Document panel (Publishing, Permalink, Page Advancement, Background Audio, Taxonomies, Story Description, Visibility)
2. Checked open panel persistence
3. Checked everything RTL and LTR

PR #10521, opened 2022-02-08
Editor: Created dummy menus for all element types

barklund:
## Context

This is stage five of #10112. The final structure starts to make sense.

This adds a folder and file structure for all the required elements as well as the structure for defining the menus.

**NB**: These are all dummy buttons and don't do anything nor necessarily look right. It is just all the elements that need to go there (with a dummy icon that might or might not be correct) in the right order.

## Todo

- [x] ~There's a known bug, where the menu width does not update when you switch to a different element (or go from single to multi and vice versa) so the menu isn't correctly centered under the element. Unselect and reselect to see the proper menu and alignment. This should be fixed within this PR.~ fixed!
- [x] ~No menu was designed in Figma for multi-selection (of different types), so I didn't add anything but also didn't hide the menu. But I think I remember seeing something about having alignment options?~ It will be an alignment menu for any multi-selection regardless of type.

## User-facing changes

_Still behind feature-flag so technically not user-facing._

| Element type | Menu |
|-|-|
| Text | Screen Shot 2022-02-08 at 17 10 07 |
| Video | Screen Shot 2022-02-08 at 17 10 02 |
| GIF | Screen Shot 2022-02-08 at 17 09 58 |
| Shape | Screen Shot 2022-02-08 at 17 34 58 |
| Sticker | Screen Shot 2022-02-08 at 17 09 50 |
| Image | Screen Shot 2022-02-08 at 17 09 45 |
| Multi | Screen Shot 2022-02-14 at 17 37 30 |

## Testing Instructions

## Checklist

---

Partially addresses #10112


packages/story-editor/src/components/floatingMenu/elements/borderRadius.js
```diff
@@ -0,0 +1,42 @@
...
+
+const BorderRadius = memo(function BorderRadius() {
+  return (
+    <>
+      <IconButton
+        Icon={Icons.Corner}
+        title={__('Change border radius', 'web-stories')}
+        onClick={() => {}}
+      />
+      <Separator />
```
barklund, line 36: This element includes its own separator, because this element and separator is optional, depending on whether the current shape can have border-radius at all (if non-rectangular, this button and separator will be hidden).

miina:
> No menu was designed in Figma for multi-selection (of different types), so I didn't add anything but also didn't hide the menu. But I think I remember seeing something about having alignment options?

Would be good to get UX feedback about it -- perhaps there won't be an edit bar for multi-selection if it's not in Figma?
Otherwise we'd need to display "Multiple" in the menu bar and that could potentially make it quite long if all the values happen to be "Multiple".

packages/design-system/src/components/contextMenu/components/button.js
```diff
@@ -121,6 +130,7 @@ function ButtonWithRef(
       role="menuitem"
       $isIconMenu={isIconMenu}
       $isHorizontal={isHorizontal}
+      $forcePadding={forcePadding}
```
miina: From recent discussions: These props shouldn't need being transient, see also https://xwp.slack.com/archives/CCCAUAH6F/p1643891495809269 and https://github.com/GoogleForCreators/web-stories-wp/issues/10468 and https://github.com/GoogleForCreators/web-stories-wp/issues/6669#issuecomment-793531492

packages/story-editor/src/components/floatingMenu/menus/selector.js
```diff
@@ -0,0 +1,60 @@
...
+import Video from './video';
+
+const FloatingMenuSelector = memo(function FloatingMenuSelector({
+  selectionTypes,
+}) {
+  if (selectionTypes.length > 1) {
+    return <Mixed />;
+  }
+
+  const [selectionType] = selectionTypes;
```
miina: Same here as in the other PR: `selectedTypes` would be more accurate here.

packages/story-editor/src/components/floatingMenu/menus/mixed.js
```diff
@@ -0,0 +1,36 @@
...
+ * External dependencies
+ */
+import { memo } from '@googleforcreators/react';
+
+/**
+ * Internal dependencies
+ */
+import { Dismiss, Separator } from '../elements';
+
+const FloatingTextMenu = memo(function FloatingTextMenu() {
```
miina: Should be something else, `MultiSelectionMenu` or something similar perhaps?

packages/story-editor/src/components/floatingMenu/elements/toggleUnderline.js
```diff
@@ -0,0 +1,39 @@
...
+import { memo } from '@googleforcreators/react';
+import { Icons } from '@googleforcreators/design-system';
+import { __ } from '@googleforcreators/i18n';
+
+/**
+ * Internal dependencies
+ */
+import { IconButton } from './shared';
+
+const ToggleUnderline = memo(function ToggleUnderline() {
```
miina: Do all these need memoizing currently? 🤔 Could you clarify why?
Isn't it a recommended practice (e.g. by Kent C Dodds) to initially not memo anything to avoid any hidden issues and add `memo` if only if really needed?

barklund: Yes, you are right. The elements were all just a last-minute copy-paste, not thought through. Because the menus are properly memoized and are all completely static, memoizing the elements is irrelevant. The elements themselves will often re-render because of stateful hooks, but never because the parent re-renders.

packages/design-system/src/components/contextMenu/components/button.js
barklund: Excellent points. I have reworked it in c9c0c43dba.

packages/story-editor/src/components/floatingMenu/menus/selector.js
barklund: Yes, fixed in 9529cf1f73.

packages/story-editor/src/components/floatingMenu/menus/mixed.js
barklund: Oh yes, forgot to rename this one. It's temporary anyway. Fixed in 391bb33717.

packages/story-editor/src/components/floatingMenu/elements/toggleUnderline.js
barklund: Fixed in e93628fe29.

packages/story-editor/src/components/floatingMenu/elements/trim.js
```diff
@@ -0,0 +1,38 @@
...
+ */
+import { Icons } from '@googleforcreators/design-system';
+import { __ } from '@googleforcreators/i18n';
+
+/**
+ * Internal dependencies
+ */
+import { IconButton } from './shared';
+
+const Trim = function Trim() {
```
miina: Here and elsewhere
```suggestion
function Trim() {
```

(unless you had a reason for using both `const` and `function` 😛

packages/story-editor/src/components/floatingMenu/elements/textAlign.js
```diff
@@ -0,0 +1,39 @@
...
+import { memo } from '@googleforcreators/react';
+import { Icons } from '@googleforcreators/design-system';
+import { __ } from '@googleforcreators/i18n';
+
+/**
+ * Internal dependencies
+ */
+import { IconButton } from './shared';
+
+const TextAlign = memo(function TextAlign() {
```
miina:
```suggestion
function TextAlign() {
```

miina (approved):

merapi (commented):
LG so far.

I was wondering about this `const FloatingShapeMenu = memo(function FloatingShapeMenu() {`

image

image

And here is the component tree:

image


The last one looks the cleanest? (if you use memo)

packages/story-editor/src/components/floatingMenu/elements/trim.js
barklund: Fixed in a85296e32c.

packages/story-editor/src/components/floatingMenu/elements/textAlign.js
barklund: Fixed in a85296e32c.

barklund:
> I was wondering about this `const FloatingShapeMenu = memo(function FloatingShapeMenu() {`

I was trying to do it identically as to when we're using `forwardRef`, and because you cannot `forwardRef` something with `propTypes` (for weird reasons), it has to be written in this slightly weird manner.

I just kept the same syntax for ease of use, but I actually don't think there are any problems with memoizing components with `propTypes`, so we could do it like that, for sure.

merapi (approved):
https://github.com/GoogleForCreators/web-stories-wp/pull/10521/commits/a85296e32cf445bdf7ec3412dfa1d369f052a341 👍🏻

PR #10524, opened 2022-02-09
Add filter web_stories_markup to allow markup modification before render

freibergerb:
## Context

I would like to be able to filter the markup before render, on our use case we need this to be able to convert urls to the ones configure on our cdn.

## Summary

Adds a filter to allow the possibility of markup modification. 

## Relevant Technical Choices

I believe a filter would be the best choice to allow customisation of the markup.

## To-do

We could add unit tests.

## User-facing changes

None

## Testing Instructions

You would need to use `add_filter` and modify the markup within your own customisation / plugin.

1.

## Checklist

---

Fixes #

### DIFF
--- a/includes/Renderer/Story/HTML.php
+++ b/includes/Renderer/Story/HTML.php
@@ -74,6 +74,16 @@ class HTML {
 		$markup = $this->print_analytics( $markup );
 		$markup = $this->print_social_share( $markup );
 
+		/**
+		 * Filters the markup
+		 *
+		 * @param string $markup Markup.
+		 *
+		 * @since 1.18.0
+		 *
+		 */
+		$markup = apply_filters( 'web_stories_markup', $markup );
+
 		return $markup;
 	}
 


swissspidy:
What kinda URLs do you beed to replace in your stories? Why are they wrong in the first place?

Ideally we would have a much more targeted filter, not a catch-all one that allows completely changing the whole markup.

freibergerb:
@swissspidy thank you for the quick reply.

I first would like to address the urls part of it. They aren't wrong, they are like that by design. Our sites are unaccessible to the public directly through the VIPGO platform. The only way to get there is through our company's VPN. So as you can see we do have different URL's there to the ones which are public facing, let's say news.com.au. This is our CDNs provider public facing URL, which is responsible to hit VIPGO and then serve the content. I cannot go into much + detail publicly.

Yes, adding that filter there wasn't our first option either, I was digging around and I couldn't find any other Wordpress core filter we could use as we aren't working directly with a `WP_Query`, instead getting the content from `$post->post_content` when building the `$markup`.
The only filter I found looking at core is one that isn't even documented but it didn't work as expected, which is https://developer.wordpress.org/reference/hooks/post_field/.

As the markup is stored in a field we need to translate these urls, usually with WP you would have filter for most things you are generating on the fly, so I thought first about placing it in the Story Model object to isolate a bit more https://github.com/GoogleForCreators/web-stories-wp/blob/main/includes/Model/Story.php#L220

Doing it, solves most of it, I'm left with two other places to fix:
https://github.com/GoogleForCreators/web-stories-wp/blob/main/includes/Discovery.php#L91-L92
`rsd_link` relies on `site_url()`: I can use the `site_url ` filter for this.
`wlwmanifest_link` relies on `includes_url`: I can use the `includes_url` filter for this.

If placing on the Model sounds like a better solution we could work with that and add two more filters?

Let me know what you think.

swissspidy:
Thanks for the additional information, that's super helpful!

In that case, I recommend writing a custom AMP sanitizer and hooking into the `web_stories_amp_sanitizers` filter to add it.

This sanitizer must extend the `Google\Web_Stories_Dependencies\AMP_Base_Sanitizer` class.

It will operate on the final DOMDocument instance containing the final story output. You can use that to go through the document and replace all instances of URLs where needed.

Since sanitizers also can add/change URLs (for example links to the publisher logo), this approach ensures that you really catch all URLs needing replacement. This might not be the case when using a catch-all filter like proposed here.

There's some documentation available at https://amp-wp.org/documentation/playbooks/sanitizers/. Just the filter & class names are different in our plugin, but the rest stays the same.

Here's an example of our own custom sanitizer:

https://github.com/GoogleForCreators/web-stories-wp/blob/f89e3063a28d7e16f931854f092f7fe050369f43/includes/AMP/Story_Sanitizer.php
https://github.com/GoogleForCreators/web-stories-wp/blob/f89e3063a28d7e16f931854f092f7fe050369f43/includes/AMP/Traits/Sanitization_Utils.php

freibergerb:
Thank you @swissspidy that's very helpful.

I've implemented that and it works quite well, however I'm still a bit divided by the two different options.

On one hand I see how more appropriate and elegant it's to implement with the Sanitizer. The only trick part is making sure I can gather all the possible options on where tag's attributes can have a link. So far I've been checking `<link>, , <amp-img>, <amp-video>, <video>` then within these tags you can have `src, srcset, poster, artwork`. I'm just concern at how reliable this would be and questioning how can I make sure I'm covering it all!? Any tips here? I've been looking at https://amp.dev/documentation/guides-and-tutorials/learn/spec/amphtml/ .

Said that, I know that the filter on markup seems a bit of a brute force change, but it also work, and it has so many less lines of code to accomplish the same. What's your main concern around it? Security and Design?

freibergerb:
@swissspidy we had a quick chat internally and decided to go with the Sanitizer as per your suggestion. If we end up having any issues or see need for this in the future we can open a new thread and discuss.

I'm closing the PR, thank you for your help and input on this!

PR #10525, opened 2022-02-09
Code Quality Updates after bumping PHP version requirement

### LINKED ISSUE #10151: Code Quality Updates after bumping PHP version requirement
## Task Description

* Remove some compat code we have for PHP 7.0 or even PHP 5.6
* Use newer language features / simplify some code where applicable (the IDE will easily help with this)

Examples:

* `void` return type

spacedmonkey:
## Summary

Make tweaks to PHP to use new features in PHP 7.2. Implement new linting rules to use these features going forward. 

## Relevant Technical Choices

```

 $instantiation = $this->instantiation; // Because uniform variable syntax not supported in PHP 5.6. 
 $service       = $instantiation();
```
Could not be removed without error. 

https://github.com/GoogleForCreators/web-stories-wp/blob/7274e02b0fe5c2d10c03c64d68c1e0db5329f3a8/includes/AMP/Output_Buffer.php#L195-L201

Kept as removing does not improve code. 

No linting rule was found for `list` function. Manaully fixed in PR. 

## Testing Instructions

1.

## Checklist

---

Fixes #10151

### DIFF
--- a/includes/AMP/Meta_Sanitizer.php
+++ b/includes/AMP/Meta_Sanitizer.php
@@ -59,7 +59,7 @@ class Meta_Sanitizer extends AMP_Meta_Sanitizer {
 	 *
 	 * @return void
 	 */
-	protected function ensure_boilerplate_is_present() {
+	protected function ensure_boilerplate_is_present(): void {
 		$style  = null;
 		$styles = $this->dom->xpath->query( './style[ @amp-boilerplate ]', $this->dom->head );
 
@@ -116,7 +116,7 @@ class Meta_Sanitizer extends AMP_Meta_Sanitizer {
 	 *
 	 * @return string[] Stylesheets, where first is contained in style[amp-boilerplate] and the second in noscript>style[amp-boilerplate].
 	 */
-	protected function get_boilerplate_stylesheets() {
+	protected function get_boilerplate_stylesheets(): array {
 		return [
 			'body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}',
 			'body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}',
--- a/includes/Integrations/AMP.php
+++ b/includes/Integrations/AMP.php
@@ -45,7 +45,7 @@ class AMP extends Service_Base implements HasRequirements {
 	/**
 	 * Slug of the AMP validated URL post type.
 	 */
-	const AMP_VALIDATED_URL_POST_TYPE = 'amp_validated_url';
+	public const AMP_VALIDATED_URL_POST_TYPE = 'amp_validated_url';
 
 	/**
 	 * Settings instance.
@@ -91,7 +91,7 @@ class AMP extends Service_Base implements HasRequirements {
 	 *
 	 * @return void
 	 */
-	public function register() {
+	public function register(): void {
 		add_filter( 'option_amp-options', [ $this, 'filter_amp_options' ] );
 		add_filter( 'amp_supportable_post_types', [ $this, 'filter_supportable_post_types' ] );
 		add_filter( 'amp_to_amp_linking_element_excluded', [ $this, 'filter_amp_to_amp_linking_element_excluded' ], 10, 4 );
@@ -232,7 +232,7 @@ class AMP extends Service_Base implements HasRequirements {
 	 * @param array     $error Validation error being sanitized.
 	 * @return null|bool Whether sanitized.
 	 */
-	public function filter_amp_validation_error_sanitized( $sanitized, $error ) {
+	public function filter_amp_validation_error_sanitized( $sanitized, $error ): ?bool {
 		// Skip sanitization for missing publisher logos and poster portrait images.
 		if (
 			( isset( $error['node_type'], $error['node_name'], $error['parent_name'] ) ) &&
@@ -318,7 +318,7 @@ class AMP extends Service_Base implements HasRequirements {
 	 *
 	 * @return string|null
 	 */
-	protected function get_request_post_type() {
+	protected function get_request_post_type(): ?string {
 		// phpcs:disable WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
 
 		if ( did_action( 'wp' ) && is_singular() ) {
@@ -370,7 +370,7 @@ class AMP extends Service_Base implements HasRequirements {
 	 * @param int $post_id Post ID for Validated URL Post.
 	 * @return string|null Post type or null if validated URL is not for a singular post.
 	 */
-	protected function get_validated_url_post_type( $post_id ) {
+	protected function get_validated_url_post_type( $post_id ): ?string {
 		if ( empty( $post_id ) ) {
 			return null;
 		}
--- a/includes/REST_API/Stories_Autosaves_Controller.php
+++ b/includes/REST_API/Stories_Autosaves_Controller.php
@@ -179,7 +179,7 @@ class Stories_Autosaves_Controller extends WP_REST_Autosaves_Controller implemen
 	 *
 	 * @param WP_Post         $post Post object.
 	 * @param WP_REST_Request $request Request object.
-	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
+	 * @return WP_REST_Response Response object.
 	 */
 	public function prepare_item_for_response( $post, $request ) {
 		$response = parent::prepare_item_for_response( $post, $request );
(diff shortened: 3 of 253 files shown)


phpcs.xml.dist
```diff
@@ -86,15 +86,24 @@
...
   </rule>
 
+  <rule ref="SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingNativeTypeHint">
+    <properties>
+      <property name="enableStaticTypeHint" value="false" />
+      <property name="enableMixedTypeHint" value="false" />
+      <property name="enableUnionTypeHint" value="false" />
+    </properties>
+        <exclude-pattern>web-stories.php</exclude-pattern>
+        <exclude-pattern>includes/compat/Web_Stories_Compatibility.php</exclude-pattern>
```
swissspidy: Nit: incorrect indentation.

tests/phpunit/integration/includes/Fixture/DummyInterface.php
```diff
@@ -3,6 +3,5 @@
 namespace Google\Web_Stories\Tests\Integration\Fixture;
 
 interface DummyInterface {
-
-	public function get_dummy();
+	public function get_dummy(); // phpcs:ignore SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingNativeTypeHint
```
swissspidy, line 6: Could just to `: void` here and in `DummyClassWithDependency`?

includes/REST_API/Stories_Media_Controller.php
```diff
@@ -302,7 +302,7 @@ protected function prepare_items_query( $prepared_args = [], $request = null ) {
 	 * @param WP_REST_Request $request Request object.
 	 * @return WP_REST_Response Response object.
 	 */
-	public function prepare_item_for_response( $post, $request ) {
+	public function prepare_item_for_response( $post, $request ): WP_REST_Response {
```
swissspidy, line 306: I thought we wanted to avoid these because of the `apply_filters()` usage

spacedmonkey: This method is not returned a filter is it?

tests/phpunit/integration/includes/Fixture/DummyInterface.php
spacedmonkey: No, the downstream method doesn't return void.

includes/REST_API/Stories_Media_Controller.php
swissspidy: Someone could filter `web_stories_rest_prepare_attachment` below to return something else than a `WP_REST_Response`.

My understanding is that because of this we have never added these return type hints to these methods in the past.

Please correct me if I'm wrong

spacedmonkey: I can add an ignore on this line then.

includes/AMP/Meta_Sanitizer.php
```diff
@@ -116,7 +116,7 @@ protected function ensure_boilerplate_is_present() {
 	 *
 	 * @return string[] Stylesheets, where first is contained in style[amp-boilerplate] and the second in noscript>style[amp-boilerplate].
```
timarney, line 117: doc block says string here - vs array

includes/Integrations/AMP.php
```diff
@@ -232,7 +232,7 @@ public function add_amp_content_sanitizers( $sanitizers ) {
 	 * @param array     $error Validation error being sanitized.
 	 * @return null|bool Whether sanitized.
```
timarney, line 233: Should this be flipped  ?

bool|null

includes/Renderer/Stories/Renderer.php
```diff
@@ -199,7 +199,7 @@ public function render( array $args = [] ): string {
 	 * @return Story|null
 	 */
 	#[\ReturnTypeWillChange]
```
timarney: Thinking this can be removed `#[\ReturnTypeWillChange]`

includes/Renderer/Stories/Renderer.php
```diff
@@ -259,7 +259,7 @@ public function rewind() {
 	 * @return void
 	 */
 	#[\ReturnTypeWillChange]
```
timarney: Thinking this can be removed `#[\ReturnTypeWillChange]`

includes/Integrations/AMP.php
spacedmonkey: Thisis unrelated to my change and should be done in another PR

includes/AMP/Meta_Sanitizer.php
spacedmonkey: This is unrelated to my change and should be done in another PR

spacedmonkey: This means an array of strings `string[]` doesn't it?

includes/REST_API/Stories_Media_Controller.php
spacedmonkey: Fixed

includes/Renderer/Stories/Renderer.php
spacedmonkey: Removed

includes/Integrations/AMP.php
swissspidy: Why would this need to be flipped? What difference does it make? Just for aesthetics?

includes/REST_API/Stories_Media_Controller.php
```diff
@@ -300,7 +304,7 @@ protected function prepare_items_query( $prepared_args = [], $request = null ) {
 	 *
 	 * @param WP_Post         $post    Attachment object.
 	 * @param WP_REST_Request $request Request object.
-	 * @return WP_REST_Response Response object.
+	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
```
swissspidy: Since when can this suddenly be a WP_Error? 🤔  Nothing has changed in the code.

Is this just a sneaky way of avoiding a lint warning  instead of ignoring them?

spacedmonkey: It was always possible this was `WP_Error`, because of the `web_stories_rest_prepare_attachment` filter. 

I am just making clear, it could either be a `WP_REST_Response` or `WP_Error`. If anything that is returned from that filter, that it will just break everything.

includes/AMP/Meta_Sanitizer.php
timarney: Your right I misread that

includes/Integrations/AMP.php
timarney: For the complied code shouldn't matter.

Just for aesthetics --- correct

includes/REST_API/Stories_Media_Controller.php
swissspidy: Since a filter could return literally anything, wouldn't it make more sense for this to be `mixed`?

But I am actually -1 on doing this change here.

`prepare_item_for_response` and `web_stories_rest_prepare_attachment` are not supposed to return errors. They expect proper response objects.

Let's keep the PHPDoc at `@return WP_REST_Response` and ignore any PHPCS warnings about missing return type hints.

includes/Integrations/AMP.php
swissspidy: I don't see a need for changing this, but no strong opinion.

If we'd want to change this for aesthetics this should be something that can be enforced automatically using PHPCS

spacedmonkey: I think we shouldn't change this. Seems fine as it is.

includes/REST_API/Stories_Autosaves_Controller.php
```diff
@@ -178,9 +179,9 @@ public function register_routes() {
 	 *
 	 * @param WP_Post         $post Post object.
 	 * @param WP_REST_Request $request Request object.
-	 * @return WP_REST_Response Response object.
+	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
```
spacedmonkey:
```suggestion
	 * @return WP_REST_Response Response object.
```

includes/REST_API/Stories_Base_Controller.php
```diff
@@ -120,7 +120,7 @@ protected function prepare_item_for_database( $request ) {
 	 *
 	 * @param WP_Post         $post Post object.
 	 * @param WP_REST_Request $request Request object.
-	 * @return WP_REST_Response Response object.
+	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
```
spacedmonkey:
```suggestion
	 * @return WP_REST_Response Response object.
```

includes/REST_API/Stories_Controller.php
```diff
@@ -67,7 +67,7 @@ class Stories_Controller extends Stories_Base_Controller {
 	 *
 	 * @param WP_Post         $post Post object.
 	 * @param WP_REST_Request $request Request object.
-	 * @return WP_REST_Response Response object.
+	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
```
spacedmonkey:
```suggestion
	 * @return WP_REST_Response Response object.
```

includes/REST_API/Stories_Media_Controller.php
spacedmonkey:
```suggestion
	 * @return WP_REST_Response Response object.
```

includes/REST_API/Stories_Media_Controller.php
```diff
@@ -177,7 +177,11 @@ public function create_item( $request ) {
 		}
 
 		$new_response = $this->prepare_item_for_response( $attachment, $request );
-		$data         = $new_response->get_data();
+		if ( is_wp_error( $new_response ) ) {
+			return $new_response;
+		}
```
spacedmonkey:
```suggestion
```

includes/REST_API/Stories_Media_Controller.php
spacedmonkey: Okay, I have reverted the changes and added back the type hinting. There is type hint makes sense here. If you return anything other than a `WP_REST_Response`, it will fail anyway. You agree?

swissspidy (approved):

PR #6533, opened 2021-02-25
Reverted to older popup logic

barklund:
## Context

Reverting to (maybe?) older popup logic, as the previous logic was broken in the editor.

### QA

This PR can be tested by following these steps:

1. Hover the grid view icon in the carousel menu
2. Observe tooltip appearing just over button and not at the top of the browser

### UAT

- [x] UAT should use the same steps as above.

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testiing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [x] This PR contains automated tests (unit, integration, and/or e2e) to verify the code works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [x] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #6496

PR #10539, opened 2022-02-10
Design System: Bugfix for dropdown positioning regression in popup

### LINKED ISSUE #10538: Settings: DropDown opening in wrong position
## Bug Description

A regression snuck through recent efforts to deduplicate popup usage (#10089 ) 

## Expected Behaviour

Open dropdown menu should show either directly below or above (depending on space) its affiliated input. 

## Steps to Reproduce

Go to Settings and open dropdowns after you have scrolled. 

## Additional Context

An important part of this update: https://github.com/GoogleForCreators/web-stories-wp/pull/6533/files was overwritten

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

BrittanyIRL:
## Context

Popups were deduplicated recently (#10089 ) - the offsets that were updated to respect the newer needs of the color picker overruled the initial fix for this positioning issue found here #6533. 

## Summary

Since our popup is a true renaissance component and is the foundation of sooo many different things, we need an escape hatch for when the view scrolls. Because this app is mostly a fixed height, it's rare that we need this update - only really on the dropdowns. 

## Relevant Technical Choices

- Add a new prop called `ignoreMaxOffsetY` to provide a popup option that respects page scroll. 
- Set to false by default, when present it will ignore the found `maxYOffset` in the popup positioning. 
- Useful when scroll will occur on an element. Only comes up in dashboard pages since editor is a fixed height. This is an updated version of the original fix that was overwritten - that fix is in #6533

## User-facing changes

| 😢  | 😄  |
| --- | --- | 
| Screenshot 2022-02-10 at 15 51 29 | Screen Shot 2022-02-10 at 9 30 11 AM |

## Testing Instructions

- See that dropDowns are staying close to their affiliated select buttons (ie: Dashboard Settings, Editor animation selection, Editor authors) 
- See that popups used in editor still show up how they were meant to be - should see no impact on these (ie: links on canvas, color popup, three dot menus on media, tooltips) 
- And then check all of this again for RTL

1.

## Checklist

---

Fixes #10538

### DIFF
--- a/packages/design-system/src/components/popup/index.js
+++ b/packages/design-system/src/components/popup/index.js
@@ -80,6 +80,7 @@ function Popup({
   topOffset = DEFAULT_TOPOFFSET,
   zIndex = DEFAULT_POPUP_Z_INDEX,
   noOverFlow = false,
+  ignoreMaxOffsetY,
 }) {
   const [popupState, setPopupState] = useState(null);
   const isMounted = useRef(false);
@@ -103,12 +104,21 @@ function Popup({
       }
       setPopupState({
         offset: anchor.current
-          ? getOffset(placement, spacing, anchor, dock, popup, isRTL, topOffset)
+          ? getOffset({
+              placement,
+              spacing,
+              anchor,
+              dock,
+              popup,
+              isRTL,
+              topOffset,
+              ignoreMaxOffsetY,
+            })
           : {},
         height: popup.current?.getBoundingClientRect()?.height,
       });
     },
-    [anchor, dock, isRTL, placement, spacing, topOffset]
+    [anchor, dock, isRTL, placement, spacing, topOffset, ignoreMaxOffsetY]
   );
   useEffect(() => {
     // If the popup height changes meanwhile, let's update the popup, too.
@@ -185,6 +195,7 @@ Popup.propTypes = {
   topOffset: PropTypes.number,
   zIndex: PropTypes.number,
   noOverFlow: PropTypes.bool,
+  ignoreMaxOffsetY: PropTypes.bool,
 };
 
 export { Popup, PLACEMENT };
--- a/packages/design-system/src/components/dropDown/index.js
+++ b/packages/design-system/src/components/dropDown/index.js
@@ -195,6 +195,7 @@ export const DropDown = forwardRef(
             refCallback={positionPlacement}
             fillWidth={popupFillWidth}
             zIndex={popupZIndex}
+            ignoreMaxOffsetY
           >
             {menu}
           </Popup>
--- a/packages/design-system/src/components/popup/utils/getOffset.js
+++ b/packages/design-system/src/components/popup/utils/getOffset.js
@@ -23,18 +23,14 @@ export function getXOffset(
   spacing = 0,
   anchorRect,
   dockRect,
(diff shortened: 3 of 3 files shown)


packages/design-system/src/components/popup/index.js
```diff
@@ -103,12 +104,21 @@ function Popup({
       }
       setPopupState({
         offset: anchor.current
-          ? getOffset(placement, spacing, anchor, dock, popup, isRTL, topOffset)
+          ? getOffset({
```
BrittanyIRL, line 107: this is just way too many props to pass them individually, just making it an object instead.

samwhale (approved):
Seeing the fix work locally. 

Found a weird interaction - but may be out of scope. Maybe another ticket?

Opening the lowest dropdown, scrolling up, and then clicking another dropdown jolts the page and doesn't open the new dropdown that was clicked.


https://user-images.githubusercontent.com/22185279/153458934-8b473b4f-fdb1-49ca-8f89-8dec6b4ad935.mp4

BrittanyIRL:
> Seeing the fix work locally.
> 
> Found a weird interaction - but may be out of scope. Maybe another ticket?
> 
> Opening the lowest dropdown, scrolling up, and then clicking another dropdown jolts the page and doesn't open the new dropdown that was clicked.
> 
>  close.open.mp4

Super interesting, I'm able to replicate this but very very sparingly. It kind of makes sense that it needs to close first and if it's unaware of that extra popup then it wouldn't open. I don't see this as a super high priority just because it's not an always thing and it's not really impacting usability. I'll make a separate lower priority issue for this.

------
Edit 

Issue created: https://github.com/GoogleForCreators/web-stories-wp/issues/10540

BrookeGraham:
This lgtm. Didn't find any other regressions, but did notice a couple issues that are happening on main as. Probably outta scope, so maybe we can make a couple tickets. 

When in RTL `Tooltip` is getting a stale value for `placement` or something to that effect. On elements that have a `BOTTOM` placement would be cutoff by the bottom of the screen, it updates to having a `TOP` placement, but only on every other hover. In some instances, it doesn't get updated and instead the anchor point is pushed up instead. 
Kapture 2022-02-10 at 10 21 06

Screen Shot 2022-02-10 at 9 59 45 AM


Similarly when in RTL if a `TOOLTIP` is cutoff the left edge of the screen we run into an infinite loop. Likely due to `bottom` and `end` placement overrides fighting one another? 
Kapture 2022-02-10 at 10 23 48


One other Tooltip place issue I saw was for `Insert Text` quick action, it isn't getting moved when in RTL.
Screen Shot 2022-02-10 at 10 00 27 AM

BrookeGraham (approved):
🚀

BrittanyIRL:
> This lgtm. Didn't find any other regressions, but did notice a couple issues that are happening on main as. Probably outta scope, so maybe we can make a couple tickets.

These are great finds. Can you make tickets for them?

BrookeGraham:
> > This lgtm. Didn't find any other regressions, but did notice a couple issues that are happening on main as. Probably outta scope, so maybe we can make a couple tickets.
> 
> These are great finds. Can you make tickets for them?

yeah!

PR #10548, opened 2022-02-11
Video Element: Extract reusable captions component

spacedmonkey:
## Summary

Breakout from  #10514 based on this comment https://github.com/GoogleForCreators/web-stories-wp/pull/10514#discussion_r803090360. 

This a little code tidy up and make a reusable component. 

## Testing Instructions

1.

## Checklist

---

Fixes #


timarney:
Nice - cleans up the code and makes it more readable at a glance.

swissspidy:
I see why you added this and yes extracting repeated patterns into its own reusable component technically makes things more DRY in theory.

However, IMO just replacing a few `Array.prototype.map` calls doesn't really add much value.

Especially because this touches a lot of different aspects of the editor (design panel, element display, video trimming, AMP output). This abstraction causes a tighter coupling between all these aspects.

This is detrimental to our decoupling efforts, where we want to split these aspects into their own packages. See #8125 and #8777 for example. Some initial work on this is actually already in progress in #10452.

This package separation would mean either moving such a `Captions` component to its own package (or the design system, but that feels wrong) so all others can use it, or, more straightforward, just don't do this change at all.

With this context, I'm definitely -1 on making this change.

We can revisit this after the decoupling and package organization work has further progressed.

spacedmonkey:
Okay, maybe just scale this back a little, so the repeat logic in `video/display.js`, `video/trim.js` and `video/output.js` can grouped together. These are all inside the video element and are unlikely to to decoupled, right?

swissspidy:
> Okay, maybe just scale this back a little, so the repeat logic in `video/display.js`, `video/trim.js` and `video/output.js` can grouped together. These are all inside the video element and are unlikely to to decoupled, right?

Yeah these would stay together.

But even there I still don't see much benefit in doing that. No strong opinion though, but I feel like there are better places in our code bases in more dire need of a refactor :-)

packages/story-editor/src/elements/video/display.js
```diff
@@ -118,19 +119,7 @@ function VideoDisplay({ previewMode, box: { width, height }, element }) {
...
-                label={label}
-                // Hides the track from the user.
-                // Displaying happens in MediaCaptionsLayer instead.
-                kind="metadata"
-                src={src}
-                key={key}
-                default={i === 0}
-              />
-            ))}
+          <Captions tracks={tracks} defaultKind="metadata" />
```
swissspidy: This completely changes the semantics.

Before, we _force_ the `kind` to be `metadata`. Now this just says "use metadata as the default if none is provided"

packages/story-editor/src/elements/video/display.js
```diff
@@ -118,19 +119,7 @@ function VideoDisplay({ previewMode, box: { width, height }, element }) {
           data-leaf-element="true"
         >
           {url && <source src={url} type={resource.mimeType} />}
-          {tracks &&
-            tracks.map(({ srclang, label, track: src, id: key }, i) => (
-              <track
-                srcLang={srclang}
-                label={label}
-                // Hides the track from the user.
-                // Displaying happens in MediaCaptionsLayer instead.
```
swissspidy, line 130: This useful comment is now lost

packages/story-editor/src/elements/video/display.js
spacedmonkey:
```suggestion
          <Captions tracks={tracks} />
```

packages/story-editor/src/elements/video/display.js
spacedmonkey: Is this really a problem?

swissspidy: It helps explain why it‘s needed, so yes

packages/story-editor/src/elements/video/captions.js
```diff
@@ -0,0 +1,43 @@
...
+function Captions({ tracks, defaultKind = 'metadata' }) {
+  if (!tracks) {
+    return null;
+  }
+
+  return tracks.map(({ srclang, label, kind, track, id: key }, i) => (
+    <track
+      srcLang={srclang}
+      label={label}
+      kind={kind || defaultKind}
```
spacedmonkey:
```suggestion
      // Hides the track from the user
      // Displaying happens in MediaCaptionsLayer instead.
      kind={kind || defaultKind}
```

swissspidy: That comment should be in `display.js`, not here

swissspidy:
Converting to draft to save some resources like Percy snapshots until all open comments are resolved

packages/story-editor/src/elements/video/captions.js
spacedmonkey: Comment moved.

packages/story-editor/src/elements/video/display.js
swissspidy: @spacedmonkey This still is not fixed.

packages/story-editor/src/elements/video/display.js
```diff
@@ -117,20 +118,13 @@ function VideoDisplay({ previewMode, box: { width, height }, element }) {
...
-                default={i === 0}
-              />
-            ))}
+          {
+            url && <source src={url} type={resource.mimeType} />
+
+            // Hides the track from the user
+            // Displaying happens in MediaCaptionsLayer instead.
+          }
+          <Captions tracks={tracks} />
```
swissspidy: This is where you need to actually set `kind="metadata"`

The other components don't need that.

`defaultKind` doesn't really achieve that

swissspidy:
There are still lint errors here.

Also some e2e test failures, but maybe those are just flakey.

spacedmonkey:
This is still awaiting a review.

timarney:
I haven't kept on on the comments for this one but LGTM.

PR #10554, opened 2022-02-12
Media: Do not trim all instances of video

### LINKED ISSUE #10409: Media: Video Trimming affects all instances of video
## Bug Description

This is a follow-up to https://github.com/GoogleForCreators/web-stories-wp/pull/9945#issuecomment-1007467673

When trimming a video within a story, currently all instances of that video within the story are being trimmed.

The `useProcessMedia` hook uses `updateElementsByResourceId` to update all instances of a video item within the story.

This makes sense for optimization and probably also muting. But for video trimming this is confusing, as it's not really expected behavior as a user.

## Expected Behaviour

When trimming one instance of a video, it should only trim the current instance, not all instances of it.

We should not use `updateElementsByResourceId` for trimming.

## Steps to Reproduce

1. Drag video from media library to a page
2. Drag the same video from media library to a page (same page or new one)
3. Trim first video
4. Observe that both are now trimmed once completed

## Additional Context

- Plugin Version: 1.16.0
- WordPress Version: 5.9
- Operating System: macOS
- Browser: Chrome

timarney:
## Context

When trimming a video within a story, currently all instances of that video within the story are being trimmed.

See: https://github.com/GoogleForCreators/web-stories-wp/issues/10409

## Summary

The existing code uses `canvasResourceId` & `updateExistingElements(canvasResourceId,  ...)` to updating existing elements.

See: https://github.com/GoogleForCreators/web-stories-wp/pull/10554/files#diff-284f5d5d8d1215d90a8fa6d991491806595892db0619411fbbea3b629868d366R66

When adding multiple copies of the same video element the canvasResourceId ends up getting duplicated "not unique"

When calling updateExistingElements with the canvasResourceId --- any elements with that ID will get updated

| Element # 1  |      Element # 2     | 
|----------|:-------------:|
| 153723757-df3bedd4-41c8-4a58-85b5-11a834e552c2 |  153723775-d4c56683-ea13-47d7-b8f3-a2788db846e7 |

## Relevant Technical Choices

This PR updates the trimming functions to update existing elements by element ID (which should be unique).

## To-do

Write tests as needed

## User-facing changes

Bug fix for existing trim functionality 

## Testing Instructions

1) Drag video from media library to a page
2)  Drag the same video from media library to a page (same page or new one)
3) Trim first video
4) Observe that only the item that selected to be trimmed was updated

No

No

No

## Checklist

Fixes #10409

### DIFF
--- a/packages/story-editor/src/app/media/utils/useProcessMedia.js
+++ b/packages/story-editor/src/app/media/utils/useProcessMedia.js
@@ -66,11 +66,11 @@ function useProcessMedia({
   );
 
   const copyResourceDataByElementId = useCallback(
-    ({ oldResource, resource }) => {
-      const { id, alt } = oldResource;
+    ({ elementId, oldResource, resource }) => {
+      const { alt } = oldResource;
 
       updateElementById({
-        elementId: id,
+        elementId,
         properties: () => {
           return {
             type: resource.type,
@@ -219,7 +219,7 @@ function useProcessMedia({
    * @param {string} end Time stamp of end time of new video. Example '00:02:00'.
    */
   const trimExistingVideo = useCallback(
-    ({ resource: oldResource, elementId, start, end }) => {
+    ({ resource: oldResource, canvasResourceId, elementId, start, end }) => {
       const { id: resourceId, ...oldResourceWithoutId } = oldResource;
       const { src: url, mimeType, poster, isMuted, isOptimized } = oldResource;
 
@@ -243,9 +243,10 @@ function useProcessMedia({
       const onUploadSuccess = ({ resource }) => {
         const oldCanvasResource = {
           alt: oldResource.alt,
-          id: elementId,
+          id: canvasResourceId,
         };
         copyResourceDataByElementId({
+          elementId,
           oldResource: oldCanvasResource,
           resource,
         });
@@ -253,7 +254,7 @@ function useProcessMedia({
       };
 
       const onUploadProgress = ({ resource }) => {
-        const newResourceWithCanvasId = { ...resource, id: elementId };
+        const newResourceWithCanvasId = { ...resource, id: canvasResourceId };
         updateExistingElement(elementId, {
           ...newResourceWithCanvasId,
         });
@@ -293,7 +294,7 @@ function useProcessMedia({
             ...oldResourceWithoutId,
             trimData,
           },
-          originalResourceId: elementId,
+          originalResourceId: canvasResourceId,
           posterFile,
         });
       };
--- a/packages/story-editor/src/app/media/utils/useProcessMedia.js
+++ b/packages/story-editor/src/app/media/utils/useProcessMedia.js
@@ -85,7 +85,7 @@ function useProcessMedia({
     [updateElementById]
   );
 
-  const updateExistingElement = useCallback(
+  const updateExistingElementById = useCallback(
     (elementId, resource) => {
       updateElementById({
         elementId,
@@ -102,7 +102,7 @@ function useProcessMedia({
     [updateElementById]
   );
 
-  const updateExistingElements = useCallback(
+  const updateExistingElementsByResourceId = useCallback(
     (resourceId, resource) => {
       updateElementsByResourceId({
         id: resourceId,
@@ -152,7 +152,7 @@ function useProcessMedia({
       const { id: resourceId, src: url, mimeType } = oldResource;
 
       const onUploadError = () =>
-        updateExistingElements(resourceId, {
+        updateExistingElementsByResourceId(resourceId, {
           isOptimized: false,
         });
 
@@ -166,7 +166,7 @@ function useProcessMedia({
       // TODO: Confirm which properties exactly need to be updated.
       const onUploadProgress = ({ resource }) => {
         const oldResourceWithId = { ...resource, id: oldResource.id };
-        updateExistingElements(resourceId, {
+        updateExistingElementsByResourceId(resourceId, {
           ...oldResourceWithId,
         });
       };
@@ -176,7 +176,7 @@ function useProcessMedia({
 
         // This video was optimized before, no need to optimize it again.
         if (optimizedResource) {
-          updateExistingElements(resourceId, optimizedResource);
+          updateExistingElementsByResourceId(resourceId, optimizedResource);
           return;
         }
 
@@ -201,7 +201,7 @@ function useProcessMedia({
       })();
     },
     [
-      updateExistingElements,
+      updateExistingElementsByResourceId,
       copyResourceData,
       updateOldTranscodedObject,
       deleteMediaElement,
@@ -231,12 +231,12 @@ function useProcessMedia({
       };
 
       const onUploadStart = () =>
-        updateExistingElement(elementId, {
+        updateExistingElementById(elementId, {
           trimData,
         });
 
       const onUploadError = () =>
-        updateExistingElement(elementId, {
+        updateExistingElementById(elementId, {
           trimData: oldResource.trimData || {},
         });
 
@@ -255,7 +255,7 @@ function useProcessMedia({
 
       const onUploadProgress = ({ resource }) => {
         const newResourceWithCanvasId = { ...resource, id: canvasResourceId };
-        updateExistingElement(elementId, {
+        updateExistingElementById(elementId, {
           ...newResourceWithCanvasId,
         });
       };
@@ -301,7 +301,7 @@ function useProcessMedia({
       return process();
     },
     [
-      updateExistingElement,
+      updateExistingElementById,
       copyResourceDataByElementId,
       postProcessingResource,
       uploadMedia,
@@ -319,7 +319,7 @@ function useProcessMedia({
       const { src: url, mimeType, poster, isOptimized } = oldResource;
 
       const onUploadError = () => {
-        updateExistingElements(resourceId, {
+        updateExistingElementsByResourceId(resourceId, {
           isMuted: false,
         });
       };
@@ -333,7 +333,7 @@ function useProcessMedia({
       // TODO: Confirm which properties exactly need to be updated.
       const onUploadProgress = ({ resource }) => {
         const oldResourceWithId = { ...resource, id: oldResource.id };
-        updateExistingElements(resourceId, {
+        updateExistingElementsByResourceId(resourceId, {
           ...oldResourceWithId,
         });
       };
@@ -343,7 +343,7 @@ function useProcessMedia({
 
         // This video was muted before, no need to mute it again.
         if (mutedResource) {
-          updateExistingElements(resourceId, mutedResource);
+          updateExistingElementsByResourceId(resourceId, mutedResource);
           return;
         }
 
@@ -384,7 +384,7 @@ function useProcessMedia({
       })();
     },
     [
-      updateExistingElements,
+      updateExistingElementsByResourceId,
       copyResourceData,
       updateOldMutedObject,
       postProcessingResource,
@@ -412,7 +412,7 @@ function useProcessMedia({
       // TODO: Confirm which properties exactly need to be updated.
       const onUploadProgress = ({ resource }) => {
         const oldResourceWithId = { ...resource, id: oldResource.id };
-        updateExistingElements(resourceId, {
+        updateExistingElementsByResourceId(resourceId, {
           ...oldResourceWithId,
         });
       };
@@ -447,7 +447,7 @@ function useProcessMedia({
       updateOldTranscodedObject,
       deleteMediaElement,
       postProcessingResource,
-      updateExistingElements,
+      updateExistingElementsByResourceId,
       uploadMedia,
     ]
   );
--- a/packages/story-editor/src/app/media/utils/useProcessMedia.js
+++ b/packages/story-editor/src/app/media/utils/useProcessMedia.js
@@ -38,13 +38,17 @@ function useProcessMedia({
   const {
     actions: { getOptimizedMediaById, getMutedMediaById },
   } = useAPI();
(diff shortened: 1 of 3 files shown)


includes/Media/Video/Trimming.php
```diff
@@ -74,18 +74,22 @@ public function register_meta() {
...
+							'start'     => [
 								'description' => __( 'Start time.', 'web-stories' ),
 								'type'        => 'string',
 							],
-							'end'      => [
+							'end'       => [
 								'description' => __( 'End time.', 'web-stories' ),
 								'type'        => 'string',
 							],
+							'elementId' => [
```
timarney: Was getting a REST error --- will see if this can be removed

swissspidy: This definitely needs to be removed.

You're getting the warnings because you are accidentally passing `elementId` to the underlying REST API call.

This happens because you added `elementId` to `trimData` in `trimExistingVideo()`. That needs to be removed.

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -42,9 +42,14 @@ function useProcessMedia({
     updateElementsByResourceId: state.actions.updateElementsByResourceId,
   }));
 
+  const { updateElementById } = useStory((state) => ({
```
swissspidy: Can just merge this with the `useStory()` call above.

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -178,38 +220,42 @@ function useProcessMedia({
...
+          oldResource: oldCanvasResource,
+          resource,
+        });
         postProcessingResource(resource);
       };
 
       const onUploadProgress = ({ resource }) => {
-        const newResourceWithCanvasId = { ...resource, id: canvasResourceId };
-        updateExistingElements(resourceId, {
+        const newResourceWithCanvasId = { ...resource, id: elementId };
```
swissspidy: This would override the resource ID with the element ID. We don't want that. Instead, we want to preserve the original resource ID.

timarney: Noting: We have a reference to the `original` resourceId under `original`

```
const trimData = {
        original: resourceId,
        start,
        end,
        elementId, // I might be able to remove this completely
      };
```

includes/Media/Video/Trimming.php
timarney: Noting this was added to the trim data --- to keep a temporary reference (while trimming) to the item selected to be trimmed  ... after trimming it should be removed.

If there's a better way to hold this reference that would be ideal.

timarney: Given most of the code in the PR has been updated to use updateElementById vs updating by Resource ID we might not need to hold the elementId in the trim data at all.

spacedmonkey:
Storing the element id isn't going to work here. What if the same trimmed video is used in multiple place in one story? Or in multiple stories. This fix doesn't feel workabout.

timarney:
@spacedmonkey 

> Storing the element id isn't going to work here. 

The code as is isn't finished yet --- what I'm trying to accomplish here is keeping a reference the element that the user is intending to be trimmed 👉 but only while the element is being trimmed 👈 .   You would have more context around holding a longer reference.

With the current code on main multiple elements can end up with the same CanvasResourceID --- when a trim is requested and using the CanvasResourceID as the item to be trimmed they all get trimmed.

I did a test setting a window variable to hold the reference to the the item to be trimmed see https://github.com/GoogleForCreators/web-stories-wp/pull/10555.
```
// for demo purposes only --- this will not remain in the code
      if (window.TEMP_TRIM_ID) {
        if (element.id == window.TEMP_TRIM_ID && element.resource?.id === id) {
          console.log(`found trim item ${element.id}`, element);
```

===

This PR is trying to go down down that road (and the trimming part works) --- the larger change here is `updateElementById` potentially I don't need to hold the elementID (temp variable) in the trim data at all.

swissspidy:
> Storing the element id isn't going to work here. What if the same trimmed video is used in multiple place in one story? Or in multiple stories. This fix doesn't feel workabout.

Not sure I follow you here 🤔 

We don't care about the same video being used in multiple stories. And we also don't care about the same trimmed video being used multiple times _in the same story_. The whole point of this ticket is to not care about it.

includes/Media/Video/Trimming.php
spacedmonkey: Why does this need to saved at all. Couldn't we just delete it before submitted to api?

```
delete trimData.elementId;
```

swissspidy: It doesn't need to be saved, so yes it obviously needs to be removed at some point.

I think we are all on the same page that this is just transient data.

timarney: > Why does this need to saved at all. Couldn't we just delete it before submitted to api?

yes for sure 

> I think we are all on the same page that this is just transient data.

yes

spacedmonkey: You can delete it here

https://github.com/GoogleForCreators/web-stories-wp/blob/27c547c73de98ae361f340cc025122ece0441b9f/packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js#L247

packages/story-editor/src/components/videoTrim/provider.js
```diff
@@ -56,15 +56,15 @@ function VideoTrimProvider({ children }) {
       return;
     }
     const lengthInSeconds = Math.round(endOffset / 1000 - startOffset / 1000);
+
```
spacedmonkey: New line?

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -248,15 +293,15 @@ function useProcessMedia({
             ...oldResourceWithoutId,
             trimData,
           },
-          originalResourceId: canvasResourceId,
+          originalResourceId: elementId,
```
spacedmonkey: This now seems wrong. You are using a element id in place of a resource id.

packages/story-editor/src/app/media/utils/useProcessMedia.js
spacedmonkey: Why are we using the element id in place of a resource id here. This is really confusing to me.

packages/story-editor/src/components/videoTrim/provider.js
```diff
@@ -56,15 +56,15 @@ function VideoTrimProvider({ children }) {
...
     trimExistingVideo({
       resource: {
         ...resource,
         length: lengthInSeconds,
         lengthFormatted: getVideoLengthDisplay(lengthInSeconds),
       },
       // This is the ID of the resource that's currently on canvas and needs to be cloned.
-      // It's only different from the above resource, if the canvas resource is a trim of the other.
-      canvasResourceId: element.resource.id,
```
spacedmonkey, line 67: Why does this need to be removed?

packages/story-editor/src/components/videoTrim/provider.js
timarney: removed

packages/story-editor/src/components/videoTrim/provider.js
timarney: As far as I can see it's not used anywhere else.  Correct me if I'm wrong but it's only used as a parameter for calling `trimExistingVideo `

Which I updated to 
```
const trimExistingVideo = useCallback(
    ({ resource: oldResource, elementId, start, end }) => {
 ```

File: packages/story-editor/src/app/media/utils/useProcessMedia.js (hunk 5 above)
Comment (timarney) [reply packages/story-editor/src/app/media/utils/useProcessMedia.js]: This one was going to ask you about and I want to dig a little further to understand that that is doing.

We have a reference to the `"original_id: resourceId"` under `"additionalData"` which is passed to that function.   So is "originalResourceId" just a temporary reference ?

```
 additionalData: {
            original_id: resourceId,
            web_stories_is_muted: isMuted,
            web_stories_media_source: isOptimized
              ? 'video-optimization'
              : 'editor',
          },
```

File: packages/story-editor/src/components/videoTrim/provider.js (hunk 6 above)
Comment (spacedmonkey) [reply packages/story-editor/src/components/videoTrim/provider.js:67]: It is used here https://github.com/GoogleForCreators/web-stories-wp/blob/e00e50da7a02d956b993f4038ca1c15d46eda3ef/packages/story-editor/src/app/media/utils/useProcessMedia.js#L211

I don't love the idea of elemnt and resource having the same id. It doesn't make it a unique identifier anymore and makes this code confusing and hard to follow.

Comment (swissspidy) [reply packages/story-editor/src/components/videoTrim/provider.js:67]: That ID could theoretically be derived from the element.

As discussed yesterday I am going through this PR in detail. Haven't gotten around to do that yet, but it's high up in my todo list.

Let's hold off on reviewing anything until then.

File: packages/story-editor/src/app/media/utils/useProcessMedia.js (hunk 5 above)
Comment (spacedmonkey) [reply packages/story-editor/src/app/media/utils/useProcessMedia.js]: Why was this changed? Why does `originalResourceId` have to be the elementId?

**Comment by swissspidy** [comment]
@timarney Just had an initial look at this. As @spacedmonkey already pointed out, the resource ID should not be overridden with the element ID, so I made that change in 298ee26614e8951a5745f4ae8e46f3cb904ff090.

Now this already works quite well.

A couple of minor nits in this PR:

* Adding `elementId` to `trimData` doesn't feel great, but I can't think of a better place right now without adding new fields everywhere in the media upload queue. So I suppose we can live with it.
* The naming of all these callbacks (`copyResourceData` / `copyResourceDataByElementId` / `updateExistingElement` / `updateExistingElements`) is a bit confusing, maybe we can find better names

Also a general observation:

When adding the same video multiple times to the canvas and trimming only one of them, each video's "Trim" button changes to "Trimming..." and gets disabled, because technically the underlying resource actually is being trimmed.

However, trimming one video doesn't really affect the others, so their buttons shouldn't get disabled in that case.

Not sure what's a good way to address that, but it feels like it can be its own ticket as it could be a bigger lift. @spacedmonkey Would appreciate your thoughts on that.

File: packages/story-editor/src/app/media/utils/useProcessMedia.js [hunk 7]
```diff
@@ -38,13 +38,17 @@ function useProcessMedia({
...
+    (state) => ({
+      updateElementsByResourceId: state.actions.updateElementsByResourceId,
+      updateElementById: state.actions.updateElementById,
+    })
+  );
 
   const copyResourceData = useCallback(
     ({ oldResource, resource }) => {
       const { id, alt } = oldResource;
+
```
spacedmonkey, line 51:
```suggestion
```

spacedmonkey:
> The naming of all these callbacks (copyResourceData / copyResourceDataByElementId / updateExistingElement / updateExistingElements) is a bit confusing, maybe we can find better names

Naming here is important to make this code readable. 

> When adding the same video multiple times to the canvas and trimming only one of them, each video's "Trim" button changes to "Trimming..." and gets disabled, because technically the underlying resource actually is being trimmed.

Yeah, I remember discussing this at the time. It is extremely unlikely the a user would be trimming the same video at the time. But the implementation blocks that by design. Consider the user story. 

1. User insides video. 
2. User trims video. Trim starts 
3. User insides video again. 
4. User tries to trim video, functionality is disabled. 

Now that update is based on element id and resource id, it does open the door to allow for multiple trimming of the same video. I do think this should be done in a follow up ticket.

spacedmonkey (approved):
Approved with one nic pic.

packages/story-editor/src/app/media/utils/useProcessMedia.js
swissspidy: What's wrong with some extra breathing room?

swissspidy:
> Naming here is important to make this code readable.

Any ideas for better names?

> Now that update is based on element id and resource id, it does open the door to allow for multiple trimming of the same video. I do think this should be done in a follow up ticket.

I'll create the ticket, but I am curious to hear some initial thoughts on how it could be fixed.

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -61,6 +65,43 @@ function useProcessMedia({
...
+              alt,
+            },
+          };
+        },
+      });
+    },
+    [updateElementById]
+  );
+
+  const updateExistingElement = useCallback(
```
spacedmonkey:
```suggestion
  const updateExistingElementById = useCallback(
```

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -61,6 +65,43 @@ function useProcessMedia({
...
+              ...resource,
+            },
+          };
+        },
+      });
+    },
+    [updateElementById]
+  );
+
   const updateExistingElements = useCallback(
```
spacedmonkey:
```suggestion
  const updateExistingElementsByResourceId = useCallback(
```

spacedmonkey:
I have add some recommendations on naming. https://github.com/GoogleForCreators/web-stories-wp/pull/10554#discussion_r809282631 https://github.com/GoogleForCreators/web-stories-wp/pull/10554#discussion_r809282991

timarney:
Let me know if there's any action items to update on this one or if we'll move to new tickets.

spacedmonkey:
Other than maybe renaming those functions, I think we are good to go.

timarney:
> Other than maybe renaming those functions, I think we are good to go.

Updated

swissspidy (approved):

spacedmonkey (approved):

swissspidy:
Follow-up: #10641

PR #10573, opened 2022-02-14
Add AMP sanitizer for adding semantic heading tags

### LINKED ISSUE #10562: Semantic Headings: Update existing stories to use new
## Feature Description

The new semantic headings feature only applies to newly created stories or when updating existing ones.

Ideally older stories benefit from this feature as well.

We can achieve this by using our custom AMP sanitizer to update stories on the fly.

swissspidy:
## Context

The new semantic headings feature only applies to newly created stories or when updating existing ones.

To make the most out of it, we want to "upgrade" existing stories on the fly using an AMP sanitizer.

For context, AMP sanitizers are run on the frontend when viewing stories to ensure they're valid AMP.

## Summary

This adds a new AMP sanitizer copying the JS logic for adding semantic heading tags to that existing stories benefit from this new feature as well, without having to update the stories in the editor.

## User-facing changes

No visual changes

## Testing Instructions

1. **Disable** the semantic headings experiment
2. Create a new story and insert a few pages with a few text elements on each, using the "Paragraph" preset
3. For each text element, change the font size to something larger or smaller, so you end up with all sorts of differently-sized text elements
3. Preview the story
4. The text elements should be wrapped in `<h1>`/`<h2>`/`<h3>`/`<p>` tags, depending on their font sizes

No

No

No

## Checklist

---

Fixes #10562

### DIFF
--- a/includes/AMP/Sanitization.php
+++ b/includes/AMP/Sanitization.php
@@ -27,6 +27,7 @@
 namespace Google\Web_Stories\AMP;
 
 use DOMElement;
+use Google\Web_Stories\Experiments;
 use Google\Web_Stories\Model\Story;
 use Google\Web_Stories\Settings;
 use Google\Web_Stories\Story_Post_Type;
@@ -62,15 +63,24 @@ class Sanitization {
 	private $settings;
 
 	/**
+	 * Experiments instance.
+	 *
+	 * @var Experiments Experiments instance.
+	 */
+	private $experiments;
+
+	/**
 	 * Analytics constructor.
 	 *
 	 * @since 1.12.0
 	 *
-	 * @param Settings $settings Settings instance.
+	 * @param Settings    $settings    Settings instance.
+	 * @param Experiments $experiments Experiments instance.
 	 * @return void
 	 */
-	public function __construct( Settings $settings ) {
-		$this->settings = $settings;
+	public function __construct( Settings $settings, Experiments $experiments ) {
+		$this->settings    = $settings;
+		$this->experiments = $experiments;
 	}
 
 	/**
@@ -437,10 +447,11 @@ class Sanitization {
 			];
 
 			$sanitizers[ Story_Sanitizer::class ] = [
-				'publisher_logo' => $story->get_publisher_logo_url(),
-				'publisher'      => $story->get_publisher_name(),
-				'poster_images'  => array_filter( $poster_images ),
-				'video_cache'    => $video_cache_enabled,
+				'publisher_logo'    => $story->get_publisher_logo_url(),
+				'publisher'         => $story->get_publisher_name(),
+				'poster_images'     => array_filter( $poster_images ),
+				'video_cache'       => $video_cache_enabled,
+				'semantic_headings' => $this->experiments->is_experiment_enabled( 'semanticHeadingTags' ),
 			];
 		}
 
--- a/includes/AMP/Traits/Sanitization_Utils.php
+++ b/includes/AMP/Traits/Sanitization_Utils.php
@@ -115,6 +115,112 @@ trait Sanitization_Utils {
 	}
 
 	/**
+	 * Transforms paragraphs to use semantic heading tags if needed.
+	 *
+	 * This logic here mirrors the getTextElementTagNames() function in the editor
+	 * in order to change simple <p> tags into <h1>, <h2> or <h3>, depending on font size.
+	 *
+	 * Only relevant for older stories that haven't been updated in a while.
+	 *
+	 * @since 1.18.0
+	 *
+	 * @param Document|AMP_Document $document   Document instance.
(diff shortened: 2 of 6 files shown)


includes/AMP/Traits/Sanitization_Utils.php
```diff
@@ -114,6 +114,139 @@ private function transform_a_tags( &$document ): void {
...
+
+			if ( ! preg_match( '/font-size:([^em]+)em/', $style, $matches ) ) {
+				continue;
+			}
+
+			// Contains the font-size in em.
+			// This is basically reversing the dataToFontSizeY() logic. Example:
+			// 0.582524em roughly equals 36 editor pixels: 0.582524 * 618 / 10 = 35.9999px.
+			$font_size_in_em = $matches[1];
+			$font_size_in_px = round( $font_size_in_em * $page_height / 10, 0 );
```
spacedmonkey, line 194: Do we need to calculate this on every request? Especially if it never changes?

includes/AMP/Traits/Sanitization_Utils.php
```diff
@@ -114,6 +114,139 @@ private function transform_a_tags( &$document ): void {
...
+
+		/**
+		 * The <amp-story-page> element
+		 *
+		 * @var DOMElement $page The <amp-story-page> element
+		 */
+		foreach ( $pages as $page ) {
+			$h1 = $page->getElementsByTagName( 'h1' );
+			$h2 = $page->getElementsByTagName( 'h2' );
+			$h3 = $page->getElementsByTagName( 'h3' );
```
spacedmonkey, line 146: Instead of doing this, a string search might be less expensive, resource wise.

swissspidy: We're operating on a DOMDocument instance here, can't really perform a string search on that.

includes/AMP/Traits/Sanitization_Utils.php
swissspidy: Feels more explanatory than just a plain comment with some hardcoded values. And it's not like it's slow.

includes/AMP/Traits/Sanitization_Utils.php
```diff
@@ -114,6 +114,139 @@ private function transform_a_tags( &$document ): void {
...
+
+			if ( $font_size_in_px >= 36 && ! $has_h1 ) {
+				$this->change_tag_name( $text_el, 'h1' );
+				$has_h1 = true;
+				continue;
+			}
+
+			if ( $font_size_in_px >= 27 ) {
+				$this->change_tag_name( $text_el, 'h2' );
+			} elseif ( $font_size_in_px >= 21 ) {
```
timarney, line 204: Wondering if H3s should cover more sizes (21-26) vs H2s will take all the other sizes once there's an H1

When I tested I ended up with a single H3 

Screen Shot 2022-02-15 at 5 52 05 AM

swissspidy: Great question. Since this is just a copy of the original implementation on the JS side to match our text presets (see the epic for context), any discussion about the algorithm is best held in a separate ticket.

swissspidy:
@timarney @spacedmonkey Friendly ping, I've re-requested your review here 2 days ago

includes/AMP/Traits/Sanitization_Utils.php
```diff
@@ -114,6 +114,139 @@ private function transform_a_tags( &$document ): void {
...
+	 *
+	 * Only relevant for older stories that haven't been updated in a while.
+	 *
+	 * @since 1.18.0
+	 *
+	 * @param Document|AMP_Document $document   Document instance.
+	 * @param bool                  $is_enabled Whether the feature is enabled.
+	 * @return void
+	 */
+	private function use_semantic_heading_tags( &$document, $is_enabled ): void {
```
spacedmonkey:
```suggestion
	private function use_semantic_heading_tags( &$document, bool $is_enabled ): void {
```

includes/AMP/Sanitization.php
```diff
@@ -61,16 +62,25 @@ class Sanitization {
...
 	 * @since 1.12.0
 	 *
-	 * @param Settings $settings Settings instance.
+	 * @param Settings    $settings    Settings instance.
+	 * @param Experiments $experiments Experiments instance.
 	 * @return void
 	 */
-	public function __construct( Settings $settings ) {
-		$this->settings = $settings;
+	public function __construct( Settings $settings, Experiments $experiments ) {
```
spacedmonkey, line 81: I am surprised this change it not mean updating or creating new tests.

swissspidy: We use the injector for tests. New dependencies are automatically added.

spacedmonkey (approved):

Issue #6690, opened 2021-03-09
Refactor Design Panel inputs and update logic


miina:
## Description

With the new Floating menu logic (#10112), many buttons will have a simple structure that's also more performant due to using specific selectors (e.g. button renders only if the value specific to that button changes, not if anything changes). The buttons, however, are quite often duplicated now since these are visually the same and also behave the same in both the floating menu and the design panel (except for multi-selection handling).

Ideally we would have a button "library" that we can just reuse anywhere so that the Design Panel would also have the new simplified logic.

This can be done in batches that can be defined once #10112 is ready.

The current design panel updating logic needs refactoring anyway, an example:
We are currently using `pushUpdate` with the second arg `true` for every input since otherwise the change is not applied at all. This seems to cause lagging as well, for example in case of text areas (Edit: the TextArea design system component now has its internal `onChange` handling which avoids lagging but is not ideal system either since the `onChange` that comes from the editor is ignored until `onBlur`). 
Additionally, if we need to trim values with `onBlur`, this currently would cause a separate history entry.

The `pushUpdate` logic should be reviewed/adjusted, if we need to use it with the second arg always `true`, it doesn't really fulfil its purpose.

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

littlemilkstudio:
referenced update logic here. Hopefully it helps out this ticket!

https://github.com/GoogleForCreators/web-stories-wp/pull/11389

littlemilkstudio:
outlined some perf exploration here. Hopefully can be used to help dictate some design decisions around the update logic rewrite as well:
#11390

littlemilkstudio:
outlined some perf exploration here. Hopefully can be used to help dictate some design decisions around the update logic rewrite as well:
#11390

PR #10581, opened 2022-02-15
Editor: Added functional flip toggles to floating menu

barklund:
## Context

This is the sixth (and final) PR for #10112. Future work on #10112 will after the completion of this PR be split into multiple tickets for each remaining part of the menus.

## Summary

This adds some general plumbing useful for future work on all elements.

## Relevant Technical Choices

* The `useProperties` hook can be this simple, because we are only going to use it for single element selection.
* The logic in `useFlip` is only extracted into an external file and not just made inline in the flip elements because it works in general for two different elements.

## Todo

* [ ] Add karma test

## User-facing changes

| Vertical flipped | Horizontal flipped | Screenshot |
|-|-|-|
| ❌  | ❌  | Screen Shot 2022-02-15 at 15 12 05 |
| ✅  | ❌  | Screen Shot 2022-02-15 at 15 12 09 |
| ✅  | ✅  | Screen Shot 2022-02-15 at 15 12 18 |
| ❌  | ✅  | Screen Shot 2022-02-15 at 15 12 22 |

## Testing Instructions

## Checklist

---

Partially addresses #10112


packages/story-editor/src/components/floatingMenu/elements/flipHorizontal.js
```diff
@@ -24,13 +24,16 @@ import { __ } from '@googleforcreators/i18n';
  * Internal dependencies
  */
 import { IconButton } from './shared';
+import useFlip from './shared/useFlip';
 
 function FlipHorizontal() {
+  const { horizontal, toggleHorizontal } = useFlip();
```
barklund: I really like how simple this component is. That's the goal for all of these menu elements!

packages/design-system/src/components/contextMenu/components/button.js
```diff
@@ -107,6 +109,14 @@ function ButtonWithRef(
   const autoGeneratedId = useMemo(uuidv4, []);
   const elementId = id || autoGeneratedId;
 
+  // Only supply a type if isToggled is boolean, otherwise leave type as undefined
+  const type =
+    typeof isToggled === 'boolean'
+      ? isToggled
+        ? BUTTON_TYPES.SECONDARY
+        : BUTTON_TYPES.TERTIARY
+      : undefined;
```
barklund: This is slightly copied logic from the `ToggleButton`, but it's not a direct copy, so I feel it's okay to include here. It could be generalized to somewhere, but on the other hand, it's so simple!

miina: cc @samwhale @BrittanyIRL for visibility.

packages/story-editor/src/components/floatingMenu/elements/flipHorizontal.js
```diff
@@ -24,13 +24,16 @@ import { __ } from '@googleforcreators/i18n';
  * Internal dependencies
  */
 import { IconButton } from './shared';
+import useFlip from './shared/useFlip';
 
 function FlipHorizontal() {
```
miina, line 29: Visually, these look exactly the same as the buttons in the design panel. And the functionality should be the same as well (if I understand correctly) -- would it make sense to unify these so that we could reuse the same buttons in both places? And for other upcoming buttons as well where applicable. Otherwise we'll have the same looks and logic implemented in two places. (I might not know some of the substantial differences here of course). Thoughts?

barklund: That's not a bad idea, but a lot of work at this point. Let's separate this into a new roadmap item, with the purpose of simplifying and optimizing the data flow in the design panel.

packages/design-system/src/components/contextMenu/components/button.js
barklund: Note that as far as I can tell, the menu buttons in the context menu use the button from the design-system, though never one of the specific types of buttons (primary, secondary, etc), but rather rely on the default common styles only.

packages/story-editor/src/components/floatingMenu/elements/flipHorizontal.js
miina: Based on the grooming discussion, updated this existing issue: https://github.com/GoogleForCreators/web-stories-wp/issues/6690

We can define it better once the menu bar is in place.

miina: ^ cc @littlemilkstudio for visibility since I know you also had some ideas for improving the Design Panel performance specifically, or were planning to work on it. We were discussing today in the grooming to follow up later (#6690) with using similar structure/logic for the design panel as Morten is doing here in this PR (reusing the same components as well) for not duplicating the buttons + improving performance.

packages/design-system/src/components/contextMenu/components/button.js
```diff
@@ -74,6 +73,8 @@ const StyledButton = styled(BaseButton)`
  * @param {boolean} props.dismissOnClick If to call onDismiss when clicking.
  * @param {Function} props.onFocus Focus event handler.
  * @param {boolean} props.forcePadding Force padding to button regardless of mode.
+ * @param {boolean} props.isToggled `true` if a toggled pushbutton, `false` if an untoggled
+ * pushbutton, leave unset if not a pushbutton.
```
merapi:
```suggestion
 * @param {boolean|null} props.isToggled `true` if a toggled pushbutton, `false` if an untoggled
 * pushbutton, leave unset if not a pushbutton.
```
`null` or `undefined`?

merapi: and `isToggled: PropTypes.bool,` in PropTypes (can't comment there).

merapi (approved):
Small nits and 🚢 🇮🇹

packages/design-system/src/components/contextMenu/components/button.js
barklund: Hm, an isn't an optional property just typed as the optional type or do we really need to do `|undefined` for each of them?

I will add the prop-type though!

barklund: Done in 8dbbc317c1.

miina (approved):

packages/design-system/src/components/contextMenu/components/button.js
merapi: To be JSDoc purists we should probably use:
image

For me, this particular case is a little special because if you leave it unset it will result in this 3rd state (while bool only has 2 states).

packages/design-system/src/components/contextMenu/components/button.js
BrookeGraham: This is a nit, but `isToggled` to me implies that the button is toggled "on" whereas `isToggle` might convey that the button is a toggle-style button.

packages/story-editor/src/components/floatingMenu/elements/flipHorizontal.js
BrittanyIRL: #6690 might be a good thing for us in pea pod to handle next release so that we can get some better shared context of these menu updates.

packages/story-editor/src/components/floatingMenu/elements/shared/useUpdateSelected.js
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Internal dependencies
+ */
+import { useStory } from '../../../../app';
+
+function useUpdateSelected() {
+  const { updateSelectedElements } = useStory(
+    ({ actions: { updateSelectedElements } }) => ({ updateSelectedElements })
+  );
```
BrittanyIRL: nit: destructuring this is redundant

```suggestion
  const updateSelectedElements = useStory(({ actions: { updateSelectedElements } }) => updateSelectedElements);
```

BrittanyIRL (commented):
when i click from one flip to the other so that both are toggled true, i expect the bg to be the same but the second one that becomes toggled true sticks on hover/focus bg color.


https://user-images.githubusercontent.com/10720454/154529223-5ef3a883-1ff7-4364-81e4-5d1e3d9d5ae1.mp4

packages/story-editor/src/components/floatingMenu/elements/flipHorizontal.js
maxyinger: @miina thanks for tagging. Yea, I haven't started looking into performance wins for the design panel yet, was kinda waiting for the dust to settle here before I started. Happy to audit any PRs for performance improvements if you'd like.

The only goal there is to not have the entire panel re-render every time any input updates. I'm pretty fine with whatever architectural decisions get us there. I know the logic with all the pushUpdate stuff is pretty complex so it might make it tough.

packages/story-editor/src/components/floatingMenu/elements/shared/useUpdateSelected.js
```diff
@@ -0,0 +1,37 @@
...
+ * External dependencies
+ */
+import { useCallback } from '@googleforcreators/react';
+
+/**
+ * Internal dependencies
+ */
+import { useStory } from '../../../../app';
+
+function useUpdateSelected() {
```
maxyinger: This hook seems a little redundant. Seems like an extra layer of indirection just to switch an action from named params to not. I feel like this also adds a little opaqueness to someone coming onto the codebase when we could just use the action directly. Thoughts?

packages/story-editor/src/components/floatingMenu/elements/shared/useUpdateSelected.js
maxyinger: Oh nice. didn't know `updateSelectedElements` existed as an action. now seeing that it's functionality embedded into the `updateElements` action. Was gonna create this action for perf in some areas, but nice to see it's already there to utilize.

packages/story-editor/src/components/floatingMenu/elements/shared/useFlip.js
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Internal dependencies
+ */
+import useProperties from './useProperties';
+import useUpdateSelected from './useUpdateSelected';
+
+function useFlip() {
+  const { flip } = useProperties(['flip']);
+  const updateSelected = useUpdateSelected();
```
maxyinger: as far as perf goes for this PR, this seems like the two key areas that would effect it and both `useUpdateSelected` & `useProperties` look pretty lean as far as using `useContextSelector` well. nice!

packages/design-system/src/components/contextMenu/components/button.js
samwhale: 💭 Honestly I don't think this should live in the design system's `Button`. The `Button` api is pretty flexible, and I don't think the responsibility of solving this specific case is the `Button`'s. I think that the parent component should decide and pass in the type.

Ex: In the future if a new toggle button were needed where the types were different - this would need to be changed since it spreads to all uses of the base `Button`.

Keeping this outside of the button would keep the api lightweight and flexible. I think the wrapper would look something like:

```js
function ToggleButton({ isToggled, ...props }) {
    
    const type =
      typeof isToggled === 'boolean'
        ? isToggled
          ? BUTTON_TYPES.SECONDARY
          : BUTTON_TYPES.TERTIARY
        : undefined;

    return (
        <Button {...props} type={type} />
    )
}
```

---

However, I know that this is practically done and the suggestion above isn't necessary to make this work. So no pressure either way.

samwhale (approved):

packages/story-editor/src/components/floatingMenu/elements/shared/useUpdateSelected.js
barklund: It was just to simplify the usage of the action in a lot of components that are going to use this and make things more DRY. So rather than doing all of this in every component:

```js
  const { updateSelectedElements } = useStory(
    ({ actions: { updateSelectedElements } }) => ({ updateSelectedElements })
  );
  const onClick = () => {
    ...
    updateSelectedElements({ properties });
  }
```

It would be just this:

```js
  const updateSelectedElements = useUpdateSelected();
  const onClick = () => {
    ...
    updateSelectedElements(properties);
  }
```

The difference between named arguments and positional was actually not the primary purpose. It was to DRY up the whole selector for the function, which is pretty lengthy.

I would gladly remove the `useCallback` part of it so it's just a shorthand for getting the action from the context.

barklund:
> when i click from one flip to the other so that both are toggled true, i expect the bg to be the same but the second one that becomes toggled true sticks on hover/focus bg color.

That's a general UX problem we have with toggle buttons. The same thing happens in the design panel for these exact same buttons. I did notice it, but I'm not going to solve this here.

We should seek to have a different UX to display focus on toggle buttons rather than just a background change, as it's almost impossible to discern the difference between focus from toggled in this fashion.

packages/story-editor/src/components/floatingMenu/elements/shared/useUpdateSelected.js
barklund: Also with the default identify function being `shallowEqual`? I hadn't actually tried this, but just kind of assumed that we could only select an object of things, not a single thing, while still having memoization work correctly.

barklund: But if that works, this hook is definitely pretty useless, as it was mainly to get around this slightly annoying restriction.

Also, we should do that throughout the codebase - I think we have potentially hundreds of these spread throughout the codebase.

packages/design-system/src/components/contextMenu/components/button.js
barklund: > This is a nit, but `isToggled` to me implies that the button is toggled "on" whereas `isToggle` might convey that the button is a toggle-style button.

@BrookeGraham, that's also the intent of the code. If isToggled is set to true or false, it is a toggle button and it's toggled based on the truthiness. If isToggled is not set at all, it is not a toggle button at all. It's my way of decoding both the fact that this is a toggle button and that this toggle button is respectively toggled on or off into a single prop.

@samwhale, that's a good point. I moved it completely out of the component now and into a local button in the floating menu codebase in 3c9de899da.

packages/story-editor/src/components/floatingMenu/elements/shared/useUpdateSelected.js
barklund: Seems to work great. Optimized this in fbdaaec8ba .

packages/story-editor/src/components/floatingMenu/elements/shared/useUpdateSelected.js
barklund: Optimized this in fbdaaec8ba  (and completely removed this file).

packages/design-system/src/components/contextMenu/components/button.js
barklund: @samwhale, just re-read your comment. This code didn't actually live in the design-system's `Button` but in the design-system's context menu's `Button` so it was a bit more specific than that. Regardless, I have moved it out of the design-system entirely, as having toggl-able context menu buttons (for now) is specific to the floating design menu.

samwhale: Ah, I missed that! but great, I like the change 😄

PR #10582, opened 2022-02-15
Status Panel: make options a Dropdown

### LINKED ISSUE #10579: Visibility Panel: Redesign
In the latest editor redesign, the visibility panel is getting updated from the legacy 3 radio buttons into a dropdown. This ticket is to handle that update in this panel. **Functionality on select will remain the same as this is required by WordPress.** 

Figma lives here: 
- [story details modal](https://www.figma.com/file/ZrO763yvMWJjYeAPsRnNfW/_Editor---Sprint-Feature-Level-Designs_2022?node-id=57%3A14792)

## Acceptance Criteria 
- Update e2e tests accordingly 
- Regression testing on panel visibility as used in both editor and story details publish modal 
- keyboard accessibility

BrookeGraham:
## Context

In the latest editor redesign, the visibility panel is getting updated from the legacy 3 radio buttons into a dropdown. This ticket is to handle that update in this panel. Functionality on select will remain the same as this is required by WordPress.

## Summary
Updating the Status Panel to show visibility options in a dropdown. 

## Relevant Technical Choices
Added a `visibility` to story state. This is so that we can keep `Document` pane and `Publish` modal in sync since the `Status` panel is being used in both places. 

## To-do
- e2e test

## User-facing changes

|Before|After|
|--|--|
|Screen Shot 2022-02-15 at 4 16 44 PM|Screen Shot 2022-02-17 at 11 42 21 AM|
|Screen Shot 2022-02-17 at 11 17 32 AM|Screen Shot 2022-02-17 at 11 47 00 AM

### New Modal
|Public|Password|
|--|--|
|Screen Shot 2022-02-16 at 2 46 37 PM|Screen Shot 2022-02-17 at 11 49 15 AM|

## Testing Instructions

1. Regression on Status (Visibility) Panel. It should all work the same as staging, we are just showing in a dropdown format now. 

no

no

no

## Checklist

---

Fixes #10579

### DIFF
--- a/packages/design-system/src/components/dropDown/index.js
+++ b/packages/design-system/src/components/dropDown/index.js
@@ -31,7 +31,6 @@ import { __, sprintf } from '@googleforcreators/i18n';
 /**
  * Internal dependencies
  */
-import { THEME_CONSTANTS } from '../../theme';
 import { Menu, MENU_OPTIONS } from '../menu';
 import { Popup, PLACEMENT } from '../popup';
 import { DropDownContainer, Hint } from './components';
@@ -200,14 +199,7 @@ export const DropDown = forwardRef(
             {menu}
           </Popup>
         )}
-        {hint && (
-          <Hint
-            hasError={hasError}
-            size={THEME_CONSTANTS.TYPOGRAPHY.TEXT_SIZES.SMALL}
-          >
-            {hint}
-          </Hint>
-        )}
+        {hint && <Hint hasError={hasError}>{hint}</Hint>}
       </DropDownContainer>
     );
   }
--- a/packages/wp-story-editor/src/components/documentPane/index.js
+++ b/packages/wp-story-editor/src/components/documentPane/index.js
@@ -23,6 +23,7 @@ import {
   BackgroundAudioPanel,
   TaxonomiesPanel,
 } from '@googleforcreators/story-editor';
+import styled from 'styled-components';
 
 /**
  * Internal dependencies
@@ -60,9 +61,14 @@ export function PublishModalDocumentPane() {
 // Isolated Status Panel should not collapse and
 // should have its own name to prevent collapse
 // based on other implementations that have default name
+
+const IsolatedPanel = styled(StatusPanel)`
+  padding: 0;
+`;
+
 export function IsolatedStatusPanel() {
   return (
-    <StatusPanel
+    <IsolatedPanel
       nameOverride="storyDetailsStatus"
       canCollapse={false}
       isPersistable={false}
--- a/packages/wp-story-editor/src/components/documentPane/index.js
+++ b/packages/wp-story-editor/src/components/documentPane/index.js
@@ -66,6 +66,7 @@ export function IsolatedStatusPanel() {
       nameOverride="storyDetailsStatus"
       canCollapse={false}
       isPersistable={false}
+      popupZIndex={11}
     />
   );
 }
--- a/packages/wp-story-editor/src/components/documentPane/status/status.js
+++ b/packages/wp-story-editor/src/components/documentPane/status/status.js
@@ -32,6 +32,11 @@ import {
   useIsUploadingToStory,
 } from '@googleforcreators/story-editor';
 
+/**
+ * Internal dependencies
+ */
+import { VISIBILITY, STATUS } from '../../../constants';
+
 function StatusPanel({
   nameOverride,
   popupZIndex,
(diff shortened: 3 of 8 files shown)


packages/wp-story-editor/src/components/documentPane/status/status.js
```diff
@@ -38,7 +37,7 @@ const InputRow = styled(Row)`
   margin-left: 34px;
 `;
 
-function StatusPanel({ nameOverride }) {
+function StatusPanel({ nameOverride, popupZIndex }) {
```
BrookeGraham: I decided to pass `popupZIndex` as a prop so that when we render `StatusPanel` in the sidebar Document Pane, it the dropdown doesn't have to have the same zIndex as is needed when it is rendered in the modal. If we had let them be the same (ie 11 is need on the modal) then tooltips would get placed behind the dropdown.

BrittanyIRL (commented):
Thanks for jumping on this so quickly. Functionality is working great! 

One small thing - let's line the text input up with the dropdown, I know that you kept it how it's been but I think lining it up is more in line with the new designs 

Screen Shot 2022-02-15 at 9 40 53 PM
Screen Shot 2022-02-15 at 9 42 32 PM

BrookeGraham:
> One small thing - let's line the text input up with the dropdown

Totally. I also realized the toggle was still appearing. So I fixed that just haven’t pushed it up yet.

BrittanyIRL:
> Totally. I also realized the toggle was still appearing. So I fixed that just haven’t pushed it up yet.

oh thank you! that was on my list of things to do today, that was my bad when i renamed the `name` to `overrideName` (or whatever it was) that those 2 props weren't getting considered from the isolated status panel that i added.

packages/wp-story-editor/src/components/documentPane/status/status.js
```diff
@@ -78,19 +80,25 @@ function StatusPanel({ nameOverride }) {
...
+      });
+      return;
     }
 
-    return 'public';
-  }, [status, hasPassword]);
+    updateStory({
+      properties: { visibility: 'public' },
+    });
+  }, [password, status, updateStory]);
```
BrookeGraham: Hi @swissspidy From what I gathered, we were setting `visibility` based on the story's status and whether or not a password was present. Basically, if a status was anything but `private` then it's `visibility` would be set to `public` _unless_ there was a `password` present. Since we now will be using this component in multiple places, I have added `visibility` to the story's state in order to keep the two instances of the component in sync. I am still setting `visibility` based on the presence of a `password`, so I wanted to make sure I was understanding that bit of logic correctly.

packages/wp-story-editor/src/components/documentPane/index.js
```diff
@@ -60,12 +61,18 @@ export function PublishModalDocumentPane() {
 // Isolated Status Panel should not collapse and
 // should have its own name to prevent collapse
 // based on other implementations that have default name
+
+const IsolatedPanel = styled(StatusPanel)`
+  padding: 0;
```
BrookeGraham, line 66: This is to remove the `SimplePanel`'s padding.

packages/wp-story-editor/src/components/documentPane/status/status.js
```diff
@@ -206,27 +214,34 @@ function StatusPanel({ nameOverride }) {
       <SimplePanel
         name={nameOverride || 'status'}
         title={__('Visibility', 'web-stories')}
+        canCollapse={canCollapse}
+        isPersistable={isPersistable}
```
BrittanyIRL, line 232: thank you for adding these!

BrittanyIRL (approved):
dig this update, thanks for doing is so thoughtfully. It's going to make it so much easier to check for a password when we need to since the new story detail has these fields required before actually publishing. 

what do you think about storing 'private', 'protected' and 'public' as constants to avoid matching typed strings through out `document/status/status`?

BrittanyIRL (commented):
looks like one unit test needs updating 

Screen Shot 2022-02-17 at 3 25 36 PM

packages/wp-story-editor/src/components/documentPane/status/test/status.js
```diff
@@ -86,42 +95,71 @@ describe('statusPanel', () => {
...
   it('should always render the "Public" visibility option', () => {
     arrange({
       publish: false,
     });
-    expect(screen.getByLabelText('Public')).toBeInTheDocument();
+    clickDropdown();
+
+    expect(
+      screen.getByRole('option', {
+        name: 'Selected Public',
```
BrittanyIRL: Just `Public` is what's getting looked for. 
Screen Shot 2022-02-17 at 3 25 36 PM

packages/design-system/src/components/dropDown/index.js
```diff
@@ -200,14 +199,7 @@ export const DropDown = forwardRef(
             {menu}
           </Popup>
         )}
-        {hint && (
-          <Hint
-            hasError={hasError}
-            size={THEME_CONSTANTS.TYPOGRAPHY.TEXT_SIZES.SMALL}
```
BrookeGraham, line 206: We want `PRESET_SIZES` not `TEXT_SIZES`

packages/wp-story-editor/src/components/documentPane/status/test/status.js
BrookeGraham: I am a ding dong.

BrookeGraham: There are no options when it is disabled  🤦🏻‍♀️

BrookeGraham:
>what do you think about storing 'private', 'protected' and 'public' as constants to avoid matching typed strings through out document/status/status?

Yeah. Let's do it.

packages/wp-story-editor/src/components/documentPane/status/status.js
swissspidy: Thanks for flagging! Yep, that's how it basically works in WP. So from what I can see this looks good 👍

PR #10589, opened 2022-02-16
Code Quality: Check if global constants are defined before using

mohdsayed:
## Context

Global node constants added in https://github.com/GoogleForCreators/web-stories-wp/pull/10487 are not defined when working with a standalone package

## Testing Instructions

1.

## Checklist

---

Fixes #

### DIFF
--- a/packages/story-editor/src/app/helpCenter/useHelpCenter/effects.js
+++ b/packages/story-editor/src/app/helpCenter/useHelpCenter/effects.js
@@ -94,7 +94,10 @@ export function deriveUnreadTipsCount(previous, next) {
 }
 
 export function deriveAutoOpen(previous, next) {
-  if (WEB_STORIES_DISABLE_QUICK_TIPS === 'true') {
+  if (
+    typeof WEB_STORIES_DISABLE_QUICK_TIPS !== 'undefined' &&
+    WEB_STORIES_DISABLE_QUICK_TIPS === 'true'
+  ) {
     return {};
   }
   if (isInitialHydrate(previous, next)) {
@@ -107,7 +110,10 @@ export function deriveAutoOpen(previous, next) {
 // If there are any unread tips, we respect users last open setting.
 // If all tips are read, we want the popup closed regardless of user setting.
 export function deriveInitialOpen(persisted) {
-  if (WEB_STORIES_DISABLE_QUICK_TIPS === 'true') {
+  if (
+    typeof WEB_STORIES_DISABLE_QUICK_TIPS !== 'undefined' &&
+    WEB_STORIES_DISABLE_QUICK_TIPS === 'true'
+  ) {
     return {};
   }
   const hasUnreadTips = Boolean(persisted?.unreadTipsCount);
--- a/packages/story-editor/src/components/library/panes/text/textSets/textSet.js
+++ b/packages/story-editor/src/components/library/panes/text/textSets/textSet.js
@@ -113,7 +113,9 @@ function TextSet({ id, elements, translateY, translateX, ...rest }, ref) {
     [onClick]
   );
 
-  const renderImages = WEB_STORIES_DISABLE_OPTIMIZED_RENDERING !== 'true';
+  const renderImages =
+    typeof WEB_STORIES_DISABLE_OPTIMIZED_RENDERING !== 'undefined' &&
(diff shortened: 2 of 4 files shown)


swissspidy (approved):

swissspidy:
I definitely recall discussing this, but not sure where that was. But definitely important to fix this, so thanks!

miina:
@swissspidy / @sayedtaqui Looks like this Karma test has been failing since this PR: `should add text color based on background`.
Was there a reason for merging this with failing tests, e.g. did you see it failing already before?

(Not 100% sure this PR caused it but that's the first that fails with this on `main`)

miina:
Update: reverted the changes in the PR and the test doesn't fail locally anymore.

miina:
It also seems to cause some other failures, e.g. `Text Panel should have no aXe violations` in `libraryTabs.karma.js`.

Could we revert the PR and recreate it so that it doesn't cause Karma failures?

packages/story-editor/src/components/library/panes/text/textSets/textSet.js
```diff
@@ -113,7 +113,9 @@ function TextSet({ id, elements, translateY, translateX, ...rest }, ref) {
     [onClick]
   );
 
-  const renderImages = WEB_STORIES_DISABLE_OPTIMIZED_RENDERING !== 'true';
+  const renderImages =
+    typeof WEB_STORIES_DISABLE_OPTIMIZED_RENDERING !== 'undefined' &&
```
miina, line 117: Since the default should be `false`, shouldn't it be ok here to check for `WEB_STORIES_DISABLE_OPTIMIZED_RENDERING !== 'true';` only since it also behaves correctly if not set?

Edit: meaning that `renderImages` should be `true` even if `WEB_STORIES_DISABLE_OPTIMIZED_RENDERING` is `undefined`.

packages/story-editor/src/app/helpCenter/useHelpCenter/effects.js
```diff
@@ -94,7 +94,10 @@ export function deriveUnreadTipsCount(previous, next) {
 }
 
 export function deriveAutoOpen(previous, next) {
-  if (WEB_STORIES_DISABLE_QUICK_TIPS === 'true') {
+  if (
+    typeof WEB_STORIES_DISABLE_QUICK_TIPS !== 'undefined' &&
+    WEB_STORIES_DISABLE_QUICK_TIPS === 'true'
```
miina, line 99: Doesn't `WEB_STORIES_DISABLE_QUICK_TIPS === 'true'` already mean that `typeof WEB_STORIES_DISABLE_QUICK_TIPS !== 'undefined'`? 🤔  Or would it give an error otherwise?

packages/story-editor/src/components/library/panes/text/textSets/textSet.js
swissspidy: You mean `const renderImages =
    typeof WEB_STORIES_DISABLE_OPTIMIZED_RENDERING === 'undefined' || ! WEB_STORIES_DISABLE_OPTIMIZED_RENDERING`?

packages/story-editor/src/app/helpCenter/useHelpCenter/effects.js
swissspidy: There would  be an error otherwise. Try running `WEB_STORIES_DISABLE_QUICK_TIPS === 'true'` in your dev tools

swissspidy:
@miina haven't spotted them. Sounds like some checks need to be adjusted. Should be an easy fix in a new PR.

packages/story-editor/src/components/library/panes/text/textSets/textSet.js
miina: Or changing the constant name (and usage):
`const omitRenderingImages = typeof WEB_STORIES_DISABLE_OPTIMIZED_RENDERING !== 'undefined' && WEB_STORIES_DISABLE_OPTIMIZED_RENDERING === 'true';`

Or
`const renderImages = typeof WEB_STORIES_DISABLE_OPTIMIZED_RENDERING === 'undefined' || WEB_STORIES_DISABLE_OPTIMIZED_RENDERING !== 'true';`

That would perhaps fix the test failures as well since these seem to be all related to the text panel.

PR #10590, opened 2022-02-16
Telemetry: track the WordPress environment

### LINKED ISSUE #10551: Telemetry: track the WordPress environment
## Feature Description

In WordPress 5.5 the function [wp_get_environment_type](https://developer.wordpress.org/reference/functions/wp_get_environment_type/) was added. This allows hosting providers and devs to provide a way to define the environment, like production or dev. This data may be especially useful for data submitting the analytics data that is tracked. There could even be argument to say that analytics data should not be collected on dev environments as these are by their design, not as stable as production. This data may also be useful in filtering out errors, by other showing that happens on production ( public ) servers.

spacedmonkey:
## Summary

Add a new user property called `serverEnvoriment` that is submitted to the analytics. 

## User-facing changes

Screenshot 2022-02-16 at 12 17 32

## Testing Instructions

1.

## Checklist

---

Fixes #10551

### DIFF
--- a/includes/Tracking.php
+++ b/includes/Tracking.php
@@ -196,6 +196,7 @@ class Tracking extends Service_Base {
 			'wpVersion'          => get_bloginfo( 'version' ),
 			'phpVersion'         => PHP_VERSION,
 			'isMultisite'        => (int) is_multisite(),
+			'serverEnvironment'  => wp_get_environment_type(),
 			'adNetwork'          => $ad_network,
 			'analytics'          => $analytics,
 			'activePlugins'      => $active_plugins,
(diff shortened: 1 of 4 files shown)


includes/Tracking.php
```diff
@@ -196,6 +196,7 @@ private function get_user_properties(): array {
 			'wpVersion'          => get_bloginfo( 'version' ),
 			'phpVersion'         => PHP_VERSION,
 			'isMultisite'        => (int) is_multisite(),
+			'serverEnvironment'  => wp_get_environment_type(),
```
swissspidy, line 199: We also need to set this in `packages/tracking/src/enableTracking.js`

```
dimension20: 'serverEnvironment',
```

spacedmonkey: Fixed https://github.com/GoogleForCreators/web-stories-wp/pull/10590/commits/4fdf8036a7bf51ac364029a06373564086687632

swissspidy (approved):

PR #10595, opened 2022-02-16
New Story Details Modal: Load Users

BrittanyIRL:
## Context

## Summary

Accidentally introduced a new bug on the new story details modal (so this is a preemptive fix since that's not in use yet). Missed where the users are being fetched from `wp-story-editor` and threaded through to assigning author in publish panel. 

## Relevant Technical Choices

Limiting request to load users based on users array being empty and isUsersLoading being false instead of the document tab. This means that users are only getting fetched once regardless of tab changes.

Also silencing 2 console warnings that were bumming me out while testing publish panel in the modal. 

## User-facing changes

Now users will populate for the story details modal regardless of if user has navigated to the document tab in the editor first. 

| Broken | Fixed | 
| --- | --- |
| https://user-images.githubusercontent.com/10720454/154348528-c61102a0-2cde-4115-bb92-2ad72edee859.mp4 | https://user-images.githubusercontent.com/10720454/154348580-1e0d90ed-338c-4515-92ff-e5a5469ef539.mp4 |

## Testing Instructions

Verify that users are getting fetched one time and both author dropdowns get populated.

1.

## Checklist

---

Partially addresses #10115

### DIFF
--- a/packages/story-editor/src/components/form/dateTime/datePicker.js
+++ b/packages/story-editor/src/components/form/dateTime/datePicker.js
@@ -163,7 +163,7 @@ function DatePicker({ currentDate, onChange, onViewChange }) {
 DatePicker.propTypes = {
   onChange: PropTypes.func.isRequired,
   onViewChange: PropTypes.func,
-  currentDate: PropTypes.string,
+  currentDate: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
 };
 
 export default DatePicker;
--- a/packages/wp-story-editor/src/components/documentPane/publish/author.js
+++ b/packages/wp-story-editor/src/components/documentPane/publish/author.js
@@ -55,10 +55,11 @@ function Author() {
   const [visibleOptions, setVisibleOptions] = useState(null);
 
   useEffect(() => {
-    if (tab === 'document') {
-      loadUsers();
+    if (users.length === 0 && !isUsersLoading) {
+      return loadUsers();
     }
-  }, [tab, loadUsers]);
+    return () => {};
+  }, [loadUsers, users, isUsersLoading]);
 
   const getAuthorsBySearch = useCallback(
     (search) => {
--- a/packages/wp-story-editor/src/components/documentPane/publish/author.js
+++ b/packages/wp-story-editor/src/components/documentPane/publish/author.js
@@ -32,7 +32,7 @@ function Author() {
     actions: { getAuthors },
   } = useAPI();
   const {
-    state: { tab, users, isUsersLoading },
+    state: { users, isUsersLoading },
     actions: { loadUsers },
   } = useInspector();
   const { isSaving, author, updateStory } = useStory(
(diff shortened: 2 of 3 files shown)


packages/wp-story-editor/src/components/documentPane/publish/author.js
```diff
@@ -55,10 +55,11 @@ function Author() {
   const [visibleOptions, setVisibleOptions] = useState(null);
 
   useEffect(() => {
-    if (tab === 'document') {
-      loadUsers();
+    if (users.length === 0 && !isUsersLoading) {
```
miina: `loadUsers` already includes that check as well: https://github.com/GoogleForCreators/web-stories-wp/blob/d530f151e0afc617b2ce9097d48dd1f5c77411f5/packages/story-editor/src/components/inspector/inspectorProvider.js#L111-L114

packages/story-editor/src/components/form/dateTime/datePicker.js
```diff
@@ -163,7 +163,7 @@ function DatePicker({ currentDate, onChange, onViewChange }) {
 DatePicker.propTypes = {
   onChange: PropTypes.func.isRequired,
   onViewChange: PropTypes.func,
-  currentDate: PropTypes.string,
+  currentDate: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
miina, line 166: Just curious -- in which case is it a number?

BrittanyIRL: the first time you go to the document pane in a new story. `floatingDate` in `PublishTime` is true so the display date is used, which is a number (`{floatingDate: true, displayDate: 1645112356708, date: '2022-02-17T15:39:12'}` - whereas after the story's been saved in any way the floatingDate is false, it seems to be consistent, so the `date` string is used instead. Since the date capture is working as expected I didn't want to tamper with what value is getting used since both are valid to create a new Date.

packages/wp-story-editor/src/components/documentPane/publish/author.js
BrittanyIRL: this is good to know, thank you! the check i added is redundant so we can just call `getUsers` here and let that decide when to actually get users.

BrookeGraham (approved):
Nice

maxyinger (approved):

packages/wp-story-editor/src/components/documentPane/publish/author.js
```diff
@@ -55,10 +55,8 @@ function Author() {
   const [visibleOptions, setVisibleOptions] = useState(null);
 
   useEffect(() => {
-    if (tab === 'document') {
-      loadUsers();
-    }
-  }, [tab, loadUsers]);
+    loadUsers();
+  }, [loadUsers]);
```
samwhale, line 59: 👍 verified this only runs when the document panel is open _or_ the publish modal is open

samwhale (approved):

PR #10585, opened 2022-02-16
New Stories Detail Modal: Cover Preview UI

BrittanyIRL:
## Context

Part of the story details modal feature. 

## Summary

Gives the user a preview of what their story will show up as in Discover and Search results as part of the new story details modal that gets viewed when 'publish' is clicked in the editor. 

**Example of previews as seen in Chrome Search results**
Screen Shot 2022-02-15 at 3 27 15 PM

## Relevant Technical Choices

Mostly this is just a lot of css.
- Thought about sharing some styles with the dashboard since we're kind of copying the cover styles as seen in Search, but there's some extra info there that we don't need here that isn't consistent with the actual cover previews (who the author is, is it a draft, is it locked). So I opted for just putting all the styles in one place. 
- Set the base structure up for editing the featured media from the new 'edit' button that's overlayed on the cover preview in figma so that I knew I got all the variants right for styling it. Hooking up the functionality is more involved. Discussed possibly moving this to v2 of the modal with Andrew who was ok with it. Will circle back after everything else here is done. Already going to cut it close. 
  - Checking to see if user has `hasUploadMediaAction` capabilities before rendering 'edit' button, that functionality I'll do separate since it requires sharing code that's already in use, new ticket for that lives here: https://github.com/GoogleForCreators/web-stories-wp/issues/10584

## To-do

There's this, that I mentioned above - to add the [ability to edit the featured image](https://github.com/GoogleForCreators/web-stories-wp/issues/10584) from the cover preview

I found a few loose ends that I will take care of separately that are noticeable here:
- available users to set as author doesn't get fetched until the 'document' panel has been clicked on in the editor 
- discrepancy between onChange vs onBlur with story description in modal vs in panel - change that and title to also be on blur implementations 
- Checklist needs to be an icon w/ tooltip just like in the footer 
- There's a 1px discrepancy on the edge of the modal that has the document panels  

## User-facing changes

N/A All behind a feature flag 

## Testing Instructions

Storybook is helpful here, added some new knobs (thanks for that update, @BrookeGraham !!) 
- You can see different variants of the cover preview depending on what all the user's provided (featured media, publisher logo, site title, story title) 

In the editor, enable the new story details modal experiment and then hit 'publish'
- should see cover preview
- look at it with and without a featured image, publisher logo, site title, story title 
- repeat in RTL

| LTR | RTL |
|--- | --- |
| Screen Shot 2022-02-15 at 5 15 51 PM | Screen Shot 2022-02-15 at 5 15 37 PM |

[...]

PR #10598, opened 2022-02-16
New Story Details Modal: Checklist Button Swap to Icon 

BrittanyIRL:
## Context

## Summary

Updates the button style that opens checklist from detail story modal. It should be an icon with a tooltip, not a button that says "checklist"

## Relevant Technical Choices

Use `ToggleButton` from footer for checklist button in modal, works the same just doesn't "toggle", the UI is what we want to match. Pass in optional popupZIndexOverride to account for our popup in a popup situation like other tooltips in modal.
- Move `ChecklistCountProvider` up out of footer and into canvas `NavLayer` - that is the nearest component to both header and footer - so that i can access checklist count in modal. 
- Add one new value to state in checklist count context called `hasChecklist`, I need to know if the checklist even exists in the header and i don't want to pass through the raw component as that is overkill - so by checking if the component is getting threaded through from `wp-story-editor` in `checklistCountProvider` I can call the new hook I wrote called `useHasChecklist` to grab a boolean that tells me if i should bother grabbing the checklist count and rendering the button in the modal. Will add e2e test separately.

## User-facing changes

Detail story modal now has a checklist button that shows the checklist icon and a tooltip (see video) AND it is tracking the priority issues present in the story just like the checklist's toggle button. Figma doesn't show that, this was a change in design directive today verified in slack - I think it's good to have the continuity. 

https://user-images.githubusercontent.com/10720454/154359177-e379cf60-c406-478b-ac20-a8ff32b28509.mp4

Screen Shot 2022-02-16 at 4 51 21 PM

## Testing Instructions

Verify that the checklist button in the story details modal (experiment turned out) looks like the above. Clicking it closes the modal and opens the checklist.

1.

## Checklist

---

Partially addresses #10115

### DIFF
--- a/packages/story-editor/src/components/canvas/navLayer.js
+++ b/packages/story-editor/src/components/canvas/navLayer.js
@@ -27,24 +27,29 @@ import Footer from '../footer';
 import DirectionAware from '../directionAware';
 import PageSideMenu from './pageSideMenu';
 import { FooterArea, HeadArea, Layer, PageMenuArea, Z_INDEX } from './layout';
+import { ChecklistCountProvider } from '../checklist';
 
 function NavLayer({ header, footer }) {
   return (
-    <Layer
-      pointerEvents="none"
-      zIndex={Z_INDEX.NAV}
-      onMouseDown={(evt) => evt.stopPropagation()}
+    <ChecklistCountProvider
+      hasChecklist={Boolean(footer?.secondaryMenu?.checklist)}
     >
-      <HeadArea pointerEvents="initial">{header}</HeadArea>
-      <DirectionAware>
-        <PageMenuArea>
-          <PageSideMenu />
-        </PageMenuArea>
-      </DirectionAware>
-      <FooterArea pointerEvents="initial">
-        <Footer footer={footer} />
-      </FooterArea>
-    </Layer>
+      <Layer
+        pointerEvents="none"
+        zIndex={Z_INDEX.NAV}
+        onMouseDown={(evt) => evt.stopPropagation()}
+      >
+        <HeadArea pointerEvents="initial">{header}</HeadArea>
+        <DirectionAware>
+          <PageMenuArea>
+            <PageSideMenu />
+          </PageMenuArea>
+        </DirectionAware>
+        <FooterArea pointerEvents="initial">
+          <Footer footer={footer} />
+        </FooterArea>
+      </Layer>
+    </ChecklistCountProvider>
   );
 }
 
--- a/packages/story-editor/src/components/canvas/navLayer.js
+++ b/packages/story-editor/src/components/canvas/navLayer.js
@@ -23,11 +23,11 @@ import Proptypes from 'prop-types';
 /**
  * Internal dependencies
  */
+import { ChecklistCountProvider } from '../checklist';
 import Footer from '../footer';
 import DirectionAware from '../directionAware';
 import PageSideMenu from './pageSideMenu';
 import { FooterArea, HeadArea, Layer, PageMenuArea, Z_INDEX } from './layout';
-import { ChecklistCountProvider } from '../checklist';
 
 function NavLayer({ header, footer }) {
   return (
--- a/packages/story-editor/src/components/publishModal/mainContent/checklistButton.js
+++ b/packages/story-editor/src/components/publishModal/mainContent/checklistButton.js
@@ -22,19 +22,12 @@ import { __ } from '@googleforcreators/i18n';
  * Internal dependencies
  */
 import { ISSUE_TYPES } from '../../checklist/constants';
-import {
-  useCategoryCount,
-  useHasChecklist,
-} from '../../checklist/countContext';
+import { useCategoryCount } from '../../checklist/countContext';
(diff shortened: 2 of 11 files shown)


packages/story-editor/src/components/canvas/navLayer.js
```diff
@@ -27,24 +27,29 @@ import Footer from '../footer';
...
+import { ChecklistCountProvider } from '../checklist';
 
 function NavLayer({ header, footer }) {
   return (
-    <Layer
-      pointerEvents="none"
-      zIndex={Z_INDEX.NAV}
-      onMouseDown={(evt) => evt.stopPropagation()}
+    <ChecklistCountProvider
+      hasChecklist={Boolean(footer?.secondaryMenu?.checklist)}
```
BrittanyIRL, line 35: Open to suggestions if ya'll have a better/more straightforward way to communicate if the checklist even exists to the story details modal.

packages/story-editor/src/components/publishModal/stories/index.js
```diff
@@ -96,7 +97,9 @@ export const _default = (args) => {
             },
           }}
         >
-          <PublishModal {...args} />
+          <ChecklistCountProvider hasChecklist={args.hasChecklist}>
```
BrittanyIRL, line 100: args is getting destructured in #10585 so this is mostly to prevent a conflict, i'll add it to the destructure in whatever PR is merged last between these two

packages/story-editor/src/components/canvas/navLayer.js
maxyinger: Looks like this is how we're checking whether or not to render the other checklist button too, so I think this is what we got to work with 🤷

packages/story-editor/src/components/publishModal/mainContent/checklistButton.js
```diff
@@ -0,0 +1,50 @@
...
+} from '../../checklist/countContext';
+import { ToggleButton } from '../../toggleButton';
+
+const ChecklistButton = ({ handleReviewChecklist }) => {
+  const hasChecklist = useHasChecklist();
+  if (!hasChecklist) {
+    return null;
+  }
+
+  const priorityCount = useCategoryCount(ISSUE_TYPES.PRIORITY);
```
BrookeGraham, line 30: move this above early return to avoid breaking hook rules.

maxyinger (approved):
lgtm

BrookeGraham (approved):
One small comment. Looks great.

packages/story-editor/src/components/publishModal/mainContent/checklistButton.js
BrittanyIRL: whoops, thank you!

packages/story-editor/src/components/canvas/navLayer.js
samwhale: Looks like the count provider can be moved above the checklist provider since it doesn't depend on the checklist provider

💭 Is there a way to derive it from the `value` in the `ChecklistCountProvider`?

```js
// packages/story-editor/src/components/checklist/countContext/checkCountContext.js

function ChecklistCountProvider({ children }) {
  const value = useState(INITIAL_STATE);

  // Grab the issues and see if any of the issue types have any issues registered
  const hasChecklist = Object.values(value).some((issues) => issues.length);

  return (
    <CountContext.Provider value={value}>{children}</CountContext.Provider>
  );
}
ChecklistCountProvider.propTypes = {
  children: PropTypes.node,
};
```

But I faintly remember that some issues don't show until certain checkpoints have been passed 🤔. Maybe that would make the above approach not work?

packages/story-editor/src/components/publishModal/mainContent/checklistButton.js
```diff
@@ -0,0 +1,50 @@
...
+import {
+  useCategoryCount,
+  useHasChecklist,
+} from '../../checklist/countContext';
+import { ToggleButton } from '../../toggleButton';
+
+const ChecklistButton = ({ handleReviewChecklist }) => {
+  const hasChecklist = useHasChecklist();
+  if (!hasChecklist) {
+    return null;
```
samwhale: 💭 Could use the hook inside `MainContent` and conditionally render the button there? What do you think about having the hook there vs here?

samwhale (commented):
A few prop validation and hook rule errors but once those are fixed then I think this works well.

Makes sense that we need to lift the count provider up so that we can get it in the publish modal too

packages/story-editor/src/components/publishModal/mainContent/checklistButton.js
BrittanyIRL: yes, i actually just pushed that up after brooke pointed out my linting issue. thanks!

packages/story-editor/src/components/canvas/navLayer.js
BrittanyIRL: yeah, this gets weird because we need to know ahead of issues existing because the checklist is conditionally rendered based on if `footer.secondaryMenu.checklist` in `FooterLayout`.

PR #10600, opened 2022-02-17
Text Sets: fix global constant check

swissspidy:
Follow-up to #10589.


miina (approved):
Tests seem to be passing now 🎉

mohdsayed (approved):

PR #10603, opened 2022-02-17
Links: Add special handling for Instagram when retrieving metadata

### LINKED ISSUE #10451: Links: Add special handling for Instagram when retrieving metadata
## Feature Description

As requested by the user in [this support topic](https://wordpress.org/support/topic/about-link-information-from-another-site-in-editor/#post-15324280), could we add some special handling for Instagram links to fix the text link optional description box auto populating with text.

## Alternatives Considered

N/A

## Additional Context

User initially asked if it was possible to disable this automatically fetching text for the optional description box as well as the optional brand icon. After discussing with @swissspidy he suggested that we could add a special handling for Instagram links to fix the text link optional description box auto populating with text.

spacedmonkey:
## Summary

Add a special workaround for instagram based urls in link preview. 

## Testing Instructions

1. Create a story. 
2. Insert a image element. 
3. Add an instagram link to image element. Example: https://www.instagram.com/googleforcreators
4. See title `Instagram - @googleforcreators` or and no link icon

## Checklist

---

Fixes #10451

### DIFF
--- a/includes/REST_API/Link_Controller.php
+++ b/includes/REST_API/Link_Controller.php
@@ -165,7 +165,7 @@ class Link_Controller extends REST_Controller implements HasRequirements {
 		// See https://github.com/GoogleForCreators/web-stories-wp/issues/10451.
 		$matches      = [];
 		$query_string = wp_parse_url( $url, PHP_URL_QUERY );
-		$check_url    = str_replace( "?$query_string", '', $url );
+		$check_url    = \is_string( $query_string ) ? str_replace( "?$query_string", '', $url ) : $url;
 		if ( preg_match( '~^https?://(www\.)?instagram\.com/([^/]+)/?$~', $check_url, $matches ) ) {
 			$data['title'] = sprintf(
 				/* translators: %s: Instagram username. */
--- a/includes/REST_API/Link_Controller.php
+++ b/includes/REST_API/Link_Controller.php
@@ -161,10 +161,21 @@ class Link_Controller extends REST_Controller implements HasRequirements {
 			'description' => '',
 		];
 
+		// Do not request instagram.com, as it redirects to a login page.
+		// See https://github.com/GoogleForCreators/web-stories-wp/issues/10451.
+		$url_host = wp_parse_url( $url, PHP_URL_HOST );
+		if ( \is_string( $url_host ) && false !== strpos( $url_host, 'instagram.com' ) ) {
+			set_transient( $cache_key, wp_json_encode( $data ), $cache_ttl );
+			$response = $this->prepare_item_for_response( $data, $request );
+
+			return rest_ensure_response( $response );
+		}
+
 		$args = [
 			'limit_response_size' => 153600, // 150 KB.
 			'timeout'             => 7, // phpcs:ignore WordPressVIPMinimum.Performance.RemoteRequestTimeout.timeout_timeout
 		];
+
 
 		/**
 		 * Filters the HTTP request args for link data retrieval.
--- a/includes/REST_API/Link_Controller.php
+++ b/includes/REST_API/Link_Controller.php
@@ -163,22 +163,17 @@ class Link_Controller extends REST_Controller implements HasRequirements {
 
 		// Do not request instagram.com, as it redirects to a login page.
 		// See https://github.com/GoogleForCreators/web-stories-wp/issues/10451.
-		$parse_url = wp_parse_url( $url );
-		if ( \is_array( $parse_url ) ) {
-			$url_host    = (string) str_replace( 'www.', '', $parse_url['host'] ?? '' );
-			$path_pieces = explode( '/', $parse_url['path'] ?? '' );
-			$path_pieces = array_filter( $path_pieces );
-			if ( \count( $path_pieces ) === 1 && false !== strpos( $url_host, 'instagram.com' ) ) {
-				$data['title'] = sprintf(
-					/* translators: %s: Instagram username. */
-					__( 'Instagram - @%s', 'web-stories' ),
-					array_shift( $path_pieces )
-				);
-				set_transient( $cache_key, wp_json_encode( $data ), $cache_ttl );
-				$response = $this->prepare_item_for_response( $data, $request );
+		$matches = [];
+		if ( preg_match( '~^https?://(www\.)?instagram\.com/(.*)/?$~', $url, $matches ) ) {
+			$data['title'] = sprintf(
+				/* translators: %s: Instagram username. */
+				__( 'Instagram - @%s', 'web-stories' ),
+				$matches[2]
+			);
+			set_transient( $cache_key, wp_json_encode( $data ), $cache_ttl );
+			$response = $this->prepare_item_for_response( $data, $request );
 
-				return rest_ensure_response( $response );
-			}
+			return rest_ensure_response( $response );
 		}
 
 		$args = [
--- a/includes/REST_API/Link_Controller.php
+++ b/includes/REST_API/Link_Controller.php
@@ -164,7 +164,7 @@ class Link_Controller extends REST_Controller implements HasRequirements {
 		// Do not request instagram.com, as it redirects to a login page.
 		// See https://github.com/GoogleForCreators/web-stories-wp/issues/10451.
 		$matches = [];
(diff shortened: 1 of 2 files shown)


includes/REST_API/Link_Controller.php
```diff
@@ -161,11 +161,22 @@ public function parse_link( $request ) {
...
+
+			return rest_ensure_response( $response );
+		}
+
 		$args = [
 			'limit_response_size' => 153600, // 150 KB.
 			'timeout'             => 7, // phpcs:ignore WordPressVIPMinimum.Performance.RemoteRequestTimeout.timeout_timeout
 		];
 
+
```
spacedmonkey:
```suggestion
```

swissspidy:
On the ticket I mentioned we should add a custom description instead of nothing. Is that not possible?

This is already a good improvement though, so having a custom description is just a bonus.

spacedmonkey:
> On the ticket I mentioned we should add a custom description instead of nothing. Is that not possible?

How would that work? There are other kind of url on the instagram domain, like linking to a photo, like this. 
- https://www.instagram.com/p/CZ8wpcBNsh1/
- https://www.instagram.com/stories/tagdrinks/2776282053056194615/
- https://www.instagram.com/reel/CRZI8PnCfvw/
- https://www.instagram.com/stories/highlights/18119512600136497/
- https://help.instagram.com/581066165581870
- https://www.instagram.com/about/jobs/
- https://www.instagram.com/legal/privacy/

It can be hard to map the url structures to what title we wish it to be. We could only get user handle from the url, but the worry is there is a url like this  https://www.instagram.com/about/jobs/. TLDR, I think any solution would be risky and breakable.

swissspidy:
Just handling links to profiles, e.g. `https://www.instagram.com/googleforcreators/`, would be good enough. The rest is negligible.

swissspidy (approved):

spacedmonkey:
> Just handling links to profiles, e.g. https://www.instagram.com/googleforcreators/, would be good enough. The rest is negligible.

Okay, I will take a look later.

includes/REST_API/Link_Controller.php
```diff
@@ -163,12 +163,22 @@ public function parse_link( $request ) {
 
 		// Do not request instagram.com, as it redirects to a login page.
 		// See https://github.com/GoogleForCreators/web-stories-wp/issues/10451.
-		$url_host = wp_parse_url( $url, PHP_URL_HOST );
-		if ( \is_string( $url_host ) && false !== strpos( $url_host, 'instagram.com' ) ) {
-			set_transient( $cache_key, wp_json_encode( $data ), $cache_ttl );
-			$response = $this->prepare_item_for_response( $data, $request );
+		$parse_url = wp_parse_url( $url );
```
timarney: Maybe not in this PR but should this code be broken into a function for reuse outside of the `instagram.com` 

Possibly looping over an array of social sites as needed.

spacedmonkey: I don't think that is in scope. But should be interesting for the future.

includes/REST_API/Link_Controller.php
```diff
@@ -161,6 +161,26 @@ public function parse_link( $request ) {
 			'description' => '',
 		];
 
+		// Do not request instagram.com, as it redirects to a login page.
+		// See https://github.com/GoogleForCreators/web-stories-wp/issues/10451.
+		$parse_url = wp_parse_url( $url );
+		if ( \is_array( $parse_url ) ) {
```
swissspidy: Why not just use a regex? Would be more readable IMO

spacedmonkey: Regexs are expensive and hard to read. Parse url is easy to use and read.

swissspidy: This is definitely harder to read.

includes/REST_API/Link_Controller.php
```diff
@@ -161,6 +161,26 @@ public function parse_link( $request ) {
...
+					/* translators: %s: Instagram username. */
+					__( 'Instagram - @%s', 'web-stories' ),
+					array_shift( $path_pieces )
+				);
+				set_transient( $cache_key, wp_json_encode( $data ), $cache_ttl );
+				$response = $this->prepare_item_for_response( $data, $request );
+
+				return rest_ensure_response( $response );
+			}
+		}
```
swissspidy:
```suggestion
		// Do not request instagram.com, as it redirects to a login page.
		// See https://github.com/GoogleForCreators/web-stories-wp/issues/10451.
		$matches = [];
		if ( preg_match( '~^https?://(www\.)?instagram\.com/(.*)/?$~', $url, $matches ) ) {
			$data['title'] = sprintf(
				/* translators: %s: Instagram username. */
				__( 'Instagram - @%s', 'web-stories' ),
				$matches[2]
			);
			set_transient( $cache_key, wp_json_encode( $data ), $cache_ttl );
			$response = $this->prepare_item_for_response( $data, $request );

			return rest_ensure_response( $response );
		}
```

This is more concise and readable in my opinion.

swissspidy (approved):

timarney (approved):

felipebochehin87:
Hi @swissspidy! I kindly need your support here, because there's something I found: when we add a URL such as _https://www.instagram.com/bocha1711/?hl=en_ (that's my personal profile), the description is still displayed as **Login * Instagram** (as can be seen below). @spacedmonkey investigated and noticed that it's because this URL has _?hl=en_ in the end. The same profile without this _?hl=en_ works. My question is: should we work on this now in this PR or should I create a new PR for this change?

Thanks!

image (53)

includes/REST_API/Link_Controller.php
```diff
@@ -161,6 +161,21 @@ public function parse_link( $request ) {
 			'description' => '',
 		];
 
+		// Do not request instagram.com, as it redirects to a login page.
+		// See https://github.com/GoogleForCreators/web-stories-wp/issues/10451.
+		$matches = [];
+		if ( preg_match( '~^https?://(www\.)?instagram\.com/([^/]+)/?$~', $url, $matches ) ) {
```
swissspidy: I see two simple solutions to account for the scenario Felipe has mentioned:

**Option A:** Remove query params before doing the regex.

```suggestion
		$url_without_query = explode( '?', $url )[0];
		if ( preg_match( '~^https?://(www\.)?instagram\.com/([^/]+)/?$~', $url_without_query, $matches ) ) {
```

(Yes, `$matches = []` can be safely removed! My bad for suggesting that initially. There is no need to define that variable beforehand!)

**Option B:** Update regex to account for query param.

```suggestion
		$matches = [];
		if ( preg_match( '~^https?://(www\.)?instagram\.com/([^/]+)/?(\?.*)?$~', $url, $matches ) ) {
```

swissspidy: I tend towards option A because it allows us to keep the regex simple and understandable.

tests/phpunit/integration/tests/REST_API/Link_Controller.php
```diff
@@ -259,6 +283,89 @@ public function test_empty_url(): void {
...
+
+		// Subsequent requests is cached and so it should not cause a request.
+		rest_get_server()->dispatch( $request );
+
+		$this->assertEquals( $num, $this->request_count );
+		$this->assertNotEmpty( $data );
+		$this->assertEqualSetsWithIndex( $expected, $data );
+	}
+
+	public function data_instagram_urls(): array {
```
swissspidy, line 307: Neat 👍

tests/phpunit/integration/tests/REST_API/Link_Controller.php
```diff
@@ -259,6 +283,89 @@ public function test_empty_url(): void {
...
+		return [
+			'Instagram profile url'                   => [
+				'url'      => self::URL_INSTAGRAM,
+				'expected' => [
+					'title'       => 'Instagram - @googleforcreators',
+					'image'       => '',
+					'description' => '',
+				],
+				'num'      => 0,
```
swissspidy: Nit: maybe rename to `request_count`?

spacedmonkey: Fixed.

swissspidy (approved):

PR #7766, opened 2021-05-28
Media: Transcode videos inserted from WordPress media modal

spacedmonkey:
## Summary

All for existing video that can be transcode to be shown in the default WordPress media picker. Once the video is select, insert a placeholder into content. From there, generate poster image and start the transcoding process. 

Some video types, like mpg, the video will not have height and width data, so a default height and width will need to be set. 

The allow list of transcodable mime type will have to moved to PHP, as it is needed for the jetpack integration. A filter is not required for this list of transcodable mime types. 

This PR will likely be effected by https://github.com/google/web-stories-wp/pull/7699 https://github.com/google/web-stories-wp/pull/7691 https://github.com/google/web-stories-wp/pull/7317. I would like to see all these PRs merged first, so this PR can be rebased and retested. 

## To-do

- [ ] Add some kind of test
- [ ] Fix existing tests.

#### Upload MOV ( that generates height and width )

https://user-images.githubusercontent.com/237508/119982431-bc5b8680-bfb6-11eb-918f-a6bd0aef6f1c.mov

#### Upload mpg  ( that generates doesn't height and width )
https://user-images.githubusercontent.com/237508/119982391-afd72e00-bfb6-11eb-8905-1cf89776117d.mov

### QA

- [ ] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1. Enable Video optimization in media picker feature flag. 
2. Create new story. 
3. Click upload button. 
4. Upload, .mov or .mpg file. 
5. Wait for file to upload. 
6. Once complete, click insert in to page. 
7. See video is inserted. 
8. See placeholder is inserted.
9. See update once poster is generated, element may change size and position. 
10. See that video controls are not visible. 
11. Once trancode is finished, video should appear in sidebar and video should be playable. 

This functionality should be work, if users disables transcoding options. See dashboard -> settings. 

1. Enable Video optimization in media picker feature flag. 
2. Go to media in sidebar. 
3. Upload .mov or .mpg file by clicking "add new"
2. Create new story. 
3. Click upload button. 
5. Click on media library. 
6. Once complete, click insert in to page. 
7. See video is inserted. 
8. See placeholder is inserted.
9. See update once poster is generated, element may change size and position. 
10. See that video controls are not visible. 
11. Once trancode is finished, video should appear in sidebar and video should be playable. 

### UAT

- [ ] UAT should use the same steps as above.

This PR can be tested by following these steps:

1.

## Checklist

[...]

PR #10604, opened 2022-02-17
Media: Move list of transcodable mime types to JS 

### LINKED ISSUE #10591: Media: Move list of transcodable mime types to JS
## Task Description

We currently have this hardcoded list of video mime types supported by FFmpeg in PHP here that we simply pass to the editor via the inline config:

https://github.com/GoogleForCreators/web-stories-wp/blob/19aa34528cadcb2b0554d76ccb19ff3537c7cf39/includes/Media/Types.php#L175-L202

It does not make sense for this to live there because this list is hardcoded and is coupled to the FFmpeg binary. Passing this via config just makes any custom editor implementation (in a decoupled context) unnecessarily cumbersome.

**To-do**

* Remove this hardcoded list from PHP
* Define a new `TRANSCODABLE_MIME_TYPES` array in `packages/story-editor/src/app/media/constants.js`
* `export * from './constant.js` in `packages/story-editor/src/app/media/index.js`
* Change all usage of `transcodableMimeTypes` from config to use this const again

timarney:
## Context

Moves transcodable mime types out of PHP / config and into a constants file 

See: https://github.com/GoogleForCreators/web-stories-wp/issues/10591#issue-1140124787

## Summary

Moves the mime types into `packages/story-editor/src/app/media/index.js`  + updates code to pull from that location vs pulling from the config.

## To-do

N/A

## User-facing changes

N/A

## Testing Instructions

@spacedmonkey will test the VideoPress / JetPack code see:
https://github.com/GoogleForCreators/web-stories-wp/pull/10604#issuecomment-1046830698

1.

No

No

No

## Checklist

---
...starting with JS files will add removal of PHP code soon.

Fixes #10591


packages/story-editor/src/elements/media/constants.js
```diff
@@ -53,3 +53,23 @@ export const MEDIA_PANELS = [
   PanelTypes.ANIMATION,
   PanelTypes.LINK,
 ];
+
+export const TRANSCODABLE_MIME_TYPES = [
```
swissspidy: This must be in a different location, please check the ticket description.

includes/Integrations/Jetpack.php
```diff
@@ -184,12 +184,6 @@ public function filter_ajax_query_attachments_args( $args ) {
 				// Load filter at 15, so it load after Media\Media\wp_prepare_attachment_for_js which is loaded at 10.
 				add_filter( 'wp_prepare_attachment_for_js', [ $this, 'filter_admin_ajax_response' ], 15, 2 );
 			}
-
-			$allowed_mime_types_transcodable = array_merge( $allowed_mime_types, $this->types->get_allowed_transcodable_mime_types() );
-			if ( ! array_diff( $allowed_mime_types_transcodable, $args['post_mime_type'] ) ) {
-				// Load filter at 15, so it load after Media\Media\wp_prepare_attachment_for_js which is loaded at 10.
-				add_filter( 'wp_prepare_attachment_for_js', [ $this, 'filter_admin_ajax_response' ], 15, 2 );
-			}
```
swissspidy, line 190: Hmm I completely forgot about this.

This was added in #7766.

Quote:

> The allow list of transcodable mime type will have to moved to PHP, as it is needed for the jetpack integration.

For context:

Jetpack VideoPress modifies the WP media modal query args quite a bit and forces every VideoPress video to have the mime type `video/videopress` (which is obviously not a real mime type). We want to hijack into that process to disguise them as MP4s (which they are and which is something our editor understands). This is what `filter_admin_ajax_response()` does.
Now, we want to do that at the right time, i.e. only when this media modal is used within the story editor, not interfering with other screens.

This method here tries to do that by being clever, trying to compare the `post_mime_type` query arg with something that the editor might send.

For this section here specifically I am honestly not sure what purpose it serves. I hope @spacedmonkey remembers.

We need to find a way to target Jetpack VideoPress without comparing the list of transcodable mime types.

includes/Admin/Editor.php
```diff
@@ -215,7 +215,7 @@ public static function get_requirements(): array {
 	 * @param WP_Post    $post    Current post object.
 	 * @return bool|mixed Whether the editor has been replaced.
 	 */
-	public function replace_editor( $replace, $post ) {
+	public function replace_editor( $replace, $post ) { // phpcs:ignore SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingNativeTypeHint
```
swissspidy: Weird, did you get PHPCS errors otherwise? Why now and why not on `main` as well? 🤔

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
```diff
@@ -67,8 +68,8 @@ const {
 } = Icons;
 
 export const MediaPicker = ({ render, ...props }) => {
+  const allowedTranscodableMimeTypes = TRANSCODABLE_MIME_TYPES;
```
swissspidy: This assignment is a bit pointless :-)

packages/story-editor/src/app/media/utils/test/useFFmpeg.js
```diff
@@ -44,25 +45,7 @@ function arrange({ userSettingEnabled, hasUploadMediaAction }) {
...
-      'video/x-h261',
-      'video/x-h263',
-      'video/x-m4v',
-      'video/x-matroska',
-      'video/x-mjpeg',
-      'video/x-ms-asf',
-      'video/x-msvideo',
-      'video/x-nut',
-    ],
+    allowedTranscodableMimeTypes: TRANSCODABLE_MIME_TYPES,
```
swissspidy: This is unneeded. With `TRANSCODABLE_MIME_TYPES` being a constant in the editor now it never needs to be part of the config context. So this line can be removed.

packages/story-editor/src/app/media/utils/useFFmpeg.js
```diff
@@ -128,9 +129,9 @@ const FFMPEG_SHARED_CONFIG = [
  * @return {FFmpegData} Functions and vars related to FFmpeg usage.
  */
 function useFFmpeg() {
+  const allowedTranscodableMimeTypes = TRANSCODABLE_MIME_TYPES;
```
swissspidy: Pointless assignment here as well

packages/story-editor/src/app/uploader/test/useUploader.js
```diff
@@ -53,25 +54,7 @@ function setup(args) {
...
-      'video/x-h261',
-      'video/x-h263',
-      'video/x-m4v',
-      'video/x-matroska',
-      'video/x-mjpeg',
-      'video/x-ms-asf',
-      'video/x-msvideo',
-      'video/x-nut',
-    ],
+    allowedTranscodableMimeTypes: TRANSCODABLE_MIME_TYPES,
```
swissspidy: This can be removed here as well

packages/story-editor/src/components/form/mediaUploadButton.js
```diff
@@ -27,10 +27,12 @@ import PropTypes from 'prop-types';
  */
 import { useConfig, useLocalMedia } from '../../app';
 import useFFmpeg from '../../app/media/utils/useFFmpeg';
+import { TRANSCODABLE_MIME_TYPES } from '../../elements/media/constants';
 
 function MediaUploadButton({ onInsert, renderButton, buttonInsertText }) {
+  const allowedTranscodableMimeTypes = TRANSCODABLE_MIME_TYPES;
```
swissspidy: Ditto about pointless assignment

packages/story-editor/src/components/library/test/_utils/index.js
```diff
@@ -50,25 +51,7 @@ export async function arrange({ mediaResponse = [] }) {
...
-      'video/x-h261',
-      'video/x-h263',
-      'video/x-m4v',
-      'video/x-matroska',
-      'video/x-mjpeg',
-      'video/x-ms-asf',
-      'video/x-msvideo',
-      'video/x-nut',
-    ],
+    allowedTranscodableMimeTypes: TRANSCODABLE_MIME_TYPES,
```
swissspidy: Ditto about removing this since it's unneeded

packages/story-editor/src/karma/fixture/fixture.js
```diff
@@ -100,25 +101,7 @@ export const FIXTURE_DEFAULT_CONFIG = {
...
-    'video/x-h261',
-    'video/x-h263',
-    'video/x-m4v',
-    'video/x-matroska',
-    'video/x-mjpeg',
-    'video/x-ms-asf',
-    'video/x-msvideo',
-    'video/x-nut',
-  ],
+  allowedTranscodableMimeTypes: TRANSCODABLE_MIME_TYPES,
```
swissspidy: Same here. Unneeded & can be removed

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
timarney: I'll fix these up later today ...

To clarify in these case use TRANSCODABLE_MIME_TYPES direct or ?

swissspidy: Yes

timarney:
Couple of tests failing for `useQuickActions.js` around video optimization -- 

Old code used 

```
allowedTranscodableMimeTypes: ['videoMimeType']
```

https://github.com/GoogleForCreators/web-stories-wp/runs/5254791008?check_suite_focus=true

swissspidy:
Looks like the fixture data in that test needs to use an actual proper mime type, not `videoMimeType`

includes/Admin/Editor.php
```diff
@@ -234,7 +234,7 @@ public function replace_editor( $replace, $post ) {
 		}
 
 		return $replace;
-	}
+	}
```
swissspidy:
```suggestion
	}
```

includes/Integrations/Jetpack.php
spacedmonkey: I remember why this code is needed. Consider the following scenario.

- Upload a mov. 
- Videopress converts video to mp4 and uploads to CDN. 
- The attachement's mime type would be video/quicktime. 

This code is needed, to see if the original file was a video file of any kind that has been converted mp4 are usable in editor.

timarney:
> Looks like the fixture data in that test needs to use an actual proper mime type, not videoMimeType

Tried that locally the other day --- but will see what CI does.

includes/Integrations/Jetpack.php
spacedmonkey: Looking at it, if a video has been transcoded to video, why do we care about the orignal video type? 

We could do this. 
```
if ( \in_array( self::VIDEOPRESS_MIME_TYPE, $args['post_mime_type'], true ) ) {
     	add_filter( 'wp_prepare_attachment_for_js', [ $this, 'filter_admin_ajax_response' ], 15, 2 );
}
```

That should be okay.

timarney:
Will dig a little deeper for the failing tests --- they were introduce here:
https://github.com/GoogleForCreators/web-stories-wp/pull/9534

spacedmonkey:
JS unit test failures seem valid breakages. 

Once this is ready for review / testing, I will test against videopress myself.

timarney:
> JS unit test failures seem valid breakages

Seems to have been a issue with the mock for `jest.mock('../../../media');`

packages/story-editor/src/app/highlights/quickActions/test/useQuickActions.js
```diff
@@ -93,8 +93,13 @@ jest.mock('@googleforcreators/media', () => ({
   },
 }));
 
+jest.mock('../../../media', () => ({
+  ...jest.requireActual('../../../media'),
+  useLocalMedia: jest.fn(),
+  TRANSCODABLE_MIME_TYPES: ['video/mp4'],
```
swissspidy: This is not needed for the tests to work.

`useQuickActions` only uses `TRANSCODABLE_MIME_TYPES` this to check whether `optimizeVideo` needs to be called first.

The tests only check whether `postProcessingResource` is ultimately called.

swissspidy (approved):
LGTM with one nit.

Pending more testing with VideoPress

packages/story-editor/src/app/highlights/quickActions/test/useQuickActions.js
swissspidy: @timarney I meant **only** line 99 is not needed; not the whole mock. Now the tests are failing again.

spacedmonkey (approved):
I did some testing on videopress test site. Seems to work as expected. One thing I noticed ( unreleated to this change ), ALL of the .mov I uploaded to the media library ( not from our editor) stripped out the audio. I don't know why videopress does this. But it is not related to editor, just something to note. 

This looks good to go. 👍

PR #10610, opened 2022-02-18
PHPCS: Fix `ReturnTypeHint` sniff configuration

swissspidy:
## Context

@timarney pointed out issues with the PHPCS configuration, where there were unexpected lint errors when running PHPCS on PHP 8+.

The reason for that is that https://github.com/slevomat/coding-standard has different defaults depending on the PHP version.

We were overriding those defaults in our config, however the config was wrong, so it didn't actually work.

## Summary

This PR fixes the PHPCS configuration so that it works again as expected, even on PHP 8.

In the process, this implicitly enabled the `SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingNativeTypeHint` rule which reported a few errors in our tests. This PR addresses those.

This disables some other rules that we don't need (yet) and haven't enabled before.

**Note:** I want to enable `SlevomatCodingStandard.TypeHints.ReturnTypeHint.UselessAnnotation` in a future PR. Makes our inline docs a bit cleaner by removing redundant information.

## Relevant Technical Choices

Formats the XML file using Prettier.

## Testing Instructions

1. Run PHPCS locally on PHP 8+ and see no lint errors.

## Checklist

---

Fixes #

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -83,7 +83,7 @@
   <config name="testVersion" value="7.2-" />
   <rule ref="PHPCompatibilityWP" />
 
-  <!-- The main plugin file should be parsable by PHP < 7.0. -->
+  <!-- The main plugin file should be parsable by PHP 5.6. -->
   <rule ref="Generic.Arrays.DisallowLongArraySyntax.Found">
     <exclude-pattern>web-stories.php</exclude-pattern>
     <exclude-pattern
@@ -108,7 +108,7 @@
       name="SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingTraversableTypeHintSpecification"
     />
 
-    <!-- The main plugin file should be parsable by PHP < 7.0. -->
+    <!-- The main plugin file should be parsable by PHP 5.6. -->
     <exclude-pattern>web-stories.php</exclude-pattern>
     <exclude-pattern
     >includes/compat/Web_Stories_Compatibility.php</exclude-pattern>
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -95,17 +95,25 @@
     >tests/phpunit/integration/includes/REST_Setup.php</exclude-pattern>
   </rule>
 
-  <rule
-    ref="SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingNativeTypeHint"
-  >
+  <rule ref="SlevomatCodingStandard.TypeHints.ReturnTypeHint">
     <properties>
       <property name="enableStaticTypeHint" value="false" />
       <property name="enableMixedTypeHint" value="false" />
       <property name="enableUnionTypeHint" value="false" />
     </properties>
+    <exclude
+      name="SlevomatCodingStandard.TypeHints.ReturnTypeHint.UselessAnnotation"
+    />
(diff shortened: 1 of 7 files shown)


phpcs.xml.dist
```diff
@@ -95,17 +95,25 @@
...
       <property name="enableUnionTypeHint" value="false" />
     </properties>
+    <exclude
+      name="SlevomatCodingStandard.TypeHints.ReturnTypeHint.UselessAnnotation"
+    />
+    <exclude
+      name="SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingTraversableTypeHintSpecification"
+    />
+
+    <!-- The main plugin file should be parsable by PHP < 7.0. -->
```
spacedmonkey:
```suggestion
    <!-- The main plugin file should be parsable by PHP < 7.2. -->
```

swissspidy: For context, this comment is just copied from above.

spacedmonkey:
Should we change this 

https://github.com/GoogleForCreators/web-stories-wp/blob/0b001ca2d0972b367b408c8fdb1c47b6d5a86c79/.github/workflows/lint-php.yml#L45

To PHP 8.0. This will match what we use locally, be faster and flag issue like this?

timarney:
Tested locally --- now showing `No fixable errors were found` ✅

swissspidy:
> Should we change this
> 
> https://github.com/GoogleForCreators/web-stories-wp/blob/0b001ca2d0972b367b408c8fdb1c47b6d5a86c79/.github/workflows/lint-php.yml#L45
> 
> To PHP 8.0. This will match what we use locally, be faster and flag issue like this?

Yeah let's try that in a separate PR 👍  Wanna give it a go?

spacedmonkey:
> Yeah let's try that in a separate PR 👍 Wanna give it a go?

Issue created - https://github.com/GoogleForCreators/web-stories-wp/issues/10611

tests/phpunit/integration/tests/REST_API/Stories_Users_Controller.php
```diff
@@ -123,7 +123,7 @@ public function test_filter_user_query_capabilities_query_supported(): void {
 		);
 	}
 
-	public function filter_rest_user_collection_params( $query_params ) {
+	public function filter_rest_user_collection_params( array $query_params ): array {
```
spacedmonkey, line 126: This runs throught a filter right? Thought we were not adding type hints to filter callbacks?

spacedmonkey (approved):

timarney:
Noticed PHP 8.0 - WP latest (experimental) (in random order) seems to be failing
https://github.com/GoogleForCreators/web-stories-wp/runs/5248004403?check_suite_focus=true

swissspidy:
@timarney Yep saw that too. This test is allowed to fail. See #8792 for context.

This random-order test can detect scenarios where the test makes assumptions that are only fulfilled by global state manipulated by a previous test, but are not true in the actual logic to be tested.

In case such a randomly ordered test run fails, the random seed can be found in the GHA log so that this particular ordering can be replicated locally.

So ideally we'd investigate it in a dedicated ticket to improve our test robustness.

PR #10613, opened 2022-02-18
Code Quality: Remove useless `@return` PHPDoc annotations

swissspidy:
## Context

As mentioned on #10610, I want to remove redundant PHPDoc annotations from our code that are already documented using the actual return type hints.

## Summary

Enables the `SlevomatCodingStandard.TypeHints.ReturnTypeHint.UselessAnnotation` rule by default.

Makes our inline docs a bit cleaner by removing redundant information.

## Testing Instructions

1.

## Checklist

---

Fixes #

### DIFF
--- a/includes/AMP/Canonical_Sanitizer.php
+++ b/includes/AMP/Canonical_Sanitizer.php
@@ -53,8 +53,6 @@ class Canonical_Sanitizer extends AMP_Base_Sanitizer {
 	 * Sanitize the HTML contained in the DOMDocument received by the constructor.
 	 *
 	 * @since 1.1.0
-	 *
-	 * @return void
 	 */
 	public function sanitize(): void {
 		$canonical_url = $this->args['canonical_url'];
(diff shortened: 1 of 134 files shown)


spacedmonkey:
This is a personal thing, but I like having the return type in the PHPDocs. Here are some reasons. 

- Not all code analysis and code sniffing tools support return types. 
- PHPDoc return types normally have a comment, improving documentation. 
- Some functions and methods that do not have returns types will need phpdocs return type, meaning in consistency in the codebase.

swissspidy:
Thanks for sharing your feedback, I appreciate it!

It is worth noting that all our tooling supports return type hints. Type declarations were introduced in PHP 7.0, **7 years ago**. If a tool does not support that, we should absolutely not use it and run far, far away.

Regarding comments, this change here only removes/flags useless PHPDoc `@return` annotations _without a comment_. If we need a comment to explain something better, we can (and should) still do that without problems.

As for consistency, I am not really concerned about that because we don't have to worry about that ourselves. Our tooling is taking care of that and tells us when we need to add something or can safely remove it.

spacedmonkey (approved):
As noted in our sync, I don't feel strongly about this change. I can understand why it is an improve. Happy to merge it

Issue #10621, opened 2022-02-18
Story Details Modal: e2e Tests


BrittanyIRL:
Separating out e2e tests for the new story details modal from #10115.

Let's write some new tests!
- Swap of 'publish' and 'schedule' in button copy based on date selected 
- interacting with different inputs (title, description, slug, taxonomy...) 
- changing cover image
- clicking on checklist

PR #10622, opened 2022-02-18
New Story Details Modal: Enable Experiment By Default

BrittanyIRL:
## Context

## Summary

Enables the new story details modal that opens when publish is clicked. 

## Relevant Technical Choices

Tweak to private publish option in `wp-story-editor` so that old checklist dialog doesn't pop up on top of the new modal, that 'do you want to review the checklist' dialog will be sunset in favor of this new flow. 

## To-do

I will add e2e tests specifically for this modal separately: https://github.com/GoogleForCreators/web-stories-wp/issues/10621

These are the outstanding issues for this feature: 
- https://github.com/GoogleForCreators/web-stories-wp/issues/10623 
- https://github.com/GoogleForCreators/web-stories-wp/issues/10618
- https://github.com/GoogleForCreators/web-stories-wp/issues/10619
- https://github.com/GoogleForCreators/web-stories-wp/issues/10584 (not part of this release) 

## User-facing changes

Now when you click "publish" on a story in the editor you see this new fancy 💅  modal! 
Screen Shot 2022-02-18 at 3 07 21 PM

https://user-images.githubusercontent.com/10720454/154770259-c3c7f5b7-b461-48ad-9866-4e46ac3b8c52.mp4

https://user-images.githubusercontent.com/10720454/154770720-abfe789a-4a1c-4051-a67c-2575cb37f02a.mp4

## Testing Instructions

This is the modal that shows up before stories get published. Most of this functionality is lifted from the editor and just consolidated into one place as the key info that a user should add or at least see before publishing a story. 

#### Cover Preview
- Has the potential to show the publisher logo in upper left (LTR) corner, the poster image (featured image), story title, and the site title (accessed via WordPress profile). If any of these are missing, the view should fall back gracefully (aka no crashes or weird styles. This design is based on Story covers you'd see in Google Search results for real stories, so it breaks from Figma - it's intentional. 

#### Checklist 
- This is a button that when you click it the story details modal collapses and the checklist expands (or stays open if it's already open). The button is counting priority issues present in the story. It should maintain consistency with the checklist button that you can see in the editor. 
  - A caveat to this that I'm still sorting through is when you try to publish a 'bad' story (less than 4 pages) the count of issues "wrong" with a story doesn't show in the editor's checklist button. So, this causes an inconsistency because the count of priority issues will always show when > 0 in the story details modal. 

#### Publish Button Inside of Story Details Modal 
- ~*NEW!* this button is disabled when there's no title 😲  - before this point you could publish stories with no title. Now you cannot. (One caveat of this is within the visibility panel -> private publish, but that's a WordPress thing).~ (Removed this, but didn't redo the screencast) 
- Button copy should be "publish" unless publish date is in the future, then it should be "Schedule". This button should also be in parallel with the publish button of the editor that is visible behind the modal. 
  - You may notice that the publish button on the editor, if there's priority issues in the checklist to look at and story has more than 4 pages, that there's an exclamation icon and a tooltip - that's missing right now from the dialog because it was an edge case I forgot. I'm making a new ticket for it but that logic is tied to the mentioned checklist button count so it makes sense to separate it out. 
 
#### Center Pane 
- Story Title functions same as in main view.
- Story description functions the same as the document pane's version. 
- Visibility works same as in document pane. 

#### Document Pane 
- Panels are limited to: Publishing, Permalink, Page Advancement, Background, Audio, Taxonomies  (Visibility panel is in the center pane)
- Functionality here should be exactly the same as in the document panel on the editor 

### Usability 
- So all of the above should be usable in LTR and RTL both by cursor and keyboard.
- Contributors are the only user role that can see the editor but can't publish stories, so you should still see submit for review not publish on the editor for them. 

No

Nothing new, but it does add a new point for data collection of `review_prepublish_checklist` (which is tracked from the publish button already when users opt to check the checklist for outstanding issues before publishing)

No

## Checklist

---

Partially addresses #10115


packages/story-editor/src/components/publishModal/publishModal.js
```diff
@@ -109,6 +109,7 @@ function PublishModal({ isOpen, onPublish, onClose, publishButtonCopy }) {
       overlayStyles={{
         backgroundColor: theme.colors.inverted.interactiveBg.modalScrim,
       }}
+      id="modal-story-details-publish"
```
swissspidy: Do we need this, can't we target this modal in some other way?

swissspidy:
> I will add e2e tests specifically for this modal separately: #10621

Curious, what's the reasoning for doing these _after_ enabling the feature?

maxyinger (approved):

packages/story-editor/src/components/publishModal/publishModal.js
BrittanyIRL: This was how the previous publish modal is getting targeted by e2e so i was just mirroring that. can swap it to a testid.

BrittanyIRL:
> 

Not a bad question. To make the release deadline, I decided to split out the e2e tests from here so that this branch could get QA'd before the bug bash. While QA is working through the lengthy manual testing as outlined above, my plan is to write the e2e tests (today) based on this same branch behind the scenes. 

#10115 was pretty under scoped when I received it, on Friday we had pretty dang near completion so I separated out the remaining work to avoid an "8 point ticket" getting worked on for another (4th) week, dipping into a 3rd sprint, and to more accurately track where time is going. Hindsight I should have converted 10115 into an epic of its own and made individual issues, but I didn't.

samwhale (approved):
Great pr description

BrookeGraham (approved):
💅

packages/story-editor/src/components/publishModal/publishModal.js
swissspidy: A testid would be more appropriate. Or even better targeting aria roles where possible.

BrittanyIRL: yeah i'll clean all of this up over here: https://github.com/GoogleForCreators/web-stories-wp/pull/10669

PR #10643, opened 2022-02-21
Prevent endless loop on empty font array in `FontProvider`

### LINKED ISSUE #10596: FontProvider creating endless loop
Standalone editor, font provider calling getFonts function and creating endless loop.
This code block creating endless loop.

https://github.com/GoogleForCreators/web-stories-wp/blob/2f73bb644cf2313672e344891c02c7c838be08d4/packages/story-editor/src/app/font/fontProvider.js#L54-L85

## Steps to Reproduce
https://codesandbox.io/s/laughing-heyrovsky-eefv2

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

mohdsayed:
## Context

See https://github.com/GoogleForCreators/web-stories-wp/issues/10596

## Testing Instructions

1.

## Checklist

---

Fixes https://github.com/GoogleForCreators/web-stories-wp/issues/10596

### DIFF
--- a/packages/story-editor/src/app/font/fontProvider.js
+++ b/packages/story-editor/src/app/font/fontProvider.js
@@ -18,19 +18,13 @@
  * External dependencies
  */
 import PropTypes from 'prop-types';
-import {
-  useCallback,
-  useEffect,
-  useRef,
-  useState,
-} from '@googleforcreators/react';
+import { useCallback, useRef, useState } from '@googleforcreators/react';
 import { CURATED_FONT_NAMES } from '@googleforcreators/fonts';
 import { useFeature } from 'flagged';
 
 /**
  * Internal dependencies
  */
-import { trackError } from '@googleforcreators/tracking';
 import loadStylesheet from '../../utils/loadStylesheet';
 import { FONT_WEIGHT_NAMES } from '../../constants';
 import { useAPI } from '../api';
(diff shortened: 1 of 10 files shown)


packages/story-editor/src/components/panels/design/textStyle/fontPicker.js
```diff
@@ -57,18 +64,33 @@ const FontPicker = forwardRef(function FontPicker(
...
   }));
 
+  useEffect(() => {
+    mounted.current = true;
+
+    getCustomFonts(mounted);
+    getCuratedFonts(mounted);
+
+    return () => {
+      mounted.current = false;
```
swissspidy: This mounted ref is not needed here

mohdsayed: Oh yes, it's no longer needed because the state update happens in `FontProvider` which doesn't unmount. 👍

packages/story-editor/src/app/font/fontProvider.js
```diff
@@ -51,73 +45,51 @@ function FontProvider({ children }) {
...
-          }
-
-          const formattedFonts = _customFonts.map((font) => ({
-            id: font.family,
-            name: font.family,
-            value: font.family,
-            ...font,
-          }));
+  const getCustomFonts = useCallback(async () => {
+    if (customFonts || !getFonts || !isCustomFontsEnabled) {
```
swissspidy, line 49: The `getFonts` checks here and below might not be needed anymore. I now recall that I originally added it because of `storyPageToDataUrl`, which uses `FontProvider` to load fonts, but doesn't need custom fonts or curated fonts.

So with this effect gone, this safeguard would be not needed anymore.

mohdsayed: So if `getFonts` is not provided, we will not render `FontPicker` component dropdown at all?

mohdsayed: Like this https://github.com/GoogleForCreators/web-stories-wp/pull/10643/commits/a27975166e512bd5494e5876fb28c60b1581811c

swissspidy: No I meant

```diff
-if (customFonts || !getFonts || !isCustomFontsEnabled) {
+if (customFonts || !isCustomFontsEnabled) {
```

swissspidy: I meant just removing the `|| !getFonts` check but I suppose it makes sense like this 👍

mohdsayed: https://github.com/GoogleForCreators/web-stories-wp/commit/a27975166e512bd5494e5876fb28c60b1581811c looks wrong because if we remove `getFonts` condition, the standalone editor will throw an error when `getFonts` isn't provided. To prevent this I had added `fonts.length > 0` so that `FontPicker` doesn't get rendered and `getFonts` is never called, but actually, that was a mistake because `fonts` will not have any value until `FontPicker` is rendered so font picked will not render in any case. Reverting the commit changes.

swissspidy (approved):

Issue #10546, opened 2022-02-10
Tooltip maximum depth error on hover


BrookeGraham:
## Bug Description
 When in `RTL` if a `TOOLTIP` is cutoff the left edge of the screen AND cutoff by the bottom of the screen we run into an infinite loop. Likely due to `bottom` and `end` placement overrides fighting one another? 

## Expected Behaviour

App shouldn't crash when hovering over a tooltip. 

## Steps to Reproduce

In `RTL` mode, Open Layers panel and scroll until the `duplicate` and `trash can` icons are just in view. 
Hover the `duplicate` icon and leave mouse positioned over icon. 
Wait a few seconds and you'll get the exceeded max depth error. 

## Screenshots
![Kapture 2022-02-10 at 10 23 48](https://user-images.githubusercontent.com/1820266/153462358-55c2c2e5-7739-4c7e-bff5-df6ee8db7350.gif)

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

amovar18:
@BrookeGraham looked into this i think it could be solved by adding `ignoreMaxOffsetY` just in `Tooltip` component inside the story-editor.

felipebochehin87:
Tested against https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin, using PR #10668.

1. Opened the app on RTL mode
2. Hovered over the Duplicate Layer icon
3. Waited for a few seconds - max depth error was not displayed

PR #10665, opened 2022-02-21
Design System: Tooltip fix stale placement

### LINKED ISSUE #10544: Tooltip getting stale dynamicPlacement state in RTL
## Bug Description

When in RTL `Tooltip` is getting a stale value for `placement` or something to that effect. On elements that have a `BOTTOM` placement and they would be cutoff by the bottom of the screen, `Tooltip` updates to having a `TOP` placement, but only on every other hover. In some instances, it doesn't get updated and instead the anchor point is pushed up instead. 

## Expected Behaviour

`Tooltip` should show in the correct placement every time you hover to see it. 

## Steps to Reproduce

In `RTL` open the Background  Audio panel and position window so that it is just below the `trash can` and  `play` icons. Hover icons and notice it switches from `BOTTOM` to `TOP` position on ever other hover event. 

## Screenshots

Kapture 2022-02-10 at 10 21 06

Screen Shot 2022-02-10 at 9 59 45 AM

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

BrookeGraham:
## Context
When a `tooltip` is near the bottom of the screen and has a `placement` value of `BOTTOM` we update the placement via `dynamicPlacement` to be `TOP` this avoids the content of tooltip being cutoff by the bottom of the window. However, we weren't always getting the correct (most recent) values for whether or a not a tooltip was showing outside of the window. 

## Summary
The dynamic placement for updating `bottom` to `top` was being calculated with `popupRect.bottom`. This was checked against `window.innerHeight` to determine whether the `tooltip` was being cutoff by the bottom of the window. This would get out of sync when the placement of the tooltip changes. 

## Relevant Technical Choices
Instead of basing the calculation on the `popupRect`'s `bottom` value we are now calculating the values ourselves using `offset.popupHeight` in conjunction with the anchor's `y` offset. This helps avoid stale values (ie the height of a popup will be the same regardless of where we place it) and allows the `offset.y` to utilize  `ignoreMaxOffsetY` and `Math.max(topOffset, Math.min(offsetY, maxOffsetY))`.

## User-facing changes
|Before|After|
|--|--|
|before|after|

## Testing Instructions

1. Add background audio if one is not a part of the story
2. Position the `trash` icon and `play` icon just above the bottom edge of the window
3. Hover the `trash` icon or `play` icon and see that the tooltip is placed above the icons
4. Repeatedly hover to ensure the placement isn't  switching 
5. move the panel up so that there is enough space below the icons to see the full tooltip
6. Hover the `trash` icon or `play` icon and see that the tooltip is placed below the icons
7. Repeatedly hover to ensure the placement isn't  switching 
Note: there is still the issue of the editor crashing when in RTL and the tooltip is cutoff both at the bottom and on the left. Addressing in #10546 

no

no

no

## Checklist

---

Fixes #10544

### DIFF
--- a/packages/design-system/src/components/tooltip/index.js
+++ b/packages/design-system/src/components/tooltip/index.js
@@ -37,7 +37,7 @@ import { THEME_CONSTANTS } from '../../theme';
 import { Text } from '../typography';
 import { RTL_PLACEMENT } from '../popup/constants';
 import { noop } from '../../utils';
-import { SvgForTail, Tail, SVG_TOOLTIP_TAIL_ID } from './tail';
+import { SvgForTail, Tail, SVG_TOOLTIP_TAIL_ID, TAIL_HEIGHT } from './tail';
 
 const SPACE_BETWEEN_TOOLTIP_AND_ELEMENT = 8;
 // For how many milliseconds is a delayed tooltip waiting to appear?
@@ -177,14 +177,13 @@ function Tooltip({
   const positionPlacement = useCallback(
     ({ offset }) => {
       // check to see if there's an overlap with the window's bottom edge
-      const neededVerticalSpace = offset.bottom;
+      const neededVerticalSpace = offset.y + offset.height + TAIL_HEIGHT;
       const shouldMoveToTop =
         dynamicPlacement.startsWith('bottom') &&
         neededVerticalSpace >= window.innerHeight;
       // check that the tooltip isn't cutoff on the left edge of the screen.
       // right-cutoff is already taken care of with `getOffset`
       const isOverFlowingLeft = offset.popupLeft < 0;
-
       if (shouldMoveToTop && !isOverFlowingLeft) {
         setDynamicPlacement(PLACEMENT.TOP);
       } else if (shouldMoveToTop && isOverFlowingLeft) {
--- a/packages/design-system/src/components/tooltip/index.js
+++ b/packages/design-system/src/components/tooltip/index.js
@@ -39,7 +39,7 @@ import { RTL_PLACEMENT } from '../popup/constants';
 import { noop } from '../../utils';
 import { SvgForTail, Tail, SVG_TOOLTIP_TAIL_ID, TAIL_HEIGHT } from './tail';
 
-const SPACE_BETWEEN_TOOLTIP_AND_ELEMENT = 8;
+const SPACE_BETWEEN_TOOLTIP_AND_ELEMENT = TAIL_HEIGHT;
 // For how many milliseconds is a delayed tooltip waiting to appear?
 const DELAY_MS = 1000;
 // For how many milliseconds will triggering another delayed tooltip show instantly?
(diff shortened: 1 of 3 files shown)


packages/design-system/src/components/tooltip/index.js
```diff
@@ -177,7 +177,7 @@ function Tooltip({
   const positionPlacement = useCallback(
     ({ offset }) => {
       // check to see if there's an overlap with the window's bottom edge
-      const neededVerticalSpace = offset.bottom;
+      const neededVerticalSpace = offset.y + offset.popupHeight + TAIL_HEIGHT;
```
BrookeGraham: Tail height is needed regardless of whether `hasTail` is true or not because the popup is placed in the same spot, ie 8 px below the anchor.

samwhale (approved):
Looking good!

packages/design-system/src/components/tooltip/tail.js
```diff
@@ -27,7 +27,7 @@ import { PLACEMENT } from '../popup';
 
 export const SVG_TOOLTIP_TAIL_ID = 'tooltip-tail';
 const TAIL_WIDTH = 34;
-const TAIL_HEIGHT = 8;
+export const TAIL_HEIGHT = 8;
```
samwhale, line 30: 💭 Probably not worth updating but it seems like there should be two different variables here and they happen to be the same value:
1. The tail height (8px)
2. The tooltip's distance away from the anchor element (also 8px)

In the future it may be beneficial to split these up, if ever the tooltip should be placed a little further/closer to the anchor.

But I don't think it's worth updating now. Maybe it would be helpful to add a comment saying the extra displacement is necessary because the tooltip is always 8px from the anchor?

BrookeGraham: Yeah, Tooltip has the const `SPACE_BETWEEN_TOOLTIP_AND_ELEMENT = 8;` perhaps the cleanest way to update / sync these would be to edit `const SPACE_BETWEEN_TOOLTIP_AND_ELEMENT = TAIL_HEIGHT`?

BrookeGraham: Or vise versa?

samwhale: If the space is dependent specifically on the tooltip height then I'd do something like

```js
const SPACE_BETWEEN_TOOLTIP_AND_ELEMENT = TAIL_HEIGHT + extraOffset;
```

If it isn't I'd do 
```js
const SPACE_BETWEEN_TOOLTIP_AND_ELEMENT = 8;
```

BrookeGraham: I think I got us confused. `const SPACE_BETWEEN_TOOLTIP_AND_ELEMENT = 8;` Already exists on the Tooltip component [here](https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/design-system/src/components/tooltip/index.js#L42).

Since there is both 
`const SPACE_BETWEEN_TOOLTIP_AND_ELEMENT = 8;`
and 
`const TAIL_HEIGHT = 8;`

I would think we can keep them in sync by updating one to equal the other. I am assuming the `SPACE_BETWEEN_TOOLTIP_AND_ELEMENT` should be the same as the `TAIL_HEIGHT` so that there is enough room for the tail to be placed. 

But now I see just how confusing this is. I _think_ this would be easier to follow if we did something like 
``` 
const TAIL_HEIGHT = 8;
...
const SPACE_BETWEEN_TOOLTIP_AND_ELEMENT = TAIL_HEIGHT;
...
//  In order to check if there's an overlap with the window's bottom edge we need the overall height of the tooltip
//  from the anchor's y position along with the amount of space between the anchor and the tooltip content. 
 const neededVerticalSpace = offset.y + offset.popupHeight + SPACE_BETWEEN_TOOLTIP_AND_ELEMENT;
```

BrittanyIRL (approved):
I like the idea you came up with based on the thread here with Samuel!

maxyinger (approved):

PR #10667, opened 2022-02-21
Editor: Added functional border-radius control to new design menu

### LINKED ISSUE #10624: UI Simplification: Floating Menu: Border radius element
## Feature Description

Make the border radius element of the floating menu functional.

* Design: [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=11394%3A457049)
* Menus: Shape
* Optional rendering: This only renders for rectangular shapes that don't already have varied corner radii. If either condition is not met, this element does not render at all.
* Notes: This element has a following separator inside, that only renders, if the control renders (otherwise the menu would have double separators in a row)

## Additional Context

Screen Shot 2022-02-18 at 18 27 20

barklund:
## Context

This adds a functional border-radius control to the new design menu.

This control is only present for non-rectangular shapes with identical (locked) radii. If the radii are unlocked, but still happen to be identical, the control is still not rendered, as it is the lock flag that determines whether the control is displayed or not.

## User-facing changes

_Not actually user-facing as it is still behind a feature-flag_
| Shape | Screenshot |
|-|-|
| Non-rectangular | Screen Shot 2022-02-21 at 18 24 50 |
| Varied radii | Screen Shot 2022-02-21 at 18 24 54 |
| Rectangular with equal radii | Screen Shot 2022-02-21 at 18 24 58 

## Testing Instructions

## Checklist

---

Fixes #10624


packages/story-editor/src/components/floatingMenu/menu.js
```diff
@@ -42,9 +42,19 @@ const FloatingMenu = memo(
...
+        node.style.setProperty('--width', `${bounds.width.toFixed(2)}px`);
+        node.style.setProperty('--height', `${bounds.height.toFixed(2)}px`);
+        node.style.width = 'var(--width)';
+      };
+      updateSize();
+      // If the menu children list changes, update the size again
+      const observer = new MutationObserver(updateSize);
+      const menu = node.querySelector('[role=menu]');
+      observer.observe(menu, { childList: true });
+      return () => observer.disconnect();
```
barklund, line 57: Note that this was necessary, because changing the border-radius lock toggle in the design panel causes the design menu to change size, but because this previously only recalculated as the selection changed, this change in size was not incorporated. Now this observer will recalculate the size whenever a control is added to or removed from the menu.

However, if we in the future have controls that change size based on external input, we need to re-evaluate this (as this only observes whether nodes are added to or removed from the menu node).

packages/story-editor/src/components/panels/design/sizePosition/radius.js
```diff
@@ -32,7 +32,7 @@ import { useCommonObjectValue, focusStyle } from '../../shared';
 import { MULTIPLE_DISPLAY_VALUE, MULTIPLE_VALUE } from '../../../../constants';
 import Tooltip from '../../../tooltip';
 
-const DEFAULT_BORDER_RADIUS = {
+export const DEFAULT_BORDER_RADIUS = {
```
barklund, line 35: Should this go somewhere else? If so, where?

packages/story-editor/src/components/floatingMenu/elements/borderRadius.js
```diff
@@ -17,21 +17,63 @@
...
  */
-import { IconButton, Separator } from './shared';
+import { useStory } from '../../../app';
+import { canSupportMultiBorder } from '../../../masks';
+import { DEFAULT_BORDER_RADIUS } from '../../panels/design/sizePosition/radius';
+import { Separator, useProperties } from './shared';
+
+const StyledInput = styled(NumericInput)`
+  width: 70px;
+  flex: 0 0 70px;
```
barklund, line 34: I actually don't think both of these are needed. But the first one definitely was, otherwise the menu rendered all weird!

packages/story-editor/src/components/floatingMenu/elements/borderRadius.js
```diff
@@ -17,21 +17,63 @@
...
 /**
  * Internal dependencies
  */
-import { IconButton, Separator } from './shared';
+import { useStory } from '../../../app';
+import { canSupportMultiBorder } from '../../../masks';
+import { DEFAULT_BORDER_RADIUS } from '../../panels/design/sizePosition/radius';
+import { Separator, useProperties } from './shared';
+
+const StyledInput = styled(NumericInput)`
```
barklund, line 32: By the way, I was very happy to see that our existing input component was perfectly styled for this already and just working seemlessly in the menu. I don't know if an input in a menu should be marked up in a particular way, but don't think so. It's not uncommon to have these inputs in a menu (for text editors), so I think it's semantically fine, but there might be some details around focus, that we should add (e.g. pressing arrow keys left-right should move focus, whereas up-down changes the value).

packages/story-editor/src/components/panels/design/sizePosition/radius.js
miina: Maybe either to `elements/constants.js` or to `constants/index.js` directly.

packages/story-editor/src/karma/fixture/containers/designMenu.js
```diff
@@ -39,4 +39,8 @@ export class DesignMenu extends Container {
       ToggleButton
     );
   }
+
+  get borderRadius() {
+    return this.queryByRole('textbox', { name: 'Corner radius' });
```
miina: Nit: It's `Corner Radius` (with capital `R`) in the Design Panel, should we use the same for this, too, for consistency?

miina (approved):
🐬

packages/story-editor/src/karma/fixture/containers/designMenu.js
barklund: Fixed in 135bc61.

PR #10672, opened 2022-02-22
Context Menu: Code and performance updates for `useCopyPaste`

### LINKED ISSUE #10459: Context Menu: Code and performance updates for `useCopyPaste`
## Feature Description

@littlemilkstudio [listed some great ideas](https://github.com/GoogleForCreators/web-stories-wp/pull/10361#pullrequestreview-869371062) in #10361 for perf and code improvements for the new `useCopyPasteActions` hook.

Tasks:
- [Add `updateSelectedElements` action to story reducer](https://github.com/GoogleForCreators/web-stories-wp/pull/10361#discussion_r796786689). Should be helpful since many updates are targeting the selected elements, so there's no need to pull those into the hook from the story reducer. May be usable in other areas too.
- [Make `updateSelectedElements` stable between updates](https://github.com/GoogleForCreators/web-stories-wp/pull/10361#discussion_r796848228). Will help reduce needless re-renders.
- [Look into updating story reducer 'copy element functionality' to something like `copySelectedElementStyles`](https://github.com/GoogleForCreators/web-stories-wp/pull/10361#discussion_r796966823). May be able to add those styles in as a history action, so `ctrl+z` could be leveraged to 'un-copy' the styles. 
    - #10361 updated the story reducer to hold state called [`copiedElementState`](https://github.com/GoogleForCreators/web-stories-wp/pull/10361/files#diff-5a1e5effb14e5ac43c4ffe44c9fa64d9c28b1e8593a627d40c7cdd0bd67a40c8R36). That state may not be the best way to store the copied element styles + animations.
- [Pull out text style updates into their own pure function](https://github.com/GoogleForCreators/web-stories-wp/pull/10361#discussion_r796941465).

## Alternatives Considered

n/a

## Additional Context

n/a

maxyinger:
## Context
Part of looking through right click menu provider. Helps with a little code quality and some perf

## Summary
Some code quality improvements, and some perf improvements around primary actions:

**Selecting new Item**
| Before | After |
| ----- | ----- |
| ~9.2ms | ~0.1ms |
| Screen Shot 2022-02-22 at 9 32 03 AM |  Screen Shot 2022-02-22 at 9 30 42 AM |

**Drag to update element position**
| Before | After |
| ----- | ----- |
| ~2.9ms | ~0.1ms |
| Screen Shot 2022-02-22 at 9 33 03 AM |  Screen Shot 2022-02-22 at 9 32 50 AM |

## Relevant Technical Choices
All perf improvements are around limiting some of the expensive editable text calls when they don't need to be recomputed. the editable text calls are called in some of the hooks of the right click menu provider needed to supply callbacks to some of the right click menu actions.

As far as code quality goes, I think I did all the tasks listed in the ticket except move a `pasteCopiedElementStyles` action into the reducer. At the end of the day, this wouldn't have improved anything re-render wise in the hook or amount of lines the callback is. After looking at it more in depth, it felt a little arbitrary since it wouldn't/couldn't handle pasting of text elements, so figured for now its fine to just leave all the pasting copied element styles logic contained in the same place in this callback defined in the `useCopyPasteActions` hook :shrug: . 

## User-facing changes
NA

## Testing Instructions
NA

1.

## Checklist

---

Fixes #10459


BrittanyIRL (approved):
~0.1ms vs ~9.2ms is an awesome improvement here!! Pulled the code down and did a little local testing too. Looks good. Super interesting how just being a little more careful with what we need and where can make improvements. Thanks for thorough walk through, as ever.

packages/story-editor/src/components/panels/design/textStyle/stylePresets/useApplyStyle.js
```diff
@@ -30,10 +30,8 @@ function useApplyStyle({ pushUpdate }) {
   // Update size and position if relevant values have changed.
   usePresubmitHandler(getUpdatedSizeAndPosition, []);
 
-  const { selectedElements } = useStory(({ state: { selectedElements } }) => {
-    return {
-      selectedElements,
-    };
+  const selectedTextElements = useStory(({ state: { selectedElements } }) => {
+    return selectedElements?.filter(({ type }) => type === 'text');
```
miina: Q: If `selectedElements` is `undefined` here, would https://github.com/GoogleForCreators/web-stories-wp/pull/10672/files#diff-e37fde4c9173759792145832a2359437534a2fd118fa83f85160e7fd69a66cf1R106 error? Since it would be called with `undefined`.

packages/story-editor/src/app/rightClickMenu/hooks/useCopyPasteActions.js
```diff
@@ -80,41 +107,42 @@ const useCopyPasteActions = () => {
...
-              ...updates,
-            });
-            return {
-              ...updates,
-              ...sizeUpdates,
-            };
-          },
-        });
-      }
+      updateSelectedElements({
```
miina, line 131: Just curious (since using similar updater in another PR) -- why is `updateSelectedElements` better than `updateElementsById`?

miina (approved):

BrookeGraham (approved):
Nice! I like it.

packages/story-editor/src/components/panels/design/textStyle/stylePresets/useApplyStyle.js
maxyinger: ah good catch. will add some behavior to account for that

packages/story-editor/src/app/rightClickMenu/hooks/useCopyPasteActions.js
maxyinger: yea great question. so if we need to update selected elements these two are logically equivalent:
```js
const { updateElementsById, selectedElements } = useStory(({ state, actions }) => ({
  updateElementsById: actions.updateElementsById,
  selectedElements: state.selectedElements,
}));

const someCallback = useCallback((...args) => {
  updateElementsById({
    elements: selectedElements,
    properties: args
  })
}, [selectedElements, updateElementsById]);
```
vs
```js
const { updateSelectedElements } = useStory(({ actions }) => ({
  updateSelectedElements: actions.updateSelectedElements,
}));

const someCallback = useCallback((...args) => {
  updateSelectedElements({
    properties: args
  })
}, [updateSelectedElements]);
```
the difference is the latter won't force a re-render on any component consuming this hook for the lifetime of the app since all of our actions are stable, the first one will force a re-render every time `selectedElements` updates.

Here specifically, we're not getting any of the benefit because we're still pulling in `selectedElements` into this hook for other logic, but at least it heads toward that direction.

miina: > Here specifically, we're not getting any of the benefit because we're still pulling in selectedElements into this hook for other logic, but at least it heads toward that direction.

That's why I was wondering why one is better than the other, makes sense now, thanks for the explanation!

samwhale (approved):
Nice this is rad

PR #10677, opened 2022-02-22
Performance: Improve performance on first selection

### LINKED ISSUE #10674: Performance: First Selection of an element after load, Frame Elements
The initial selection of an element after a fresh load in the editor always took a while. I always through this case when auditing element selection out because I thought it was just associated with initial load time.

Finally decided to load the editor, wait a couple seconds till the main thread wasn't doing any work, then audit. Found out that the first selection actually just takes a long time. Seeing between 700ms-900ms total on the main thread (seen up to 1300ms when selecting quickly after load). ~570ms of total react commit time:

The design panels we're still waiting to get into here: https://app.zenhub.com/workspaces/web-stories-5e94d3aced449034e1e9b226/issues/googleforcreators/web-stories-wp/10072

but they're contributing to ~300ms: 

Another ~80ms is contributed to by the layout thrashing issue caused by Moveable under the hood of single selection here:

We're following this in these tickets:
https://github.com/GoogleForCreators/web-stories-wp/pull/10396
https://app.zenhub.com/workspaces/web-stories-5e94d3aced449034e1e9b226/issues/googleforcreators/web-stories-wp/10448

And then we got ~40ms from the frames layer:

this ticket is to fix the frames layer. we should be able to shave off that ~40ms by moving `isSingleElement` into the video `Control` element instead of forcing a re-render in all frame elements

maxyinger:
## Context
Found this auditing some perf stuff

## Summary
Found out that initial selection of components was from some components initially mounting. See writeup [here](https://app.zenhub.com/workspaces/web-stories-5e94d3aced449034e1e9b226/issues/googleforcreators/web-stories-wp/10674)

This prevents the render of every frame element on first selection:
| Before | After |
| --- | --- |
| ~550ms | ~500ms |
| Screen Shot 2022-02-22 at 1 01 41 PM | Screen Shot 2022-02-22 at 12 46 40 PM |

## Relevant Technical Choices
Just pulling in needed data from `useStory` into places that needed vs a higher level where every instance doesn't need it.

## Testing Instructions
NA

1.

## Checklist

---

Fixes #10674

### DIFF
--- a/packages/story-editor/src/elements/video/controls.js
+++ b/packages/story-editor/src/elements/video/controls.js
@@ -22,20 +22,25 @@ import PropTypes from 'prop-types';
 /**
  * Internal dependencies
  */
+import { useStory } from '../../app/story';
 import StoryPropTypes from '../../types';
 import PlayPauseButton from './playPauseButton';
 
 function VideoControls({
   box,
   isSelected,
-  isSingleElement,
   isEditing,
   isTransforming,
   elementRef,
   element,
 }) {
-  const isActive =
-    isSelected && !isTransforming && isSingleElement && !isEditing;
+  const isActive = useStory(
+    ({ state }) =>
+      state.selectedElementIds.length === 1 &&
+      isSelected &&
+      !isTransforming &&
+      !isEditing
+  );
 
   return (
     <PlayPauseButton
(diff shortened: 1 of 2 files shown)


packages/story-editor/src/elements/video/controls.js
```diff
@@ -22,20 +22,25 @@ import PropTypes from 'prop-types';
...
   element,
 }) {
-  const isActive =
-    isSelected && !isTransforming && isSingleElement && !isEditing;
+  const isActive = useStory(
+    ({ state }) =>
+      state.selectedElementIds.length === 1 &&
+      isSelected &&
+      !isTransforming &&
+      !isEditing
```
samwhale, line 42: Cool, never thought to use props in the context selectors.

samwhale (approved):
This is dope

BrookeGraham (approved):
🦖

PR #10681, opened 2022-02-22
Status Panel: Visibility dropdown styling updates

### LINKED ISSUE #10680: Status Panel: dropdown style updates
## Feature Description
When selecting `Private` from the `Visibility` panel, you will never see the helper / hint text that describes what this selection means. We want to update the dropdown so that each item also contains that additional text as a second line.

BrookeGraham:
## Context

When selecting Private from the Visibility panel, you will never see the helper / hint text that describes what this selection means. We want to update the dropdown so that each item also contains that additional text as a second line.

## To-do
n/a

## User-facing changes
||Open Modal|Hover|Focus|
|--|--|--|--|
|After|Screen Shot 2022-02-22 at 1 47 27 PM|Screen Shot 2022-02-22 at 1 21 30 PM|Screen Shot 2022-02-22 at 1 21 22 PM|
|Before|Screen Shot 2022-02-22 at 3 54 16 PM|Screen Shot 2022-02-22 at 3 50 46 PM|Screen Shot 2022-02-22 at 3 50 53 PM|

## Testing Instructions

1.

no

no

no

## Checklist

---

Fixes #10680

### DIFF
--- a/packages/e2e-tests/src/specs/editor/contributorUser.js
+++ b/packages/e2e-tests/src/specs/editor/contributorUser.js
@@ -35,12 +35,14 @@ describe('Contributor User', () => {
 
     await expect(page).toClick('li[role="tab"]', { text: 'Document' });
 
-    await expect(page).toMatchElement('button', { text: 'Public' });
-    await expect(page).not.toMatchElement('li[role="option"]', {
-      text: 'Private Visible to site admins & editors only',
+    await expect(page).toMatchElement('button[disabled]', {
+      text: /^Public/,
     });
     await expect(page).not.toMatchElement('li[role="option"]', {
-      text: 'Password Protected Visible only to those with the password.',
+      text: /^Private/,
+    });
+    await expect(page).not.toMatchElement('li[role="option"]', {
+      text: /^Password Protected/,
     });
   });
 });
--- a/packages/e2e-tests/src/specs/editor/contributorUser.js
+++ b/packages/e2e-tests/src/specs/editor/contributorUser.js
@@ -37,10 +37,10 @@ describe('Contributor User', () => {
 
     await expect(page).toMatchElement('button', { text: 'Public' });
     await expect(page).not.toMatchElement('li[role="option"]', {
-      text: 'Private',
+      text: 'Private Visible to site admins & editors only',
     });
     await expect(page).not.toMatchElement('li[role="option"]', {
-      text: 'Password Protected',
+      text: 'Password Protected Visible only to those with the password.',
     });
   });
 });
--- a/packages/wp-story-editor/src/components/documentPane/status/status.js
+++ b/packages/wp-story-editor/src/components/documentPane/status/status.js
@@ -61,12 +61,16 @@ const StyledListItem = styled.li`
     background-color: ${theme.colors.border.defaultNormal};
   }
 `;
-const LabelText = styled.span`
+const LabelText = styled(Text).attrs({
+  as: 'span',
+})`
   font-size: 14px;
   color: ${theme.colors.fg.primary};
 `;
 
-const HelperText = styled.span`
+const HelperText = styled(Text).attrs({
+  as: 'span',
+})`
   color: ${theme.colors.fg.tertiary};
 `;
 
--- a/packages/wp-story-editor/src/components/documentPane/status/status.js
+++ b/packages/wp-story-editor/src/components/documentPane/status/status.js
@@ -18,9 +18,20 @@
  * External dependencies
  */
 import PropTypes from 'prop-types';
-import { useCallback, useEffect, useState } from '@googleforcreators/react';
+import {
+  useCallback,
+  useEffect,
+  useState,
+  forwardRef,
+} from '@googleforcreators/react';
 import { __ } from '@googleforcreators/i18n';
-import { Input, DropDown } from '@googleforcreators/design-system';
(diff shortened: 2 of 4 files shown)


BrittanyIRL:
Looks like JS unit tests are going to need a little update: https://github.com/GoogleForCreators/web-stories-wp/runs/5296013186?check_suite_focus=true#step:6:583

BrookeGraham:
> Looks like JS unit tests are going to need a little update: https://github.com/GoogleForCreators/web-stories-wp/runs/5296013186?check_suite_focus=true#step:6:583

oh yep. Probably e2e too. Thanks

BrittanyIRL (approved):
Thanks for solving this issue so proactively! Looks good.

Approving ahead of JS unit tests update.

maxyinger (approved):

packages/e2e-tests/src/specs/editor/contributorUser.js
```diff
@@ -37,10 +37,10 @@ describe('Contributor User', () => {
 
     await expect(page).toMatchElement('button', { text: 'Public' });
     await expect(page).not.toMatchElement('li[role="option"]', {
-      text: 'Private',
+      text: 'Private Visible to site admins & editors only',
     });
     await expect(page).not.toMatchElement('li[role="option"]', {
-      text: 'Password Protected',
+      text: 'Password Protected Visible only to those with the password.',
```
samwhale: 💭 I think this was good the way it was before! This is much more specific now, which may break the test in ways we don't want.

Ex scenario: The options are visible, but the help text (`Visible to site admins & editors only`) is removed.

Then this test would still fail even though the option is visible. I think it may be better to use a less specific regular expression here:

```js
await expect(page).not.toMatchElement('li[role="option"]', {
  text: /^Private/,
});
await expect(page).not.toMatchElement('li[role="option"]', {
  text: /^Password Protected/,
});
```

Thoughts?

packages/wp-story-editor/src/components/documentPane/status/status.js
```diff
@@ -37,6 +48,37 @@ import {
...
 import { VISIBILITY, STATUS } from '../../../constants';
 
+const StyledListItem = styled.li`
+  position: relative;
+  padding: 6px 8px;
+  margin: 4px 8px;
+  align-items: center;
+  border-radius: 4px;
+  cursor: pointer;
+  ${themeHelpers.focusableOutlineCSS}
```
samwhale, line 59: 💭 nit but may not be necessary:
```suggestion
  ${themeHelpers.focusableOutlineCSS};
```

packages/wp-story-editor/src/components/documentPane/status/status.js
```diff
@@ -37,6 +48,37 @@ import {
...
+  }
+`;
+const LabelText = styled.span`
+  font-size: 14px;
+  color: ${theme.colors.fg.primary};
+`;
+
+const HelperText = styled.span`
+  color: ${theme.colors.fg.tertiary};
+`;
```
samwhale: 💭 Should these be styled `Text` elements? May not be worth the trouble since this is so easy

samwhale (commented):
Looks good, have some thoughts about the tests.

💭 I think it would be better to add one test that specifically targets the options and verifies that the help text is there. Then the other tests could be updated to match the start of the strings (`Public`, `Private`, `Password Protected`)

samwhale (approved):

packages/e2e-tests/src/specs/editor/contributorUser.js
BrookeGraham: hahaha I was just about to update this! Much better.

packages/wp-story-editor/src/components/documentPane/status/status.js
BrookeGraham: whoops. Thanks.

packages/wp-story-editor/src/components/documentPane/status/status.js
BrookeGraham: I had initially done that so that I could use `THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL` but for some reason the font-size wasn't getting applied properly. I was getting 13px? I looked into it a little but didn't see it so I just went with these spans. 🤷🏻‍♀️

packages/wp-story-editor/src/components/documentPane/status/status.js
swissspidy: I don't think this semicolon is needed 🤔 

`focusableOutlineCSS` is a complete CSS block, so this would just result in something like `&.:focus-visible, &[data-focus-visible-added] {...};` with that extra semicolon at the end

swissspidy: `${themeHelpers.focusableOutlineCSS}` does not have a semicolon anywhere else in our code base either

BrookeGraham:
Holding off on merging until after [10115](https://github.com/GoogleForCreators/web-stories-wp/issues/10115)  gets qa'ed and merged via pr [10622](https://github.com/GoogleForCreators/web-stories-wp/pull/10622)  with e2e tests for the publish modal, I will rebase to make sure I didn't miss any updates.

swissspidy (approved):
Works great!

PR #10685, opened 2022-02-23
UI Simplification: Add Saved Styles into Text Insertion panel

### LINKED ISSUE #10113: UI Simplification: Text element insertion panel - Add Saved Styles
## Feature Description
See [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=11470%3A451937)

- [x] Add Saved Styles to the Text insertion panel (add this behind a feature flag! To be removed within #10609)
- [x] Clicking on a saved style when no element is selected or non-text element is selected: Insert a new text element with the saved styles
- [x] Clicking on a saved style when text element selected: Apply the saved style to the selected text(s)
- [x] Note: Display the Insertion `+` icon on the Saved Style (when hovering or focused) if clicking on it results in adding a new element
- [x] Display the `+` icon in the panel header for saving a style as well. The icon is disabled if a text element is not selected.
- ~Performance: make sure only the Saved Styles panel re-renders when selecting/de-selecting a text element, and make sure it re-renders only if it actually needs to.~ Added a follow-up issue: https://github.com/GoogleForCreators/web-stories-wp/issues/10705
- [x] Add new tests for the Saved Styles.

## Additional Context

Screenshot 2022-02-15 at 17 48 56

miina:
## Summary
Adds Saved Styles section to the Text Insertion panel which allow either inserting a new styled text element or applying style on the text, if selected.

## To-do

- [x] Use the feature flag
- [x] Make style match Figma
- [x] Verify the panel only re-renders when relevant
- [x] Tests

## User-facing changes

When text is selected (apply mode):
Screenshot 2022-02-24 at 14 56 04

When text is not selected (insert mode):
Screenshot 2022-02-24 at 14 56 18

## Testing Instructions

Enable the feature flag for Library: Saved Styles first!

1. Open the Library Text Panel.
2. Add a text, e.g. Title and assign some styles to it -- color, font type, background color, etc.
3. Now, in the Library Saved Styles section, click the `+` icon in the panel header.
4. Verify a new style was added based on the selected Text element.
5. Unselect the text element. 
6. Verify that now the `+` icon is disabled.
7. Hover on the added style, verify that a `+` icon appears.
8. Click on the style button, verify a new text is added with the styles applied.
9. Now add a Label
10. Click on the same style button
11. Verify that instead of adding a new element, the selected Text element gets styled.
12. Test also with multi-selection and also open the Style manager from the "More Styles" link and try the same out there.
13. NB -- all this is from the Left side Text Insertion panel and not the Design panel :)

## Checklist

---

Fixes #10113


packages/story-editor/src/components/library/panes/text/stylePresets/stylePresets.js
```diff
@@ -0,0 +1,253 @@
...
+            ...sizeUpdates,
+          };
+        },
+      });
+    },
+    [updateSelectedElements]
+  );
+
+  const handleApplyStyle = useApplyStyle({ pushUpdate });
+  const { addGlobalPreset } = useAddPreset({ presetType: PRESET_TYPES.STYLE });
```
miina, line 129: This is currently re-rendering due to `useAddPreset` whenever selection changes. `useAddPreset` is dependent on `selectedElements`. Might make sense to split this helper up to have separate hooks for adding a style vs colors -- perhaps this way we can only react to when selection changes from text element to non-text element and vice versa, but not when selection changes between non-text elements.

miina: Added issue: https://github.com/GoogleForCreators/web-stories-wp/issues/10705

packages/story-editor/src/components/library/panes/text/stylePresets/stylePresets.js
```diff
@@ -0,0 +1,251 @@
...
+  width: 100%;
+  svg {
+    transform: rotate(-90deg);
+    height: 32px;
+    width: 32px;
+    margin-left: -4px;
+  }
+`;
+
+const NoStylesWrapper = styled.div`
```
miina: There are some parts of this file that are shared with the Design Panel `stylePresets`, e.g. `NoStylesWrapper`. Not sure if there's a good single place where to keep those instead. Ideas?

barklund: Hm, you did create that new style manager folder. Perhaps some related shared components could go there too?

packages/story-editor/src/components/library/panes/text/stylePresets/stylePresets.js
```diff
@@ -0,0 +1,251 @@
...
+        <>
+          <StylesWrapper ref={stylesRef}>
+            {/* We only show the insertion overlay if we're inserting -> this means selected elements are not texts */}
+            <StyleGroup
+              styles={[...textStyles].reverse().slice(0, 2)}
+              handleClick={handlePresetClick}
+              buttonWidth={STYLE_BUTTON_WIDTH}
+              {...styleItemProps}
+            />
+          </StylesWrapper>
```
barklund, line 189: Should these styles be draggable, as the text presets and text sets are? Is there already a new ticket for that? I feel it should be for consistency, but can also see how that could make things more complex because of the reuse of this section in the design panel.

packages/story-editor/src/components/library/panes/text/stylePresets/stylePresets.js
```diff
@@ -0,0 +1,251 @@
...
+            topOffset={topOffset}
+            isRTL={isRTL}
+            anchor={buttonRef}
+            isOpen={isPopupOpen}
+            placement={PLACEMENT.RIGHT_START}
+            spacing={SPACING}
+            renderContents={() => (
+              <StyleManager
+                styles={textStyles}
+                applyStyle={handlePresetClick}
```
barklund, line 209: I feel like this should also close the fly-out on click. Was there specific UX on this?

packages/story-editor/src/components/library/panes/text/stylePresets/stylePresets.js
```diff
@@ -0,0 +1,251 @@
...
+              {...styleItemProps}
+            />
+          </StylesWrapper>
+          <StyledMoreButton
+            ref={buttonRef}
+            type={BUTTON_TYPES.PLAIN}
+            size={BUTTON_SIZES.SMALL}
+            variant={BUTTON_VARIANTS.RECTANGLE}
+            onClick={() => setIsPopupOpen(true)}
+          >
```
barklund: This button should probably have an aria-haspopup="dialog" and perhaps `aria-expanded` as well.

packages/story-editor/src/components/styleManager/styleManager.js
```diff
@@ -73,7 +73,7 @@ const StyledText = styled(Text).attrs({
   color: ${({ theme }) => theme.colors.fg.primary};
 `;
 
-function StyleManager({ styles, onClose, applyStyle }) {
+function StyleManager({ styles, onClose, applyStyle, ...rest }) {
```
barklund, line 83: This dialog seems to lack focus trapping.

barklund: Also, when you delete an element using the keyboard inside the dialog, focus is lost and returned to the selected element on canvas rather than staying inside the dialog. But both of these issues probably existed before, so could be moved to follow-up issues.

barklund (approved):
This works really well! 🚢 🚀

packages/story-editor/src/components/library/panes/text/stylePresets/stylePresets.js
miina: Hm, on applying it was not supposed to close in the panel, and the idea was to have the same logic here. However, when inserting a new, it should really close, will add it.

packages/story-editor/src/components/library/panes/text/stylePresets/stylePresets.js
miina: I'm not actually sure -- will follow up with UX with this. Most of the styles are displayed only in the Popup, so should the styles be draggable from the popup as well then? This would make things a bit complex indeed.

PR #10709, opened 2022-02-24
Editor: Make mime type config less confusing

### LINKED ISSUE #10592: Editor: Make mime type config less confusing
## Task Description

We currently pass this list of mime/file types to the editor:

https://github.com/GoogleForCreators/web-stories-wp/blob/f130e0cfa2c69ddd07201db7c1b74d58b24bf341/includes/Admin/Editor.php#L363-L369

As we are working on some documentation on using the decoupled story editor, it became more obvious how confusing this config is.

* `allowedMimeTypes`: map of allowed mime types for actual story elements. For example,` allowedMimeTypes.image` might contain `image/svg+xml` or something to say that you can drop SVGs on the canvas
* `allowedImageMimeTypes` : this is used for places in the design/document tab where you can upload images (video poster, story poster, publisher logo). For example, here we would never want to allow uploading SVGs. Hence this separate config.
* `allowedAudioMimeTypes`: similarly, this is used for when you want to add background audio to a page.

In addition to that, `allowedFileTypes` is just a simple array of file types so that in the editor we can say “You are only allowed to upload jpeg, png, gif” etc. We can just get rid of this and use `getExtensionFromMimeType` from the `media` package instead.

`allowedTranscodableMimeTypes` is really pointless too, which is why I opened #10591 for that already.

**To-do**

* Rename `allowedMimeTypes` to somehting more obvious, like for example `allowedElementMimeTypes`
* Maybe rename `allowedImageMimeTypes` and `allowedAudioMimeTypes` to something more obvious too
* Make `allowedFileTypes` obsolete by assembling that on-the-fly in the editor instead, using `getExtensionFromMimeType`

spacedmonkey:
## Summary

Make the handling of mime types and file extensions much simpler, by scaling back. This scaling back includes. 

- Remove the following properties from the config object. 
    - `allowedFileTypes`
    - `allowedImageFileTypes`
    - `allowedImageMimeTypes`
    - `allowedAudioFileTypes`
    - `allowedAudioMimeTypes`
- Remove methods from type class. 
  - `get_allowed_file_types`
  - `get_allowed_image_mime_types`
  - `get_allowed_audio_mime_types`
- Remove the filter `web_stories_allowed_image_mime_types`

Make `get_allowed_mime_types` a single point of truth for all mime types that are handled by the editor / dashboard. 
Do not add image/svg to image types by a new key vector key. This means image mime should all be considered web safe. 

## Relevant Technical Choices

Now all mime types are checked against if WordPress can upload them. That means that aacs are removed, because multisite can not handle this file type. There is a unit test to prove this and a skipped test.

There are some hardcoded mime type mapping, as these mime types are return the wrong file extensions or no file extension at all. This can not avoided sadly. 

Handling of file extensions is now done in javascript and passing in file extensions to the config object. 

## Testing Instructions

1.

## Checklist

---

Fixes #10592


includes/Admin/Editor.php
```diff
@@ -344,11 +344,11 @@ public function get_editor_settings(): array {
...
 		/** This filter is documented in wp-admin/includes/post.php */
-		$show_locked_dialog = apply_filters( 'show_post_locked_dialog', true, $post, $user );
-		$nonce              = wp_create_nonce( 'wp_rest' );
-		$mime_types         = $this->types->get_allowed_mime_types();
-		$image_mime_types   = $this->types->get_allowed_image_mime_types();
-		$audio_mime_types   = $this->types->get_allowed_audio_mime_types();
+		$show_locked_dialog     = apply_filters( 'show_post_locked_dialog', true, $post, $user );
+		$nonce                  = wp_create_nonce( 'wp_rest' );
+		$mime_types             = $this->types->get_allowed_mime_types();
+		$mime_types['ws_image'] = $this->types->get_allowed_image_mime_types();
```
swissspidy: What is `ws_image` supposed to mean?

spacedmonkey: Web safe or web story. Not decided yet. 😄 

ws_image and image mime are different. web site images do not include SVGs.

swissspidy: Ah, so this is referring to things like poster images and publisher logos? What currently `allowedImageMimeTypes` is used for?

I see. So the structure is this:

```
'allowedMimeTypes' = [
  'image' => ..., // From get_allowed_mime_types (), for the image element
  'video' => ..., // From get_allowed_mime_types (), for the video element
  'audio' => ..., // // From get_allowed_mime_types (), for the audio element should we ever have one, but overridden to be used for bg audio
  'ws_image' => ... // For poster images, publisher logo, etc.
]
```

That's not very intuitive at all then and just as confusing as what we currently have.

Why not something like this

```
'allowedMimeTypes' = [
  'elements' => [
    'image' => ...,
    'video' => ...,
    'audio' => ...,
  ],
  'story' => [
    'image' => ...,
    'video' => ...,
    'audio' => ...,
  ]
]
```

Or flatter:

```
'allowedElementMimeTypes' = [
  'image' => ...,
  'video' => ...,
  'audio' => ...,
],
'allowedStoryMimeTypes' = [
  'image' => ...,
  'video' => ...,
  'audio' => ...,
]
```

spacedmonkey: Okay, but that is very long. It also doesn't take into account pages. So media can embed / linked to elements, pages and story levels. So would you want 

```
'allowedMimeTypes' = [
  'elements' => [
    'image' => ...,
    'video' => ...,
    'audio' => ...,
  ],
  'page' => [
    'image' => ...,
    'video' => ...,
    'audio' => ...,
  ]
  'story' => [
    'image' => ...,
    'video' => ...,
    'audio' => ...,
  ]
]
```

swissspidy: I don't think we need 3 different ones 🤔 

Maybe this with different names:

```
'allowedMimeTypes' = [
  'story' => [
    'image' => ...,
    'video' => ...,
    'audio' => ...,
  ],
  'metadata' => [
    'image' => ...,
    'video' => ...,
    'audio' => ...,
  ]
]
```

Edit: hmm well that doesn't work well

swissspidy: So with your suggestion I think right now it would only be like this, right?

```
'allowedMimeTypes' = [
  'elements' => [
    'image' => ...,
    'video' => ...,
  ],
  'story' => [
    'audio' => ..., // Bg audio
  ]
  'metadata' => [
    'image' => ..., // Poster images
  ]
]
```

spacedmonkey: Something like this 

```
'allowedMimeTypes' = [
  'elements' => [
    'image' => ...,
    'video' => ...,
  ],
  'pages' => [
    'audio' => ..., // Bg audio page level
  ]
  'metadata' => [
    'image' => ..., // Poster images
    'audio' => ..., // Bg audio story level
  ]
]
```

spacedmonkey: To be honest, I would go back to 
```
'allowedMimeTypes' = [
    'image' => ...,
    'video' => ...,
    'audio' => ...,
  ],
```

This web safe image thing, is only really an issue if you enable SVG support. We could have something hardcoded to remove svg from allow list, in javascript. it is only for posters and publisher logos. 

We could also do this. 

```
'allowedMimeTypes' = [
    'image' => ...,
    'video' => ...,
    'audio' => ...,
    'vector' => ['image/svg'],
  ],
```

Then we could keep image array, "clean" with only web safe mime types in the images.

swissspidy: There is no need to separate story-level and page-level bg audio.

> This web safe image thing, is only really an issue if you enable SVG support.

SVG would fall under `elements.image`, I don't see why this would need to be a separate thing.

spacedmonkey: I don't think separating into story / metadata / element makes any sense. It makes it confusing, which we are trying to get away from. 

SVGs can only be used in one context, which is image elements, all other contexts, like poster, publisher logo etc, it's does make any sense. By having image/svg in another key, it means for the contexts that support vector images, you could concat image and vector arrays together. For all other contexts, that require images that embeddable / web safe like posters, you would use images.  

We could also just not worry about in this PR and follow up when we officially support svgs.

swissspidy: > SVGs can only be used in one context, which is image elements

Which is why it should not be a separate `vectors` list.

> I don't think separating into story / metadata / element makes any sense. It makes it confusing, which we are trying to get away from.

The current `ws_image` thing is way more confusing.

Another idea off the top of my head:

```
'allowedMimeTypes' = [
  'elements' => [
    'image' => ...,
    'video' => ...,
   // ... whatever new element types we might have in the future, e.g. audio elements
  ],
  'backgroundAudio' => ...,
  'storyPoster' => ...,
  'publisherLogo' => ...
],
```

Similarly, in the dashboard, for consistency:

```
'allowedMimeTypes' = [
  'publisherLogo' => ...
],
```

spacedmonkey:
I have made some tweaks to the Type class, to remove unneeded methods. I would strongly consider removing the `get_allowed_image_mime_types` method and `web_stories_allowed_image_mime_types` as they are no longer really needed. Thoughts @swissspidy ?

swissspidy:
Yes we should remove the unused code 👍

includes/Media/Types.php
```diff
@@ -82,16 +61,22 @@ public function get_file_type_exts( array $mime_types = [] ): array {
...
-			// TODO: Update once audio elements are supported.
-			'audio' => [],
-			'video' => [
+			'audio'  => [
+				'audio/mpeg',
+				'audio/aac',
+				'audio/wav',
+				'audio/ogg',
+			],
+			'text'   => [ 'text/vtt' ],
```
swissspidy: What's the reason for adding this?

`text/vtt` should not be passed via configuration, that one is never going to change. It's why it's not passed via config today.

spacedmonkey: So it runs through the mime type filtering to ensure that this file type can be uploaded. At the moment, these is no check for this and we just guess that vvt files can be uploaded.

swissspidy: You mean if for some reason the WP site blocks VTT uploads?

Related: `text` is very ambiguous. `captions` would make more sense, no? It's not like other `text/` types could ever be in this array.

We should rename this.

tests/phpunit/integration/tests/Media/Types.php
```diff
@@ -40,29 +40,46 @@ public function set_up(): void {
 	 * @covers ::get_allowed_mime_types
 	 */
 	public function test_get_allowed_mime_types(): void {
+		if ( is_multisite() ) {
+			$this->markTestSkipped( 'This should be skipped if is multisite.' );
+		}
```
swissspidy: The proper way to do this is by adding a `@group ms-excluded` annotation. We should update `phpunit-integration-multisite.xml.dist` accordingly.

includes/REST_API/Hotlinking_Controller.php
```diff
@@ -429,7 +429,7 @@ public function get_item_schema(): array {
 			return $this->add_additional_fields_schema( $this->schema );
 		}
 
-		$allowed_mime_types = $this->types->get_allowed_mime_types();
+		$allowed_mime_types = $this->get_media_types();
```
timarney: Wondering about the naming here -- any reason not to say get_mime_types here?

Code says it returns an `Array of supported media types.` and it does .... but the variable says $mime_type --- so feels like we're just switching back and forth between mime types or media types .. bit of metal overhead we could avoid?

includes/Media/Types.php
spacedmonkey: Fixed.

includes/REST_API/Hotlinking_Controller.php
spacedmonkey: Rename to get_allowed_mime_types to be clear.

packages/story-editor/src/app/media/test/useUploadMedia.js
```diff
@@ -71,6 +71,8 @@ jest.mock('../../config', () => ({
     allowedMimeTypes: {
       image: [],
       video: [],
+      text: [],
```
spacedmonkey:
```suggestion
      caption: [],
```

packages/story-editor/src/components/panels/shared/captionsPanelContent.js
```diff
@@ -76,11 +76,15 @@ function CaptionsPanelContent({
...
 }) {
   const {
+    allowedMimeTypes: { caption: allowedTextMimeTypes },
     capabilities: { hasUploadMediaAction },
     MediaUpload,
   } = useConfig();
 
-  if (!hasUploadMediaAction && !tracks.length) {
+  if (
+    (!hasUploadMediaAction && !tracks.length) ||
```
timarney: I didn't dig into the code that calls this etc... but any concern of hitting a `cannot read property length ...`  type of error here?

spacedmonkey: Not an issue, I default the param to an empty array.

timarney (approved):

packages/story-editor/src/components/panels/shared/captionsPanelContent.js
timarney: Unlikely but a null value could still throw 

```
function append(value, array = []) {

  if(!arr.length){
      console.log("on no")
  }

  array.push(value)
  return array
}

append(1)  // [1]
append(2)  // [2], not [1, 2]

append(2, null) // VM121:3 Uncaught ReferenceError: arr is not defined
```

timarney: Could happen with a null ya?  

```
function append(value, myArr = []) {

  if(!myArr.length){
      console.log("on no")
  }

  myArr.push(value)
  return myArr
}

append(1)  // [1]
append(2)  // [2], not [1, 2]
append(2, null) 


// VM214:3 Uncaught TypeError: Cannot read properties of null (reading 'length')
```

includes/REST_API/Hotlinking_Controller.php
```diff
@@ -570,4 +570,23 @@ public function stream_headers( $handle, $header ): int {
...
+	 *
+	 * @since 1.19.0
+	 *
+	 * @return array<string, array> List of allowed mime types.
+	 */
+	protected function get_allowed_mime_types(): array {
+		$mime_type = $this->types->get_allowed_mime_types();
+		// TODO: Update once audio elements are supported.
+		$mime_type['audio'] = [];
+		// TODO: Add support hotlinking vvt files.
```
swissspidy:
```suggestion
		// TODO(#10515): Add support hotlinking vtt files.
```

packages/story-editor/src/components/panels/shared/captionsPanelContent.js
swissspidy: Theoretically yes, but all consumers using `<CaptionsPanelContent>` pass an empty array as a fallback, so this safeguard is not needed.

packages/story-editor/src/components/panels/shared/captionsPanelContent.js
```diff
@@ -76,11 +76,15 @@ function CaptionsPanelContent({
...
+    !allowedTextMimeTypes.length
```
swissspidy: This hides the panel if `allowedTextMimeTypes` is empty but there are tracks.
If there are some tracks we should always display this panel.

So more `(!tracks.length && (!hasUploadMediaAction || !allowedTextMimeTypes.length))`

phpunit-integration-multisite.xml.dist
```diff
@@ -23,6 +23,12 @@
 		</testsuite>
 	</testsuites>
 
+	<groups>
+  		<exclude>
+  			<group>ms-excluded</group>
+  		</exclude>
+  	</groups>
+
```
swissspidy: This is a mix of tabs and spaces. Should only use tabs.

packages/story-editor/src/components/panels/shared/captionsPanelContent.js
spacedmonkey: This if state is to allow those with upload permission and have tracks set to remove caption files. The other statement is so that caption files are not be to uploaded to hide the upload button. This logic would not do this.

phpunit-integration-multisite.xml.dist
spacedmonkey: Fixed

phpunit-integration-multisite.xml.dist
```diff
@@ -23,6 +23,12 @@
 		</testsuite>
 	</testsuites>
 
+  <groups>
+    <exclude>
+      <group>ms-excluded</group>
+    </exclude>
+  </groups>
```
swissspidy: Now they're all spaces, but they should be all tabs.

spacedmonkey:
What is happening with this pr?

swissspidy:
There are some karma test failures after merging `main`.

Also, fixed the spaces vs tabs issue for you.

packages/story-editor/src/components/panels/shared/captionsPanelContent.js
```diff
@@ -76,11 +76,15 @@ function CaptionsPanelContent({
...
   const {
+    allowedMimeTypes: { caption: allowedCaptionMimeTypes },
     capabilities: { hasUploadMediaAction },
     MediaUpload,
   } = useConfig();
 
-  if (!hasUploadMediaAction && !tracks.length) {
+  if (
+    (!hasUploadMediaAction || !tracks.length) &&
+    !allowedCaptionMimeTypes.length
```
spacedmonkey:
```suggestion
    !allowedCaptionMimeTypes?.length
```

spacedmonkey:
Karma tests are now passing. I missed one last place that had not been changed.

swissspidy (approved):

PR #10712, opened 2022-02-24
Video Trimming: Don't change state for other instances 

### LINKED ISSUE #10641: Video Trimming: Don't change state for other instances
## Feature Description

This is a follow-up to #10554 / #10409

When adding the same video multiple times to the canvas and trimming only one of them, each video's "Trim" button changes to "Trimming..." and gets disabled, because technically the underlying resource actually is being trimmed.

However, trimming one video doesn't really affect the other instances of it, so their buttons shouldn't get disabled in that case. This is different from muting or optimizing videos, which always affects all instances.

We need to somehow rework the `isResourceTrimming` check in `VideoOptionsPanel` to make that work.

My initial idea:

1. In `useProcessMedia`/`trimExistingVideo`, don't store `elementId` under `trimData`. Store it as a separate property on the upload queue item
2. In `useMediaUploadQueue`, rename `isResourceTrimming` to `isElementTrimming` and compare `item.elementId === elementId`

timarney:
## Context

Updates to ensure trimming state is only applied to the item selected to be trimmed.

See: https://github.com/GoogleForCreators/web-stories-wp/issues/10641#issue-1145607462 for full details

## Summary

Adds elementId to uploadMedia so we can check that ID against `isElementTrimming` which replaces `isResourceTrimming`

## Testing Instructions

1.

## Checklist

---

Fixes #10641


packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
```diff
@@ -75,23 +75,23 @@ function VideoOptionsPanel({ selectedElements, pushUpdate }) {
...
       isNewResourceMuting,
-      isResourceTrimming,
+      isElementTrimming,
       muteExistingVideo,
     })
   );
   const resource = getCommonValue(selectedElements, 'resource');
-  const { isMuted, id: resourceId = 0 } = resource;
-  const isTrimming = isResourceTrimming(resourceId);
+  const { isMuted, id: resourceId = 0, elementId = 0 } = resource;
```
swissspidy: `elementId` is not part of `resource`; it's he ID of the element itself. So this doesn't work. It will always be `0`.

`const elementId = getCommonValue(selectedElements, 'id');` would get you the ID

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
```diff
@@ -42,6 +42,8 @@ import { ITEM_STATUS } from './constants';
  * @param {boolean} action.payload.muteVideo Whether the video being uploaded should be muted.
  * @param {import('@googleforcreators/media').TrimData} action.payload.trimData Trim data.
  * @param {number} action.payload.originalResourceId Original resource id.
+ * @param {string} args.elementId ID of element on the canvas.
+ * @param action.payload.elementId
```
swissspidy:
```suggestion
 * @param {string} action.payload.elementId ID of element on the canvas.
```

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -234,9 +235,9 @@ function useMediaUploadQueue() {
 
   const trimVideoItem = useCallback(
     async (item) => {
-      const { id, file, additionalData = {}, trimData } = item;
+      const { id, elementId, file, additionalData = {}, trimData } = item;
 
-      startTrimming({ id });
+      startTrimming({ elementId });
```
swissspidy: I don‘t think we wanna change that; this should still operate on the queue item ID.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
```diff
@@ -311,14 +316,14 @@ export function finishMuting(
  * @param {Object} state Current state.
  * @param {Object} action Action object.
  * @param {Object} action.payload Action payload.
- * @param {string} action.payload.id Item ID.
+ * @param {string} action.payload.elementId Element ID.
  * @return {Object} New state
  */
-export function startTrimming(state, { payload: { id } }) {
+export function startTrimming(state, { payload: { elementId } }) {
```
swissspidy: As per my other comment, we should not need to change this reducer.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
timarney: Okay -- just noticed your comments so I'll make the other updates + revert this

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
timarney: reverted here - https://github.com/GoogleForCreators/web-stories-wp/pull/10712/commits/144962ae691692e2ae1f05db9ea8dd408c925b50

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
timarney: reverted here - https://github.com/GoogleForCreators/web-stories-wp/pull/10712/commits/42cb79b0040d3c77165c0879780527b852f5ddc5

packages/story-editor/src/app/media/utils/useMediaUploadQueue/reducer.js
timarney: Reverted - https://github.com/GoogleForCreators/web-stories-wp/pull/10712#pullrequestreview-893099431

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
timarney: Update here - https://github.com/GoogleForCreators/web-stories-wp/pull/10712/commits/7f576fd0dddf012865501affb514c3c5e7107e44

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
```diff
@@ -75,23 +75,24 @@ function VideoOptionsPanel({ selectedElements, pushUpdate }) {
...
-      isResourceTrimming,
+      isElementTrimming,
       muteExistingVideo,
     })
   );
   const resource = getCommonValue(selectedElements, 'resource');
   const { isMuted, id: resourceId = 0 } = resource;
-  const isTrimming = isResourceTrimming(resourceId);
+  const elementId = getCommonValue(selectedElements, 'id');
+  const isTrimming = isElementTrimming(elementId);
```
spacedmonkey: Multiple elements can be selected. So we should only do `isElementTrimming` if one element is selected. So something like this. 

```
const isTrimming = selectedElements.length === 1 ? isElementTrimming(elementId) : false;
```

timarney:
@swissspidy @spacedmonkey --- think I'm missing a couple of pieces here to finish this one up.  If you test adding two of the same video + trimming the `trim` button for the second is disabled still.  

Is that possibly because `isNewResourceProcessing` needs to be updated or is it something else?

```javascript
const isNewResourceProcessing = (resourceId) =>
      state.queue.some(
        (item) =>
          [
            ITEM_STATUS.TRANSCODING,
            ITEM_STATUS.TRIMMING,
            ITEM_STATUS.MUTING,
          ].includes(item.state) && item.originalResourceId === resourceId
      );
```

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -832,17 +830,17 @@ function useMediaUploadQueue() {
...
-        (item) =>
-          item.state === ITEM_STATUS.TRIMMING &&
-          item.originalResourceId === resourceId
-      );
+    const isElementTrimming = (elementId) =>
+      state.queue.some((item) => {
+        return (
+          item.state === ITEM_STATUS.TRIMMING && item.elementId === elementId
+        );
+      });
```
swissspidy: Why not keeping the short return?

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
swissspidy: I don't see why this is needed 🤔 

We do `const elementId = getCommonValue(selectedElements, 'id');` above

So if there is a multi-selection, `elementId` will be `MULTIPLE_VALUE` (which is the string `((MULTIPLE))`)

`isElementTrimming('((MULTIPLE))')` does not cause any harm

spacedmonkey: It seems a little wasteful there, there is no need to `isElementTrimming('((MULTIPLE))')`. It will do a lookup in state when it not needed. The `selectedElements.length === 1` in an inexpensive way of avoiding that.

swissspidy:
@timarney The button is disabled based on the outcome of `canTranscodeResource()`:

https://github.com/GoogleForCreators/web-stories-wp/blob/928311c162f7d7f0454364755e689ef099badb33/packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js#L103

https://github.com/GoogleForCreators/web-stories-wp/blob/928311c162f7d7f0454364755e689ef099badb33/packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js#L158

It is also used to disable the Mute button.

That function returns `false` if the resource is _somewhere_ in the queue already.

I need to wrap my head around this first on Monday, but I think we either need to add a second `elementId` param to `canTranscodeResource`, or we add a dedicated `canTrimElement` selector.

swissspidy:
So _ideally_ the Mute & Trim buttons are disabled if the current _resource_ is currently being muted/optimized (because these operations currently affect all elements with this resource) or if the current _element_ is being trimmed (but not another element with the same resource).

Now, i am still not sure how best to achieve that.

The `canTranscodeResource()` check is too broad. As mentioned above, that function simply checks if the resource is _somewhere_ in the queue already, without knowing what _type_ of upload it is (trimming, muting, regular upload, etc.).

Changing that would be quite an effort and thus better handled in a follow-up ticket as a nice-to-have. In other words: **it's ok for the second element's trim button to still be disabled, as long as it isn't trimmed**

timarney:
> Changing that would be quite an effort and thus better handled in a follow-up ticket as a nice-to-have. In other words: it's ok for the second element's trim button to still be disabled, as long as it isn't trimmed

With that said -- I think the trimming part of this PR is working as expected will move it to ready to review unless you see other items that need to be done here.

spacedmonkey (approved):
Works well in my testing.

swissspidy (approved):

PR #10717, opened 2022-02-25
Packages: Reorganise utility functions in more packages

### LINKED ISSUE #9639: Decoupling: Package for common utility functions

mohdsayed:
## Context

See https://github.com/GoogleForCreators/web-stories-wp/issues/9639

## Testing Instructions

1.

## Checklist

---

Fixes https://github.com/GoogleForCreators/web-stories-wp/issues/9639


spacedmonkey:
I love this! Specially the url package.

packages/dom/src/index.js
```diff
@@ -13,18 +13,7 @@
...
-    document.addEventListener('DOMContentLoaded', callback);
-  } else {
-    callback();
-  }
-}
-
-export default domReady;
+export { default as stripHTML } from './stripHTML';
+export { default as escapeHTML } from './escapeHTML';
+export { default as getCaretCharacterOffsetWithin } from './getCaretCharacterOffsetWithin';
```
swissspidy: This one might be a candidate for the rich-text package since it's used for text editing

packages/story-editor/src/app/font/actions/useLoadFontFiles.js
```diff
@@ -19,12 +19,12 @@
...
 import { useCallback } from '@googleforcreators/react';
 import { getGoogleFontURL } from '@googleforcreators/fonts';
+import { loadStylesheet } from '@googleforcreators/dom';
 
 /**
  * Internal dependencies
  */
 import cleanForSlug from '../../../utils/cleanForSlug';
-import loadStylesheet from '../../../utils/loadStylesheet';
 import { ensureFontLoaded, loadInlineStylesheet } from '../utils';
```
swissspidy: `loadInlineStylesheet` could also be something for the `dom` package?

packages/units/src/range.js
```diff
@@ -13,35 +13,9 @@
...
 
-/**
- * Takes a value and a range in the form of a tuple and clamps
- * the value to that range.
- *
- * @param {number} v - value to be clamped
- * @param {[number, number]} range - range in the form of a tuple
- * @return {number} - number within range
- */
-export const clamp = (v, range) => {
```
swissspidy, line 38: Note that this function had a different signature than the new one below

packages/units/src/range.js
```diff
@@ -13,35 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
-/**
- * Linear interpolation function.
```
swissspidy: Removed by accident?

packages/story-editor/src/app/font/actions/useLoadFontFiles.js
mohdsayed: Didn't want to add `fonts` package dependency to `dom`. But yes it can be removed by a little refactoring. Done in https://github.com/GoogleForCreators/web-stories-wp/pull/10717/commits/be17eff2a932febd7b1659a1c4a740e099507414

packages/units/src/range.js
mohdsayed: There were two versions of `clamp` defined in three different packages, one can accept an array ( `[number, number]` ) and other accepts object (`{MIN: number, MAX: number}`)

https://github.com/GoogleForCreators/web-stories-wp/blob/0f12659b6a277de64da7f65922aa5eb666a6614c/packages/story-editor/src/utils/clamp.js#L17

https://github.com/GoogleForCreators/web-stories-wp/blob/e00e50da7a02d956b993f4038ca1c15d46eda3ef/packages/dashboard/src/utils/clamp.js#L17-L22

https://github.com/GoogleForCreators/web-stories-wp/blob/e00e50da7a02d956b993f4038ca1c15d46eda3ef/packages/animation/src/utils/range.js#L38-L43

So I just combined those two versions in one and moved to `units`. Does it bother you that `range` can accept both object and array?

packages/units/src/range.js
mohdsayed: Thanks added back

packages/units/src/range.js
swissspidy: > Does it bother you that `range` can accept both object and array?

A little bit, yes :-) Can we settle on one? Perhaps the object, as it's more explicit

packages/units/src/range.js
```diff
@@ -42,26 +42,24 @@ export const lerp = (progress, range) =>
...
-export const clamp = (value, range) => {
-  const lowerBound = Array.isArray(range)
-    ? Math.min(range[0], range[1])
-    : range.MIN;
-  const upperBound = Array.isArray(range)
-    ? Math.max(range[0], range[1])
-    : range.MAX;
+export const clamp = (value, { MIN, MAX }) => {
+  const lowerBound = Math.min(MIN, MAX);
+  const upperBound = Math.max(MIN, MAX);
```
mohdsayed: This looks weird with `MIN` and `MAX` object but don't know why some tests are doing inverse value testing here and if removing it would create any side effect https://github.com/GoogleForCreators/web-stories-wp/blob/pacakges/utilities/packages/units/src/test/range.js and they fail without it.

packages/units/src/range.js
mohdsayed: https://github.com/GoogleForCreators/web-stories-wp/pull/10717/commits/58bc976f8afd04a765180d0cb592abe278dff745

packages/units/src/range.js
swissspidy: Agreed, that is weird. Doing it for the array notation makes sense, but not with `MIN` and `MAX` object.
So I'd say we can remove those tests

packages/units/src/range.js
```diff
@@ -57,8 +42,21 @@ export const clamp = (v, range) => {
...
 };
+
+/**
+ * Takes a value and a range in the form of a tuple and clamps the value to that range.
+ *
+ * @param {number} value - value to be clamped
+ * @param {Object} range - range in the form of a tuple.
+ * @param range.MIN
+ * @param range.MAX
+ * @return {number} - number within range
```
swissspidy:
```suggestion
 * @param {number} value Value to be clamped.
 * @param {Object} range Range.
 * @param {number} range.MIN Minimum value.
 * @param {number} range.MAX Maximum value.
 * @return {number} Number within range.
```

packages/wp-story-editor/src/api/story.js
```diff
@@ -28,11 +29,7 @@ import apiFetch from '@wordpress/api-fetch';
  * Internal dependencies
  */
 import { STORY_EMBED, STORY_FIELDS } from './constants';
-import {
-  base64Encode,
-  transformStoryResponse,
-  snakeToCamelCaseObjectKeys,
-} from './utils';
+import { base64Encode, transformStoryResponse } from './utils';
```
spacedmonkey, line 32: Should we move `base64Encode`, `transformStoryResponse` to `@web-stories-wp/wp-utils`?

spacedmonkey (approved):
Approved with one qusestion.

packages/wp-story-editor/src/api/story.js
swissspidy: They are only used by `wp-story-editor` and not shared with `wp-dashboard`. So no need to move to a shared package. That'd just make the `wp-dashboard` bundle bigger for no benefit.

swissspidy (approved):

packages/wp-story-editor/src/api/story.js
spacedmonkey: Would it make the bundle bigger?

swissspidy: Well the tree shaking might be stripping things out again, but I'm not betting on that with webpack v4 :)

Still, moving stuff that isn't actually shared to a shared package does not achieve anything

Issue #10111, opened 2022-01-06
UI Simplification: Move the right sidebar functionality to the left sidebar as tabs


miina:
## Feature Description
See [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=11122%3A447794)

- [x] Add additional tab row to the left sidebar to the top with the functionality of the right sidebar. The new tabs are: Insert (previously the whole left sidebar), Style (previously the Design tab), Document (previously the right sidebar Document tab).
- [x] Adjust the whole grid of the editor to match Figma and use the space of the right sidebar for the workspace. Move the Page and the top buttons accordingly.
- [x] Ensure the highlighting system and the quick-action menu still work as expected.

For now, don't change anything else in the content of the panels, even if it is so in Figma -- these will be followed up in separate issues.

## Additional Context
<img width="613" alt="Screenshot 2022-01-06 at 13 14 57" src="https://user-images.githubusercontent.com/3294597/148374634-14dca286-a051-47a1-8f2a-3073e1169df1.png">

agingoldseco:
The bottom bar is correct—I'll go through and fix those in the doc. However, there is more to think through to get combining the right and left panels just right and I'll be working on that for a lot of this week. I would start on the more ready tickets first and I'll keep you posted on progress.

miina:
@agingoldseco Alright, thanks for the update, I understood from Slack conversation that not much should change anymore. Please let me know once the final version is ready.

bmattb:
@merapi Infra pod has flagged that E2E tests will require updating in support of this UI change.

miina:
Closing as this part got fixed (there's an epic now combining all the Sidebar issues: #10817)

felipebochehin87:
Tested against https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/, using PR #10931.

Regressed all functionalities, side bar works fine RTL, LTR, Safari, Chrome, Firefox. Text, videos, images, animations, subtitles, audio, shapes, editing, publishing, preview, etc...

BrittanyIRL:
Thanks @felipebochehin87  I know that was _alot_

PR #10719, opened 2022-02-25
UI Simplification: Allow applying Presets

### LINKED ISSUE #10609: UI Simplification: Text element insertion panel - Allow applying Presets
## Feature Description
In addition to adding an element by clicking on a text Preset, it should also be possible to apply it, if a text element is selected.

Note: this feature should be released together with #10113 and not separately!

- [x] Clicking on a Preset when no element is selected or non-text element is selected: Insert a new text element
- [x] Clicking on a Preset when text element selected: Apply the preset to the selected text(s)
- [x] Note: Do not display the Insertion `+` icon on the Preset if clicking on it results in applying style. The insertion icon should be displayed only if we'll add a new element when clicking.
- [ ] Performance: make sure only the Presets re-render when selecting/de-selecting a text element, and make sure it re-renders only if it actually needs to.
- [x] Update tests (quite a few will break due to not being able to insert two elements in a row anymore)
- [x] Remove feature flag for Saved Styles added in #10113 

Note: for consecutively adding text elements, the user now has to de-select text first and then click to add text again.

miina:
## Summary
Allows applying presets when text element is selected (or multiple text elements).

## To-do

- [x] Remove the Saved Styles feature flag once #10685 gets merged.
- [x] Update tests
- [x] Add new tests

## Testing Instructions

1. Add a text element, for example Paragraph. Verify that `+` icon is displayed when hovering over the Paragraph preset.
2. Now keep the text element selected and hover over Title 1. Verify the `+` icon is not displayed anymore.
3. Click on the preset. Verify that instead of adding a new one, the preset is applied to the existing element instead -- e.g. the font size changes, etc.
4. Try the same with other presets / multiple selected texts -- when text is selected, the Preset is applied; when something else is selected (or nothing), a new element is inserted.

No

No

No

## Checklist

---

Fixes #10609


packages/story-editor/src/components/canvas/karma/quickActions.karma.js
```diff
@@ -26,7 +26,7 @@ import { ACTIONS } from '../../../app/highlights';
 import { Fixture } from '../../../karma';
 import useInsertElement from '../useInsertElement';
 
-describe('Quick Actions integration', () => {
+xdescribe('Quick Actions integration', () => {
```
miina: Note: skipping the tests is temporary in this PR, will add these back in once `main` passes -- it was necessary for detecting which tests failed due to this PR and which not.

packages/story-editor/src/components/library/panes/text/fontPreview.js
```diff
@@ -111,6 +117,15 @@ function FontPreview({ title, element, insertPreset, getPosition, index }) {
 
   const { calculateAccessibleTextColors } = usePageAsCanvas();
 
+  const { isText, updateSelectedElements } = useStory(({ state, actions }) => {
+    const isText =
+      state.selectedElements && areAllType('text', state.selectedElements);
```
barklund: Nit: This technically needs a Boolean wrapper to be consistently a bool.

```js
    const isText =
      Boolean(state.selectedElements && areAllType('text', state.selectedElements));
```

packages/story-editor/src/components/library/panes/text/fontPreview.js
```diff
@@ -161,8 +176,30 @@ function FontPreview({ title, element, insertPreset, getPosition, index }) {
     pageCanvasData,
     versionNumber,
   ]);
-
   const onClick = useCallback(() => {
+    // If we have only text(s) selected, we apply the preset instead of inserting.
+    if (isText) {
+      updateSelectedElements({
```
barklund, line 198: Why is the element being moved as well when applying a text preset? Should the element stay in place?

preset-move

And should things like italic, color, underline, all caps, and text alignment also be "reset" rather than just font family, font size, and font weight?

packages/story-editor/src/components/panels/design/textStyle/karma/stylePresets.karma.js
```diff
@@ -65,8 +65,8 @@ describe('Panel: Style Presets', () => {
       });
 
       // Verify that no styles are added currently.
-      const noStylesText = fixture.screen.getByText('No Styles Saved');
-      expect(noStylesText).toBeDefined();
+      const noStylesText = fixture.screen.getAllByText('No Styles Saved');
+      expect(noStylesText.length).toBeGreaterThanOrEqual(1);
```
barklund: Can there be more than one?

barklund: Ah, it is from the two different panels - because the text exists in both the text insertion panel and the design panel?

Perhaps it should have a test for each separately, to confirm that it exists in both panels individually. That will also make the test (correctly) fail once the panels are merged in #10111.

barklund (commented):
Seems to work fine, though I'm not convinced that all the right properties of the saved elements are applied to the existing texts.

packages/story-editor/src/components/library/panes/text/fontPreview.js
miina: No, the element should definitely not move around!

> And should things like italic, color, underline, all caps, and text alignment also be "reset" rather than just font family, font size, and font weight

Not certain about it -- will check with UX.

includes/Experiments.php
```diff
@@ -394,17 +394,6 @@ public function get_experiments(): array {
 				'group'       => 'editor',
 				'default'     => true,
 			],
-			/**
-			 * Author: @miina
-			 * Issue: #10113
-			 * Creation date: 2022-02-22
-			 */
-			[
```
swissspidy, line 392: We usually first enable a feature by default before removing the flag code completely. But I suppose for this small change here it's fine.

miina: I know, I wasn't planning to add it behind a feature flag originally at all since it's quite a small change but since UX preferred having it released together with this feature (Applying presets), then added it anyway for ensuring these two get released together.

packages/story-editor/src/components/panels/design/textStyle/karma/stylePresets.karma.js
miina: This test should actually really test for the one in the Library. Adjusted the test to do that.

packages/story-editor/src/components/library/panes/text/fontPreview.js
miina: For visibility: looks like at least the colors, font family, and text-alignment should be kept as it was in the original element.
Waiting for confirmation from @agingoldseco, see the [Slack discussion](https://xwp.slack.com/archives/C01ESBB8ZN1/p1646210559348009).

EDIT: As of current information, only the font color should be kept from the inline styles.

merapi:
It might be connected to the existing issue (shouldn't set the padding to Mixed when the color is changed, right?):

https://user-images.githubusercontent.com/21036224/156668910-a5b0ad01-2898-4bf3-abd5-c4fddfac201f.mov

miina:
> It might be connected to the existing issue (shouldn't set the padding to Mixed when the color is changed, right?):

Seems like a separate issue -- I've seen odd padding values as well (`null`) but have seen it quite randomly without understanding where it came exactly. I'll create a follow-up issue for this as it doesn't seem to be related to Applying presets.

miina:
@barklund @merapi This PR is ready for final review now -- got confirmation from UX that the text color should be kept but most of the other Text Styles not. The others would be overwritten in a similar way as the Saved Styles overwrite most of the props (but not 100% all) (yep, it could potentially cause confusion).

packages/story-editor/src/components/library/panes/text/fontPreview.js
merapi: Maybe we can just use `isText: areAllType('text', state.selectedElements),` because in the worst case `selectedElements = STABLE_ARRAY` and `Boolean(STABLE_ARRAY || any array) === true`.

merapi:
When using the keyboard there is no visible focus when you are in the "apply preset" mode.

I guess it should look like the focus on the Saved Styles:

image

miina:
> When using the keyboard there is no visible focus when you are in the "apply preset" mode.

Good catch! Fixed.

packages/story-editor/src/components/library/panes/text/fontPreview.js
miina: This extra `state.selectedElements &&` is for checking if the `state.selectedElements` is set at all, if it's `null` then `areAllType` would error since it's checking for `selectedElements.length` inside. Thoughts?

merapi: My point is: is it possible that `selectedElements` can be `null`?
https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/story-editor/src/app/story/storyProvider.js#L96

miina: Perhaps not! For some reason I remember it erroring without checking first but doesn't seem to happen currently. Changing.

merapi (approved):
This will be very confusing for long term users but the PR itself is 🏅

PR #10720, opened 2022-02-25
Media: Reload video on canvas after replacing

### LINKED ISSUE #10678: Video not reloading on canvas after replacing
## Bug Description

Originally found by @timarney:

When replacing a video on the canvas using the "Replace Media" quick action menu item, the video is not actually reloaded on the canvas.

[This StackOverflow answer](https://stackoverflow.com/questions/41303012/updating-source-url-on-html5-video-with-react) has a good explanation of why that happens.

> The video initially doesn't change because in essence you're only modifying the `<source>` element and React understands that `<video>` should remain unchanged, so it does not update it on the DOM and doesn't trigger a new load event for that source. A new load event should be triggered for `<video>`.

The quick and easy solution is to use `key={url}` to trick React into updating the video on the DOM. Like so:

In `packages/story-editor/src/elements/video/display.js`:

```diff
        <Video
          id={`video-${id}`}
+         // Force React to update the video in the DOM, causing it to properly reload if the URL changes.
+         // See https://github.com/GoogleForCreators/web-stories-wp/issues/10678
+         key={url}
          poster={poster || resource.poster}
```

Using a ref + effect as suggested on that SO post seems overkill, because initially we only want to load metadata, not the whole video.

**Note:** Replacing a video using drag and drop is not affected by this, because the video element's drop handler calls `videoEl.load()`:

https://github.com/GoogleForCreators/web-stories-wp/blob/d31f8dde36c8f50cd703dca596a61df3b5e6f7d9/packages/story-editor/src/elements/video/onDropHandler.js#L17-L26

## Expected Behaviour

The video should be reloaded on the canvas after using the "Replace Media" quick action

## Steps to Reproduce

1. Insert video on canvas
2. Use "Replace Media" quick action to replace video with another one
3. Still see old video on canvas

## Screenshots

https://user-images.githubusercontent.com/841956/155220645-2e51ee7f-be31-465d-9f98-ac4ed5547d13.mp4

## Additional Context

- Plugin Version: 1.18.0
- WordPress Version: 5.9
- Operating System: macOS
- Browser: Chrome

timarney:
## Context

When replacing a video on the canvas using the "Replace Media" quick action menu item, the video is not actually reloaded on the canvas.

See: https://github.com/GoogleForCreators/web-stories-wp/issues/10678#issue-1147381459

## Summary

Adds a [key](https://reactjs.org/docs/lists-and-keys.html#keys) attribute to the Video to ensure React get the update

## User-facing changes

Bug fix

## Testing Instructions

1) Insert video on canvas
2) Use "Replace Media" quick action to replace video with another one
3) Still see new video on canvas

No

No

No

## Checklist

---

Fixes #10678

### DIFF
--- a/packages/story-editor/src/elements/video/display.js
+++ b/packages/story-editor/src/elements/video/display.js
@@ -107,6 +107,9 @@ function VideoDisplay({ previewMode, box: { width, height }, element }) {
         // eslint-disable-next-line styled-components-a11y/media-has-caption,jsx-a11y/media-has-caption -- False positive.
         <Video
           id={`video-${id}`}
+          // Force React to update the video in the DOM, causing it to properly reload if the URL changes.
+          // See https://github.com/GoogleForCreators/web-stories-wp/issues/10678
+          key={url}
           poster={poster || resource.poster}
           style={style}
           {...videoProps}


packages/story-editor/src/elements/video/display.js
```diff
@@ -107,6 +107,9 @@ function VideoDisplay({ previewMode, box: { width, height }, element }) {
         // eslint-disable-next-line styled-components-a11y/media-has-caption,jsx-a11y/media-has-caption -- False positive.
         <Video
           id={`video-${id}`}
+          // Force React to update the video in the DOM, causing it to properly reload if the URL changes.
+          // See https://github.com/GoogleForCreators/web-stories-wp/issues/10678
+          key={url}
```
spacedmonkey, line 112: There a worry that two video with same url would be on a canvas at the same time?

spacedmonkey (approved):
Approved with one question.

packages/story-editor/src/elements/video/display.js
swissspidy: That doesn‘t matter and has no effect on this. That‘s not how the key prop works.

PR #10735, opened 2022-02-28
Revert "Move to `react-photo-album`"

swissspidy:
Reverts GoogleForCreators/web-stories-wp#10463 until we can fix the issues with drag & drop from the media library

PR #10737, opened 2022-02-28
Media: Use `react-photo-album` for media library

### LINKED ISSUE #10444: Replace `react-photo-gallery`

swissspidy:
## Context

This re-introduces #10463 which was reverted in #10735 due to failing tests.

See those PRs for full context.

Previously [react-photo-gallery](https://www.npmjs.com/package/react-photo-gallery) was used to display media elements in the library pane, it has not seen any change since 1 July 2019. Therefore it was stuck at React version `^16.8.0` which will cause 2 versions of React to be installed if are packages published.

## Summary

Replaces `react-photo-gallery` with `react-photo-album` in the media library.

## To-do

* Fix tests

## User-facing changes

None

## Testing Instructions

1. Media library should still work as expected, including dragging & dropping items to the canvas

No

No

No

## Checklist

---

Fixes #10444


spacedmonkey:
When this PR, can you ensure that highlight the difference from the original PR so this PR can be reviewed.

swissspidy:
There are no differences right now,  any new changes can be seen in the commit messages and will of course be reflected in the PR description, as per usual.

swissspidy:
See https://github.com/GoogleForCreators/web-stories-wp/pull/10463#issuecomment-1056356686

miina:
@swissspidy I think the relevant tests should be fixed now.

There's a failing e2e run but seems unrelated: https://github.com/GoogleForCreators/web-stories-wp/runs/5408325743?check_suite_focus=true

I reviewed the snapshots and things seem generally fine but there are some slight changes -- might be good if you could double check to see if these seem expected.

Also, please review the new file changes -- it's your PR so I can't request a review from you :)

packages/e2e-tests/src/specs/editor/media/insertWebMVideo.js
```diff
@@ -72,6 +72,7 @@ describe('Inserting WebM Video', () => {
     uploadedFiles.push(fileName);
 
     await page.waitForSelector('[data-testid="mediaElement-video"]');
+    await page.waitForTimeout(100);
```
miina: Note: it seems like without a small delay the correct video won't be available on time.
It's a bit hard to see what exactly is happening there since in interactive mode everything works as expected but I guess the new video will be otherwise added to the library after the click has happened already and that messes things up.
Not sure if there's a better way to fix these tests. Feel free to amend to the PR if you have ideas.

miina: One alternative could be clicking in the corner of the element to avoid the Insertion menu (and insert whatever element gets clicked on first) but not sure if that's the right approach either -- if we're waiting for the uploaded video.

spacedmonkey (approved):
Works well in testing.

packages/e2e-tests/src/specs/editor/media/insertWebMVideo.js
swissspidy: Ideally we'd use a more targeted `waitFor...` specifically for this new element.

The `await page.waitForSelector('[data-testid="mediaElement-video"]')`above is a bit pointless, as it is too generic and just waits for any video to exist in the library, which is always the case.

I'll try to find a more specific selector.

PR #10753, opened 2022-02-28
Shapes: avoid excess styled-components classes

### LINKED ISSUE #10372: Resolve the ‘Over 200 classes generated” warning
See https://github.com/google/web-stories-wp/issues/5797 
Example in logs [line 1160](https://github.com/google/web-stories-wp/runs/4766928054?check_suite_focus=true#step:6:1160)

samwhale:
## Summary

Pass in computed element styles as an attribute to the `ShapePreview`.

## Relevant Technical Choices

Removes the necessity to have Styled Components interpolate the same styles on every render.

💭 ~Still confused about _why_ the component is re-rendering so many times in the karma tests. Question:
**After a test suite is completed, does the completed test env get torn down?**~

~It looks like the location is changed but nothing is removed from the root element~: https://github.com/GoogleForCreators/web-stories-wp/blob/13fb5c0512df2129eac594552c0b4c7ea8ccc953/packages/story-editor/src/karma/fixture/fixture.js#L222-L225

[Seems like this is coming from styled components cache not being reset between tests](https://github.com/GoogleForCreators/web-stories-wp/pull/10753#discussion_r816345221).

---

Also:
🗒️ the `text` and `shape` elements both use this. I think it'd be good to come up with the same update for both.

## User-facing changes

None

## Testing Instructions

Verify that this warning is missing from the logs:

```bash
"Over 200 classes were generated for component icon__ShapePreview with the id of "icon__ShapePreview-sc-8694nc-1".
Consider using the attrs method, together with a style object for frequently changed styles.
Example:
  const Component = styled.div.attrs(props => ({
    style: {
      background: props.background,
    },
  }))`width: 100%;`

  <Component />", source: http://localhost:9876/context.js (265)
```

no

no

no

## Checklist

---

Fixes #10372

### DIFF
--- a/packages/story-editor/src/elements/shape/icon.js
+++ b/packages/story-editor/src/elements/shape/icon.js
@@ -18,12 +18,12 @@
  * External dependencies
  */
 import styled from 'styled-components';
+import { generatePatternStyles } from '@googleforcreators/patterns';
 
 /**
  * Internal dependencies
  */
 import { getMaskByType } from '../../masks';
-import { elementWithBackgroundColor } from '../shared';
 import StoryPropTypes from '../../types';
 import { useStory } from '../../app';
 
@@ -39,12 +39,15 @@ const Container = styled.div`
   background-color: ${({ theme }) => theme.colors.opacity.black10};
 `;
 
-const ShapePreview = styled.div`
+const ShapePreview = styled.div.attrs(({ backgroundColor }) => ({
+  // https://stackoverflow.com/a/58095676
+  // Removes console warning where class was re-generated over 200 times.
+  style: generatePatternStyles(backgroundColor),
+}))`
   width: 100%;
   height: 100%;
   margin: 1px;
   clip-path: ${({ maskId }) => `url(#${maskId})`};
-  ${elementWithBackgroundColor}
 `;
 
 /* 
--- a/packages/story-editor/src/elements/shape/icon.js
+++ b/packages/story-editor/src/elements/shape/icon.js
@@ -40,8 +40,8 @@ const Container = styled.div`
 `;
 
 const ShapePreview = styled.div.attrs(({ backgroundColor }) => ({
-  // https://stackoverflow.com/a/58095676
-  // Removes console warning where class was re-generated over 200 times.
+  // Prevents `ShapePreview` class from being generated with each new layer.
+  // https://styled-components.com/docs/faqs#when-to-use-attrs
   style: generatePatternStyles(backgroundColor),
 }))`
   width: 100%;


maxyinger (approved):
nice find!

packages/story-editor/src/elements/shape/icon.js
```diff
@@ -39,12 +39,15 @@ const Container = styled.div`
   background-color: ${({ theme }) => theme.colors.opacity.black10};
 `;
 
-const ShapePreview = styled.div`
+const ShapePreview = styled.div.attrs(({ backgroundColor }) => ({
+  // https://stackoverflow.com/a/58095676
+  // Removes console warning where class was re-generated over 200 times.
```
swissspidy: Not sure how valuable that comment is. Especially since this is not the only place where we use `attrs()`.

I'd prefer a link to https://github.com/GoogleForCreators/web-stories-wp/pull/10753 and/or https://styled-components.com/docs/faqs#when-to-use-attrs

packages/story-editor/src/elements/shape/icon.js
```diff
@@ -39,12 +39,15 @@ const Container = styled.div`
...
+  style: generatePatternStyles(backgroundColor),
+}))`
   width: 100%;
   height: 100%;
   margin: 1px;
   clip-path: ${({ maskId }) => `url(#${maskId})`};
-  ${elementWithBackgroundColor}
```
swissspidy, line 47: This is not the only usage of this variable. Why does only this one need to be changed / why are others not affected?

samwhale: Yeah, mentioned that in the description. We could fix it in all places, but before we do that I think it'd be best to verify that this is the change we want to make.

This happens in a few places, and I'm not sure why it's only happening to `ShapePreview`. I think this points to a worse problem where the `ShapePreview` is being re-rendered over 200 times.

There are some unanswered questions here still. So just wanted to open this up to see if others had some ideas of what could be happening here too 😄

samwhale: 💭 Unsure yet of where styled-components is adding the class to the document, but If the document styles aren’t cleared, then styled components will keep adding a new class every time the editor is rendered (as well as a new layer)

May be that after 200 layers are added, styled components can tell that they all have very similar ids + styles but they're all different class definitions, so it logs that error

samwhale: Update: verified that after the `ShapePreview` component has been rendered in the test suite 200 times, this error gets logged.

---


1. Every time a new test is started in the editor, a layer is is rendered in the layer panel with the `ShapePreview` element. This will create a `ShapePreview` element.
    - Note: this may not be 200 tests because some tests render more than one layer with a `ShapePreview`
2. Styled components keeps track of the amount of classes that are generated from each component. However, that cache ([seems like it's a cache](https://github.com/styled-components/styled-components/blob/30dab74acedfd26d227eebccdcd18c92a1b3bd9b/packages/styled-components/src/utils/createWarnTooManyClasses.ts#L12)) is not reset between tests.
3. After 200 layers with shape elements have been rendered, the warning gets logged to the console.

This happened only for shape elements because the background layer is a shape and that number gets to 200 quicker than text layers do. With all tests that are run, there are not 200 text layers rendered yet. Once that threshold is passed, the warning should be visible again.

---

This warning can show up if a user creates 200 layers of shapes or text too. So updating those styles to be passed in as attributes would fix the problem. A new class would only be created for each shape type rather than each individual shape + text layer.

200-shapes

BrookeGraham: Nice.

BrookeGraham (approved):
Great find. That makes perfect sense.

swissspidy (approved):

PR #10754, opened 2022-02-28
Editor: Added new border width and color controls to design menu

### LINKED ISSUE #10625: UI Simplification: Floating Menu: Border width and color
## Feature Description

Make the border width and color elements of the floating menu functional.

* Design: [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=11394%3A457049)
* Menus: Shape + Image + Video
* Optional rendering: Border width part of the element only renders if the element in question has a uniform border width (non-rectangular shapes can only ever have uniform border width). The border color element always renders.
* Notes: There's a dash between the two elements if both render.

## Additional Context

border-width-color

barklund:
## Context

This adds functional inputs to the design menu for border width and/or border color.

## Relevant Technical Choices

- Some optional rendering of elements which corresponds to the border panel.
  - Note that an element can never not have uniform border and not have any border at the same time, as having any border be non-zero would result in the element having at least one border.
  - Also note that non-multi-border-elements do not allow opacity in the color picker.
- Generalized the numeric input for design menu from the border-radius control.

## To-do

- [x] Default (black) color is not set when a border is set originally. Where is that done in the existing border panel?
- [x] Karma tests

## Non user-facing changes

| Allows multi-border | Has uniform borders | Has any border | Screenshot |
|-|-|-|-|
| ✅  | ✅  | ❌   | Screen Shot 2022-02-28 at 16 21 55 |
| ✅  | ✅  | ✅  | Screen Shot 2022-02-28 at 16 22 00 |
| ✅  | ❌  | ✅  | Screen Shot 2022-02-28 at 16 22 03 |
| ❌  | N/A | ❌  | Screen Shot 2022-02-28 at 16 22 06 |
| ❌  | N/A | ✅  | Screen Shot 2022-02-28 at 16 22 11 |

## Testing Instructions

## Checklist

---

Fixes #10625


packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
```diff
@@ -19,19 +19,110 @@
...
+
+  // If both controls are displayed, also add a dash between them
+  const hasBoth = hasUniformBorder && hasBorderWidth;
+
+  // NB: Note that it can never be the case, that neither input is shown, so at least
+  // one of them will always render.
+
+  const handleWidthChange = (value) =>
+    updateSelectedElements({
+      properties: ({ border: oldBorder }) => ({
```
miina, line 83: Hm, the default black border in the Design panel comes from the `DEFAULT_BORDER`, so it looks like `oldBorder` doesn't include the color here for some reason? When getting the initial border value, in theory, it should be black by default when using `DEFAULT_BORDER`.

Maybe `const { border = DEFAULT_BORDER, mask } = useProperties(['border', 'mask']);` above is not getting `border` value back as `undefined` / `null` and thus the `DEFAULT_BORDER` is not assigned correctly?

packages/story-editor/src/karma/fixture/containers/designMenu.js
```diff
@@ -43,4 +43,20 @@ export class DesignMenu extends Container {
...
+  get borderWidth() {
+    return this.queryByRole('textbox', { name: 'Border width' });
+  }
+
+  get borderColor() {
+    const region = this.queryByRole('region', {
+      name: /Color input: Border color/,
+    });
+    if (!region) {
+      return null;
```
miina, line 56: Usually we're testing `expect(() => ...).toThrow()` to verify something is not in the DOM, maybe keep it the same for consistency here and just get the element but not return `null` if not found?

packages/story-editor/src/components/floatingMenu/elements/shared/color.js
```diff
@@ -0,0 +1,63 @@
...
+  );
+}
+
+function FloatingColor(props) {
+  const { hasInputs, hasEyedropper } = props;
+  const width = getWidth(hasInputs, hasEyedropper);
+  return (
+    <Color
+      width={width}
+      maxHeight={362}
```
miina: Nit: To a constant?

packages/story-editor/src/components/floatingMenu/elements/shared/color.js
```diff
@@ -0,0 +1,63 @@
...
+      pickerPlacement={PLACEMENT.TOP_END}
+      isMinimal
```
miina: Hm, I saw it before already but forgot to ask -- what does `isMinimal` mean? E.g. if we don't use it, what happens then? It's not obvious from just the name.

packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
```diff
@@ -19,19 +19,110 @@
...
+    border.left === border.top &&
+    border.left === border.bottom;
+
+  // We only allow editing border color, if at least one border has a non-zero width
+  const hasBorderWidth = getHasBorderWidth(border);
+
+  // If both controls are displayed, also add a dash between them
+  const hasBoth = hasUniformBorder && hasBorderWidth;
+
+  // NB: Note that it can never be the case, that neither input is shown, so at least
```
miina: Is this comment a leftover from somewhere? Not sure about which part it is here 🤔

packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
```diff
@@ -19,19 +19,110 @@
...
+  const { border = DEFAULT_BORDER, mask } = useProperties(['border', 'mask']);
+  const updateSelectedElements = useStory(
+    (state) => state.actions.updateSelectedElements
+  );
+
+  // Only multi-border elements support border opacity
+  const canHaveBorderOpacity = canSupportMultiBorder({ mask });
+
+  // We only allow editing the current border width, if all borders are identical
+  const hasUniformBorder =
```
miina, line 66: Q: Wouldn't `lockedWidth` already show that?

miina (commented):
Works great generally!

There's currently the bug of initial black border not being set and this resulting in a transparent border as the default. Saw it as a todo in the PR, too!

packages/story-editor/src/karma/fixture/containers/designMenu.js
barklund: I used `query*` specifically to avoid them throwing, because the elements are not supposed to always be on screen.

packages/story-editor/src/components/floatingMenu/elements/shared/color.js
barklund: The whole item is compressed in size when displayed with inputs (and optionally opacity too). Compare the menu with and without `isMinimal` below (when `hasInputs` is true):

| `isMinimal` | Screenshot |
|-|-|
| ✅  | Screen Shot 2022-03-02 at 15 24 21 |
| ❌  | Screen Shot 2022-03-02 at 15 19 23 |

Alternate names are welcome. We could simply name it `isInDesignMenu` as this should always be true if in the design menu and always false if not.

packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
barklund: Even with border ratio lock off, the borders can still be identical. I felt that we could still display the input as long as the border widths were identical for easy access.

packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
barklund: Nope, it refers to the two above bools, `hasUniformBorder` and `hasBorderWidth`. I have made it more explicit in fc71c7b847.

packages/story-editor/src/components/floatingMenu/elements/shared/color.js
barklund: Well, it's only ever used this once and it's a pure magic number, so no real difference either way. But moved it in 1004d6ea66.

packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
barklund: The `border` variable is not used inside the `handleWidthChange` logic. But I can explicitly set it - done in 8cf3b8eb92.

packages/design-system/src/icons/index.js
```diff
@@ -90,6 +91,7 @@ export { default as FloppyDisk } from './floppy_disk.svg';
 export { default as Gear } from './gear.svg';
 export { default as GearWithGauge } from './gear_with_gauge.svg';
 export { default as Keyboard } from './keyboard.svg';
+export { default as Launch } from './launch.svg';
```
barklund, line 94: Just moved this to its proper alphabetical place. This list is manually sorted.

packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
miina: Ah, good point!

packages/story-editor/src/components/floatingMenu/elements/shared/color.js
miina: > We could simply name it `isInDesignMenu` as this should always be true if in the design menu and always false if not.

Yep, something that makes it more clear when to use it would be great, `isInDesignMenu` sgtm.

miina (approved):

packages/story-editor/src/components/floatingMenu/elements/shared/color.js
barklund: Renamed in 001b60bedf

PR #10704, opened 2022-02-24
Editor: New design for color picker

barklund:
## Context

This updates the color picker to the new slimmer, leaner design.

This also adds a ton of new configuration options for the color input and the color picker, that ar required for the color inputs in the new design menu.

And while working on these elements, I made a number of other optimizations as well (including optimizing CSS and DOM structure).

**Note:** Screenshots are incoming (once I can run a virtual machine again when I get a proper charger for my laptop), so please just clone and check for yourself. The storybook stories are the best way to play around with this (Stories Editor > Components > Form > Color > *)

## Summary

* A lot of the design decisions (spacing, etc) has been made on the fly, because the full color picker was not designed by UX. Figma files only existed for the basic color picker, not the custom color picker.

## Relevant technical choices

* Updated swatches to only render 1 CSS class in total, not 1 per swatch (by moving the actual swatch pattern styles to `style` rather than the class)

## To-do

* [x] Add (comprehensive!) storybook stories for color input in both design panel (left and right) as well as new design menu.
* [x] Add before and after screenshots
* [x] Fix tests

## User-facing changes

The color picker as opened from the design menu is a lot slimmer in general now.

| Picker | Saved colors? | Before | After |
|-|-|-|-|
| Basic | Yes | Screen Shot 2022-02-24 at 16 34 27 | Screen Shot 2022-02-24 at 16 32 17 |
| Basic | No | Screen Shot 2022-02-24 at 16 34 35 | Screen Shot 2022-02-24 at 16 33 06 |
| Custom, solid | Yes | Screen Shot 2022-02-24 at 16 34 31 | Screen Shot 2022-02-24 at 16 33 21 |
| Custom, solid | No | Screen Shot 2022-02-24 at 16 34 40 | Screen Shot 2022-02-24 at 16 33 15 |
| Custom, gradient | Yes | Screen Shot 2022-02-24 at 16 34 20 | Screen Shot 2022-02-24 at 16 32 05 |

## Non-user-facing changes

Color picker as opened from the new design menu looks like the following (note this is still behind feature flag).

| Picker | Eyedropper | Gradient | Screenshot |
|-|-|-|-|
| Basic | No | No | Screen Shot 2022-02-24 at 16 43 48 |
| Custom | No | No | Screen Shot 2022-02-24 at 16 43 56 |
| Basic | Yes | No | Screen Shot 2022-02-24 at 16 44 37 |
| Custom | Yes | No | Screen Shot 2022-02-24 at 16 44 41 |
[...]

Issue #10698, opened 2022-02-23
Keyboard focus doesn't shift focus correctly for Empty State contextual menu


bmattb:
## Bug Description

Originally reported by @BrittanyIRL 

Keyboard focus doesn’t shift to the selected pane when I click on Browse Page Templates but it does for other panes. 

Additionally @spacedmonkey found that "if I select page template once. Change tab and then why to go to page template again, nothing happens" and @BrookeGraham found that "similar with moving away from images tab and then selecting Choose Image or Video from right click menu" 
![image10](https://user-images.githubusercontent.com/58738530/155392150-0b5a0b1c-e6ba-44c7-bf70-538edefe8b1c.gif)

@merapi flagged that these are all likely the same issue.

## Steps to Reproduce

1. Open a new Story
2. right click within the canvas of the first page to trigger the "Empty State" contextual menu
3. Select "Browse Page Templates" and note that focus isn't changed as expected.

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

miina:
@BrittanyIRL Since this is related to the general focus management of elements/canvas (#10850) that you'll be looking at, if I understood correctly, is this an issue that would make sense for Pea pod?

BrittanyIRL:
> @BrittanyIRL Since this is related to the general focus management of elements/canvas (#10850) that you'll be looking at, if I understood correctly, is this an issue that would make sense for Pea pod?

i think this falls into that, absolutely.

BrittanyIRL:
Likely connected to #11291 - none of these are working on repeat actions. they used to, so something regressed here in the past few releases.

Issue #10626, opened 2022-02-18
UI Simplification: Slim color picker


barklund:
## Feature Description

This concerns updating the color picker to the new and slimmer design.

Design: [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=11394%3A457271)
Notes: This does not include design for the custom color picker state but it should be similarly sized and contents should be scaled to fit

## Additional information

<img width="604" alt="Screen Shot 2022-02-18 at 18 36 35" src="https://user-images.githubusercontent.com/637548/154774974-a9631196-1fba-431c-8dce-07a2b0803dd7.png">

felipebochehin87:
Tested against https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/, using PR #10704.

Regressed color picker for elements - all working as expected

Issue #10486, opened 2022-02-06
E2E Tests: broken Percy snapshots


swissspidy:
## Task Description

For some reason the Percy snapshot for the `Story Background Audio` e2e test is always broken, see screenshots below.

It would be nice to find out why & fix it.

**Note:** testing locally is preferred so that we don't wastefully upload tons of snapshots to Percy.

![Screenshot 2022-02-07 at 00 14 28](https://user-images.githubusercontent.com/841956/152705651-ea007f84-4bcf-43b5-9374-3e90292102ee.png)

![73eef55ce8d431f35a87748e8fd83479fae959803fc32dbcef3915cce42493ab](https://user-images.githubusercontent.com/841956/152705644-0326bf71-bbcd-4e1f-836b-185369d8b552.png)

miina:
Another seemingly broken snapshot, for "Get Started Story":
<img width="402" alt="Screenshot 2022-02-09 at 12 54 28" src="https://user-images.githubusercontent.com/3294597/153184693-d4ec63f5-0cec-4e0c-bf31-18212995de4f.png">

timarney:
I can replicate the "broken snapshot" screen running Percy locally (not via CI) --- so I can dig deeper on that part to start.

https://github.com/GoogleForCreators/web-stories-wp/issues/10486#issuecomment-1033631848

For "Story Background Audio" I get this for an output

![4ebe5dfed03a3d482e98cdff3193ed92c96ace96ea3fa3d6ebaeb69a7ad9801d](https://user-images.githubusercontent.com/62242/154378039-f8a60088-3b52-4d78-849e-847c294cd374.png)


---
Same test screens outside of the editor render properly.

![46482f530f7d5f70480c2e150ae752eccc827d02f58022dbd50aa0ac4b7fa705](https://user-images.githubusercontent.com/62242/154383166-b23728a5-62f8-40da-8a95-dd31f6bcb863.png)

---


---

As for the "unstyled" Admin --- unclear as to why that's happening.  I contacted Percy support to see if that turns up anything.

I did find some screens online that seems to be similar.  This appears to be related to load-styles.php possibly around `CONCATENATE_SCRIPTS` issues or Cross origin related (http vs https)

https://wordpress.org/support/topic/admin-pages-unstyled-and-unformatted/


[...]

PR #10760, opened 2022-03-01
UI Simplification: Move the right sidebar functionality to the left sidebar as tabs

### LINKED ISSUE #10111: UI Simplification: Move the right sidebar functionality to the left sidebar as tabs

merapi:
## New naming conventions (general, for all)

image

We will now have just one panel on the left side, so we might want to change the naming of a few components (we at least need to change the structure a bit, especially in the Fixture).
Prior to this design change, the left side panel was called `Library` but I think now, something generic like `Side Panel/Side Bar` can be a better name for a place that holds the Insert, Style, and Document tabs.
`<Inspector />` will now be under the `Style` tab, so not sure if it should change the name or just the location (in Fixture).

I'd love to hear your naming suggestions.
Some examples:
`SidePanel: Insert, Inspector, Document`
`SidePanel/SideTabs: InsertTab, StyleTab, DocumentTab`

## Summary/Relevant Technical Choices (this PR)

My initial solution was to move the right sidebar functionality to the left sidebar as tabs (as in the title), but then I've found out that it's better (less code, fewer conflicts) to move the left sidebar to the right sidebar and then move the whole thing to the left.

Fixing tests: I went through all of them with `fit` so I didn't have to wait for the whole suite to fail each time, most of the time it was about switching tabs, mostly to `Design` tab, sometimes back to `Insert` tab. More complex cases involved switching back the focus from the panel area to canvas/element.
Note: tabs sometimes change automatically e.g. by using quick action buttons that use highlight.

I've had one unexpected issue with many [failing](https://user-images.githubusercontent.com/21036224/156177155-06d09d8e-4347-4d9a-811e-bd8e8cb27260.png) tests that took me a while to debug, that was caused by IDE auto imported `import {takeSnapshot} from '@web-stories-wp/e2e-test-utils';` (still don't know why it was like this).

## To-do here or in another PR.

- [ ] Change naming for elements in the new layout (library, inspector etc.)
- [ ] Change naming for Fixture parts (designPanel etc.)
- [ ] Fix color picker positioning
- [ ] Fix date picker positioning

## Testing Instructions

1.

## Checklist

---


merapi:
@BrittanyIRL @samwhale @littlemilkstudio flagging you for visibility and suggestions about the new naming conventions.

packages/story-editor/src/components/form/color/colorInput.js
```diff
@@ -250,7 +250,8 @@ const ColorInput = forwardRef(function ColorInput(
         </Tooltip>
       )}
       <Popup
-        isRTL={isRTL}
+        // Temp fix making the color picker usable, refactor after the new color picker changes (#10626)?
+        isRTL={!isRTL}
```
barklund: #10626 was solved in #10704 and has been merged.

packages/story-editor/src/components/inspector/inspectorProvider.js
```diff
@@ -144,8 +147,13 @@ function InspectorProvider({ inspectorTabs, children }) {
     data: {
       tabs: [
         {
-          id: DESIGN,
-          title: __('Design', 'web-stories'),
+          id: INSERT,
+          title: __('Insert', 'web-stories'),
+          Pane: Library,
```
barklund, line 152: The biggest consequence of moving the library in as a pane in the inspector as-is, has to do with the width. The library was wider than the inspector, and the new library doesn't adjust well to the narrower width:

| Area | Screenshot |
|-|-|
| Saved styles | Screen Shot 2022-03-01 at 13 53 37 |
| Text sets | Screen Shot 2022-03-01 at 13 53 49 |
| Page templates | Screen Shot 2022-03-01 at 13 54 06 |
 
 The alternative is of course that we make the old inspector wider, but that would make the design of all the panels off. However, I do think that this whole area should return to the width that the library had before to make the library elements look correct.
 
 We can also fix all the panels in follow-up PR's (which should generally be pretty quick to do).

packages/story-editor/src/components/inspector/inspectorProvider.js
```diff
@@ -36,11 +36,12 @@ import { useFeature } from 'flagged';
...
 import { useHighlights } from '../../app/highlights';
-import { DOCUMENT, DESIGN, PUBLISH_MODAL_DOCUMENT } from './constants';
+import { DOCUMENT, STYLE, PUBLISH_MODAL_DOCUMENT, INSERT } from './constants';
 import Context from './context';
 import DesignInspector from './design';
+import Library from '../library';
 
-const INSPECTOR_TAB_IDS = new Set([DOCUMENT, DESIGN]);
+const INSPECTOR_TAB_IDS = new Set([INSERT, DOCUMENT, STYLE]);
 function InspectorProvider({ inspectorTabs, children }) {
```
barklund, line 45: Regarding naming, I think that we could change the name to just _Sidebar_, and that would work fine.

The only reason I invented the two names, _Library_, and _Inspector_, in the first place was just to have some identifiers. Neither name is used in the UI as far as I remember, so they are only internal anyway.

We did use _Sidebar_ at one point for the flyout menus (such as color picker), but I believe they were all changed to use our popup/modal API, so _Sidebar_ should be a "free" term again now. Note that "inspector" is used as a pre- and postfix for many elements.

We could then also rename the Library to Insert, but I don't think that's necessary. I feel like the name still reflects its purpose.

I do think all the renaming should be moved to a new ticket. Just to keep this PR as clean as it is right now (which is impressively low footprint considering the huge change in UX).

packages/story-editor/src/components/layout/index.js
```diff
@@ -60,11 +60,10 @@ const Editor = withOverlay(styled.section.attrs({
   width: 100%;
   display: grid;
   grid:
-    'lib   canv        insp' 1fr
-    'lib   supplementary insp' auto /
-    minmax(${LIBRARY_MIN_WIDTH}px, ${LIBRARY_MAX_WIDTH}px)
-    minmax(${CANVAS_MIN_WIDTH}px, 1fr)
-    minmax(${INSPECTOR_MIN_WIDTH}px, ${INSPECTOR_MAX_WIDTH}px);
+    'insp   canv          ' 1fr
```
barklund, line 56: There's a number of lint warnings in this file. Is husky not working currently? I actually feel like I haven't seen it running for a while.

packages/story-editor/src/app/highlights/provider.js
```diff
@@ -58,10 +58,11 @@ function HighlightsProvider({ children }) {
       }
 
       if (highlight) {
-        const { tab, ...highlightState } = STATES[highlight];
+        const { tab, subTab, ...highlightState } = STATES[highlight];
```
barklund: Ugh, this is getting ugly, but I guess there's no good way around it. "subTab" doesn't sound good though, IMO. But I have no better ideas. 🤷

packages/story-editor/src/components/inspector/inspectorProvider.js
```diff
@@ -36,11 +36,12 @@ import { useFeature } from 'flagged';
...
 import { useStory } from '../../app/story';
 import { useHighlights } from '../../app/highlights';
-import { DOCUMENT, DESIGN, PUBLISH_MODAL_DOCUMENT } from './constants';
+import { DOCUMENT, STYLE, PUBLISH_MODAL_DOCUMENT, INSERT } from './constants';
 import Context from './context';
 import DesignInspector from './design';
+import Library from '../library';
 
-const INSPECTOR_TAB_IDS = new Set([DOCUMENT, DESIGN]);
+const INSPECTOR_TAB_IDS = new Set([INSERT, DOCUMENT, STYLE]);
```
barklund, line 44: Should this whole thing be behind a feature flag for now? I feel like a lot of stuff is being changed without being thoroughly tested and I also feel like this is only the first of many PR's implementing this change (follow-up refactors and design rework incoming), so perhaps it should be hidden? I know it's a lot of extra work, but maybe not too much extra work?

Well, perhaps except for the highlight stuff, that might be a bit tricky.

All the karma tests can be updated to use the feature flag, so they don't have to be reverted again.

packages/story-editor/src/components/inspector/inspectorProvider.js
```diff
@@ -36,11 +36,12 @@ import { useFeature } from 'flagged';
 import { useAPI } from '../../app/api';
 import { useStory } from '../../app/story';
 import { useHighlights } from '../../app/highlights';
-import { DOCUMENT, DESIGN, PUBLISH_MODAL_DOCUMENT } from './constants';
+import { DOCUMENT, STYLE, PUBLISH_MODAL_DOCUMENT, INSERT } from './constants';
 import Context from './context';
 import DesignInspector from './design';
+import Library from '../library';
```
barklund: When things are renamed, folders should also be moved around accordingly. Either the library should go inside the inspector (to be renamed) folder or both document and design panels should move out. I feel like the latter is the proper approach. But that's work for a future refactor.

packages/story-editor/src/components/inspector/inspectorProvider.js
```diff
@@ -36,11 +36,12 @@ import { useFeature } from 'flagged';
 import { useAPI } from '../../app/api';
 import { useStory } from '../../app/story';
 import { useHighlights } from '../../app/highlights';
-import { DOCUMENT, DESIGN, PUBLISH_MODAL_DOCUMENT } from './constants';
+import { DOCUMENT, STYLE, PUBLISH_MODAL_DOCUMENT, INSERT } from './constants';
 import Context from './context';
 import DesignInspector from './design';
```
barklund, line 42: This should probably also be renamed _Style_ now?

barklund (commented):
This is really good work. Amazing that you complete this refactor with such little effort - the largest part of the PR are all the test fixes!

I have some ideas for renaming, posted inline where applicable for better threaded conversations.

Most importantly, I think we need to feature flag this work though, as I think it needs quite a bit of follow-up work to be mergeable into user-facing terrain.

packages/story-editor/src/components/layout/index.js
swissspidy: We updated it recently; you might need to run `npx husky-init` once

packages/story-editor/src/app/highlights/provider.js
merapi: I guess we can do something like:
```
tab: `${INSERT}/${PAGE_TEMPLATES.id}`
```
without `subTab`, would that be any better?

BrittanyIRL: what about `section` instead of `subTab`?

packages/story-editor/src/components/inspector/inspectorProvider.js
BrittanyIRL: big fan of `Sidebar` - i _still_ have to remind myself which is the library and which is the inspector sometimes.

packages/story-editor/src/components/inspector/inspectorProvider.js
BrittanyIRL: the highlighting update could probably be handled conditionally, if `subTab` or whatever it gets called is present then use that, if not fallback to the existing structure. I think that would make feature flagging doable? 

It might also just be something that has to look a little weird for like a week, if that's a thing everyone is ok with - pea pod can help out too to make it move faster. I agree with @barklund that this whole thing needs to be tested thoroughly.  Might be worth a slack thread to figure out a plan for the whole update since it's such a lift.

packages/story-editor/src/components/inspector/inspectorProvider.js
miina: In Figma, the contents of the panels are adjusted to be wider, so it makes sense to me to make it wider right away already, and then later in a follow-up adjust the design panels.

packages/story-editor/src/components/inspector/inspectorProvider.js
miina: > We could then also rename the Library to Insert, but I don't think that's necessary. I feel like the name still reflects its purpose.

Agreed that we could just keep Library as it is, could even keep Design as Design IMO unless Style is generally thought to be a better name. E.g. if the visual name changes again in the future, we'd probably keep the naming in the code the same and only change the visual text, so if Design still reflects what that tab does then seems ok to keep that.

miina:
To echo the others -- awesome that you managed to make such a change with so few file changes!

A few notes from testing (didn't test thoroughly, just noticed when clicking through it):
- When tabbing out of the library, the focus seems to get lost somewhere? (Previously went visibly to the canvas/workspace). E.g. tab to Shapes -> Stickers -> tab again -> not sure where the focus is.
- The Layer Panel is now visible only when the Style tab is opened -- I guess moving this would be for a follow-up issue? LMK, will create one.

packages/story-editor/src/components/inspector/inspectorProvider.js
BrittanyIRL: +1 to making this refactor and doing it as its own ticket.

BrittanyIRL:
> * When tabbing out of the library, the focus seems to get lost somewhere? (Previously went visibly to the canvas/workspace). E.g. tab to Shapes -> Stickers -> tab again -> not sure where the focus is.
> * The Layer Panel is now visible only when the Style tab is opened -- I guess moving this would be for a follow-up issue? LMK, will create one.

These both feel like good issues on their own, especially the focus bit - it feels like that's gone a bit haywire as of the last few releases and some attention to it would be nice. It's also good isolated work, so pea pod could help out without really risking stepping on any toes after this initial move is done.

miina:
> > When tabbing out of the library, the focus seems to get lost somewhere? (Previously went visibly to the canvas/workspace). E.g. tab to Shapes -> Stickers -> tab again -> not sure where the focus is.
> > The Layer Panel is now visible only when the Style tab is opened -- I guess moving this would be for a follow-up issue? LMK, will create one.

> These both feel like good issues on their own, especially the focus bit - it feels like that's gone a bit haywire as of the last few releases and some attention to it would be nice. It's also good isolated work, so pea pod could help out without really risking stepping on any toes after this initial move is done.

Sounds good to me! The initial goal was to get the Panel move done by the end of the next Sprint to get it into the next release, so help with these is definitely appreciated. @merapi is offline today, I'll check with him first just in case as well in case he had already started some work with those.

samwhale:
Incredible how little code was required here! I don't have anything to add that others haven't addressed. The idea of  the `Sidebar` name sounds good to me


I saw one thing that would be good to add a min width to the editor/sidebar so that tabs aren't cut off.

packages/story-editor/src/app/highlights/provider.js
merapi: `section` is ok, it might be a little confusing that `section` leads to a pane with multiple `section` elements (there is no 1:1 connection between them), but I guess that's fine.

image

packages/story-editor/src/components/inspector/inspectorProvider.js
merapi: > In Figma, the contents of the panels are adjusted to be wider, so it makes sense to me to make it wider right away already, and then later in a follow-up adjust the design panels.

Yep, the new design is `360px` (like the old Library), I've changed `INSPECTOR_MAX_WIDTH` to `360` and now it looks good (even better than in Figma, because Figma has old design pasted into a wider layout without re-adjusting).

Figma:
image
 
Editor:
image

packages/story-editor/src/components/inspector/inspectorProvider.js
merapi: I'm +1 with Miina [here](https://github.com/GoogleForCreators/web-stories-wp/pull/10760#discussion_r817626896), I guess you too? So leaving it as-is for now.

merapi:
> I saw one thing that would be good to add a min width to the editor/sidebar so that tabs aren't cut off.

I guess we will just set the min and max to 360 - I will confirm that with Design Pod.

Prior to this PR the min-width was 220 so the library tabs were already cut off:

image

packages/story-editor/src/app/highlights/provider.js
merapi: Done in https://github.com/GoogleForCreators/web-stories-wp/pull/10760/commits/872b439e308ef93caad29d4a3d70b90a64ab21cb

packages/story-editor/src/components/inspector/inspectorProvider.js
merapi: Done in https://github.com/GoogleForCreators/web-stories-wp/pull/10760/commits/f584130488c37cf937629c95b3c09247d531519b

packages/story-editor/src/components/layout/index.js
merapi: Done in https://github.com/GoogleForCreators/web-stories-wp/pull/10760/commits/b85b713e56b824ebc646ad5e7f7ae12935cde0eb

merapi:
> 

I will work on the Layers panel next, but we can use your help @BrittanyIRL  to fix the focus issues in general, we currently have a pretty big issue here: https://github.com/GoogleForCreators/web-stories-wp/issues/10793 not mentioning other minor issues like https://github.com/GoogleForCreators/web-stories-wp/issues/10792.

packages/story-editor/src/components/form/color/colorInput.js
BrittanyIRL: Should this have it's own follow up issue since 10626 is solved and merged?

packages/story-editor/src/components/library/panes/pageTemplates/karma/savedPageTemplates.karma.js
```diff
@@ -118,6 +118,7 @@ describe('CUJ: Page Templates: Custom Saved Templates', () => {
       await fixture.events.hover(
         fixture.editor.library.pageTemplatesPane.pageTemplates[0]
       );
+      await fixture.events.sleep(100);
```
spacedmonkey, line 121: Why is this needed?

packages/story-editor/src/components/inspector/inspectorProvider.js
BrittanyIRL: Just to keep conversations current, discussion on feature flag was moved to [Slack](https://xwp.slack.com/archives/C0135V8LYJE/p1646317678228379?thread_ts=1646235979.968759&cid=C0135V8LYJE) and decided against because it adds unnecessary time to this effort. The plan is to keep this branch current with main and merge this and the floating menus together.

packages/story-editor/src/components/library/panes/pageTemplates/karma/savedPageTemplates.karma.js
BrittanyIRL, line 121: Just curious, given the recent efforts to clean up sleeps, is this absolutely necessary to sleep? What changed in the flow that's making it newly needed?

packages/story-editor/src/components/panels/design/textStyle/stylePresets.js
```diff
@@ -125,7 +125,8 @@ function PresetPanel({ pushUpdate }) {
           </MoreButton>
           <Popup
             topOffset={topOffset}
-            isRTL={isRTL}
+            // Temp fix, should work the same way as the new color picker (after ongoing refactor is done).
```
BrittanyIRL: Same here, i think this as a new issue and the issue number linked here would be helpful.

spacedmonkey (changes requested):
I don't see an tweaks to you e2e tests. These will need quite a bit of work. 
Of the top of head, the following tests would need tweaking. 

https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/e2e-tests/src/specs/editor/pageTemplates.js
https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/e2e-tests/src/specs/editor/templates.js
https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/e2e-tests/src/specs/editor/media/insert3PMedia.js
https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/e2e-tests/src/specs/editor/media/insertMovVideo.js
https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/e2e-tests/src/specs/editor/media/insertWebMVideo.js
https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/e2e-tests/src/specs/editor/media/svg.js

I would recommend creating some helper util functions to switch to the correct tab.

BrittanyIRL (commented):
Pulled and ran locally this am and this is working quite well! 

You mentioned pea pod helping out with the following: 
- Focus issues (all lumped together): #10793 
- Keyboard navigation issues that existed prior to this update: #10792  

The outstanding work I see here, just to summarize: 
- Positioning of popups (inclusive of tooltips that are now cut off) 
- e2e test updates 
- layers panel (this should be visible on all tabs now, yes?) 

I agree with @samwhale 's point about adding a minwidth to avoid cutting off tabs. I am aware that as it stands tabs get cut off, but this redesign is a really great opportunity to improve that. 

If peapod is to help with the following, I just want to make sure that this is in a stable place before we do anything and also to confirm the flow for these updates. 

In slack, ya'll talked about keeping this branch current with main, so do you want us to do PRs to this branch for the above? 

Is it just the accessibility issues that you want any assistance on? As these are previously existing issues, to me, they could be done after this whole thing gets merged just as well. I just want to make sure because the plan to merge this and the floating menu stuff all at once feels like _alot_ of really big code change.

miina:
@BrittanyIRL 
> I just want to make sure because the plan to merge this and the floating menu stuff all at once feels like alot of really big code change.

The floating menu code is all behind a feature flag, so just removing the feature flag (or setting it enabled by default) would be done together with the panel move. It was a request from UX to release these two changes to public at the same time due to some feedback from users against the panel move -- the goal of releasing these together is resolving some of that feedback via the floating menu. Thoughts?

> layers panel (this should be visible on all tabs now, yes?)

Layer panel will actually be moved to the bottom right corner of the workspace: [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=11394%3A456057)
Since there are some Figma mocks which have it in the top bar as well then it was [confirmed with UX](https://github.com/GoogleForCreators/web-stories-wp/issues/10111#issuecomment-1008977594) just in case that it should go to the bottom indeed.

packages/story-editor/src/components/library/panes/pageTemplates/karma/savedPageTemplates.karma.js
miina: If we're waiting for the delete button to be available, perhaps we could add `waitFor` for `fixture.editor.library.pageTemplatesPane.deleteBtnByIndex(0)` where we throw an error explicitly if the button node is not available (as it's now done elsewhere instead of sleeping).

BrittanyIRL:
Thanks, @miina ! 

> The floating menu code is all behind a feature flag, so just removing the feature flag (or setting it enabled by default) would be done together with the panel move. It was a request from UX to release these two changes to public at the same time due to some feedback from users against the panel move -- the goal of releasing these together is resolving some of that feedback via the floating menu. Thoughts?

Totally makes sense to release together, just wanted to make sure i was clear on the overall plan. I still think some of the issues that have been discussed here (ie the focus issues) aren't really a blocker for this ui simplification to get merged since they were broken before these changes. So what I'm really getting at is how long will this branch need to be maintained before it can go to main to help us all move on the blocking issues before nonblocking. 

> Layer panel will actually be moved to the bottom right corner of the workspace

Thanks for clarifying and the link! We haven't really seen many of these design specs.

miina:
@BrittanyIRL 

> I still think some of the issues that have been discussed here (ie the focus issues) aren't really a blocker for this ui simplification to get merged since they were broken before these changes.

Yep, existing issues are definitely not blockers, agreed :)

> When tabbing out of the library, the focus seems to get lost somewhere? (Previously went visibly to the canvas/workspace). E.g. tab to Shapes -> Stickers -> tab again -> not sure where the focus is.

^ As far as I saw, this was one of the focus issues that came with the move, it would be great to fix this before the merge, it might also be related to some existing issues potentially, e.g. #10698 and #10793

Once this PR is final, and the Layer Panel is done + the new focus issue is resolved (probably easiest to base those off from this branch), I don't see any additional blocker issues for merging the PR.

spacedmonkey:
It is also worth noting that all the percy screenshots for the e2e tests seem super broken. I would take a look into those.

merapi:
Okay, I think we have a plan, how about merging this PR to a new branch called `sidebar` and then base all follow-ups on that? We already have quite a few comments here and it's getting messy @miina @BrittanyIRL 

Follow-ups will be (as @BrittanyIRL mentioned), at least:
> Positioning of popups (inclusive of tooltips that are now cut off)
e2e test updates
layers panel

I think any renaming should be done as a last step, which could make the merging to main easier (renaming will happen immediately after merging).

merapi:
> It is also worth noting that all the percy screenshots for the e2e tests seem super broken. I would take a look into those.

Yep, most of them will just have to be overwritten.

miina:
> Okay, I think we have a plan, how about merging this PR to a new branch called `sidebar` and then base all follow-ups on that? We already have quite a few comments here and it's getting messy @miina @BrittanyIRL

You mean because `sidebar` would be a "clean" branch to merge everything together? Sure, sgtm if it makes it easier!

packages/story-editor/src/components/panels/design/textStyle/stylePresets.js
merapi: 👍🏻  
Done in 215edae2d8a2a4e939bd2a5f0a62015d59d8c9c2

packages/story-editor/src/components/form/color/colorInput.js
merapi: 👍🏻  
Done in 215edae2d8a2a4e939bd2a5f0a62015d59d8c9c2

packages/story-editor/src/components/library/panes/pageTemplates/karma/savedPageTemplates.karma.js
merapi: It seems that the button is not instantly actionable, if you lower that to 50 it will still fail, that's weird, given the fact that there is no fade-in animation etc.

swissspidy:
@spacedmonkey Are you just referring to the known issue with Percy snapshots tracked at #10486?

packages/story-editor/src/components/library/panes/pageTemplates/karma/savedPageTemplates.karma.js
merapi: Hmm, it looks like the delete button is available, but not actionable, even if I replace `sleep` with 
```
await waitFor(() => {
    expect(fixture.editor.library.pageTemplatesPane.deleteBtnByIndex(0).node).toBeTruthy();
});
```
it will fail.

spacedmonkey:
Update: @paulschreiber seems to be right and this is a known issue.

packages/story-editor/src/components/library/panes/pageTemplates/karma/savedPageTemplates.karma.js
miina: What about explicitly throwing an error if the node is not ready? If there's no error in the `waitFor`, it doesn't actually wait. Something like:

```
await waitFor(() => {
    const button = fixture.editor.library.pageTemplatesPane.deleteBtnByIndex(0).node;
    if (!button) {
      throw new Error('node not ready');
    }
    expect(fixture.editor.library.pageTemplatesPane.deleteBtnByIndex(0).node).toBeTruthy();
});
```

miina: Ah, okay, just now realized you mentioned the button itself is available but not actionable. Hm, what's the error it fails with?

merapi: It errors at the next step (finding the delete dialog).

miina: What if you bring clicking and checking for the dialogue into the same `waitFor` as well, for example:
```
...
await waitFor(() => {
    const button = fixture.editor.library.pageTemplatesPane.deleteBtnByIndex(0).node;
    if (!button) {
      throw new Error('node not ready');
    }
    await fixture.events.click(
        fixture.editor.library.pageTemplatesPane.deleteBtnByIndex(0)
      );
    const dialog = await fixture.screen.getByRole('dialog', {
        name: 'Delete Page Template',
      });
    if (!dialog) {
      throw new Error('node not ready');
    }
}, { timeout: 9000 });
...
```

paulschreiber:
@spacedmonkey did you tag me by accident?

BrittanyIRL:
I made a new ticket for the focus regression here: https://github.com/GoogleForCreators/web-stories-wp/issues/10806 

Once this is merged to `sidebar` and the draft PR for it to go to `main` is created i think a checklist tracking remaining todos on the description with the following so that everyone can be kept on the same page regardless of our many timezones would be helpful. 

- [ ] [Focus Regressions](https://github.com/GoogleForCreators/web-stories-wp/issues/10806)  - pea pod 
- [ ] Positioning of popups (inclusive of tooltips that are now cut off) - prometheus
- [ ] e2e test updates - prometheus 
- [ ] layers panel - prometheus

packages/story-editor/src/components/library/panes/pageTemplates/karma/savedPageTemplates.karma.js
maxyinger: It might just be the main thread getting blocked cus we're generating an image for the saved template when it gets added. We could either add something in the tests like: `await fixture.waitForIdle()` that just waits for the main thread to calm down, or we could mock out some stuff so we're not generating images for saved templates in the karma tests

maxyinger: yeaup, `waitFor` is from react-testing library, while the `expect ` in our karma tests are coming from jasmine, so they're not compatible and you can't use `waitFor` how it's described in the react-testing-library docs. gotta throw an error explicitly.

If I had to guess, the culprit is the main thread getting blocked because we're generating an image when we add a saved template. Posted some possible solutions above

packages/story-editor/src/components/panels/design/link/karma/link.karma.js
```diff
@@ -259,6 +261,8 @@ describe('Link Panel', () => {
       const frame = fixture.editor.canvas.framesLayer.frame(element.id).node;
       await moveElementToBottom(frame, 0);
 
+      await fixture.events.click(fixture.editor.inspector.designTab);
+      await fixture.events.sleep(100);
```
maxyinger: the could probably easily be switched to a `waitFor` correct?

maxyinger (approved):
lgtm. awesome how this was mostly name changes and little karma test updates

packages/story-editor/src/components/library/panes/pageTemplates/karma/savedPageTemplates.karma.js
merapi: @littlemilkstudio I've tried `waitForIdle` with `requestIdleCallback` with no luck, same with removal of `queuePageImageGeneration` 🤷🏻

merapi: > What if you bring clicking and checking for the dialogue into the same `waitFor` as well, for example:

For me, it just throws `node not ready` and that's it.

packages/story-editor/src/components/panels/design/link/karma/link.karma.js
merapi: Looks like this one is redundant 👍🏻

merapi:
I'm merging this one, all remaining works is as follows (all will be merged to `sidebar`):

https://github.com/GoogleForCreators/web-stories-wp/issues/10806
https://github.com/GoogleForCreators/web-stories-wp/issues/10804
https://github.com/GoogleForCreators/web-stories-wp/issues/10803
https://github.com/GoogleForCreators/web-stories-wp/issues/10812
https://github.com/GoogleForCreators/web-stories-wp/issues/10813

packages/story-editor/src/components/library/panes/pageTemplates/karma/savedPageTemplates.karma.js
maxyinger: ah bummer. probably not the issue then. thanks for exploring and reporting back here

PR #10763, opened 2022-03-01
Karma: Fix flakey Percy snapshots due to mutating config

### LINKED ISSUE #10715: Editor: Flakey Percy Snapshots
## Task Description

I noticed some confusing Percy snapshots for the editor Karma tests. [Here is an example run](https://percy.io/google/web-stories-wp/builds/16051415/changed/904659737?browser=chrome&browser_ids=20&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1600&widths=1600).

Here's what I am seeing:

**`enableHotlinking` feature flag unexpectedly enabled**

In Karma tests, `enableHotlinking` is currently disabled by default and only active for the tests that need it, i.e. the hotlinking Karma tests.

For some reason it is also active for some completely unrelated tests, such as `Canvas keys integration should not be able to delete element from a dialog: color picker open`.

You can confirm this by seeing two media buttons with icons instead of the single "Upload" button when the feature is disabled.

ea389da95046506458df45a8fc9e29ca541fcde7707b916863c773bc20799bb0

For some reason on `main` it looks as expected, with just the single "Upload" button active for this test.

Related: #10232 might resolve this by removing this code

**`floatingMenu` feature flag unexpectedly enabled**

Next up, there are lots of screenshots on the `main` branch where you can see the floating menu bar, even though that menu bar is only enabled for one specific test (Flip toggles).

Example test where the menu bar should not be visible:

24e3697d099db69af7e4f67c1a388acc1ae6c3da26e65e8bf7561355baa33b55

**Help center count off by one**

Sometimes the count of unread tips in the help center is 7, sometimes 8. Even when the test has nothing to do with the help center, e.g. `Panel: Style Presets CUJ: Creator can Apply or Save Text Style from/to Their Preset Library`

Screenshot 2022-02-25 at 10 53 40

**Missing text sets**

Sometimes the text sets are not fully loaded before taking a snapshot, which causes a large diff:

Screenshot 2022-02-25 at 10 54 31

Can happen for any text-related test I think, I just saw it in the `TextEdit integration add a multiline text element edit mode should not change height when entering and exiting edit mode` snapshot.

samwhale:
## Context

thanks to @littlemilkstudio and @BrookeGraham for the help with this.

## Summary

`deepMerge` mutates the `defaultConfig` object. This isn't a problem when running the application, but between tests the flags that are set end up persisting since they are added directly to the `defaultConfig` object.

## Relevant Technical Choices

Start with a deep clone of the `defaultConfig` object so that the base object is not mutated.

## To-do

n/a

## User-facing changes

n/a

## Testing Instructions

1. Percy screenshots should not show features that are behind feature flags _unless_ the feature flag is enabled in the test suite.

no

no

no

## Checklist

---

Fixes #10715


packages/story-editor/src/storyEditor.js
```diff
@@ -57,7 +58,12 @@ import KeyboardOnlyOutlines from './utils/keyboardOnlyOutline';
...
 
 function StoryEditor({ config, initialEdits, children }) {
-  const _config = useMemo(() => deepMerge(defaultConfig, config), [config]);
+  // Need `cloneDeep` so that karma tests are not using a modified
+  // `defaultConfig` between runs in CI
+  const _config = useMemo(
+    () => deepMerge(cloneDeep(defaultConfig), config),
+    [config]
+  );
   const { storyId, isRTL, flags } = _config;
```
BrittanyIRL: While this is getting eyes, I'm curious if there's a specific reason `_config` is getting memoized? Pretty sure this is only ever getting run on initial app mount.

BrittanyIRL (approved):
What a find 🤯  it'll be so nice to have these snapshots be accurate.

Did you double check that adding in `cloneDeep` has a negligible difference in load time?

packages/story-editor/src/storyEditor.js
samwhale: Yeah, good call. I don't think it needs to be memoized since `config` doesn't change.

packages/dashboard/src/dashboard.js
```diff
@@ -43,7 +44,12 @@ import { KeyboardOnlyOutline } from './utils';
 import defaultConfig from './defaultConfig';
 
 function Dashboard({ config, children }) {
-  const _config = useMemo(() => deepMerge(defaultConfig, config), [config]);
+  // Need `cloneDeep` so that karma tests are not using a modified
+  // `defaultConfig` between runs in CI
```
barklund: I'm not sure I understand why karma tests are even modifiying the default config anyway. Could we fix "that" instead or is that too much work?

Could we create another config file with only `defaultFlags`, that we could just spread, which the karma tests can then edit? Then we don't need to deep clone the rest of the config, that isn't touched anyway (if karma tests only touch the flags of course). Or if karma tests do touch many properties, we could create an "extraConfig" file that is loaded in, and only that one is deep-cloned every time?

It's not a big thing, it just seems a bit useless to clone an object in the client every time some views a page for no reason.

packages/dashboard/src/dashboard.js
```diff
@@ -43,7 +44,12 @@ import { KeyboardOnlyOutline } from './utils';
...
+  const _config = useMemo(
+    () => deepMerge(cloneDeep(defaultConfig), config),
```
mohdsayed: Instead of using `cloneDeep` in more than one place, would it make sense to use it in `deepMerge` utility function? 

Just thinking, now that we are adding one more `clone-deep` package, is there any deepmerge npm package available that does the cloning already? So that we can get rid of this utility function from `design-system`.

swissspidy: Does `structuredClone` work? Browser support is really good and there are polyfills

packages/dashboard/src/dashboard.js
maxyinger: Yea so the issue is the app modifies the default config, and karma renders the app, so karma will modify the config regardless of what you do in karma land. Definitely made this issue hard to track down.

> it just seems a bit useless to clone an object in the client every time some views a page for no reason.

yeaup. agreed. we could mock defaultConfig in the fixture and just have the mock value be a deep copy of defaultConfig, that way we only experience the cost of this call in the tests and not the user facing app.

I think one thing to consider tho is that the app having the sideEffect of mutating `defaultConfig` every time feels like a bug. This would bleed over into any unit test that renders `App` with a partial config too. It's not exclusive to the karma/e2e tests

maxyinger: one simple change that could fix this without a deep clone would just be to make `defaultConfig` a function that returns an object vs a module scoped object.

ie
```js
const getDefaultConfig = () => ({
  // ...
  flags: { ... }
});
...
const _config = useMemo(
    () => deepMerge(getDefaultConfig(), config),
    [config]
  );
```
vs
```js
const defaultConfig = {
  // ...
  flags: { ... }
};
...
const _config = useMemo(
    () => deepMerge(defaultConfig, config),
    [config]
  );
```

packages/dashboard/src/dashboard.js
maxyinger: TIL about structured clone. very cool reference.

If I'm reading it correctly,  `structuredClone` only works for transferable objects:
https://developer.mozilla.org/en-US/docs/Glossary/Transferable_objects#supported_objects

Also since it moves stuff from one place to another instead of copying it, we'd corrupt `defaultConfig` once the app has mounted once. this would break the `defaultConfig` for future mounts between tests.

Posted this solution above which is a very minor change, doesn't use `deepClone` and eliminates the side effect of mutating `defaultConfig` when we mount the app:
https://github.com/GoogleForCreators/web-stories-wp/pull/10763#discussion_r817930001

packages/dashboard/src/dashboard.js
maxyinger: would just allocate new memory for the entirety of `defaultConfig` every time the app mounts vs pointing to the same memory every time the app mounts

BrookeGraham: Just throwing in my 🪙  🪙  I agree: 
>I think one thing to consider tho is that the app having the sideEffect of mutating defaultConfig every time feels like a bug. 

However y'all decide to go about it, I don't think that we should continue to mutate `defaultConfig`.

packages/dashboard/src/dashboard.js
swissspidy: > The global structuredClone() method creates a [deep clone](https://developer.mozilla.org/en-US/docs/Glossary/Deep_copy) of a given value using the [structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm).

> The method **also** allows [transferable objects](https://developer.mozilla.org/en-US/docs/Glossary/Transferable_objects) in the original value to be transferred rather than cloned to the new object. Transferred objects are detached from the original object and attached to the new object; they are no longer accessible in the original object.

So it works just fine to deep clone a plain old JavaScript objects.

packages/dashboard/src/dashboard.js
swissspidy: Yeah we should not mutate `defaultConfig`, that was simply done by accident from what I can see.

A `getDefaultConfig()` function sounds simple enough and does not inflate bundle size.

packages/dashboard/src/dashboard.js
maxyinger: Oh so it only moves transferable objects. but functions normally how you'd expect for other objects. thanks for clearing that up!

samwhale: Got rid of `cloneDeep` with Max's suggestion of making `defaultConfig` a function

swissspidy (approved):

PR #10764, opened 2022-03-01
Saved Templates: remove virtual scroll

### LINKED ISSUE #10437: Page Templates:  thumbs keep "jumping" when you try to apply a template
## Bug Description
When I scroll through the Page Layouts, apply one page, then scroll through it again and click on a different thumb layout, the scroll jumps to a different position, and does not apply the layout I want. This happens when I am using my own saved templates, not the models available. When I use the keyboard, it doesn't happen.

## Expected Behaviour
To be able to choose a template, use it and not have it jumping when I select a template on my saved models. 

## Steps to Reproduce

Apply a template, confirm, scroll and click on a different page template on my saved models. 

## Screenshots

https://drive.google.com/file/d/1kwoQXDtEwBluYODg6-GlAizJ3GoQCSft/view?usp=sharing

## Additional Context

- Plugin Version: 1.16.0
- WordPress Version: 
- Operating System: Windows 10 Pro
- Browser: Chrome 97.0.4692.99

BrookeGraham:
## Context
Saved Templates layout in the editor is using virtual scroll. When applying a template to a story, the scrolling will get stuck on the same 6 or so items. This is preventing the user from being able to use or see all saved templates. 

## Summary
Now that users have been able to generate image blobs for saved templates, we can remove the virtual scroll component from templateList.

## Relevant Technical Choices
Removed `TemplateList` in favor of `DefaultTemplateList` since most functionality needed was already present in that component.  

## To-do
- [x] check if any other components can be remove along side `templateList`.
- [x] since keyboard nav is using the `+` icon when naving up the list, the template is not fully in view, fix that.

### Before: 
https://user-images.githubusercontent.com/1820266/156233311-fcea218c-393c-4633-b2bf-08eca9237eb9.mov

### After: 
savedtemplates

## Testing Instructions

1. Regression, make sure saved templates layout works as before
2. Ensure bug no longer present: 
3. Scroll toward end of save templates list
4. Add template to story
5. Attempt to scroll back up; scrolling should be smooth. 

## Checklist

---

Fixes #10437


BrookeGraham:
Wanted to flag this item I left in the todo 
>since keyboard nav is using the + icon when navigating up the list, the template is not fully in view, fix that.

This is an existing bug in staging that may require some eyes from design. I am not sure how I can bring the entire template image into view while navigating with the keyboard. As we use the keyboard to move from template to template, we are actually focusing on the `+` button. This is due to having two buttons on the template itself. Any thoughts?
keyboard nav

miina:
> This is an existing bug in staging that may require some eyes from design. I am not sure how I can bring the entire template image into view while navigating with the keyboard. As we use the keyboard to move from template to template, we are actually focusing on the + button. This is due to having two buttons on the template itself. Any thoughts?

That bug seems to be my fault :)

This is the place where scrolling happens:
https://github.com/GoogleForCreators/web-stories-wp/blob/7fc8dfef15f5fd356907dd6bd4ab607c1b5debdb/packages/story-editor/src/utils/useRovingTabIndex/index.js#L159-L161

We would probably need to adjust this logic in case the `rovingTabIndex` logic is used for nested elements (where depth is more than 0) -- instead of scrolling to the focusable element itself, scroll to its ancestor.

We have an util for getting the ancestor of an element by the depth of nesting:
https://github.com/GoogleForCreators/web-stories-wp/blob/7fc8dfef15f5fd356907dd6bd4ab607c1b5debdb/packages/story-editor/src/utils/useRovingTabIndex/nestedNavigation.js#L26-L39

Perhaps in case of having `depth > 0`, we could instead of just doing `e.scrollIntoView(...)` do something like
```
const ancestor = getAncestorByDepth(e, depth);
ancestor.scrollIntoView(...);
```

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplateList.js
```diff
@@ -105,20 +118,35 @@ function DefaultTemplateList({ pages, parentRef, pageSize, ...rest }) {
...
-          onFocus={() => handleFocus(page.id)}
-          isActive={currentPageId === page.id}
-          onClick={() => handlePageClick(page.story)}
-          columnWidth={pageSize.width}
-          {...rest}
-        />
-      ))}
+      {handleDelete || fetchTemplates
+        ? pages.map((page, index) => (
+            <PageTemplate
```
miina: Maybe leave a comment here to explain that one is Saved Templates and then other is Default Templates?

One nit from testing:
When opening the Default templates, the first item seems to show the Title by default now, as if it was in focus, even though it's not in focus nor hovered over either:
Screenshot 2022-03-02 at 11 35 35

BrookeGraham: Oh yep. That might sense that the title is showing cause the first one isActive when first loading. Nice catch

BrookeGraham:
Thanks so much for this info @miina !

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplateList.js
```diff
@@ -37,19 +37,28 @@ import { PANE_PADDING } from '../shared';
 import { useConfig } from '../../../../app/config';
 
 import DefaultPageTemplate from './defaultPageTemplate';
+import SavedPageTemplate from './savedPageTemplate';
 
 const WrapperGrid = styled.div`
   display: grid;
   width: 100%;
   margin-left: ${PANE_PADDING};
+  margin-bottom: 12px;
```
BrookeGraham: Noticed there was no spacing at the end of the template grids, so added some margin so that it doesn't look cutoff.

packages/story-editor/src/components/library/panes/pageTemplates/savedPageTemplate.js
```diff
@@ -262,27 +232,32 @@ function PageTemplate(
...
+          <PageTemplateTitleContainer isActive={isActive}>
+            <Text
+              as="span"
+              size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL}
+            >
+              {page.title}
+            </Text>
+          </PageTemplateTitleContainer>
+        )}
+      </PreviewPageWrapper>
```
BrookeGraham: Just curious, how does one get a title added to custom saved template? I mocked it while developing to make sure styles were applied correctly. But I am unsure how a user would name a template.

BrittanyIRL: @miina do you know this? 

I feel like it can't? There's no way to title saved templates.

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplateList.js
```diff
@@ -37,19 +37,28 @@ import { PANE_PADDING } from '../shared';
...
   gap: 12px;
   grid-template-columns: ${({ columnWidth }) =>
     `repeat(auto-fit, ${columnWidth}px)`};
   grid-template-rows: ${({ rowHeight }) =>
     `repeat(minmax(${rowHeight}px, 1fr))`};
 `;
 
-function DefaultTemplateList({ pages, parentRef, pageSize, ...rest }) {
+function DefaultTemplateList({
```
BrittanyIRL: nit: now this is the only `TemplateList` can probably drop the `Default`

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplateList.js
```diff
@@ -60,11 +69,12 @@ function DefaultTemplateList({ pages, parentRef, pageSize, ...rest }) {
   const pageRefs = useRef({});
 
   const handlePageClick = useCallback(
-    (page) => {
+    ({ templateId, version, title, ...page }) => {
+      // Just using destructuring above so we don't pass unnecessary props to addPage().
```
BrittanyIRL: 👍

packages/story-editor/src/components/library/panes/pageTemplates/defaultTemplateList.js
```diff
@@ -105,20 +119,35 @@ function DefaultTemplateList({ pages, parentRef, pageSize, ...rest }) {
...
-          page={page}
-          pageSize={pageSize}
-          onFocus={() => handleFocus(page.id)}
-          isActive={currentPageId === page.id}
-          onClick={() => handlePageClick(page.story)}
-          columnWidth={pageSize.width}
-          {...rest}
-        />
-      ))}
+      {handleDelete || fetchTemplates
```
BrittanyIRL: nit: For readability it might be nice to do `const isSavedTemplateView = handleDelete || fetchTemplates` and then check for that here.

BrittanyIRL (changes requested):
See ya, Virtual Scroll! 

Really like this lift. Happy it's finally time to remove virtual scroll, makes this so much more straight forward. 

The only issue I'm seeing is I can no longer tab from a saved template + button to the trashcan button.

BrookeGraham:
>the only issue I'm seeing is I can no longer tab from a saved template + button to the trashcan button.

Dang it! I was hyper focused on the templates getting into view correctly. Will check this out.

packages/story-editor/src/components/library/panes/pageTemplates/savedPageTemplate.js
miina: Same here -- as far as I know there's no way to do that currently.

BrookeGraham:
>>the only issue I'm seeing is I can no longer tab from a saved template + button to the trashcan button.

>Dang it! I was hyper focused on the templates getting into view correctly. Will check this out.

@BrittanyIRL do you mind giving this another go?
savedtemplateskeyboard

packages/story-editor/src/components/library/panes/pageTemplates/savedPageTemplate.js
BrittanyIRL: yeah, this can probably just get removed then, no need to include it since saved templates don't have titles. if they ever do we can add it in then.

BrittanyIRL (approved):
focus is fixed, thanks!

package-lock.json
```diff
@@ -45145,7 +45145,7 @@
         "colorthief": "^2.3.2",
         "draft-js": "^0.11.7",
         "draft-js-export-html": "^1.4.1",
-        "draft-js-import-element": "https://github.com/swissspidy/draft-js-import-element",
+        "draft-js-import-element": "git+ssh://git@github.com/swissspidy/draft-js-import-element.git#d477c24df373487a2018f9e9ed08fb4bdda18e3f",
```
maxyinger: intentional?

maxyinger (approved):
This looks and works great! awesome job!

package-lock.json
BrookeGraham: Snuck in with the rebase. Thanks

samwhale (approved):
This is super dope. Works well when there are no templates (and I add some) or if some already exist.

PR #10766, opened 2022-03-01
Story Details Modal: Publish + Checklist Button Parity

### LINKED ISSUE #10623: Story Details Modal x Prepublish Checklist Parity
## Feature Description

Separating this out from #10115 to untangle the different UI states the checklist button and publish button have based on the checkpoint of the checklist. 😅  

With the introduction of the story details modal we're sunsetting the "do you want to review the checklist" modal that shows up when you click publish and have either a) under 4 pages or b) other outstanding priority checklist issues to solve. 

(PB = editor publish button, CB = checklist button, MPB = modal publish button, MCB = modal checklist button)

| Issue | Location | Visual |
| --- | --- | --- |
| Less than 4 pages | PB | Screen Shot 2022-02-18 at 3 49 05 PM |
| Less than 4 pages | CB | Screen Shot 2022-02-18 at 3 49 36 PM |
| Less than 4 pages | MPB | Screen Shot 2022-02-18 at 3 50 15 PM
| Less than 4 pages | MCB | Screen Shot 2022-02-18 at 3 50 42 PM | 
| 4 or greater pages  | PB | Screen Shot 2022-02-18 at 3 45 20 PM |
| 4 or greater pages | CB | Screen Shot 2022-02-18 at 3 47 13 PM |

My thinking is that _regardless_ of story having 4 or more pages, the publish button on MPB should have the little exclamation mark and tooltip if there are priority issues present (which there will be if story has less than 4 pages).

If my thinking is correct, then should the editor's publish button have parity with the modal publish button? Does this mean if the story has under 4 pages and I click the editor's publish button and get to the modal - the modal publish button has the icon/tooltip - does the editor's button update to match it? Does the editor checklist button update to show the priority issues as well? 

## Updated after talking through with @agingoldseco 

All that needs to happen is when 'publish' is clicked in the editor, the checkpoint should be crossed to enable priority issue notices. This keeps everything in parallel. 

## Additional Context

Separated from #10115

BrittanyIRL:
## Context

Parity that was left out of initial plan for the story details modal in 1.18 (edge case clean up)

## Summary

The prepublish checklist has checkpoints that decide when to expose sections of the checklist. The one that got out of sync is for priority issues - which exposed at 4th page add _or_ when 'publish' in header is clicked - until then no badges will show up. Basically, we should be using the same "publish" button for both instances of it so that the logic for what version of it stay parallel. 

While doing that, noticed we could improve the flow for contributors - since that was tied in with the publish button I was already working on I did it here as well because it would have been more work to do it separately. Now, if you're a contributor (local dev env we have this set up already `contributor` `password`) and you click 'submit for review' you also get the story details modal with contributor approved options to finalize before you submit. 

## Relevant Technical Choices

- new action in checkpoint context to independently update to expose priority issues. 
- Consolidate `buttonWithChecklistWarning` to remove redundancies between the two publish buttons since they should behave the same - so now most of that logic lives there instead of in the header or the dialog/getting passed in from the header. 
- Removing check for canPublish on header publish button because regardless of that ability, users should all see this story details modal and then proceed from there accordingly.

## To-do

- [x] New e2e test for contributor 

It might be nice for a dialog to pop up for contributors telling them their story was submitted for review. Drew the line on that for scope creep here :) 

Also looking forward to cleaning up the checklist provider once story details experiment is deleted next sprint.

### Checklist Parity
- The checklist toggle button on the footer and the checklist button on the story details modal should have the same badges at all times now - when one changes, so does the other 
- The publish button in the header and the publish button in the story details modal should have the same copy, tooltips, badges at all times - when one changes, so does the other 

| Checklist Parity Demos | |
| --- | --- |
| https://user-images.githubusercontent.com/10720454/156441165-edabcd56-9a46-491d-9ef9-307b6e169ed5.mp4 | https://user-images.githubusercontent.com/10720454/156441220-57e020e6-0c76-4b85-8cc3-96b747c1fd40.mp4 |

### Contributor Flow 
- Now when the 'submit for review' button is hit in the header, contributor also gets story details modal. Here they can finalize before submitting and then when they hit submit in the modal it'll have the same flow as before (now the modal closes) and header updates like usual.

| Existing contributor flow | New contributor flow  |
| --- | --- |
| https://user-images.githubusercontent.com/10720454/156441360-1074b011-0665-4d32-9227-35fdbb9cd965.mp4 | https://user-images.githubusercontent.com/10720454/156441419-4856cd2a-1d40-4222-839b-bce1b8aa6568.mp4 |

## Testing Instructions

See the above 😄  

- Make sure publish and checklist buttons stay in sync in editor and modal 
- Make sure contributor can still submit stories for review 

1.

## Checklist

---

Fixes #10623


packages/story-editor/src/components/header/buttons/buttonWithChecklistWarning.js
```diff
@@ -72,18 +75,23 @@ InnerButton.propTypes = {
...
-  text,
-  isUploading,
-  canPublish,
+  disabled,
+  hasFutureDate,
   ...buttonProps
 }) {
+  const isUploading = useIsUploadingToStory();
+  const { isSaving, canPublish, status } = useStory(({ state }) => ({
+    isSaving: state.meta?.isSaving,
```
BrittanyIRL, line 85: Some unit tests were having trouble with this, so checking if `isSaving` exists to keep tests intact

packages/story-editor/src/components/header/buttons/buttonWithChecklistWarning.js
```diff
@@ -114,22 +122,33 @@ function ButtonWithChecklistWarning({
...
+  const text = canPublish
+    ? publishText
+    : __('Submit for review', 'web-stories');
   return (
-    <Tooltip title={toolTip} hasTail>
+    <Tooltip title={toolTip} popupZIndexOverride={11} hasTail>
       <InnerButton
         text={text}
-        {...buttonProps}
+        disabled={disabled || isSaving || isUploading}
```
BrittanyIRL, line 145: pulled `disabled` out from `buttonProps` so that it can be one of these 3 values that would set the button to be disabled.

packages/story-editor/src/components/header/buttons/publish.js
```diff
@@ -109,44 +99,36 @@ function PublishButton({ forceIsSaving }) {
   }, [refreshPostEditURL, saveStory, hasFutureDate, title, canPublish]);
 
   const handlePublish = useCallback(() => {
-    if (
-      canPublish &&
-      (shouldReviewDialogBeSeen || isUpdatedPublishModalEnabled)
-    ) {
+    updateToAllowPriorityIssues();
```
BrittanyIRL: as soon as the publish button in the header is hit we want to allow priority issues to surface in the checklist.

packages/story-editor/src/components/publishModal/header/index.js
```diff
@@ -48,13 +49,16 @@ const CloseButton = styled(Button)`
   margin: 6px auto 5px 6px;
 `;
 
-const PublishButton = styled(Button)`
-  height: 32px;
-  padding: 6px 8px;
+const PublishButtonWrapper = styled.div`
   margin: 6px 6px 5px auto;
```
BrittanyIRL, line 53: opted for the shared publish button padding (`4px 6px`)

packages/story-editor/src/components/publishModal/stories/index.js
```diff
@@ -37,27 +41,46 @@ const MockDocumentPane = () => (
...
   argTypes: {
     onPublish: { action: 'onPublish clicked' },
     onClose: { action: 'onClose clicked' },
+    checkpoint: {
+      options: Object.values(PPC_CHECKPOINT_STATE),
+      control: { type: 'select' },
+      name: "Checklist's given checkpoint",
+    },
+    storyStatus: {
+      options: ['publish', 'future', 'private', 'draft'],
```
BrittanyIRL, line 67: These come from wp-story-editor, didn't feel worth importing from a separate package.

packages/story-editor/src/components/publishModal/stories/index.js
```diff
@@ -124,19 +150,20 @@ export const _default = (args) => {
...
-            <PublishModal {...args} />
-          </ChecklistCountProvider>
-          <CheckpointContext.Provider
-            value={{
-              actions: {
-                onPublishDialogChecklistRequest: noop,
-              },
-            }}
-          >
-            <ChecklistCountProvider hasChecklist={args.hasChecklist}>
```
BrittanyIRL, line 136: (snuck through a rebase, cleaning up providers)

packages/e2e-tests/src/specs/editor/storyDetailsModal/contributorUser.js
```diff
@@ -0,0 +1,171 @@
...
+      { text: 'Public' }
+    );
+    await expect(visibilityDropDownButton).toMatch('Public');
+
+    await visibilityDropDownButton.click();
+
+    await expect(page).not.toMatchElement(
+      '[aria-label=" Option List Selector"]'
+    );
+  });
```
BrookeGraham, line 80: This is checked [here](https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/e2e-tests/src/specs/editor/contributorUser.js). Do we want it tested in both places?

packages/story-editor/src/components/header/buttons/buttonWithChecklistWarning.js
```diff
@@ -114,22 +122,33 @@ function ButtonWithChecklistWarning({
...
+        : __('Publish', 'web-stories'),
+    [hasFutureDate, status]
+  );
+
+  const text = canPublish
+    ? publishText
+    : __('Submit for review', 'web-stories');
   return (
-    <Tooltip title={toolTip} hasTail>
+    <Tooltip title={toolTip} popupZIndexOverride={11} hasTail>
```
BrookeGraham: This is just a thought, since we have to pass in `popupZindexOverride={11}` for a few things specifically for this publish modal, maybe we should make it a const / enum to ensure we are always using the same zindex. (or at least using the correct value for a particular use case.  Definitely not a blocker.

BrookeGraham (approved):
Looks great.

packages/story-editor/src/components/header/buttons/buttonWithChecklistWarning.js
BrittanyIRL: yeah, i was thinking about that with this ticket here: https://github.com/GoogleForCreators/web-stories-wp/issues/10636 since stacking z index is the solution/problem here too. Seems like named z indexes for our layers of popups would be beneficial to the readability.

BrittanyIRL: ok, fixed it here :)

packages/e2e-tests/src/specs/editor/storyDetailsModal/contributorUser.js
BrittanyIRL: I think it doesn't hurt. It's also a good double check that the two versions remain in sync should some prop get changed where something gets weird this check in the modal vs the editor would catch a regression.

packages/story-editor/src/components/checklist/checkpointContext/checkpointContext.js
```diff
@@ -118,10 +118,14 @@ function ChecklistCheckpointProvider({ children }) {
     dispatch(PPC_CHECKPOINT_ACTION.ON_PUBLISH_CLICKED);
   }, []);
 
+  const updateToAllowPriorityIssues = useCallback(
+    () => dispatch(PPC_CHECKPOINT_ACTION.ON_PUBLISH_CLICKED),
+    []
+  );
```
samwhale: 💭 nit: maybe `showPriorityIssues` could work here too? I think `updateToAllow` is more verbose, so something simpler could be easier to read here. But this is fine too!

packages/story-editor/src/components/header/buttons/publish.js
```diff
@@ -109,44 +99,36 @@ function PublishButton({ forceIsSaving }) {
...
       setShowDialog(true);
       return;
     }
 
     publish();
   }, [
+    updateToAllowPriorityIssues,
     shouldReviewDialogBeSeen,
     isUpdatedPublishModalEnabled,
-    canPublish,
```
samwhale, line 124: 🤔 is this not necessary anymore?

packages/story-editor/src/components/header/buttons/publish.js
```diff
@@ -109,44 +99,36 @@ function PublishButton({ forceIsSaving }) {
...
-  const text =
-    hasFutureDate && status !== 'private'
-      ? __('Schedule', 'web-stories')
-      : __('Publish', 'web-stories');
-
   return (
     <>
       <ButtonWithChecklistWarning
         onClick={handlePublish}
-        disabled={isSaving || forceIsSaving || isUploading}
```
samwhale, line 139: what was the difference between `forceIsSaving` and `isSaving`?

samwhale (approved):
Nice, working locally for me. Left some small questions, but I think something UX wise that's confusing is there's no notification that the contributor has submitted a story for review.

💭 for the future:
1. Should a snackbar be shown or some other notification?
2. Should the contributor be navigated away?
3. Should the button be disabled until the contributor makes another change?

things for later :)

packages/story-editor/src/components/checklist/checkpointContext/checkpointContext.js
BrittanyIRL: that's a better name! thank you!

packages/story-editor/src/components/header/buttons/publish.js
BrittanyIRL: nope because `isUpdatedPublishModalEnabled` is enabled by default right now so when we clean out the feature flag all this is going to end up being is 
```
const handlePublish = useCallback(() => {
 updateToAllowPriorityIssues();
setShowDialog(true);
}, [updateToAllowPriorityIssues])
```

packages/story-editor/src/components/header/buttons/publish.js
BrittanyIRL: great question, so `forceIsSaving is for the metaboxes specific to wp-story-editor, it gets set in `wp-story-editor/src/components/header/buttons` whereas `isSaving` is based on the story itself.

samwhale: So this button should only be disabled when metaboxes are updating?

BrittanyIRL: or when media/story is updating.

samwhale: Oh I think I see where I'm confused now:

`forceIsSaving` is for metaboxes, but for media/story updates that's tracked by `isSaving` and `isUploading`, right? Those were removed here so wondering if those need to be added back in.

maxyinger (approved):
Looks great! nice work!

packages/story-editor/src/components/header/buttons/publish.js
BrookeGraham: I think that `isSaving` and `isUploaded` are now defined in the  `ButtonWithChecklistWarning` component

samwhale: Ah, yup! 😮‍💨 thank you!

PR #10779, opened 2022-03-02
Support Fast Refresh

### LINKED ISSUE #3689: React Fast Refresh
## Feature Description

[Fast Refresh](https://github.com/pmmmwh/react-refresh-webpack-plugin) (née Hot Reloading) might be a welcome enhancement for engineers working on the project.

## Additional Context

https://github.com/pmmmwh/react-refresh-webpack-plugin

---

swissspidy:
## Context

React Fast Refresh is the recommended to improve DX by enabling "hot reloading" during development.

See https://www.netlify.com/blog/2020/12/03/what-is-react-fast-refresh/ for an example 

## Summary

This adds support for fast refresh for the dashboard and story editor.

## Testing Instructions

1.

## Checklist

---

Fixes #3689


swissspidy:
@samwhale As mentioned on #3689, this is the code I have so far for enabling Fast Refresh in the editor and dashboard. It seems to work already, but haven't properly tested it. At least on the dashboard, when using `npm run serve`, changes I made to some UI labels were reflected automatically without having to reload the page.

packages/story-editor/src/utils/copyPaste.js
```diff
@@ -149,5 +146,3 @@ export function addElementsToClipboard(page, elements, animations, evt) {
     `<!-- ${serializedPayload} -->${htmlContent}`
   );
 }
-
-export { getOffsetCoordinates as getPastedCoordinates };
```
swissspidy, line 150: The changes in this file here are unrelated, but webpack complained about it when I started debugging Fast Refresh.

This re-export was completely unused and doesn't make sense anyway.

samwhale:
Got tied up in some other stuff so posting an update real quick. Will work on this more next week:

Seems like fast refresh with that plugin works for the dashboard. However, the editor is crashing on first visit and on hard refreshes of the page.

It seems like anything nothing is exported from the `/story` folder, so any import from the story folder fails: https://github.com/GoogleForCreators/web-stories-wp/blob/08e7fb89ee55bc0adf862b470c7a3

Two exceptions happen:

---

first 

```bash
Uncaught TypeError: Cannot read properties of undefined (reading 'STORY_EVENTS')
    at Module.STORY_EVENTS (useLoadStory.js:159:1)
    at Module.STORY_EVENTS (useHistoryReducer.js:109:1)
    at registerExportsForReactRefresh (RefreshUtils.js:147:1)
    at Object.executeRuntime (RefreshUtils.js:182:1)
    at $ReactRefreshModuleRuntime$ (index.js:49:1)
    at Module../packages/story-editor/src/app/index.js (index.js:49:1)
    at Module.options.factory (react refresh:6:1)
    at __webpack_require__ (bootstrap:24:1)
    at fn (hot module replacement:61:1)
    at Module../packages/story-editor/src/elements/text/edit.js (display.js:330:1)
```

Looking at the imported modules, I'm seeing that when webpack's `bootstrap.js` is trying to import the modules - it is not finding anything and we get the first error from above: `Uncaught TypeError: Cannot read properties of undefined (reading 'STORY_EVENTS')`.
Screen Shot 2022-03-18 at 1 23 50 PM
Screen Shot 2022-03-18 at 1 24 13 PM
9aeaa2a09a2/packages/story-editor/src/app/story/index.js#L19-L30

---

and second:

```bash
Uncaught TypeError: webStories.initializeStoryEditor is not a function
```

This one seems strange because the `webStories` object is getting defined, but `initializeStoryEditor` is not getting attached to it.

In the browser console:
image

---

Wasn't seeing the wp-story-editor index file run, but it is loaded in to the browser: https://github.com/GoogleForCreators/web-stories-wp/blob/08e7fb89ee55bc0adf862b470c7a39aeaa2a09a2/packages/wp-story-editor/src/index.js#L62-L71

Hoping that fixing the first error above will solve the second, but we'll see next week

samwhale:
a898f4d fixed the bug we were seeing with a failed socket connection, but looks like the plugin should handle it too: https://github.com/pmmmwh/react-refresh-webpack-plugin/pull/552. Confused about why it's broken, but will come back to it later once I figure out the editor

samwhale:
Spent today working on this again. Unfortunately, wasn't able to crack it - so I wanted to post some findings.

---

I've been able to confirm that the exports defined here are causing the app to crash on load: https://github.com/GoogleForCreators/web-stories-wp/blob/b4dfac65fd276db7aa94f0ed2264f7720b6f3d88/packages/story-editor/src/app/index.js#L30-L37

Stubbing these out lets the rest of the app load.

Further - I found that if they are stubbed out (and the app only renders a `<div />` so that the stubs aren't used) we end up seeing that the imports are eventually found. 

```js
// packages/story-editor/src/storyEditor.js

function StoryEditor({ config, initialEdits, children }) {
  const _config = deepMerge(getDefaultConfig(), config);

  return <div />;
}
```

```js
// packages/story-editor/src/app/index.js

let useStory = () => {};
let useStoryTriggers = () => {};
let useStoryTriggerListener = () => {};
let useStoryTriggersDispatch = () => {};
let storyReducers = {};
let STORY_EVENTS = {};
import * as obj from './story';
console.log('app/index.js', obj); // will log the correct imports !!!
```

It seems that the initial pass at requiring `/story` fails, but after it ends up trying again it can find them.

---

It looks like webpack is seeing the code and adding it to the bundle. I see the definition for `STORY_EVENTS` in the compiled file `assets/js/packages_story-editor_src_index_js.js` with this plugin and the current setup.
image

---

So it seems like some wires aren't getting connected correctly. Narrowing it down further, I found that when webpack's `require` is being run it doesn't end up running this code: https://github.com/GoogleForCreators/web-stories-wp/blob/b4dfac65fd276db7aa94f0ed2264f7720b6f3d88/packages/story-editor/src/app/story/index.js#L23-L28.

Further, if I move that import statement to be before the `storyReducers` import statement on line 19 - then we don't see the `storyReducers` module being imported at all before the crash. Below I am logging out the modules being required using webpack's require fn.

|Normal|`storyTriggers` being imported before `storyReducers`|
|--|--|
|image|image|

Moving those imports ends up preventing the app from looking in the `storyReducer/reducers` folder before crashing.

---

I still don't know _why_ it's failing. I'll give myself another day at this before moving onto something else and taking a break from it.

swissspidy:
@merapi @sayedtaqui This is a long shot, but perhaps one of you has some ideas here regarding these imports?

mohdsayed:
> @samwhale As mentioned on #3689, this is the code I have so far for enabling Fast Refresh in the editor and dashboard. It seems to work already, but haven't properly tested it. At least on the dashboard, when using `npm run serve`, changes I made to some UI labels were reflected automatically without having to reload the page.

@swissspidy Was it automatically reloading the page to reflect changes, or without page reload?

mohdsayed:
After merging the latest main all of the above issues are resolved ( probably because of recent package organization PRs ), and the editor, dashboard both automatically reload the page whenever there is a new change in any file ( I think because it reads `window.location.origin` ), but yes the socket connection fails.

webpack.config.cjs
```diff
@@ -304,6 +312,17 @@ const templateParameters = (compilation, assets, assetTags, options) => ({
...
   ...sharedConfig,
+  devServer: !isProduction
+    ? {
+        devMiddleware: {
+          writeToDisk: true,
+        },
+        hot: true,
+        allowedHosts: 'all',
+        host: 'localhost',
+        port: 8080,
```
mohdsayed: @samwhale I see https://github.com/GoogleForCreators/web-stories-wp/pull/10779#issuecomment-1074335024, but why are we specifically using port `8080` ?

swissspidy:
> Was it automatically reloading the page to reflect changes, or without page reload?

IIRC there was no page reload, but its' been a few weeks now, so not 100% certain.

I did use `port: 'auto'` though without any issues.

FWIW I am also using localwp.com for local development and not the Docker container from the repo—maybe that makes a difference.

mohdsayed:
> > Was it automatically reloading the page to reflect changes, or without page reload?
> 
> IIRC there was no page reload, but its' been a few weeks now, so not 100% certain.
> 
> I did use `port: 'auto'` though without any issues.
> 
> FWIW I am also using localwp.com for local development and not the Docker container from the repo—maybe that makes a difference.

Automatic page refresh is fine however I am wondering how we are even expecting it to hot reload when the Webpack dev server is running on some other port http://localhost:8080 and serving content from `web-stories-wp/public` directory which doesn't exist, and we are running WordPress instance on a totally different URL 🤔  .

swissspidy:
Just tried it again with `port: 'auto'` and `npm run serve`.

Both the dashboard and editor do hot reload just fine!

No socket errors, no page refreshes

Screenshot 2022-03-31 at 12 10 19

Screenshot 2022-03-31 at 12 08 57

So... this seems to work? 🤷

mohdsayed:
> Just tried it again with `port: 'auto'` and `npm run serve`.
> 
> 
> 
> Both the dashboard and editor do hot reload just fine!
> 
> 
> 
> No socket errors, no page refreshes
> 
> 
> 
> Screenshot 2022-03-31 at 12 10 19
> 
> 
> 
> Screenshot 2022-03-31 at 12 08 57
> 
> 
> 
> So... this seems to work? 🤷 
> 
> 

Awesome, I am using flywheel for local development and it doesn't work for me. What are you using?

swissspidy:
https://localwp.com/

mohdsayed:
@swissspidy 

💡  Ah, I was making changes to `wp-story-editor` in `initializeStoryEditor` where it was reloading the page. It does hot reloading when updating the core packages with `port: "auto"` . Works for both editor and dashboard. Yay! 🥳 🎉

webpack.config.cjs
samwhale: I was seeing an error about the web socket connection failing. `auto` wasn't working for me locally for some reason, even though it should have.

> I did use `port: 'auto'` though without any issues.
> 
> FWIW I am also using localwp.com for local development and not the Docker container from the repo—maybe that makes a difference.

I'm using the docker container - so maybe it had to do with that?

samwhale:
😕 it's not working for me locally for both the dashboard and the editor now. I'm using docker, and here's what I'm seeing when running `npm run serve`:
image

My steps:
1. Start docker
2. Run `npm run env:start`
3. In a separate terminal window run `npm run serve`
4. Navigate to the dashboard or editor urls

Dashboard:
image

Editor: 
image

---

💭 It seems like y'all have a good setup without docker. Do we need to be using docker at all then? If the setup needs to change in a very complex way then maybe it'd be better to leave it?

mohdsayed:
Installed docker and tried and it does hot reloading for me. Not showing those errors in my console.


https://user-images.githubusercontent.com/6297436/161222524-31a72f2a-b7a9-4bee-81aa-7abe39b8dd5b.mp4


Console is clean

image

swissspidy:
Same here; no issues whatsoever with the Docker environment.

@samwhale Could you perhaps share some details about your env to see if there's any specific difference that causes this issue for you? Like Node version (after `nvm install`), Docker version, OS, etc.

samwhale:
> Same here; no issues whatsoever with the Docker environment.
> 
> @samwhale Could you perhaps share some details about your env to see if there's any specific difference that causes this issue for you? Like Node version (after `nvm install`), Docker version, OS, etc.

That's great to hear that's it's probably just my env then. Here's my specs:
OS: macOS Big Sur 11.6.2
Docker version: 4.0.1
Node version: 16.4.2

I'll get somebody else on pea pod to test it today to see if it's just me :). may just be that I have my env setup incorrectly. Maybe restarting from scratch would help.

samwhale:
Worked on this with the team and it all works for us now, which is great. I'm very confused about why it wasn't working before.

The only thing we were all seeing was this log:

image

This happened because `port` was set to `auto`.
image

In `getUrlFromParts`, a new URL tries to be created (`new URL("/ws", "ws://localhost:auto")`, but fails and yields this error:
image

Then the socket connection fails. But fast refresh works! I don't know what changed 😅

timarney:
Testing using the `wp-env` env

Works but than a delayed crash

https://user-images.githubusercontent.com/62242/161589683-f4f2a809-9d2d-4b55-aeea-b1805b7cc1ab.mp4

BrittanyIRL (approved):
Well this is exciting. Everything works great for me locally. I'm not seeing the crash that Tim reported earlier today.

swissspidy:
Thanks everyone for testing so far!

Since this is a new feature that doesn't affect regular builds and seems to work reliably in most cases, I am going ahead to merge this so we can test this a bit more in day to day use.
If crashes should be more frequent we can take another stab at it.

PR #10790, opened 2022-03-03
[TRY][WIP]: Introduce data package using redux-toolkit

mohdsayed:
## Context

See https://github.com/GoogleForCreators/web-stories-wp/issues/10614

## Testing Instructions

1.

## Checklist

---

Fixes #


packages/wp-story-editor/src/index.js
```diff
@@ -59,6 +60,7 @@ import { GlobalStyle } from './theme.js';
 
 window.webStories = window.webStories || {};
 window.webStories.domReady = domReady;
+window.webStories.data = data;
```
mohdsayed, line 63: This may allow third-party plugin developers to use data like

```js
webStories.data.select( 'sliceName' );
webStories.data.dispatch('sliceName', 'action', value);
webStories.data.subscribe();
```

packages/story-editor/src/app/store/setupReducers.js
```diff
@@ -0,0 +1,26 @@
...
+ * External dependencies
+ */
+import { combineReducers, replaceReducer } from '@googleforcreators/data';
+import { elementReducer } from '@googleforcreators/elements';
+
+const rootReducer = combineReducers({
+  element: elementReducer,
+});
+
+replaceReducer(rootReducer);
```
maxyinger, line 26: Just cus I'm unfamiliar with redux, would this mean the only place we can setup the reducer is here in `story-editor`? I'm assuming `replaceReducer` completely replaces the stores reducer?

mohdsayed: Yes, that's correct, it completely replaces the reducers and we can only set them up here. If we see a need to add reducers elsewhere, I think we can create `injectReducer` using [this approach](https://redux.js.org/usage/code-splitting#reducer-injection-approaches).

PR #10805, opened 2022-03-04
Update Composer dependencies and PHPStan configuration

swissspidy:
Updates dependencies and fixes some newly reported PHPStan errors.

Removes now obsolete copied PHPStan extension 

Changes namespace of custom PHPStan extensions


includes/Infrastructure/ServiceContainer/SimpleServiceContainer.php
```diff
@@ -53,6 +53,10 @@ public function get( $id ): Service {
 
 		$service = $this->offsetGet( $id );
 
+		if ( ! $service instanceof Service ) {
```
spacedmonkey: Can you explain this change and maybe by adding a unit test.

spacedmonkey:
This PR description needs a little more detail to explain these changes and why they are needed. There should also be unit / integration test for changes in [SimpleServiceContainer.php](https://github.com/GoogleForCreators/web-stories-wp/pull/10805/files#diff-ae326a9b3811396dbf43b7b09cfc3e39a3171e0ccf85092ff33a1af2e7c672bd)

includes/Infrastructure/ServiceContainer/SimpleServiceContainer.php
swissspidy: `$service = $this->offsetGet( $id );` above calls `ArrayObject:offsetGet()`, which returns `false|mixed`. PHPStan complains about that because this method here obviously returns `Service` and nothing else.

This can be resolved by either a `@var Service $service` docblock or this simple check here.

In fact, I should probably do the former because it's not possible to put anything other than a Service into this.

swissspidy: Addressed now.

spacedmonkey (approved):
SGTM

PR #10823, opened 2022-03-07
Publish Modal: Fix issues with apostrophes in site name

### LINKED ISSUE #10809: Publish Modal: Apostrophe in site name not displayed correctly
My site name is Let's Camp S'more and webstories is pulling the name with apostrophes as code. 

So `Let's` is `Let&#039;s`

What can be done?

swissspidy:
## Context

As reported at #10809, site titles with apostrophes in them are not correctly displayed in the pre-publish modal.

This is because WordPress saves them to the database as HTML entities, but we need the raw values.

## Summary

This PR fixes the issues with single quotes in the site name by using `html_entity_decode`

## Relevant Technical Choices

This PR also includes some other related tweaks I made while looking into this.

1. Call the `web_stories_publisher_name` filter earlier, within `load_from_post()`, so it doesn't run in a getter method. `get_publisher_name()` isn't supposed to modify the publisher name and run this filter every time it's called.
2. Do not escape publisher name / site title within `get_publisher_name`. Any escaping should be done as late as possible.
3. Do not escape `$lang` when setting the `html[lang]` attribute. The `setAttribute()` call does the escaping for us.

## User-facing changes

**Before**

Screenshot 2022-03-07 at 15 52 53

**After**

Screenshot 2022-03-07 at 16 13 40

## Testing Instructions

1. In the WordPress settings, change site title to include single/double quotes, e.g. `S'mores`.
4. Create a new story
5. Click on "Publish" to open the pre-publish modal
6. Verify that the site title is displayed correctly

No.

Any attributes printed on the page are automatically escaped through DOMDocument usage in PHP.

On the client-side, escaping is done automatically through the DOM/React.

No

No

## Checklist

---

Fixes #10809


spacedmonkey:
This really feels like a good place for a unit test.

swissspidy:
PHP test added for the `html_entity_decode` change

spacedmonkey (approved):

PR #10826, opened 2022-03-07
CORS Check: Initial implementation

### LINKED ISSUE #10706: CORS Check: Initial implementation
## Feature Description

See #10529 for more context.

In a nutshell, we want to detect potential issues when creators are using a custom CDN for their media but haven't set proper CORS headers.

The initial implementation can be modelled after our existing [`StatusCheck` component](https://github.com/GoogleForCreators/web-stories-wp/blob/e2101cc98226d7400867d4d67ac9d654a65f402f/packages/wp-story-editor/src/components/statusCheck/statusCheck.js) in the editor, which looks like this:

Screenshot 2022-02-24 at 11 44 24

### Implementation Brief

**Reminder:** since this is behind a feature flag we don't need extra UX for this (yet).

**Dialog**

* [ ] Add new `corsCheck` experiment / feature flag in `Experiments.php`
* [ ] Create new `CorsCheck` component analogous `StatusCheck` within `packages/wp-story-editor/src/components/corsCheck`
	* [ ] Add storybook story

Preliminary wording for this initial implementation:

```
**Unable to load media**

We detected a potential misconfiguration that prevents media items from loading correctly. This may be due to media being hosted  on an external CDN. <a>Learn how to address this</a> or <a>submit a new support topic</a> for additional help.
```

For now, the first link can go to `https://wp.stories.google/docs`, the second one (wrapped in `__()` to `https://wordpress.org/support/plugin/web-stories/`

**Algorithm**

Rough idea of this can work:

1. Create a new `getMediaForCorsCheck` function in `packages/wp-story-editor/src/api/media.js` that fetches a bunch of media items (let's say 10) via `apiFetch()`  and only loads the `source_url` field (`?_fields=source_url`)
2. Call that function in the dialog to get a list of media URLs.
3. Loop through the URLs. For each URL:
	1. Compare media URL with current origin
	2. _Match?_ -> do nothing
	3. _No match?_ -> perform `HEAD` request on the media URL to see if it can be accessed
	4. If any of the requests fail, bail early and display the dialog

For the looping part, something like this should work (

```js
let hasError;

try {
  await Promise.all(
    mediaUrls
      .filter((url) => url.startsWith(location.origin))
      .map((url) =>
        fetch(url, {
          method: 'HEAD',
        })
      )
  );
} catch (err) {
  hasError = true;
}
```

It's not perfect, but good enough for a first implementation.

**Tests**

* [ ] Add e2e tests for the dialog, presumably with a test plugin in `packages/e2e-tests/src/plugins` to mock a REST API response with some URLs that triggers the dialog
* [ ] Add JS unit tests for the dialog

spacedmonkey:
## Summary

Add a basic check for cross site origin for resources in the media library that are requested from another url. 

These feature is behind feature flag at this time. 

To test this PR, active the E2E Tests CORS error plugin. 

## User-facing changes
Screenshot 2022-03-09 at 11 30 35

## Testing Instructions

1.

## Checklist

---

Fixes #10706


packages/wp-story-editor/src/components/corsCheck/corsCheck.js
```diff
@@ -0,0 +1,64 @@
...
+        mediaItems = await getMediaForCorsCheck();
+      } catch (err) {
+        return;
+      }
+      if (!mediaItems?.length) {
+        return;
+      }
+      try {
+        await Promise.all(
+          mediaItems.map((url) => fetch(url, { method: 'HEAD' }))
```
swissspidy: Note that on the ticket I suggested filtering `mediaItems` by comparing with `location.origin` to avoid unnecessary HTTP requests to images on the same origin.
Is the worry here HTTP redirects? I see how that check would not work in that case, but not sure how likely that is.

spacedmonkey: Fixed.

packages/e2e-tests/src/plugins/web-stories-cors-error.php
```diff
@@ -0,0 +1,46 @@
...
+ * Plugin Name: E2E Tests CORS error
+ * Plugin URI:  https://github.com/googleforcreators/web-stories-wp
+ * Description: CORS errors on purpose for testing.
+ * Author:      Google
+ * Author URI:  https://opensource.google.com
+ *
+ * @package   Google\Site_Kit
+ * @copyright 2022 Google LLC
+ * @license   https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link      https://sitekit.withgoogle.com
```
swissspidy:
```suggestion
 * @copyright 2022 Google LLC
 * @license   https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
```

We should fix this for the other e2e test plugins as well. Can be done in this PR too.

includes/Experiments.php
```diff
@@ -299,6 +299,17 @@ public function get_experiments(): array {
...
 			],
+			/**
+			 * Author: @spacedmonkey
+			 * Issue: #10706
+			 * Creation date: 2022-03-07
+			 */
+			[
+				'name'        => 'enableCORSCheck',
+				'label'       => __( 'CORS check', 'web-stories' ),
+				'description' => __( 'Add a check in the editor for cors errors.', 'web-stories' ),
```
swissspidy:
```suggestion
				'description' => __( 'Add a check in the editor for CORS errors.', 'web-stories' ),
```

packages/e2e-tests/src/specs/wordpress/corsCheck.js
```diff
@@ -0,0 +1,62 @@
...
+ */
+import { addAllowedErrorMessage } from '../../config/bootstrap.js';
+
+describe('CORS check', () => {
+  withExperimentalFeatures(['enableCORSCheck']);
+  let removeCORSErrorMessage;
+  let removeResourceErrorMessage;
+
+  beforeAll(async () => {
+    await activatePlugin('e2e-tests-cors-error');
```
swissspidy: Can use the `withPlugin()` util here?

packages/wp-story-editor/src/api/media.js
```diff
@@ -69,6 +69,19 @@ export function getMedia(
...
+// Important: Keep in sync with REST API preloading definition.
+export function getMediaForCorsCheck(config) {
+  const path = addQueryArgs(config.api.media, {
+    context: 'edit',
+    per_page: 10,
+    _fields: 'source_url',
+  });
+
+  return apiFetch({ path }).then((attachments) =>
+    attachments.map((attachment) => attachment?.source_url)
```
swissspidy: Nit: `attachment` is not `null` here, so no need for optional chaining.

```suggestion
    attachments.map((attachment) => attachment.source_url)
```

packages/e2e-tests/src/plugins/web-stories-cors-error.php
```diff
@@ -0,0 +1,46 @@
...
+		return $response;
+	}
+	/**
+	 * Response data.
+	 *
+	 * @var array $data
+	 */
+	$data = $response->get_data();
+
+	$data['source_url']             = 'https://wp.stories.google/img/302e7662-696.jpeg'; // TODO: Find a better example of an image with cors issue.
```
swissspidy: WDYT about using https://ps.w.org/web-stories/assets/banner-772x250.png? Should be pretty stable.

The ones from `wp.stories.google/img` are not reliable as the URLs change often.

packages/e2e-tests/src/specs/wordpress/corsCheck.js
```diff
@@ -0,0 +1,62 @@
...
+    removeCORSErrorMessage = addAllowedErrorMessage(
+      'has been blocked by CORS policy'
+    );
+    removeResourceErrorMessage = addAllowedErrorMessage(
+      'Failed to load resource'
```
swissspidy, line 45: Is this from the same fetch request for the image? Or something else?

packages/wp-story-editor/src/components/corsCheck/corsCheck.js
```diff
@@ -0,0 +1,66 @@
...
+        return;
+      }
+      if (!mediaItems?.length) {
+        return;
+      }
+      try {
+        await Promise.all(
+          mediaItems
+            .filter((url) => !url.startsWith(location.origin))
+            .map((url) => fetch(url, { method: 'HEAD' }))
```
swissspidy: The e2e teat plugin will cause `mediaItems` to be an array with containing the same URL twice.

We can safely filter out duplicate URLs here.

```suggestion
          [
            ...new Set(
              mediaItems.filter((url) => !url.startsWith(location.origin))
            ),
          ].map((url) => fetch(url, { method: 'HEAD' }))
```

This way, only 1 request to `https://wp.stories.google/img/302e7662-696.jpeg` is made.

packages/wp-story-editor/src/components/corsCheck/corsCheck.js
```diff
@@ -0,0 +1,66 @@
...
+import { useFeature } from 'flagged';
+
+/**
+ * Internal dependencies
+ */
+import CorsCheckFailed from './corsCheckFailed';
+
+function CorsCheck() {
+  const [showDialog, setShowDialog] = useState(false);
+  const [doCheck, setDoCheck] = useState(true);
```
swissspidy: If you use an empty dependencies array for the `useEffect` below (and suppress the lint error for it), then this state is not needed, because the effect will only run once.

packages/e2e-tests/src/plugins/web-stories-cors-error.php
spacedmonkey:
```suggestion
	$data['source_url']             = 'https://ps.w.org/web-stories/assets/banner-772x250.png';
```

packages/e2e-tests/src/specs/wordpress/corsCheck.js
spacedmonkey: Not exactly sure. But when I run this headless locally, i see this.

```
  ● CORS check › should see media dialog

    expect(jest.fn()).not.toHaveErrored(expected)

    Expected mock function not to be called but it was called with:
    ["Failed to load resource: net::ERR_FAILED"]

      at Object.assertExpectedCalls (../../../node_modules/@wordpress/jest-console/build/@wordpress/jest-console/src/index.js:36:4)
```

packages/e2e-tests/src/plugins/site-kit-analytics.php
```diff
@@ -5,11 +5,27 @@
  * Description: Utility plugin to mock sitekit analytics
  * Author:      Google
  * Author URI:  https://opensource.google.com
+ * License: Apache License 2.0
+ * License URI: https://www.apache.org/licenses/LICENSE-2.0
  *
- * @package   Google\Site_Kit
- * @copyright 2020 Google LLC
+ * @copyright 2022 Google LLC
```
spacedmonkey:
```suggestion
 * @copyright 2020 Google LLC
```

packages/e2e-tests/src/plugins/site-kit-adsense.php
```diff
@@ -5,11 +5,27 @@
  * Description: Utility plugin to mock sitekit adsense
  * Author:      Google
  * Author URI:  https://opensource.google.com
+ * License: Apache License 2.0
+ * License URI: https://www.apache.org/licenses/LICENSE-2.0
  *
- * @package   Google\Site_Kit
- * @copyright 2020 Google LLC
+ * @copyright 2022 Google LLC
```
spacedmonkey:
```suggestion
 * @copyright 2020 Google LLC
```

packages/e2e-tests/src/plugins/web-stories-disable-media3p.php
```diff
@@ -5,11 +5,27 @@
  * Description: Disable 3P media.
  * Author:      Google
  * Author URI:  https://opensource.google.com
+ * License: Apache License 2.0
+ * License URI: https://www.apache.org/licenses/LICENSE-2.0
  *
- * @package   Google\Site_Kit
- * @copyright 2021 Google LLC
+ * @copyright 2022 Google LLC
```
spacedmonkey:
```suggestion
 * @copyright 2021 Google LLC
```

packages/e2e-tests/src/specs/wordpress/corsCheck.js
spacedmonkey: @swissspidy are you happy with me leaving this in?

swissspidy: It's important to know why a certain error is being suppressed / what it belongs to.

I just checked this myself to confirm that it's indeed the console error message for the `https://ps.w.org/web-stories/assets/banner-772x250.png` URL.

So that should be fine to leave it in.

Adding a comment explaining this would be helpful though.

swissspidy (approved):
Overall LGTM

packages/e2e-tests/src/specs/wordpress/corsCheck.js
```diff
@@ -0,0 +1,61 @@
...
+
+describe('CORS check', () => {
+  withExperimentalFeatures(['enableCORSCheck']);
+  withPlugin('e2e-tests-cors-error');
+
+  let removeCORSErrorMessage;
+  let removeResourceErrorMessage;
+
+  beforeAll(() => {
+    removeCORSErrorMessage = addAllowedErrorMessage(
```
spacedmonkey, line 40:
```suggestion
    // Ignore CORS error, this is present in the test by design. 
    removeCORSErrorMessage = addAllowedErrorMessage(
```

packages/e2e-tests/src/specs/wordpress/corsCheck.js
```diff
@@ -0,0 +1,61 @@
...
+      'has been blocked by CORS policy'
+    );
+    removeResourceErrorMessage = addAllowedErrorMessage(
```
spacedmonkey, line 44:
```suggestion
    // Ignore resource failing to load. This is only present because of the CORS error.
    removeResourceErrorMessage = addAllowedErrorMessage(
```

swissspidy:
There are now lint errors that need to be addressed

packages/migration/scripts/utils/updateTemplates.js
```diff
@@ -23,7 +23,6 @@ import { join, resolve } from 'path';
 /**
  * Internal dependencies
  */
-// eslint-disable-next-line import/no-unresolved
```
swissspidy, line 26: This line is required. It will be flagged as unneeded once you run a migration, which is misleading, but that doesn't mean it can be removed.

spacedmonkey: I reverted this change. It was removed my the linter and was committed by mistake.

timarney (approved):

PR #10831, opened 2022-03-07
Editor: Select canvas elements when they are focused

### LINKED ISSUE #10806: UI Simplification: Sidebar Focus Regressions
## Bug Description

Noted by Miina in PR #10760 : 

> When tabbing out of the library, the focus seems to get lost somewhere? (Previously went visibly to the canvas/workspace). E.g. tab to Shapes -> Stickers -> tab again -> not sure where the focus is.

([comment](https://github.com/GoogleForCreators/web-stories-wp/pull/10760#issuecomment-1058999666))

## Expected Behaviour

No focus regressions in new sidebar layout. 

## Additional Context

This ticket is to ensure that the new sidebar that @merapi is working on has no focus regressions. There's some outstanding issues, as noted here: #10698  and #10793 - to keep scope tight here, it's really just focused on making sure nothing new is regressing in this sidebar branch. @miina has requested that before the sidebar gets merged to main new focus issues are resolved. 

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

samwhale:
## Context

I think this has been around for a while. Looking at staging it’s there too, so I tracked down the commit. It looks like it started with [this PR](https://github.com/GoogleForCreators/web-stories-wp/pull/10388).

Now when tabbing out of the library, the canvas receives focus. However - there is no event handle on the background frame that sets that element as the 'selected element' in the story.

This is an example of one approach to fix the perceived loss of focus.

## Summary

This 

## Relevant Technical Choices

I think there's two approaches here that would both solve the problem:
1. (as demonstrated in this PR) we can add event handlers to the frame that catch the focus and select the element when it is focused. This has the benefit of being able to change focus from elements using `Tab`. However, this may not be a good thing, so would love your feedback.
2. we can focus the canvas container using `useFocusCanvas`. This has the benefit of working the same way the editor did before.

## To-do

n/a

## User-facing changes

|Before|After|
|--|--|
|focus-hide|tabbin|

## Testing Instructions

1. Add an element to the canvas
2. Focus something in the library (any tab)
3. Press `Tab` until you get to the canvas
4. Should see the background focused
5. Every time you press `Tab` from there, if the next focused target is an element on the canvas then that element should be set as 'selected'
6. Once you tab out of the canvas, the previously selected element should remain selected

## Checklist

---

Fixes #10806


samwhale:
Note: this has existed for over 60 commits and should not block @merapi's work :). This can be merged into main instead of the `sidebar` branch too.

samwhale:
Chatted with @miina and this is not needed. Focus behavior is not blocking in the sidebar pr. Check ticket for more information.

Issue #10824, opened 2022-03-07
Performance: Page Zoom, integrate with page canvas provider


maxyinger:
This ticket may be a little more exploratory, but following [10730](https://app.zenhub.com/workspaces/web-stories-5e94d3aced449034e1e9b226/issues/googleforcreators/web-stories-wp/10730) it would be nice to see if we can bring in the image generation action used for the page zoom into the page canvas provider, so we can share the same generated image with other components in the app (page carousel, also any contrast calculations).

this should hopefully cut down on the amount of expensive calls to generate images from story pages in the app.

littlemilkstudio:
Turns our zoom doesn't actually generate a canvas/image, just scales up the dom nodes. Will use this ticket as an opportunity to integrate the pageCanvasProvider with the eyedropper instead

littlemilkstudio:
Welp, the eye dropper is not hard to integrate with the page canvas cache to supply page canvas/image data, however, the existing code is tightly coupled to the size of images and nodes in the editor.

Currently it basically makes a lot of assumptions that the story page DOM node is the same pixel size as the image data we reference. With the page canvas provider it is not guaranteed that the image will be the same size as the story page DOM node.

There's definitely approaches that can be taken that use some linear interpolation to accommodate for differences in coordinate system between DOM events and the image data, but updating the logic in the eyedropper is a larger lift and out of scope for this ticket right now.

Closing this issue with this synopsis. If anyone wants to create a ticket for integrating the Page Canvas Provider into the eyedropper so that the eyedropper interacts with the page image cache, it's definitely doable, but should be marked as a 5 to accommodate for the updates needed in the eyedropper business logic.

Issue #10822, opened 2022-03-07
Performance: Story carousel integrate with page canvas provider


maxyinger:
Following [this ticket 10730](https://app.zenhub.com/workspaces/web-stories-5e94d3aced449034e1e9b226/issues/googleforcreators/web-stories-wp/10730) We should have infrastructure, to integrate the generation of story pages for page carousels into the page canvas provider. 

Should help cut down on expensive calls to generate images. Should also be pretty simple to integrate with the background task api like the current behavior.

More context: https://github.com/googleforcreators/web-stories-wp/issues/9749

BrittanyIRL:
@felipebochehin87 there's testing directions in the PR for this (#11207) - the same PR will also fix #9958.

felipebochehin87:
Tested against https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/, using PR #11207.

1. Created a new story
2. Added a background image to the first page - thumbnail updated
3. Duplicated page 1 to create page 2 - thumbnail displayed as expected
4. Changed the background image on page 2 - thumbnail displayed as expected
5. Duplicated page 2 to create page 3 - thumbnail displayed as expected
6. Changed images on 3 pages - thumbnail displayed as expected
7. Checked the thumbnails displayed on the checklist - thumbnails displayed as expected
8. Checked the preview page - thumbnails displayed as expected

Issue #10821, opened 2022-03-07
Performance: Saved template image generation integrate with page canvas provider


maxyinger:
Following [this ticket](https://app.zenhub.com/workspaces/web-stories-5e94d3aced449034e1e9b226/issues/googleforcreators/web-stories-wp/10730) We should have infrastructure set up so that we can pull a previously generated story page image if it's already been generated for a contrast calculation.

If we don't have the most up to date page already generated, we can default to the existing approach.

Issue #10292, opened 2022-01-20
Performance: LibraryMoveable Discovery


maxyinger:
We get a lot of re-renders from LibraryMoveable whenever we drag to move an object. This may not be avoidable, but goal of this ticket is to look into this and document possible solutions or why the current approach is necessary:
![Screen Shot 2022-01-20 at 1.29.15 PM.png](https://images.zenhubusercontent.com/5ea0a3e2f739259b2b0b092a/a81f1db7-b9c9-467f-a4b0-6ae64a353d37)

swissspidy:
Related / Previously: #9580 #9882

maxyinger:
@swissspidy Oh dope. Thanks for linking the previous work here!

maxyinger:
Started looking into why some of the library panels, like text & shapes, have empty re-renders when selecting new elements. LibraryMoveable ended up being the root culprit.

Looks like a lot of things are set up well in here thanks to previous perf work in the area. 

Seems the empty re-renders are now being caused by the addition of the adaptive colors for text sets feature. Through `LibraryMoveable -> useInsertTextSet -> usePageAsCanvas`, theres a dependency on `selectedElementIds` to generate an image from the story to test color accessibility against.

I've currently come up with two options for this, but trying to think of others that I like more:
1) add a `TextSetLibraryMoveable` wrapper around `LibraryMoveable` that calls the `useInsertTextSet` hook and passes the necessary callback into `LibraryMoveable`. Only use `TextSetLibraryMoveable` for text sets.
  - [pro] Just uses composition adjustments which are pretty straight forward
  - [pro] Fixes empty re-renders in shape panel
  - [con] Still will deal with the same empty re-renders in the text panel where `useInsertTextSet` will still be called in the scope of the `TextSetLibraryMoveable` component

2) Getting latest context values through a ref without imposing any forced renders
```
const selectedElementIdsRef([]);
useStorySelector(({ state }) => {
  selectedElementIdsRef.current = state.selectedElementIds;
  return true;
})
```
  - [pro] This will fix all empty re-renders from element selection in LibraryMoveable components used through out the library panels
  - [pro] Doesn't break any behavior since the `selectedElementIds` are just used for callbacks
  - [con] We're already dealing with the issue of putting side effects in our selectors with new versions of `use-context-selector` coming (#10528) . This could potentially compound tech debt here. It also may be fine with the proposed use case since the selector never triggers an update :shrug:

[...]

PR #10842, opened 2022-03-07
Performance: Fix empty renders of LibraryMoveable

### LINKED ISSUE #10730: Performance: Fix empty renders of LibraryMoveable
Found while doing discovery around [panel & library moveable perf](https://app.zenhub.com/workspaces/web-stories-5e94d3aced449034e1e9b226/issues/googleforcreators/web-stories-wp/10292)

**Fix Empty re-renders**
Currently there's empty re-renders stemming around pulling in `selectedElementIds` into `usePageAsCanvas` which ends up being in the component definition scope of `LibraryMoveable`. I went over some alternative options in the discovery ticket above, but I think we can best remedy this by pulling this state into a provider and using refs for the callback scope vs doing it in this hook. this should limit only the provider to re-rendering when selected elements update vs every component consuming `usePageAsCanvas`. This also leads to the next task which is kind of coupled with this architecturally.

**Remove usage of currentPageHash**
Looks like `currentPageHash` (and pageCanvasData) which also holds `currentPageHash` are only used in `library/panes/text/fontPreview.js` & `utils/usePageAsCanvas.js`

`JSON.stringify()`ing the whole page consumes a lot of memory and subsequently induces more garbage collection. As far as I can tell, this is just used to indicate that the page has changed. I think we can re-architecture this so we can just look at referential equality of page.elements updating or not to justify generating a new image to reference when doing color calculations.

We can probably mirror some of the stuff we're doing in the `PageDataUrlProvider` provider too and use something similar to `storyPageToDataUrl` to generate the page without passing a ref to an element around with the `CanvasProvider`.

In the worst case scenario where we can't eliminate the usage of `currentPageHash` and still maintain the same functionality, we can at least decouple its memoization from the other values it is generated with in the `useMemo` call so that it is generated less frequently

**Note**
This is kinda two separate tickets tied into one, but the technical implementation to achieve both the goals is kind of coupled, so just creating this larger ticket to accommodate for it

maxyinger:
## Context
Noticed while doing some performance work that we had different performance metrics for the same actions depending on what library panel was open. Specifically the shapes panel and the text panel around selection. Looked into it and found out it was our `LibraryMoveable` component. Theorized some possible approaches and the exploration in this ticket #10292.

Ultimately created this ticket #10730 as the best approach to fix the empty re-renders in the library panels. 

Also saw a possible win with the new approach to centralize some of our `html-to-canvas` usage. This sets up the foundation for that and hopefully we can share more generated canvases from story pages amongst components in the editor that need it. Created these follow tickets to look into that. Measured how long our `htmlToImage.toCanvas()` call took and on our test story with 50 shape elements it took ~1300ms, so any way we can minimize this call is a big win:

* #10822
* #10821
* #10824

Also with this new architecture we were able to eliminate some complexity in the logic, passing around of refs to a provider, and having to `JSON.stringify()` the entire story page every time it updated 🙌 . Relevant prior art here #10054

## Summary
Restructures how our smart colors are generating and passing around canvases to read color data off of them. Has a couple benefits that I'll go over more in the technical choices, but the big performance win here is limiting empty re-renders on selection when the text or shape panel are open.

| Selection update, text panel open | Selection update, shapes panel open |
| ---- | ----- | 
|  Screen Shot 2022-03-08 at 3 10 15 PM |  Screen Shot 2022-03-08 at 3 10 05 PM |

**Selection update, text panel open**
| Before | After |
| ----- | ----- |
| react commit  | react commit |
| ~46ms | ~26ms |
| Screen Shot 2022-03-08 at 2 47 01 PM |  Screen Shot 2022-03-08 at 2 48 01 PM |
| total js | total js |
| ~103ms | ~82ms |
| Screen Shot 2022-03-08 at 2 54 05 PM |  Screen Shot 2022-03-08 at 2 55 06 PM |

**Selection update, shape panel open**
| Before | After |
| ----- | ----- |
| react commit  | react commit |
| ~86ms | ~26ms |
| Screen Shot 2022-03-08 at 3 01 54 PM |  Screen Shot 2022-03-08 at 3 02 18 PM |
| total js | total js |
| ~145ms | ~87ms |
|  Screen Shot 2022-03-08 at 3 07 18 PM |  Screen Shot 2022-03-08 at 3 07 28 PM |

## Relevant Technical Choices
- Central provider responsible for holding and invalidating page canvas caches `PageCanvasProvider` 
- Creates & uses `storyPageToCanvas` util which simplifies some of the logic by keeping the current state of the user facing story canvas ui state agnostic from generating an image with the story markup.
- Makes `calculateAccessibleTextColors` stable throughout the life of the application
- Extracts & creates `useIdleQueue` from `pageUrlProvider`
- Simplifies cache invalidation logic to use react state vs storing & looking at `pageHash`
- Removes `pageHash`
- Removes `usePageAsCanvas`
- Cleans up some now unneeded state in the `CanvasProvider`

## User-facing changes
NA

## Testing Instructions
Go through and see that text smart colors behave the same as on `main`:
Screen Shot 2022-03-08 at 3 32 04 PM

**Adaptive color action**

1. add text on poor contrast backgrounds
2. select text
3. click on  `adaptive colors` action in the design panel with the paint bucket (highlighted above)
4. see that text becomes better contrast on the background

**Adaptive color toggle**

1. set the background color to black
2. toggle on `Adaptive Text Colors` in the library (highlighted above)
3. add a preset (like `Title 1`) and see that it gets applied to the page white
4. repeat on a white background and see that text is black

## Checklist

---

Fixes #10730


packages/story-editor/src/components/library/panes/text/textPane.js
```diff
@@ -85,7 +85,9 @@ function TextPane(props) {
   const { getPosition, insertPreset } = useInsertPreset({
     shouldUseSmartColor,
   });
-  const { generateCanvasFromPage } = usePageAsCanvas();
+  const generateDefferedCurrentPageCanvas = usePageCanvas(
+    ({ actions }) => actions.generateDefferedCurrentPageCanvas
```
swissspidy: Nit: deffered -> deferred

packages/story-editor/src/app/pageCanvas/getPixelDataFromCanvas.js
```diff
@@ -0,0 +1,47 @@
...
+        ? attrs.height
+        : calculateTextHeight(attrs, attrs.width),
+    },
+    canvas.width,
+    canvas.height - 2 * safeZoneDiff
+  );
+  const { x, y: origY, width, height } = box;
+  const y = origY + safeZoneDiff;
+  const pixelData = ctx.getImageData(x, y, width, height).data;
+  return pixelData;
```
maxyinger, line 44: didn't really touch any code here. this was logic was just moved directly into a function.

packages/story-editor/src/app/pageDataUrls/pageDataUrlsProvider.js
```diff
@@ -40,94 +32,32 @@ import Context from './context';
 
 function PageDataUrlProvider({ children }) {
   const [dataUrls, setDataUrls] = useState({});
-  const taskQueue = useRef([]);
-  const isTaskQueueRunning = useRef(false);
-  const currentTask = useRef([null, null]);
+  const queueIdleTask = useIdleTaskQueue();
```
maxyinger, line 35: I pretty much pulled out the IdleTaskQueue from this code so we could use it in other places. cut down on a lot of the code here, but I didn't alter the behavior of the Idle task queue at all, so most of that code hasn't really changed, just been moved around.

@swissspidy just pinging you here cus you had the suggestion initially and wanted to notify you that this is a util now available for consumption in other places

packages/story-editor/src/app/story/storyProvider.js
```diff
@@ -57,33 +57,26 @@ function StoryProvider({ storyId, initialEdits, children }) {
...
       return {
-        currentPageId: null,
-        currentPageIndex: null,
-        currentPageNumber: null,
-        currentPage: null,
-        currentPageHash: null,
+        currentPageId: current,
+        currentPageIndex: index,
+        currentPageNumber: number,
+        currentPage: page,
```
maxyinger: Didn't change anything here beyond removing the `currentPageHash` which we don't need anymore

maxyinger: we could honestly remove this memoization call directly too and just assign the variable directly in the block scope of the component and it wouldn't change the behavior at all.

This is because the actual page would maintain the same reference as it does with the memoization between renders and the rest of the values are primitives. also without the `JSON.stringify()` none of the assignment calls here are expensive, so no need to limit them

packages/story-editor/src/components/panels/design/textStyle/karma/textStyle.other.karma.js
```diff
@@ -86,9 +86,14 @@ describe('Text Style Panel', () => {
...
-      await fixture.events.click(
-        fixture.editor.canvas.framesLayer.frames[1].node
+      await fixture.events.mouse.clickOn(
+        fixture.editor.canvas.framesLayer.frames[1].node,
+        10,
+        10
       );
+      // await fixture.events.click(
+      //   fixture.editor.canvas.framesLayer.frames[1].node
+      // );
```
maxyinger, line 108: oops. I'll delete this. old remnant. my bad

packages/story-editor/src/storyEditor.js
```diff
@@ -82,16 +83,18 @@ function StoryEditor({ config, initialEdits, children }) {
...
+                                          <GlobalStyle />
+                                          <DevTools />
+                                          <DefaultMoveableGlobalStyle />
+                                          <CropMoveableGlobalStyle />
+                                          <ModalGlobalStyle />
+                                          <CalendarStyle />
+                                          <KeyboardOnlyOutlines />
+                                          {children}
+                                        </PageDataUrlProvider>
+                                      </PageCanvasProvider>
```
maxyinger, line 97: Just adding the `PageCanvasProvider` here. explicitly put it above the `PageDataUrlProvider` so we can eventually use a previously generated canvas that exists in the cache to get the saved template data url:

#10821

packages/story-editor/src/utils/storyPageToCanvas.js
```diff
@@ -0,0 +1,54 @@
...
+  const canvas = await htmlToImage.toCanvas(node, {
+    ...options,
+    fontEmbedCss: '',
+    pixelRatio: 1,
+  });
+
+  cleanup();
+
+  return canvas;
+}
```
maxyinger, line 52: mimics util for `PageUrlProvider` only difference here is we want to include the full bleed page aspect ratio for the generated canvas vs the 2/3 aspect ratio for the saved template thumbnail images.

packages/story-editor/src/utils/storyPageToNode.js
```diff
@@ -0,0 +1,176 @@
...
+
+  document.body.appendChild(bufferRoot);
+  const cleanup = () => {
+    unmountComponentAtNode(bufferRoot);
+    document.body.removeChild(bufferRoot);
+  };
+
+  return [node, cleanup];
+}
+export default storyPageToNode;
```
maxyinger, line 176: pulled out directly from `storyPageToDataUrl` so we can use it for both `storyPageToDataUrl`  & `storyPageToCanvas`. didn't change the behavior at all beyond exposing the `renderFullHeightThumb` prop so that the generated node could be full bleed or not

packages/story-editor/src/utils/useIdleTaskQueue.js
```diff
@@ -0,0 +1,117 @@
...
+        clearQueuedTask(id);
+      };
+    },
+    [runTaskQueue, clearQueuedTask]
+  );
+
+  return queueIdleTask;
+}
+
+export default useIdleTaskQueue;
```
maxyinger, line 117: pulled directly out of `PageDataUrlProvider`. didn't update the logic here at all. mainly just updated comments

packages/story-editor/src/utils/test/useIdleTaskQueue.js
```diff
@@ -0,0 +1,102 @@
...
+      expect(task2Tuple[1]).toHaveBeenCalledTimes(1);
+      expect(task3Tuple[1]).toHaveBeenCalledTimes(0);
+
+      await runAllIdleCallbacks();
+      expect(task1Tuple[1]).toHaveBeenCalledTimes(1);
+      expect(task2Tuple[1]).toHaveBeenCalledTimes(1);
+      expect(task3Tuple[1]).toHaveBeenCalledTimes(1);
+    });
+  });
+});
```
maxyinger, line 102: These unit tests are basically the same as the unit tests for `PageDataUrlProvider` because most of the tested logic in the `PageDataUrlProvider` is just testing the functionality of the `IdleTaskQueue`

maxyinger (commented):
Just commented on where I was moving around code so you can focus on actual changes 😄

packages/story-editor/src/components/library/panes/text/fontPreview.js
```diff
@@ -124,60 +117,33 @@ function FontPreview({ title, element, insertPreset, getPosition, index }) {
...
-      // If the element is positioned under the previous element (not default position),
-      // no new image generation needed.
-      if (atts.y !== element.y) {
-        presetDataRef.current.skipCanvasGeneration = true;
-      }
-
-      presetDataRef.current.autoColor = await calculateAccessibleTextColors(
-        { ...element, ...atts },
-        false /* isInserting */
+  const onClick = useCallback(async () => {
```
miina, line 151: The `onClick` of `FontPreview` has some updates so this will need updating with `main` now.

miina (approved):
LGTM and as far as I can see everything works as previously but would definitely be good to have more eyes on this. 

There are some tests related to smart color (e.g. `should add text color based on background`) which have a delay of `800` ms currently to wait for canvas generation + color generation. Wondering if those delays could be reduced now?

packages/story-editor/src/app/pageCanvas/pageCanvasProvider.js
```diff
@@ -0,0 +1,205 @@
...
+import getPageWithoutSelection from './getPageWithoutSelection';
+
+/**
+ * @typedef {import('../../../types').Page} Page
+ */
+
+function PageCanvasProvider({ children }) {
+  const queueIdleTask = useIdleQueue();
+  const [pageCanvasMap, actions] = usePageCanvasMap();
+  const { getSnapshotCanvas, setSnapshot } = usePageSnapshot();
```
samwhale: 💭 was really helpful to chat with you about why these two caches are here. Would be awesome to get some documentation here (if there isn't any already).

I found it hard to understand why these were here until you walked me through it, so maybe others will too!

packages/story-editor/src/utils/useIdleTaskQueue.js
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    const [taskId, task] = taskQueue.current.shift();
+    const idleCallbackId = requestIdleCallback(async () => {
+      await task();
+      currentTask.current = [null, null];
+      runTaskQueue();
+    });
+    currentTask.current = [taskId, idleCallbackId];
+  }, []);
```
samwhale, line 57: Cool to see that we have a fallback in `requestIdleCallback` for safari users too: https://github.com/GoogleForCreators/web-stories-wp/blob/c0a81152b0174c7daa8ff6f6bc1f93047d129431/packages/story-editor/src/utils/idleCallback.js#L18-L37

packages/story-editor/src/app/pageCanvas/pageCanvasProvider.js
```diff
@@ -0,0 +1,205 @@
...
+      value={{
+        state: {
+          pageCanvasMap,
+        },
+        actions: {
+          calculateAccessibleTextColors,
+          generateDeferredPageCanvas,
+          generateDeferredCurrentPageCanvas,
+        },
+      }}
```
samwhale, line 191: 💭 does it matter if the object on `178` is being re-created every render? I guess it shouldn't matter as long as the references in `state` and `actions` are being pulled out of the object, and the parent object reference isn't being compared directly in children, right?

packages/story-editor/src/app/pageCanvas/test/usePageCanvas.js
```diff
@@ -0,0 +1,270 @@
...
+}
+
+describe('usePageCanvas', () => {
+  let mockStoryContext;
+  let currentPage;
+
+  beforeEach(() => {
+    storyPageToCanvas.mockImplementation(
+      () =>
+        'lets pretend this is a canvas. I wont tell the computer if you dont 🤫'
```
samwhale, line 62: lmao

packages/story-editor/src/app/pageCanvas/test/usePageCanvas.js
```diff
@@ -0,0 +1,270 @@
...
+
+        // cache populates with generated canvas for currentPage
+        await act(async () => {
+          const [usePageCanvasCurrent] = result.current;
+          await usePageCanvasCurrent.actions.calculateAccessibleTextColors(
+            element
+          );
+        });
+        expect(storyPageToCanvas).toHaveBeenCalledTimes(1);
+      });
```
samwhale: 💭 should this also check to see if the canvas cache is not populated at the beginning?

packages/story-editor/src/app/pageCanvas/test/usePageCanvas.js
```diff
@@ -0,0 +1,270 @@
...
+        // calculateAccessibleTextColors knows to exclude the element we're testing
+        // from the generated canvas. It holds this canvas generated from a partial
+        // of the page in the snapshot cache.
+        await act(async () => {
+          const [usePageCanvasCurrent] = result.current;
+          await usePageCanvasCurrent.actions.calculateAccessibleTextColors(
+            selectedElement
+          );
+        });
+        expect(storyPageToCanvas).toHaveBeenCalledTimes(1);
```
samwhale, line 177: 🤔 I guess it's not possible to test that a partial snapshot vs a whole snapshot was created since both use the same `storyPageToCanvas` fn, right?

packages/story-editor/src/app/story/storyProvider.js
samwhale: let's do that!

packages/story-editor/src/utils/storyPageToNode.js
```diff
@@ -0,0 +1,176 @@
...
+ * @typedef {import('../../../types').Page} Page
+ */
+
+/**
+ * Takes a story page and generates a DOM node containing the rendered
+ * page. Returns a tuple containing the page DOM node and a cleanup
+ * method to remove the returned DOM node.
+ *
+ * **IMPORTANT:** Not calling the returned `cleanup()` method after use of
+ * page DOM node will result in memory leak.
```
samwhale, line 122: 👍

samwhale (approved):
thanks for walking us through this last week. Looking through it a second time after seeing the motivation for changes (especially the addition of the two caches) helped to make this easier to understand.

maxyinger:
> There are some tests related to smart color (e.g. should add text color based on background) which have a delay of 800 ms currently to wait for canvas generation + color generation. Wondering if those delays could be reduced now?

I don't think there's any speed increase here. all the changes relative to this were just for code quality. We're still generating a new canvas for both the actions in the specified karma test, so just gonna leave it as is at 800 for now

packages/story-editor/src/app/pageCanvas/pageCanvasProvider.js
maxyinger: correct. If we're using selectors well, there's no difference if you were to memoize the object held in the providers value

packages/story-editor/src/app/pageCanvas/test/usePageCanvas.js
maxyinger: sure we can add it in there

packages/story-editor/src/app/pageCanvas/test/usePageCanvas.js
maxyinger: we could look at the story page it was called with and compare it to the story we're using

packages/story-editor/src/components/panels/design/textStyle/karma/textStyle.other.karma.js
swissspidy: @littlemilkstudio friendly ping :) before we merge this by accident

swissspidy (approved):
Kudos, nice work!

swissspidy:
Looking at the Percy snapshots now...

Looks like the backdrop from the media edit mode now goes all the way over the carousel? Is that new?

Screenshot 2022-03-16 at 09 50 47

See https://percy.io/google/web-stories-wp/builds/16492104/changed/928097483

miina:
> Looking at the Percy snapshots now...
> Looks like the backdrop from the media edit mode now goes all the way over the carousel? Is that new?

@swissspidy I think #10929 will fix it (it was from before already)

PR #10844, opened 2022-03-08
Floating menu: Text color

### LINKED ISSUE #10632: UI Simplification: Floating menu: Font color
## Feature Description

Make the font color element in the floating design menu functional.

Design: [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=11394%3A457149)
Menus: Text
Notes: This should re-use the logic for a dropdown from elsewhere but note that this color element does not include a hex input field.

## Additional information

text-color

miina:
## Summary
- Adds color picker to the text element floating menu to choose the text color.

## To-do

- [x] Tests

## User-facing changes
When there are multiple colors in a text:
Screenshot 2022-03-09 at 13 15 03

When there is single color in the text
Screenshot 2022-03-09 at 13 14 18

## Testing Instructions

1. Add a text
2. Verify you can assign color to the text from the floating menu
3. Verify the eyedropper also works

Note: general known bug: eyedropper doesn't close unless the user chooses a color.

Tracks when color is set from the menu:
```
trackEvent('floating_menu', {
        name: 'set_text_color',
        element: 'text',
      });
```

No

## Checklist

---

Fixes #10632


packages/story-editor/src/components/form/color/colorInput.js
```diff
@@ -242,9 +248,19 @@ const ColorInput = forwardRef(function ColorInput(
                 </Text>
               </TextualPreview>
             ) : (
-              <ChevronContainer>
-                <Icons.ChevronDown />
-              </ChevronContainer>
+              <>
+                {/* In case of design menu, we display Mixed value even without inputs */}
+                {isInDesignMenu && isMixed && (
```
miina: @barklund Per UX, we should display `Mixed` in the menu if the text includes more than one color, and it should look similar to the Design Panel.

PTAL at these changes here -- feels a bit hacky but not sure if there's a better way.

packages/story-editor/src/components/floatingMenu/elements/textColor.js
```diff
@@ -17,20 +17,51 @@
...
  * Internal dependencies
  */
-import { IconButton } from './shared';
+import useRichTextFormatting from '../../panels/design/textStyle/useRichTextFormatting';
+import updateProperties from '../../inspector/design/updateProperties';
+import { useStory } from '../../../app';
+import { Color, useProperties } from './shared';
 
 function TextColor() {
+  const { content, type } = useProperties(['content', 'type']);
```
barklund: Why select `type`? It's always `'text'`?

packages/story-editor/src/components/form/color/colorInput.js
barklund: The chevron is located wrong now. It should be vertically centered, but is too high now (both with and without the mixed label):

| Before | After |
|-|-|
| Screen Shot 2022-03-08 at 15 22 06 | Screen Shot 2022-03-08 at 15 24 15 |


As for the code, I think we can remove the `isInDesignMenu && ` and just say that we always show mixed even if inputs are off. Currently this would only happen in the design menu anyway, but if we were later to have no-input color pickers elsewhere, the same thing should probably happen.

packages/story-editor/src/components/floatingMenu/elements/textColor.js
```diff
@@ -17,20 +17,51 @@
...
+  const pushUpdate = useCallback(
+    (update) => {
+      updateSelectedElements({
+        properties: (element) => {
+          const updates = updateProperties(element, update, true);
+          return {
+            ...element,
+            ...updates,
+          };
+        },
```
barklund: I'm pretty sure the updates are merged into the existing object, so no need to spread the old object:

```js
        properties: (element) => updateProperties(element, update, true),
```

packages/story-editor/src/components/floatingMenu/elements/textColor.js
```diff
@@ -17,20 +17,51 @@
...
-      onClick={() => {}}
+    <Color
+      label={__('Text color', 'web-stories')}
+      value={color}
+      allowsSavedColors
+      onChange={handleSetColor}
+      hasInputs={false}
+      hasEyeDropper
+      allowsOpacity={false}
     />
```
barklund: I feel we should incorporate the warning about bad contrast into the design menu as well. Either directly in the color picker or as a warning next to the color input in the design menu. Because when we move the design panel to a (mostly hidden) tab, users won't see it anymore.

Screen Shot 2022-03-08 at 15 31 35

But that's probably a follow-up ticket needing UX lead.

barklund (approved):
Works great, the design is just a bit off.

packages/story-editor/src/components/floatingMenu/elements/textColor.js
```diff
@@ -17,20 +17,51 @@
...
-      title={__('Change text color', 'web-stories')}
-      onClick={() => {}}
+    <Color
+      label={__('Text color', 'web-stories')}
+      value={color}
+      allowsSavedColors
+      onChange={handleSetColor}
+      hasInputs={false}
+      hasEyeDropper
+      allowsOpacity={false}
```
merapi: Shouldn't that be `true`? in the Design panel, you can choose text color with opacity.

image
image

merapi (commented):
Not sure if it belongs here, but those Mixed values should look the same?

image

Should the eyedropper also be here?

image

packages/story-editor/src/components/floatingMenu/elements/textColor.js
miina: Oh, good catch! I remembered that it was not allowed but actually just gradients aren't allowed 💡

miina:
> Not sure if it belongs here, but those Mixed values should look the same?

I'm not sure about it -- initially I made it the same but then reverted. In the design panel, 'Mixed'  is a placeholder, however, in the design menu it's just for information, not an input. So seemed like maybe better not to have it the same color as placeholders.

> Should the eyedropper also be here?

Yep, I actually thought this would be a separate issue 🤔 But looks like it's not!

merapi:
Is that a known bug? I couldn't find any related ticket.

https://user-images.githubusercontent.com/21036224/157419541-c6348cd9-76cb-4b53-9b7d-cb80636b34cd.mov

miina:
> Is that a known bug? I couldn't find any related ticket.

Yes, that's a known bug and it got fixed within #10845 -- haven't pushed the merge yet.

miina:
@merapi Your comments should be addressed now, thank you!

One note not related this PR: It looks like clicking "Esc" de-selects the element when in eyedropper mode but doesn't stop eyedropping. Do you remember if there was a way to stop eyedropping that was initially implemented? Not sure if there was a regression somewhere or it's a bug that wasn't noticed.

packages/story-editor/src/components/floatingMenu/layer.js
```diff
@@ -73,6 +77,15 @@ function FloatingMenuLayer() {
     return () => setMoveableMount(null);
   }, [setMoveableMount]);
 
+  // When eyedropper is active, always hide the floating menu, too.
+  useEffect(() => {
```
miina, line 81: @barklund FYI Also added this here to dismiss the floating menu when eyedropping a color.

packages/story-editor/src/components/floatingMenu/elements/textColor.js
```diff
@@ -17,20 +17,50 @@
...
 
 function TextColor() {
+  const { content } = useProperties(['content']);
+  const updateSelectedElements = useStory(
+    (state) => state.actions.updateSelectedElements
+  );
+
+  const pushUpdate = useCallback(
+    (update) => {
+      trackEvent('floating_menu', {
```
miina, line 40: Added tracking, too -- maybe it'll be useful to detect how much the menu is used. Should we add these for all buttons? Added it to replacing media as well but it's currently nowhere else. Thoughts?

packages/story-editor/src/components/floatingMenu/elements/textColor.js
miina: Indeed! 💡

packages/story-editor/src/components/form/color/colorInput.js
miina: Makes sense, done!

packages/story-editor/src/components/floatingMenu/elements/textColor.js
miina: Added issue: https://github.com/GoogleForCreators/web-stories-wp/issues/10868

packages/story-editor/src/components/floatingMenu/elements/textColor.js
barklund: Uh, good idea. Let's add a ticket for that!

miina: Added: https://github.com/GoogleForCreators/web-stories-wp/issues/10871

PR #10847, opened 2022-03-08
Sidebar: New Layers Panel

### LINKED ISSUE #10804: Sidebar: Move Layers panel out of the Sidebar
## Feature Description

PR should be based on/merged to the `sidebar` branch.

merapi:
## Context

Part of editor redesign. 

## Summary

Lifts the layers panel from the inspector pane (which was combined with the design pane to create the new sidebar) and puts it opposite the sidebar as part of the editor's footer. 

## Relevant Technical Choices

Added a few more props to the shared `ToggleButton` that's used for the many popups of the footer to add a variant with copy visible/no icon and prevent tooltip from showing. 

Added `<DirectionAware>` to the footer so that the padding there is consistent and the layers panel wouldn't require additional work to be readable in RTL. I think it looks better now. 

## To-do

- [x] Fix tests
- [x] Adjust colors
- [x] Check a11y

Opted against moving all of the layers panel files into their own directory (following the suit of the helpcenter, keyboard nav and checklist) because it makes it harder to grok the updates.

Talked to @agingoldseco about the new toggle design, we're going to do that [later as a nonblocking issue.](https://github.com/GoogleForCreators/web-stories-wp/issues/10891)

The layer panel accessibility is not an issue here, it would be introducing new use patterns should we expose it to the keyboard at this point which would bloat the scope significantly. It was discussed at length 5 months ago in the design slack channel. There are keyboard shortcuts for everything that the layers panel can do. If we decide that now the layers panel should be keyboard accessible that would be its own scope of work and own item on the roadmap. 

## User-facing changes

| LTR | RTL | 
| --- | --- | 
| Screen Shot 2022-03-11 at 11 23 36 AM | Screen Shot 2022-03-11 at 11 24 13 AM |

With Meta boxes 
| LTR | RTL | 
| --- | --- | 
| Screen Shot 2022-03-11 at 1 38 26 PM | Screen Shot 2022-03-11 at 1 38 37 PM |

## Testing Instructions

Use the layers panel, now on the opposite side as the sidebar. 
You should be able to expand and collapse it with cursor and keyboard.
Confirm layer selection and behavior.
Check LTR and RTL. 
Install and activate 'yoast seo' to see meta box show up. 

1.

## Checklist

---

Fixes #10804

### DIFF
--- a/packages/story-editor/src/components/footer/footerLayout.js
+++ b/packages/story-editor/src/components/footer/footerLayout.js
@@ -28,6 +28,7 @@ import Carousel from './carousel';
 import PrimaryMenu from './primaryMenu';
 import SecondaryMenu from './secondaryMenu';
 import { MAX_MENU_WIDTH } from './constants';
+import DirectionAware from '../directionAware';
 
 const Wrapper = styled.section`
   position: relative;
@@ -52,17 +53,19 @@ const Area = styled.div`
 
 function FooterLayout({ footer }) {
   return (
-    <Wrapper aria-label={__('Workspace Footer', 'web-stories')}>
-      <Area area="carousel">
-        <Carousel />
-      </Area>
-      <Area area="primary">
-        <PrimaryMenu />
-      </Area>
-      <Area area="secondary">
-        <SecondaryMenu menu={footer?.secondaryMenu} />
-      </Area>
-    </Wrapper>
+    <DirectionAware>
+      <Wrapper aria-label={__('Workspace Footer', 'web-stories')}>
+        <Area area="carousel">
+          <Carousel />
+        </Area>
+        <Area area="primary">
+          <PrimaryMenu />
+        </Area>
+        <Area area="secondary">
+          <SecondaryMenu menu={footer?.secondaryMenu} />
+        </Area>
+      </Wrapper>
+    </DirectionAware>
   );
 }
 
--- a/packages/story-editor/src/components/footer/footerLayout.js
+++ b/packages/story-editor/src/components/footer/footerLayout.js
@@ -24,11 +24,11 @@ import PropTypes from 'prop-types';
 /**
  * Internal dependencies
  */
+import DirectionAware from '../directionAware';
 import Carousel from './carousel';
 import PrimaryMenu from './primaryMenu';
 import SecondaryMenu from './secondaryMenu';
 import { MAX_MENU_WIDTH } from './constants';
-import DirectionAware from '../directionAware';
 
 const Wrapper = styled.section`
   position: relative;
--- a/packages/story-editor/src/components/footer/layers/layers.js
+++ b/packages/story-editor/src/components/footer/layers/layers.js
@@ -51,7 +51,7 @@ function Layers() {
     <>
       <Popup
         isOpen={isOpen}
-        anchor="right"
+        placement="right"
         ariaLabel={__('Layers Panel', 'web-stories')}
       >
         <StyledNavigationWrapper ref={ref} isOpen={isOpen}>
--- a/packages/story-editor/src/components/footer/layers/layers.js
+++ b/packages/story-editor/src/components/footer/layers/layers.js
@@ -1,0 +1,112 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * External dependencies
+ */
+import { useState } from '@googleforcreators/react';
+import { __ } from '@googleforcreators/i18n';
+import {
+  Icons,
+  Text,
+  NotificationBubble,
+  THEME_CONSTANTS,
+  BUBBLE_VARIANTS,
+} from '@googleforcreators/design-system';
+
+/**
+ * Internal dependencies
+ */
+import styled, { css } from 'styled-components';
+import { LayerPanel } from '../../panels/design';
+import useLayers from '../../panels/design/layer/useLayers';
+import Popup from '../../secondaryPopup';
+
+const Container = styled.div`
+  position: absolute;
+  right: 0;
+  bottom: 0;
+  width: 300px;
+  border-radius: ${({ theme }) => theme.borders.radius.small};
+  background-color: ${({ theme }) => theme.colors.bg.secondary};
+  overflow: auto;
+`;
+
+export const LabelText = styled(Text)`
+  color: ${({ theme }) => theme.colors.fg.primary};
+  padding-right: 8px;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: hidden;
+`;
+
+const Button = styled.button`
+  border: 1px solid ${({ theme }) => theme.colors.border.defaultNormal};
+  background-color: ${({ theme }) => theme.colors.opacity.footprint};
+  border-radius: ${({ theme }) => theme.borders.radius.small};
+  padding: 5px 8px;
+  cursor: pointer;
+  display: flex;
+`;
+
+export const ChevronWrap = styled.div(
+  ({ theme, isOpen }) => css`
+    color: ${theme.colors.fg.primary};
+    width: 32px;
+    height: 32px;
+    margin: -4px;
+
+    ${isOpen &&
+    css`
+      transform: rotate(180deg);
+    `}
+  `
+);
+
+function Layers() {
+  const layers = useLayers();
+  const [isOpen, setIsOpen] = useState(false);
+
+  return (
+    <>
+      <Popup isOpen={isOpen} anchor="right" zIndex={9999}>
+        <Container>
+          <LayerPanel />
+        </Container>
+      </Popup>
+      <Button
+        aria-haspopup
+        ariaLabel={__('Layers', 'web-stories')}
+        onClick={() => setIsOpen((state) => !state)}
+      >
+        <LabelText size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL}>
+          {__('Layers', 'web-stories')}
+        </LabelText>
+        <NotificationBubble
+          data-testid="panel-badge"
+          notificationCount={layers?.length}
+          variant={BUBBLE_VARIANTS.SECONDARY}
+          aria-hidden
+        />
+        <ChevronWrap isOpen={isOpen}>
+          <Icons.ChevronDownSmall />
+        </ChevronWrap>
+      </Button>
+    </>
+  );
+}
+
+export default Layers;
--- a/packages/story-editor/src/components/footer/layers/layers.js
+++ b/packages/story-editor/src/components/footer/layers/layers.js
@@ -33,7 +33,6 @@ const LAYERS_PANEL_WIDTH = 300;
 
 const StyledNavigationWrapper = styled(NavigationWrapper)`
   width: ${LAYERS_PANEL_WIDTH}px;
-  right: inherit;
   left: -${LAYERS_PANEL_WIDTH}px;
 `;
 
@@ -53,7 +52,6 @@ function Layers() {
       <Popup
         isOpen={isOpen}
         anchor="right"
-        zIndex={Z_INDEX_FOOTER}
         ariaLabel={__('Layers Panel', 'web-stories')}
       >
         <StyledNavigationWrapper ref={ref} isOpen={isOpen}>
--- a/packages/story-editor/src/components/footer/layers/layers.js
+++ b/packages/story-editor/src/components/footer/layers/layers.js
@@ -29,13 +29,6 @@ import useLayers from '../../panels/design/layer/useLayers';
 import Popup, { NavigationWrapper } from '../../secondaryPopup';
 import { ToggleButton } from '../../toggleButton';
 
-const LAYERS_PANEL_WIDTH = 300;
-
-const StyledNavigationWrapper = styled(NavigationWrapper)`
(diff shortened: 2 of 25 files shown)


packages/story-editor/src/components/secondaryPopup/index.js
```diff
@@ -46,13 +46,17 @@ const transitionStyles = {
 const Controller = styled.div`
   position: absolute;
   top: -12px;
-  left: 0px;
+  ${({ zIndex }) => (zIndex ? `z-index: ${zIndex};` : '')}
```
merapi: I had to set a higher z-index here (didn't want to change z-indexes in Page Side Menu) in order to prevent this:

https://user-images.githubusercontent.com/21036224/157312306-b299a49e-8688-4822-9b83-754525af740b.mov

packages/story-editor/src/components/footer/primaryMenu.js
```diff
@@ -50,6 +51,7 @@ function PrimaryMenu() {
       <MenuItems id="primary-menu-items">
         <ZoomSelector />
         <GridViewButton />
+        <Layers />
```
barklund, line 56: If you have metaboxes (from other plugins), the alignment is a bit off (between layers button and layers panel).

IMO the layers panel button should be last (right-most) and the metaboxes button should be moved a bit, but UX might have a better suggestion.

Screen Shot 2022-03-08 at 15 37 31

merapi: I've noticed that panels on the left side of the footer are not always aligned to the button:

Checklist:
image

Shortcuts:
image

Not sure why - maybe to save the canvas overflow? that might be especially useful for the Layers panel that will be visible more often than the Shortcuts (so we want to cover as little of the canvas as possible).
I will ask Design pod about that.

packages/story-editor/src/components/secondaryPopup/index.js
```diff
@@ -79,6 +85,8 @@ function Popup({
           role="dialog"
           aria-label={ariaLabel}
           state={state}
+          anchor={anchor}
```
miina: In the design-system `Popup`, anchor is a `ref` of the element that has the popup and `spacing` in the format of `{ x, y }` is used for spacing and `placement` is used to define if it's on the left (or somewhere else) relative to the anchor. Could we use the same logic here for consistency? E.g. could we use `spacing` with the relevant values and/or `placement` instead of `anchor`. Thoughts?

packages/story-editor/src/components/panels/design/layer/layerPanel.js
```diff
@@ -17,37 +17,24 @@
...
  * Internal dependencies
  */
-import { Panel, PanelTitle, PanelContent } from '../../panel';
+import { PanelContent } from '../../panel';
 import useInspector from '../../../inspector/useInspector';
-import { TAB_HEIGHT, TAB_VERTICAL_MARGIN } from '../../../tabview';
 import LayerList from './layerList';
 import useLayers from './useLayers';
 
+const HEAD_AREA_HEIGHT = 64;
```
miina: Is the `HEAD_AREA_HEIGHT` related to the WordPress admin menu? Or does it include the WP admin menu + the buttons? Or just buttons?

In case it includes the WP admin menu, the value should come from `useConfig`,  e.g.
```
const { styleConstants: { topOffset } = {} } = useConfig();
```

If it's the menu area with buttons, let's use the `HEADER_HEIGHT` from the general constants file instead.

packages/story-editor/src/components/footer/layers/layers.js
```diff
@@ -0,0 +1,112 @@
...
+
+export const LabelText = styled(Text)`
+  color: ${({ theme }) => theme.colors.fg.primary};
+  padding-right: 8px;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: hidden;
+`;
+
+const Button = styled.button`
```
miina: Hm, we should really use a button from the design-system here -- whenever there's a button without variant in Figma, would be good to confirm with UX to check which button it is. The only thing that I can see in Figma right now is that it's a rectangle button of size small, unfortunately variant hasn't been defined.

Otherwise we'd need to also add hover/focus etc. here separately.
From [Figma](https://www.figma.com/file/Ha6kk9yXDN1rB80NzcDf2J/_Editor---Design-System-%22Style-Guide%22?node-id=3821%3A81609), it looks like it's a quaternary button, let's use that instead. Then we also wouldn't probably need to style the `LabelText` separately.

merapi: It makes sense, I've looked at other buttons in the footer and they all didn't match this one so I've started to create my own style (just got the permissions to access this Figma file you linked) - I will see what can be done here :)

packages/story-editor/src/components/footer/layers/layers.js
```diff
@@ -0,0 +1,112 @@
...
+  `
+);
+
+function Layers() {
+  const layers = useLayers();
+  const [isOpen, setIsOpen] = useState(false);
+
+  return (
+    <>
+      <Popup isOpen={isOpen} anchor="right" zIndex={9999}>
```
BrittanyIRL: setting the zIndex here to 9999 is alot. I'm addressing the zIndex here: https://github.com/GoogleForCreators/web-stories-wp/pull/10861 - following this format would be very helpful. i'm currently blocked by Karma tests from merging it. You can follow the same pattern as the other footer popups. I'll write up more specifically what this is here shortly.

packages/story-editor/src/components/footer/primaryMenu.js
BrittanyIRL: is this in figma?

packages/story-editor/src/components/footer/layers/layers.js
BrittanyIRL: So this is actually more in line with the other popups in the footer: 
Screen Shot 2022-03-10 at 10 53 16 AM

BrittanyIRL: The only reason the caret is on the zoom is because it's a dropdown(up) menu. I was able to reuse the `ToggleButton` that the rest of the footer popups utilize with only a few modifications. I'll screenshot the diff here. 

I'd like to request you just leave the z index be and not further tinker with it. It's been really hard to untangle and my PR will resolve your issues with it. 
Screen Shot 2022-03-10 at 10 55 36 AM
Screen Shot 2022-03-10 at 10 55 25 AM

miina: Good point that it's actually a Toggle!

BrittanyIRL: Amy might be fine with the design change to cater to consistency. 

That said, if one notification bubble has a bg and a circle then they _all_ should and that could be its own lower priority ticket.

packages/story-editor/src/components/footer/layers/layers.js
barklund: `Number.MAX_SAFE_INTEGER` **FTW** 🚀 🌔 .

packages/story-editor/src/components/footer/primaryMenu.js
BrittanyIRL: I will talk to Amy about alignment.

packages/story-editor/src/components/footer/footerLayout.js
```diff
@@ -52,17 +53,19 @@ const Area = styled.div`
...
-        <Carousel />
-      </Area>
-      <Area area="primary">
-        <PrimaryMenu />
-      </Area>
-      <Area area="secondary">
-        <SecondaryMenu menu={footer?.secondaryMenu} />
-      </Area>
-    </Wrapper>
+    <DirectionAware>
```
BrittanyIRL, line 56: Added `<DirectionAware>` to make spacing consistent in RTL, otherwise the layers panel toggle is right up against the edge of the browser

packages/story-editor/src/components/footer/layers/layers.js
```diff
@@ -0,0 +1,74 @@
...
+import { LayerPanel } from '../../panels/design';
+import useLayers from '../../panels/design/layer/useLayers';
+import Popup, { NavigationWrapper } from '../../secondaryPopup';
+import { ToggleButton } from '../../toggleButton';
+
+const LAYERS_PANEL_WIDTH = 300;
+
+const StyledNavigationWrapper = styled(NavigationWrapper)`
+  width: ${LAYERS_PANEL_WIDTH}px;
+  left: -${LAYERS_PANEL_WIDTH}px;
```
BrittanyIRL: `SecondaryPopup` was created with the left (in LTR) side of the footer in mind, so everything is aligned to the left. This is quick and dirty way to align to the right side. It's the only one that does this so I'd rather not dirty up the shared popup for 1 case

packages/story-editor/src/components/footer/layers/layers.js
```diff
@@ -0,0 +1,74 @@
...
+`;
+
+const Container = styled.div`
+  background-color: ${({ theme }) => theme.colors.bg.secondary};
+  overflow: auto;
```
BrittanyIRL, line 34: ensures scroll

packages/story-editor/src/components/panels/design/layer/layerPanel.js
BrittanyIRL: was able to ditch this because the shared popup is accounting for available height already.

packages/story-editor/src/components/secondaryPopup/index.js
BrittanyIRL: We should be good here now. @merapi 's idea to add anchor to control the left vs right value was great, that combined with an override in the layer set up and we're gucci. No need to further tweak anything.

packages/story-editor/src/components/footer/primaryMenu.js
merapi: Here is the Slack discussion: https://xwp.slack.com/archives/C01ESBB8ZN1/p1646935295628609?thread_ts=1646930940.483089&cid=C01ESBB8ZN1

packages/story-editor/src/components/footer/layers/layers.js
merapi: Just pasting the last input from Amy:

image

(so Q-styled button with up/down arrow).
But now since you made the change maybe just double-check with her.

packages/story-editor/src/components/footer/layers/layers.js
merapi: I thought `9999` is some kind of magic number, we use it a lot 😉  (I guess it just spread like a virus)

image

packages/story-editor/src/components/footer/primaryMenu.js
BrittanyIRL: Talked to amy about what I have for this and she's fine with it as is. Alignment is either to the left or the right, not the center ever. Looks good with and without meta buttons - I added screenshots to the description.

packages/story-editor/src/components/footer/layers/layers.js
BrittanyIRL: Chatted with amy, thanks for the screenshot of your chat it's helpful (slack won't open the links). 

we're going to handle the chevron/updates from the existing toggle button separate to unblock work and be consistent with other buttons with same behavior. (#10891 )

packages/story-editor/src/components/secondaryPopup/index.js
miina: Hm, still not convinced about the naming.

Since `anchor` is a `ref` in the "default" `Popup` then could we for consistency use different naming for the placement here? E.g. why not use `placement` instead as in the default `Popup`? 🤔 Or `anchorPlacement` at least?

The logic itself seems all good but just the naming is inconsistent with the general `Popup`. Thoughts?

packages/story-editor/src/components/footer/layers/layers.js
maxyinger: we could also do a default prop in `NavigationWrapper` that would allow us to not have to edit any other instance like so:
```
const NavigationWrapper = styled.div`
  ...
  ${({ alignRight = false }) => (alignRight ? 'right: 0' : 'left: 0')}
```
But I'm cool with this as is since it's pretty isolated and straight forward.

A third option here that I think would be the equivalent is just doing:
```
const StyledNavigationWrapper = styled(NavigationWrapper)`
  right: 0;
  left: auto;
```

packages/story-editor/src/components/panels/design/layer/layer.js
```diff
@@ -112,35 +112,35 @@ const LayerButton = styled(Button).attrs({
...
-    background: ${({ theme }) => theme.colors.interactiveBg.secondaryPress};
+    background: ${({ theme }) => theme.colors.interactiveBg.tertiaryPress};
   }
   :active + * {
     --background-color: ${({ theme }) =>
-      theme.colors.interactiveBg.secondaryPress};
+      theme.colors.interactiveBg.tertiaryPress};
     --background-color-opaque: ${({ theme }) =>
-      rgba(theme.colors.interactiveBg.secondaryPress, 0)};
+      rgba(theme.colors.interactiveBg.tertiaryPress, 0)};
```
maxyinger, line 142: apologies for the styling here being a little complicated 😅 

Looks like something is out of sync here tho with the hover styles over the layer actions:
| Staging | This Branch |
| --- | ---|
| Screen Shot 2022-03-14 at 9 04 51 AM |  Screen Shot 2022-03-14 at 9 05 03 AM |

Looks like it is do to the background color of the popup changing.

maxyinger (approved):

packages/story-editor/src/components/secondaryPopup/index.js
samwhale: I'll update this to `placement`

packages/story-editor/src/components/footer/layers/layers.js
```diff
@@ -0,0 +1,67 @@
...
+  const ref = useRef();
+
+  return (
+    <>
+      <Popup
+        isOpen={isOpen}
+        placement="right"
+        ariaLabel={__('Layers Panel', 'web-stories')}
+      >
+        <NavigationWrapper alignRight ref={ref} isOpen={isOpen}>
```
BrittanyIRL, line 50: like the `alignRight` update

BrittanyIRL (approved):
Thanks for getting this finished, @samwhale !!

samwhale (approved):

PR #10858, opened 2022-03-08
UI Simplification: Floating menu - hook up 'More' button

### LINKED ISSUE #10761: UI Simplification: Floating menu: More button
## Feature Description

Make the more button element in the floating design menu functional. Clicking this button should open and highlight the design panel for the current selection.

Design: [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=11394%3A457149)
Menus: All single-selection (not multiple)
Notes: Should work for all single-selection menus and might have to work a little differently for each depending on the design panel tab.

## Additional information
more-button

samwhale:
## Summary

Hook up `More` button in the floating menu.

## Relevant Technical Choices

When adding the highlight functionality to the size and position panel (titled `Selection` in the UI), saw that other code added event listeners in the ref callback. This is rather strange - and react provides an api for hooking into events, so I went with that instead.

## To-do

n/a

## User-facing changes

more-button

## Testing Instructions

1. enable floating menu experiment
2. Add any element
3. click on element to show floating menu
4. click `more` button

Should highlight the size and position (titled `Selection`) panel and focus the first input.

Note: the floating menu isn't shown for background images.

## Checklist

---

Fixes #10761

### DIFF
--- a/packages/story-editor/src/app/highlights/states.js
+++ b/packages/story-editor/src/app/highlights/states.js
@@ -35,6 +35,9 @@ import {
 const keys = {
   STORY_TITLE: 'STORY_TITLE',
 
+  // Inspector tabs
+  STYLE_PANE: 'STYLE_PANE',
+
   // INSPECTOR
   ANIMATION: 'ANIMATION',
   ASSISTIVE_TEXT: 'ASSISTIVE_TEXT',
@@ -59,6 +62,12 @@ const keys = {
 export const STATES = {
   [keys.STORY_TITLE]: {
     focus: true,
+  },
+
+  // Inspector tabs
+  [keys.STYLE_PANE]: {
+    focus: true,
+    tab: STYLE,
   },
 
   // Inspector
--- a/packages/story-editor/src/components/floatingMenu/elements/karma/more.karma.js
+++ b/packages/story-editor/src/components/floatingMenu/elements/karma/more.karma.js
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * External dependencies
+ */
+import STICKERS from '@googleforcreators/stickers';
+/**
+ * Internal dependencies
+ */
+import { Fixture } from '../../../../karma';
+
+const stickerTestType = Object.keys(STICKERS)[0];
+
+describe('More button', () => {
+  let fixture;
+
+  beforeEach(async () => {
+    fixture = new Fixture();
+    fixture.setFlags({ floatingMenu: true });
+
+    await fixture.render();
+    await fixture.collapseHelpCenter();
+  });
+
+  afterEach(() => {
+    fixture.restore();
+  });
+
+  it('media element: should focus the size and position panel when the more button is clicked', async () => {
+    // add image to canvas
+    await fixture.events.mouse.clickOn(
+      fixture.editor.library.media.item(0),
+      20,
+      20
+    );
+
+    // click more button
+    await fixture.events.click(fixture.editor.canvas.designMenu.more);
+
+    // verify that first input in size and position panel is focused
+    expect(document.activeElement).toBe(fixture.editor.inspector.designTab);
+  });
+
+  it('text element: should focus the size and position panel when the more button is clicked', async () => {
+    // add text to canvas
+    await fixture.editor.library.textTab.click();
+    await fixture.events.click(fixture.editor.library.text.preset('Paragraph'));
+
+    // click more button
+    await fixture.events.click(fixture.editor.canvas.designMenu.more);
+
+    // verify that first input in size and position panel is focused
+    expect(document.activeElement).toBe(fixture.editor.inspector.designTab);
+  });
+
+  it('shape element: should focus the size and position panel when the more button is clicked', async () => {
+    // add text to canvas
+    await fixture.editor.library.shapesTab.click();
+    await fixture.events.click(fixture.editor.library.shapes.shape('Triangle'));
+
+    // click more button
+    await fixture.events.click(fixture.editor.canvas.designMenu.more);
+
+    // verify that first input in size and position panel is focused
+    expect(document.activeElement).toBe(fixture.editor.inspector.designTab);
+  });
+
+  it('sticker element: should focus the size and position panel when the more button is clicked', async () => {
+    // add text to canvas
+    await fixture.editor.library.shapesTab.click();
+    await fixture.events.click(
+      fixture.editor.library.shapes.sticker(stickerTestType)
+    );
+
+    // click more button
+    await fixture.events.click(fixture.editor.canvas.designMenu.more);
+
+    // verify that first input in size and position panel is focused
+    expect(document.activeElement).toBe(fixture.editor.inspector.designTab);
+  });
+});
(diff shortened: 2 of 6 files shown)


packages/design-system/src/components/input/numericInput.js
```diff
@@ -17,82 +17,90 @@
...
-    onChange,
-    max,
-    min,
-    value,
-    isIndeterminate: originalIsIndeterminate,
-  });
-
-  useKeyDownEffect(
-    inputRef,
+export const NumericInput = forwardRef(
```
samwhale: A big diff - but added `forwardRef` and combined both refs using the `useCombinedRefs` hook.

packages/story-editor/src/components/floatingMenu/elements/karma/more.karma.js
```diff
@@ -0,0 +1,211 @@
...
+    const elementFrame = fixture.editor.canvas.framesLayer.frame(id).node;
+
+    // open floating menu
+    await fixture.events.click(elementFrame);
+
+    // click more button
+    await fixture.events.click(fixture.editor.canvas.designMenu.more);
+  }
+
+  it('image element: should focus the size and position panel when the more button is clicked', async () => {
```
samwhale: This is the same test for each element. Not sure if we get much value for testing each one since the button always does the same thing

If y'all want an individual test for each element type that's fine too.

Note: no gif exists in the `__static__` folder yet. Tried adding one, and it turned into more work than it was worth, so we can always come back to that later and enable that test when the asset it added.

packages/story-editor/src/components/panels/design/sizePosition/sizePosition.js
```diff
@@ -256,6 +293,8 @@ function SizePositionPanel(props) {
         )}
         <Area area="x">
           <NumericInput
+            ref={highlightRef}
+            onKeyDown={handleCancelHighlight}
```
samwhale: There's a different pattern around the codebase that adds the event listener in the ref callback

like this: https://github.com/GoogleForCreators/web-stories-wp/blob/fb965df3af6bec607dc5e6dbb97856adcac1b2c8/packages/story-editor/src/components/panels/design/imageAccessibility/imageAccessibility.js#L75-L80

But react provides an api for doing this using its own event management. This does the same thing without needing to manually attach and remove event listeners from this component.

packages/story-editor/src/karma/fixture/containers/designPanel/sizePosition.js
```diff
@@ -27,6 +27,14 @@ export class SizePosition extends AbstractPanel {
     super(node, path);
   }
 
+  get xPosition() {
+    return this.getByRole('textbox', { name: /^X position$/i });
+  }
+
+  get yPosition() {
+    return this.getByRole('textbox', { name: /^X position$/i });
```
BrookeGraham: shouldn't this be `name:  /^Y position$/i` 🤔

packages/story-editor/src/app/highlights/states.js
```diff
@@ -102,6 +103,10 @@ export const STATES = {
     focus: true,
     tab: DESIGN,
   },
+  [keys.SIZE_AND_POSITION]: {
```
BrittanyIRL: heads up this is changing in Marcin's sidebar update which isn't merged into main yet. https://github.com/GoogleForCreators/web-stories-wp/pull/10760/files#diff-ed4b4a7647ac8a4bb4a9d16aaf241422d72b46cf0fc6af9bd2211cbb7bdc1c55L103

packages/story-editor/src/components/floatingMenu/elements/karma/more.karma.js
```diff
@@ -0,0 +1,211 @@
...
+      fixture.editor.inspector.designPanel.sizePosition.xPosition
+    );
+  });
+
+  it('text element: should focus the size and position panel when the more button is clicked', async () => {
+    const text = await fixture.act(() => insertElement('text', textData));
+
+    await clickMoreButtonForElement(text);
+
+    // Karma test wasn't working unless the button is clicked twice.
```
BrittanyIRL: weird!

packages/story-editor/src/app/highlights/states.js
samwhale: I'll update that when it gets merged in 👍

packages/story-editor/src/components/floatingMenu/elements/karma/more.karma.js
BrookeGraham: Hmm 🤔

BrookeGraham (approved):
One small question, otherwise looks good to me.

BrittanyIRL (approved):
So I asked Amy to confirm where the highlight goes based on a demo of your branch, just to double check like we'd briefly mentioned to do during our design sync and she shared this: 
Screen Shot 2022-03-09 at 3 25 01 PM

The update to the sidebar Marcin is working through gives each element it's own "section" - and this is what was apparently meant in that ticket about highlighting said section.  I'm so sorry this wasn't communicated better, that's on me. 

That said, I kind of feel inclined to merge this and make a follow up ticket that updates it according to the side panel move just to avoid a stale branch. What do you think?

packages/story-editor/src/karma/fixture/containers/designPanel/sizePosition.js
samwhale: lol yes, nice catch

packages/story-editor/src/components/floatingMenu/elements/karma/more.karma.js
samwhale: Super weird and I'm not sure why 😬

packages/story-editor/src/components/floatingMenu/elements/karma/more.karma.js
miina: Could we click on the actual items in the library to make this more similar to the real usage instead of using `insertElement` with predefined data? 

Edit: Then the element gets selected automatically as well, so there's no need to click on the frame first. It's something we've generally been trying to do to keep the testing as close to real usage as possible.

packages/story-editor/src/components/floatingMenu/elements/karma/more.karma.js
miina: @samwhale
How does it fail? Does it look like the click is not triggered at all, or does setting the active element take time? For example, when adding a delay before verifying the size and position panel is focused, does it work then? Or does it never get highlighted with just one click.

In case of the latter:
Does it work if we first move the cursor above the button and then do the click, e.g. something like
```
await fixture.events.mouse.seq(({ moveRel, moveBy, down, up }) => [
      moveRel(moreButton, '50%', '50%', { steps: 10 }),
      down(),
      up(),
    ]);
```

or first moving the cursor to the button, then waiting for a bit and then clicking? Just trying to understand if something is still loading and that's the issue.

If we merge the PR with a double click, would be good to create a follow-up issue for when the feature is fully ready to look into it then.

miina: One thing I can see when running Karma locally is that the Text element is already selected after using the hook to insert it. Clicking on it again enters the edit mode -- perhaps it might take a bit of time to change the state and load the text editor and that's why clicking the button via code will not react right away.

When switching the insertion hook out to clicking the element in the library to add it, it will be selected by default -> no need to click it again -> the test should work with just one click, too.
For example, this works locally:

```
    await fixture.editor.library.textTab.click();
    await fixture.events.click(fixture.editor.library.text.preset('Paragraph'));
    await fixture.events.click(fixture.editor.canvas.designMenu.more);

    // verify that first input in size and position panel is focused
    expect(document.activeElement).toBe(
      fixture.editor.inspector.designPanel.sizePosition.xPosition
    );
```

samwhale: Updating it to what you have here fixed it!

samwhale:
This is branched off of the [`sidebar` branch](https://github.com/GoogleForCreators/web-stories-wp/tree/sidebar). Until that is merged in, this is blocked.

The reason it's branched off of that is because some internals of the highlight api have slightly changed. Rather than make @merapi deal with these merge conflicts by merging straight to main, this is fine to wait until after he is done.

> The update to the sidebar Marcin is working through gives each element it's own "section" - and this is what was apparently meant in that ticket about highlighting said section. I'm so sorry this wasn't communicated better, that's on me.

Chatted with @BrittanyIRL about this and we think this is a good path forward.

maxyinger (approved):

samwhale:
Now that #10931 is merged, this is no longer blocked. I had attached the wrong issues to this, so those were removed.

BrittanyIRL (approved):

Issue #10825, opened 2022-03-07
Floating menu: Color picker positioning


miina:
## Feature Description
Position the color picker so that it would always fit to the screen. Currently it's only positioned above the menu and can be cut off if there's not enough room:
<img width="525" alt="Screenshot 2022-03-07 at 15 19 10" src="https://user-images.githubusercontent.com/3294597/157074064-27fc73d4-0c69-4ea6-ae9a-f8792b27ef2f.png">

miina:
@mariana-k It looks like @BrookeGraham is already addressing this issue as part of #10803, this would mean no additional work is required here.
Once @BrookeGraham confirms we can close the issue as getting fixed as part of #10803

BrookeGraham:
I was just about comment here. Thanks @miina. Yes, I have to update the spacing for color picker anyways, so I this will be fixed @mariana-k. I'll be sure to tag you when I get that pr pushed up. It's wip atm. Draft pr is #10883

BrookeGraham:
Fixed via #10883

Issue #10803, opened 2022-03-04
Sidebar: Fix positioning of popups (color/date picker, tooltips)


merapi:
## Feature Description

Currently, we only have color/date pickers in the right column of the editor and their positioning is not flexible (they always open on the left side of the input), so with the new Sidebar on the left side, the color picker opens out of the viewport area, I've introduced a temporary fix so the picker is at least visible but we need to change the positioning a bit. It was [discussed](https://xwp.slack.com/archives/C01ESBB8ZN1/p1645461514518299) on Slack and it should look like this:

![image](https://user-images.githubusercontent.com/21036224/157626076-231c4c06-398e-42e4-a9ad-bef427101aee.png)

Date picker case is easier, just the anchor point needs to change, right now it's top/right, should be top/left:
![image](https://user-images.githubusercontent.com/21036224/157627543-b0c5b148-9d17-41b7-a5b1-42bd6d93ad11.png)

https://github.com/GoogleForCreators/web-stories-wp/pull/10760#discussion_r818761362

PR should be based on/merged to the `sidebar` branch.

BrittanyIRL:
@merapi - @miina flagged this for us to pick up to help you out. 

I need you to add a real description to this ticket.

BrookeGraham:
I would love to get started on this tomorrow, but I am not sure I follow. What all is needed here?

miina:
@BrookeGraham As far as I understand, the issue is the following:

Currently, in the `sidebar` branch, the right sidepanel was essentially moved to the left side, instead of updating all the tooltips/color picker/ date picker positions, it was just "fixed" by setting `isRTL=!isRTL`, for example: https://github.com/GoogleForCreators/web-stories-wp/pull/10760/files#diff-113a0edff03852bd3a3a51edc102345048b579c514aeed5fb9d899794f08c499R254

The goal of this issue is to remove those temporary fixes and position the popups as they actually should be, considering that the sidebar is now on the left instead.

@merapi - can you confirm?

merapi:
> @merapi - can you confirm?

Yes, I've added a proper description that shows the correct positioning.

BrookeGraham:
Awesome! Thanks so much y'all.

BrookeGraham:
Would someone mind pinging me in the slack thread? (@brk) Since I was added using Slack connect, I do not have access to links that are based on XWP's slack domain.  Not a big if not. I am pretty sure I have enough context now.

[...]

PR #10859, opened 2022-03-08
Editor: Add text alignment control in design menu

### LINKED ISSUE #10630: UI Simplification: Floating menu: Text align
## Feature Description

Make the text align element in the floating design menu functional.

Design: [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=11394%3A457149) + [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=11394%3A457514)
Menus: Text
Notes: This includes a dropdown / submenu. This could perhaps use the submenu functionality that already exists in the context menu, but it would have to be generalized to also work for horizontal menus too.

## Additional information

text-aligment
Screen Shot 2022-02-18 at 18 48 16

barklund:
## Context

This makes the text alignment control functional including keyboard interaction

## To-do

- [ ] Tests

## Non-user-facing changes

| Alignment | Menu closed | Menu open | Menu open with keyboard focus |
|-|-|-|-|
| `left` | Screen Shot 2022-03-08 at 18 20 51 | Screen Shot 2022-03-08 at 18 21 37 | Screen Shot 2022-03-08 at 18 21 03 |
| `center` | Screen Shot 2022-03-08 at 18 21 11 | Screen Shot 2022-03-08 at 18 21 15 | Screen Shot 2022-03-08 at 18 21 23 |
| `right` | Screen Shot 2022-03-08 at 18 21 44 | Screen Shot 2022-03-08 at 18 21 48 | Screen Shot 2022-03-08 at 18 21 56 |
| `justify` | Screen Shot 2022-03-08 at 18 22 02 | Screen Shot 2022-03-08 at 18 22 08 | Screen Shot 2022-03-08 at 18 22 14 |

## Testing Instructions

Adds tracking in the following format when invoked:
```
    trackEvent('floating_menu', {
        name: `set_text_alignment_${value}`,
        element: 'text',
      });
```

## Checklist

---

Fixes #10630


packages/design-system/src/components/contextMenu/components/subMenuTrigger.js
```diff
@@ -50,6 +50,8 @@ function SubMenuTrigger({
   isRTL = false,
```
miina: Looks like this component is not used for the submenu, are the changes here still needed?

packages/story-editor/src/components/floatingMenu/elements/textAlign.js
```diff
@@ -17,21 +17,158 @@
...
+const SubMenuContainer = styled.div`
+  position: absolute;
+  top: calc(var(--height) + ${OFFSET_Y}px);
+  z-index: 9999;
+`;
+
+const StyledMenuButton = styled(ContextMenuComponents.MenuButton)`
+  flex: 0 0 60px;
+  height: 36px;
+  border: 1px solid #393d3f;
```
miina: Nit: Is there a similar color from the design system? 🤔

miina (approved):
Works great! 🎠

miina:
From testing but unrelated to this PR:
The font size input seems to interfere with the keyboard navigation in the menu. E.g. it seems to be impossible to move out of the font size input using left-right arrow keys (was trying to get into the text alignment button from there to test it out). Should we "manually" add event listeners for left-right arrow keys to that input? Thoughts?

cc @BrookeGraham (as it's related to the font size input even though perhaps generally related to numeric inputs, too)

BrookeGraham:
>From testing but unrelated to this PR:
The font size input seems to interfere with the keyboard navigation in the menu. E.g. it seems to be impossible to move out of the font size input using left-right arrow keys (was trying to get into the text alignment button from there to test it out). Should we "manually" add event listeners for left-right arrow keys to that input? Thoughts?

The right and left arrow should traverse you though each character within the (numeric) input so that part is working fine. We should however, be able to go from the font size input to the text alignment dropdown via tab.  The align dropdown button currently has tabIndex={-1} so we should just update that to 0

miina:
> The right and left arrow should traverse you though each character within the (numeric) input so that part is working fine. We should however, be able to go from the font size input to the text alignment dropdown via tab. The align dropdown button currently has tabIndex={-1} so we should just update that to 0

Hm, the thing is that currently, the floating menu is using arrow keys for navigating between the buttons and tab will tab to another section, as in case of other context menus. Then we'd need to rework that logic, it would be a bit different than other context menus. Thoughts, @barklund?

barklund:
@BrookeGraham, @miina:

> > The right and left arrow should traverse you though each character within the (numeric) input so that part is working fine. We should however, be able to go from the font size input to the text alignment dropdown via tab. The align dropdown button currently has tabIndex={-1} so we should just update that to 0
> 
> Hm, the thing is that currently, the floating menu is using arrow keys for navigating between the buttons and tab will tab to another section, as in case of other context menus. Then we'd need to rework that logic, it would be a bit different than other context menus. Thoughts, @barklund?

Yeah, I have been thinking about this. Google Docs and others do this by allowing you to use arrow keys to move between the menu elements, including moving focus to an input field. But the input field acts like a menu item button, so you have to press enter when focus is on the input field menu item to actually go into typing in the input field. When there, arrow keys work as normal in a numeric input field, moving the cursor and stepping the value. Then you can press escape to move back out of the input field, moving focus back to the menu item and you can again use arrow keys to move between menu items.

Another related problem is, that the focus management handled by the menu does not follow actual focus. If I press a button and then press arrow keys to move to the next button, the arrow keys will take me to a different button because the arrow key focus and actual focus are not in sync.

We should create those issues as 1 or 2 new tickets.

packages/design-system/src/components/contextMenu/components/subMenuTrigger.js
barklund: Good point, I ended up scrubbing using that one. Will revert the changes.

packages/story-editor/src/components/floatingMenu/elements/textAlign.js
barklund: Oh, I forgot about going back to that one. 🤦

barklund: It's just `border.defaultNormal` 🙄. Done in 9bc94c32b2

packages/design-system/src/components/contextMenu/components/subMenuTrigger.js
barklund: Reverted in 43e08670cb.

miina:
Issue added for the keyboard navigation: https://github.com/GoogleForCreators/web-stories-wp/issues/10872

BrookeGraham:
Ah. I see where some confusion lies. I think since we have named this component `FloatingMenu`,  we are expecting them to behave as menus with menu items, but they aren't really menus at all. The input for font size, and probably other items within this floating bar make them forms. We should be thinking of them as such. Which is a total bummer because that would mean that we would have to refactor the `floating menu` to be a `floating form`. But I definitely don't wanna block this on the keyboard nav issue. I would suggest that for now, we override `tabIndex` to get the functionality we need and use that followup ticket to nail this down and refactor `floating menu` to fix the issue of them not being menus. Either not use context menu or something. 

@miina @barklund @BrittanyIRL thoughts?

barklund:
> Ah. I see where some confusion lies. I think since we have named this component `FloatingMenu`, we are expecting them to behave as menus with menu items, but they aren't really menus at all. The input for font size, and probably other items within this floating bar make them forms. We should be thinking of them as such. Which is a total bummer because that would mean that we would have to refactor the `floating menu` to be a `floating form`. But I definitely don't wanna block this on the keyboard nav issue. I would suggest that for now, we override `tabIndex` to get the functionality we need and use that followup ticket to nail this down and refactor `floating menu` to fix the issue of them not being menus. Either not use context menu or something.

I don't think so. Having inputs in a menu is not uncommon. Menu bars can have search inputs or similar. I don't think it makes it a form, it can still be considered a menu bar. 

Note in this gif how the Google Docs menu bar works. I can use arrow keys to move between the elements in the menu, not tab. Using arrow keys, i just move over the font size input as if it was a button. Only once I press enter on the "input button", do I get to interact with the input.

gdocs-menu-bar-input

Google Docs then decides to always move keyboard focus back to the document whenever you select something in the menu (even when doing so with the keyboard), so as soon as I press enter inside the input field to commit my choice of font size, focus moves to the document.

I'm not sure if we want to do the same. We could alternatively move focus back to the "input button" rather than the input itself, so arrow keys would once again work. We could then have <key>esc</key> work in the same way except that any input would be reset to the original state before moving focus back to the input button.

I think that's a better solution (but probably more work) than redefining how a menu bar works from an a11y standpoint.

(@miina, I copied this to the new ticket as well)

BrookeGraham:
>Having inputs in a menu is not uncommon. Menu bars can have search inputs or similar. I don't think it makes it a form, it can still be considered a menu bar.

The google docs example you used here is actually a `role=toolbar` not a `menu` like we currently have the floating bars set up. 

>I think that's a better solution (but probably more work) than redefining how a menu bar works from an a11y standpoint.

If y'all are married to the idea of navigating with arrows, (I personally found that odd, and would expect to use tab) then the aria role of toolbar is going to be a better representation of what this feature actually is. 

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/toolbar_role

packages/design-system/src/icons/align_text_center.svg
```diff
@@ -1,3 +1,3 @@
-<svg width="32" height="32" viewBox="0 0 32 32" fill="none" xmlns="http://www.w3.org/2000/svg">
+<svg width="32" height="32" viewBox="0 0 32 32" fill="none" xmlns="http://www.w3.org/2000/svg" data-type="center">
```
barklund: This is kind of a cheat, but it works!

I saw no better way to test in code, that the actual icon displayed was the correct one. The only alternative was to hardcode the SVG paths, but that definitely seemed like a worse option.

More semantic alternatives include putting a title on the whole SVG (`<svg ...><title>Align text center</title><path .../></svg>`) or on the individual elements (`<svg ...><path ...><title>Align text center</title></path></svg>`), but we strip this kind of metadata from our SVG's in svgo, so that's no good. The `data-` property however survives.

barklund:
> > Having inputs in a menu is not uncommon. Menu bars can have search inputs or similar. I don't think it makes it a form, it can still be considered a menu bar.
> 
> The google docs example you used here is actually a `role=toolbar` not a `menu` like we currently have the floating bars set up.
> 
> If y'all are married to the idea of navigating with arrows, (I personally found that odd, and would expect to use tab) then the aria role of toolbar is going to be a better representation of what this feature actually is.

Yeah, that is an excellent point. "Toolbar" is the correct term for this kind of thing. We should change the role for the floating design menu to be toolbar in #10872 and adopt all the relevant interactions from [that spec](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/toolbar_role).

packages/story-editor/src/components/floatingMenu/elements/textAlign.js
```diff
@@ -17,21 +17,160 @@
...
+      setTimeout(() => node.focus());
+    }
+  };
+
+  const CurrentIcon =
+    ALIGNMENTS.find(({ value }) => value === textAlign)?.icon ||
+    Icons.AlignTextLeft;
+
+  const handleTextAlign = (value) => () =>
+    updateSelectedElements({
```
miina, line 125: Could we add tracking here already now since it hasn't been merged yet?
Added [an issue for it as well](https://github.com/GoogleForCreators/web-stories-wp/issues/10871) for adding it everywhere but if we already have an open PR, then ... WDYT?

packages/story-editor/src/components/floatingMenu/elements/karma/textAlign.karma.js
```diff
@@ -0,0 +1,112 @@
...
+describe('Design Menu: Text alignment', () => {
+  let fixture;
+
+  beforeEach(async () => {
+    fixture = new Fixture();
+    fixture.setFlags({ floatingMenu: true });
+    try {
+      await fixture.render();
+    } catch {
+      // ignore
```
miina: Curious, why do we need a catch here?

packages/design-system/src/icons/align_text_center.svg
miina: Could it alternatively be a prop passed to the icon directly, e.g. `<Icon data-type={textAlign} />` or something similar?

miina (approved):
Works well!

From testing:
When there is not enough room for the text-align flyout, it will be displayed partially on top of the floating menu, and scrolling does't update the positioning once it has room again.

This situation can be created when zooming for example which has a few more oddities (not related directly to this PR) -- e.g. the floating menu is always visible when scrolling. Not sure what's the expected behaviour with zooming but might be good to create a follow-up issue for future to double check with UX.

Floating menu is visible even though the element itself is scrolled somewhere away:
Screenshot 2022-03-10 at 10 59 34

alignment flyout on top of the floating menu since it didn't have enough room at the time of opening:
Screenshot 2022-03-10 at 10 58 44

merapi:
About the keyboard interaction: I cannot open it via keyboard.

https://user-images.githubusercontent.com/21036224/157650261-550dfd84-bdf3-45a2-8ba3-3b1445e847fd.mov

and even If I open it via mouse the keyboard navigation seems broken, can you check it?

miina:
> About the keyboard interaction: I cannot open it via keyboard.

@merapi That's related to the comments starting from [here](https://github.com/GoogleForCreators/web-stories-wp/pull/10859#issuecomment-1062652565). Currently, the font size input works differently than other inputs. There's an issue to fix it: #10872 

You should be able to get into the alignment menu by first clicking on the More button for example and then using `Shift + Tab` or arrow keys (not sure which works, it's a bit of a mess currently).

merapi:
> > About the keyboard interaction: I cannot open it via keyboard.
> 
> @merapi That's related to the comments starting from [here](https://github.com/GoogleForCreators/web-stories-wp/pull/10859#issuecomment-1062652565). Currently, the font size input works differently than other inputs. There's an issue to fix it: #10872
> 
> You should be able to get into the alignment menu by first clicking on the More button for example and then using `Shift + Tab` or arrow keys (not sure which works, it's a bit of a mess currently).

I think my issue is unrelated - I'm not using arrow keys, just tab/shift+tab just like you described with the More button.

miina:
> I think my issue is unrelated - I'm not using arrow keys, just tab/shift+tab just like you described with the More button.

Hah, okay, looks like since the Font color input was added, it's not possible to get into the Text alignment button anymore using just the keyboard.

The tabIndex of that button is -1 by default and since some of the items support only tabbing and not arrow keys then it's just not possible anymore to navigate into it using the arrow keys. #10872 is still the issue to fix it generally.

For now, as @BrookeGraham suggested, it might make sense to just set the button's tabIndex to `0` though until the navigation gets sorted out to at least being able to test it.

It can be tested separately when clicking on the text alignment button and then pressing Up key to close the flyout . But getting into the button seems to be impossible now indeed. 🤷

packages/design-system/src/icons/align_text_center.svg
barklund: Hm, I don't feel like that actually addresses the issue of whether the right icon is displayed. But I guess that's not a job for the test. It's just to make sure that the TextAlign component has correctly picked up on which icon to display. So I guess this is a good enough solution. Though `aria-label` probably works better because it's actually semantic, except not really ever exposed by a screen reader because the button has its own label.

packages/story-editor/src/components/floatingMenu/elements/textAlign.js
barklund: Added in 5362692ce.

packages/story-editor/src/components/floatingMenu/elements/karma/textAlign.karma.js
barklund: Ah, it's because otherwise it doesn't work on my local setup (because of some font loading error thrown). I have to remember to remove these before pushing. Fixed in a145fc96d.

packages/design-system/src/icons/align_text_center.svg
barklund: Changed in a145fc96d.

barklund:
> From testing: When there is not enough room for the text-align flyout, it will be displayed partially on top of the floating menu, and scrolling does't update the positioning once it has room again.

I'm not going to do anything more about positioning in this PR. Yes it can be off, but it's only in really rare edge cases and we already have a ticket for it: #10825 (though it says it's specifically about color picker, it really should be about all the flyouts).

> For now, as @BrookeGraham suggested, it might make sense to just set the button's tabIndex to 0 though until
> the navigation gets sorted out to at least being able to test it.

I'm also going to ignore this for now. If it's only for testing purposes, you can manipulate focus some other way to see it in action. It does work (if the rest of the components worked correctly) - you just now have to trust me.

BrookeGraham:
>I'm not going to do anything more about positioning in this PR. Yes it can be off, but it's only in really rare edge cases and we already have a ticket for it: https://github.com/GoogleForCreators/web-stories-wp/issues/10825 (though it says it's specifically about color picker, it really should be about all the flyouts).

I think #10803 may solve #10825 ? I will be workin on it today. Seems like maybe a duplicate ticket / issue?

packages/story-editor/src/components/floatingMenu/elements/textAlign.js
```diff
@@ -17,21 +17,165 @@
...
+      setTimeout(() => node.focus());
+    }
+  };
+
+  const { icon: CurrentIcon, label: currentLabel } =
+    ALIGNMENTS.find(({ value }) => value === textAlign) || ALIGNMENTS[0];
+
+  const handleTextAlign = (value) => () => {
+    trackEvent('floating_menu', {
+      name: `set_text_alignment_${value}`,
```
barklund, line 122: @swissspidy,@miina: would this be the proper way to do it or should the actual alignment value be (another) property on the object?

miina:
> I'm not going to do anything more about positioning in this PR. Yes it can be off, but it's only in really rare edge cases and we already have a ticket for it: https://github.com/GoogleForCreators/web-stories-wp/issues/10825 (though it says it's specifically about color picker, it really should be about all the flyouts).

@barklund Yep, definitely not for this PR, that's why I mentioned it might be good to have an issue for resolving Zoom issues! Looks like this can only really happen in Zoom mode currently. 

> I think https://github.com/GoogleForCreators/web-stories-wp/issues/10803 may solve https://github.com/GoogleForCreators/web-stories-wp/issues/10825 ? I will be working on it today. Seems like maybe a duplicate ticket / issue?

@BrookeGraham Hm, not sure really, the Color Picker in the floating menu is currently always displayed above it, even if there's not enough room. The issue with the sidebar popups is a bit different -- these are currently positioned using `isRTL=!isRTL` so that it would compensate for the sidebar move. Would fixing those also cover the Color Picker of the floating menu/toolbar? I know that @mariana-k assigned the Color Picker of floating menu issue for herself currently, might be good to coordinate with her if you plan to fix that one as well. Thoughts?

packages/story-editor/src/components/floatingMenu/elements/textAlign.js
miina: Sgtm as is if @swissspidy is ok with it -- in some other places we're also setting the name dynamically like this.

swissspidy: I think it's fine.

We can do a more holistic analysis for floating menu bar telemetry later, to ensure coherent naming.

PR #10860, opened 2022-03-09
Floating Menu: Font Size, fix text element selection size

### LINKED ISSUE #10855: Floating Menu: Font Size not updating text selection box size
## Bug Description

 It looks like the text element size is not adjusting currently when changing the font size. When changing from the design panel (both when increasing / decreasing font size), it does change so it seems to be a bug with the floating menu logic only.

## Expected Behaviour

When the font size changes, the box that surrounds the text element should expand / contrast along with it. 
## Additional Context

In the design panel, `usePresubmitHandler` is the part that before submitting the values, makes sure the text element is adjusted properly as well, this only works in the design panel though together with pushUpdate.
Probably we'd need to use something like `getUpdatedSizeAndPosition` to update the size/position of the text element together with updating the font size (position is relevant if the element is rotated).

Might be worth refactoring to use the same component in both places, similar to the Font Family refactor ticket [#10628](https://github.com/GoogleForCreators/web-stories-wp/issues/10628)

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

BrookeGraham:
## Context
When the font size changes, the box that surrounds the text element should expand / contrast along with it.

## Summary
This component can be used to replace the font size `StyledNumericInput` inside `FontControls`, but we would need to find a width for that input that works for both spots. The current width of `StyledNumericInput` is too narrow for 3 digit numbers. But this FontSize component is still a little wide. I am happy to chat with @agingoldseco to find a size that will allow us to pull this out and use it in both places. 

## Relevant Technical Choices
Grabbed `getUpdatedSizeAndPosition` thanks Miina. 

## User-facing changes
|before|after|
|--|--|
|font size|after|

## Testing Instructions

1. select some text and change the font size. 
2. Check that the containing box size adjusts to fit the text. 

no

no

no

## Checklist

---

Fixes #10855

### DIFF
--- a/packages/story-editor/src/components/floatingMenu/elements/fontSize.js
+++ b/packages/story-editor/src/components/floatingMenu/elements/fontSize.js
@@ -18,11 +18,14 @@
  * External dependencies
  */
 import { __ } from '@googleforcreators/i18n';
+import { useCallback } from '@googleforcreators/react';
 
 /**
  * Internal dependencies
  */
 import { useStory } from '../../../app';
+import getUpdatedSizeAndPosition from '../../../utils/getUpdatedSizeAndPosition';
+import updateProperties from '../../inspector/design/updateProperties';
 import { focusStyle, inputContainerStyleOverride } from '../../panels/shared';
 import { MIN_MAX } from '../../panels/design/textStyle/font';
 // TODO: https://github.com/GoogleForCreators/web-stories-wp/issues/10799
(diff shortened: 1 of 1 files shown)


miina (approved):

packages/story-editor/src/components/floatingMenu/elements/fontSize.js
```diff
@@ -36,18 +39,31 @@ function FontSize() {
...
-    });
+  const pushUpdate = useCallback(
+    (update) => {
+      updateSelectedElements({
+        properties: (element) => {
+          const updates = updateProperties(element, update, true);
+          const sizeUpdates = getUpdatedSizeAndPosition({
+            ...element,
+            ...updates,
+          });
```
samwhale, line 50: oh nice, I didn't know this was a thing that existed 👍

samwhale (approved):

packages/story-editor/src/components/floatingMenu/elements/fontSize.js
BrookeGraham: Me either, Miina pointed me to it.

BrittanyIRL (approved):

PR #10870, opened 2022-03-09
Editor: Fixed empty story state when there is an attachment

### LINKED ISSUE #10700: A page with Page Attachment should not be considered empty state
## Bug Description

A Page with Page Attachment (CTA) is currently considered empty state.

## Expected Behaviour
If a Story has 1 page and that Page has a Page Attachment with no other changes, it is currently considered empty state and displays the empty state menu.
If Page Attachment is present, it shouldn't be considered empty state.

## Steps to Reproduce
1. Add a new Story
2. Select Background from the Layer Panel
3. Add Page Attachment
4. See the empty state UX still present.

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

marianakyrkosh:
## Context

An empty Story text and icon should disappear when an attachment is added to a Story.

## Summary

Currently, even if the Page Attachment is added, the Empty State helper is still shown — but it shouldn’t, since the Page is already changed really.

You can add the Page Attachment locally by:
1. Create a new story.
2. Click on the “Background” list item in the right bottom corner of the screen (The one under “Layers”).
3. Now the Page gets selected and you should be able to see the “Page Attachment” section in the right-side panel.

## User-facing changes

Before: 
Screenshot 2022-03-09 at 15 32 19
After:

image

## Testing Instructions

1. Create 1 Story.
2. Add a link in the Attachments.
3. Empty story text and icon should disappear.

## Checklist

---

Fixes #10700


packages/story-editor/src/app/story/utils/isEmptyStory.js
```diff
@@ -40,6 +40,11 @@ const isEmptyStory = (pages) => {
     return false;
   }
 
+  const hasNoAttachment = pages[0].pageAttachment?.length === 0;
```
miina: Nit: we usually prefer "positive" variable names for readability, e.g.
`const hasPageAttachment` and then just check for `if (hasPageAttachment)`.

The page attachment is actually an object that consists of different props, e.g. if we remove a Page Attachment from a Page, it will still stay as a prop but with empty values:
```
"pageAttachment": {
        "url": "",
        "icon": "",
        "needsProxy": false
      }
```

It can also be `null` if not set.

We should probably check here for the length of the `url` instead.

merapi: ➕  for the "positive" variable names (in most cases), I'm first to blame here (see `hasNoExtraElements` and `hasNoBackgroundMedia` although this one is tricky)

image

packages/story-editor/src/app/story/utils/test/isEmptyStory.js
```diff
@@ -54,4 +54,12 @@ describe('isEmptyStory', () => {
     storyWithExtraElements[0].elements.push({});
     expect(isEmptyStory(storyWithExtraElements)).toBeFalse();
   });
+
+  it('should not count story with an attachment as empty', () => {
+    const storyWithAttachment = [createPage()];
+    const testUrl = 'https://testurl.com';
+
+    storyWithAttachment[0].pageAttachemnt.push({ url: testUrl });
+    expect(isEmptyStory(storyWithAttachment)).toBeFalse();
```
miina, line 64: Would be good to test that it returns `true` correctly as well, for example when `pageAttachment` is `null` (or not set) and also when it was set but then deleted, then the `pageAttachment` value would be:
```
    {
        url: '',
        icon: '',
        needsProxy: false,
      }
```

barklund (approved):
Pending the extra tests @miina asked for, this looks good to me! :shipit:

packages/story-editor/src/app/story/utils/test/isEmptyStory.js
marianakyrkosh: @miina thanks for the suggestion! I've added the tests

miina (approved):

Issue #10872, opened 2022-03-09
Floating menu: Keyboard Navigation to Enter and Exit


miina:
## Feature Description
Generally, the floating menu is using roving index keyboard navigation as other context menus (navigating with arrow keys, tab moves to the next section), however, since we have a text input in the menu, this doesn't work exactly as expected.

Some thoughts from @barklund:

> Google Docs and others do this by allowing you to use arrow keys to move between the menu elements, including moving focus to an input field. But the input field acts like a menu item button, so you have to press enter when focus is on the input field menu item to actually go into typing in the input field. When there, arrow keys work as normal in a numeric input field, moving the cursor and stepping the value. Then you can press escape to move back out of the input field, moving focus back to the menu item and you can again use arrow keys to move between menu items.
>
> Note in this gif how the Google Docs menu bar works. I can use arrow keys to move between the elements in the menu, not tab. Using arrow keys, i just move over the font size input as if it was a button. Only once I press enter on the "input button", do I get to interact with the input.
>
> ![gdocs-menu-bar-input](https://user-images.githubusercontent.com/637548/157508949-0148ba2c-94bb-4ff8-a4b5-0b0e8d7b0710.gif)
>
> Google Docs then decides to always move keyboard focus back to the document whenever you select something in the menu (even when doing so with the keyboard), so as soon as I press enter inside the input field to commit my choice of font size, focus moves to the document.
>
> I'm not sure if we want to do the same. We could alternatively move focus back to the "input button" rather than the input itself, so arrow keys would once again work. We could then have <key>esc</key> work in the same way except that any input would be reset to the original state before moving focus back to the input button.
>
> Another related problem is, that the focus management handled by the menu does not follow actual focus. If I press a button and then press arrow keys to move to the next button, the arrow keys will take me to a different button because the arrow key focus and actual focus are not in sync.

See also the [comment thread and thoughts](https://github.com/GoogleForCreators/web-stories-wp/pull/10859#issuecomment-1063266518) from @BrookeGraham 

The current idea (as of 10th of March) is to change the role of the menu to "toolbar" and implement the relevant parts from [its spec](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/toolbar_role).

maxyinger:
@BrittanyIRL  Just to clarify, the expected scope of this ticket is:
- can `enter` while element is both selected & focused to access the toolbar (for all element types)
[...]

PR #10861, opened 2022-03-09
Floating Menu: Render Depth Resolved

BrittanyIRL:
## Context

Floating menus should be behind the Footer popups.

## Summary

Put simply, the initial layout of the canvas wasn't anticipating this many popups and the natural order of zindexing got weird. 

## Relevant Technical Choices

- Don't set `z-index` on the navLayer `Layer` so that the contents of that layer have a fighting chance at claiming their own positioning. The sidebar menu and the footer are both children of this layer, so if the z-index is set on the `Layer` (the parent) then we can't get the floating menu (which is at the same level as the `Layer` to be under the footer popovers and over the sidebar menu.  
It looks like this: 
```
Layout Layer
  > PageMenu Area > Side Menu
  > Footer > Help Center/Checklist/KeyboardNav 
Floating Menu
```

- Also bringing the `SideMenu` `z-index` down from 999 to 3 to make it more manageable.
- Move `z-index` for layers to a constant file so that its easier to tell the ordering and we don't have to hunt around for another 4 hours in the future for all of these 👯 

## To-do

N/A

## User-facing changes

`z-index` is one of those things that usually just gets updated when one specific thing is wrong, I tried hard to not mess with the order any more than I had to. There should be no user facing changes.

Behind the floating menu feature flag we get the following: 

| View | New | Old | 
| -- | -- | -- |
| Help Center | Screen Shot 2022-03-08 at 5 33 11 PM | Screen Shot 2022-03-08 at 5 34 34 PM |
| Checklist | Screen Shot 2022-03-08 at 5 13 37 PM | Screen Shot 2022-03-08 at 5 34 28 PM | 
| Keyboard Nav | Screen Shot 2022-03-08 at 5 33 18 PM | Screen Shot 2022-03-08 at 5 34 21 PM |

## Testing Instructions

Now the floating menus seen when an element is selected in the canvas should be _behind_ the checklist/help center/ keyboard nav when any of those are open while still remaining _in front_ of the side bar menu. 

Please also consider the other ways the z index is used here, I check drag and tooltips and zoom and the layers panel but I appreciate any heads up from people who have different context about regressions. 

- [x] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1.

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
[...]

PR #10856, opened 2022-03-08
Editor: Fixed layer order of design menu relative to edit layer

barklund:
## Context

This simply moves the layering of the design menu (was `z-index:2`, now `z-index:4`) to be above the edit layer (`z-index:3`).

Note that there are still other layering issues to be resolved in #10636.

## Non-user-facing changes

Screen Shot 2022-03-08 at 16 00 34

## Testing Instructions

-[x ] This is a non-user-facing change and requires no QA

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/googleforcreators/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [x] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/googleforcreators/web-stories-wp/tree/main/docs#testing))
- [x] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #10845

PR #10875, opened 2022-03-10
Floating menu: Add Text formatting toggles

### LINKED ISSUE #10631: UI Simplification: Floating menu: Text format toggles
## Feature Description

Make the text format toggle elements in the floating design menu functional.

Design: [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=11394%3A457149) + [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=11394%3A457514)
Menus: Text
Notes: This includes both the bold, italic, and underline toggles. This should use the existing functionality available in the rich-text helper utils. This should work both in and out of text edit-mode (and correctly reflect whether current selection is bold/italic/underlined while in edit-mode).

## Additional information

text-toggles

miina:
## Summary
Makes bold, italic, underline functional

## Relevant Technical Choices
Changed the `tabIndex` of the three toggles to `0` -- it was impossible to get into those with keyboard otherwise.
To be followed up in #10872 

## User-facing changes
Screenshot 2022-03-10 at 13 33 44

## Testing Instructions

1. Add a Text element
2. Verify the bold, italic, underline toggles work as expected.
3. Verify the toggles work when selecting only part of the text as well.

Adds tracking in the following format for each toggle:
```
    trackEvent('floating_menu', {
        name: 'set_bold',
        element: 'text',
      });
```

## Checklist

---

Fixes #10631


packages/story-editor/src/components/floatingMenu/elements/toggleBold.js
```diff
@@ -19,18 +19,47 @@
...
+
   return (
     <IconButton
+      isToggled={isBold}
       Icon={Icons.LetterBBold}
-      title={__('Toggle bold text', 'web-stories')}
-      onClick={() => {}}
+      title={__('Toggle bold', 'web-stories')}
+      onClick={() => handleClickBold(!isBold)}
+      tabIndex="0"
```
miina, line 41: FYI: Added the tabIndexes for now since it was impossible to get into the toggles otherwise using keyboard -- some of the menu items seem to only allow tabbing and some using arrow keys. This can be adjusted in #10872.

BrittanyIRL (approved):

packages/story-editor/src/components/floatingMenu/elements/toggleBold.js
```diff
@@ -19,18 +19,47 @@
...
+      updateSelectedElements({
+        properties: (element) => updateProperties(element, update, true),
+      });
+    },
+    [updateSelectedElements]
+  );
+  const {
+    textInfo: { isBold },
+    handlers: { handleClickBold },
+  } = useRichTextFormatting([{ content, type: 'text' }], pushUpdate);
```
barklund: These are very similar in the three components, but I guess there's no good way to generalize it?

Maybe it could be generalized like:

```js
const { isToggled, toggle } = useTextToggle({
  currentValue: 'isBold',
  toggleFunction: 'handleClickBold',
  eventName: 'set_bold'),
});
```

It's not a huge gain, but I feel it makes sense here? Especially because the component here only needs the two values, `isToggled` and `toggle`, to render the button, and the rest is just internal stuff.

packages/story-editor/src/components/floatingMenu/elements/toggleBold.js
```diff
@@ -19,18 +19,47 @@
...
+
   return (
     <IconButton
+      isToggled={isBold}
       Icon={Icons.LetterBBold}
-      title={__('Toggle bold text', 'web-stories')}
-      onClick={() => {}}
+      title={__('Toggle bold', 'web-stories')}
```
barklund, line 39: Hmm, why doesn't the tooltip show up on keyboard focus? I'm pretty sure it did earlier, but it seems to be a problem with all the buttons using this `IconButton` component now (which the eyedropper doesn't use). 

| Without tooltip | With tooltip |
|-|-|
| Screen Shot 2022-03-10 at 16 50 40 | Screen Shot 2022-03-10 at 16 50 32 |

barklund (approved):
Works great!

packages/story-editor/src/components/floatingMenu/elements/toggleBold.js
miina: Hm, not sure -- added an issue: https://github.com/GoogleForCreators/web-stories-wp/issues/10884

packages/story-editor/src/components/floatingMenu/elements/toggleBold.js
miina: Added in [2c20684](https://github.com/GoogleForCreators/web-stories-wp/pull/10875/commits/2c20684a57cae47d9ccea8f031675d1e8368f2e8)

BrittanyIRL (changes requested):
Looks like a karma test related to this is failing. 

Screen Shot 2022-03-11 at 9 12 50 AM

miina:
> Looks like a karma test related to this is failing.

@BrittanyIRL Looks like the test is failing because there was a regression and the floating menu is now below the edit layer again. It got fixed in #10856 but there have been some additional changes meanwhile.

This results in clicking a format toggle exiting Text edit mode instead of clicking the button.

Looks like it changed in this PR: #10861

Could you take a look to see if you can also see this regression, perhaps a follow-up issue is needed for #10861 ?

To reproduce on `main`:
1. Add a Text
2. Enter edit mode
3. Select part of the text
4. Try to open the toolbar color picker to assign color to the selection (click on it once).
5. See the edit mode exiting instead of opening the color picker.

miina:
Moving it back to Review as it can't be QA-ed before the regression gets fixed, it won't be possible to apply any style on inline selection of a text otherwise.

miina:
Karma tests are passing now again after merging `main`.
(e2e test failures are unrelated)

BrittanyIRL (approved):
Glad this was able to flag the issue with positioning menus early.

PR #10883, opened 2022-03-11
Popups in sidebar and floating menu repositioning

### LINKED ISSUE #10803: Sidebar: Fix positioning of popups (color/date picker, tooltips)

### LINKED ISSUE #10825: Floating menu: Color picker positioning

BrookeGraham:
## Context

Now that we are moving the panels to one side, custom use cases of Popup need their placements updated. 

## Summary

Double checked all use cases of tooltip and pop throughout the editor. Updated Color Picker popup , Publish time popup, Color swatch tooltip, Eye dropper tooltip, Alignment tooltip, Style Presets popup.

This pr is branched off `sidebar`

## Relevant Technical Choices
- `ColorInput` was using `dock` ref for placement. Made this dynamic, since we are reusing color in the floating menu. 
- added `dock` to `Tooltip` so we can place the first alignment tooltip at the left most of the screen (in RTL mode, this is set to bottom end since dock would force it in the center of the canvas)
- added yet another flag to `Popup` this one for resetting the x -offset. In the case of RTL, "pure" uses of `Popup`, can render off screen eg. floating menu's color picker. This forces the popup inward.
- created a const file for spacing requirements depending on color input's use case. 

## To-do
n/a

## User-facing changes
||LTR|RTL|
|--|--|--|
|Color swatch tooltip|color swatch|color swatch RTL|
|Color dropper tooltip|color dropper|eyedropper RTL|
|Color picker sidebar|color picker|color picker RTL|
|Color picker floating menu default|floating menu picker default|floating menu RTL|
|Floating Menu edge case|floating menu picker|floating menu RTL edge|
|Alignment tooltip|alignment|alignment RTL|
|Style presets popup|style presets|style presets RTL|
|Publish time popup|publish|publish RTL|

## Testing Instructions

1. Please check all tooltips and popups I may have missed. Known updates to: 
- Color swatch tooltip
- Color dropper tooltip
- Color picker sidebar
- Color picker floating menu
- Alignment tooltip
- Style presets popup
- Publish time popup  (used in publish modal and sidebar)

no

no

no

## Checklist

---

Fixes #10803 
Fixes #10825


packages/story-editor/src/components/floatingMenu/context/provider.js
```diff
@@ -25,8 +26,11 @@ import PropTypes from 'prop-types';
 import Context from './context';
 
 function FloatingMenuProvider({ handleDismiss, children }) {
+  const designMenuRef = useRef(null);
```
samwhale: 💭 there's a ref to the design pane in [the inspector provider](https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/story-editor/src/components/inspector/inspectorProvider.js#L76). It's a different provider but since it's a reference to the design pane you could always grab it from there instead of needing to set up another one.

---

💭 But rather than checking where it's being rendered, we could follow the pattern of the [`Smart Popover` component](https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/design-system/src/components/contextMenu/styled.js#L55-L84) that checks its dimensions and renders itself in the correct place

BrookeGraham: the `designMenu` is referring to the `floating menu` not the design pane. And really, if we just make `dock` `null`, it'll render where it is supposed to. since the anchor is in the floating menu. So I am not sure we even need a ref to the floating menu at all.

packages/story-editor/src/components/panels/design/alignment/alignment.js
```diff
@@ -210,7 +211,13 @@ function ElementAlignmentPanel({ selectedElements, pushUpdate }) {
...
     >
       <ElementRow ref={ref}>
-        <Tooltip title={__('Distribute horizontally', 'web-stories')}>
+        <Tooltip
+          dock={!isRTL ? ref : null}
+          placement={
+            isRTL ? TOOLTIP_PLACEMENT.BOTTOM_START : TOOLTIP_PLACEMENT.BOTTOM
+          }
+          title={__('Distribute horizontally', 'web-stories')}
+        >
```
BrookeGraham: I added `dock` to tooltip so that we can pass it along to `Popup`. This allowed me to dock the tooltip to the `ElementRow` That in turn pushes the tooltip inside the editor's body while still attempting to center the tooltip. However, that would mean in RTL it would render on the left of the sidebar since it was docked to that whole row element.  We don't want that. So in the case of RTL, we kinda have to live with this tooltip being placed `bottom_start` until we can update all tooltip inside the editor.  Basically all the tooltips that get used in the editor will appear under the admin sidebar. They aren't constrained to be within the bounds of the editor.

BrookeGraham: Screen Shot 2022-03-14 at 9 43 39 AM
Screen Shot 2022-03-14 at 9 43 22 AM

packages/design-system/src/components/popup/utils/getOffset.js
```diff
@@ -128,7 +133,12 @@ export function getOffset({
   const { x: spacingH = 0, y: spacingV = 0 } = spacing || {};
 
   // Horizontal
-  const offsetX = getXOffset(placement, spacingH, anchorRect, dockRect, isRTL);
+  const offsetX = () => {
+    if (resetXOffset && popupRect?.left <= 0) {
+      return isRTL ? width : 0;
+    }
+    return getXOffset(placement, spacingH, anchorRect, dockRect, isRTL);
+  };
```
BrookeGraham, line 145: Since tooltips update the placement of a popup, and since some popups can have negative left values, I created this `resetXOffset` flag to limit when we would potentially reset the offset to the width
Screen Shot 2022-03-14 at 1 48 47 PM
 of the popup. This forces the popup to start at the left edge of the screen in in RTL.

packages/story-editor/src/components/floatingMenu/context/provider.js
BrookeGraham: i ended up removing this

packages/wp-story-editor/src/components/documentPane/publish/publishTime.js
```diff
@@ -136,7 +136,7 @@ function PublishTime() {
         topOffset={TOOLBAR_HEIGHT}
         anchor={dateFieldRef}
         isOpen={showDatePicker}
-        placement={PLACEMENT.BOTTOM_END}
+        placement={PLACEMENT.BOTTOM}
```
BrookeGraham: just remove, bottom is default

packages/story-editor/src/components/form/color/constants.js
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+export const SPACING = {
+  FLOATING_MENU: { x: 0, y: 12 },
+  DEFAULT_SIDEBAR: { x: 230, y: 0 },
+  SIDEBAR_WITHOUT_EYEDROPPER: { x: 208, y: 0 },
+  IS_RTL: { x: 12, y: 0 },
+};
```
BrookeGraham, line 23: The `dock` feature in Popups is based on the left of the element we dock to. In the future it would be handy to be able to dock to the right and still have all the RTL, anchor.width, and placement details be able to do the math to render it were we would expect. That is a bit much for this pr. So we can create a follow up ticket to address.

packages/story-editor/src/components/panels/design/alignment/alignment.js
```diff
@@ -210,7 +211,13 @@ function ElementAlignmentPanel({ selectedElements, pushUpdate }) {
       ariaLabel={__('Alignment', 'web-stories')}
     >
       <ElementRow ref={ref}>
-        <Tooltip title={__('Distribute horizontally', 'web-stories')}>
+        <Tooltip
+          dock={!isRTL ? ref : null}
+          placement={
+            isRTL ? TOOLTIP_PLACEMENT.BOTTOM_START : TOOLTIP_PLACEMENT.BOTTOM
+          }
```
BrookeGraham, line 218: In retrospect, I can update the `spacing` when in RTL so that it mirrors LTR

packages/story-editor/src/components/panels/design/textStyle/stylePresets.js
```diff
@@ -82,6 +81,7 @@ function PresetPanel({ pushUpdate }) {
   const [isPopupOpen, setIsPopupOpen] = useState(false);
   const { isRTL, styleConstants: { topOffset } = {} } = useConfig();
   const hasPresets = textStyles.length > 0;
+  const SPACING = { x: isRTL ? 12 : 46 };
```
samwhale, line 84: Does the popup here need `isRTL` and the difference in spacing?

Also - can this live in the constants file with the other spacing constants?

samwhale (approved):
🤔  The popup has become quite complicated. After having both popups combined and moved to the design system - I think it's time to invest in making the pop up easier to use.

This current approach adds to the complicated API that the popup has developed - but considering that it's blocking the sidebar work, I think it makes sense to get this through and come back to address the popup component in the future.

packages/story-editor/src/components/panels/design/textStyle/stylePresets.js
BrookeGraham: Yes, it is docked to the `inspector` element. Since docking is currently based on the left of the element (in conjunction with the where the anchor is), when in RTL that spacing only needs to give the appearance of a little bit of margin, hence the 12 vs 46. I think it is fine to leave here since the other spacing constants are related to the color picker

packages/design-system/src/components/popup/utils/getOffset.js
BrookeGraham: Not happy about adding this extra prop. It was needed because in the layers panel has tooltips that get a negative `left` value, so I can't check them directly. This feels like a bug in the calculations we are using to determine x-offsets. We should investigate why those tooltips are getting wacky values, and then remove this extra `resetXOffset` prop.

packages/story-editor/src/components/form/color/constants.js
BrookeGraham: Okay. @littlemilkstudio helped out here. We discovered that the math being done in `getOffset` is wrong. [Seen here.](https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/design-system/src/components/popup/utils/getOffset.js#L28-L32) When we are using the `dock` feature, we are checking for `dockRect?.left` but in the case where the element has a left value of zero, which inspector now has, it skips over it and uses `anchor?.left` instead. In addition, leftAligned and rightAligned means different things when used with tooltips vs popup (that uses the dock feature). I will create a bug fix ticket to document these findings a bit better.

BrookeGraham: Added ticket: #10928

maxyinger: Awesome. Thanks for documenting & creating the ticket!

BrookeGraham: Please add any details I may have missed, or that you think would be helpful for whoever tackles this. I might be a little too _in it_ to objectively note all the nuances.

maxyinger (approved):
All the behavior here seems to be working for me! Thanks for creating a follow up ticket and tackling this

packages/design-system/src/components/popup/utils/getOffset.js
```diff
@@ -25,6 +25,10 @@ export function getXOffset(
   dockRect,
   isRTL
 ) {
+  // TODO: https://github.com/GoogleForCreators/web-stories-wp/issues/10928
+  // doctRect?.left can have a valid value of zero, however in this case, the falsey
+  // value gets ignored and we end up using anchor.left. Additionally, if using `dock`
```
BrittanyIRL, line 30: this is such a great find.

BrittanyIRL (approved):
😲  Wow this is a lot to sort out. I appreciate the notes that you added and the follow up bug seems like a great jumping off point to take a step back and separate out all of these so that it's easier to maintain. 

Fixes work as you say.

packages/design-system/src/components/popup/utils/getOffset.js
BrookeGraham: Thank @littlemilkstudio  I couldn't figure how the anchor width was messing with the calcs. But those eagle eyes caught it.

Issue #10892, opened 2022-03-11
Layout Layer Reorganized


BrittanyIRL:
The initial Layout Layer structure needs to get an update now that we have so many layers in there that need to overlap visually but remain usable. 

## Context 

Recent redesigns have shifted the editor from its initial structure of 2 panels where everything was controlled for elements on the canvas to 1 sidepanel with menus appearing on top of the canvas to also edit the elements. 

There was this ticket here: https://github.com/GoogleForCreators/web-stories-wp/issues/10636 to "fix" the layer order so that the floating menus would be behind the footer's popups (helpcenter, keyboard nav, checklist) and on top of the sidebar menus that appear on the side of the canvas. This was tricky because the footer popups and the sidebar menus are all children of the `Layout` layer. The Floating menu is at the same level as `Layout`. To fix this I'd removed z index from getting set on the `Layout` to that each set of menus could have it's own priority. This does resolve the visual ordering but it completely breaks the functionality when a floating menu is used and another popup is opened from the floating menu. 

I undid my fix here: https://github.com/GoogleForCreators/web-stories-wp/issues/10636 where you can see the ordering push the menus/popups within the `Layout` layer under the floating menu. 

## Acceptance Criteria 
- All footer popups are visually in front of the floating menu 
- Floating menu is in front of the sidebar menu 
- Elements need to retain selection. 

### Problem demoed with existing z index solution: 

https://user-images.githubusercontent.com/10720454/158215361-fa0e7d96-65e2-4b55-b66d-eeea33d101fc.mp4

So the structure is like so: 
- Canvas layout 
  - Selection Canvas
    - Frames Layer > Layout Page Area (canvas)
    - Frames Layer > Layout NavArea (nav arrows)  
  - Layer Grid (menu time)
    - Head Area 
    - Page Menu Area 
    - Footer Area 
  - Section Menu Wrapper (Floating menus) 

So the floating menus are outside of the canvas layout entirely, when the page area gets a double click to enter "text mode" focus gets lost when z index is updated for the menus to get their desired visual positioning. 

Does floating menu need to be moved inside selection canvas?

PR #10893, opened 2022-03-11
Bug Fix: Move floating menu in front of popups so that functionality is fixed

BrittanyIRL:
## Context

Miina pointed out here: https://github.com/GoogleForCreators/web-stories-wp/pull/10861#discussion_r824902184 and here https://github.com/GoogleForCreators/web-stories-wp/pull/10875#issuecomment-1065285614 that fixing the layering visually actually makes features completely unusable. 
 
## Summary

move popups behind floating menu so that things can be edited.

## To-do

Getting the layering right here is going to be a much bigger lift than fixing the z index, we need to restructure the HTML. Ticket created here: #10892 but for now, the floating menus are just going to need to be on top. 

## Testing Instructions

1.

## Checklist

---

Fixes #

### DIFF
--- a/packages/story-editor/src/components/floatingMenu/elements/karma/text.karma.js
+++ b/packages/story-editor/src/components/floatingMenu/elements/karma/text.karma.js
@@ -96,10 +96,7 @@ describe('Design Menu: Text Styles', () => {
       );
     });
 
-    // This is failing due to keyboard problems
-    // TODO #10872 https://github.com/GoogleForCreators/web-stories-wp/issues/10872
-    // eslint-disable-next-line jasmine/no-disabled-tests
-    xit('should allow changing text color for a selection from the design menu', async () => {
+    it('should allow changing text color for a selection from the design menu', async () => {
       // Enter edit-mode
       await fixture.events.keyboard.press('Enter');
       await fixture.screen.findByTestId('textEditor');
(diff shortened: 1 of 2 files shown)


miina (approved):
Thanks for the temp fix!

Maybe let's enable this test here, too, it should be working now with the floating menu being available in text edit mode: https://github.com/GoogleForCreators/web-stories-wp/pull/10861/files#diff-8b48e6a473c1a371bc199f0cdd4fb91e2b49420347116552ec193a5ee1c6ccedR102

samwhale (approved):
💭 the temp fix works, so in the future do you think that opening those menus should de-select the elements?

BrittanyIRL:
> Thanks for the temp fix!
> 
> Maybe let's enable this test here, too, it should be working now with the floating menu being available in text edit mode: https://github.com/GoogleForCreators/web-stories-wp/pull/10861/files#diff-8b48e6a473c1a371bc199f0cdd4fb91e2b49420347116552ec193a5ee1c6ccedR102

sure, just wanted to get this fix up for you in case it was blocking you before I got in for the day.

BrittanyIRL:
> 💭 the temp fix works, so in the future do you think that opening those menus should de-select the elements?

Whatever the real fix is it needs to keep the element selected so that the popups work. I was honestly caught off guard by the side effects of fixing the z index to be isolated. Where I'm at without digging deeper is that we either need to rework the entire "layer" that holds alllllll of these things so that they have independence or maybe the popups can get a higher z index? That to me feels like just pushing the root problem further along. 

All to say, I am unsure right now but fixing this will be easier now that the entire floating menu is in. 

I'm hoping this can be a really short term fix, but at least now the tests and functionality are all intact.

PR #10900, opened 2022-03-14
[WIP] Try testing library.

### LINKED ISSUE #10473: Discovery: Consider using Puppeteer Testing Library
## Task Description

With Puppeteer Testing Library ([`pptr-testing-library`](https://www.npmjs.com/package/pptr-testing-library)) we could bring our e2e test suite closer to our karma and unit test suites, making it easier to write write tests for all three targets.

Let's do a brief evaluation and perhaps rewrite a few e2e tests to see how well it works. If successful, we can update all of the tests.

spacedmonkey:
## Testing Instructions

1.

## Checklist

---

Fixes #10473


packages/e2e-tests/src/specs/editor/media/insertWebMVideo.js
```diff
@@ -70,23 +74,19 @@ describe('Inserting WebM Video', () => {
 
     const fileName = await uploadMedia('small-video.webm');
     uploadedFiles.push(fileName);
-
-    await page.waitForSelector(
-      `[data-testid="mediaElement-video"] [src*="${fileName}"`
```
swissspidy, line 75: How did this selector ever work? There's a missing square bracket at the end.

Edit: looks like I added this in #10737 😅

packages/e2e-tests/src/specs/editor/media/insertWebMVideo.js
```diff
@@ -70,23 +74,19 @@ describe('Inserting WebM Video', () => {
...
-    );
+    const document = await getDocument(page);
+    await getByTestId(document, 'mediaElement-video');
```
swissspidy: This would need to be in a `waitFor()`, otherwise this is not equivalent to `waitForSelector()`. Also the `[src*="${fileName}]` part is missing.

packages/e2e-tests/src/specs/editor/media/insertWebMVideo.js
spacedmonkey: This will not work at all and should be removed. This is looking for a media element is sidebar with child element with src of filename. But now we use poster images instead of video in the media library, this selector will never return anything.

packages/e2e-tests/src/specs/editor/media/insertWebMVideo.js
spacedmonkey: Still exploring this library. This is a WIP and draft.

packages/e2e-tests/src/specs/editor/media/insertWebMVideo.js
swissspidy: IIRC `fileName` is the file name without extension. So `[src*="${fileName}]` matches also the poster image, so it should work in theory.

swissspidy:
Closing as per the ticket.

PR #10918, opened 2022-03-15
Custom Fonts: Improve Accessibility on Settings screen

### LINKED ISSUE #10878: Custom Fonts: Improve Accessibility on Settings screen
## Bug Description

In #9966 we added a custom fonts section to the settings screen, to allow adding custom fonts by URL. It did not contain any a11y work, but there was also no follow-up ticket for that yet.

This feature right now is completely unusable with keyboard or assistive technologies. It's impossible to tab into it in the first place.

## Expected Behaviour

The `Current Fonts` list needs proper `role="list"` and `role="listitem"` attributes and keyboard support so that users can tab into it, select fonts using ArrowUp/ArrowDown keys and delete a font using the keyboard as well.

## Screenshots

Screenshot 2022-03-10 at 18 53 07
Screenshot 2022-03-10 at 18 53 01

## Additional Context

- Plugin Version: 1.18.0
- WordPress Version: 5.9
- Operating System: macOS
- Browser: Chrome

timarney:
## Context

Adds keyboard navigation to list of Custom Fonts 

Details: https://github.com/GoogleForCreators/web-stories-wp/issues/10878#issue-1165544840

## Summary

Adds `listbox` / aria functionality to the Font List --- tabbing + arrow keys  

Per the comment this PR doesn't use `useRovingTabIndex` to handle tabs + arrows

> useRovingTabIndex probably doesn't work well, as you've pointed out, because it is a bit overkill and only handles arrow navigation, but not things like aria-selected, which we need.

## To-do

I don't see any test written for the existing `custom fonts feature `

There are some starter utils functions used in the e2e tests for #10808 . Those could be used as a starter to test this. 

The remove font utility function needs to be updated to remove fonts by font family vs the first font in the list.

## Testing Instructions

1. Enable the custom fonts feature 
2. Add some fonts 
```
https://overpass-30e2.kxcdn.com/overpass-extrabold-italic.ttf
https://overpass-30e2.kxcdn.com/overpass-regular.ttf
https://overpass-30e2.kxcdn.com/overpass-semibold-italic.ttf
```

3. Use the keyboard to navigate the font list + delete fonts

## Checklist

---

Fixes #10878


swissspidy:
A couple of notes from preliminary testing:

1. Functionality-wise it works well with a keyboard!
2. `useRovingTabIndex` probably doesn't work well, as you've pointed out, because it is a bit overkill and only handles arrow navigation, but not things like `aria-selected`, which we need.
3. The `tabIndex={0}` and `tabIndex={-1}` can probably be removed, they don't do anything AFAICT
4. The listbox needs [`aria-activedescendant`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-activedescendant) pointing to the ID of the currently selected font. So yes, every `role=option` item needs an ID. This should work:  
  ```
  id={`font-${id}`}
  ```
5. We don't need the `.active` `className`. To target the active item with CSS we can just use `[aria-selected="true"]`

timarney:
Sounds good .

timarney:
I kept the `tabIndex={0}` on the "listbox" without it I'm getting this scrolling behavior  when I use the Arrow keys

See 👉  https://user-images.githubusercontent.com/62242/158698069-edead673-978b-4890-8d30-9f7abb5bf424.mp4

timarney:
The font deletion for the e2e tests I wrote for this seem to pass / fail --- will look deeper into why they don't always clean themselves up.  When they fail to delete that causes issues with the other tests as the code tries to re-add the same fonts etc... 

https://github.com/GoogleForCreators/web-stories-wp/runs/5593306519?check_suite_focus=true#step:9:22

☝️ 

This was when I was using promise all map

timarney:
Moving this to ready to review so I can troubleshoot the e2e test with some snapshots.

packages/e2e-test-utils/src/customFonts.js
```diff
@@ -13,11 +13,54 @@
...
+  } catch (e) {
+    return [];
+  }
+};
 
 /**
- * Internal dependencies
+ * Get custom font list
+ *
+ * @return {Promise<[]>}
```
swissspidy: To address lint warnings:

```suggestion
 * @return {Promise<[Object<{name: string, url: string}]>} Fonts list.
```

packages/e2e-test-utils/src/customFonts.js
```diff
@@ -13,11 +13,54 @@
...
+  return optionsText.map((option) => {
+    const font = option.split('\n');
+    return { name: font[0], url: font[1] };
+  });
+}
+
+/**
+ * Get custom font list
+ *
+ * @return {Promise<[]>}
```
swissspidy: To address lint warnings:

```suggestion
 * @return {Promise<[Object<{name: string, url: string}]>} Fonts list.
```

packages/e2e-test-utils/src/takeSnapshot.js
```diff
@@ -20,6 +20,9 @@
 import percySnapshot from '@percy/puppeteer';
 
 async function takeSnapshot(page, name, options) {
+  // @todo -- remove --- for local debug
+  // await page.screenshot({ path: `./${name}-screenshot.png`, fullPage: true });
```
swissspidy: I often do this too locally 👍 

But let's remove it :)

packages/e2e-test-utils/src/customFonts.js
```diff
@@ -13,11 +13,54 @@
...
+};
 
 /**
- * Internal dependencies
+ * Get custom font list
+ *
+ * @return {Promise<[]>}
  */
-import visitSettings from './visitSettings';
+export const getSelectedFontList = async () => {
```
swissspidy: There's only ever one selected font, no? So why return an array? This is a bit misleading.

packages/e2e-tests/src/specs/editor/publishPanel/adminUser.js
```diff
@@ -79,7 +79,7 @@ describe('Publish panel in document tab', () => {
     expect(optionListBeforeSearch).toBeDefined();
     expect(optionListBeforeSearch.length).toBeGreaterThan(1);
 
-    await page.keyboard.type('auth');
+    await page.keyboard.type('auth', { delay: 10 }); // was seeing some typing / input issues
```
swissspidy: When/where did this happen? I've never seen this test fail.

packages/e2e-tests/src/specs/editor/fontCheck/fontCheck.js
```diff
@@ -141,7 +144,7 @@ describe('Font Check', () => {
     await expect(page).toMatchElement('h2', { text: 'Dashboard' });
   });
 
-  it('should redirect to dashboard when pressing ESC', async () => {
+  it('should redirect to dashboard when pressing ESC !', async () => {
```
swissspidy: ?

packages/e2e-test-utils/src/customFonts.js
```diff
@@ -35,20 +77,19 @@ export const addCustomFont = async (fontUrl) => {
 
   await page.keyboard.type(fontUrl);
   await expect(page).toClick('button', { text: 'Add Font' });
-  // TODO(#10878): Target font by font name.
-  await page.waitForSelector('[aria-label="Remove font"]');
+  await page.waitForTimeout(300);
```
swissspidy: What is the timeout for?

It's preferable to wait for a specific event or selector instead of magic timeout numbers (e.g. network response, element appearing, element changing state, etc.)

packages/e2e-test-utils/src/customFonts.js
timarney: Good point

packages/e2e-test-utils/src/customFonts.js
spacedmonkey:
```suggestion
  await page.waitForTimeout(300);
  await expect(page).toMatch(fontUrl);
```

packages/e2e-tests/src/specs/editor/publishPanel/adminUser.js
timarney: I'll remove and see if it's more stable now after the other fixes

I was noticing when entering 3 fonts in a row the typing would sometime miss chars --- I think the underlying issue could have been the field losing focus.

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
```diff
@@ -0,0 +1,113 @@
...
+const removeAllFonts = async () => {
+  await visitSettings();
+  let fonts = await getFontList();
+
+  if (fonts.length === 0) {
+    return;
+  }
+
+  await asyncForEach(fonts, async (font) => {
+    await removeCustomFont(font.name);
```
spacedmonkey: Instead of this could do. 

```
await Promise.all( fonts.map( ( { name } ) => removeCustomFont(name) ) );
```

Might be a little simplier to read / understand.

packages/e2e-tests/src/specs/editor/fontCheck/fontCheck.js
timarney: Leftover from debugging -- watch mode to trigger a change --- will remove.

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
```diff
@@ -0,0 +1,113 @@
...
+    const font1 = findByUrl(fonts, OPEN_SANS_CONDENSED_LIGHT_URL);
+    expect(font1.name).toStrictEqual(OPEN_SANS_CONDENSED_LIGHT);
+    const font2 = findByUrl(fonts, OPEN_SANS_CONDENSED_BOLD_URL);
+    expect(font2.name).toStrictEqual(OPEN_SANS_CONDENSED_BOLD);
+
+    await page.keyboard.press('Tab');
+    await page.keyboard.press('ArrowDown');
+    await page.keyboard.press('ArrowDown');
+    await page.keyboard.press('ArrowUp');
+    await takeSnapshot(page, `Select custom font`);
```
spacedmonkey: Do we need a screenshot here?

packages/e2e-tests/src/specs/editor/fontCheck/fontCheck.js
spacedmonkey: ?

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
timarney:
```
await Promise.all( fonts.map( ( { name } ) => removeCustomFont(name) ) );
```

So this is an interesting one.

I was using promise.all map --- but the font deletion would sometimes happen and other times fail.

I would need to play around with it further but I did notice when capturing some screens that the "next" deletion looked like it was trying to fire before the first dialog box "delete font" click was happening.  In turn one or more fonts would still end up in the list.

With the async for each --- the code awaits properly + runs in order without the next delete firing pre the first one being finished.

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
timarney: Will remove -- I don't think we need it.

packages/wp-dashboard/src/components/editorSettings/customFonts/index.js
```diff
@@ -242,6 +258,32 @@ function CustomFontsSettings({
     [handleOnSave]
   );
 
+  useEffect(() => {
+    const el = currentFontsRowsRef.current[`row-${currentFontsFocusIndex}`];
```
spacedmonkey, line 262: What is happening here?

packages/e2e-tests/src/specs/editor/fontCheck/fontCheck.js
timarney: Yup flagged above --- I should have flipped this PR back to draft 

https://github.com/GoogleForCreators/web-stories-wp/pull/10918#issuecomment-1072258228

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
spacedmonkey: If they need to run it order, then just do a for loop and add an ignore to the jslint that comes up. No need for another function.

packages/wp-dashboard/src/components/editorSettings/customFonts/index.js
```diff
@@ -242,6 +258,32 @@ function CustomFontsSettings({
...
+        setCurrentFontsFocusIndex((index) => Math.max(0, index - 1));
+      } else if (key === 'ArrowDown') {
+        evt.preventDefault();
+        setCurrentFontsFocusIndex((index) =>
+          Math.min(customFonts.length - 1, index + 1)
+        );
+      }
+    },
+    [customFonts]
+  );
```
spacedmonkey, line 290: A comment might be useful here. This code isn't clear.

packages/wp-dashboard/src/components/editorSettings/customFonts/index.js
timarney: That's a ref back to the row that **needs to get focus**

currentFontsRowsRef.current is an array of refs to each one of the rows

packages/wp-dashboard/src/components/editorSettings/customFonts/index.js
```diff
@@ -280,9 +322,25 @@ function CustomFontsSettings({
...
+              aria-activedescendant={
+                currentFontsActiveId ? currentFontsActiveId : customFonts[0]?.id
+              }
+            >
+              {customFonts.map(({ id, family, url }, index) => (
+                <FontRow
+                  id={`font-${id}`}
+                  ref={(el) =>
+                    (currentFontsRowsRef.current[`row-${index}`] = el)
+                  }
```
spacedmonkey: ?

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
timarney: Needs to be async --- need to wait for the previous font deleting to finish 

press delete 
font dialog -> delete

repeat

packages/wp-dashboard/src/components/editorSettings/customFonts/index.js
timarney: Just pushed a comment --- let me know if it's helpful and or feel free to push a commit

packages/wp-dashboard/src/components/editorSettings/customFonts/index.js
timarney: Added some comments per the other comment below

packages/e2e-tests/src/specs/editor/publishPanel/adminUser.js
timarney: removed.

packages/e2e-tests/src/specs/editor/fontCheck/fontCheck.js
timarney: removed.

packages/wp-dashboard/src/components/editorSettings/customFonts/index.js
timarney: Same type of logic is here:
https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/design-system/src/components/datalist/list/list.js#L186

packages/wp-dashboard/src/components/editorSettings/customFonts/index.js
timarney: Same logic as here I added a small comment

For the refs:
https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/design-system/src/components/datalist/list/list.js#L270

packages/e2e-test-utils/src/customFonts.js
timarney: There seems to be a slight delay before the font shows up.

I'll re-test.  If I can remove I will 

====

Else ...

Agreed --- just thinking about this.

Maybe we pass in the font name as well and use --- I don't really want to add another param but it would ensure the font has been added.


```javascript
await page.waitForSelector(`[aria-label="Delete ${fontName}"]`);
```

packages/e2e-test-utils/src/customFonts.js
timarney: Updated here:
https://github.com/GoogleForCreators/web-stories-wp/pull/10918/files#diff-15102bcca0cb4bffec2cea43a88ada5c191041b40f7e784266ac7b5cd2161a1dR55

packages/e2e-test-utils/src/customFonts.js
timarney: Removed --- seems to be working.

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
timarney: @spacedmonkey your right for loop is all that's needed here - 

Tested here : https://codesandbox.io/s/reverent-fermi-h1vgex

timarney: Pushed the update

packages/e2e-test-utils/src/customFonts.js
```diff
@@ -13,11 +13,60 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+function parseText(optionsText) {
+  return optionsText.map((option) => {
+    const font = option.split('\n');
+    return { name: font[0], url: font[1] };
```
spacedmonkey:
```suggestion
    const [ name, url ] = option.split('\n');
    return { name, url };
```

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
```diff
@@ -0,0 +1,105 @@
...
+  if (fonts.length === 0) {
+    return;
+  }
+
+  for (let i = 0; i < fonts.length; i++) {
+    // eslint-disable-next-line no-await-in-loop -- async for
+    await removeCustomFont(fonts[i].name);
+  }
+
+  await page.waitForTimeout(500);
```
spacedmonkey: Why is this wait needed here?

packages/e2e-test-utils/src/customFonts.js
```diff
@@ -13,11 +13,60 @@
...
+ */
+export const getSelectedFont = async () => {
+  try {
+    const optionsText = await page.evaluate(() =>
+      Array.from(
+        document.querySelectorAll(
+          'div[role=listbox] [aria-selected=true] div:first-child'
+        ),
+        (element) => element.innerText
+      )
```
spacedmonkey: Why are querying all the selectors here. Why can't we just do `document.querySelector` instead of `document. querySelectorAll`

spacedmonkey: I understnad what is happening here. I still find it weird, we are querying all of elements, when we only need one them.

packages/e2e-test-utils/src/customFonts.js
```diff
@@ -35,20 +83,19 @@ export const addCustomFont = async (fontUrl) => {
...
  */
-export const removeCustomFont = async () => {
-  await visitSettings();
-  // TODO(#10878): Target font by font name.
-  await page.hover('[aria-label="Remove font"]');
-  await page.click('[aria-label="Remove font"]');
+export const removeCustomFont = async (fontName) => {
+  await page.waitForSelector(`[aria-label="Delete ${fontName}"]`);
+  await page.focus(`[aria-label="Delete ${fontName}"]`);
+  await page.click(`[aria-label="Delete ${fontName}"]`);
```
spacedmonkey:
```suggestion
  const selector = `[aria-label="Delete ${fontName}"]`;
  await page.waitForSelector(selector);
  await page.focus(selector);
  await page.click(selector);
```

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
```diff
@@ -0,0 +1,105 @@
...
+  const found = arr.filter((o) => o.url === val);
+  if (found) {
+    return found[0];
+  }
+  return [];
+};
+
+const removeAllFonts = async () => {
+  await visitSettings();
+  let fonts = await getFontList();
```
spacedmonkey: I don't think this needs to be a let, we can just have two consts here. I don't see the value of reusing the variable here.

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
```diff
@@ -0,0 +1,105 @@
...
+
+const FONT_BASE_URL = `${process.env.WP_BASE_URL}/wp-content/e2e-assets`;
+const OPEN_SANS_CONDENSED_LIGHT = 'Open Sans Condensed Light';
+const OPEN_SANS_CONDENSED_LIGHT_URL = `${FONT_BASE_URL}/OpenSansCondensed-Light.ttf`;
+const OPEN_SANS_CONDENSED_BOLD = 'Open Sans Condensed Bold';
+const OPEN_SANS_CONDENSED_BOLD_URL = `${FONT_BASE_URL}/OpenSansCondensed-Bold.ttf`;
+const OPEN_SANS_CONDENSED_LIGHT_ITALIC_URL = `${FONT_BASE_URL}/OpenSansCondensed-LightItalic.ttf`;
+
+const findByUrl = (arr, val) => {
+  const found = arr.filter((o) => o.url === val);
```
spacedmonkey: Could you using `.find` here?

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
```diff
@@ -0,0 +1,105 @@
...
+  expect(fonts).toHaveLength(0);
+};
+
+describe('Custom Fonts', () => {
+  withExperimentalFeatures(['customFonts']);
+
+  beforeEach(async () => {
+    await visitSettings();
+    await removeAllFonts();
+  });
```
spacedmonkey, line 62:
```suggestion
  beforeEach(async () => {
    await visitSettings();
 });
    
 beforeAll(async () => {   
    await removeAllFonts();
 });
```

We don't see to do this before each test. Reset back to no font at start and after each test. That should be enough

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
```diff
@@ -0,0 +1,105 @@
...
+    await page.keyboard.press('ArrowUp');
+    const selected = await getSelectedFont();
+    expect(selected.name).toStrictEqual(OPEN_SANS_CONDENSED_LIGHT);
+
+    const listbox = await page.$('[role="listbox"]');
+    const ariaActiveDescendant = await page.evaluate((el) => {
+      const id = el.getAttribute('aria-activedescendant');
+      return document.getElementById(id).getAttribute('aria-selected');
+    }, listbox);
+    expect(ariaActiveDescendant).toBe('true');
```
spacedmonkey, line 94: Can we just use a selector here. 

```
const ariaActiveDescendant = await page.$('[role="listbox"] [aria-selected="true"]');
```

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
```diff
@@ -0,0 +1,105 @@
...
+    await page.keyboard.press('Tab');
+
+    const deleteLabel = await page.evaluate(() =>
+      document.activeElement.getAttribute('aria-label')
+    );
+
+    expect(deleteLabel).toBe(`Delete ${OPEN_SANS_CONDENSED_LIGHT}`);
```
spacedmonkey:
```suggestion
     await expect(page).toMatchElement(`button[aria-label="Delete ${OPEN_SANS_CONDENSED_LIGHT}"]`);
```

This seems simpler.

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
timarney: The font list seems to have a slight delay with rendering --- without

```
 Expected length: 0
    Received length: 1
    Received array:  [{"name": "Open Sans Condensed Light Italic", "url": "http://localhost:8899/wp-content/e2e-assets/OpenSansCondensed-LightItalic.ttf"}]
```

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
timarney: I think that would tell us that there's an option with aria-selected true but won't let us know if the 
`aria-activedescendant` we're setting matches the `id` of the `aria-selected` item.

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
timarney: Updated 

https://github.com/GoogleForCreators/web-stories-wp/pull/10918/files/9bf20a7d87aa6f3b690d8ff6a3aa17924d647bc5#diff-15102bcca0cb4bffec2cea43a88ada5c191041b40f7e784266ac7b5cd2161a1dR18

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
timarney: Sure - https://github.com/GoogleForCreators/web-stories-wp/pull/10918/files#diff-8a323053c0cb1326f851722b15dc1f8f75dd5a7336feec051e3dd91f41babda3R35

packages/e2e-test-utils/src/customFonts.js
timarney: Updated to work better for working with single element `document.querySelector`

packages/e2e-test-utils/src/customFonts.js
```diff
@@ -35,20 +79,23 @@ export const addCustomFont = async (fontUrl) => {
...
+export const removeCustomFont = async (fontName) => {
+  const selector = `[aria-label="Delete ${fontName}"]`;
+  await page.waitForSelector(selector);
+  await page.focus(selector);
+  await page.click(selector);
   await page.waitForSelector('[role="dialog"]');
   await expect(page).toClick('button', { text: 'Delete Font' });
+  // add a small delay seems to take a moment for the font list to catchup
+  // return from dialog + re-focus on list
+  await page.waitForTimeout(1000);
```
swissspidy: Can we instead wait for dialog to disappear and for the font to disappear from the list too?

We should avoid `waitForTimeout` where possible

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
swissspidy: Can we instead wait for the list to be empty or disappear.

We should avoid `waitForTimeout` where possible.

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
```diff
@@ -0,0 +1,101 @@
...
+    await removeAllFonts();
+  });
+
+  afterEach(async () => {
+    await visitSettings();
+    await removeAllFonts();
+  });
+
+  it('should add a fonts and handle keyboard navigation', async () => {
+    await addCustomFont(OPEN_SANS_CONDENSED_LIGHT_URL); ////
```
swissspidy:
```suggestion
    await addCustomFont(OPEN_SANS_CONDENSED_LIGHT_URL);
```

packages/e2e-test-utils/src/customFonts.js
timarney: Yes --- but there's also that delay for the font list rendering --- which I'm hoping we can get rid of as well

https://github.com/GoogleForCreators/web-stories-wp/pull/10918#discussion_r831231006

Maybe I can wait for the dialog to disappear and use 

https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#pagewaitforfunctionpagefunction-options-args

Looks like there's some other code doing something similar 

```
// wait for token to render
  await page.waitForFunction(
    (originalNumChildren) =>
      document.getElementById('web_story_tag-input').parentNode.children
        .length > originalNumChildren,
    {},
    numChildren
  );
```

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
timarney: Need to figure out the delay for for the font list rendering / the items to get updated.

packages/e2e-test-utils/src/customFonts.js
```diff
@@ -35,20 +79,33 @@ export const addCustomFont = async (fontUrl) => {
...
+  await page.click(selector);
   await page.waitForSelector('[role="dialog"]');
   await expect(page).toClick('button', { text: 'Delete Font' });
+  await page.waitForFunction(
+    (originalNumChildren) => {
+      const count =
+        document.querySelector('div[role=listbox]')?.children?.length;
+      return !count || count === originalNumChildren - 1 ? true : false;
+    },
+    { polling: 300 },
```
timarney: If the latest commit works I'll see about removing the polling

packages/e2e-test-utils/src/customFonts.js
timarney: Removed 

https://github.com/GoogleForCreators/web-stories-wp/pull/10918/files#diff-15102bcca0cb4bffec2cea43a88ada5c191041b40f7e784266ac7b5cd2161a1dR102

Now checking count of child nodes

packages/e2e-test-utils/src/customFonts.js
timarney: Removed https://github.com/GoogleForCreators/web-stories-wp/pull/10918/commits/f8edcdff1d8c35ecf14211493a63b6699b27c05a

packages/e2e-tests/src/specs/dashboard/settings/adminUser/customFonts.js
timarney: Removed this as well via https://github.com/GoogleForCreators/web-stories-wp/pull/10918/files#diff-15102bcca0cb4bffec2cea43a88ada5c191041b40f7e784266ac7b5cd2161a1dR102

swissspidy (approved):
Thanks for taking the time to improve the assertions!

I just made a tiny tweak to the `waitForFunction()` call and added a Percy snapshot while at it.

This looks good to go now :)

spacedmonkey (approved):

PR #10920, opened 2022-03-15
Context menus: Move tooltip out of the MenuIcon component

### LINKED ISSUE #10884: Floating menu: Icon buttons not showing tooltip when in focus
Originally reported by @barklund: https://github.com/GoogleForCreators/web-stories-wp/pull/10875/files#r824197428

## Bug Description
`IconButton` component shows a tooltip when hovering but not when focused.

## Expected Behaviour
`IconButton` should show a tooltip when focused, too.

## Steps to Reproduce
Focus on a flip toggle with keyboard -- see the tooltip not showing.

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

miina:
## Summary
Fixes the issue where element in focus with keyboard does not show the tooltip.

## Relevant Technical Choices
Moves `Tooltip` out of the `MenuIcon` component to `MenuButton` component.
Since the `Tooltip` was displayed inside the `MenuButton` and never wrapped it, it only worked when hovering, never when focusing.

## Testing Instructions

1. Focus on the quick-action menu items or Page menu items using keyboard. Verify that the tooltips display when focusing.
2. Add a text and focus on the Bold / Italic / Underline in the floating design menu (toolbar). Verify the tooltips display when focusing.
3. Verify that tooltips show correctly when hovering as well.

No

No

No

## Checklist

---

Fixes #10884


packages/design-system/src/components/contextMenu/components/button.js
```diff
@@ -26,6 +26,7 @@ import { useMemo, forwardRef } from '@googleforcreators/react';
  */
 import { Button as BaseButton } from '../../button';
 import { useContextMenu } from '../contextMenuProvider';
+import { Tooltip, TOOLTIP_PLACEMENT } from '../../tooltip';
```
miina: See the "Relevant Technical Choices" for the reason of the changes here.
@BrittanyIRL / @samwhale Let me know if you have any concerns with the change.

packages/design-system/src/components/contextMenu/components/button.js
```diff
@@ -84,6 +85,9 @@ const Button = forwardRef(function Button(
     onFocus,
     forcePadding = false,
     dismissOnClick = true,
+    children,
+    title,
```
samwhale: 💭  since this is related to the tooltip and not the button, I think updating this to `tooltipTitle` would be more descriptive.
```suggestion
    tooltipTitle,
```

packages/design-system/src/components/contextMenu/components/button.js
```diff
@@ -84,6 +85,9 @@ const Button = forwardRef(function Button(
     onFocus,
     forcePadding = false,
     dismissOnClick = true,
+    children,
```
samwhale: 🤔  I don't think `children` needs to be destructured here since it'd be passed into the button through `props`
```suggestion
```

packages/design-system/src/components/contextMenu/components/button.js
```diff
@@ -122,19 +126,23 @@ const Button = forwardRef(function Button(
...
+        isHorizontal={isHorizontal}
+        forcePadding={forcePadding}
+        onBlur={handleBlur}
+        onClick={handleClick}
+        onFocus={handleFocus}
+        {...props}
+      >
+        {children}
+      </StyledButton>
+    </Tooltip>
```
samwhale: 💭  the right click menus don't have tooltips - so I think it'd be better to only render the tooltip when it's needed.

```js
  const Button = <StyledButton
      ref={ref}
      id={elementId}
      tabIndex={focusedId === elementId ? 0 : -1}
      role="menuitem"
      isIconMenu={isIconMenu}
      isHorizontal={isHorizontal}
      forcePadding={forcePadding}
      onBlur={handleBlur}
      onClick={handleClick}
      onFocus={handleFocus}
      {...props}
    />;

  if (title) {
    return (
       <Tooltip placement={tooltipPlacement} title={title}>
         <Button/>
       </Tooltip>
    )
  }

  return <Button />;
```

samwhale (commented):
Oh nice catch - not being able to see those without the keyboard is a bad thing.

💭  I'm not sure that the `Button` needs the tooltip. The quick action menu is the only thing that needs it, right? It could be added directly to `packages/story-editor/src/components/canvas/pageSideMenu.js`.

I'm not convinced that the `Button` needs to be expanded to be able to render tooltips when tooltips could be added on an as-needed basis through composability. Thoughts?

packages/design-system/src/components/contextMenu/components/button.js
miina: Hm, it didn't work otherwise for the quick-action menu though.

samwhale: oh weird! Okay then ignore me here 😄

miina:
> I'm not convinced that the Button needs to be expanded to be able to render tooltips when tooltips could be added on an as-needed basis through composability. Thoughts?

@samwhale The new floating menu / toolbar is also using these and needs tooltips. 

However, I can of course move these directly to the usage and remove from the `Button`. True that it's not used in the right-click menu so might be a better option.

packages/design-system/src/components/contextMenu/components/button.js
miina: @samwhale I've moved the `Tooltip` out of the `MenuButton` now as well per your note that the `Button` can be just wrapped into `Tooltip` as needed as we generally do instead of including it into the component.

packages/story-editor/src/components/floatingMenu/elements/textAlign.js
```diff
@@ -101,7 +102,10 @@ function TextAlign() {
   const [offsetLeft, setOffsetLeft] = useState(0);
 
   // Record left position of this button in the parent design menu
-  useEffect(() => setOffsetLeft(buttonRef.current.offsetLeft + OFFSET_X), []);
+  useEffect(
+    () => setOffsetLeft(buttonRef.current.parentNode.offsetLeft + OFFSET_X),
```
miina, line 106: Using `parentNode` here now as the button is wrapped into a tooltip. An alternative would be creating a new wrapper around the tooltip as well and using a `ref` for that, however, this might need some adjustment anyway related to #10872 cc @barklund

packages/design-system/src/components/contextMenu/components/icon.js
```diff
@@ -39,32 +34,22 @@ const IconWrapper = styled.div`
...
       <VisuallyHidden>{title}</VisuallyHidden>
-      <TooltipWrapper>
-        <Tooltip placement={placement} title={title} {...props}>
-          <IconWrapper className={className}>{children}</IconWrapper>
-        </Tooltip>
-      </TooltipWrapper>
+      <IconWrapper className={className} {...props}>
+        {children}
+      </IconWrapper>
     </>
```
samwhale, line 48: 👍  I think this makes the icon look much nicer too!

packages/story-editor/src/components/canvas/pageSideMenu.js
```diff
@@ -104,21 +105,20 @@ function PageSideMenu() {
...
+                    {...quickAction}
                   >
-                    <Icon />
-                  </ContextMenuComponents.MenuIcon>
-                </ContextMenuComponents.MenuButton>
+                    <ContextMenuComponents.MenuIcon title={label}>
+                      <Icon />
+                    </ContextMenuComponents.MenuIcon>
+                  </ContextMenuComponents.MenuButton>
+                </Tooltip>
```
samwhale, line 121: 👍  I like this!

samwhale (approved):
Nice, I think this is great!

BrittanyIRL (changes requested):
Works great for focus but causes a double tooltip for cursor on hover?

seeing double

miina:
> Works great for focus but causes a double tooltip for cursor on hover?

@BrittanyIRL Hm, I'm not seeing it locally nor in the QA environment 🤔 Could you double check if you still see it / if you can see it in the QA environment as well?
tooltips

BrittanyIRL:
> > Works great for focus but causes a double tooltip for cursor on hover?
> 
> @BrittanyIRL Hm, I'm not seeing it locally nor in the QA environment 🤔 Could you double check if you still see it / if you can see it in the QA environment as well? tooltips

well it's fine now, how odd!!

BrittanyIRL (approved):

PR #10922, opened 2022-03-15
Code Quality: Remove errant `fontWeight` references

swissspidy:
## Context

This was noticed in ##10886

`fontWeight` was inlined to the text element’s `content` in the `v0017_inlineTextProperties` migration.

Yet, text sets and presets still had this unused property.

## Summary

This PR simply cleans up this unused property to avoid confusion.

## User-facing changes

None

## Testing Instructions

1.

## Checklist

---

Fixes #


packages/story-editor/src/components/library/panes/text/textPresets.js
```diff
@@ -62,7 +62,6 @@ const DEFAULT_PRESET = {
     'text preset content',
     'web-stories'
   ),
-  fontWeight: FONT_WEIGHT.NORMAL,
```
miina, line 65: These are actually currently used in `FontPreview` since `fontWeight` is taken directly from the presets. If we remove it here, we should get the `fontWeight` from the content instead in `FontPreview`.

swissspidy: Ah, thanks for catching that!

In that case, I suppose we can use `getTextInlineStyles` for that, which is already being used in `FontPreview`

The alternative would be to make sure it's discarded when actually inserting the preset, but that seems fragile.

swissspidy: Addressed now in 6a046aef788060818ac09e05562f2c4334f3ce8c

Previews look better now

miina (approved):

PR #10937, opened 2022-03-16
Improve handling of corrupt page templates data

### LINKED ISSUE #10936: Saved templates not visible
## Bug Description

Saved templates are not appearing in the editor view. The list is empty and there is no errors in the console. I noticed that the API result of saved templates has a lot of templates but there is two types of items in the list. One type has `story_data` as an object with data on it and other type is `story_data` as an array without data on it. Could the view be failing silently because of the empty `story_data` arrays? On my other environments the saved template views are working properly and there is no empty `story_data` arrays in the response.

## Expected Behaviour

Saved templates should appear in the template list just like the default templates does.

## Steps to Reproduce

I cannot provide the steps to reproduce because I don't know what happened. Since the third of February the saved template list in the editor view has been empty. Could this be related to some unsuccessful migration?

## Screenshots

Here is a screenshot of the editor view with console open showing the results of `/wp-json/web-stories/v1/web-story-page/?context=edit&per_page=100&page=1&_web_stories_envelope=true&_fields=id%2Cstory_data%2Cbody%2Cheaders&_embed=wp%3Afeaturedmedia&_locale=user` API request.
Screenshot 2022-03-16 at 10 42 55

## Additional Context

- Plugin Version: 17.0.1
- WordPress Version: 5.9.2
- Operating System: macOS Big Sur 11.6
- Browser: Chrome, Brave, Firefox

swissspidy:
## Context

As reported in #10936, the Saved Templates library tab is empty if there is 1 page template with empty `story_data`.

`story_data` is empty if the story data stored in `post_content_filtered` is corrupt and not valid JSON anymore.

This leads to some JS errors being thrown and silenced, causing the whole tab to be empty.

## Summary

This PR improves handling of corrupt page templates by ensuring certain object keys are always present, even if `story_data` is empty.

## To-do

In the future, if this becomes a more frequent issue, we might want to look into ways to surface this to the users.

## Testing Instructions

1. Create a new story with some content
2. Save as a template
3. Modify the template in the database, emptying the `post_content_filtered` column
4. Go to the editor and to the Saved Templates list
5. Verify that the all saved templates appear as expected

No

Adds new error tracking for when loading page templates fails.

No

## Checklist

---

Fixes #10936

### DIFF
--- a/packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
+++ b/packages/story-editor/src/components/library/panes/pageTemplates/pageTemplatesPane.js
@@ -31,6 +31,7 @@ import {
   LOCAL_STORAGE_PREFIX,
 } from '@googleforcreators/design-system';
 import { DATA_VERSION, migrate } from '@googleforcreators/migration';
+import { trackError } from '@googleforcreators/tracking';
 
 /**
  * Internal dependencies
@@ -152,7 +153,8 @@ function PageTemplatesPane(props) {
           setNextTemplatesToFetch(nextTemplatesToFetch + 1);
         }
       })
-      .catch(() => {
+      .catch((err) => {
+        trackError('saved_templates', err.message);
         setNextTemplatesToFetch(false);
         setSavedTemplates((_savedTemplates) => _savedTemplates ?? []);
       })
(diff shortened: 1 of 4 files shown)


spacedmonkey:
If the data is invalid in the database. Could we just fix this at the PHP level. If the data is invalid, either remove api or empty an empty array / object?

swissspidy:
Mhm yeah maybe; not sure what's the best way there.

`Stories_Base_Controller` already does _some_ handling and returning `null` if it's invalid JSON. Returning an empty object would not change anything and still cause client-side issues.

So this type of hardening here is required nonetheless.

Also worth noting that `Stories_Base_Controller` is also used for stories themselves, so need to be careful not to paper over errors there.

In short, I think PHP-level changes need some thorough thinking. We can open a ticket to discuss this further.

maxyinger (approved):

spacedmonkey (approved):

PR #10957, opened 2022-03-17
Autoset as Background: Show snackbar message only once

### LINKED ISSUE #10782: Autoset as Background: Do not display snackbar message all the time
## Feature Description

This is a follow-up to #10458

As noted [on the forums](https://wordpress.org/support/topic/suggestion-option-to-hide-image-background-popup/), we display a `Full bleed images and videos are automatically set as background. Double click to scale and position at any time.` snackbar message **every time** the user performs this action.

4VTVnr4

This is annoying. We should only display it once and then store that in localStorage so we don't display it again in the future.

After the first time the user will know how it works.

marianakyrkosh:
## Context

After setting a background image a message is shown: "Full bleed images and videos are automatically set as background. Double click to scale and position at any time". This message is displayed every time a background image is set. It should be shown only once.

## Summary

A property web_stories_background_is_set_dialog_dismissed is added to localStorage after a background message is shown, so the message is not shown anymore.

## User-facing changes

A background message is shown only once.
image
image
image

## Testing Instructions

1. Add an image to the story.
2. Drag the image to cover whole area of the story so the image is set as a background.
3. A message is shown.
4. A property web_stories_background_is_set_dialog_dismissed is added to localStorage.
5. Remove the background, drag a new image to cover the whole area of the Page so the image is set as a background. 
7. The message should not be shown.

## Checklist

---

Fixes #10782


.husky/pre-commit
```diff
@@ -1,8 +1,4 @@
 #!/bin/sh
-
-export NVM_DIR="$HOME/.nvm"
```
miina: @mariana-k I guess this change is not intentional?

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
```diff
@@ -39,19 +39,26 @@ function useFullbleedMediaAsBackground({ selectedElement }) {
...
+      if (!isBackgroundMessageShown) {
+        showSnackbar({ 
+          message: __(
+            'Full bleed images and videos are automatically set as background. Double click to scale and position at any time.',
+            'web-stories'
+          ),
+          dismissible: true,
+        });
+        isBackgroundMessageShown = true;
+        localStorage.setItem('web_stories_is_background_message_shown', isBackgroundMessageShown);
```
swissspidy: I don't see much value in `isBackgroundMessageShown` being a `let`. Could probably just be this:

```suggestion
        localStorage.setItem('web_stories_is_background_message_shown', true);
```

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
```diff
@@ -39,19 +39,26 @@ function useFullbleedMediaAsBackground({ selectedElement }) {
   const { showSnackbar } = useSnackbar();
 
   const handleFullbleedMediaAsBackground = (target) => {
+    let isBackgroundMessageShown = localStorage.getItem('web_stories_is_background_message_shown');
```
miina: We are generally using `localStore` from the `design-system` package for managing the local storage. You could check how we're using it by just searching for `localStore` in the `story-editor` package, one example is in this component: https://github.com/GoogleForCreators/web-stories-wp/blob/ad0e6193cdaea8bf3348fe8305c73f117140bcdb/packages/story-editor/src/components/colorPicker/confirmationDialog.js#L49

Since getting stuff from local storage is not very fast then it's also using state for it there.

.husky/pre-commit
marianakyrkosh: Reverted the file

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
marianakyrkosh: Done

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
marianakyrkosh: Changed the implementation like Miina suggested

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
```diff
@@ -25,8 +33,17 @@ import { MEDIA_ELEMENT_TYPES } from '@googleforcreators/elements';
  */
 import isTargetCoveringContainer from '../../../utils/isTargetCoveringContainer';
 import { useStory, useCanvas } from '../../../app';
+import { CONFIRMATION_BACKGROUND_MESSAGE_STORAGE_KEY } from './constants';
```
miina: `utils` folders are usually for utils only, so the constants file would be better outside of it. Since `canvas` doesn't have its own constants file and this constant is used only in this file, I think we could just move this constant directly to this file instead of creating a separate file for it. Thoughts?

On the name -- maybe `BACKGROUND_AUTOSET_MESSAGE_STORAGE_KEY` could be a more understandable name?

miina: Or `DISMISS_BACKGROUND_AUTOSET_MESSAGE_STORAGE_KEY`

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
```diff
@@ -46,13 +63,23 @@ function useFullbleedMediaAsBackground({ selectedElement }) {
...
+            'web-stories'
+          ),
+          dismissible: true,
+        });
+        localStore.setItemByKey(
+          LOCAL_STORAGE_PREFIX[CONFIRMATION_BACKGROUND_MESSAGE_STORAGE_KEY],
+          !isBackgroundSnackbarMessageDismissed
+        );
+        setIsBackgroundSnackbarMessageDismissed(
+          !isBackgroundSnackbarMessageDismissed
```
miina: Since we're never setting it back to `false` then for readability, we could use just `true` here and above.

marianakyrkosh: Done

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
marianakyrkosh: Done

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
```diff
@@ -16,7 +16,15 @@
...
-import { useSnackbar } from '@googleforcreators/design-system';
+import { useState } from '@googleforcreators/react';
+import {
+  useSnackbar,
+  LOCAL_STORAGE_PREFIX,
+  localStore,
+} from '@googleforcreators/design-system';
+/**
+ * Internal dependencies
+ */
```
miina: Internal dependencies is now twice in the file. It should always be only once and only above actual internal dependencies, currently,  this one is above external deps.

packages/story-editor/src/components/canvas/utils/useFullbleedMediaAsBackground.js
```diff
@@ -27,6 +35,16 @@ import isTargetCoveringContainer from '../../../utils/isTargetCoveringContainer'
 import { useStory, useCanvas } from '../../../app';
 
 function useFullbleedMediaAsBackground({ selectedElement }) {
+  const DISMISS_BACKGROUND_AUTOSET_MESSAGE_STORAGE_KEY =
```
miina: Nit: Let's move this above the function for consistency, we usually keep these out of the functions. (When using constants inside functions, we'd use camelcase.)

miina (approved):
Two nits left, otherwise looks good and works great!

PR #11021, opened 2022-03-22
Floating Menus: Shapes, only show border options when locked

### LINKED ISSUE #10992: Floating Menu: Border Should be Hidden When Locked
Per UX guidelines confirmed by @miina the border and corner radius in floating menus should have the same functionality. Right now they don't because of implementation inconsistency (https://github.com/GoogleForCreators/web-stories-wp/issues/10640#issuecomment-1072788853) 

So let's fix it. 

## Acceptance Criteria 
Corner Radius: Visible in floating menu when locked. 
Border Width (and color? confirm with UX): Visible in floating menu when locked. Hidden from menu when not locked.
Test to verify

BrookeGraham:
## Context
Floating menus for shapes had inconsistencies between when to show corner radius, border width, and border color. These features should work on the same logic for when to appear. 

Design is undecided on the border color functionality inside the floating menu when border radius or border widths are unlocked. For now, all these options including border color should be hidden if they are unlocked and the values are different. 

## Relevant Technical Choices
Since some shapes don't have `border.locked`, eg. circles, we need to check `!border.locked` or `!hasUniformBorder`. I choose to wait until both conditions were met. This allows the user to toggle borders open in the panel, but doesn't remove from floating menu until the sides are actually different. Made radius work with that same logic. 
 

## To-do
@agingoldseco is still deciding what to do with colors / what to do when it is unlocked or there are variable width sides. 

## User-facing changes
|example||
|--|--|
|Circle|circle|
|Radius|corner|
|Border|border|

## Testing Instructions

1. add a shape
2. change border toggle to unlocked
3. change the width of at least one side
4. once one side is different, the border options should be removed from floating toolbar. 
5. repeat for radius

no

no

no

## Checklist

---

Fixes #10992

### DIFF
--- a/packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
+++ b/packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
@@ -28,7 +28,7 @@ import { canSupportMultiBorder } from '@googleforcreators/masks';
  */
 import { useStory } from '../../../app';
 import { DEFAULT_BORDER } from '../../panels/design/border/shared';
-import { Input, Color, useProperties } from './shared';
+import { Input, Color, Separator, useProperties } from './shared';
 
 const Container = styled.div`
   display: flex;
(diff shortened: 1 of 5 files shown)


packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
```diff
@@ -63,25 +63,23 @@ function BorderWidthAndColor() {
...
-  // If both controls are displayed, also add a dash between them
-  const hasBoth = hasUniformBorder && hasBorderWidth;
+  // Only multi-border elements support border opacity
+  const canHaveBorderOpacity = canSupportMultiBorder({ mask });
 
-  // NB: Note that it can never be the case, that both of the above bools, hasUniformBorder
-  // and hasBorderWidth, are false. If so, neither input would be shown. But because they
-  // partially contradict each other, one of the inputs will always render.
+  // We only allow editing border color, if at least one border has a non-zero width
+  const hasBorderWidth = getHasBorderWidth(border);
```
samwhale, line 82: 💭 would be great to add some unit tests here to make sure the input renders when it's supposed to

BrookeGraham: Yeah I wasn't sure if I should add some since the requirements might change?

BrookeGraham: I can add them though.

BrittanyIRL (approved):
Thanks for thinking through a solution that is consistent for the time being and also for leaving notes!

packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
samwhale: Sure! Let's add something simple for now. If/when the requirements change then we can come back and remove those.

BrookeGraham: There were some tests already. Fixed ones needed the new logic and added one more.

maxyinger (approved):

Issue #11024, opened 2022-03-22
Floating Menu: Cannot select border color with eyedropper 


BrittanyIRL:
## Bug Description

If I select an element with a border and use the floating menu's border color to open the color popup and select the eyedropper, when I select the color I want from the canvas, that color is not applied as the border color. 

This works fine if I use the eyedropper from the color popup for the border in the style panel. 

## Expected Behaviour

Should be able to use eyedropper to select new border color in the floating menu. 

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

BrittanyIRL:
@felipebochehin87 I detailed testing on the PR.

felipebochehin87:
Thanks @BrittanyIRL :)

I tested and I still get some weird behavior (testing PR #11075):

When I pick a color from the canvas using the eye dropper located on the side menu, the color is not set instantly. I need to drag and drop the element (in this case, the square) so that the border color is changed. Please see video below:

1. I select a color using the floating menu - color is set
2. I click on the eye dropper (from the Style menu) and select a color from the background or any other image - color is not set
3. I drag and drop the element - the color selected on step 2 is set

https://images.zenhubusercontent.com/235435637/45bcd9db-a838-47b3-a1e1-7cea68faceff/2022_03_28_08_35_51.mp4

BrittanyIRL:
Thanks, Felipe! 

It looks like that's also happening in staging 😬  happy to see if I can't fix it here though. 

It might be impacting more than just borders since that seems to be outside of floating menus.

samwhale:
@felipebochehin87 I think we fixed that! Would you mind checking this when you have some time?

felipebochehin87:
Tested against https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/, using PR #11075.

1. Added a shape
2. Added border
3. Changed color of the border using the color picker from floating menu - color changed as expected
4. Changed color of the border using the color picker from side menu (Style) - color changed as expected
5. Changed color of the border using the eye dropper from side menu (Style) - color changed as expected

https://images.zenhubusercontent.com/235435637/37592032-a8cf-418d-80ef-e6b17101a363/11024.mp4

felipebochehin87:
@BrookeGraham @BrittanyIRL it's happening again... Should we reopen this or create a new bug?

BrittanyIRL:
[...]

PR #11025, opened 2022-03-22
Floating Menu: Fix "Saved Color" (gradient) as border color causes crash

### LINKED ISSUE #11019: Floating Menu: Apply "Saved Color" as border color causes crash
## Bug Description

Found by @LuckynaSan  during bug bash, confirmed by Brittany 

> Editor crashes when you add a masked border to an element & select a saved color on the floating menu bar.

Note, this isn't an issue if you apply color for a border in the panel because saved colors aren't available there for border colors. Furthermore, it seems like it's just gradients causing the crash. Makes sense. 

## Steps to Reproduce

1) Insert element 2) Apply border from floating menu bar 3) open color picker 4) Select saved color (green in this case) 5) Editor crashes with error message below
## Screenshots

Screen Shot 2022-03-22 at 10 42 00 AM

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

BrittanyIRL:
## Context

Saved colors can include gradients, but gradients shouldn't be border colors - that was causing the crash noted in bug bash.

## Summary

I chose to fix this by not showing saved colors in the border color popup of the floating menu because the border color popup in the style pane doesn't show saved colors either and this felt more consistent to me. 

That said, we should discuss (if this hasn't happened yet and I am just unaware) allowing saved styles for borders. I tried that first and it's fine as long as `allowsGradient` is false. There's be an update necessary in `useAddPreset` to grab a color from a border - right now that doesn't exist so adding a new saved color doesn't work. 

## Relevant Technical Choices

- Fixes a casing typo that doesn't really matter because it's false but still `hasEyeDropper` not `hasEyedropper`
- `pickerHasEyedropper` is set to false on the border color level to override the shared eyedropper setting in the floating menu color component.
- `allowsSavedColors` is overridden also so that it doesn't show by default. 

## To-do

While debugging this found that the eyedropper in the border color popup isn't actually getting the border color. I set `pickerHasEyedropper` to false so it's hidden for now but we should fix it. https://github.com/GoogleForCreators/web-stories-wp/issues/11024

## User-facing changes

| Broken | Fixed | 
| --- | --- | 
| https://user-images.githubusercontent.com/10720454/159569958-a9704fd0-ccc2-43e4-aa85-2770587d6b0f.mp4 | https://user-images.githubusercontent.com/10720454/159570083-84cdc6c2-1ce3-4bf8-8cea-a7208c76230a.mp4 |

## Testing Instructions

Give elements borders and border colors and see that it doesn't crash. 

1.

## Checklist

---

Fixes #11019

### DIFF
--- a/packages/story-editor/src/components/colorPicker/basicColorPicker.js
+++ b/packages/story-editor/src/components/colorPicker/basicColorPicker.js
@@ -197,41 +197,40 @@ function BasicColorPicker({
             </StyledText>
           </EyedropperWrapper>
         )}
-        <SavedColors>
-          {allowsSavedColors && (
-            <>
-              <Label id="colorpicker-story-colors-title">
-                {__('Current story', 'web-stories')}
-              </Label>
-              <BasicColorList
-                color={color}
-                colors={storyColors}
-                handleClick={handleClick}
-                isLocal
-                allowsOpacity={allowsOpacity}
-                allowsGradient={allowsGradient}
-                aria-labelledby="colorpicker-story-colors-title"
-                isEditMode={isEditMode}
-                data-testid="saved-story-colors"
-                changedStyle={changedStyle}
-              />
-              <Label id="colorpicker-saved-colors-title">
-                {__('Saved colors', 'web-stories')}
-              </Label>
-              <BasicColorList
-                color={color}
-                colors={savedColors}
-                isGlobal
-                handleClick={handleClick}
-                allowsOpacity={allowsOpacity}
-                allowsGradient={allowsGradient}
-                aria-labelledby="colorpicker-saved-colors-title"
-                isEditMode={isEditMode}
-                changedStyle={changedStyle}
-              />
-            </>
-          )}
-        </SavedColors>
+        {allowsSavedColors && (
+          <SavedColors>
+            <Label id="colorpicker-story-colors-title">
+              {__('Current story', 'web-stories')}
+            </Label>
+            <BasicColorList
+              color={color}
+              colors={storyColors}
+              handleClick={handleClick}
+              isLocal
+              allowsOpacity={allowsOpacity}
+              allowsGradient={allowsGradient}
+              aria-labelledby="colorpicker-story-colors-title"
+              isEditMode={isEditMode}
+              data-testid="saved-story-colors"
+              changedStyle={changedStyle}
+            />
+            <Label id="colorpicker-saved-colors-title">
+              {__('Saved colors', 'web-stories')}
+            </Label>
+            <BasicColorList
+              color={color}
+              colors={savedColors}
+              isGlobal
+              handleClick={handleClick}
+              allowsOpacity={allowsOpacity}
+              allowsGradient={allowsGradient}
+              aria-labelledby="colorpicker-saved-colors-title"
+              isEditMode={isEditMode}
+              changedStyle={changedStyle}
+            />
+          </SavedColors>
+        )}
+
         <DefaultColors>
           <Label id="colorpicker-default-colors-title">
             {__('Default', 'web-stories')}
--- a/packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
+++ b/packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
@@ -137,8 +137,10 @@ function BorderWidthAndColor() {
           value={border.color || BLACK}
           onChange={handleColorChange}
           hasInputs={false}
-          hasEyeDropper={false}
+          hasEyedropper={false}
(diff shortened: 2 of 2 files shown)


packages/story-editor/src/components/colorPicker/basicColorPicker.js
```diff
@@ -197,41 +197,40 @@ function BasicColorPicker({
...
-                allowsOpacity={allowsOpacity}
-                allowsGradient={allowsGradient}
-                aria-labelledby="colorpicker-saved-colors-title"
-                isEditMode={isEditMode}
-                changedStyle={changedStyle}
-              />
-            </>
-          )}
-        </SavedColors>
+        {allowsSavedColors && (
```
BrittanyIRL, line 200: Check for allowing saved colors before rendering the container to avoid an orphaned border when its just default colors showing.

packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
```diff
@@ -137,8 +137,10 @@ function BorderWidthAndColor() {
           value={border.color || BLACK}
           onChange={handleColorChange}
           hasInputs={false}
-          hasEyeDropper={false}
+          hasEyedropper={false}
```
samwhale, line 140: nice catch

packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
```diff
@@ -137,8 +137,10 @@ function BorderWidthAndColor() {
...
           allowsOpacity={canHaveBorderOpacity}
+          pickerHasEyedropper={false} // override shared floating menu Color component TODO https://github.com/GoogleForCreators/web-stories-wp/issues/11024
+          allowsSavedColors={false} // if this changes to true, need to change allowsGradient to false because borders cannot be gradient.
```
samwhale: 💭 if one of `allowGradient` or `allowSavedColors` must be false, then we should add that in here:
```suggestion
          pickerHasEyedropper={false} // override shared floating menu Color component TODO https://github.com/GoogleForCreators/web-stories-wp/issues/11024
          // Borders cannot be gradients.
          // One of `allowSavedColors` or `allowGradient` must be false.
          allowsSavedColors={!allowGradient} 
          allowGradient={allowGradient}
```

packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
```diff
@@ -137,8 +137,10 @@ function BorderWidthAndColor() {
           value={border.color || BLACK}
           onChange={handleColorChange}
           hasInputs={false}
-          hasEyeDropper={false}
+          hasEyedropper={false}
           allowsOpacity={canHaveBorderOpacity}
```
BrookeGraham, line 141: I think this is gonna need to be false as well 🤔 Since I can go to custom colors, it will show gradient as an option then 💥 

Screen Shot 2022-03-22 at 2 50 28 PM
Screen Shot 2022-03-22 at 2 48 11 PM

packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
samwhale: Also - can we add a small unit test here?

samwhale (approved):
This works great! Would be nice if we could add a small unit test to prevent future regressions.

packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
BrookeGraham: Or rather, `allowsGradient` should be false as well

BrookeGraham (changes requested):
gradient

packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
BrittanyIRL: oh yup, you're super right. i completely ignored that button.

BrittanyIRL:
> This works great! Would be nice if we could add a small unit test to prevent future regressions.

I think this is going to need follow up (see #11024 so I'm going to hold on that to make sure design has time to figure out what they want where)

BrittanyIRL:
Thanks for the catch, Brooke!
All fixed
Screen Shot 2022-03-22 at 2 02 03 PM

samwhale:
Oh heck yeah, Brooke! Great catch there.

BrookeGraham:
Only caught it cause I swear I had already fixed it. 🤷

BrookeGraham (approved):
🌈

PR #11026, opened 2022-03-22
Experiments: Turn Floating Menus off by default to properly go through QA

BrittanyIRL:
Turn experiment for floating menu back off because i did not send this through qa before merging it earlier today, got a little trigger happy. Will immediately make another PR to go through QA and then merge. Sorry everyone!


swissspidy:
Why not just let this go through QA as-is? No need to create 2 more PRs just for disabling & re-enabling it again. Felipe can just test the `main` branch.

BrittanyIRL:
> 

If you're cool with that then so am I, this was just against our process. It did go through bug bash this am pretty well.

Issue #10930, opened 2022-03-15
Media Recording (Video/Audio/Image Capture)


swissspidy:
## Summary

Creating custom media today is a big pain due to asset management etc.

We want to explore adding video/audio recording capabilities to the story editor, so that you could record yourself, for example for narration.

## References

Quick prototype:

https://user-images.githubusercontent.com/841956/158474813-e816259e-1a1e-4d73-954c-11222265e183.mov

## Acceptance Criteria

* Video capture with or without audio -> insert as video to story
* Image capture -> capture current frame to take a picture -> add as image to story
* Audio-only capture -> upload as background audio

### Does this epic have any performance impact?

N/A

### Does this epic have telemetry?

Yes

nicholassykes:
@swissspidy I wonder if the scope of this feature might be widened a little to include screen video recording in combination with webcam and audio so that users can provide a show and tell? Like Loom..

swissspidy:
Thanks for the suggestion! We will consider this as an enhancement after building the initial feature. But I‘d imagine it‘s a bit difficult to get a useful screen recording in the size of a story.

PR #11027, opened 2022-03-22
Video Recording: Add initial implementation

### LINKED ISSUE #10958: Video Capture: Initial implementation
## Feature Description

As per #10930 we want to support video/audio capture.

This ticket is for adding the initial plumbing for this feature, based on the existing proof of concept.

We can then build off of this initial implementation.

swissspidy:
## Context

As per #10958 / #10930, we want to add support for webcam/microphone recording.

## Summary

This adds a the foundation for the following work to enable video/audio/image recording in the editor.

This is very rough around the edges. There's absolutely no UX for this so far and some known bugs.

But it gets the job done and unblocks all future work in this area.

## Relevant Technical Choices

* Fixes a bug in `getVideoLength` which returned `NaN` for recorded videos because the video duration was returned as `Infinity`.
* Uses `@wmik/use-media-recorder` instead of implementing all this logic myself. While it's relatively new and marked as a beta, it seems to work quite well. We can revisit this choice later if needed.

## To-do

* Tests
* UX
* Telemetry
* Bugfixes
* ... and lots more

Lots of polish is needed in future PRs. And of course actual UX.

## User-facing changes

Once the experiment is enabled, there is a new recording button to open the dialog for webcam recording.

https://user-images.githubusercontent.com/841956/158474813-e816259e-1a1e-4d73-954c-11222265e183.mov

## Testing Instructions

1.

No

No, something for future PRs

No

## Checklist

---

Fixes #10958

### DIFF
--- a/packages/media/src/getVideoLength.js
+++ b/packages/media/src/getVideoLength.js
@@ -26,7 +26,10 @@ import getVideoLengthDisplay from './getVideoLengthDisplay';
  * @return {Object<{length: number, lengthFormatted: string}>} Video length information.
  */
 function getVideoLength(video) {
-  const duration = !isNaN(video.duration) ? video.duration : 0;
+  // If the element's media doesn't have a known duration—such as for live media streams—the value of duration is +Infinity.
+  // See https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/duration
+  const isNumber = !isNaN(video.duration) && Infinity !== video.duration;
+  const duration = isNumber ? video.duration : 0;
   const length = Math.round(duration);
   const lengthFormatted = getVideoLengthDisplay(length);
   return {
--- a/packages/story-editor/src/components/library/panes/media/local/mediaRecording/modal.js
+++ b/packages/story-editor/src/components/library/panes/media/local/mediaRecording/modal.js
@@ -1,0 +1,192 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * External dependencies
+ */
+import { __ } from '@googleforcreators/i18n';
+import {
+  Text,
+  Toggle,
+  Button,
+  THEME_CONSTANTS,
+  BUTTON_TYPES,
+  BUTTON_SIZES,
+} from '@googleforcreators/design-system';
+import {
+  useState,
+  useRef,
+  useCallback,
+  useMemo,
+  useEffect,
+} from '@googleforcreators/react';
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+import useMediaRecorder from '@wmik/use-media-recorder';
+import { blobToFile } from '@googleforcreators/media';
+import { v4 as uuidv4 } from 'uuid';
+
+/**
+ * Internal dependencies
+ */
+import Dialog from '../../../../../dialog';
+import useUploadWithPreview from '../../../../../canvas/useUploadWithPreview';
+
+const Video = styled.video`
+  width: 100%;
+  height: 100%;
+`;
+
+const VideoWrapper = styled.div`
+  width: 240px;
+  aspect-ratio: 9 / 16;
+  background: #000;
+  margin: 0 auto;
+`;
+
+function Modal({ isOpen, onClose }) {
+  const videoRef = useRef();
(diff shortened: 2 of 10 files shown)


packages/story-editor/src/components/library/panes/media/local/mediaRecording/modal.js
```diff
@@ -0,0 +1,182 @@
...
+} from '@googleforcreators/react';
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+import useMediaRecorder from '@wmik/use-media-recorder';
+import { blobToFile } from '@googleforcreators/media';
+
+/**
+ * Internal dependencies
+ */
+import { v4 as uuidv4 } from 'uuid';
```
spacedmonkey: Move to External dependencies

packages/story-editor/src/components/library/panes/media/local/mediaRecording/modal.js
```diff
@@ -0,0 +1,182 @@
...
+`;
+
+function Modal({ isOpen, onClose }) {
+  const videoRef = useRef();
+  const [file, setFile] = useState(null);
+  const [enableVideo, setEnableVideo] = useState(true);
+  const [enableAudio, setEnableAudio] = useState(true);
+
+  const onStop = useCallback((blob) => {
+    const f = blobToFile(blob, 'test-recording.webm', 'video/webm');
```
spacedmonkey, line 67: Why are we using webm here? We mostly use mp4s in this editor. Also, can be name the file as the timestamp or something.

spacedmonkey:
https://user-images.githubusercontent.com/237508/159749873-34e6b8dd-914f-4103-9a1e-0ab065771473.mov

When I open the dialog for the first time, I don't see a dialog asking me to allow for camera use. Is this a mistake?

spacedmonkey:
Is there a possibility to add an option to select which camera to use. Like in google hangout?
Screenshot 2022-03-23 at 16 38 28

spacedmonkey:
What happens if the user do not have a camera at all? Some error handling or hide the camera button?

packages/story-editor/src/components/library/panes/media/local/mediaRecording/modal.js
swissspidy: For this initial implementation I stuck with the mime type Chrome uses for media recording, which is webm. So it returns a webm, which is why this is a webm.

In future versions this needs to be dynamic.

Determining the file name is something I'll bring up with UX as well.

swissspidy:
> When I open the dialog for the first time, I don't see a dialog asking me to allow for camera use. Is this a mistake?

Right now the dialog opens when you first click on the "Start recording" button.

This is likely to change once we have UX.

> Is there a possibility to add an option to select which camera to use. Like in google hangout?

Yes this is technically possible and something that will be considered for UX, but out of scope for this PR.

> What happens if the user do not have a camera at all? Some error handling or hide the camera button?

This also will be considered for UX, but out of scope for this PR.

spacedmonkey:
Works well when video transcoding is disabled 👍

spacedmonkey:
One thing I found odd, when I open the dialog, record a video and then inserted it. When I opened dialog again, the prevous video I record was still there. I think the dialog should reset after is closed. If I want to see the video I recorded before, I see that in my media library.

packages/media/src/getVideoLength.js
```diff
@@ -26,7 +26,8 @@ import getVideoLengthDisplay from './getVideoLengthDisplay';
  * @return {Object<{length: number, lengthFormatted: string}>} Video length information.
  */
 function getVideoLength(video) {
-  const duration = !isNaN(video.duration) ? video.duration : 0;
+  const isNumber = !isNaN(video.duration) && Infinity !== video.duration;
```
spacedmonkey, line 31: I would add a comment to explain why this is needed.

packages/story-editor/src/components/library/panes/media/local/mediaRecording/modal.js
spacedmonkey: Maybe we should add a todo here, to change this to an different format.

spacedmonkey:
This is GREAT for a first pass. My only feedback, is that on close, the video / dialog to reset. Other than this, it is basically ready to go behind a feature.

merapi (approved):
Good PoC, just one thing that was already mentioned: changing the recording device(s), because for me OBS is first on the list and I need to go into Chrome settings to change that - some people will struggle here.

image

swissspidy:
> I think the dialog should reset after is closed.

Yeah that's a known issue that will need to get addressed. Likely will get fixed automatically once we change the UX flow.

timarney:
As per other comments I think this is great as a POC.

Just to capture first thoughts as a user...

> For the initial screen the "black box" pre giving the camera permission made me think something was broken.   Given the Camera and Mic are already toggled to the on state. Might suggest moving that prompt to when you first press "the camera" button.  

Screen Shot 2022-03-25 at 4 52 30 AM

timarney (approved):

packages/story-editor/src/components/library/panes/media/local/mediaRecording/modal.js
swissspidy: No need for a todo. I'll create tickets for this.

PR #11030, opened 2022-03-22
Floating Menu: Fix Overlapping Section between 'colors' and 'more'.

### LINKED ISSUE #11018: Floating Menus: Overlapping 'Mixed' Input
## Bug Description

Found in bug bash by @samwhale  and confirmed by @BrittanyIRL 

Screen Shot 2022-03-22 at 10 34 49 AM

If a text element has more than 1 color it gets the 'mixed' input, this input overlaps with the 'more' button. If the input swaps to 'mixed' it should expand the allowed input space. This needs to happen for other languages too! 

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

BrittanyIRL:
## Context

Bug identified in bugbash today where if a color input is 'mixed' (2 colors or more in an element) then that color input overlaps the 'more' button. 

## Summary

Checks for multiple color value in color input to override set width, allowing input to expand when needed.

## Relevant Technical Choices

When color value is `MULTIPLE_VALUE` override the asked for width in the input so that 'mixed' (or whatever the translated version of that is) can take up its needed space. 

## User-facing changes

| Broken | Fixed | 
| --- | --- | 
| Screen Shot 2022-03-22 at 2 45 37 PM | Screen Shot 2022-03-22 at 2 46 28 PM | 

## Testing Instructions

This Color section I removed width from is shared by all the color inputs so some regression testing around the color inputs in the various areas of the editor will be necessary. The ones I checked and am aware of before submitting this for review: Floating menu elements, style panel for borders, background color, foreground color, text color combos.

color input change size

1.

## Checklist

---

Fixes #11018

### DIFF
--- a/packages/story-editor/src/components/form/color/color.js
+++ b/packages/story-editor/src/components/form/color/color.js
@@ -35,10 +35,10 @@ import {
 /**
  * Internal dependencies
  */
+import { MULTIPLE_VALUE } from '../../../constants';
 import useEyedropper from '../../eyedropper';
 import Tooltip from '../../tooltip';
 import { focusStyle } from '../../panels/shared';
-import { MULTIPLE_VALUE } from '../../../constants';
 import applyOpacityChange from './applyOpacityChange';
 import OpacityInput from './opacityInput';
 import ColorInput from './colorInput';
--- a/packages/story-editor/src/components/form/color/color.js
+++ b/packages/story-editor/src/components/form/color/color.js
@@ -47,6 +47,8 @@ import { SPACING } from './constants';
 const containerCss = css`
   display: flex;
   align-items: center;
+  width: ${({ width, ignoreWidth }) =>
+    (ignoreWidth && 'inherit') || (width ? `${width}px` : `100%`)};
 `;
 
 const Container = styled.section`
@@ -104,6 +106,7 @@ const Color = forwardRef(function Color(
   },
   ref
 ) {
+  console.log({ label, value });
   const handleOpacityChange = useCallback(
     (newOpacity) => onChange(applyOpacityChange(value, newOpacity)),
     [value, onChange]
@@ -128,8 +131,17 @@ const Color = forwardRef(function Color(
       ? TOOLTIP_PLACEMENT.BOTTOM
       : TOOLTIP_PLACEMENT.BOTTOM_START;
 
+  // Sometimes the value of the set color is "(MULTIPLE)", this is the only time the color comes back a string not an object.
+  // When there's multiple colors the input displays "Mixed" (in english) and takes up a different amount of space.
+  // By checking here to ignore that value based on mixed colors we prevent overlaps in any language too.
+  const ignoreSetWidth = typeof value === 'string';
   return (
-    <Container aria-label={containerLabel} isInDesignMenu={isInDesignMenu}>
+    <Container
+      aria-label={containerLabel}
+      isInDesignMenu={isInDesignMenu}
+      width={width}
+      ignoreWidth={ignoreSetWidth}
+    >
       {hasEyedropper && (
         <Tooltip
           title={tooltip}
--- a/packages/story-editor/src/components/form/color/color.js
+++ b/packages/story-editor/src/components/form/color/color.js
@@ -52,9 +52,12 @@ const containerCss = css`
 const Container = styled.section`
   ${containerCss}
   gap: ${({ isInDesignMenu }) => (isInDesignMenu ? 6 : 8)}px;
-  width: ${({ width, ignoreWidth }) =>
-    (ignoreWidth && 'inherit') || (width ? `${width}px` : `100%`)};
+  width: ${({ width }) => (width ? `${width}px` : 'inherit')};
 `;
+Container.propTypes = {
+  isInDesignMenu: PropTypes.bool,
+  width: PropTypes.string,
+};
 
 const ColorInputsWrapper = styled.div`
   ${containerCss}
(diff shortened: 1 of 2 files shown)


packages/story-editor/src/components/form/color/colorInput.js
```diff
@@ -275,7 +275,10 @@ const ColorInput = forwardRef(function ColorInput(
                 {/* We display Mixed value even without inputs */}
                 {isMixed && (
                   <MixedLabel>
-                    <Text size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL}>
+                    <Text
+                      size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL}
+                      as="span"
```
BrittanyIRL, line 277: no reason for a paragraph tag to live here, just updating to pass as a span.

packages/story-editor/src/components/form/color/color.js
```diff
@@ -52,7 +52,6 @@ const containerCss = css`
 const Container = styled.section`
   ${containerCss}
   gap: ${({ isInDesignMenu }) => (isInDesignMenu ? 6 : 8)}px;
-  width: ${({ width }) => (width ? `${width}px` : `100%`)};
```
BrookeGraham, line 55: If we remove the `width` override here, `width` should no longer be needed as a prop. BUT I added this because we were getting weird sizing on the input on other use cases in the floating menu without it.
Screen Shot 2022-03-22 at 4 03 52 PM

BrookeGraham (dismissed):
This leaves the input as the wrong size in other places.

packages/story-editor/src/components/form/color/color.js
BrittanyIRL: ah drag!

packages/story-editor/src/components/form/color/color.js
```diff
@@ -132,8 +135,17 @@ const Color = forwardRef(function Color(
       ? TOOLTIP_PLACEMENT.BOTTOM
       : TOOLTIP_PLACEMENT.BOTTOM_START;
 
+  // Sometimes the value of the set color is "(MULTIPLE)", this is the only time the color comes back a string not an object.
+  // When there's multiple colors the input displays "Mixed" (in english) and takes up a different amount of space.
+  // By checking here to ignore that value based on mixed colors we prevent overlaps in any language too.
+  const ignoreSetWidth = typeof value === 'string';
```
BrookeGraham: 🤔 could we do the check in the floating menu  > elements > shared > color. Then just set the width appropriately for when value === string

BrookeGraham: 💭 would still have to account for translations though 🤔

swissspidy: We should directly compare with the `MULTIPLE_VALUE` / `MULTIPLE_DISPLAY_VALUE` constant in the editor, not just checking the type.

There's plenty of precedence in the editor for this kind of comparison.

packages/story-editor/src/components/form/color/color.js
```diff
@@ -52,7 +52,8 @@ const containerCss = css`
 const Container = styled.section`
   ${containerCss}
   gap: ${({ isInDesignMenu }) => (isInDesignMenu ? 6 : 8)}px;
-  width: ${({ width }) => (width ? `${width}px` : `100%`)};
+  width: ${({ width, ignoreWidth }) =>
+    (ignoreWidth && 'inherit') || (width ? `${width}px` : `100%`)};
```
samwhale: 🤔  So this feels a little strange to me. The `Color` component takes a `width` prop, but if there's a "multiple" value then the component needs to ignore the passed in width.

I tried removing lines 55 and 56, as well as removing lines 145 (so that `width` isn't passed to the div. Seems to me like everything worked as intended? The color input looks the same in the floating menu and in the style pane.

It seems to me like letting the container decide its own width would work here, but lmk if I'm missing a case.

samwhale: If the `containerCss` has `display: flex` then I think `width: 100%` is redundant since flexbox will stretch the section to fill the parent too?

samwhale (approved):
Works locally for me. Just left a question about implementation

packages/story-editor/src/components/form/color/color.js
maxyinger: no I think you're right. I think the width stuff within the floating menu is a little wonky. I think it's there for calculating the bounds of where we can place it, but then we also explicitly set it? I'm not sure what the purpose of explicitly setting the width is. also feels diffferent to need all the mutation observers we have there to gather this info, but I think I'm lacking some context 🤷

maxyinger (approved):

packages/story-editor/src/components/form/color/color.js
BrookeGraham: >I tried removing lines 55 and 56, as well as removing lines 145 (so that width isn't passed to the div. Seems to me like everything worked as intended? The color input looks the same in the floating menu and in the style pane. 

Without exciplty setting it, the flex extends both containers where color is used. So we end up with extra wide `more` button and and extra wide opacity input. 

But, we could still use `inherit` instead of the extra prop. 

something like 
```
  width: ${({ width }) => (width ? `${width}` : `inherit`)};

```
Where `width` is string that contains units. eg. `width = "128px"`  or `width='inherit"` depend on if it is mixed or not. 

Screen Shot 2022-03-24 at 9 39 23 AM

BrittanyIRL: ok, i think we're good here - removed the `100%` fallback and need for an extra prop.

PR #11033, opened 2022-03-23
Design System:  fix getoffset calculation for Popups

### LINKED ISSUE #10928: Popup's getOffset util calculation incorrect when using dock
## Bug Description

When moving the color picker popup to `dock` on the right of the new `sidebar` component it was discovered in #10883 that the `Popup`'s util function `getOffset` is [calculating the offset incorrectly](https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/design-system/src/components/popup/utils/getOffset.js#L28-L32). Currently, it check's for a truthy value for `dockRect?.left`. However, `dockReck.left` can be and now is a falsey value of zero. This means it unintentionally uses the anchor's left value instead of the dock's left value.  In this sidebar color picker use case, the dock element being passed into the `Popup` is a reference to the entire `inspector` element since we want it docked to right of the entire `sidebar`. Now that this element is on the left side of the screen, the dockRect's left value is zero. This is also the case for style presets.

We can not however simply change the check for dockRect since the it would still make the math for  `leftAligned` (and `rightAligned`) incorrect. When `dock`ing a popup, we shouldn't be adding the width of the anchor, as the anchor is no longer relevant to positioning. It should be noted however, that in the use case for tooltips, `rightAligned` would need to account for anchor width. 

It should also be noted that `leftAligned` and `rightAligned` mean something different when talking about where to `dock`, vs. `placement` of the popup -> tooltip. In addition to RTL mode. 

Currently, the  `Popup` component has 17 props, with another one being added to differentiate "normal popups" from the tooltip's implementation. While `Tooltip` _only_ sends down 8 to Popup, and yet has 16 of it's own props, trying to account for all use cases is getting problematic. Regressions are hard to manage since there are vastly different ui needs between Popups, Tooltips, and Dropdown. With so much complexity and the use cases diverging it may be time to split apart `Tooltip` from `Popup` and perhaps `Dropdown`as well.  

## Expected Behaviour

We expect the Popup to render in relation to the dock element whenever a dock ref is present. We also expect the offset calculations be based on the referenced element and that those calculations take into account whether we are docked on the right of an element or on the left. 
 
## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

### LINKED ISSUE #11020: Popup Sidebar Regression: Overlap
## Bug Description

Seems a regression occurred somewhere in the sidebar work because these popups _were_ addressed. Now they are overlapping again. 

Text color is fine but now border and bg are overlapping. 

| Border | Text color | Background |
| --- | --- | --- | 
| Screen Shot 2022-03-22 at 10 52 21 AM | Screen Shot 2022-03-22 at 10 52 28 AM | Screen Shot 2022-03-22 at 10 52 35 AM | 

## Additional Context

PR that originally fixed: https://github.com/GoogleForCreators/web-stories-wp/pull/10883 

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

BrookeGraham:
## Context

The background color picker popup is not respecting offset spacing in relation to it's docked element. 

## Summary
The background color picker popup was getting incorrect values for the inspector's `dock` `getXOffset`. Since those calculations were incorrect anyway, see #10928  I decided to fix the calculation instead of continuing to override spacing. 
If a Popup has an element it is docked against, the offset should be based on that dock regardless of it's values.

## Relevant Technical Choices
Updated the logic for calculating offset via `dock` element
Reset the spacing for color input and saves styles popups to only reflect expected margin from the docked element

### Before
| Border | Text color | Background |
| --- | --- | --- | 
| Screen Shot 2022-03-22 at 10 52 21 AM | Screen Shot 2022-03-22 at 10 52 28 AM | Screen Shot 2022-03-22 at 10 52 35 AM | 

### After 
| Border | Text color | Background |
| --- | --- | --- | 
|border|Screen Shot 2022-03-22 at 7 15 14 PM|background|

## Testing Instructions

1. Open each color picker; check if it is rendered in the correct spot. 
- background color
- shape color
- text color
- border color shape
- image filters
- image borders
- floating menus
2. Check Saved Styles popup is docked to the sidebar as well
3. Check for regressions on all tooltips in editor and on dashboard (Sorry it's a lot) 
4. Don't forget RTL

no

no

no

## Checklist

---

Fixes #11020 
Partially Fixes #10928

### DIFF
--- a/packages/design-system/src/components/popup/utils/getOffset.js
+++ b/packages/design-system/src/components/popup/utils/getOffset.js
@@ -25,7 +25,6 @@ export function getXOffset(
   dockRect,
   isRTL
 ) {
-  // TODO: https://github.com/GoogleForCreators/web-stories-wp/issues/10928
   // doctRect.left can have a valid value of zero, if dockRect exists, it takes precedence.
   const leftAligned = (dockRect ? dockRect.left : anchorRect.left) - spacing;
   const rightAligned = (dockRect ? dockRect.right : anchorRect.right) + spacing;
--- a/packages/design-system/src/components/popup/utils/getOffset.js
+++ b/packages/design-system/src/components/popup/utils/getOffset.js
@@ -30,7 +30,7 @@ export function getXOffset(
   const leftAligned = (dockRect ? dockRect.left : anchorRect.left) - spacing;
   const rightAligned = (dockRect ? dockRect.right : anchorRect.right) + spacing;
   const centerAligned = dockRect
-    ? dockRect.left
+    ? dockRect.left + dockRect.width / 2
     : anchorRect.left + anchorRect.width / 2;
 
   switch (placement) {
--- a/packages/design-system/src/components/popup/utils/getOffset.js
+++ b/packages/design-system/src/components/popup/utils/getOffset.js
@@ -26,14 +26,12 @@ export function getXOffset(
   isRTL
 ) {
   // TODO: https://github.com/GoogleForCreators/web-stories-wp/issues/10928
-  // doctRect?.left can have a valid value of zero, however in this case, the falsey
-  // value gets ignored and we end up using anchor.left. Additionally, if using `dock`
-  // anchor.width shouldn't need be added
-  const leftAligned = (dockRect?.left || anchorRect.left) - spacing;
-  const rightAligned =
-    (dockRect?.left || anchorRect.left) + anchorRect.width + spacing;
-  const centerAligned =
-    (dockRect?.left || anchorRect.left) + anchorRect.width / 2;
+  // doctRect.left can have a valid value of zero, if dockRect exists, it takes precedence.
+  const leftAligned = (dockRect ? dockRect.left : anchorRect.left) - spacing;
+  const rightAligned = (dockRect ? dockRect.right : anchorRect.right) + spacing;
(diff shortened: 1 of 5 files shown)


packages/design-system/src/components/popup/utils/getOffset.js
```diff
@@ -25,15 +25,12 @@ export function getXOffset(
...
-  const rightAligned =
-    (dockRect?.left || anchorRect.left) + anchorRect.width + spacing;
-  const centerAligned =
-    (dockRect?.left || anchorRect.left) + anchorRect.width / 2;
+  // doctRect.left can have a valid value of zero, if dockRect exists, it takes precedence.
+  const leftAligned = (dockRect ? dockRect.left : anchorRect.left) - spacing;
+  const rightAligned = (dockRect ? dockRect.right : anchorRect.right) + spacing;
+  const centerAligned = dockRect
+    ? dockRect.left + dockRect.width / 2
+    : anchorRect.left + anchorRect.width / 2;
```
samwhale, line 33: 🥳

samwhale (approved):
Excelente

BrittanyIRL (approved):
🦈

maxyinger (approved):
💪

PR #11067, opened 2022-03-24
Background Audio: Do not always loop audio in the editor

### LINKED ISSUE #11046: Background Audio: Audio always loops when playing in the editor
## Bug Description

Originally reported by @BrookeGraham 
 in the design panel, if I uncheck the loop box, then hit the play button, the audio still loops until I hit the stop button. The loop checkbox does not impact the preview functionality as one would expect

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

spacedmonkey:
## Summary

Ensure that the looping checkbox is respected by audio player. Ensure that play / pause button is reset when the audio player is finished. 

This change, also improves browser compatibility, as browsers like chrome allow for audio to be paused via media control or by media buttons on the keyboard.  

## User-facing changes

https://user-images.githubusercontent.com/237508/159926210-88ee5f41-4cbf-4f31-8cb3-d9d0bee499c5.mov

## Testing Instructions

1. Create a story
2. Add background audio to a page.
3. Set looping to disabled.
4. Play audio. 
5. When audio is finished, it should not loop and play / pause button should reset. 

#### Media controls
1. Create a story
2. Add background audio to a page.
3. Set looping to disabled.
4. Play audio. 
5. Use media controls in Chrome to pause audio. 
6. Play / pause button should reset. 

## Checklist

---

Fixes #11046


packages/story-editor/src/components/audioPlayer/index.js
```diff
@@ -115,41 +144,21 @@ function AudioPlayer({ title, src, mimeType, tracks = [], audioId }) {
...
-              variant={BUTTON_VARIANTS.SQUARE}
-              aria-label={__('Play', 'web-stories')}
-              onClick={handlePlayPause}
-            >
-              <Icons.PlayFilled />
-            </StyledButton>
-          </Tooltip>
-        )}
-      </div>
+      <Tooltip
```
spacedmonkey, line 147: This part is not needed, but it a nice refactor, to stop lots of repeated code.

packages/story-editor/src/components/audioPlayer/index.js
```diff
@@ -86,19 +91,43 @@ function AudioPlayer({ title, src, mimeType, tracks = [], audioId }) {
 
     if (isPlaying) {
       player.pause();
-      setIsPlaying(false);
     } else {
-      player
-        .play()
-        .then(() => setIsPlaying(true))
-        .catch(() => {});
+      player.play();
```
swissspidy: `.play()` returns a Promise, which means it can get rejected. Which is why we'd need the `catch(() => {})` here to avoid throwing errors.

See https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play

BrookeGraham (approved):
Looks great. Sounds even better 🔈

timarney:
Nice -- works well

timarney (approved):

PR #11069, opened 2022-03-24
Video: Improved handling of video play button

### LINKED ISSUE #11068: Video: Improved handling of video play button
## Feature Description

Discovered while working on https://github.com/GoogleForCreators/web-stories-wp/issues/11046. 

Videos can be play / paused using media controls in the browser. ( See video ). But the play / pause button does not respect that. 

See this video, where I am using media keys on my laptop's keyboard. 

https://user-images.githubusercontent.com/237508/159931296-4f769a88-52dc-41d9-9bd7-eeb22afdeb49.mov

spacedmonkey:
## Summary

Videos can be play / paused using media controls in the browser. ( See video ). But the play / pause button does not respect that.

## Relevant Technical Choices

This PR, make code simplier, but treating the video tags, play / pause events as a point of trust on if the video is playing or not. Removing a lot of logic. 

## User-facing changes

https://user-images.githubusercontent.com/237508/160124232-76d50e02-ec2f-4b43-8529-f2be8eadb628.mov

## Testing Instructions

1. Create story
2. Insert a video ( video must have audio ).
3. Start video playing. 
4. Use chrome media controls / media keys on keyboard. 

## Checklist

---

Fixes #11068

### DIFF
--- a/packages/element-library/src/video/playPauseButton.js
+++ b/packages/element-library/src/video/playPauseButton.js
@@ -223,7 +223,7 @@ function PlayPauseButton({
       videoNode.pause();
       setShowControls(true);
     } else {
-      videoNode.play();
+      videoNode.play().catch(() => {});
     }
   };
   useKeyDownEffect(
--- a/packages/element-library/src/video/playPauseButton.js
+++ b/packages/element-library/src/video/playPauseButton.js
@@ -183,7 +183,7 @@ function PlayPauseButton({
       videoNode.removeEventListener('play', onVideoPlay);
       videoNode.removeEventListener('pause', onVideoPause);
     };
-  }, [shouldResetOnEnd, getVideoNode, id]);
+  }, [getVideoNode, id]);
 
   const checkShowControls = useDebouncedCallback(() => {
     if (!isPlayAbove) {
--- a/packages/element-library/src/video/playPauseButton.js
+++ b/packages/element-library/src/video/playPauseButton.js
@@ -159,20 +159,29 @@ function PlayPauseButton({
       return undefined;
     }
 
-    const onVideoPlay = () => setIsPlaying(true);
-    const onVideoPause = () => setIsPlaying(false);
     const onVideoEnd = () => {
       videoNode.currentTime = 0;
       setIsPlaying(false);
     };
 
+    videoNode.addEventListener('ended', onVideoEnd);
+    return () => videoNode.removeEventListener('ended', onVideoEnd);
+  }, [shouldResetOnEnd, getVideoNode, id]);
+
+  useEffect(() => {
+    const videoNode = getVideoNode();
+    if (!videoNode) {
+      return undefined;
+    }
+
+    const onVideoPlay = () => setIsPlaying(true);
+    const onVideoPause = () => setIsPlaying(false);
+
     videoNode.addEventListener('play', onVideoPlay);
     videoNode.addEventListener('pause', onVideoPause);
-    videoNode.addEventListener('ended', onVideoEnd);
     return () => {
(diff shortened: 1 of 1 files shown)


packages/element-library/src/video/playPauseButton.js
```diff
@@ -170,12 +162,29 @@ function PlayPauseButton({
...
+    const onVideoPlay = () => setIsPlaying(true);
+    const onVideoPause = () => setIsPlaying(false);
+
+    videoNode.addEventListener('play', onVideoPlay);
+    videoNode.addEventListener('pause', onVideoPause);
+    return () => {
+      videoNode.removeEventListener('play', onVideoPlay);
+      videoNode.removeEventListener('pause', onVideoPause);
+    };
+  }, [shouldResetOnEnd, getVideoNode, id]);
```
spacedmonkey:
```suggestion
  }, [getVideoNode, id]);
```

packages/element-library/src/video/playPauseButton.js
```diff
@@ -216,13 +221,9 @@ function PlayPauseButton({
...
     if (isPlaying) {
       videoNode.pause();
-      setIsPlaying(false);
       setShowControls(true);
     } else {
-      videoNode
-        .play()
-        .then(() => setIsPlaying(true))
-        .catch(() => {});
+      videoNode.play();
```
swissspidy: Same here as in https://github.com/GoogleForCreators/web-stories-wp/pull/11067#discussion_r834356976

`.play()` returns a Promise, which means it can get rejected. Which is why we'd need the `catch(() => {})` here to avoid throwing errors.

See https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play

spacedmonkey: Good catch!

Worth noting there are a couple other place without a catch. 

https://github.com/GoogleForCreators/web-stories-wp/blob/e00e50da7a02d956b993f4038ca1c15d46eda3ef/packages/story-editor/src/components/videoTrim/useVideoNode.js#L57

https://github.com/GoogleForCreators/web-stories-wp/blob/e00e50da7a02d956b993f4038ca1c15d46eda3ef/packages/story-editor/src/components/videoTrim/useVideoNode.js#L78

swissspidy: If you could create a ticket for that & assign to Prometheus pod, that would be great!

spacedmonkey: Ticket created https://github.com/GoogleForCreators/web-stories-wp/issues/11085

spacedmonkey:
This is ready for review.

timarney (approved):

swissspidy:
@spacedmonkey Works fine for regular videos, but is this supposed to work for muted videos as well? It doesn't in my testing and I guess that kinda makes sense. Would be good to clarify that in the QA instructions.

spacedmonkey:
> Works fine for regular videos, but is this supposed to work for muted videos as well? It doesn't in my testing and I guess that kinda makes sense. Would be good to clarify that in the QA instructions.

It is a good point. Media control in chrome only show if sound is playing. But media keys would work.

swissspidy:
Media keys did not work for muted videos in my testing (macOS Monterey, Chrome v100)

spacedmonkey:
> Media keys did not work for muted videos in my testing (macOS Monterey, Chrome v100)

You are right. It is just related to videos with sound. I thought I saw it working but I was wrong. But it does work with audio. So that is a win.

swissspidy (approved):

PR #11075, opened 2022-03-24
Floating Menu: Fix Border Color Updates

### LINKED ISSUE #11024: Floating Menu: Cannot select border color with eyedropper

BrittanyIRL:
## Context

Found in bugbash 1.19 - using the eyedropper from the floating menu to update a border color wasn't working. Immediate fix was to remove the eyedropper (#11025) to prevent crashes while getting to the bottom of the inconsistent updates here.

## Summary

This is actually a pretty interesting new case (thanks @littlemilkstudio for spending an hour talking through this and getting to the bottom of it!). Before the floating menus the color picker stays mounted when the eyedropper's in use. Floating menu introduces a new flow where the color picker unmounts. If it's unmounted, the debounced change isn't going to ever get found to actually tell the element to update the color. Since border is the only color picker variant in the floating menu that has the color picker _in_ the popup (as opposed to in the floating menu itself) it's only an issue there. 

## Relevant Technical Choices

`basic` color picker variant: 
- Track if the color picker is mounted with a ref. When it's false and color is changed don't use the debounced change, use the passed in change. The only time the debounce is really needed is on drag interactions. 

`custom` color picker variant: 
- So it's receiving the change handler from the basic color picker, that's why the debounce is actually there - since custom has the dragging on the gradient and slide. That said, it only ever processes that change when a `generatedColor` is true on that existing effect (line 77). This is a problem if the eyedropper's used because (spoiler) it's unmounted now! 
- Move the onChange to a callback of its own, keep existing `updateCurrentColor` to trigger the useColor hook and update as needed and then check if it's unmounted - if it is (and the only way this is true is with floating menu + eyedropper) and there's an rgb key to the value then go ahead and trigger that passed in `handleColorChange` with a manually set color object to propagate changes. 

In the border color floating menu: 
- Fix typo on `hasEyedropper`
- borders can't have gradients 
- borders don't show saved colors (for continuity with sidepanel border color options) 
- added `changedStyle`  for consistency with other border color instance and so that the changedStyle doesn't default to 'background' here. It largely seems to work without it but I figure mirroring the existing border color is good because this whole thing is rather complicated.

## To-do

- Tests probably would be nice, but given the time crunch for release we'll see. 

## User-facing changes

- Can use eyedropper to pick out the border color you want via floating menu. 

https://user-images.githubusercontent.com/10720454/160026202-5df8b424-fbb7-4b80-9cff-7f052bdb6609.mp4

## Testing Instructions

- Please test the color pickers. Make sure that the border color picker in the sidepanel and the floating menu mirror each other. You should be able to set a border color via the eyedropper, color circles, custom gradient, set opacity, type in a hex. 
- Make sure color picker for foreground is fine.
- Make sure color picker from sidepanel also works as expected.

1.

## Checklist

---

Fixes #11024


packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
```diff
@@ -135,8 +136,11 @@ function BorderWidthAndColor() {
               value={border.color || BLACK}
               onChange={handleColorChange}
               hasInputs={false}
-              hasEyeDropper={false}
               allowsOpacity={canHaveBorderOpacity}
+              hasEyedropper={false}
+              allowsGradient={false}
+              allowsSavedColors={false}
+              changedStyle={'border-color'}
```
samwhale: Don't need string interpolation here
```suggestion
              changedStyle="border-color"
```

packages/story-editor/src/components/colorPicker/colorPicker.js
```diff
@@ -118,9 +119,25 @@ function ColorPicker({
...
-      onDebouncedChange(newColor);
+      if (isMounted.current) {
+        onDebouncedChange(newColor);
+      } else {
+        onChange(newColor);
+      }
+
       selectedElementIds.forEach((id) => {
         pushTransform(id, {
           color: newColor,
```
samwhale, line 143: 😕 I'm confused about why the `onChange` is debounced but the transforms are immediately pushed onto the elements (when the picker is mounted)

This was before these changes, but it seems a little strange, no?

packages/story-editor/src/components/colorPicker/customColorPicker.js
```diff
@@ -80,6 +88,21 @@ function CustomColorPicker({
...
+      // If using the eyedropper in floating menu, the popup unmounts
+      // so the generatedColor won't ever be updated
+      // check for unmount and if so, we know this change event
+      // if from the eyedropper so we can just grab the rgb and
+      // trigger the rest of the change for the element.
+      if (!isMounted.current && e?.rgb) {
+        handleColorChange({ color: e.rgb });
+      }
+    },
+    [updateCurrentColor]
```
samwhale: This is missing a dependency
```suggestion
    [handleColorChange, updateCurrentColor]
```

packages/story-editor/src/components/colorPicker/customColorPicker.js
```diff
@@ -115,7 +138,7 @@ function CustomColorPicker({
       )}
       <CurrentColorPicker
         color={currentColor}
-        onChange={updateCurrentColor}
+        onChange={handleColorPickerChange}
         showOpacity={allowsOpacity}
         hasEyedropper={hasEyedropper}
```
samwhale, line 143: This change works well. Considering the time pressure and the complexity of de-coupling the eye dropper and color picker, I think this is good to go.

---

💭  I think the source of the problem here is that the color picker has any awareness of the eye dropper at all. The color picker only needs to know if it should render the eye dropper button. However, it is also given a prop that determines if the eye dropper is active. Due to that, this component has become very complicated.

The color picker code smells funky to me because the eyedropper lives deep in the color picker. The eyedropper should exist as a separate component to the color picker. However, the existence of the eye dropper in the color picker necessitates the color picker being mounted to run a debounced change when changing colors.
https://github.com/GoogleForCreators/web-stories-wp/blob/3cc4ee5c5c02eb9dc2e652660a5d7a731aa3babe/packages/story-editor/src/components/colorPicker/currentColorPicker.js#L158-L160

Due to that, the color picker needs to know if the eye dropper is being used. `isEyeDropperActive` is used to determine, on focus out, if the color picker should close. Through this - the color picker is given awareness of an external component's existence. The color picker shouldn't make decisions about its own visibility - the parent should be in charge of that. In this case, the `ColorInput`.
https://github.com/GoogleForCreators/web-stories-wp/blob/3cc4ee5c5c02eb9dc2e652660a5d7a731aa3babe/packages/story-editor/src/components/form/color/colorInput.js#L288-L310

The visibility of the popup should be decided in `onClose`, which is defined in the parent. However, the visibility is decided deeper in the actual color picker. I think that decision should be lifted.
https://github.com/GoogleForCreators/web-stories-wp/blob/3cc4ee5c5c02eb9dc2e652660a5d7a731aa3babe/packages/story-editor/src/components/colorPicker/colorPicker.js#L135-L142

---

None of this was a problem before because the color picker never unmounted when using the eye dropper. However, this new use case highlights the consequences of creating the eye dropper in the color picker popup.

samwhale (approved):
Thanks for adding the helpful comments

packages/story-editor/src/components/colorPicker/colorPicker.js
BrittanyIRL: It's because of the custom color picker where you can drag - because this sets up the on change for allllll of the variants of the color picker.

packages/story-editor/src/components/colorPicker/customColorPicker.js
BrittanyIRL: I really need to get my linter to start working again, thanks 🙈

packages/story-editor/src/components/colorPicker/colorPicker.js
maxyinger: I also think the transform stuff doesn't cause a large state update, whereas the `onChange` does.

maxyinger (approved):

packages/story-editor/src/components/colorPicker/customColorPicker.js
```diff
@@ -80,6 +88,21 @@ function CustomColorPicker({
     }
   }, [color, load]);
 
+  const handleColorPickerChange = useCallback(
+    (e) => {
+      updateCurrentColor(e);
+      // If using the eyedropper in floating menu, the popup unmounts
+      // so the generatedColor won't ever be updated
+      // check for unmount and if so, we know this change event
+      // if from the eyedropper so we can just grab the rgb and
```
swissspidy:
```suggestion
      // is from the eyedropper so we can just grab the rgb and
```

PR #11099, opened 2022-03-28
Fix issues related to selecting all text

### LINKED ISSUE #11034: Unable to delete a text selection - Safari
## Bug Description

Highlighting all text with `Cmd + a` and pressing `Backspace` does not delete the text. It only deletes one letter at a time

image3

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

### LINKED ISSUE #11061: Selecting all text and pasting does not replace text
## Bug Description

It does not seem to happen consistently, but most of the time when I select all of the text in a text element using Cmd+A and then paste some new text, the existing one is not replaced.

## Expected Behaviour

Cmd+A and pasting should replace text

## Steps to Reproduce

1. Insert text from preset
2. Select all
3. Paste some text from the clipboard
4. See that existing text was not replaced, just appended

## Screenshots

Towards the end of the video you see it better:

https://user-images.githubusercontent.com/841956/159796985-258fccba-823e-49f5-9ad7-c0dbdfc5f69c.mov

## Additional Context

- Plugin Version: 1.19.0-alpha
- WordPress Version: 5.9
- Operating System: macOS Monterey
- Browser: Chrome

merapi:
## Summary

There are issues with selecting all text that were reported in #11034 and #11061.

There might be a different/better solution (because CMD+A generally works, it's more a problem with selection/focus), but this one is easy and harmless and most importantly - fixes all the issues with selecting all text.

I tell CMD+A to explicitly select all text in the text element.

## Testing Instructions

1. Add a text element
2. Click on it once (in the middle of the text)
3. Select all text by CMD+A
4. Paste or delete text
5. It should work as expected (and was broken before)

## Checklist

---

Fixes #11034
Fixes #11061

### DIFF
--- a/packages/story-editor/src/karma/element-library/text/edit.karma.js
+++ b/packages/story-editor/src/karma/element-library/text/edit.karma.js
@@ -151,9 +151,9 @@ describe('TextEdit integration', () => {
       });
 
       it('should select all text and delete it', async () => {
-        // Testing this on the BROKEN version
         await fixture.events.mouse.clickOn(frame, 30, 5); // enter the edit mode by clicking
 
+        // Needed because of https://github.com/puppeteer/puppeteer/issues/1313
         if (navigator.userAgentData.platform === 'macOS') {
           document.execCommand('selectAll'); // not the same as mod+a, but does work on macOS
         } else {
--- a/packages/story-editor/src/karma/element-library/text/edit.karma.js
+++ b/packages/story-editor/src/karma/element-library/text/edit.karma.js
@@ -154,11 +154,27 @@ describe('TextEdit integration', () => {
       fit('should select all text and delete it', async () => {
         // Testing this on the BROKEN version
         await fixture.events.mouse.clickOn(frame, 30, 5); // enter the edit mode by clicking
-        await fixture.events.keyboard.shortcut('mod+a'); // doesn't work on macOS, maybe will work on Ubuntu
-        //document.execCommand("selectAll"); // selects all, but it's not the same as MOD+A, so we can't use it
-        await fixture.events.keyboard.type('BAR');
-        // result: BAR, expected: Lorem...BAR...ipsum
-        expect(frame.querySelector('p').innerHTML).not.toEqual('BAR');
+
+        if (navigator.userAgentData.platform === 'macOS') {
+          document.execCommand('selectAll'); // not the same as mod+a
+        } else {
+          await fixture.events.keyboard.shortcut('mod+a'); // doesn't work on macOS, works on Ubuntu
+        }
+        const text = '461';
+        await fixture.events.keyboard.type(text);
+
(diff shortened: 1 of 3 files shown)


spacedmonkey:
Can you provide a bit more context in the PR description please.

merapi:
> Can you provide a bit more context in the PR description please.

I've updated the default title and added some details about the solution, more details about the issue itself can be found in the linked tickets.

swissspidy:
Can we add Karma tests for this change?

merapi:
> Can we add Karma tests for this change?

It's tricky, I believe you saw that: https://github.com/puppeteer/puppeteer/issues/1313
I didn't test it on Ubuntu, maybe we can skip the test locally on macOS, WDYT?

```js
// Testing this on the BROKEN version
await fixture.events.mouse.clickOn(frame, 30, 5); // enter the edit mode by clicking
await fixture.events.keyboard.shortcut('mod+a'); // doesn't work on macOS, maybe will work on Ubuntu
document.execCommand("selectAll"); // selects all, but it's not the same as MOD+A, so we can't use it
await fixture.events.keyboard.type('BAR');
// result: BAR, expected: Lorem...BAR...ipsum
```

swissspidy:
> I didn't test it on Ubuntu, maybe we can skip the test locally on macOS, WDYT?

That could work, yeah. Better than no tests at all.

merapi:
@swissspidy We now have a test for it, PTAL at those `navigator` and `document` lines, are they ok or want to put them somewhere in the Fixture?

packages/story-editor/src/karma/element-library/text/edit.karma.js
```diff
@@ -149,6 +149,32 @@ describe('TextEdit integration', () => {
...
+
+      it('should select all text and delete it', async () => {
+        // Testing this on the BROKEN version
+        await fixture.events.mouse.clickOn(frame, 30, 5); // enter the edit mode by clicking
+
+        if (navigator.userAgentData.platform === 'macOS') {
+          document.execCommand('selectAll'); // not the same as mod+a, but does work on macOS
+        } else {
+          await fixture.events.keyboard.shortcut('mod+a'); // doesn't work on macOS, works on Ubuntu
+        }
```
swissspidy, line 161:
```suggestion
        // Needed because of https://github.com/puppeteer/puppeteer/issues/1313
        if (navigator.userAgentData.platform === 'macOS') {
          document.execCommand('selectAll'); // not the same as mod+a, but does work on macOS
        } else {
          await fixture.events.keyboard.shortcut('mod+a'); // doesn't work on macOS, works on Ubuntu
        }
```

swissspidy:
I think it's OK like that 👍

swissspidy:
I think it's OK like that 👍

swissspidy (approved):

PR #11114, opened 2022-03-29
Floating Menu: Fix "Saved Color" (gradient) as border color causes crash (Again)

### LINKED ISSUE #11019: Floating Menu: Apply "Saved Color" as border color causes crash

BrittanyIRL:
## Context

Re-adding #11025 which had changes that were overwritten in a rebase that are necessary to avoid crashing the editor. 

## Summary

See that these are the same changes as previously added: https://github.com/GoogleForCreators/web-stories-wp/pull/11025/files#diff-3e101509bd8aebce46f213f43dc1e13358fabc6f9cccef75d7d4cce88f8de35c

## User-facing changes

Adding border colors via floating menu no longer crashes editor

## Testing Instructions

Give elements borders and border colors and see that it doesn't crash.

1.

## Checklist

---

Fixes #11019

### DIFF
--- a/packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
+++ b/packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
@@ -134,8 +134,11 @@ function BorderWidthAndColor() {
             value={border.color || BLACK}
             onChange={handleColorChange}
             hasInputs={false}
-            hasEyeDropper={false}
+            hasEyedropper={false}
             allowsOpacity={canHaveBorderOpacity}
+            allowsGradient={false}
+            pickerHasEyedropper={false} // override shared floating menu Color component TODO https://github.com/GoogleForCreators/web-stories-wp/issues/11024
+            allowsSavedColors={false}
           />
         </>
       )}
--- a/packages/story-editor/src/components/form/color/color.js
+++ b/packages/story-editor/src/components/form/color/color.js
@@ -56,7 +56,7 @@ const Container = styled.section`
 `;
 Container.propTypes = {
   isInDesignMenu: PropTypes.bool,
-  width: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  width: PropTypes.number,
 };
 
 const ColorInputsWrapper = styled.div`
@@ -141,7 +141,7 @@ const Color = forwardRef(function Color(
     <Container
       aria-label={containerLabel}
       isInDesignMenu={isInDesignMenu}
-      width={!ignoreSetWidth && width ? width : ''}
+      width={!ignoreSetWidth && width ? width : null}
     >
       {hasEyedropper && (
         <Tooltip
--- a/packages/story-editor/src/components/form/color/color.js
+++ b/packages/story-editor/src/components/form/color/color.js
@@ -56,7 +56,7 @@ const Container = styled.section`
 `;
 Container.propTypes = {
   isInDesignMenu: PropTypes.bool,
-  width: PropTypes.string,
+  width: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
 };
 
 const ColorInputsWrapper = styled.div`
@@ -141,7 +141,7 @@ const Color = forwardRef(function Color(
     <Container
       aria-label={containerLabel}
       isInDesignMenu={isInDesignMenu}
-      width={!ignoreSetWidth && width}
+      width={!ignoreSetWidth && width ? width : ''}
     >
       {hasEyedropper && (
         <Tooltip
--- a/packages/story-editor/src/components/form/color/test/color.js
+++ b/packages/story-editor/src/components/form/color/test/color.js
@@ -24,10 +24,11 @@ import { renderWithTheme } from '@googleforcreators/test-utils';
 /**
  * Internal dependencies
  */
-import Color from '../color';
-import applyOpacityChange from '../applyOpacityChange';
 import { ConfigProvider } from '../../../../app/config';
 import getDefaultConfig from '../../../../getDefaultConfig';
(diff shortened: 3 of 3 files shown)


packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
```diff
@@ -134,8 +134,11 @@ function BorderWidthAndColor() {
             value={border.color || BLACK}
             onChange={handleColorChange}
             hasInputs={false}
-            hasEyeDropper={false}
+            hasEyedropper={false}
             allowsOpacity={canHaveBorderOpacity}
+            allowsGradient={false}
+            pickerHasEyedropper={false} // override shared floating menu Color component TODO https://github.com/GoogleForCreators/web-stories-wp/issues/11024
+            allowsSavedColors={false}
```
BrittanyIRL, line 141: same set of updates as found here: https://github.com/GoogleForCreators/web-stories-wp/pull/11025/files#diff-3e101509bd8aebce46f213f43dc1e13358fabc6f9cccef75d7d4cce88f8de35c

packages/story-editor/src/components/form/color/color.js
```diff
@@ -56,7 +56,7 @@ const Container = styled.section`
 `;
 Container.propTypes = {
   isInDesignMenu: PropTypes.bool,
-  width: PropTypes.string,
+  width: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
BrittanyIRL: this is just silencing a console warning pre release while i'm here.

swissspidy: Is this because we're now passing an empty string to `width`? Can't we pass `null` instead?

Edit: or is it because of `MIXED_VALUE`?

swissspidy:
Can we add tests to prevent this in the future?

packages/story-editor/src/components/form/color/color.js
merapi: Not sure why it was `string` before, but changing that to `width: PropTypes.number,` seems to work for all cases.

packages/story-editor/src/components/form/color/color.js
```diff
@@ -141,7 +141,7 @@ const Color = forwardRef(function Color(
     <Container
       aria-label={containerLabel}
       isInDesignMenu={isInDesignMenu}
-      width={!ignoreSetWidth && width}
+      width={!ignoreSetWidth && width ? width : ''}
```
merapi: With `width: PropTypes.number,` this is no longer needed.

merapi (approved):
Fixes the crash ✅ 
Karma test won't hurt I guess.

Nit: The color picker should be smaller?

image

image

packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
swissspidy: Also same as #11075 it seems?

packages/story-editor/src/components/form/color/color.js
BrittanyIRL: linter was complaining about it becoming a boolean, happy to return it back to how it was.

BrittanyIRL:
> Nit: The color picker should be smaller?
> 
> image
> 
> image

I don't really see how the color picker size is impacted by this PR - would you make a new ticket for it and reference the intended size?

packages/story-editor/src/components/form/color/color.js
swissspidy: Or just return null instead of an empty string?

BrittanyIRL:
```
  ["Warning: Received `%s` for a non-boolean attribute `%s`.·
    If you want to write it to the DOM, pass a string instead: %s=\"%s\" or %s={value.toString()}.·
    If you used to conditionally omit it with %s={condition && value}, pass %s={condition ? value : undefined} instead.%s", "false", "width", "width", "false", "width", "width", "width", "
```

packages/story-editor/src/components/floatingMenu/elements/borderWidthAndColor.js
BrittanyIRL: that one actually reenables the color picker, i noticed that my hotfix to avoid crashing was gone when i rebased that PR. Felipe identified an (unrelated) issue where the eyerdropper in the sidepanel outside of the color picker is out of sync with updating the canvas which took that PR back into in progress, so it's not going to make the release. Whereas this is just eliminating saved colors and eyedropper from floating menu color picker in the mean time.

packages/story-editor/src/components/form/color/color.js
swissspidy: Is the width only needed for CSS or also in the DOM?
If the former then we might want to use transient props.

maxyinger (approved):

merapi:
> I don't really see how the color picker size is impacted by this PR - would you make a new ticket for it and reference the intended size?

| Before this PR | After this PR |
|----------------|---------------|
| image | image |

image

You have just hidden those buttons? 🤔

Issue #10515, opened 2022-02-08
Media: Add support for hotlinking VTT files


bmattb:
## Feature Description

As a publisher who leverages remote asset hosting we depend on “insert by URL” functionality to support using videos in our Web Stories. As such, it would be great for the plugin to also support the use of remotely hosted VTT files by being able to supply a URL for these as well in the Caption VTT file selector.

swissspidy:
Sounds like a reasonable feature request. Should be easy to implement. Just needs design.

For reference, right now we have a simple button that opens a media picker:
![Screenshot 2022-02-09 at 14 27 39](https://user-images.githubusercontent.com/841956/153211026-4f98374a-3d9c-4cd9-bd93-6707e79b988d.png)

Suggestion:

We could change this button so that it opens a dropdown with two options:

1. Upload file - same as today
2. Insert from URL - this will open a modal with an input field for the URL (similar to media hotlinking)

cc @aaskedall @gajasinski

gajasinski:
@swissspidy seems reasonable. @agingoldseco @aaskedall for thoughts.

Happy to prioritize this accessibility work.

swissspidy:
@agingoldseco @aaskedall Friendly ping, this one still needs some design

Right now what we have is this:

https://user-images.githubusercontent.com/237508/161045350-f28b6eb5-5e2b-47f9-a58e-aa138d55893a.mov

The wording can probably improved and perhaps we can also use buttons with icons instead of labels.

agingoldseco:
Any reason people might want to rename it after upload? I can see that, reopening the link modal, or viewing info about it being reasons to make it clickable. Otherwise, I can work on ways to just make it look less clickable. 

I actually don't mind the text buttons—they're straightforward and go with the rest of the panel. Making them a dropdown might be needlessly adding a click. 

We could potentially add a little blurb about captions to explain why they're good to include for accessibility.

swissspidy:
Renaming is not necessary / not relevant. But being able to click on it to view details (or even replace it) would be nice. Right now it's not possible to do that. You can't even tell whether `wst024_page2_my_captions.vtt` (from the screenshot above) is from my WP media library or hotlinked from somewhere else.

> We could potentially add a little blurb about captions to explain why they're good to include for accessibility.

Just noting that we already have a checklist item for captions.

bmattb:
@agingoldseco have you had a chance to work on this at all?

spacedmonkey:
[...]

PR #11122, opened 2022-03-30
Media: Initial implementation for captions hotlinking 

### LINKED ISSUE #11081: Media: Initial implementation for captions hotlinking
## Feature Description

Splitting this out of #10515 as discussed this week.

While we don't yet have UX for this feature, we can already implement a first version of captions hotlinking behind a feature flag.

spacedmonkey:
## Summary

Add a basic implement of to allow users to hotlink caption files for audio / video element. 

## Relevant Technical Choices

Created a reusable component that can be used for other parts of the code that might need hotlinking support. 

## User-facing changes

https://user-images.githubusercontent.com/237508/161045350-f28b6eb5-5e2b-47f9-a58e-aa138d55893a.mov

## Testing Instructions

1. Enable caption hotlinking experiment. 
2. Create a story. 
3. Insert a video. 
4. Go to caption settings. 
5. Click Link to file. 
6. See dialog. 
7. Try input invalid urls. 
8. Click insert. 
9. See error. 
10. Insert valid url. example https://assets.computingcorner.co.uk/uploads/2022/03/test.vtt
11. Click insert. 
12. See captions functionality work correctly. 

## Checklist

---

Fixes #11081


includes/Experiments.php
```diff
@@ -383,6 +383,17 @@ public function get_experiments(): array {
...
+			/**
+			 * Author: @spacedmonkey
+			 * Issue: #11081
+			 * Creation date: 2022-03-30
+			 */
+			[
+				'name'        => 'captionHotlinking',
+				'label'       => __( 'Caption hotlinking', 'web-stories' ),
+				'description' => __( 'Enable hotlinking of captions', 'web-stories' ),
+				'group'       => 'general',
```
swissspidy:
```suggestion
				'group'       => 'editor',
```

Since this only applies to the editor, not to both editor and dashboard.

packages/e2e-tests/src/specs/editor/backgroundAudio.js
```diff
@@ -155,5 +156,49 @@ describe('Background Audio', () => {
 
       await expect(page).toMatch('test.vtt');
     });
+    describe('Hotlink captions', () => {
+      withExperimentalFeatures(['captionHotlinking']);
+
+      it('should allow adding background audio with captions', async () => {
```
swissspidy, line 167: This test does not actually add captions..?

Could try to hotlink captions from `${process.env.WP_BASE_URL}/wp-content/e2e-assets/test.vtt`

packages/story-editor/src/utils/getErrorMessage.js
```diff
@@ -0,0 +1,39 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * External dependencies
+ */
+import { __, sprintf } from '@googleforcreators/i18n';
+
+function getErrorMessage(code, description) {
```
swissspidy: `getErrorMessage` is a very generic name and `story-editor/src/utils` is for more generic utils, but this function is specific to hotlinking.

Let's move this file to `packages/story-editor/src/components/library/panes/media/local/hotlink`

packages/story-editor/src/components/panels/shared/hotLinkModel.js
```diff
@@ -0,0 +1,203 @@
+/*
```
swissspidy: Let's change the file name: `hotLinkModel.js` -> `hotlinkModal.js` (`e` -> `a`, lowercase `l`)

packages/story-editor/src/components/panels/shared/hotLinkModel.js
```diff
@@ -0,0 +1,203 @@
...
+import useCORSProxy from '../../../utils/useCORSProxy';
+import getErrorMessage from '../../../utils/getErrorMessage';
+
+const InputWrapper = styled.form`
+  margin: 16px 4px;
+  width: 470px;
+  height: 100px;
+`;
+
+function HotlinkModal({
```
swissspidy: There's already a very similar component with the same name in `packages/story-editor/src/components/library/panes/media/local/hotlink/hotlinkModal.js`, used for the existing media hotlinking.

Can we abstract the shared logic into a common component or something to make things more DRY?

swissspidy (commented):
Discovered #11134 while testing this.

packages/e2e-tests/src/specs/editor/backgroundAudio.js
spacedmonkey: So I tried that and it failed. The proxy fail with localhost requests for some reason. How about adding an e2e test plugin to hook the response. Would that be good enough?

swissspidy: Sure! As long as the test actually does what it promises :-)

packages/story-editor/src/components/panels/shared/hotLinkModel.js
spacedmonkey: I did consider this, but felt out of scope of this change. Also without confirmed UX, doing this doesnt make sense. I think we should do this once final UX is in place.

swissspidy: Fair enough. Then this needs a ticket so we don't forget about it.

spacedmonkey:
E2E tests are failing for unreleated reasons it seems.

packages/story-editor/src/components/panels/shared/hotLinkModel.js
spacedmonkey: Ticket created. https://github.com/GoogleForCreators/web-stories-wp/issues/11143

swissspidy (approved):

packages/story-editor/src/components/panels/shared/hotlinkModal.js
```diff
@@ -0,0 +1,205 @@
...
+  const isDisabled = errorMsg || !link || isInserting;
+  const primaryText = isInserting
+    ? __('Inserting…', 'web-stories')
+    : __('Insert', 'web-stories');
+
+  let description = __('No file types are currently supported.', 'web-stories');
+  if (allowedFileTypes.length) {
+    description = sprintf(
+      /* translators: %s is a list of allowed file extensions. */
+      __('You can insert %s.', 'web-stories'),
```
timarney, line 76: Should this be a little more descriptive ?

Current `You can insert vtt.`

Maybe - You can insert a .vtt file

packages/story-editor/src/components/panels/shared/hotlinkModal.js
```diff
@@ -0,0 +1,205 @@
...
+      secondaryText={__('Cancel', 'web-stories')}
+      primaryRest={{ disabled: isDisabled }}
+    >
+      <InputWrapper onSubmit={onSubmit}>
+        <Input
+          ref={inputRef}
+          onChange={({ target: { value } }) => onChange(value)}
+          value={link}
+          hint={errorMsg?.length ? errorMsg : description}
+          hasError={Boolean(errorMsg?.length)}
```
timarney, line 186: Given this happens with the input maybe outside the scope of this PR but when using VoiceOver the error isn't announced.

Should there maybe be a standalone error with a live region to announce the input error for this use case ?

swissspidy: Good call!

I created #11150 the other day to improve our usage of aria-live regions. Will add these types of errors as an example there.

packages/story-editor/src/components/panels/shared/hotlinkModal.js
swissspidy: Agreed.

I just flagged this in #10515 to get some input from Amy & Andrew.

PR #11126, opened 2022-03-30
Accessibility: Resolve issue with duplicate aria-label in reused color component

### LINKED ISSUE #11028: Accessibility: Potential issues with duplicate `aria-label` in reused components
## Task Description

Consider the following scenario:

Screenshot 2022-03-22 at 22 37 00

Here, the color picker for the text element is displayed twice, once in the sidebar and once in the floating toolbar. Both have the same `aria-label="Color input: Text color" attribute.

According to the [`landmark-unique`](https://dequeuniversity.com/rules/axe/4.3/landmark-unique?application=axe-puppeteer) aXe rule this violates a best practice which says those labels should be unique.

Example run: https://github.com/GoogleForCreators/web-stories-wp/runs/5647019338?check_suite_focus=true

We should look into this and perhaps either (selectively) disable this rule or find ways to make these labels unique but still keep them useful.

Any thoughts appreciated.

BrittanyIRL:
## Context

Duplicate landmarks were flagged in e2e now that floating menus are a thing (see https://github.com/GoogleForCreators/web-stories-wp/runs/5647019338?check_suite_focus=true as an example) 

## Summary

The color input is shared across the floating menus and the style panel. When the style panel is open at the same time as the floating menu the label appears duplicate at the container level causing a (moderate) accessibility violation. 

I think there's value in this rule and don't want to turn it off just because of 1 shared input that can be visible in two places at once. My solution here is to make the container label that prefixes the colon of the label (formerly just `Color input: color`) so that the location of the color input could be put in instead. So now we have 2 variants: `Color input: {label}` (default) and `Color input (floating menu): {label}` - could easily have a third that is anything else `super awesome color input : {label}`. Anyway, you get the point. This also keeps all our existing test infrastructure intact. 

## Relevant Technical Choices

Pass in an optional containerLabelBase to distinguish between which variant is in use and avoid duplication

## To-do

So far as I can tell running aXe locally on the editor, this is the only reused component getting caught in this issue. When e2e runs, check to see if there are others. 

Need to update some karma fixture labels

## User-facing changes

Just label changes.

## Testing Instructions

If you have aXe devtools you can run a comparison on main vs this branch. Insert an element into the canvas and with it selected open the style panel for it. Make sure that `text` tab is open and then run aXe on the whole page. 

| main | this branch | 
| --- | --- | 
| Screen Shot 2022-03-30 at 3 23 07 PM | Screen Shot 2022-03-30 at 3 22 13 PM |

This is the issue we're addressing here: 
Screen Shot 2022-03-30 at 3 23 21 PM

Alternatively, check and see that the color containers are labeled like so: 
| design panel | floating menu | 
| --- | --- | 
| Screen Shot 2022-03-30 at 3 33 15 PM | Screen Shot 2022-03-30 at 3 32 58 PM |
| Screen Shot 2022-03-30 at 3 31 28 PM | Screen Shot 2022-03-30 at 3 32 02 PM |

1.

No

No

No

## Checklist

---

Fixes #11028


packages/story-editor/src/components/floatingMenu/elements/shared/color.js
```diff
@@ -52,6 +53,7 @@ function FloatingColor(props) {
       allowsSavedColors
       allowsSavedColorDeletion={false}
       pickerHasEyedropper={!hasEyedropper}
+      containerLabelBase={__('Color input (floating menu)', 'web-stories')}
```
samwhale, line 56: 💭 Since this could be read by screen reader, maybe this would make more sense?
```suggestion
      containerLabelBase={__('Floating Menu Color input', 'web-stories')}
```

samwhale (approved):

BrookeGraham (approved):
🎨

packages/story-editor/src/components/floatingMenu/elements/shared/color.js
samwhale: or (Color input for floating menu)

maxyinger (approved):

PR #11136, opened 2022-03-31
Fix center snapping

### LINKED ISSUE #11053: Bring back center snapping and fix shortcut bug
## Bug Description

Since updating to 1.16 there have been complaints about snapping no longer working as expected, and this was also observed in yesterday's bug-bash.

In my testing I have noted the following:
1. this issue seems to happen early in story creation
2. This issue seems to self resolve sometimes
3. This issue is very challenging to consistently reproduce.
4. This issue has existed since v1.16 at least

Issues found:
1. Center snapping guides do not appear in the editor consistently
2. Object snapping doesn't seem to occur consistently
3. Placing an image as background by using edge fill fails

Discussion with another XWP developer it was noted:

> Cached assets probably.
First time or two it doesn't load, but then all the assets are loaded and cached and then it does.

Here are some examples of my findings: (all in chrome on macos)

**v1.16 snapping test:** Not working 
https://images.zenhubusercontent.com/235435637/6358a2d5-cf3c-43e6-a4a1-a90ee6330526/snapping_test___v1_16.mov

**v1.18.1 snapping test:** Working
https://images.zenhubusercontent.com/235435637/a82d641a-27bf-4f40-9477-9bee7a53dadf/snapping_test___v1_18_1.mov
## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

merapi:
## Summary

This PR fixes two issues with snapping:

 ### 1) Center/middle snapping
`react-moveable` replaced `snapCenter` with `snapDirections` ([and others](https://github.com/daybrush/moveable/releases/tag/0.27.0)) which has `(default: { left: true, top: true, right: true, bottom: true })` without `center/middle`. The upstream change was made on 2021-11-09 and it went unnoticed.
If they had used proper semantic versioning it wouldn't have happened? @swissspidy 

Before:

https://user-images.githubusercontent.com/21036224/161091050-1758cc40-5588-4395-88d6-1b0ff5cfc138.mp4

After:

https://user-images.githubusercontent.com/21036224/161091406-32e63891-7a3e-4350-a7c3-2de793fa46e1.mp4

 ### 2) Bugged "Disable snapping and drop targeting"

Especially when the user switched windows using MOD+TAB, the MOD key state was still set to pressed even if that was no longer the case (and that caused no snapping and drop targeting at all). A temporary "workaround" was to press MOD once after you return to the window. Fix in this PR sets the key state to not-pressed when you blur the window - not sure if there is any better solution?

https://user-images.githubusercontent.com/21036224/161091826-a7d8b013-dda5-4631-bfc5-aa55b11c5909.mp4

3.

## Testing Instructions

1. Check if the centerline snapping of the element is working.
2. Try to reproduce the issue with the "sticky" MOD key.

## Checklist

---

Fixes #11053

### DIFF
--- a/packages/design-system/src/components/keyboard/index.js
+++ b/packages/design-system/src/components/keyboard/index.js
@@ -25,6 +25,7 @@ import {
   useState,
   useContext,
   useBatchingCallback,
+  useCallback,
 } from '@googleforcreators/react';
 /**
  * Internal dependencies
@@ -181,6 +182,17 @@ export function useKeyUpEffect(
  */
 export function useIsKeyPressed(refOrNode, keyNameOrSpec, deps = undefined) {
   const [isKeyPressed, setIsKeyPressed] = useState(false);
+
+  const handleBlur = useCallback(() => {
+    setIsKeyPressed(false);
+  }, []);
+  useEffect(() => {
+    window.addEventListener('blur', handleBlur);
+    return function () {
+      window.removeEventListener('blur', handleBlur);
+    };
+  }, [handleBlur]);
+
   //eslint-disable-next-line react-hooks/exhaustive-deps -- Pass through provided deps.
   useKeyDownEffect(refOrNode, keyNameOrSpec, () => setIsKeyPressed(true), deps);
   //eslint-disable-next-line react-hooks/exhaustive-deps -- Pass through provided deps.
--- a/packages/story-editor/src/components/canvas/karma/snapping.karma.js
+++ b/packages/story-editor/src/components/canvas/karma/snapping.karma.js
@@ -57,5 +57,17 @@ describe('Snapping integration', () => {
 
       await fixture.snapshot('Design space guideline hidden');
     });
+
+    it('should show canvas center guidelines when moved to the center', async () => {
+      await fixture.events.mouse.seq(({ moveRel, moveBy, down }) => [
+        moveRel(frame1, 10, 10),
+        down(),
+        moveBy(10, 150, { steps: 6 }),
+      ]);
+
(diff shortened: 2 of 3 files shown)


packages/design-system/src/components/keyboard/index.js
```diff
@@ -181,6 +182,17 @@ export function useKeyUpEffect(
...
+  const handleBlur = useCallback(() => {
+    setIsKeyPressed(false);
+  }, []);
+  useEffect(() => {
+    window.addEventListener('blur', handleBlur);
+    return function () {
+      window.removeEventListener('blur', handleBlur);
+    };
+  }, [handleBlur]);
+
```
merapi, line 195: Do we want to create a custom hook?

barklund (approved):
Works great - nice and easy fix for once!

packages/design-system/src/components/keyboard/index.js
swissspidy: Maybe In the future if we need to reuse it 👍

swissspidy:
A bit unfortunate with that Moveable change 😞 Could you think of a good way to test this behavior / prevent this from happening again in the future?

swissspidy (approved):

merapi:
> A bit unfortunate with that Moveable change 😞 Could you think of a good way to test this behavior / prevent this from happening again in the future?

We can prevent the same issue from happening again by checking if the snapping lines are present when moving (or via snapshot diff, but that's sometimes skipped). It can check all lines, not only center. Moving to QA and will work on it.

PR #11138, opened 2022-03-31
Discovery: Splitting apart Popup, Tooltip and Dropdown

### LINKED ISSUE #6921: Discovery: Separating Popups by Intention
(Initially focused on tooltips) 
We're currently using portals for our tooltips, historically this is because we need tooltips to overflow out of elements in some occasions across the editor and dashboard. With the redesign though we should know that this is still a design choice we need to follow because portals are more expensive than inline elements and if we can avoid them it might be worth it.  The other thing to consider here are the tails, so this whole thing might be unnecessary but we should do a little discovery and know for sure. 

More info: https://github.com/google/web-stories-wp/pull/6887#discussion_r599128462

Fast forward a year from making this ticket and the underlaying popup of the tooltip in the design system has about 19 different props getting passed to it for various edge case scenerios. This is getting really wacky to maintain and is the source of at least 1 regression per month. We can do better. Let's separate concerns. 

This ticket is to do discovery around alternative methods to handle our many many many popup situations (tooltips, context menus, floating menus, footer popups, dropdown menus, color picker...)

BrookeGraham:
## Context

With `Popup` having to deal with `Tooltip`, `Dropdown` `Search` and "basic" implementations like `Color input`, `PublishTime` `DataList` etc. it has become a bit complicated to maintain each use case. Looking to see what splitting it apart would look like. 

## Relevant Technical Choices
`Dropdown` `Color Input` and `Tooltip` now each have there own implementations of a `Popup`

## To-do

- ~~pull out Dropdown~~
- ~~pull out Search~~
- ~~What about Color Input ?~~
- RTL tooltip are a bit whacky. 
- getOffset should account for the edge of a screen whether or not in RTL
- Investigate `Search` and `DataList` - should these really be `Dropdown`s or does `Popup` still make sense? Or should these uses also be their own implementation? 

Fixes #6921

### DIFF
--- a/packages/design-system/src/components/popup/utils/getOffset.js
+++ b/packages/design-system/src/components/popup/utils/getOffset.js
@@ -115,7 +115,7 @@ export function getOffset({
   dock,
   popup,
   isRTL,
-  topOffset,
+  topOffset = 0, // set default then we won't have to send it in with Tooltip
   ignoreMaxOffsetY,
   resetXOffset,
 }) {
--- a/packages/design-system/src/components/popup/utils/getOffset.js
+++ b/packages/design-system/src/components/popup/utils/getOffset.js
@@ -149,10 +149,12 @@ export function getOffset({
   const offsetY = getYOffset(placement, spacingV, anchorRect);
   const maxOffsetY =
     bodyRect.height + bodyRect.y - height - getYTransforms(placement) * height;
+  // maybe pull in leftOffset from config to account to LTR wp-admin bar?
+  const leftEdge = popupRect?.left <= 0 ? (popupRect?.width + spacingH) / 2 : 0;
 
   // Clamp values
   return {
-    x: Math.max(0, Math.min(offsetX(), maxOffsetX)),
+    x: Math.max(leftEdge, Math.min(offsetX(), maxOffsetX)),
     y: ignoreMaxOffsetY
       ? offsetY
       : Math.max(topOffset, Math.min(offsetY, maxOffsetY)),
--- a/packages/design-system/src/components/popup/utils/getOffset.js
+++ b/packages/design-system/src/components/popup/utils/getOffset.js
@@ -135,7 +135,10 @@ export function getOffset({
 
   // Horizontal
   const offsetX = () => {
-    if (resetXOffset && popupRect?.left <= 0) {
+    // resetXOffset was added because the layers panel popup is getting weird x values in the negative range. So if we
+    // just check for popupRect.left <= 0 we end up pushing the tooltips inside the layers panel off to the edge. We
+    // should figure out why the layer's panel is getting whacked values.
+    if (resetXOffset && popupRect && popupRect.left <= 0) {
       return isRTL ? width : 0;
     }
     return getXOffset(placement, spacingH, anchorRect, dockRect, isRTL);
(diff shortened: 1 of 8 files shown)


packages/story-editor/src/components/form/color/colorInput.js
```diff
@@ -38,8 +42,9 @@ import {
   THEME_CONSTANTS,
   Swatch,
   Icons,
-  Popup,
```
BrookeGraham, line 41: I think now that I have removed `Tooltip` and `Dropdown` as use cases from `Popup` it would be fine to leave `Color Input` in. It isn't really doing anything different than what a "normal" `Popup` would be.

packages/design-system/src/components/tooltip/index.js
```diff
@@ -223,6 +236,77 @@ function Tooltip({
...
+      return undefined;
+    }
+    isPopupMounted.current = true;
+    positionPopup();
+    // Adjust the position when scrolling.
+    document.addEventListener('scroll', positionPopup, true);
+    return () => {
+      document.removeEventListener('scroll', positionPopup, true);
+      isPopupMounted.current = false;
+    };
```
BrookeGraham, line 290: Do we ever really expect to scroll a tooltip?

packages/design-system/src/components/popup/utils/getOffset.js
```diff
@@ -146,10 +149,12 @@ export function getOffset({
   const offsetY = getYOffset(placement, spacingV, anchorRect);
   const maxOffsetY =
     bodyRect.height + bodyRect.y - height - getYTransforms(placement) * height;
+  // maybe pull in leftOffset from config to account to LTR wp-admin bar?
+  const leftEdge = popupRect?.left <= 0 ? (popupRect?.width + spacingH) / 2 : 0;
```
BrookeGraham, line 153: This will account for the left edge getting cutoff but only when `center` aligned. The left / right tooltip are still a bit outta whack.

packages/design-system/src/components/tooltip/index.js
```diff
@@ -111,28 +117,29 @@ function Tooltip({
   onFocus = noop,
   onBlur = noop,
   isDelayed = false,
-  forceAnchorRef = null,
+  // forceAnchorRef = null, needed for WithLink so that the url tooltip hovers the element, and isn't anchor
+  // to just the canvas. Also used in Slider, but I couldn't trigger it.
+  // could remove tooltipProps, do we need it anymore
```
BrookeGraham, line 122: 💭 We could just use `dock` instead. It takes precedence in getOffset, so no additional logic would be needed here for forceAnchor vs anchor

packages/story-editor/src/components/panels/design/layer/layer.js
```diff
@@ -342,6 +342,7 @@ function Layer({ element }) {
               title={__('Duplicate Layer', 'web-stories')}
               hasTail
               isDelayed
+              placement={'bottom-end'}
```
BrookeGraham, line 345: In addition to updating this placement, we should be importing `Tooltip` from `Story-Editor` to get the correct placements. This should fix the negative values issue as well.

BrittanyIRL:
Closing since this is reference discovery. Thank you so much for the initial effort here!!

PR #11144, opened 2022-04-01
Improve "Failed to save the story" message

### LINKED ISSUE #11082: Improve "Failed to save the story" message
## Feature Description

Splitting this out of #10963

If we fail to save the story, we do not currently provide any context about why that happens.

For example, if the REST API returns a 5xx or 4xx status, we should be able to say that in the message.

Let's:

* Get the `err` from the `catch` and see if we can enhance the snackbar message. Maybe with `err.message`?
* Call `console.log(err)` so it's easier to spot this in dev tools; helps with support
* Call `trackError()` so we get telemetry on these errors.

Here's where we display the message:

https://github.com/GoogleForCreators/web-stories-wp/blob/3cc4ee5c5c02eb9dc2e652660a5d7a731aa3babe/packages/story-editor/src/app/story/actions/useSaveStory.js#L108-L113

We could even look into adding an e2e test for this, with a test plugin that mocks typical scenarios like 5xx or 4xx status. Like we do for status check.

timarney:
## Context

Adds more specific error message for when a story fails to save.

See https://github.com/GoogleForCreators/web-stories-wp/issues/11082#issue-1180689949 for more details

## Summary

Updates failed save story to handle 400, 401, 403, 500 errors + track errors

## Testing Instructions

1.

No 

Yes --- tracks error here 
https://github.com/GoogleForCreators/web-stories-wp/pull/11144/files#diff-268fab14c97529f4d480ac6c70e2082129e453aceff794f2237dcc6e3bad9180R152

No

## Checklist

---

Fixes #11082

### DIFF
--- a/packages/e2e-tests/src/specs/dashboard/dashboard.js
+++ b/packages/e2e-tests/src/specs/dashboard/dashboard.js
@@ -21,6 +21,9 @@ import {
   takeSnapshot,
   withRTL,
   visitDashboard,
+  createNewStory,
+  insertStoryTitle,
+  publishStory,
 } from '@web-stories-wp/e2e-test-utils';
 
 const percyCSS = `.dashboard-grid-item-date { display: none; }`;
@@ -43,8 +46,12 @@ describe('Stories Dashboard', () => {
   });
 
   it('should choose sort option for display', async () => {
+    // dropdown needs a story for filtering
+    await createNewStory();
+    await insertStoryTitle('Stories Dashboard test - story');
+    await publishStory();
     await visitDashboard();
-    await expect(page).toClick('[aria-label="Dashboard"]', {
+    await expect(page).toClick('[aria-label="Dashboard (active view)"]', {
       text: 'Dashboard',
     });
     const sortButtonSelector =
--- a/packages/e2e-tests/src/specs/editor/saveStory.js
+++ b/packages/e2e-tests/src/specs/editor/saveStory.js
@@ -29,7 +29,7 @@ import {
  */
 import { addAllowedErrorMessage } from '../../config/bootstrap';
 
-describe('Save story snackbar', () => {
+describe('Saving Story', () => {
   let removeErrorMessage;
   let stopRequestInterception;
   let mockResponse;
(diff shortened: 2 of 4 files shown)


timarney:
@spacedmonkey @swissspidy 

I'm having an issue getting an e2e test running properly for this ...

For real testing if I force an error from the `Stories_Controller.php` i.e.

Screen Shot 2022-04-04 at 7 31 07 AM

I get the expected result (SnackBar message with a server error)

| Response  |      Result    |  
|----------|:-------------:|
| real-response |  Screen Shot 2022-04-04 at 7 24 24 AM | 


<hr>

In Jest I can see the Mock response I setup seems to be working but no SnackBar message shows --- I also see Jest has a expected not to 500 error.  Is there something for Jest I need to adjust to allow the 500 in this case?


### Jest 

| Response  |      Result    |  
|----------|:-------------:|
| test-response |  jest|

swissspidy:
Apologies for the delay here.

> Is there something for Jest I need to adjust to allow the 500 in this case?

Yes. See 6caa3ab8e214cdabd036dd3602a5a37474bc9d02

**Explanation:** 

We use a package called [`@wordpress/jest-console`](https://npmjs.com/package/@wordpress/jest-console) to automatically surface browser console messages to Jest. This allows us to catch potentially unexpected errors (which would indicate a bug). In this case of course we are expecting this log message because we are forcing it ourselves.

I also took the liberty of refining the snackbar messages a little bit.

Not sure how well it works because I don't think we really have access to the HTTP status code in the `catch` block right now.

timarney:
Rebased this to latest main

packages/e2e-tests/src/specs/dashboard/dashboard.js
```diff
@@ -43,8 +46,12 @@ describe('Stories Dashboard', () => {
   });
 
   it('should choose sort option for display', async () => {
+    // dropdown needs a story for filtering
```
timarney: Noticed the dashboard test was failing ...

The dropdown didn't have a story to work with.

timarney: I cherry picked that commit for this PR

https://github.com/GoogleForCreators/web-stories-wp/pull/11227

swissspidy:
While debugging an issue reported via support, I noticed that errors thrown in `getStoryMarkup` (where we assemble the full HTML for a story) were not caught. I just fixed this now in 62d1df4d9b51ba030e7fe28231f2c9549368a8c2 as part of this PR. Works like a charm:

Screenshot 2022-04-11 at 15 37 10

swissspidy (approved):

Issue #10846, opened 2022-03-08
Shopping Experience


swissspidy:
## Summary

The Web Stories format is launching a new set of components for a configurable, templated shopping experience within stories.

<img src="https://user-images.githubusercontent.com/3860311/152416815-61a63a88-76c0-4ae8-9300-a7ad277dbe7a.jpg" alt="amp-story-shopping demo" />

We want to bring this to WordPress as well. In a first step using an integration with WooCommerce.

## References

* [Preview / I2S](https://github.com/ampproject/amphtml/issues/37570)
* [Demo story](https://wsdemos.uc.r.appspot.com/shopping)
* [**Figma**](https://www.figma.com/file/ZrO763yvMWJjYeAPsRnNfW/_Editor---Sprint-Feature-Level-Designs_2022?node-id=503%3A15921)

### Does this epic have telemetry?

Yes, definitely needed!

Issue #11157, opened 2022-04-01
Shopping: Settings Input Validation


swissspidy:
For shopping, users are able to enter their Shopify host/domain and API key on the settings page.

We need to validate these values to ensure they adhere to the required format.

For example, the host/domain should have the format `pascals-test-store.myshopify.com` with no protocol or pathname or trailing slash.

Issue #11154, opened 2022-04-01
Shopping: Product Query


swissspidy:
We need an easy way to easily query products from WooCommerce or Shopify, depending on what has been set up by the user.

For WooCommerce it means using the provided PHP functions/classes. For Shopify it means writing GraphQL queries.

In the POC these queries were simply hardcoded. But we need to make this more robust and support more features such as sorting by various fields.

Perhaps a `Product_Query` class makes sense that supports both providers. Open for feedback in this regard.

Maybe we can even make the product query logic filterable/extensible so that other plugins could hook into it.

Documentation:

* https://shopify.dev/api/storefront/2022-01/queries/products
* https://github.com/woocommerce/woocommerce/wiki/wc_get_products-and-WC_Product_Query

timarney:
Makes sense --- seems like enough there to start a prototype.

--- would be good to define a few things like fields we'll sort on

> features such as sorting by various fields.

For searching are we taking into account any status, type , inventory, maybe product tags or categories?

Looking at Shopify sorting 

<img width="218" alt="Screen Shot 2022-04-21 at 7 25 37 AM" src="https://user-images.githubusercontent.com/62242/164448259-d0fdf7f7-3ebd-40dc-abcd-c6bfa0c49cc5.png">

swissspidy:
For the sorting we'll only need sorting by name, price, and (modified) date (see #11253 ), both ascending and descending. At least for now that should suffice.

spacedmonkey:
@swissspidy @timarney  I have created a POC of what I have discussed above here. https://github.com/GoogleForCreators/web-stories-wp/pull/11340

Let me know if you like what I have done and I will add unit tests.

PR #11158, opened 2022-04-01
Shopping: Add initial implementation

### LINKED ISSUE #11152: Shopping: Initial implementation
Build initial implementation based on the POC shared a while back: https://drive.google.com/file/d/1M7U3feb-tXOIwMdh-kTP24A03NzMPQUl/view

This is gonna be a very rough version that does not match latest designs.

swissspidy:
## Context

The aim for this PR is to implement a very rough initial version of the shopping integration based on [the POC shared a while back](https://drive.google.com/file/d/13lN9YLNBtWaODBof36oyZwD-HfKccjP5/view)

## Summary

This gets the ball rolling by laying the foundation for future work on the shopping functionality.

**Wanna test this?**

Two options:

1. Set up a WooCommerce store on your site and create some products
2. Ping me for some Shopify API credentials for testing

Note: need to disable AMP sanitization (e.g. by commenting out the relevant line in `Output_Buffer`) for now since this is not stable yet.

## Relevant Technical Choices

**New element attributes: `supportsRotation`, `lockDimensions`**

Product tags on a page cannot be rotated or resized. To support this I created these new attributes on the element level (see `defaultAttributes` and `SingleSelectionMoveable`.

**WooCommerce stubs**

The REST API controller in PHP uses some WooCommerce functions like `wc_get_products` and `get_woocommerce_currency`. Thanks to `php-stubs/woocommerce-stubs` PHPStan will recognize them better.

**No GraphQL PHP SDK**

This PR does not use a dedicated GraphQL API library for talking to the Shopify GraphQL API.

Existing solutions are either rather outdated/unmaintained or have higher PHP version requirement. Also, it would unnecessarily add bloat to this plugin, since we only really need a simple product query and nothing else.

## To-do

There are tons of open to-dos; basically everything currently open under the #10846 epic, but also all sorts of tests etc. These can be added once the design has been finalized.

## User-facing changes

Screenshot 2022-04-04 at 21 28 54
Screenshot 2022-04-04 at 21 28 51
Screenshot 2022-04-04 at 21 28 18

## Testing Instructions

1.

## Checklist

---

Fixes #11152


timarney:
Will do more testing tomorrow --- with your API key 

Testing using my dev store  --- I'm getting some broken results (likely store settings).


Screen Shot 2022-04-04 at 8 19 56 PM

includes/REST_API/Products_Controller.php
```diff
@@ -0,0 +1,489 @@
...
+			[
+				'headers' => [
+					'Content-Type'                      => 'application/graphql',
+					'X-Shopify-Storefront-Access-Token' => $access_token,
+				],
+				'body'    => $query,
+			]
+		);
+
+		if ( WP_Http::OK !== wp_remote_retrieve_response_code( $response ) ) {
```
timarney, line 277: It would be good to dump some errors here for dev troubleshooting

packages/wp-dashboard/src/components/editorSettings/shopping/shopify/index.js
```diff
@@ -0,0 +1,242 @@
...
+  SettingsTextInput,
+  TextInputHelperText,
+  VisuallyHiddenLabel,
+} from '../../components';
+
+export const TEXT = {
+  HOST_CONTEXT: sprintf(
+    /* translators: 1: first example. 2: second example. */
+    __(
+      'Without <code>%1$s</code> or <code>%2$s</code>. Example: %3$s. <a>Find your URL</a>.',
```
timarney, line 46: Noting the `Without https:` here is easy to miss when cutting and pasting a store name --- fine for this PR but should add validation around this.

swissspidy: There is #11157 for this

includes/REST_API/Products_Controller.php
swissspidy: Not sure about dumping some errors as we don't typically do that elsewhere (yet), but definitely would need more robust error handling. Probably as part of #11154.

swissspidy:
> Testing using my dev store --- I'm getting some broken results (likely store settings).

Oh neat, you have a dev store!

That actually looks fine :-)

Products are getting populated in the dropdown, which is great.

The floating menu for product tags on canvas hasn't been implemented yet (pending UI design), so that currently appears empty.

And the "Shop Now" text at the bottom of the canvas (below the arrow) currently lacks some text shadow, so right now it's white on white

timarney:
Should I be seeing this show up on the front-end yet? 

Going to approve as is as I think this PR covers the setup work as intended and we can start picking up the todo.

| Admin   |      Front-end      |  
|----------|:-------------:|
| Screen Shot 2022-04-05 at 5 54 04 AM |  Screen Shot 2022-04-05 at 5 54 42 AM |

timarney (approved):
Nice work this PR covers a lot of ground to get this feature rolling

 🎉

swissspidy:
> Should I be seeing this show up on the front-end yet?

Yes but you need to disable AMP sanitization (see PR description for how to do that). Since this new component is not stable yet in AMP, it causes AMP validation errors, so the sanitization strips that out.

merapi (approved):
Looks good 👍🏻   (I'm holding off with some comments because the final design is not there yet).

I've installed WooCommerce and removed `$this->sanitization->sanitize_document( $dom );` for preview to work.

image

includes/REST_API/Products_Controller.php
```diff
@@ -0,0 +1,489 @@
...
+	 */
+	public function register_routes(): void {
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base,
+			[
+				[
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => [ $this, 'get_items' ],
+					'permission_callback' => [ $this, 'permissions_check' ],
```
spacedmonkey:
```suggestion
					'permission_callback' => [ $this, 'get_items_permissions_check' ],
```

includes/REST_API/Products_Controller.php
```diff
@@ -0,0 +1,489 @@
...
+	}
+
+	/**
+	 * Checks if a given request has access to get and create items.
+	 *
+	 * @since 1.20.0
+	 *
+	 * @return true|WP_Error True if the request has read access, WP_Error object otherwise.
+	 */
+	public function permissions_check() {
```
spacedmonkey:
```suggestion
	public function get_items_permissions_check() {
```

spacedmonkey: We should allow the core conversion here and use this naming.

includes/REST_API/Products_Controller.php
```diff
@@ -0,0 +1,489 @@
...
+		if ( ! $this->story_post_type->has_cap( 'edit_posts' ) ) {
+			return new \WP_Error(
```
spacedmonkey, line 125: Why slash this and not use the class?

includes/REST_API/Products_Controller.php
```diff
@@ -0,0 +1,489 @@
...
+		$shopify_host         = $this->settings->get_setting( Settings::SETTING_NAME_SHOPIFY_HOST );
+		$shopify_access_token = $this->settings->get_setting( Settings::SETTING_NAME_SHOPIFY_ACCESS_TOKEN );
+
+		if ( ! empty( $shopify_host ) && ! empty( $shopify_access_token ) ) {
+			return $this->get_items_shopify( $request );
+		}
+
+		if ( function_exists( 'wc_get_products' ) ) {
+			return $this->get_items_woocommerce( $request );
+		}
```
spacedmonkey, line 156: What is happening with the ordering here? Is shopify seen as more important than WOO? Can you have both in installed at the same time?

includes/REST_API/Products_Controller.php
swissspidy: We do it like this literally everywhere in the code base, IIRC it's a PHPCS config.

includes/REST_API/Products_Controller.php
swissspidy: Right now, yes.

If this changes in design/UX the code will have to be adapted accordingly.

includes/REST_API/Products_Controller.php
```diff
@@ -0,0 +1,489 @@
...
+
+	/**
+	 * Retrieves all products.
+	 *
+	 * @since 1.20.0
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
+	 */
+	public function get_items( $request ) {
```
spacedmonkey, line 143: Currently this endpoint does not respect fields that are passed to it. A `prepare_item_for_response` method should be added. Also `get_items_shopify` and `get_items_woocommerce` should really just return an array / object of data or WP_Error. Formatting of that data should happen outside of these methods, so this code is more testable and DRY.

includes/REST_API/Products_Controller.php
```diff
@@ -0,0 +1,489 @@
...
+
+		/**
+		 * Filters the shopify products data TTL value.
+		 *
+		 * @since 1.20.0
+		 *
+		 * @param int $time Time to live (in seconds). Default is 1 day.
+		 * @param WP_REST_Request $request Full details about the request.
+		 */
+		$cache_ttl = apply_filters( 'web_stories_shopify_data_cache_ttl', DAY_IN_SECONDS, $request );
```
spacedmonkey: What happens if I product is removed or is out of stock? Caching for a day seems like a long time. What is the product is removed for safety reasons and is a limited "drop". Is there a cache control header sent by the api, maybe we could respect that.

includes/REST_API/Products_Controller.php
spacedmonkey: I think we should try and work out what the error was here and show different errors. 
Example, 
1. auth error. 
2. Rate limit error
3. 404 / missing item error. 
4. 500 error. 

If we are proxying there api, I think it would be extremely useful for debug, if we don't just swallow that error.

includes/REST_API/Products_Controller.php
```diff
@@ -0,0 +1,489 @@
...
+				];
+			}
+
+			$results[] = [
+				'productId'            => $product['id'],
+				'productTitle'         => $product['title'],
+				'productBrand'         => $product['vendor'],
+				// TODO: Maybe eventually provide full price range.
+				// See https://github.com/ampproject/amphtml/issues/37957.
+				'productPrice'         => $product['priceRange']['minVariantPrice']['amount'],
```
spacedmonkey, line 317: How does this work? What if you have a product that has variants from $1 to $1000. Isn't this prices from $1? Is there anyway to show there are variants and prices may different. 

Worth noting that some countries is a little matter if you do not show the correct price next to the item.

includes/REST_API/Products_Controller.php
```diff
@@ -0,0 +1,489 @@
...
+				'productPriceCurrency' => $product['priceRange']['minVariantPrice']['currencyCode'],
+				'productImages'        => $images,
+				// TODO: Remove once optional, or find way to actually populate.
+				// See https://github.com/ampproject/amphtml/pull/37474.
+				'aggregateRating'      => [
+					'ratingValue' => 0,
+					'reviewCount' => 0,
+					'reviewUrl'   => $product['onlineStoreUrl'],
+				],
+				'productDetails'       => $product['description'],
```
spacedmonkey, line 320: What this descriptioin has html or other markup in it? Do we have to clean this?

includes/REST_API/Products_Controller.php
```diff
@@ -0,0 +1,489 @@
...
+			}
+
+			$results[] = [
+				'productId'            => $product['id'],
+				'productTitle'         => $product['title'],
+				'productBrand'         => $product['vendor'],
+				// TODO: Maybe eventually provide full price range.
+				// See https://github.com/ampproject/amphtml/issues/37957.
+				'productPrice'         => $product['priceRange']['minVariantPrice']['amount'],
+				'productPriceCurrency' => $product['priceRange']['minVariantPrice']['currencyCode'],
```
spacedmonkey, line 318: How doe this work? Can you request this in different currencies? does it default to us dollars? Should be have an option to force to a currency, like if you are publisher in the UK, to force pounds etc?

includes/REST_API/Products_Controller.php
swissspidy: Yes, but that's something for #11154, as indicated in the PR description. This code is still very rough because of that.

includes/REST_API/Products_Controller.php
swissspidy: Yeah we'll probably lower this to a couple of minutes or so in the future

includes/REST_API/Products_Controller.php
```diff
@@ -0,0 +1,489 @@
...
+				'status'  => 'publish',
+				'limit'   => 100,
+				'orderby' => 'date',
+				'order'   => 'DESC',
+				's'       => $request['search'],
+			]
+		);
+
+		foreach ( $products as $product ) {
+			$images    = array_map(
```
spacedmonkey: Can you prime caches first. Like this

```
_prime_post_cache( $product->get_gallery_image_ids() );
```

includes/REST_API/Products_Controller.php
swissspidy: Product ranges are not supported yet by the Web Stories format itself in, which is why this TODO links to https://github.com/ampproject/amphtml/issues/37957. Once the format supports it, this will need updating.

There's nothing we can do about it right now.

includes/REST_API/Products_Controller.php
swissspidy: Great question.

This simply takes the price and currency as configured in Shopify. Whatever you set up there, we use.

If you're a publisher in the UK you would surely configure the shop to use the same. Providing overrides is out of scope for this implementation.

includes/REST_API/Products_Controller.php
swissspidy: Hmm is this really needed?

We can take a closer look at this in #11154

includes/REST_API/Products_Controller.php
swissspidy: Good question, haven't tested this extensively. We can do that in #11154 or some other follow-up ticket

includes/REST_API/Products_Controller.php
```diff
@@ -0,0 +1,489 @@
...
+				static function( $image_id ) {
+					$url = wp_get_attachment_url( $image_id );
+					return [ 'url' => $url ];
```
spacedmonkey:
```suggestion
					$url = wp_get_attachment_url( $image_id );
					$alt = get_post_meta( $image_id, '_wp_attachment_image_alt', true );
					return [ 'alt' => $alt, 'url' => $url ];
```

Should we also add alt here as well?

includes/REST_API/Products_Controller.php
```diff
@@ -0,0 +1,489 @@
...
+			];
+		}
+
+		return rest_ensure_response( $results );
+	}
+
+	/**
+	 * Retrieves the publisher logo's schema, conforming to JSON Schema.
+	 *
+	 * @since 1.12.0
```
spacedmonkey: Incorrect version.

includes/REST_API/Products_Controller.php
```diff
@@ -0,0 +1,489 @@
...
+		 * Shopify GraphQL API response.
+		 *
+		 * @var ShopifyGraphQLResponse $result
+		 */
+		$result = json_decode( wp_remote_retrieve_body( $response ), true );
+
+		// TODO: Error handling.
+		if ( isset( $result['errors'] ) ) {
+			return new \WP_Error( 'rest_unknown', __( 'Error fetching products', 'web-stories' ), [ 'status' => 404 ] );
+		}
```
spacedmonkey, line 291: We should not swallow all errors here. Can you at least see what the error and return different errors in our API.

swissspidy: Yes, that's why there is a TODO for this here.

See also https://github.com/GoogleForCreators/web-stories-wp/pull/11158#discussion_r842238474

includes/REST_API/Products_Controller.php
```diff
@@ -0,0 +1,489 @@
...
+	 * Retrieves all Shopify products.
+	 *
+	 * @SuppressWarnings(PHPMD.ExcessiveMethodLength)
+	 *
+	 * @since 1.20.0
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
+	 */
+	protected function get_items_shopify( $request ) {
```
spacedmonkey, line 171: I kind of feel like this code should maybe live in another class. This code is related to get shopify products and is not related to the endpoints. Having this as another class, should be massively useful for mocking and testing.

swissspidy: Yes. This is what #11154 is mainly for.

includes/REST_API/Products_Controller.php
spacedmonkey: It is much better for performance. Instead of loading one attachment at a time, load them all at once. See this example from code. 

https://github.com/WordPress/wordpress-develop/blob/05141500261c845fb4dca2da28c967dc94c5e551/src/wp-includes/media.php#L1821

includes/REST_API/Products_Controller.php
swissspidy: Just checked. This does not have markup in it. See https://shopify.dev/api/storefront/2022-04/objects/Product#field-product-description

spacedmonkey: I would still strip tags. The worry here is two fold, 1, that markup is added later. 2. Some level man in the middle attack, that would add markup.

includes/REST_API/Products_Controller.php
spacedmonkey: I still have questions on how this work. What if you have a product and want to show different prices were location? Same you are an British publisher but want to show an item in dollars if you are viewing in the USA. For example. I think these limitations need to be defined somewhere, so user understand how this feature works.

swissspidy: The format does not support this, so this would need to be filed as a feature request there.

spacedmonkey:
If woo is installed, the options to use shopify should be hidden.

includes/REST_API/Products_Controller.php
```diff
@@ -0,0 +1,506 @@
...
+			/*
+			 * Warm the object cache with post and meta information for all found
+			 * images to avoid making individual database calls.
+			 */
+			_prime_post_caches( $product_image_ids, false, true );
+
+			$images = array_map(
+				static function( $image_id ) {
+					$url = wp_get_attachment_url( $image_id );
+					$alt = get_post_meta( $image_id, '_wp_attachment_image_alt', true );
```
spacedmonkey, line 371: This should be cast to a string to be sure.

swissspidy:
> If woo is installed, the options to use shopify should be hidden.

Not necessarily. This is out of scope for this PR and dependent on further UX rounds.

PR #11160, opened 2022-04-01
Editor: Removed superfluous data load

### LINKED ISSUE #10246: Video Trimmer: don't call `getVideoData` on mount
## Feature Description

`useVideoTrimMode()` calls `getVideoData` when element selection changes. So whenever I select a video element, it calls this function which causes an HTTP request to be made.

This seems a bit wasteful. Especially since this call is already made when actually toggling trim mode, which I think should suffice.

## Additional Context

https://github.com/google/web-stories-wp/blob/456348dc59a45d22783a3a587768adaaec821a90/packages/story-editor/src/components/videoTrim/useVideoTrimMode.js#L99-L103

barklund:
## Summary

I'm not sure why this effect was ever there? Maybe a kind of premature optimization or just for debugging purposes? Anyway, it doesn't seem to be required in any case (regardless of a video having an original or not).

## Testing Instructions

## Checklist

---

Fixes #10246

### DIFF
--- a/packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
+++ b/packages/story-editor/src/components/videoTrim/useVideoTrimMode.js
@@ -17,12 +17,7 @@
 /**
  * External dependencies
  */
-import {
-  useCallback,
-  useMemo,
-  useState,
-  useEffect,
-} from '@googleforcreators/react';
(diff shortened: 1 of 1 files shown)


merapi (approved):
I looked around and was also not sure what it was needed for.

Issue #10519, opened 2022-02-08
Layer selection incorrectly detects rotated elements under the cursor


merapi:
## Bug Description

~Layer selection in the right-click menu seems to detect elements under the cursor incorrectly. Please check the video.
Extracted from https://github.com/GoogleForCreators/web-stories-wp/pull/10458#discussion_r801805256~ Fixed in #11163

Note that this also happens for auto-setting background. Rotated media is sometimes set as background even if it doesn't fully cover the page.

## Expected Behaviour
- ~When using the "Select Layer" option in the right-click menu, it should only display the elements directly under the cursor and not include nearby rotated elements~ Fixed in #11163
- When rotated media is resized to be covering the Page mostly but not fully (e.g leaving the corners out), it should not be auto-set as background.

## Steps to Reproduce
1. Add a media element to a page
2. Rotate it
3. Try to cover the entire canvas to make it background image and see that it does not always trigger correctly (triggers before actually covering the full canvas)

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

felipebochehin87:
Tested against https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/, using PR #11163.

1. Added multiple rectangles (one over the other)
2. Right clicked on different parts of the canvas - where we have intersection of elements, we have displayed rectangles (the number of rectangles that are at the position that has been right clicked) on Select Layer & where we have only one rectangle, we only see one rectangle on Select Layer


https://user-images.githubusercontent.com/96055551/162016340-1eb766db-1f12-4b76-96ec-1a9841adda69.mp4

PR #11163, opened 2022-04-01
Editor: Fix intersection calculation when right-clicking elements

barklund:
## Summary

This fixes the intersection calculation done when right-clicking on the canvas and will now correctly calculate the polygons for all elements (still not taking masks/shapes into consideration though).

This is only a partial fix for #10519, as it only addresses the right-click menu calculation.

## Relevant Technical Choices

- Manually creating polygons from original element sizes and positions rather than `getBoundingClientRect`, as it was trouble for rotated elements.
- Also utilizing a lot more features of the SAT library for easier polygon creation.

## User-facing changes

Clicking just above a rotated element will no longer show the element as being below the cursor (e.g. show a list of elements below the cursor, because there's only the background there):

| Before | After |
|-|-|
| Screen Shot 2022-04-01 at 16 11 58 | Screen Shot 2022-04-01 at 16 11 32 |

## Testing Instructions

1. Add a number of (rectangular only) elements
2. Rotate, flip, and move them around so they overlap in various places
3. Right-click any position on the canvas and observe that a list of elements under the cursor only appears in the correct places and that the correct elements are in said list.

## Checklist

---

Partially addresses #10519

### DIFF
--- a/packages/story-editor/src/app/rightClickMenu/useLayerSelect.js
+++ b/packages/story-editor/src/app/rightClickMenu/useLayerSelect.js
@@ -29,12 +29,12 @@ import { Icons } from '@googleforcreators/design-system';
 import { trackEvent } from '@googleforcreators/tracking';
 import styled from 'styled-components';
 import { getDefinitionForType } from '@googleforcreators/elements';
+import { useUnits } from '@googleforcreators/units';
 
 /**
  * Internal dependencies
  */
 import useStory from '../story/useStory';
-import createPolygon from '../canvas/utils/createPolygon';
 import { useCanvas } from '../canvas';
 import { useConfig } from '../config';
 
@@ -57,6 +57,8 @@ function useLayerSelect({ menuItemProps, menuPosition, isMenuOpen }) {
     })
   );
 
+  const getBox = useUnits(({ actions: { getBox } }) => getBox);
+
   useEffect(() => {
     // Close submenu if the menu itself also closes.
     if (!isMenuOpen) {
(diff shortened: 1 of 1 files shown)


packages/story-editor/src/app/rightClickMenu/useLayerSelect.js
```diff
@@ -29,12 +29,12 @@ import { Icons } from '@googleforcreators/design-system';
 import { trackEvent } from '@googleforcreators/tracking';
 import styled from 'styled-components';
 import { getDefinitionForType } from '@googleforcreators/elements';
+import { useUnits } from '@googleforcreators/units';
 
 /**
  * Internal dependencies
  */
 import useStory from '../story/useStory';
-import createPolygon from '../canvas/utils/createPolygon';
```
barklund, line 37: So I did not use this function, as it cannot correctly calculate the box because it doesn't have all the information.

We could convert this to a hook, `useElementPolygon`, so it can access the units provider and get the relevant information from there and then use this same hook for the other part of #10519 (for the is-element-covering-entire-canvas calculation), but I wasn't sure if that was the right way to go.

@merapi, you worked on that second part. What do you think is the best way to port this work to that feature?

packages/story-editor/src/app/rightClickMenu/useLayerSelect.js
```diff
@@ -70,21 +72,55 @@ function useLayerSelect({ menuItemProps, menuPosition, isMenuOpen }) {
...
+        bgPolygon.pos.x + box.x + box.width / 2,
+        bgPolygon.pos.y + elementYOffset + box.y + box.height / 2
+      );
+      const elementBox = new SAT.Box(center, box.width, box.height);
+      const polygon = elementBox.toPolygon();
+      const offset = new SAT.Vector(-box.width / 2, -box.height / 2);
+      polygon.setOffset(offset);
+      polygon.setAngle((rotationAngle * Math.PI) / 180);
+      return { element, polygon };
+    });
```
barklund, line 115: Does this require more comments? I feel it's pretty straight-forward except for the center/offset part. The [SAT documentation is decent](https://github.com/jriecken/sat-js#satvector-aliased-as-satv), so could link to that? It would be a fragile link though.

packages/story-editor/src/app/rightClickMenu/useLayerSelect.js
```diff
@@ -70,21 +72,55 @@ function useLayerSelect({ menuItemProps, menuPosition, isMenuOpen }) {
...
+    const intersectingElements = elementsWithPolygons
```
barklund, line 116: Decided to split it up into two statements, First calculating all the polygons for the elements, and second doing the intersection tests. This simplified debugging.

merapi (approved):
Works great ✅ 

About handling the auto-set bg: some parts can be used in both places, so if you can unify them then 👌🏻 
Maybe we can do `4*pointInPolygon` check for each canvas corner for the BG math?

PR #11164, opened 2022-04-01
Dashboard: Add Default View to Router

### LINKED ISSUE #11055: Accessibility: Skip to content link does not work
## Bug Description

When using keyboard navigation, a "Skip to content" link is shown. When pressing enter, the url changes and but nothing happens. Worse in dashboard, the content is cleared. See video. 

## Screenshots

https://user-images.githubusercontent.com/237508/159733308-af0bfb75-a879-424a-b856-2bf48da1ca12.mov

https://user-images.githubusercontent.com/237508/159733286-a288b33d-2b43-4c80-966d-7cfd05c2a4d8.mov

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

BrittanyIRL:
## Context

This came up because when you skip to "main content" using `tab` when the dashboard loads as a keyboard user the Stories Dashboard doesn't actually allow the main content of the Dashboard  because the `Dashboard` has a path of `/` so it needs be an exact match otherwise it'll always show it since every other route has a path that will partially match `/`.

## Summary

Now `post_type=web-story&page=stories-dashboard#wpbody-content` will return the dashboard, as will `post_type=web-story&page=stories-dashboard#` and `post_type=web-story&page=stories-dashboard#/ajsdflkajdlkajsdfklasdf`.

## Relevant Technical Choices

Our router is an in-house solution because it is so minimal. I chose to add a new prop for `<Route>` called `isDefault`. This gave me the wiggle room necessary to have the dashboard base path of `/` be both `exact` and `isDefault` so that if it's exact it'll render and if it's not it will only render when no other routes also fit the criteria. I didn't want to mess with the base path for the entire dashboard. 

This means I also added complexity to the router provider and initial mount of the Dashboard. Now we track the available routes and tell the provider what the default path is so that we can use it when `<Route>` calls for it. 

## To-do

N/A

## User-facing changes

(note the `&page=stories-dashboard#wpbody-content` in the url path)

| Before | After | 
| --- | --- | 
| Screen Shot 2022-04-01 at 4 17 42 PM |  Screen Shot 2022-04-01 at 4 17 13 PM |

## Testing Instructions

Head to the Stories Dashboard and verify that Dashboard loads by default, that "Explore Templates" and "Template Details" still work as expected. 
Now use the keyboard to navigate to "Main Content" skip button (tab through WP admin toolbar, if you just loaded the page it'll be the first tab). Hit "Enter" on "Main Content" and you should still see the dashboard. 

Type in some wrong paths off of the dashboard base path `&page=stories-dashboard` like `&page=stories-dashboard#/jalkfjalkjflakdjflk` and see that the dashboard default view is still showing. 

Also test a user that has no stories yet. What happens there is on the first page load the user is sent to "Explore Templates", the user interacting with "Main Content" for keyboard user is no longer the initial load so it'll actually bring the user to Dashboard Stories instead. 

No

No

No

## Checklist

---

Fixes #11055

### DIFF
--- a/packages/dashboard/src/app/router/route.js
+++ b/packages/dashboard/src/app/router/route.js
@@ -89,30 +89,25 @@ export function matchPath({
   // If there is and the current path fits 1 or less
   // of the available routes then render it as default
   if (isDefault) {
-    const availableRoutesPresentInCurrentPath = availableRoutes.filter(
-      (route) => currentPath.startsWith(route)
-    );
-    if (availableRoutesPresentInCurrentPath.length <= 1) {
+    const isValidDefaultRoute =
+      availableRoutes.filter((route) => currentPath.startsWith(route)).length <=
+      1;
+
+    if (isValidDefaultRoute) {
       return defaultRoute;
     }
   }
 
   const isExactMatch = currentPath === matchUrl;
-
   if (exact && !isExactMatch) {
     return null;
   }
-
   return matchUrl;
 }
 
 function Route({ component, path, exact = false, isDefault = false }) {
   const { availableRoutes, currentPath, defaultRoute } = useRouteHistory(
-    ({ state: { availableRoutes, currentPath, defaultRoute } }) => ({
-      availableRoutes,
-      currentPath,
-      defaultRoute,
-    })
+    ({ state }) => state
   );
 
   const match = matchPath({
--- a/packages/dashboard/src/app/router/route.js
+++ b/packages/dashboard/src/app/router/route.js
@@ -67,22 +67,63 @@ export function resolveRoute(route) {
   }
 }
 
-export function matchPath(currentPath, path, exact = false) {
+export function matchPath({
+  currentPath,
+  path,
+  availableRoutes,
+  defaultRoute,
+  isDefault = false,
+  exact = false,
+}) {
   const match = new RegExp(`^${path}`).exec(currentPath);
-  if (!match) {
+
+  // allow paths through that don't match if they are default, not narrowed out yet.
+  if (!match && !isDefault) {
     return null;
   }
-  const matchUrl = match[0];
+
+  // Find all available url matches
+  const matchUrl = match?.[0];
+
+  // Check if there is a default before looking for more exact matches
+  // If there is and the current path fits 1 or less
+  // of the available routes then render it as default
+  if (isDefault) {
+    const availableRoutesPresentInCurrentPath = availableRoutes.filter(
+      (route) => currentPath.startsWith(route)
+    );
+    if (availableRoutesPresentInCurrentPath.length <= 1) {
+      return defaultRoute;
+    }
+  }
+
   const isExactMatch = currentPath === matchUrl;
+
   if (exact && !isExactMatch) {
     return null;
   }
+
   return matchUrl;
 }
 
-function Route({ component, path, exact }) {
-  const { state } = useRouteHistory();
-  const match = matchPath(state.currentPath, path, exact);
+function Route({ component, path, exact = false, isDefault = false }) {
+  const { availableRoutes, currentPath, defaultRoute } = useRouteHistory(
+    ({ state: { availableRoutes, currentPath, defaultRoute } }) => ({
+      availableRoutes,
+      currentPath,
+      defaultRoute,
+    })
+  );
+
+  const match = matchPath({
+    currentPath,
+    availableRoutes,
+    defaultRoute,
+    path,
+    exact,
+    isDefault,
+  });
+
   if (!match) {
     return null;
   }
--- a/packages/dashboard/src/app/router/routerProvider.js
+++ b/packages/dashboard/src/app/router/routerProvider.js
@@ -26,6 +26,10 @@ import {
 } from '@googleforcreators/react';
 import PropTypes from 'prop-types';
 import { createHashHistory } from 'history';
+/**
+ * Internal dependencies
+ */
+import { APP_ROUTES } from '../../constants';
 
 export const RouterContext = createContext({ state: {}, actions: {} });
 
@@ -46,7 +50,7 @@ function RouterProvider({ children, ...props }) {
     history.current.location.pathname
   );
   const [availableRoutes, setAvailableRoutes] = useState([]);
-  const [defaultRoute, setDefaultRoute] = useState();
+  const defaultRoute = APP_ROUTES.DASHBOARD;
 
   const activeRoute = useMemo(
     () => getActiveRoute({ availableRoutes, currentPath, defaultRoute }),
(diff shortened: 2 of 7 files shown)


packages/e2e-tests/src/specs/dashboard/dashboard.js
```diff
@@ -42,6 +42,27 @@ describe('Stories Dashboard', () => {
     await takeSnapshot(page, 'Stories Dashboard', { percyCSS });
   });
 
+  it('should be able to skip to main content of Dashboard', async () => {
+    await visitDashboard();
+
+    // If there are no existing stories, the app goes to the templates page instead.
+    // Account for both here, but then force-visit the dashboard.
```
swissspidy: To account for the "force-visit" part this is missing the

```js
await expect(page).toClick('[aria-label="Main dashboard navigation"] a', {
  text: 'Dashboard',
});
```

part

packages/e2e-tests/src/specs/dashboard/dashboard.js
```diff
@@ -42,6 +42,27 @@ describe('Stories Dashboard', () => {
...
+    // When navigating to Dashboard, immediately use keyboard to
+    // tab to WordPress shortcut of "Main Content"
+    page.keyboard.press('Tab');
+    page.keyboard.press('Enter');
+
+    await expect(page).toMatchElement('h2', {
+      text: /(Dashboard|Explore Templates)/,
+    });
+
+    await takeSnapshot(page, 'Stories Dashboard on Main Content', { percyCSS });
```
swissspidy: A detail, but could we perhaps also test whether the skip link is visible and that the next Tab goes straight to the "Create New Story" link

maxyinger (approved):
nice catch

packages/e2e-tests/src/specs/dashboard/dashboard.js
BrittanyIRL: Thanks! Realized I actually don't want to force anything and we just want to see what happens with the navigation by keyboard once we're in the Stories Dashboard. Fixed this up and added better comments.

packages/dashboard/src/app/router/routerProvider.js
```diff
@@ -34,6 +34,8 @@ function RouterProvider({ children, ...props }) {
   const [currentPath, setCurrentPath] = useState(
     history.current.location.pathname
   );
+  const [availableRoutes, setAvailableRoutes] = useState([]);
+  const [defaultRoute, setDefaultRoute] = useState();
```
samwhale: 💭 This is always `APP_ROUTES.DASHBOARD`, so I don't think the state is needed here. This could be passed into `matchParams` as `defaultRoute`

But lmk if I'm missing why this may need to be different

packages/dashboard/src/components/interfaceSkeleton/index.js
```diff
@@ -87,6 +89,23 @@ const InterfaceSkeleton = ({ additionalRoutes }) => {
     });
   }, [addInitialFetchListener, push, currentPath, canViewDefaultTemplates]);
 
+  // Only set the available routes & default route on initial mount
```
samwhale, line 92: 💭 if it's only being set on initial mount, could this be set using a state hook rather than an effect?

Does this need to happen in an effect?

packages/dashboard/src/app/router/route.js
```diff
@@ -67,22 +67,58 @@ export function resolveRoute(route) {
...
+  availableRoutes,
+  defaultRoute,
+  isDefault = false,
+  exact = false,
+}) {
   const match = new RegExp(`^${path}`).exec(currentPath);
-  if (!match) {
+
+  // allow paths through that don't match if they are default, not narrowed out yet.
+  if (!match && !isDefault) {
```
samwhale, line 81: So a route may not match but it could be the default?

packages/dashboard/src/app/router/route.js
```diff
@@ -67,22 +67,58 @@ export function resolveRoute(route) {
...
+
+  // Find all available url matches
+  const matchUrl = match?.[0];
+
+  // Check if there is a default before looking for more exact matches
+  // If there is and the current path fits 1 or less
+  // of the available routes then render it as default
+  if (isDefault) {
+    const isValidDefaultRoute =
+      availableRoutes.filter((route) => currentPath.startsWith(route)).length <=
```
samwhale, line 93: 💭 if there's no match, should this show a 404 page? I guess it's weird because it's a nested app in the wordpress app.

Maybe going to an unregistered route should reset the url to the current route?

like going to `http://localhost:8899/wp-admin/edit.php?post_type=web-story&page=stories-dashboard#/not-a-place` should route to `http://localhost:8899/wp-admin/edit.php?post_type=web-story&page=stories-dashboard#/`?

samwhale (approved):
Works great for me! Some small questions

BrookeGraham (approved):

packages/dashboard/src/components/interfaceSkeleton/index.js
maxyinger: I'm not sure if I'm 100% following. could you elaborate on the state hook approach?

I think another approach might be making `availableRoutes` a memoized value here and adding an `availableRoutes` prop to `Route`:
```js
const availableRoutes = useMemo(() => {
  // ...
  return [...Object.values(APP_ROUTES), ...additionalPaths];
}, [additionalPaths]);

...
<Route
  availableRoutes={availableRoutes}
    // ... whatever other props
/>
```

maxyinger (approved):

packages/dashboard/src/components/interfaceSkeleton/index.js
samwhale: Oops yeah sorry, I had morning brain when I wrote this.

I now see why we need to be able to set this. `additionalPaths` may change.

Either way is fine with me, sorry!

packages/dashboard/src/app/router/route.js
BrittanyIRL: yup! so `stories-dashboard#wpbody-content` comes back as `match = null` so `!match` is `true` - the thing is, that I want to find out if there's no match _and_ if it's default we want to check and see if the currentPath starts with one of our available routes. It means that there's no chance that this currentPath would be represented by another available path that would render instead.

packages/dashboard/src/app/router/route.js
BrittanyIRL: I suppose that's an option, yes. Both work the same though. As soon as you navigate from some bogus route that's now defaulted to the Dashboard stories view you'll end up in a proper route. We handle routes this same way for detail template views so my current implementation is consistent.

PR #10531, opened 2022-02-09
Blocks/Renderer: Ensure images have `srcset` and `sizes` attributes

spacedmonkey:
## Summary

Add `srcset` and `sizes` attributes to all images that are rendered in web stories. 

## Testing Instructions

- [x] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1.

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/googleforcreators/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [x] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/googleforcreators/web-stories-wp/tree/main/docs#testing))
- [x] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #7983

PR #11191, opened 2022-04-05
Embeds: avoid empty `srcset` and `sizes` attributes

swissspidy:
## Context

@fellyph reported to me an issue with the Web Stories block and AMP validation.

AMP does not allow empty `srcset=""` attributes.

Introduced in #10531

## Summary

This PR prevents AMP validation errors due to such empty `srcset` and `sizes` attributes.

## Relevant Technical Choices

Changed default properties of the `Story` model to use empty strings instead of `null`, to avoid having to check against `null` everywhere and to prevent the need for type casting.

## To-do

None

## User-facing changes

N/A

## Testing Instructions

1. Embed a single story using the Web Stories block
2. Preview the post
3. Verify that the story image does not have empty `srcset=""` and `sizes=""` attributes

No

No

No

## Checklist

---

Fixes #

### DIFF
--- a/includes/Model/Story.php
+++ b/includes/Model/Story.php
@@ -109,21 +109,21 @@ class Story {
 	 *
 	 * @var string
 	 */
-	protected $poster_portrait;
+	protected $poster_portrait = '';
 
 	/**
 	 * Date for the story.
 	 *
 	 * @var string
 	 */
-	protected $date;
+	protected $date = '';
 
 	/**
 	 * Author of story.
 	 *
 	 * @var string
 	 */
-	protected $author;
+	protected $author = '';
 
 	/**
 	 * Story constructor.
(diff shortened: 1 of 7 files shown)


includes/Model/Story.php
```diff
@@ -230,14 +230,14 @@ public function get_poster_srcset(): string {
 	 * @since 1.0.0
 	 */
 	public function get_title(): string {
-		return (string) $this->title;
+		return $this->title;
```
spacedmonkey, line 233: The reason these were put into first place, is that if a developer sets this to false by mistake, ths would give a type error. Is there a worry this might happen? What does it hurt, ensuring these are the correct type?

spacedmonkey, line 233: The reason these were put into first place, is that if a developer sets this to false by mistake, ths would give a type error. Is there a worry this might happen? What does it hurt, ensuring these are the correct type?

spacedmonkey (approved):

includes/Model/Story.php
swissspidy: The only way these could be set is through the `load_from_post()` method, where title is always a string. Other fields like the `url` are converted to strings in `load_from_post()`.

PR #11192, opened 2022-04-05
Floating Menus: adjust input widths

### LINKED ISSUE #10799: Tweak Floating Menu Styles
## Feature Description
Let's clean up some styling tweaks on `Floating Menu`  once more components are in place. See [comment](https://github.com/GoogleForCreators/web-stories-wp/pull/10786#discussion_r819085047)

- [ ] make font input width smaller than ./shared `Input`, but larger than standard `NumericInput`
- [ ] check with @agingoldseco about the opacity input unit.

## Additional Context

The [Figma Designs](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=11394%3A457172) show the input with no unit, but the opacity input in the original design panel has `%` unit.

| without | with | 
| --- | --- | 
| Screen Shot 2022-03-10 at 5 24 28 PM | Screen Shot 2022-03-10 at 5 23 15 PM |

BrookeGraham:
## Context
All of the numeric inputs were sharing styles, this was causing some inputs to be wider than necessary. Design also requested that the % unit be added to opacity and the `More` button have a larger font and smaller padding.  

## User-facing changes
||NEW|OLD|
|--|--|--|
|video|video|video old|
|sticker|sticker|sticker old|
|text|text|text old|
|image|image|image old|
|shape|shape|shape old|

## Testing Instructions

1. Open each different type of floating menu: image, text, shape, sticker, multiple, video. 
2. See the updated inputs.
3. `Font size`, `Opacity`, and `More`

no

no

no

## Checklist

---

Fixes #10799

### DIFF
--- a/packages/story-editor/src/components/floatingMenu/elements/layerOpacity.js
+++ b/packages/story-editor/src/components/floatingMenu/elements/layerOpacity.js
@@ -17,8 +17,9 @@
 /**
  * External dependencies
  */
-import { Icons } from '@googleforcreators/design-system';
-import { __ } from '@googleforcreators/i18n';
+import styled from 'styled-components';
+import { __, _x } from '@googleforcreators/i18n';
+import { Icons, NumericInput } from '@googleforcreators/design-system';
 import { trackEvent } from '@googleforcreators/tracking';
 
 /**
@@ -26,7 +27,12 @@ import { trackEvent } from '@googleforcreators/tracking';
  */
 import { useStory } from '../../../app';
 import { MIN_MAX } from '../../panels/design/sizePosition/opacity';
-import { Input, useProperties } from './shared';
+import { useProperties } from './shared';
+
+const Input = styled(NumericInput)`
+  width: 82px;
+  flex: 0 0 82px;
+`;
 
 function LayerOpacity() {
   const { opacity, type } = useProperties(['opacity', 'type']);
@@ -50,6 +56,7 @@ function LayerOpacity() {
   return (
     <Input
       suffix={<Icons.ColorDrop />}
+      unit={_x('%', 'Percentage', 'web-stories')}
       value={opacity || 0}
       aria-label={__('Opacity in percent', 'web-stories')}
       onChange={handleOpacityChange}
--- a/packages/story-editor/src/components/floatingMenu/elements/shared/text.js
+++ b/packages/story-editor/src/components/floatingMenu/elements/shared/text.js
@@ -25,6 +25,8 @@ import { ContextMenuComponents } from '@googleforcreators/design-system';
 const Button = styled(ContextMenuComponents.MenuButton)`
   font-size: 14px;
   padding: 2px 12px;
+  font-weight: normal;
+  letter-spacing: normal;
 `;
 
 const TextButton = memo(function TextButton({ text, ...rest }) {
--- a/packages/story-editor/src/components/floatingMenu/elements/shared/text.js
+++ b/packages/story-editor/src/components/floatingMenu/elements/shared/text.js
@@ -19,13 +19,19 @@
  */
 import PropTypes from 'prop-types';
 import { memo } from '@googleforcreators/react';
+import styled from 'styled-components';
 import { ContextMenuComponents } from '@googleforcreators/design-system';
+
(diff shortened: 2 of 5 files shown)


packages/story-editor/src/components/floatingMenu/elements/layerOpacity.js
```diff
@@ -50,6 +56,7 @@ function LayerOpacity() {
   return (
     <Input
       suffix={<Icons.ColorDrop />}
+      unit={_x('%', 'Percentage', 'web-stories')}
```
BrittanyIRL, line 59: yay!

packages/story-editor/src/components/floatingMenu/elements/shared/text.js
```diff
@@ -19,13 +19,19 @@
  */
 import PropTypes from 'prop-types';
 import { memo } from '@googleforcreators/react';
+import styled from 'styled-components';
 import { ContextMenuComponents } from '@googleforcreators/design-system';
 
+const Button = styled(ContextMenuComponents.MenuButton)`
+  font-size: 14px;
```
BrittanyIRL, line 26: ah, this makes me sad because we've done so good for so long not having one off font size updates. 

I think this context menu button is based on the default button so you could pass in `size: BUTTON_SIZES.SMALL` as an attribute here and get the same impact with the added bonus of keeping line height consistent. 

`const Button = styled(ContextMenuComponents.MenuButton).attrs({size: BUTTON_SIZES.SMALL})`

Screen Shot 2022-04-05 at 5 02 35 PM

BrittanyIRL (approved):
Having units here is a big win, thanks for the back and forth with Amy to figure out what all needed specific updates. 

one nit about just using the prebaked size in the button and one question but I'm not a designer so I trust you've had these conversations. 

I get that these are actually aligned well but the bolded "more" is throwing me off. Is this intentional? Design was good with this? 
Screen Shot 2022-04-05 at 5 09 42 PM

packages/story-editor/src/components/floatingMenu/elements/shared/text.js
BrookeGraham: I was a bit confused here too. Cause `BUTTON_SIZE.SMALL` is not the same as `PRESET_SIZES.SMALL`. The `PRESET_SIZES` under typography has `label` listed as font size 14 but the small button was getting size 12. But upon further investigation it looks like the `ContextMenuComponent.MenuButton` is actually using paragraph, which makes sense since the text here isn't a label. in `contextMenu -> components -> styles -> menuItemStyles`
```themeHelpers.expandPresetStyles({
      preset:
        theme.typography.presets.paragraph[
          THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.X_SMALL
        ],
      theme,
```

That being said, even if I try to update the `attrs` for `ContextMenuComponent.MenuButton` to `MEDIUM` it doesn't get overridden. And it didn't seem like the right move to try and mess with context menu styling, so I am not sure _how_ I can use any of the theming here 🤷🏻‍♀️

BrookeGraham:
>bolded "more" is throwing me off. Is this intentional?

It was not. I didn't add any font weight so I am not sure why it would have gone from 400 to 500, which is what it is doing now. When I screen shoot with Amy she said it looked good, but this does seem like an oversight. So I will get it back to 400 and check in with her.

BrookeGraham:
Updated weight!
Screen Shot 2022-04-06 at 9 30 01 AM

packages/story-editor/src/components/floatingMenu/elements/shared/text.js
BrittanyIRL: Sad. Suppose this was an oversight that just got missed somewhere along earlier. No reason to sink more time into it here with your update works just fine. Thank you for trying :)

BrookeGraham: Yeah I think it comes back to the using context menu as the underlaying structure for the floating menu that makes things a bit more difficult.

samwhale (approved):
Dope!

Issue #11000, opened 2022-03-21
Focusable Canvas Elements


BrittanyIRL:
## Feature Description

All canvas elements should be focused on tab. Right now, this only works for elements without masks.

## Additional Context

Historical Context: 

PR: #5968 - https://github.com/GoogleForCreators/web-stories-wp/pull/5968
Issue: #2038

## Acceptance Criteria

- All elements focusable regardless of masking
- Karma tests reflecting the above

BrookeGraham:
Hi @felipebochehin87 

For testing we will be using keyboard navigation and we'll also want to make sure that a screen reader announces how to enter, exit and navigate within the canvas grouping. 

1. Add one of every type of element to the canvas
2. Tab in the canvas until you reach the focus container, and outline should appear just below the editor's header (just below the title and publish buttons). You should hear a screen reader tell you to how to enter the canvas. 
3. Press Enter to "get into" the canvas layer, you should hear a screen reader tell you how to navigate while inside the canvas, as well as tell you how to exit it. 
4. Press Tab or Shift + Tab. Each time you should now be focused on a new element and it should become 'selected' in the editor. All types of elements should be focusable, including the background. 
5. Press `escape` to leave the canvas and resume being able to nav to `previous button` etc.

PR #11193, opened 2022-04-05
Canvas: Canvas as a focusable group

### LINKED ISSUE #11001: Canvas as a Focusable Group
## Feature Description

We can drastically improve keyboard users’ lives by making the canvas container a focusable group - just like we do templates, text sets, and the dashboard grid. This means that you’d tab to the canvas and see the focus outline. Then you hit ‘enter’ to go into the canvas and tab through elements. You can tab all you want and your focus will stay inside of the canvas. Hitting ‘esc’ will shift you back to the canvas container, then tabbing from there will bring you to the next element after the canvas.  This means users don’t have to tab through their entire canvas to get to the footer anymore or accidentally tab out of the canvas. 

## Acceptance Criteria

- Canvas container has a focusable outline and necessary aria 
- ‘enter’ on the canvas container gives keyboard nav access to canvas 
- ‘esc’ moves keyboard focus to the canvas container. 
- ‘tabbing’ inside of the canvas while focus is active moves you around elements. 
- Karma tests reflecting the above

samwhale:
## Context

Keyboard accessibility

## Summary

Add a focusable group around the canvas. This removes the canvas and the elements from the normal keyboard flow, but a user may enter into it to navigate it with keyboard.

Thanks to @BrookeGraham for pairing with me on this!

## Relevant Technical Choices

1. #11175 implemented focus group architecture. Used that for focusing the canvas vs focusing the editor
2. Added a `FocusContainer` rather than using the existing `<section />`. The reason being that the focus border was getting covered by the header. The new `FocusContainer` shows the entire focus border without having to worry about the header bar.

## To-do

n/a

## User-facing changes

image

## Testing Instructions

1. Go to the editor
3. Tab over to the canvas. Should see a new container focused
4. Tabbing from there should take you to the `Previous Page` button
5. While focused on the canvas, pressing `Enter` will move focus to the background element (won't see it selected until #11000 is implemented).
6. Should be able to tab around cyclically (forward and back)

no

no

no

## Checklist

---

Fixes #11001


packages/story-editor/src/components/canvas/frameElement.js
```diff
@@ -236,11 +240,10 @@ function FrameElement({ id }) {
         />
       )}
       <Wrapper
-        ref={elementRef}
+        ref={combinedFocusGroupRef}
         data-element-id={id}
         {...box}
-        // eslint-disable-next-line styled-components-a11y/no-noninteractive-tabindex -- Needed for being able to focus on the selected element on canvas, e.g. for entering edit mode.
-        tabIndex={0}
```
samwhale, line 243: Now frame elements are part of a focus group. They no longer need a tab index of 0.

packages/story-editor/src/components/canvas/framesLayer.js
```diff
@@ -163,9 +186,36 @@ function FrameElements() {
...
+      <FocusContainer
+        id={canvasId}
+        data-testid="canvas-focus-container"
+        ref={canvasRef}
+        aria-label={FOCUS_CONTAINER_MESSAGE}
+        // eslint-disable-next-line styled-components-a11y/no-noninteractive-tabindex -- Container used to separate elements from normal tab order.
+        tabIndex={0}
+      />
+      {/* Hide canvas from normal tab flow. Use focus group to focus canvas */}
+      <FrameElements role="main" aria-labelledby={canvasId} />
```
samwhale: 🤔  Seems like the main content area of the editor is the canvas - so the [`main` role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/main_role) felt like the correct role to put here.

Thoughts? [`region` seemed to vague](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/region_role), and [`application` seemed too complex](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/application_role) since it tells screen readers that "no traditional HTML interpretation techniques should be used"

packages/story-editor/src/components/canvas/editLayerFocusManager/constants.js
```diff
@@ -15,6 +15,7 @@
  */
 const FOCUS_GROUPS = {
   EDIT_ELEMENT: 'EDIT_ELEMENT',
+  EDIT_CANVAS: 'EDIT_CANVAS',
```
BrittanyIRL: nit: just `CANVAS`? I think `EDIT_ELEMENT` was because it's the actively being edited element.

packages/story-editor/src/components/canvas/framesLayer.js
BrittanyIRL: hmmm. #wp-body has the main role, so this would nest a main role in a main role. the display and frames layers are regions - those both live inside the canvas. 

I vote this is a region. 

Also, having this be a `main` causes aXe to get angry about multiples: https://github.com/GoogleForCreators/web-stories-wp/runs/5858290694?check_suite_focus=true#step:9:575

packages/story-editor/src/components/canvas/karma/canvasKeyboardNavigation.karma.js
```diff
@@ -0,0 +1,147 @@
...
+
+  it('should not focus the canvas while tabbing through the editor', async () => {
+    const focusContainer = fixture.screen.getByTestId('canvas-focus-container');
+
+    // start focused on media pane searchbar
+    await fixture.events.focus(fixture.editor.library.media.searchBar);
+
+    // tab until focus reaches the canvas container
+    let count = 0;
+    while (count < 50) {
```
BrittanyIRL: just curious, how'd you land on 50?

BrittanyIRL (approved):
💯 💯  Well this works great. Soon as the aXe issue is resolved I'm good with this. One nit and one question otherwise.

packages/story-editor/src/components/canvas/karma/canvasKeyboardNavigation.karma.js
samwhale: Picked a number that felt large but not too large - but 50 is a very large overshoot. It only takes 4 tabs to get to the focus container area, so we could always bring this down. Just wanted to have an upper bound on this loop

maybe 10 or 15 is good :)

packages/story-editor/src/components/canvas/editLayerFocusManager/constants.js
maxyinger: That or `ELEMENT_SELECTION` because eventually we'll have a `PAGE_SELECTION` group I think?

packages/story-editor/src/components/canvas/frameElement.js
```diff
@@ -178,6 +190,16 @@ function FrameElement({ id }) {
...
 
+  /**
+   * Announce keyboard options on element.
+   *
+   * Using a live region because an `aria-label` would remove
+   * any labels/content that would be read from children.
+   */
+  const handleFocus = useCallback(() => {
+    speak(FRAME_ELEMENT_MESSAGE);
+  }, [speak]);
```
maxyinger, line 201: Ahh thanks for doing this. I should reference this and do the same in the floating menu PR 🙌

packages/story-editor/src/components/canvas/frameElement.js
```diff
@@ -178,6 +190,16 @@ function FrameElement({ id }) {
   );
   const handleMediaClick = useDoubleClick(NOOP, handleMediaDoubleClick);
 
+  /**
+   * Announce keyboard options on element.
+   *
+   * Using a live region because an `aria-label` would remove
+   * any labels/content that would be read from children.
```
BrookeGraham, line 197: Nice comment.

maxyinger (approved):
Excellante.

packages/story-editor/src/components/canvas/framesLayer.js
```diff
@@ -152,6 +173,7 @@ function FrameElements() {
         )}
         onContextMenu={onOpenMenu}
         onScroll={onScroll}
+        {...props}
```
BrookeGraham: 💭  Did we end up needing `props` ?  I didn't see anything being sent in?

packages/story-editor/src/components/canvas/karma/canvasKeyboardNavigation.karma.js
```diff
@@ -0,0 +1,147 @@
...
+      if (document.activeElement === focusContainer) {
+        break;
+      }
+
+      count++;
+    }
+
+    if (count >= 15) {
+      throw new Error('Could not find focus container.');
+    }
```
BrookeGraham: Nit: We never pulled this into shared func. But I can do that with #11000

packages/story-editor/src/components/canvas/framesLayer.js
```diff
@@ -41,12 +44,30 @@ import { Layer, NavNextArea, NavPrevArea, PageArea } from './layout';
...
+} from './editLayerFocusManager';
 
 const FramesPageArea = styled(PageArea)`
   pointer-events: initial;
 `;
 const marginRatio = 100 * (DESIGN_SPACE_MARGIN / PAGE_WIDTH);
 
+const FocusContainer = styled.div`
+  grid-row: 2 / -1;
+  grid-column: 1 / -1;
```
BrookeGraham, line 60: nit: a comment here about the header could be helpful for our future selves, but totally not necessary

BrookeGraham (approved):
🌈

packages/story-editor/src/components/canvas/framesLayer.js
samwhale: Yeah, we don't need this here. I can remove this

packages/story-editor/src/components/canvas/karma/canvasKeyboardNavigation.karma.js
samwhale: Oh good catch! I'll do that real quick

samwhale:
> Oooh. Just noticed that tabbing from 3p media pane, and text pane results in one extra tab needed. Not sure where the focus is going.

So this gets hung up on an anchor tag somewhere in that pane. For now I think that's out of scope, but I did fix the problem where the `FramesLayer` was getting focused instead of the focus container.

PR #11197, opened 2022-04-05
Code Quality: Clean up Checkpoint Provider

### LINKED ISSUE #10470: Code Quality: Clean up Checkpoint Provider Access
## Feature Description

When #10115 is wrapped up and no longer behind a feature flag we'll be able to gut a chunk of the checklist context. The new story details modal that is part of 10115 is reusing the reviewChecklist action in the checkpoint hook, that can get cleaned up too. Other areas that can get cleaned up now: Publish button - doesn't need to know status of any checks anymore.

BrittanyIRL:
## Context

When the prepublish checklist was initially done, there was a dialog that showed up asking if users were sure they were ready to publish. This dialog would show up based on if there were high priority issues present in their story or if they had yet to unlock the various checkpoints of it. 

Now we have a new modal that is seen when "publish" is clicked from the editor no matter what so this little dialog is unnecessary anymore. Most of the logic in the provider that was controlling when to have that dialog show up is actually still important to controlling the automatic opening of the checklist conditionally as well as the little icon and tooltip text for the publish buttons. So this is largely just renaming and doing some minor clean up. 

When I was first implementing the new modal I thought more of the checkpoint provider would get destroyed. 

## Summary

Renamed everything related to `reviewDialog` and made it more so about what was happening up stream that is influencing the flow - so `reviewChecklist`  or `hasHighPriorityIssues` - things that are easier to also have elsewhere should we need them without their naming losing meaning. 

## Relevant Technical Choices

Aside from some renaming, the only thing that's changing here is the `private` publish flow. This was the only place that the review checklist dialog would show up. I checked on how posts work with the private flow,  my update here has Story private publish working that same way. With the review checklist dialog you'd 1) select 'private' from the dropdown for visibility 2) get the default WP alert asking if you're sure, assuming you say yes 3) the review checklist dialog. Now you just get the first two steps. 

I did this because if you're in the story details modal and you hit 'private' that's already 2 things popped up in your browser - the story details modal and the new alert asking if you're sure. It just feels excessive. I talked a while ago with Amy about it and this is just killing it officially. 

If this isn't ok for some reason I don't know, we can put it back. 

## To-do

I found a bug about focus while I was working on this: https://github.com/GoogleForCreators/web-stories-wp/issues/11196

## User-facing changes

If I did this right, then there are no user facing changes. 

## Testing Instructions

Please make sure that the publish flow is still working as expected for admins and contributors.
 As well as:  publishing with different visibility and going from story details modal to review checklist.

1.

No

No

No

## Checklist

---

Fixes #10470


packages/wp-story-editor/src/components/documentPane/status/status.js
```diff
@@ -242,16 +241,7 @@ function StatusPanel({
...
-          if (
-            !isUpdatedPublishModalEnabled &&
-            shouldReviewDialogBeSeen &&
-            !isAlreadyPublished
-          ) {
-            setShowReviewDialog(true);
-            properties.visibility = VISIBILITY.PRIVATE;
-          } else {
-            publishPrivately();
-          }
```
BrittanyIRL, line 254: @swissspidy  tagged you on this PR for this bit here, if anyone has intel on keeping this it's going to be you :) 

(copied from my PR description) 
```
Aside from some renaming, the only thing that's changing in this PR is the private publish flow. This was the only place that the review checklist dialog would show up. I checked on how posts work with the private flow, my update here has Story private publish working that same way. With the review checklist dialog you'd 1) select 'private' from the dropdown for visibility 2) get the default WP alert asking if you're sure, assuming you say yes 3) the review checklist dialog. Now you just get the first two steps.

I did this because if you're in the story details modal and you hit 'private' that's already 2 things popped up in your browser - the story details modal and the new alert asking if you're sure. It just feels excessive. I talked a while ago with Amy about it and this is just killing it officially.

If this isn't ok for some reason I don't know, we can put it back.
```

swissspidy: Thanks for the ping!

This sounds reasonable to me.

Private stories are only visible for logged-in users. The review checklist/dialog is mostly relevant for public content though as we want to ensure eligibility for appearance on Search and Discover.

The one edge case here is when you want to make an existing _private_ story _public_. And in that case we show the dialog again, so that's all good!

swissspidy (approved):

packages/story-editor/src/components/checklist/checkpointContext/checkpointContext.js
```diff
@@ -71,14 +71,9 @@ function ChecklistCheckpointProvider({ children }) {
   );
 
   const [highPriorityCount, setHighPriorityCount] = useState();
-  const [reviewDialogRequested, setReviewDialogRequested] = useState();
+  const [reviewChecklistRequested, setReviewChecklistRequested] = useState();
```
samwhale, line 74: Nit: feels like this name is a little confusing to me. This keeps track of if the checklist should show the contents, right? Maybe to make it a little easier to follow it could be named to something like `showAllChecklistItems` or something 🤔 

But not a big deal. Works well as it is.

samwhale (approved):

BrookeGraham (approved):
🌅  nice.

maxyinger (approved):
less code is the best code  🙌

packages/story-editor/src/components/checklist/checkpointContext/checkpointContext.js
BrittanyIRL: It actually tracks if the checklist should popup when it's requested from outside of the toggle button. At the point that this is true the contents of the popup will always be visible.

PR #11205, opened 2022-04-06
Font Check: Automatically update font metrics

### LINKED ISSUE #11147: Font Check: Automatically update font metrics
## Feature Description

I think this has been discussed before but I couldn't find a ticket for it anymore, hence creating this new one.

The current `FontCheck` component in 1.19.0 is super nice. It automatically detects if a story contains a missing font and then allows you to replace it.

Fonts can be missing if it's either a custom font that was deleted or one that was pulled from Google Fonts.

More often than not, fonts on Google Fonts simply get _updated_, not deleted. For example if a font is now available in a new font weight or the metrics have slightly changed.

#10734 is a good example of such an update.

It would be really nice to update `FontCheck` so that it goes through all fonts in a story and then updates them if we have newer information on file.

In order to not cause history updates in the editor for these font updates, we can update `ELEMENT_PROPS_TO_IGNORE` in  `useHistoryEntry.js` accordingly.

timarney:
## Context

When Google Fonts get updated (i.e. #10734 ) we need a way to update "in use" font metrics weights etc...

## Summary

Updates the Font check code from #10808 to handle font updates vs only looking for missing fonts.

This also adds properties to  useHistoryEntry.js -> `ELEMENT_PROPS_TO_IGNORE` to avoid history updates.

## User-facing changes

In use fonts will receive updated properties

## Testing Instructions

**Visual Testing** 

1. Add a story and some text on the screen 
2. Use the [editor dev tools](https://github.com/GoogleForCreators/web-stories-wp/blob/main/docs/devtools.md) to adjust the text element's font information (e.g. change value of `fallbacks` or `metrics` or `weights`) and click on "Load"
3. Afterwards, save the story and refresh the page
4. Open the dev tools again and verify that the font information has been reset to the values from _before_
5. Verify that the undo and redo button are disabled (aka history has not been updated)

No

No

No

## Checklist

---

Fixes #11147


packages/wp-story-editor/src/components/fontCheck/index.js
```diff
@@ -34,7 +34,11 @@ export const FontCheck = () => {
   const { isStoryLoaded } = useStory(({ state: { pages } }) => ({
     isStoryLoaded: pages.length > 0,
   }));
+  const [fontsChecked, setFontsChecked] = useState(false);
   const storyPages = useStory(({ state: { pages } }) => pages);
+  const { updateElementsByFontFamily } = useStory(({ actions }) => ({
```
swissspidy: Why are these separate and not combined with the `useStory()` selector above?

packages/wp-story-editor/src/components/fontCheck/index.js
```diff
@@ -58,6 +62,22 @@ export const FontCheck = () => {
       const inUseFonts = getInUseFontsForPages(storyPages);
       const allFonts = await getFonts({ include: inUseFonts.join(',') });
 
+      if (!fontsChecked) {
```
swissspidy: Hmm why was this new state variable needed? Why did it fire too often before?

Can we accomplish the same by changing the deps array below to `[isStoryLoaded]`\*? (and ignore lint warnings)

\* or maybe to `[isStoryLoaded, updateElementsByFontFamily]`

packages/story-editor/src/app/story/effects/useHistoryEntry.js
```diff
@@ -36,6 +36,9 @@ const ELEMENT_PROPS_TO_IGNORE = [
   'resource.isMuted',
   'resource.posterId',
   'resource.poster',
+  'resource.font.metrics',
+  'resource.font.weights',
+  'resource.font.variants',
```
swissspidy, line 41: What about `font.fallbacks` and `font.styles`?

swissspidy: #10819 and #10408 are examples where a font's fallbacks have changed

swissspidy: #10031 is an example where styles and variants have changed

swissspidy:
Regarding testing:

Might be easiest to test this be using [the secret editor devtools](https://github.com/GoogleForCreators/web-stories-wp/blob/main/docs/devtools.md) to update an element's font information, save the story, and then reload the page.

As for the history updates, the goal would be for the undo/redo buttons to still be _disabled_ after font properties have updated. Is that the case?

Screenshot 2022-04-07 at 15 27 13

(The video seems to indicate yes)

Can we add Jest JS tests for this? To show that fonts get updated and check expected history state.

Theoretically could add such tests also for the actual font check dialog as well, though we also have e2e tests there, so not super urgent

packages/wp-story-editor/src/components/fontCheck/index.js
```diff
@@ -58,6 +62,22 @@ export const FontCheck = () => {
...
+            (font) => font.family === fontFamily
+          );
+          if (matchedFont) {
+            // ensure `in use` font element uses the latest metrics, weights etc...
+            updateElementsByFontFamily({
+              family: fontFamily,
+              properties: { font: matchedFont },
+            });
+          }
+        });
```
swissspidy: Nit: a `for .. of` loop is more suitable here than `.map()` as we don't use the return values here.

packages/wp-story-editor/src/components/fontCheck/index.js
timarney: I think the storyPages dependancy ends up cycling so yes I think removing it from the array will fix this

Maybe from the this part of the `updateElementsByFontFamily`...

```javascript
const updatedPages = state.pages.map((page) => {
    const updatedElements = page.elements.map((element) => {
      if (element?.font?.family === family) {
        return updateElementWithUpdater(element, propertiesOrUpdater);
      }
      return element;
    });

    return {
      ...page,
      elements: updatedElements,
    };
  });
```

timarney:
> Regarding testing:
> 
> Might be easiest to test this be using [the secret editor devtools](https://github.com/GoogleForCreators/web-stories-wp/blob/main/docs/devtools.md) to update an element's font information, save the story, and then reload the page.
> 
> As for the history updates, the goal would be for the undo/redo buttons to still be _disabled_ after font properties have updated. Is that the case?
> 
> Screenshot 2022-04-07 at 15 27 13
> 
> (The video seems to indicate yes)
> 
> Can we add Jest JS tests for this? To show that fonts get updated and check expected history state.
> 
> Theoretically could add such tests also for the actual font check dialog as well, though we also have e2e tests there, so not super urgent


The editor dev tools work perfect for testing this --- I updated the testing instructions 

For Jest coverage ---  I previously added a test checking that the props get updated .  I can look into doing something similar but with history as well. 

https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/story-editor/src/app/story/useStoryReducer/test/updateElementsByFontFamily.js#L260

timarney:
Noting the failing e2e test for Dashboard options is fixed here:

https://github.com/GoogleForCreators/web-stories-wp/pull/11144/files#r847216208

swissspidy:
> For Jest coverage --- I previously added a test checking that the props get updated .

That just covers the functionality of the reducer, but not any of the logic in the font check dialog.

> Do we have any Jest test with history at the moment?

Searching for `HistoryContext.Provider` and `HistoryProvider` in the code base reveals couple of instances.

Though we can probably wrap the test directly in `HistoryProvider` and then check that `hasNewChanges` is still `false` after updating fonts.

timarney:
> Though we can probably wrap the test directly in HistoryProvider and then check that hasNewChanges is still false after updating fonts.

Which test are you referring to in this case?

swissspidy:
> Which test are you referring to in this case?

A yet to be written test for the Font Check dialog that covers more than just the `updateElementsByFontFamily` reducer.

Currently there is no test that verifies this flow.

packages/story-editor/src/index.js
```diff
@@ -40,6 +40,7 @@ export { default as getStoryPropsToSave } from './app/story/utils/getStoryPropsT
 export { default as useRefreshPostEditURL } from './utils/useRefreshPostEditURL';
 export { default as FontContext } from './app/font/context';
 export { default as useLoadFontFiles } from './app/font/actions/useLoadFontFiles';
+export { HistoryProvider, useHistory } from './app/history';
```
swissspidy: Nit: could also just add `export * from './app/history';` on line ~28 for consistency

packages/e2e-tests/src/specs/editor/fontCheck/fontCheckMetrics.js
```diff
@@ -0,0 +1,134 @@
...
+    await page.reload({ waitUntil: ['networkidle0', 'domcontentloaded'] });
+    const textElAfter = await expect(page).toMatchElement('p', {
+      text: 'Title 1',
+    });
+    const fontFamilyAfter = await page.evaluate(
+      (el) => getComputedStyle(el).font,
+      textElAfter
+    );
+    expect(fontFamilyAfter).toContain('Rock Salt');
+    //expect(fontFamilyAfter).toContain("sans-serif");
```
timarney: Intercepting the request with the updated metrics isn't quite working for this test --- for sure we can check the style / font family (in the dom) to see if the updates were applied --- might look at trying to use the dev tools for that part.

timarney:
Tried a couple of different approaches for an e2e test --- got pretty close but not quite working 

Version 1)

Tried intercepting the request to push new metrics

https://github.com/GoogleForCreators/web-stories-wp/pull/11205/commits/9e9fd34c4f86368a97059bf1f2900be1f197b0fc#diff-70cd65049bafa0f506c5a444143268e01b09fa70f6bfa412871edc10f538c0cdR89

Version 2)

Tried loading the metrics via the dev tools

https://github.com/GoogleForCreators/web-stories-wp/pull/11205/commits/0205a51c4c7f7274017f3ae45e3a49416710a802#diff-70cd65049bafa0f506c5a444143268e01b09fa70f6bfa412871edc10f538c0cdR189

Neither version is working the same as manually testing.

I did find an easy way to check for updated metrics

Changing the fallback font updates the font family --- which I check as part of the e2e test


Screen Shot 2022-04-13 at 9 41 48 PM

timarney:
Screen recordings to demo what I'm seeing (where the test is messing up)

Manually testing using Story dev tools - 

 - Open dev tools ✅ 
 - Update the Text to load (from cursive to sans-serif) ✅ 
 - Press load ✅ 
 -  Font family fallback updates to  sans-serif ✅ 

https://user-images.githubusercontent.com/62242/163382326-b9f6424e-cd88-4fe1-a761-5bb9ca41e8b2.mp4

**Puppeteer  E2E test**
- Open dev tools ✅ 
- Update the Text to load (from cursive to sans-serif) ✅ 
- Press load --- note in the video the value swaps back at this point to `cursive` ❌ 
-  Font family fallback remains as cursive :(



https://user-images.githubusercontent.com/62242/163382779-50c97b8a-c874-40a9-bd5f-da38e6210c52.mp4

timarney:
@swissspidy if you have any idea what might be happening here let me know see -> https://github.com/GoogleForCreators/web-stories-wp/pull/11205#issuecomment-1099097054

swissspidy:
@timarney The approach with request interception was very promising.

The only real issue there was that `mockResponse` was incorrect. It needs to be a proper Response object with `status` and `body` properties.

Plus there were some incorrect selectors like `button[aria-label="Redo Changes" disabled]`. It would need to be `button[aria-label="Redo Changes"][disabled]`

I just did some cleanup on the test now and it seems to pass fine locally.

timarney:
Okay ... I'll take another look with the updated commits.  If we can get the interception working we could avoid the dev tools.

swissspidy:
The interception _is_ working.

The devtools are only used to compare before/after state, nothing else.

swissspidy:
@timarney Did you override my fixes again now?

swissspidy:
@timarney Heads up! Currently working on some updates to help make the tests a bit less flakey. Please do not rebase/override my changes.

packages/e2e-tests/src/specs/editor/fontCheck/fontCheckMetrics.js
```diff
@@ -55,21 +55,30 @@ async function addStoryWithFont(title) {
...
+
+  await page.keyboard.down(isMacOS ? 'Meta' : 'Control');
   await page.keyboard.down('Shift');
   await page.keyboard.down('Alt');
   await page.keyboard.down('J');
   await page.keyboard.up('J');
   await page.keyboard.up('Alt');
   await page.keyboard.up('Shift');
-  await page.keyboard.up('Meta');
+  await page.keyboard.down(isMacOS ? 'Meta' : 'Control');
```
timarney, line 81: 👀 - nice catch

swissspidy: Took me a looooong time to figure that one out 🤦 

I confused it with the "Mod" key we use for keyboard shortcuts that does this switch automatically. But here we have to do this ourselves apparently.

swissspidy (dismissed):

packages/e2e-tests/src/specs/editor/fontCheck/fontCheckMetrics.js
spacedmonkey: This looks very similar too 
https://github.com/GoogleForCreators/web-stories-wp/blob/e00e50da7a02d956b993f4038ca1c15d46eda3ef/packages/karma-fixture/src/events.js#L556-L582

Any chance we could reuse it.

packages/wp-story-editor/src/components/fontCheck/index.js
```diff
@@ -49,25 +51,41 @@ export const FontCheck = () => {
 
   const closeDialog = () => setShowDialog(false);
 
-  useEffect(() => {
-    (async () => {
-      if (!isStoryLoaded) {
-        return;
-      }
+  useEffect(
+    () => {
```
spacedmonkey, line 55:
```suggestion
  useEffect(() => {
```

packages/e2e-tests/src/specs/editor/fontCheck/fontCheckMetrics.js
```diff
@@ -0,0 +1,205 @@
...
+  await publishStory();
+  // Somehow reloading the page after creating the story here helps
+  // with making the test more stable, i.e. toggling dev tools
+  // works more reliably.
+  await page.reload();
+}
+
+function isPlatformMacOS() {
+  return page.evaluate(() => {
+    const { platform } = window.navigator;
```
spacedmonkey, line 60: Should we be using window here?

packages/e2e-tests/src/specs/editor/fontCheck/fontCheckMetrics.js
```diff
@@ -0,0 +1,205 @@
...
+  await page.keyboard.down('Alt');
+  await page.keyboard.down('J');
+  await page.keyboard.up('J');
+  await page.keyboard.up('Alt');
+  await page.keyboard.up('Shift');
+  await page.keyboard.down(isMacOS ? 'Meta' : 'Control');
+
+  // Dev Tools were not open before, but now they should be open.
+  if (!areDevToolsOpen) {
+    await page.waitForSelector('#web-stories-editor textarea');
```
spacedmonkey, line 85: This string `#web-stories-editor textarea` is reused a lot. Should it be a const.

packages/e2e-tests/src/specs/editor/fontCheck/fontCheckMetrics.js
```diff
@@ -0,0 +1,205 @@
...
+        (response) =>
+          //eslint-disable-next-line jest/no-conditional-in-test -- False positive.
+          response.url().includes('web-stories/v1/fonts') &&
+          response.status() === 200
+      ),
+      page.reload(),
+    ]);
+
+    const newStoryData = await getCurrentStoryData();
+    const fontAfter = newStoryData.pages[0].elements[1].font;
```
spacedmonkey, line 179:
```suggestion
    const fontAfter = newStoryData.pages[0]?.elements[1]?.font;
```

packages/e2e-tests/src/specs/editor/fontCheck/fontCheckMetrics.js
```diff
@@ -0,0 +1,205 @@
...
+        false
+    );
+    await expect(canUndo).toBeFalse();
+
+    const canRedo = await page.evaluate(
+      () =>
+        document.querySelector('button[aria-label="Redo Changes"]').disabled ===
+        false
+    );
+    await expect(canRedo).toBeFalse();
```
spacedmonkey: These seem like odd checks. Wouldn't selector above cover this?

packages/e2e-tests/src/specs/editor/fontCheck/fontCheckMetrics.js
```diff
@@ -0,0 +1,205 @@
...
+  beforeAll(async () => {
+    await page.setRequestInterception(true);
+    stopRequestInterception = addRequestInterception((request) => {
+      if (request.url().includes('web-stories/v1/fonts') && mockResponse) {
+        request.respond(mockResponse);
+        return;
+      }
+
+      request.continue();
+    });
```
spacedmonkey, line 114: Still of mocking the response like this, I think it might be better to do this in PR with a little e2e plugin.

packages/e2e-tests/src/specs/editor/fontCheck/fontCheckMetrics.js
```diff
@@ -0,0 +1,205 @@
...
+
+    await Promise.all([
+      page.waitForResponse(
+        (response) =>
+          //eslint-disable-next-line jest/no-conditional-in-test -- False positive.
+          response.url().includes('web-stories/v1/fonts') &&
+          response.status() === 200
+      ),
+      page.reload(),
+    ]);
```
spacedmonkey, line 176: Why promise all here? Should we wait until one is finish then then next?

spacedmonkey (changes requested):
I have some questions that need to be actions.

packages/wp-story-editor/src/components/fontCheck/index.js
swissspidy: Code formatting is handled by Prettier.

packages/e2e-tests/src/specs/editor/fontCheck/fontCheckMetrics.js
swissspidy: Can you elaborate on this suggestion?

I don't see where these would ever be `null` at this stage.

packages/e2e-tests/src/specs/editor/fontCheck/fontCheckMetrics.js
swissspidy: Yes they are indeed redundant. I think I added these originally before adding the above selector and didn't remove them afterwards.

packages/e2e-tests/src/specs/editor/fontCheck/fontCheckMetrics.js
swissspidy: That would not work reliably, because our fonts request is done super early on mount, and `page.reload()` waits for `load`/`DOMContentLoaded`, so if we'd do `await page.reload(); await pageWaitForResponse()` then the request might have already finished at this point. So using `Promise.all()` avoids this race condition.

packages/e2e-tests/src/specs/editor/fontCheck/fontCheckMetrics.js
swissspidy: I actually find this to be a pretty good use case for request interception, since it's only needed for a part of the test.

A plugin is more useful if it's needed for the whole duration of the test, which is not the case here.

One drawback is that request interception does currently not work in Firefox, but I don't see us running tests against Firefox anytime soon.

packages/e2e-tests/src/specs/editor/fontCheck/fontCheckMetrics.js
swissspidy: I don't see a benefit in using a const here, but that's a bit of a personal preference I guess.

packages/e2e-tests/src/specs/editor/fontCheck/fontCheckMetrics.js
swissspidy: Yes, `navigator` is a property of the `window` global.

Worth noting that this is in a `page.evaluate()` call, so this code is actually executed in the context of the browser page, not Node.

packages/e2e-tests/src/specs/editor/fontCheck/fontCheckMetrics.js
swissspidy: Oh, nice find. And good question.

The only way to reuse this would be by copying it completely to our `e2e-test-utils` package.

Since we rarely use the keyboard in e2e tests and this here is the first time we've needed the Mac checks, I don't currently see the need for doing that. But we can do that once we need it more often.

swissspidy (approved):

timarney:
@swissspidy @spacedmonkey --- looks like the questions have responses anything else needed to wrap this ?

Issue #11121, opened 2022-03-30
Perf: Page Canvas Provider


BrittanyIRL:
We do a lot of image generation of story pages within the editor to accomplish tasks like reading pixel color data for accessibility, rendering 1 image node over entire story for performance, zooming into story etc.

Currently most of these features generate an image locally, then reference that locally generated image to achieve required functionality.

In #10842 we added a `PageCanvasProvider` which is a centralized place for generating images from pages and managing their cache. This follows [the architecture proposed in this thread here](https://github.com/GoogleForCreators/web-stories-wp/issues/9749#issuecomment-992667669)

Right now, the `PageCanvasProvider` only has adoption for the smart color for text feature. However, with this new centralized method of generating and caching images, we should be able to adopt `PageCanvasProvider` into all features that require an image of the story page.

This should benefit us two-fold:
1. Allowing for less expensive calls to generate images for the story page (more cache hits). 
2. Help with code quality. Once features adopt `PageCanvasProvider` they should no longer be responsible for image generation and caching of story page image logic.

PR #11207, opened 2022-04-06
Performance: Story carousel integrate with page canvas provider

### LINKED ISSUE #9958: Editor: Carousel thumbnails not consistently updating
There are a couple of issues I have been able to consistently create with the editor page thumbnail images. 

**Issue 1:** 
When first creating a new story you are given a blank page. If you add a new background image to that page the thumbnail image for the page below remains blank. If you then add a new page or duplicate that page then the page 1 thumbnail will stay blank until you click on it to go back to page 1 which then seems to refresh the thumbnail and the image shows. 

I presume that this should not be happening because from any newly added page after the first the thumbnail image updates straight after adding a background image to the page. 

**Issue 2:** 
The thumbnail image of a duplicated page reverts back to it's previous background image if it has been changed prior to duplicating the page once more. If you duplicate page 1 for page 2, update the page 2 background image and duplicate page 2 for page 3 then the thumbnail image for page 2 reverts back to the original image from the duplicated page 1.  

When going back to page 2 the background image is correct on the page but the thumbnail is wrong. If you double-click the background image on the page and move it slightly then the thumbnail corrects itself.

## Expected Behaviour

**Issue 1:** Background image added to page 1 should update on thumbnail image.

**Issue 2:** The page 2 thumbnail image does not revert back to the original duplicated image and remains as the updated background image.

## Steps to Reproduce

**Issue 1:** 
1. Create a new Web Story.
2. Add a background image to the first page.

**Issue 2:**
1. Create a New Story.
2. Add a Background image to page1.
3. Duplicate page 1 to create page 2.
4. Change the Background image for page 2.
5. Duplicate page 2 to create page 3.

## Video recording

The following video shows both issues which I am able to reproduce consistently:
https://user-images.githubusercontent.com/91896826/145550008-71b9f115-e295-40fb-9df3-7e8577047085.mp4

### LINKED ISSUE #10822: Performance: Story carousel integrate with page canvas provider

maxyinger:
## Context
Part of this perf epic #11121 

## Summary
Updates page previews to just use the page canvas provider for obtaining generated images.

Should also fix this issue as a byproduct of integrating the new approach:
#9958 

## Relevant Technical Choices
Utilizes the background task queue and integrates canvas generation with smart colors. So... if the smart colors generate a canvas for their task, the carousel can use it. Also, there should never be more than one background task request to generate a page canvas at a time.

Also updates all image thumbnails to just use an image if it's available. This should work fine in both the grid & the checklist and should come at no additional cost. Should honestly help by cutting down on number of rendered nodes in both those places

## To-do
NA

## User-facing changes
Fixes stale carousel image issue

## Testing Instructions

(original bug reproduction scenario)

**Issue 1:**

1. Create a new Web Story.
2. Add a background image to the first page.

**Issue 2:**

1. Create a New Story.
2. Add a Background image to page1.
3. Duplicate page 1 to create page 2.
4. Change the Background image for page 2.
5. Duplicate page 2 to create page 3.

**Other Testing Area**
This PR also updates the thumbnails in the story grid view & the checklist, so ensure that you don't see regressions in any of those places either.

## Checklist

---

Fixes #10822 
Fixes #9958


packages/story-editor/src/utils/storyPageToCanvas.js
```diff
@@ -43,7 +43,7 @@ async function storyPageToCanvas(page, { width = 400, ...options }) {
   const canvas = await htmlToImage.toCanvas(node, {
     ...options,
     fontEmbedCss: '',
-    pixelRatio: 1,
+    pixelRatio: window.devicePixelRatio,
```
swissspidy: This is the default value: https://github.com/bubkoo/html-to-image/blob/30000f088ef94878e2fa67dfcf89a09c13a0e14c/src/util.ts#L140

maxyinger: ahh. right on. will update

BrittanyIRL: so it should be `pixelRatio: window.devicePixelRatio || 1`?

packages/story-editor/src/components/footer/pagepreview/index.js
```diff
@@ -92,61 +89,52 @@ const PreviewWrapper = styled.div`
...
+  position: absolute;
   width: 100%;
+  left: 0;
+
+  // image will always be vertically centered
+  // so we don't need to alter height of image
+  // when it's in the fullbleed ratio and the
+  // thumbnail is in the smaller ratio
+  top: 50%;
+  transform: translateY(-50%);
```
BrittanyIRL, line 103: 👍

packages/story-editor/src/utils/storyPageToCanvas.js
swissspidy: Since that is the default, my understanding is that this line can be removed

maxyinger: we can just omit passing in the `pixelRatio` arg completely cus the package defaults to the behavior we want

BrittanyIRL (approved):
This is so great, I'm glad that you made time to make this update after it was talked about so long ago and then put aside. Getting to just reuse the page canvas provider here is awesome. 

Working locally for me.

I just triggered a rerun for the[ 1 failing karma test ](https://github.com/GoogleForCreators/web-stories-wp/runs/5856833275?check_suite_focus=true#step:7:773)but there might be something that needs updating there?  As well as in unit tests,[ here](https://github.com/GoogleForCreators/web-stories-wp/runs/5856829572?check_suite_focus=true#step:6:71).

swissspidy:
The one thing that stands out in the Percy snapshots is that all the carousel images now have a focus outline, because that's what the active pages have on the canvas. Sounds like a bug to me.
Otherwise the carousel images now finally look correct 🎉

merapi (approved):
It's great you did it. I can't see any issues with thumbnails generation now 🧙🏻🪄

BTW: Does anybody know why we crop the thumbnail like this:
image

littlemilkstudio:
@merapi 
> BTW: Does anybody know why we crop the thumbnail like this:

I think just to match the ratio that the story thumbnails display at in the dashboard & out in the wild? @BrittanyIRL & @swissspidy might be able to elaborate on this better 🤷

merapi:
Most devices look like this:
image

In the Dashboard, we have a poster image, not a thumbnail generated from the story, right?
I will forward this to UX.

samwhale (approved):

packages/story-editor/src/app/pageCanvas/usePageCanvas.js
```diff
@@ -23,8 +23,8 @@ import { useContextSelector, identity } from '@googleforcreators/react';
  */
 import Context from './context';
 
-function useCalculateAccessibleTextColors(selector) {
+function usePageCanvas(selector) {
   return useContextSelector(Context, selector ?? identity);
 }
 
-export default useCalculateAccessibleTextColors;
+export default usePageCanvas;
```
samwhale, line 30: haha nice catch

samwhale (commented):
Nice this is way simpler. Good job dude!

BrittanyIRL:
> @merapi
> 
> > BTW: Does anybody know why we crop the thumbnail like this:
> 
> I think just to match the ratio that the story thumbnails display at in the dashboard & out in the wild? @BrittanyIRL & @swissspidy might be able to elaborate on this better 🤷

yeah dashboard is poster images so this would be separate. Seems like the top and bottom safe area aren't getting cut on the thumbnail vs in the preview.

agingoldseco:
> > @merapi
> > > BTW: Does anybody know why we crop the thumbnail like this:
> > 
> > 
> > I think just to match the ratio that the story thumbnails display at in the dashboard & out in the wild? @BrittanyIRL & @swissspidy might be able to elaborate on this better 🤷
> 
> yeah dashboard is poster images so this would be separate. Seems like the top and bottom safe area aren't getting cut on the thumbnail vs in the preview.

I think this cropping maximizes the amount of useful detail you see in the thumbnails to differentiate pages while saving some space at the bottom of the screen by not including the very top and bottom. It makes sense to me.

packages/story-editor/src/app/pageCanvas/usePageCanvas.js
maxyinger: 🙈

maxyinger: 🤫

merapi:
> I think this cropping maximizes the amount of useful detail you see in the thumbnails to differentiate pages while saving some space at the bottom of the screen by not including the very top and bottom. It makes sense to me.

@agingoldseco Alright, if we ever decide to show full canvas - we don't have to sacrifice the vertical space, we can just make the thumbnails a little bit more narrow to keep the same ratio, like this:
thumbs-ratio

packages/story-editor/src/components/footer/pagepreview/index.js
BrookeGraham: great comment

BrookeGraham (approved):

packages/story-editor/src/utils/storyPageToCanvas.js
```diff
@@ -43,7 +43,6 @@ async function storyPageToCanvas(page, { width = 400, ...options }) {
   const canvas = await htmlToImage.toCanvas(node, {
     ...options,
     fontEmbedCss: '',
-    pixelRatio: 1,
```
barklund, line 46: Hm, I think we needed this one for some reason. Or is it removed because it has now been included up-stream in the original library? CC @merapi

BrittanyIRL: For reference: https://github.com/GoogleForCreators/web-stories-wp/pull/11207#discussion_r844165130

maxyinger: we want `window.devicePixelRatio || 1` here so the images we generate don't look blurry in the grid or thumbnails, this is the what the library defaults to so we can omit it. Currently this is only used for thumbnails and smart colors and doesn't cause regressions in either.

PR #11213, opened 2022-04-07
Editor: Fix issues with saving the demo story

### LINKED ISSUE #11151: Unable to save demo story (REST API preloading side effect)
## Bug Description

When you activate the Web Stories plugin, there is a banner that directs you to `/wp-admin/post-new.php?post_type=web-story&web-stories-demo=1`, which displays a Getting Started story.

It turns out that saving this story is not currently possible. When you try it, it breaks. Not that this something we intend users to do, but there's still an actual bug in there that is worth solving.

From what I can see, something related to the publisher logo causes the global post object/ID to be overridden with the currently active publisher logo ID (`web_stories_active_publisher_logo` option), hence breaking saving of the story, because it tries to access an attachment and not a story.

For example, on my site `web_stories_active_publisher_logo` is `7381`. If I do `var_dump( $GLOBALS['post'] );` here:

https://github.com/GoogleForCreators/web-stories-wp/blob/2d71fc965a317b06b280ed65448d705d1d892eb7/includes/templates/admin/edit-story.php#L128-L132

Then I get `object(WP_Post)#1573 (24) { ["ID"]=> int(7381)....` — not the current story.

Here's what I found out so far:

When `web-stories-demo` is set, we call `rest_preload_api_request()` here:

https://github.com/GoogleForCreators/web-stories-wp/blob/2d71fc965a317b06b280ed65448d705d1d892eb7/includes/templates/admin/edit-story.php#L128-L132

`$story_query_params` is defined as follows:

https://github.com/GoogleForCreators/web-stories-wp/blob/2d71fc965a317b06b280ed65448d705d1d892eb7/includes/templates/admin/edit-story.php#L93-L99

Now, because of this `wp:publisherlogo` array item and this definition in the controller:

https://github.com/GoogleForCreators/web-stories-wp/blob/2d71fc965a317b06b280ed65448d705d1d892eb7/includes/REST_API/Stories_Controller.php#L606-L618

... and because we filter this meta value to always return the currently active publisher logo, this will essentially embed the `/web-stories/v1/media/7381` response in the `Story_Controller` REST API response.

So something in `Stories_Media_Controller` or `WP_REST_Attachments_Controller` must be affecting the global post object.

This side effect causes the saving of the demo story to fail.

## Expected Behaviour

There should be no unexpected side effects caused by REST API preloading that reset the global `$post`.

Saving the demo story should be possible.

We can add an e2e test to verify this.

## Steps to Reproduce

1. Visit `/wp-admin/post-new.php?post_type=web-story&web-stories-demo=1`
2. Try to save story
3. See error

## Additional Context

- Plugin Version: 1.19.0
- WordPress Version: 5.9.2
- Operating System: macOS
- Browser: Chrome 100

spacedmonkey:
## Summary

Move the backup of post object to before the preload call to ensure it backuped and restored correctly. 

## Relevant Technical Choices

Extend existing tests to ensure there was no error on saving. 

## Testing Instructions

1.

## Checklist

---

Fixes #11151

### DIFF
--- a/includes/templates/admin/edit-story.php
+++ b/includes/templates/admin/edit-story.php
@@ -122,6 +122,13 @@ $story_query_params = [
 	),
 ];
 
+/*
+ * Ensure the global $post remains the same after API data is preloaded.
+ * Because API preloading can call the_content and other filters, plugins
+ * can unexpectedly modify $post.
+ */
+$backup_global_post = $post;
+
 if ( empty( $_GET['web-stories-demo'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 	$preload_paths[] = $story_initial_path . build_query( $story_query_params );
 } else {
@@ -141,13 +148,6 @@ if ( empty( $_GET['web-stories-demo'] ) ) { // phpcs:ignore WordPress.Security.N
  * @param WP_Post  $post          Post being edited.
  */
 $preload_paths = apply_filters( 'web_stories_editor_preload_paths', $preload_paths, $post );
-
-/*
- * Ensure the global $post remains the same after API data is preloaded.
- * Because API preloading can call the_content and other filters, plugins
- * can unexpectedly modify $post.
- */
-$backup_global_post = $post;
 
 $preload_data = array_reduce(
 	$preload_paths,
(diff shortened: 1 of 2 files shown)


swissspidy (approved):
LGTM

PR #11223, opened 2022-04-08
Floating menu: Keyboard Navigation to Enter and Exit

### LINKED ISSUE #10872: Floating menu: Keyboard Navigation to Enter and Exit

maxyinger:
## Context
Part of accessibility work on canvas and floating menu

## Summary
Adds `ctrl+alt+p` to enter into the floating menu when an element is focused and selected, `esc` to return focus back to the original element in the frames layer.

## Relevant Technical Choices
NA

## To-do
NA

## User-facing changes
User can now navigate into the floating menu with keyboard by pressing `ctrl+alt+p` while an element is selected. They can navigate out of the floating menu and back to the element by pressing `esc`.

## Testing Instructions

1. with an element selected & focused, press `ctrl+alt+p` and see that focus goes inside the floating menu
2. now press `esc` and see that focus leaves the floating menu

## Checklist

---

Fixes #10872

### DIFF
--- a/packages/design-system/src/components/contextMenu/menu.js
+++ b/packages/design-system/src/components/contextMenu/menu.js
@@ -23,6 +23,7 @@ import {
   useMemo,
   useRef,
   useCombinedRefs,
+  forwardRef,
 } from '@googleforcreators/react';
 import styled, { css } from 'styled-components';
 /**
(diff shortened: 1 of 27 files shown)


packages/story-editor/src/components/canvas/karma/elementKeyboardNavigation.js
```diff
@@ -0,0 +1,68 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * Internal dependencies
+ */
+import { Fixture } from '../../../karma';
+import { useStory } from '../../../app/story';
+
+fdescribe('Canvas Element - keyboard navigation', () => {
```
BrittanyIRL: linter caught this too ^^

packages/design-system/src/components/contextMenu/menu.js
```diff
@@ -102,178 +103,188 @@ function getFocusableChildren(parent, isSubMenu) {
...
+     * Clicking in the wrapper should not automatically focus the first
+     * focusable element.
+     */
+    const handleFocus = useCallback(
+      (evt) => {
+        onFocus(evt);
+        const menuChildren = menuRef.current.children || [];
+        if (
+          menuRef.current === evt.target &&
+          ![...menuChildren].some((child) => child.contains(evt.target))
```
samwhale: 🤔 for an empty array, doesn't `![...menuChildren].some((child) => child.contains(evt.target))` come out as `true`?

image

I guess it doesn't matter too much since there's no focusable children? Maybe there should be a length check to make sure that the rest of that logic isn't run?
```suggestion
          menuChildren.length && ![...menuChildren].some((child) => child.contains(evt.target))
```

packages/design-system/src/components/contextMenu/menu.js
```diff
@@ -102,178 +103,188 @@ function getFocusableChildren(parent, isSubMenu) {
...
-    { key: ['esc', 'down', 'up', 'left', 'right'] },
-    handleKeyboardNav,
-    [handleKeyboardNav]
-  );
+    useKeyDownEffect(
+      menuRef,
+      {
+        key: dismissOnEscape
+          ? ['esc', 'down', 'up', 'left', 'right']
+          : ['down', 'up', 'left', 'right'],
```
samwhale, line 261: ✨

packages/story-editor/src/components/canvas/frameElement.js
```diff
@@ -87,7 +97,17 @@ const EmptyFrame = styled.div`
 
 const NOOP = () => {};
 
+const FRAME_ELEMENT_MESSAGE = sprintf(
+  /* translators: 1: Ctrl Key 2: Alt Key */
+  __('To enter floating menu, press %1$s+%2$s+p.', 'web-stories'),
```
samwhale: 💭 When chromevox is reading this out, it reads out the `+` symbol too. I think the `+` symbols could be removed

```suggestion
  __('To enter floating menu, press %1$s %2$s p.', 'web-stories'),
```

packages/story-editor/src/components/canvas/frameElement.js
```diff
@@ -178,6 +199,16 @@ function FrameElement({ id }) {
...
   const handleMediaClick = useDoubleClick(NOOP, handleMediaDoubleClick);
 
+  /**
+   * Announce keyboard options on element.
+   *
+   * Using a live region because an `aria-label` would remove
+   * any labels/content that would be read from children.
+   */
+  const handleFocus = useCallback(() => {
+    speak(FRAME_ELEMENT_MESSAGE);
```
samwhale: 💭 The background element does not have a floating menu, so I think this needs to be prevented for the background element.

packages/story-editor/src/components/floatingMenu/menu.js
```diff
@@ -72,7 +93,10 @@ const FloatingMenu = memo(
             isHorizontal
             isSecondary
             isAlwaysVisible
-            disableControlledTabNavigation
+            tabIndex={-1}
+            ref={focusGroupRef}
+            dismissOnEscape={false}
+            onFocus={announceKeyboardControls}
```
samwhale, line 99: 👍

packages/story-editor/src/components/canvas/karma/elementKeyboardNavigation.js
```diff
@@ -0,0 +1,68 @@
...
+    await fixture.events.click(rectangleShape);
+
+    // Click on one of the added shapes
+    const elementId = await fixture.renderHook(() =>
+      useStory(({ state }) => state.currentPage.elements[1].id)
+    );
+    const wrapperEl = getCanvasElementWrapperById(elementId);
+    await fixture.events.click(wrapperEl);
+    expect(document.activeElement).toBe(wrapperEl);
+
```
samwhale: 🤔 I think the shape element will be selected when you click on the button on line 49. This part of the test may not be necessary

samwhale (commented):
This is looking good! A few things that need some small changes. Test is great!

samwhale:
🤔 I think the failing js tests are [fixed in `main` already](https://github.com/GoogleForCreators/web-stories-wp/pull/11193/files#diff-bbead4800ca38928ee901c7b13d50185ce9e61084905a64d4b6c703ce6a4527dR104-R106). Merging that in should fix them

packages/design-system/src/components/contextMenu/menu.js
maxyinger: ah good catch. thanks

packages/story-editor/src/components/canvas/karma/elementKeyboardNavigation.js
maxyinger: yup. was able to remove it. good catch

BrittanyIRL (approved):

samwhale (approved):
Awesome!

samwhale:
Note: exiting the floating menu focus group and going back to the canvas focus group will be done in a separate ticket.

Issue #11210, opened 2022-04-06
Shopping: Products JSON Schema Validation


swissspidy:
## Feature Description

Context: https://github.com/ampproject/amphtml/issues/37746

What is JSON Schema: https://json-schema.org/

Current schema: https://github.com/ampproject/amphtml/blob/main/examples/amp-story/shopping/product.schema.json

Possibly helpful utils for testing:

https://www.npmjs.com/package/jest-json-schema
https://github.com/martin-helmich/phpunit-json-assert

Using this JSON schema for shoppable products, we can verify that the data we return from WooCommerce & Shopify matches the expected format. So most likely this is something for the PHPUnit integration tests.

But this could also be useful for an e2e test, e.g.:

1. Set up WooCommerce with some products
2. Create a new story and add some products
3. Preview the story
4. Take the resulting shopping config
5. Compare the config against the schema

So for that a Jest helper might be useful.

Right now it doesn't look like the schema is hosted somewhere publicly, so we could just create a simply GitHub action workflow to regularly copy the file to this repo for easier usage.

jshamble:
I'll be adding public-facing documentation with a link that points to the product json schema [here](https://github.com/ampproject/amphtml/pull/38042).

timarney:
After digging into this a bit and starting down the PHP unit test road I think the e2e test is going to make the most sense here given the data want to match against and the way that data flows in and out / mocks etc...

swissspidy:
Related: https://github.com/GoogleForCreators/web-stories-wp/pull/11340/files#r857646203

I feel like we shouldn't write off validation in PHP tests just yet, it seems useful to ensure that the query classes return data in expected format. Even if the inputs are just mocked.

timarney:
> I feel like we shouldn't write off validation in PHP tests just yet

Agreed --- feels like there is some value on some PHP checks --- I did test the `phpunit-json` package and it seemed good.

swissspidy:
Just occurred to me that `rest_validate_value_from_schema()` _might_ be able to perform this validation as well. So maybe we don't even need a separate package for this. Needs testing though.

(For context, I stumbled upon https://meta.trac.wordpress.org/changeset/11410)

spacedmonkey:
I put together a little prototype of this - https://github.com/GoogleForCreators/web-stories-wp/tree/try/dynamic-schema. This some promising results. We would have to solve translated string problem, but this looks good. 
[...]

PR #11250, opened 2022-04-12
Shopping: Products JSON Schema Validation 

timarney:
## Context

In support of #11210 we need make the WoocCommerce plugin available with some sample products to query against to ensure a valid JSON schema is returned.

> Using this JSON schema for shoppable products, we can verify that the data we return from WooCommerce & Shopify matches the expected format. So most likely this is something for the PHPUnit integration tests.

## Summary

- Adds  WoocCommerce  plugin installation step to start script
- Adds sample product import

## To-do
- e2e test is failing which is expected as the value for product price isn't matching the schema

## Testing Instructions

- `npm run env:start`
-  note CLI outputs that WooCommerce is installed
-  note in the WP Admin - WooCommerce has imported sample products (via the sample xml file)

1. Run `npm run env:start` --- this should install WooCommerce + sample products
2. Activate the WooCommerce plugin --- ensure sample products exists

No

No

No

## Checklist

---

Partially addresses #11210


timarney:
Noting this was failing on WP 5.5 -- not sure if that's a plugin issue or something else happening.  For now I have wrapped a WP latest check around the install.

https://github.com/GoogleForCreators/web-stories-wp/runs/6061324126?check_suite_focus=true

packages/e2e-tests/src/specs/editor/shopping/woocommerce.js
```diff
@@ -0,0 +1,51 @@
...
+describe('Shopping', () => {
+  describe('Shopping schema', () => {
+    let stopRequestInterception;
+    withExperimentalFeatures(['shoppingIntegration']);
+
+
+    beforeAll(async () => {
+      await page.setRequestInterception(true);
+      stopRequestInterception = addRequestInterception((request) => {
+        // ... grab story json to validate against schema ?
```
timarney, line 36: @swissspidy @spacedmonkey 

Is there a call / endpoint I can intercept here to get the story config to compare against ?

I can see I'm getting data (via the story dev tools) after I run this e2e test

```
"type": "product",
          "product": {
            "productId": "wc-36",
            "productTitle": "T-Shirt with Logo",
            "productBrand": "",
            "productPrice": "18",
            "productPriceCurrency": "USD",
            "productImages": [],
            "aggregateRating": {
              "ratingValue": "0",
              "reviewCount": 0,
              "reviewUrl": "http://localhost:8899/product/t-shirt-with-logo"
            },
            "productDetails": "This is a simple product.",
            "productUrl": "http://localhost:8899/product/t-shirt-with-logo"
          },
```

packages/e2e-tests/src/specs/editor/shopping/schema.json
```diff
@@ -0,0 +1,281 @@
+{
```
timarney, line 1: We'll need to pull this in but adding for now to get a test working

includes/REST_API/Products_Controller.php
```diff
@@ -383,11 +383,11 @@ static function( $image_id ) {
 				'productId'            => 'wc-' . $product->get_id(),
 				'productTitle'         => $product->get_title(),
 				'productBrand'         => '', // TODO: Figure out how to best provide that.
-				'productPrice'         => $product->get_price(),
+				'productPrice'         => number_format( (float) $product->get_price(), 2 ),
```
timarney: This was failing the schema as it was coming back from WooCommerce as a string

includes/REST_API/Products_Controller.php
```diff
@@ -383,11 +383,11 @@ static function( $image_id ) {
...
 				'productPriceCurrency' => get_woocommerce_currency(),
 				'productImages'        => $images,
 				'aggregateRating'      => [
-					'ratingValue' => $product->get_average_rating(),
+					'ratingValue' => number_format( (float) $product->get_average_rating(), 2 ),
```
timarney: This was failing the schema as it was coming back from WooCommerce as a string

includes/REST_API/Products_Controller.php
```diff
@@ -387,7 +387,7 @@ static function( $image_id ) {
 				'productPriceCurrency' => get_woocommerce_currency(),
 				'productImages'        => $images,
 				'aggregateRating'      => [
-					'ratingValue' => $product->get_average_rating(),
+					'ratingValue' => (int) $product->get_average_rating(),
```
swissspidy: This looks incorrect to me.

The schema says `ratingValue` is a `number`, [which can be a floating point number](https://json-schema.org/understanding-json-schema/reference/numeric.html).

`get_average_rating()` returns a float.

This is important because the average rating could be something like 4.4. Don't want to cast this to an int and turn this into `4`.

bin/local-env/install-wordpress.sh
```diff
@@ -139,6 +139,15 @@ fi
 echo -e $(status_message "Installing and activating RTL Tester plugin...")
 wp plugin install rtl-tester --activate --force --quiet
 
+echo -e $(status_message "Installing WordPress importer...")
+wp plugin install wordpress-importer --activate --force --quiet
+
+# Only install woocommerce on latest version of WordPress.
```
swissspidy, line 145: Sounds reasonable; WooCommerce has a high WP version requirement (5.7+)

bin/local-env/install-wordpress.sh
```diff
@@ -219,3 +228,11 @@ wp option patch insert web_stories_experiments enableSVG 0
...
 wp post list --post_type=attachment --format=yaml
 wp plugin list --format=yaml
+
+# Only install woocommerce on latest version of WordPress.
+if [ "$WP_VERSION" == "latest" ]; then
+	echo -e $(status_message "Import sample woocommerce products...")
+	wp import /var/www/html/wp-content/plugins/woocommerce/sample-data/sample_products.xml --authors=skip --quiet
+	# deactivate test etc... can activate as needed
+	wp plugin deactivate woocommerce 
+fi
```
swissspidy: Note: missing newline at end of file

bin/local-env/install-wordpress.sh
```diff
@@ -219,3 +228,11 @@ wp option patch insert web_stories_experiments enableSVG 0
 wp user list --format=yaml
 wp post list --post_type=attachment --format=yaml
 wp plugin list --format=yaml
+
+# Only install woocommerce on latest version of WordPress.
+if [ "$WP_VERSION" == "latest" ]; then
+	echo -e $(status_message "Import sample woocommerce products...")
+	wp import /var/www/html/wp-content/plugins/woocommerce/sample-data/sample_products.xml --authors=skip --quiet
+	# deactivate test etc... can activate as needed
+	wp plugin deactivate woocommerce
```
swissspidy, line 237: Does the plugin need to be activated further above in the first place? Is it required for the import to work?

Otherwise we can remove this line & the `--activate` flag above.

packages/e2e-tests/src/specs/editor/shopping/woocommerce.js
```diff
@@ -0,0 +1,104 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * External dependencies
+ */
+
+// eslint-disable-next-line import/no-extraneous-dependencies -- in package.json
+import { matchers } from 'jest-json-schema';
```
swissspidy: The dependency needs to be in the dependencies of `e2e-tests/package.json`, not the main `package.json`.

Then this lint warning will go away.

packages/e2e-tests/src/specs/editor/shopping/woocommerce.js
```diff
@@ -0,0 +1,104 @@
...
+  takeSnapshot,
+} from '@web-stories-wp/e2e-test-utils';
+
+/**
+ * Internal dependencies
+ */
+import * as schema from './schema.json';
+
+// eslint-disable-next-line jest/require-hook -- matchers only for this set of tests
+expect.extend(matchers);
```
swissspidy: Let's move this to `bootstrap.js`. Should be fine to make this matcher available for all e2e tests

packages/e2e-tests/src/specs/editor/shopping/woocommerce.js
```diff
@@ -0,0 +1,104 @@
...
+    afterAll(async () => {
+      if ('latest' === WPVersion) {
+        await deactivatePlugin('woocommerce');
+      }
+    });
+
+    if (WPVersion !== 'latest') {
+      //eslint-disable-next-line jest/expect-expect, jest/no-focused-tests
+      it.only('minimum WordPress requirement not met', () => {});
+    }
```
swissspidy: Can use `minWPVersionRequired` from `e2e-utils` package.

packages/e2e-tests/src/specs/editor/shopping/woocommerce.js
```diff
@@ -0,0 +1,104 @@
...
+          const { type, product } = element;
+          //eslint-disable-next-line jest/no-conditional-in-test -- only add products
+          if (type === 'product') {
+            products.push(product);
+          }
+        });
+      });
+
+      expect(products).toHaveLength(1);
+      expect(products[0]).toMatchSchema(schema);
```
swissspidy: The product element data does not necessarily map to the product schema 1:1.

What's important is that the final product output when previewing a story is valid.

So instead of opening devtools, let's preview the story, grab the product config from the amp-story-shopping config and validate that one.

packages/e2e-tests/src/specs/editor/shopping/woocommerce.js
```diff
@@ -0,0 +1,104 @@
...
+      await takeSnapshot(page, 'insert product');
+      await publishStory();
+
+      // open story dev tools
+      await page.keyboard.down('Meta');
+      await page.keyboard.down('Alt');
+      await page.keyboard.down('Shift');
+      await page.keyboard.down('J');
+
+      await takeSnapshot(page, 'shopping schema');
```
swissspidy: Don't think we need a screenshot of the dev tools; doesn't sound useful.

packages/e2e-tests/src/specs/editor/shopping/woocommerce.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    it('should should match a valid schema', async () => {
+      await createNewStory();
+      await expect(page).toClick('[aria-controls="library-pane-shopping"]');
+      await expect(page).toClick('[aria-label="Product"]');
+      await page.keyboard.press('ArrowDown');
+      await page.keyboard.press('Enter');
+      await expect(page).toClick('button', { text: 'Insert product' });
+      await takeSnapshot(page, 'insert product');
```
swissspidy: I think we can introduce this snapshot later once we add e2e tests for product insertion in general, not for the schema test.

packages/e2e-tests/src/specs/editor/shopping/woocommerce.js
```diff
@@ -0,0 +1,104 @@
...
+      it.only('minimum WordPress requirement not met', () => {});
+    }
+
+    it('should should match a valid schema', async () => {
+      await createNewStory();
+      await expect(page).toClick('[aria-controls="library-pane-shopping"]');
+      await expect(page).toClick('[aria-label="Product"]');
+      await page.keyboard.press('ArrowDown');
+      await page.keyboard.press('Enter');
+      await expect(page).toClick('button', { text: 'Insert product' });
```
swissspidy: It would be nice to know which product this is inserting. Maybe in the dropdown we can select a few specific products by title?

includes/REST_API/Products_Controller.php
spacedmonkey:
```suggestion
					'ratingValue' => (float) $product->get_average_rating(),
```

packages/e2e-tests/src/specs/editor/shopping/schema.json
swissspidy: Happy to help with pulling this in automatically, should be trivial

bin/local-env/install-wordpress.sh
timarney: Does the plugin need to be activated further above in the first place?

```bash
> STATUS: Import sample woocommerce products...
Failed to import &#8220;V-Neck T-Shirt&#8221;: Invalid post type product<br />Failed to import 
```
☝️ 

Without activating above the import fails

packages/e2e-tests/src/specs/editor/shopping/woocommerce.js
timarney: moved https://github.com/GoogleForCreators/web-stories-wp/pull/11250/commits/93cf9536bf636c5e58b44bb9463e5c0217eb590e

packages/e2e-tests/src/specs/editor/shopping/schema.json
swissspidy: Added a GitHub Actions workflow for this now. That should do the trick.

timarney: Did you commit to this PR?  If yes think I rebased over it --- sorry.

packages/e2e-tests/src/specs/editor/shopping/woocommerce.js
timarney: Updated to pull from the preview

packages/e2e-tests/src/specs/editor/shopping/woocommerce.js
timarney: updated to type in a product name

packages/e2e-tests/src/specs/editor/shopping/schema.json
swissspidy: yes... let me try to add it again

swissspidy: My fault for pushing to your PR right away, but to prevent overrides form happening again in the future using `--force-with-lease` instead of `--force` is useful. (or merging instead of rebasing & force-pushing)

packages/e2e-tests/src/specs/editor/shopping/woocommerce.js
timarney: So ... I swapped this back to using minWPVersionRequired

The WP 5.5. e2e test run is failing as it's looking for a Plugin that doesn't exist --- any way we can avoid this outside of how I was doing it prior.


WP 5.5
```
No node found for selector: tr[data-slug="woocommerce"] .activate a

      29 |
      30 |   if (!disableLink) {
    > 31 |     await page.click(`tr[data-slug="${slug}"] .activate a`);
         |     ^
      32 |     await page.waitForSelector(`tr[data-slug="${slug}"] .deactivate a`);
      33 |   }
```

timarney:
https://github.com/GoogleForCreators/web-stories-wp/runs/6119493501?check_suite_focus=true#step:9:14

@swissspidy  I could be wrong but I think we're running into the `AMP sanitization` you mentioned in the original PR 

https://github.com/GoogleForCreators/web-stories-wp/pull/11158/#issuecomment-1088570370


Screen Shot 2022-04-21 at 5 19 04 PM

timarney:
> Note: need to disable AMP sanitization (e.g. by commenting out the relevant line in Output_Buffer) for now since this is not stable yet.

Re tested locally with the AMP sanitization off 

The e2e test is catching / working as expected.

Screen Shot 2022-04-22 at 6 07 01 AM

packages/e2e-tests/src/specs/editor/shopping/woocommerce.js
```diff
@@ -0,0 +1,64 @@
...
+        JSON.parse(
+          document.querySelector('amp-story-shopping-attachment script')
+            .textContent
+        )
+      );
+
+      await page.bringToFront();
+      await previewPage.close();
+      expect(data.items).toHaveLength(1);
+      expect(data.items[0]).toMatchSchema(schema);
```
swissspidy: Let's add a few more products at the beginning of the test and then loop through all `data.items` to do the schema validation on every product

packages/e2e-tests/src/specs/editor/shopping/woocommerce.js
```diff
@@ -0,0 +1,64 @@
...
+      await page.type('[aria-label="Product"]', 'Hoodie with Zipper');
+      await page.keyboard.press('ArrowDown');
+      await page.keyboard.press('Enter');
+      await expect(page).toClick('button', { text: 'Insert product' });
+      await publishStory();
+      const previewPage = await previewStory(page);
+      await previewPage.waitForSelector('amp-story-shopping-attachment script');
+
+      // pull product data from the DOM
+      const data = await previewPage.evaluate(() =>
```
swissspidy: A `expect(data.items).toMatchInlineSnapshot()` call here might be useful, makes it easy to see at first glance what products we're expecting to be there.

swissspidy:
I think updating the Composer dependencies helped with the AMP sanitization part.

For the failing preview I need to take a closer look on Monday. Is it working for you locally with `npm run test:e2e:interactive packages/e2e-tests/src/specs/editor/shopping/woocommerce.js`?

timarney:
> Composer dependencies

Okay that would make sense

Locally it's still getting stuck 

Screen Shot 2022-04-22 at 6 03 36 PM


Screen Shot 2022-04-22 at 6 04 09 PM

timarney:
Locally running the single test with Output buffering commented out it finishes the test.


Screen Shot 2022-04-22 at 6 15 27 PM
t

swissspidy:
Hmm alright, then I didn‘t look close enough. Will re-check on Monday.

timarney:
@swissspidy with your composer update commit merged back to this one --- this PR is now working as expected (fails on the schema check string vs number)

swissspidy:
> @swissspidy I tested locally with your composer update commit merged back to this one it's s now working as expected (fails on the schema check string vs number)

Okay and which schema check is exactly failing? Can we fix our code?

I only see this error message:

```
  ● Shopping › Shopping schema › should should match a valid schema

    TimeoutError: waiting for selector `amp-story-shopping-attachment script` failed: timeout 30000ms exceeded
```

https://github.com/GoogleForCreators/web-stories-wp/runs/6174672630?check_suite_focus=true

timarney:
> > @swissspidy I tested locally with your composer update commit merged back to this one it's s now working as expected (fails on the schema check string vs number)
> 
> Okay and which schema check is exactly failing? Can we fix our code?
> 
> I only see this error message:
> 
> ```
>   ● Shopping › Shopping schema › should should match a valid schema
> 
>     TimeoutError: waiting for selector `amp-story-shopping-attachment script` failed: timeout 30000ms exceeded
> ```
> 
> https://github.com/GoogleForCreators/web-stories-wp/runs/6174672630?check_suite_focus=true


Right for CI --- need to investigate why it's not finding that selector

Locally for me I'm getting a schema error because product price is returning a string vs the schema is looking for a number.  So yes that part can be fixed --- but maybe that fix should be in the product query PR?

swissspidy:
> So yes that part can be fixed --- but maybe that fix should be in the product query PR?

Works for me; just keep in mind that then the test needs to be disabled until that gets fixed.

includes/REST_API/Products_Controller.php
```diff
@@ -314,7 +314,8 @@ protected function get_items_shopify( $request ) {
 				'productBrand'         => $product['vendor'],
 				// TODO: Maybe eventually provide full price range.
 				// See https://github.com/ampproject/amphtml/issues/37957.
-				'productPrice'         => $product['priceRange']['minVariantPrice']['amount'],
+				// TODO: Replace hardcoded value with real price as a number (AMP schema validation).
+				'productPrice'         => 10,
```
swissspidy: Why this change instead of type casting or something?

timarney: Testing ...  if this passes will fix that up.

timarney: Successful run here https://github.com/GoogleForCreators/web-stories-wp/actions/runs/2229993362

Updated to cast as a `float` for now.

packages/e2e-tests/src/specs/editor/shopping/woocommerce.js
timarney: Added a snapshot here:

https://github.com/GoogleForCreators/web-stories-wp/pull/11250/commits/78a28b1f962742927056321fea8e7392a8f9fcad

swissspidy:
All e2e tests seem to be passing now! 🎉

swissspidy (approved):

swissspidy:
@spacedmonkey ping 👋  to review the latest version of this

.github/workflows/update-product-schema.yml
```diff
@@ -0,0 +1,40 @@
...
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 12 * * 1'
+
+env:
+  GIT_AUTHOR_EMAIL: 94923726+googleforcreators-bot@users.noreply.github.com
+  GIT_AUTHOR_NAME: googleforcreators-bot
+  GIT_COMMITTER_EMAIL: 94923726+googleforcreators-bot@users.noreply.github.com
+  GIT_COMMITTER_NAME: googleforcreators-bot
```
spacedmonkey, line 12: Should these be hardcoded or should they be setup as part of the project's env vars?

swissspidy: Great question. They are hardcoded like this in all of our GitHub Actions. This is just copy-pasted from other workflows. Don't currently see the need for moving these to env vars, but maybe I can do that in the future if we feel like it would make things easier.

PR #11251, opened 2022-04-12
Floating Menu: Add e2e tests

### LINKED ISSUE #10990: Floating Menus: Add e2e Tests
## Feature Description

Just noticed as I was setting up #10988 that we don't have any e2e tests for the new floating menu feature. We should make sure those are added. 

Ideally this would be before the feature flag is enabled but it wasn't scoped.

marianakyrkosh:
## Context

E2E tests for the floating menu

## Summary

Most of the functionality is covered by the karma tests. E2E tests are not that in depth.

## Testing Instructions

1. Run the e2e tests in the file floatingMenu.js

## Checklist

---

Fixes #10990


packages/e2e-tests/src/specs/editor/floatingMenu.js
```diff
@@ -0,0 +1,109 @@
...
+import {
+    createNewStory,
+} from '@web-stories-wp/e2e-test-utils';
+
+describe('Floating Menu', () => {
+    beforeEach(async () => {
+        await createNewStory();
+
+        // Wait for title input to load before continuing.
+        await page.waitForSelector('input[placeholder="Add title"]');
```
BrittanyIRL: nit: I think most of the other tests here do this: 
```
 await expect(page).toMatchElement('input[placeholder="Add title"]');
``` 
for the title

packages/e2e-tests/src/specs/editor/floatingMenu.js
```diff
@@ -0,0 +1,109 @@
...
+
+
+        // Wait for skeleton thumbnails in the carousel to render which gives footer time to also render
+        await page.waitForFunction(
+            () =>
+                !document.querySelector(
+                    'li[data-testid^="carousel-page-preview-skeleton"]'
+                ),
+            { timeout: 10000 } // requestIdleCallback in the carousel kicks in after 5s the latest.
+        );
```
BrittanyIRL: 🤔  Both `specs/wordpress/getStartedStory.js` and `specs/dashboard/templates/useTemplate.js` use something similar (except the timeout is 5,000 not 10,000) - typically when we see the same thing more than twice it ends up getting extracted into the e2e utils. Thoughts on doing that here?

marianakyrkosh: I've removed this function. It's not required.

packages/e2e-tests/src/specs/editor/floatingMenu.js
marianakyrkosh: I've removed this function. It's not required.

packages/e2e-tests/src/specs/editor/floatingMenu.js
```diff
@@ -0,0 +1,97 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * External dependencies
+ */
+import {
+    createNewStory,
+    waitForCarousel
```
merapi: I'm wondering what went wrong with the pre-commit hook/prettier here 🤔  (also there is `no-unused-vars` and `import/named` error).

Can you check if `npm run lint:js:fix` fixes the issues on your side?

image

packages/e2e-tests/src/specs/editor/floatingMenu.js
```diff
@@ -0,0 +1,97 @@
...
+    beforeEach(async () => {
+        await createNewStory();
+    });
+
+    it('should display text floating menu', async () => {
+        // Open a text tab
+        await expect(page).toClick('#library-tab-text');
+
+        // Add a paragraph
+        await expect(page).toClick('button[data-testid="preview-text"] span', {
```
merapi: `#library-pane-text button` should also work? without having the need to add testid.

packages/e2e-tests/src/specs/editor/floatingMenu.js
```diff
@@ -0,0 +1,97 @@
...
+            text: /^Paragraph/,
+        });
+
+        // Floating menu should show up
+        await expect(page).toMatchElement('div[data-testid="context-menu-list"]');
```
merapi: I think we should go here with something more specific (there can be other `context-menu-list` on the page that is not the floating menu).

Something like https://github.com/GoogleForCreators/web-stories-wp/blob/021448ee508994be9aabeb945113a90f83efa228/packages/story-editor/src/karma/fixture/containers/canvas.js#L83 ?

packages/e2e-tests/src/specs/editor/floatingMenu.js
marianakyrkosh: As far as I know it's good to use only test ids for tests, because for example ids can be changed and tests will fail.

packages/e2e-tests/src/specs/editor/floatingMenu.js
marianakyrkosh: I'm adding a specific testid to the floating menu wrapper

merapi: Ideally, we should locate the elements by something like a label, that way even if we change the underlying code (and e.g. remove ids/testids) the test will still work (and it has other benefits). @barklund WDYT?

packages/e2e-tests/src/specs/editor/floatingMenu.js
marianakyrkosh: @merapi I've used `#library-pane-text button` like you suggested

packages/e2e-tests/src/specs/editor/floatingMenu.js
marianakyrkosh: @merapi  I've removed the testid and  used `role="menu"`

packages/e2e-tests/src/specs/editor/floatingMenu.js
```diff
@@ -0,0 +1,102 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * External dependencies
+ */
+import { createNewStory } from '@web-stories-wp/e2e-test-utils';
+
+describe('Floating Menu', () => {
+  const floatingMenu = 'div[role="menu"]';
```
BrittanyIRL: 🤔 This might not be specific enough. There's more than 1 menu role present on the canvas. 

I think looking for the section with the associated aria label for the floating menu might be more reliable. The screenshot I attached is querying the role menu vs section with aria label. Thoughts? 


Screen Shot 2022-04-18 at 1 29 14 PM

merapi: @mariana-k I also suggest locating that element by the `Design menu` label (like in my first comment). We already have labels (we need them for a11y) so let's use them whenever possible (they point to very specific things, so we can avoid issues like this^).

packages/e2e-tests/src/specs/editor/floatingMenu.js
merapi: Ok, testids are not bad, but we don't plan to change ids or labels often (forever if possible) and a lot of tests can fail if we change any of them.

packages/e2e-tests/src/specs/editor/floatingMenu.js
marianakyrkosh: Done

packages/e2e-tests/src/specs/editor/floatingMenu.js
marianakyrkosh: Changed the selector to` aria-label="Design menu"`

merapi (approved):
We are good to go (just remove the `preview-text` testid?)

mariana-k:
> We are good to go (just remove the `preview-text` testid?)

Done

PR #11256, opened 2022-04-12
Design System: DataList partial search support

### LINKED ISSUE #10675: Author Search: Limited to matching at start
Mentioned by @barklund during bug bash. 

The author dropdown only searches for first name (or to be specific: by substring from the start of the name only), so you cannot search for a last name to select an author. That seems wrong here (while still useful for font search, where it comes from).

BrookeGraham:
## Context
The search util used by `DataList` was limited to filtering on the start of a string. Instead, we want to filter search so that if any portion of the option is present we still see it. 

## Summary
Updates option filtering function to include any portion of the option string. 

## Testing Instructions

1. Any search field using `DataList` should now be able to search for any part of the option. For example, in Fonts dropdown you can search for `shade` and see result for `bungee shade`. This should also work in the Dashboard's author dropdown search. We can now find authors by last name. 

no

no

no

## Checklist

---

Fixes #10675

### DIFF
--- a/packages/design-system/src/components/datalist/utils.js
+++ b/packages/design-system/src/components/datalist/utils.js
@@ -15,7 +15,7 @@
  */
 export const createOptionFilter = (options) => (keyword) =>
   options.filter(({ name }) =>
-    name.toLowerCase().startsWith(keyword.toLowerCase())
+    name.toLowerCase().includes(keyword.toLowerCase())
   );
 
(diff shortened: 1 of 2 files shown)


samwhale (approved):

BrittanyIRL:
Looks like this test needs updating: https://github.com/GoogleForCreators/web-stories-wp/runs/5993647078?check_suite_focus=true#step:6:13

BrittanyIRL (approved):
🐻  LGTM

PR #11270, opened 2022-04-14
Fix issues for server side rendering in Next.js

### LINKED ISSUE #11219: Errors rendering Editor in Next.js
## Bug Description

I was trying to create Story Editor in Next.js however there were several errors that appear one after the other ( Most of them are due to Next.js SSR )

image

( Gets fixed by removing [CSS import ](https://github.com/GoogleForCreators/web-stories-wp/blob/afe02f9209112c2447559dbfa797e3192b052f96/packages/story-editor/src/components/form/dateTime/datePicker.js#L20) )
Reference: https://nextjs.org/docs/messages/css-npm

image

( Gets fixed by using `globalThis` )

image

( Gets fixed by removing `"type": "module"` from `package.json` )

image

image

image

image

image

image

## Expected Behaviour

Editor and dashboard should work without any issues in NextJs.

## Steps to Reproduce

Step - 1: `npx create-next-app@latest`

Step - 2: Use snippet from [getting-started](https://github.com/GoogleForCreators/web-stories-wp/blob/main/docs/third-party-integration/story-editor/getting-started.md#step-3-use-the-storyeditor-component-to-render-the-editor) to use the editor

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

mohdsayed:
## Context

See #11219

## Testing Instructions

In web-stories-wp
1. Bundle and publish all packages locally.

In Next.js setup
1. `npx create-next-app@latest`
2. `npm i react@17 react-dom@17` as next.js is using v18
3. Use snippet from [getting-started](https://github.com/GoogleForCreators/web-stories-wp/blob/main/docs/third-party-integration/story-editor/getting-started.md#step-3-use-the-storyeditor-component-to-render-the-editor) to use the editor and dashboard
4. Install all the local dependencies ( or the ones which are updated in this PR )
```bash
npm --registry http://localhost:4873/ install --save \
@googleforcreators/media \
@googleforcreators/animation \
@googleforcreators/dashboard \
@googleforcreators/date \
@googleforcreators/design-system \
@googleforcreators/dom \
@googleforcreators/element-library \
@googleforcreators/fonts \
@googleforcreators/i18n \
@googleforcreators/masks \
@googleforcreators/migration \
@googleforcreators/moveable \
@googleforcreators/output \
@googleforcreators/react \
@googleforcreators/rich-text \
@googleforcreators/stickers \
@googleforcreators/story-editor \
@googleforcreators/templates \
@googleforcreators/text-sets \
@googleforcreators/tracking \
@googleforcreators/units \
@googleforcreators/url
@googleforcreators/elements
```
6. `npm run dev`

## Checklist

---

Fixes #11219


packages/story-editor/src/components/tabview/index.js
```diff
@@ -317,7 +317,12 @@ TabView.propTypes = {
   tabs: PropTypes.array.isRequired,
   tab: PropTypes.string.isRequired,
   tabRefs: PropTypes.objectOf(
-    PropTypes.shape({ current: PropTypes.instanceOf(Element) })
+    PropTypes.shape({
+      current:
+        typeof Element !== 'undefined'
+          ? PropTypes.instanceOf(Element)
+          : PropTypes.any,
+    }) // To handle SSR
```
swissspidy: Does `PropTypes.elementType` work instead?

packages/react/src/useFocusOut.js
```diff
@@ -53,7 +53,6 @@ function useFocusOut(ref, callback, deps) {
       node.removeEventListener('focusout', onFocusOut);
       doc.removeEventListener('pointerdown', onDocumentClick, opts);
     };
-    // eslint-disable-next-line react-hooks/exhaustive-deps -- Pass through provided deps.
```
swissspidy: We should update `additionalHooks` in the ESLint config so we can keep this

mohdsayed: I didn't remove this line, not sure why eslint removes it automatically

image

mohdsayed: I think this is why

image

swissspidy: After updating the `additionalHooks` config this should not happen anymore and we can add it back

sayedtaqui:
With these many updates, the editor and dashboard both are working in Next.js 🥳  .

packages/dashboard/src/utils/usePagePreviewSize.js
```diff
@@ -97,26 +97,28 @@ const sizeFromWidth = (
...
     ? MIN_DASHBOARD_WIDTH
     : windowWidth;
 };
 
 export default function usePagePreviewSize(options = {}) {
   const { thumbnailMode = false, isGrid } = options;
+  const container =
+    typeof document !== 'undefined' ? document.getElementById(WPBODY_ID) : null;
   // When the dashboard is pulled out of wordpress this id will need to be updated.
   // For now, we need to grab wordpress instead because of how the app's rendered
```
swissspidy: Sounds like something we should refactor now and move to `wp-dashboard` somehow...

packages/dashboard/src/utils/keyboardOnlyOutline.js
```diff
@@ -55,8 +55,11 @@ const KeyboardOnlyOutline = () => {
     }
   };
 
-  document.addEventListener('keydown', handleKeydown);
-  document.addEventListener('mousedown', handleMousedown);
+  // @todo Should this be in useEffect?
```
swissspidy: 👍

packages/story-editor/src/components/tabview/index.js
mohdsayed: It doesn't because `PropTypes.elementType` is for React Element.

image

packages/react/src/useFocusOut.js
mohdsayed: Oh I see, updated in https://github.com/GoogleForCreators/web-stories-wp/pull/11270/commits/843e32a8465dc8b8643c68345c9d9f44aa524bc7

packages/dashboard/src/utils/keyboardOnlyOutline.js
mohdsayed: Done in #11275

packages/dashboard/src/utils/usePagePreviewSize.js
mohdsayed: Missed out!  #11276

.eslintrc
```diff
@@ -685,6 +691,30 @@
...
         ]
       }
+    },
+    {
+      "files": [
+        "packages/story-editor/src/**/*.js",
+        "packages/dashboard/src/**/*.js",
+        "packages/design-system/src/**/*.js",
+        "packages/element-library/src/**/*.js"
+      ],
```
swissspidy: This probably affects a bunch more packages.

Might be easiest to do `"files": [ "packages/**/*.js" ],` and then add all non-editor packages (e.g. widget, lightbox, etc.) to `excludedFiles`

packages/dashboard/src/app/router/route.js
```diff
@@ -63,7 +63,7 @@ export function resolveRoute(route) {
      * If the url has no root append it to the current route and create
      * a nested root within the context of the dashboard app.
      */
-    return `${window.location.hash}/${route}`;
+    return `${globalThis.location?.hash}/${route}`;
```
swissspidy: On the server `globalThis.location?.hash` is `undefined`, so this will return `undefined/${route}`

Could mitigate this using something like

```suggestion
    const hash = globalThis.location?.hash || '';
    return `${hash}/${route}`;
```

packages/dashboard/src/app/router/routerProvider.js
```diff
@@ -83,8 +85,8 @@ function RouterProvider({ children, ...props }) {
         defaultRoute,
       },
       actions: {
-        push: history.current.push,
-        replace: history.current.replace,
+        push: history.current?.push,
+        replace: history.current?.replace,
```
swissspidy: Should we return `noop`s here instead of possibly `null`?

packages/dashboard/src/app/router/routerProvider.js
```diff
@@ -63,7 +65,7 @@ function RouterProvider({ children, ...props }) {
   };
 
   const [queryParams, setQueryParams] = useState(
-    parse(history.current.location.search)
+    parse(history.current?.location?.search)
```
swissspidy, line 69: If `history.current?.location?.search` is `null` then this will be `parse(null)` which will return `{ null: '' }`, which seems unexpected and like something that could possible cause bugs.

Might be worth adding JS unit tests with

```js
/**
 * @jest-environment node
 */
```

to test this provider

packages/dashboard/src/app/router/routerProvider.js
```diff
@@ -45,9 +45,11 @@ export const getActiveRoute = ({
   return matchingRoutes.length <= 1 ? defaultRoute : currentPath;
 };
 function RouterProvider({ children, ...props }) {
-  const history = useRef(props.history || createHashHistory());
+  const history = useRef(
+    props.history || (globalThis.document && createHashHistory())
+  ); // createHashHistory internally uses document.
   const [currentPath, setCurrentPath] = useState(
-    history.current.location.pathname
+    history.current?.location?.pathname
```
swissspidy: Perhaps

```suggestion
    history.current?.location?.pathname || ''
```

packages/design-system/src/components/keyboard/index.js
```diff
@@ -59,7 +59,8 @@ const globalRef = createRef();
 
 function setGlobalRef() {
   if (!globalRef.current) {
-    globalRef.current = document.documentElement;
+    globalRef.current =
+      typeof document !== 'undefined' ? document.documentElement : null;
```
swissspidy: Why not `globalThis.document?.documentElement || null`

Generally speaking, seems like this PR uses both different ways (`globalThis.document?.` vs `typeof document !== 'undefined'`). Can we settle on one?

packages/dom/src/stripHTML.js
```diff
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-const buffer = document.createElement('div');
+const buffer =
+  typeof document !== 'undefined' ? document.createElement('div') : {};
 
 export default function stripHTML(string) {
```
swissspidy, line 20: Do we need a `stripHTML` function that actually works on the server?

packages/element-library/src/video/onDropHandler.js
```diff
@@ -17,7 +17,10 @@
 function onDropHandler(dropTargetId) {
   // Auto-play video on drop
   setTimeout(() => {
-    const videoEl = document.getElementById(`video-${dropTargetId}`);
+    const videoEl =
+      typeof document !== 'undefined'
+        ? document.getElementById(`video-${dropTargetId}`)
+        : null;
```
swissspidy: Is the change here needed? This function should never be called on the server

packages/i18n/src/translateWithMarkup.js
```diff
@@ -120,8 +120,10 @@ function TranslateWithMarkup({ mapping = {}, children }) {
     );
   }
 
-  const node = new DOMParser().parseFromString(children, 'text/html').body
-    .firstChild;
+  const node =
+    typeof window !== 'undefined'
+      ? new DOMParser().parseFromString(children, 'text/html').body.firstChild
```
swissspidy: Ah, forgot about that.

Does this change break anything in `TranslateWithMarkup`?

Wonder if some Jest unit tests with `@jest-environment node` would be useful here as well.

packages/story-editor/src/app/font/actions/useLoadFontFiles.js
```diff
@@ -37,7 +37,10 @@ function useLoadFontFiles() {
     const handle = cleanForSlug(family);
     const elementId = `web-stories-${handle}-font-css`;
 
-    const hasFontLink = () => document.getElementById(elementId);
+    const hasFontLink = () =>
```
swissspidy: Is this change needed? Since this is within a callback and never called directly during render as far as I can see

packages/story-editor/src/app/story/effects/useHashState.js
```diff
@@ -34,7 +19,9 @@
 import { useEffect, useState } from '@googleforcreators/react';
 
 export function hashToParams(hash) {
-  return new URLSearchParams(hash.startsWith('#') ? hash.substr(1) : hash);
+  return new URLSearchParams(
+    hash && hash.startsWith('#') ? hash.substr(1) : hash
+  );
```
swissspidy:
```suggestion
  return new URLSearchParams(hash?.startsWith('#') ? hash.substr(1) : hash);
```

packages/story-editor/src/utils/getBlurHashFromImage.js
```diff
@@ -26,7 +26,8 @@ import Worker from './generateBlurhash.worker';
 
 const getImageData = (image) => {
   const { width, height } = image;
-  const canvas = document.createElement('canvas');
+  const canvas =
```
swissspidy: Is this change needed / is this function ever run during SSR?

packages/story-editor/src/utils/keyboardOnlyOutline.js
```diff
@@ -74,7 +74,9 @@ const KeyboardOnlyOutline = () => {
 };
 
 export function isKeyboardUser() {
```
swissspidy: This function is only ever called in an effect. Do we still need this?

packages/dashboard/src/app/router/routerProvider.js
mohdsayed: I am thinking, what would be the possible case when `history.current?.location?.search` may be `null`. We are only adding `.?` here to prevent errors during server-side rendering.

.eslintrc
mohdsayed: You are right. 
These are all the `@web-stories-wp` packages.

- e2e-tests
- activation-notice
- commander
- e2e-test-utils
- eslint-import-resolver
- glider
- imgareaselect
- jest-amp
- jest-puppeteer-amp
- jest-resolver
- karma-failed-tests-reporter
- karma-puppeteer-client
- karma-puppeteer-launcher
- stories-block
- stories-carousel
- stories-lightbox
- tinymce-button
- widget
- wp-dashboard
- wp-story-editor
- wp-utils

But not all have React as a dependency, so they would not be using `useEffect` anyways. These are the `@web-stories-wp` packages with React dependency, so added those in the `excludedFiles`.

- activation-notice
- jest-amp
- karma-fixture
- react
- stories-block
- test-utils
- tinymce-button
- wp-dashboard
- wp-story-editor
- wp-utils

packages/design-system/src/components/keyboard/index.js
mohdsayed: Yes, `globalThis.document?.documentElement || null` looks better here. At other places, I used `typeof document !== 'undefined'` because it looked more readable to write 

```js
typeof document !== 'undefined' ? document.getElementById(containerId) : null;`
```

as compared to 

```js
globalThis.document?.getElementById  ? document.getElementById(containerId) : null;`
```

Are you referring to any other instance?

packages/dom/src/stripHTML.js
mohdsayed: I think we do. Probably will have to use some external library/package for it.

packages/dashboard/src/app/router/routerProvider.js
swissspidy: During server-side rendering it will be `null`, so the initial state value would be `null` here, no?

packages/design-system/src/components/keyboard/index.js
swissspidy: I don't recall all the instances of either style. I just want to make sure we use a consistent style across the code base.

packages/dom/src/stripHTML.js
mohdsayed: There are packages available like [himalaya](https://www.npmjs.com/package/himalaya) ( 375kB ) but not very lightweight.

packages/i18n/src/translateWithMarkup.js
mohdsayed: It wouldn't break but return empty strings in the node env. 
There are alternatives like https://www.npmjs.com/package/node-html-parser, but that's 149kB

packages/dom/src/stripHTML.js
swissspidy: Wouldn't it need something more like [string-strip-html](https://www.npmjs.com/package/string-strip-html)?

Still quite large though, so as long as we don't need it, we can postpone this.

packages/dashboard/src/app/router/routerProvider.js
mohdsayed: In node env, the initial state would be `undefined` therefore it will return an empty object `{}`. Am I missing something? Will add test coverage with `@jest-environment node`

swissspidy: Ah indeed, if it's `undefined` and not `null` then it should be fine.

mohdsayed: Was trying to add test coverage for `@jest-environment node` however it doesn't look straightforward as it wouldn't run the same way as it does in SSR. When running a test in a node environment it will also run code in `useEffect`.

Though [React Hooks Testing Library](https://react-hooks-testing-library.com/) does provide [SSR support](https://react-hooks-testing-library.com/usage/ssr) for `renderHook`, however that doesn't seem to work as expected. Let me see if I can find a solution.

packages/design-system/src/components/keyboard/index.js
mohdsayed: Updated all instances in https://github.com/GoogleForCreators/web-stories-wp/pull/11270/commits/438d0fa0537d351bae387dcbf77d54938f9dbdb8

packages/dashboard/src/app/router/routerProvider.js
mohdsayed: Didn't find any solution for adding test coverage for node env! The doc suggests calling `hydrate()` after `renderHook()`, but that throws `document is not defined` error.

packages/dom/src/escapeHTML.js
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-const buffer = document.createElement('div');
+const buffer = globalThis?.document?.createElement('div') || {};
```
swissspidy:
```suggestion
const buffer = globalThis?.document?.createElement('div') || {}; // To handle SSR.
```

packages/dom/src/stripHTML.js
```diff
@@ -14,8 +14,9 @@
  * limitations under the License.
  */
 
-const buffer = document.createElement('div');
+const buffer = globalThis?.document?.createElement('div') || {};
```
swissspidy:
```suggestion
const buffer = globalThis?.document?.createElement('div') || {}; // To handle SSR.
```

packages/story-editor/src/app/canvas/canvasProvider.js
```diff
@@ -62,18 +62,20 @@ function CanvasProvider({ children }) {
...
-          }
-          setBoundingBoxes((boxes) => ({
-            ...boxes,
-            [entry.target.dataset[RECT_OBSERVATION_KEY]]:
-              entry.boundingClientRect,
-          }));
-        }
-      }),
+      globalThis.window
+        ? new window.IntersectionObserver((entries) => {
```
swissspidy: Shouldn't this be like this?

```suggestion
      globalThis.IntersectionObserver
        ? new globalThis.IntersectionObserver((entries) => {
```

santoshgistto:
Looks like  test coverage dropped :)

santoshgistto:
Looks like needs to increase code coverage still 
@sayedtaqui  any help here :)

swissspidy:
Closing for now, best to review this with a fresh PR at some point.

PR #11272, opened 2022-04-14
Fix story saving after removing background audio

### LINKED ISSUE #11229: Cannot save story after removing background audio
## Bug Description

When removing background audio from a page, `page.backgroundAudio` becomes `null`.

When generating the story markup upon saving, `backgroundAudio` is referenced a few times here:

https://github.com/GoogleForCreators/web-stories-wp/blob/bbcf177fb35dcc6fb6abd5612de2e97e470bb3ba/packages/output/src/page.js#L107-L112

This will cause `Cannot read properties of null` errors.

**Solution:** To fix this we should use optional chaining there.

⚠️ Need to add tests to prevent regressions.

❓  One might think that this default value would catch this:

https://github.com/GoogleForCreators/web-stories-wp/blob/bbcf177fb35dcc6fb6abd5612de2e97e470bb3ba/packages/output/src/page.js#L54-L55

No. Default values are only used if the value is `undefined`; not when it is `null`.

Let's remove the default value for `backgroundAudio` and `pageAttachment` to avoid this kind of confusion.

## Expected Behaviour

After removing background audio, saving a story should still work.

## Steps to Reproduce

1. Add background audio to page
2. Remove background audio again
3. Try to save story
4. See error message

## Screenshots

Screenshot 2022-04-11 at 15 37 10

## Additional Context

- Plugin Version: 1.19.0
- WordPress Version: 5.9
- Operating System: macOS
- Browser: Chrome

timarney:
## Context

When removing background audio from a page, page.backgroundAudio becomes null and in turn the Story can't save.

162777840-ee0884d5-5db7-495f-a3d7-275e70919b91

See: https://github.com/GoogleForCreators/web-stories-wp/issues/11229#issue-1200164316

## Summary
- Adds optional chaining
- Removes default values `{}`
- Add e2e test to match the add / remove background audio use case

## Testing Instructions

1. Add background audio to page
2. Remove background audio again
3. Save story

No

No

No

## Checklist

---

Fixes #11229

### DIFF
--- a/packages/e2e-tests/src/specs/editor/backgroundAudio.js
+++ b/packages/e2e-tests/src/specs/editor/backgroundAudio.js
@@ -24,6 +24,8 @@ import {
   takeSnapshot,
   withExperimentalFeatures,
   withPlugin,
+  insertStoryTitle,
+  publishStory,
 } from '@web-stories-wp/e2e-test-utils';
 
 const VTT_URL = `${process.env.WP_BASE_URL}/wp-content/e2e-assets/test.vtt`;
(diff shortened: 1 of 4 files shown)


packages/e2e-tests/src/specs/editor/backgroundAudio.js
```diff
@@ -217,5 +219,36 @@ describe('Background Audio', () => {
         });
       });
     });
+
+    describe('Add and remove background audio', () => {
```
swissspidy, line 223: ❤️

swissspidy:
Let's also add a JS unit test to `packages/output/src/test/page.js` with `backgroundAudio` being `null`, and one with `pageAttachment` being null.

Otherwise this looks good 👍

packages/output/src/test/page.js
```diff
@@ -1384,6 +1384,26 @@ describe('Page output', () => {
...
+        autoAdvance: false,
+        defaultPageDuration: 7,
+      };
+
+      const { container } = render(<PageOutput {...props} />);
+      const page = container.querySelector('amp-story-page');
+      await expect(page).toBeInTheDocument();
+      expect(page).not.toContain(
+        'background-audio="https://example.com/audio.mp3"'
+      );
```
swissspidy:
```suggestion
      expect(page).not.toContain('background-audio=');
```

A little bit more robust

swissspidy (dismissed):

spacedmonkey (approved):

PR #11279, opened 2022-04-15
Editor: Fixed bugs when panning flipped background image

### LINKED ISSUE #10723: RTL: Background image - flip horizontally not working as expected
## Bug Description
Originally reported by  @BrookeGraham:
When selecting flip background horizontally, and you attempt to move the background, the image is backwards. 

## Additional Context

- Plugin Version:
background-flip

- WordPress Version:
- Operating System:
- Browser:

barklund:
## Summary

There was a weird bug when you had a flipped image and trying to pan it. It seemed a bit inconsistent and was reported as only a bug in RTL, but was actually a bug also in LTR.

The main confusion arises from the fact, that the problem only occurs on the second drag inside the pan- container. The problem occured with both vertical and horizontal flips.

This PR fixes this, by updating as soon as the contain loads (and removes the double-flipping of coordinates).

This also fixes a related issue, which occurred if flipping while already editing a background media element. The flipped image would only actually appear flipped once dragged at least one time after pressing the flip toggle.

## User-facing changes

| Direction | Before | After |
|-|-|-|
| LTR | bg-pan-ltr-before | bg-pan-ltr-after |
| RTL | bg-pan-rtl-before | bg-pan-rtl-afte |

## Testing Instructions

#### Panning a flipped background image
1. Add a background image (preferably landscape format)
2. Flip it (either and both directions)
3. Double-click the image to enter edit-mode
4. Pan the image right and left. Make sure to pan multiple times. Observe that panning works correctly
5. Do the above in both LTR and RTL writing modes

#### Flipping a background image while editing
1. Add a background image
2. Double-click the image to enter edit-mode
3. Click the flip-toggles in either the style panel or the floating design menu
4. Observe that flipping looks correct while in edit-mode
5. Do the above in both LTR and RTL writing modes

## Checklist

---

Fixes #10723


packages/element-library/src/media/editPanMoveable.js
```diff
@@ -102,14 +104,7 @@ function EditPanMoveable({
...
       throttleDrag={0}
       onDrag={({ dist }) => {
-        let [tx, ty] = dist;
-        if (flip?.vertical) {
-          ty = -ty;
-        }
-        if (flip?.horizontal) {
-          tx = -tx;
-        }
-        translateRef.current = [tx, ty];
```
barklund, line 112: Note that the flip toggles was taken into consideration twice in this file - once in `onDrag` and again in `onDragEnd`. By making sure to apply the correctly transform from the start, this stuff here is not necessary.

packages/element-library/src/media/editPanMoveable.js
```diff
@@ -61,6 +61,8 @@ function EditPanMoveable({
     moveableRef.current?.updateRect();
   });
 
+  useEffect(update, [update]);
```
barklund, line 64: This runs on load, but also if the flip toggles change while the edit-mode is mounted, which fixes the second bug mentioned.

merapi (approved):
It works flawlessly now, I tried to break it in every possible way - still strong 💪🏻  (can we skip QA if I did "Testing Instructions" and even more?).

PR #11285, opened 2022-04-18
Canvas: Support selecting element by clicking on the border

### LINKED ISSUE #5325: Allow selecting element by clicking on the outer border

marianakyrkosh:
## Context

An element should be selected by clicking on it's border too.

## Summary

Previously a border of an element was not clickable and one should have clicked on the element itself to select it, not on its border. Now clicking on a border selects an element.

## User-facing changes

Before:
image

After:
image

## Testing Instructions

1. Add an element to the story, e.g. a shape or an image.
2. Add a border to the element.
3. Click on the border.
4. The element is selected.

## Checklist

---

Fixes #5325


merapi (changes requested):
It shouldn't change the rendering of the element like this (look at the image crop in my example), the element should still look like the one on the Right but the selection box should contain the border.

Left: this PR, Right: main branch
image

swissspidy:
FYI I updated the PR title because we don't usually prefix PR titles with the issue number. Makes changelogs a bit easier to read.

merapi:
Will be fixed here: https://github.com/GoogleForCreators/web-stories-wp/pull/11628

Issue #11252, opened 2022-04-12
Shopping: Floating Menu


swissspidy:
## Feature Description

* [**Figma**](https://www.figma.com/file/ZrO763yvMWJjYeAPsRnNfW/_Editor---Sprint-Feature-Level-Designs_2022?node-id=503%3A15921)

Adding a dropdown to the floating menu for products that allows changing the displayed product

<img width="283" alt="Screenshot 2022-04-12 at 14 41 23" src="https://user-images.githubusercontent.com/841956/162964552-60859347-70ef-4a8b-ba31-ab8dca066d72.png">

Trash button: removes the element

PR #11289, opened 2022-04-18
Editor: Hide design menu if not enabled for element type

### LINKED ISSUE #11116: Floating Menu: Do not render menu if there are no items
## Feature Description

If an element type does not specify floating menu items, an empty menu bar is being rendered.

Screenshot (see the gray rectangle below the white dot):

Screenshot 2022-03-29 at 16 45 21

We should avoid that.

Ideally we move components that render these items to the element definition so that the element defines what menu items it has. Similar to how it defines its list of supported panels.

Related:

* #11064
* #10865

barklund:
## Context

This adds a new config option to element types, that hides the design menu if not enabled for a particular element type. Note that this setting only applies to single selection of said element type. If the user multi-selects either solely or partially elements with this setting off, the multi-select menu will still appear.

## User-facing changes

| Before | After |
|-|-|
| Screen Shot 2022-04-18 at 11 56 01 | Screen Shot 2022-04-18 at 11 55 39 |

## Testing Instructions

1. Enable shopping experiment and set the proper settings
2. Add a product to the page
3. Observe that no design menu is displayed just below the menu

## Checklist

---

Fixes #11116

### DIFF
--- a/packages/element-library/src/product/constants.js
+++ b/packages/element-library/src/product/constants.js
@@ -35,6 +35,8 @@ export const defaultAttributes = {
 
 export const hasEditMode = false;
 
+export const hasDesignMenu = false;
+
 export const isMedia = false;
 
 export const canFlip = true;
--- a/packages/element-library/src/gif/constants.js
+++ b/packages/element-library/src/gif/constants.js
@@ -29,6 +29,7 @@ export {
   isMaskable,
   isMedia,
   hasEditMode,
+  hasDesignMenu,
   editModeGrayout,
 } from '../media/constants';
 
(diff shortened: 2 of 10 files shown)


packages/element-library/src/product/constants.js
```diff
@@ -35,6 +35,8 @@ export const defaultAttributes = {
 
 export const hasEditMode = false;
 
+export const hasDesignMenu = false;
```
swissspidy, line 38: Just noting that this will change soon-ish in #11252.

But this PR still makes sense to make the design menu more robust, so thanks for tackling this!

merapi (approved):

merapi (approved):
👍🏻 

I think we can skip QA on this?

PR #11302, opened 2022-04-18
Performance: Saved template image generation integrate with page canvas provider

### LINKED ISSUE #10821: Performance: Saved template image generation integrate with page canvas provider

maxyinger:
## Context
Allows saved page templates to access the page canvas cache when clicking `save page as template`.

## Summary
`save page as template` button will now use the existing page canvas if it exists instead of generating a new canvas from the page. If there's no up to date page canvas in the cache, then the saved template will generate the canvas itself and create the image off of that.

**TLDR**
Save page as templates now takes advantage of the shared canvas cache.

## Relevant Technical Choices
we're only accessing the canvas cache for generating the dataURL so the `save page as template` button checks to see if there's a valid canvas for this page, and if there is, converts it to a dataURL so the new saved template thumb has access to the image on first render.

The saved template thumb is responsible for taking that dataURL, converting it to a .png file, and uploading it to WP. The reason I went with making the saved template thumb responsible for uploading the image to WP was to reduce latency when clicking the button (with this approach, we don't have to wait for more requests to come back before adding the template to the saved templates panel)

## To-do
NA

## User-facing changes
NA

## Testing Instructions
You could always `console.log` to see when functions are firing, otherwise it can be a little hard to identify whats an image and what's a fully rendered story. regardless, here's how the cache behaves.

**Populating page canvas cache before creation**
- If you're on a story page, then switch to another page, the canvas thumb should now generate an image for the non-active page and store it in the cache.
- Travel back to the original page once the canvas thumb has generated the image for it
- Go to templates and click `add page as template`
- You should see no calls to `html-to-canvas` cus the canvas already exists in the cache
- The saved template should just pull a dataURL off the canvas and use that as the image instead of generating a canvas
- the saved template should generate a .jpg file off of that dataURL, and post to WP to store that .jpg with the saved template post
- the saved template should receive a `src` back that points to the image now on the server and use that to render the image
- the saved template should always render the image from the server on all subsequent loads.

**Note**
It's possible to populate the canvas cache with the current page via smart colors right now to. One way to test this is to go to the header presets and turn on smart colors, then hover over a preset, this should generate a canvas for the cache so that we can read color accessibility info off the page. If you don't actually add one of the presets & just hover over it, the cache for the current page should not be broken and can be used for the saved template image.

**No canvas in cache**
If there's no cache hits for the current version of the page you're creating, the saved template thumbnails will behave as they did before and generate a canvas from the fully rendered story themselves.

1.

## Checklist

---

Fixes #10821

### DIFF
--- a/packages/story-editor/src/components/library/panes/pageTemplates/savedPageTemplate.js
+++ b/packages/story-editor/src/components/library/panes/pageTemplates/savedPageTemplate.js
@@ -111,9 +111,9 @@ function SavedPageTemplate(
   const queuePageImageGeneration = usePageDataUrls(
     ({ actions }) => actions.queuePageImageGeneration
   );
-  const pageDataUrl = usePageDataUrls(
-    ({ state: { dataUrls } }) => dataUrls[page.id]
-  );
+  const pageDataUrl =
+    usePageDataUrls(({ state: { dataUrls } }) => dataUrls[page.id]) ||
+    page.pregeneratedPageDataUrl;
   const [isActive, setIsActive] = useState(false);
 
   useFocusOut(ref, () => setIsActive(false), []);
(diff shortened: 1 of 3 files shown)


BrookeGraham (approved):
Nice. Thanks for the detailed testing / how it works 🤯 Feels great.

BrittanyIRL (approved):
Awesome!!!

PR #11308, opened 2022-04-19
REST API: Prime attachment caches all at once

spacedmonkey:
## Summary

Discovered while working on other things. Instead of loading term and meta one at a time in the REST API, prime all caches in one go. This remove over half queries on editor. 

#### Before
Screenshot 2022-04-19 at 14 59 37

#### After
Screenshot 2022-04-19 at 14 59 50

## Testing Instructions

1.

## Checklist

---

Fixes #


spacedmonkey:
> Is it somehow possible to test the performance aspect of this at all, i.e. that the prime_post_caches method is being called and has an affect? Perhaps by comparing wpdb::get_num_queries() or something?

I thought about that, but what regression would that be testing? After we implement this functionality when will that number will hopefully never change. What is the point of run this test every time we make a php change? Does that test add value?

swissspidy:
It's less about the exact number of queries but more about verifying that this change actually does prime the cache. So that we don't accidentally regress this again in the future.

spacedmonkey:
> It's less about the exact number of queries but more about verifying that this change actually does prime the cache. So that we don't accidentally regress this again in the future.

But then we are testing `_prime_post_cache`? The logic we are creating, is just to get the ids that need to be primed, which tests already prove work. Once those ids are passed to `_prime_post_cache` it takes over.

swissspidy:
Perhaps more testing that `_prime_post_cache` is called

I was just thinking that it would be too easy to introduce a regression here and wondering how we can prevent that.

If there's no good way to do that in WP then that's fine by me.

spacedmonkey:
> If there's no good way to do that in WP then that's fine by me.

I can't find a good way to get test. Any query count maybe broken by newer versions of WP, as they may run another query. 
We know the ids that are being passed to the function, that is the best we can do.

swissspidy (approved):

swissspidy (approved):

PR #11280, opened 2022-04-15
Design System: Tooltip separation

BrookeGraham:
## Context
`Tooltip`s currently use `Popup` for it's portal, this has lead to several regressions, weird single use prop edge cases, and is very hard to maintain `Popup` vs `Tooltip`. We are pulling `Popup` from the `Tooltip` component in order to separate concerns a little better. 

## Summary
Let `Tooltip` create it's own portal. 

## Relevant Technical Choices
Now the `Dashboard` and `Story Editor` each pull in the `BaseTooltip` from `Design System` and export their own version of the `Tooltip`. This is so that we can grab `isRTL` and `leftOffset` from `useConfig` and send down the correct derived placement and offsets without having to manually grab these every time we want to use a `Tooltip`. All current uses of the Tooltip have been updated to come from their respective package. If `BaseTooltip` gets used directly from the `Design System` for example with the `Media Input` where it gets used in both places (editor and dashboard) you will have to pass along `isRTL` and `leftOffset`. Speaking of `leftOffset`.

We are utilizing `leftOffset` from the config to account for the `MENU_FOLDED_WIDTH`. I choose the `MENU_FOLDED_WIDTH` as opposed to `MENU_WIDTH` since it is smaller, and when the menu is expanded, you will usually not obfuscate the menu items with a tooltip. 

In the Dashboard, the `Tooltip` has `ignoreMaxOffsetY` set to true always. We can remove this with #11200 

Placement of the `Tooltip` is no longer "flipped" if rendered offscreen, for example when in RTL.  Now the offset will always force you inside the screen. 

## To-do

~~Edge case with `isRTL` arrow setting on `right` and `left` placements~~

I noticed that some implementation of `Popup` are not getting correct `isRTL`. Will create a follow up ticket to address those.  Ticket created #11303 
## Testing Instructions

- [ ] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1. Just regression, we want to make sure all tooltip display as expected and that `Popup`s didn't regress. 

### Does this PR have a security-related impact?
no

### Does this PR change what data or activity we track or use?
no

### Does this PR have a legal-related impact?
no

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/googleforcreators/web-stories-wp/blob/main/docs/accessibility-testing.md))
[...]

PR #11325, opened 2022-04-21
Design System: Popup clean up

### LINKED ISSUE #11201: Layers Panel Using Wrong Tooltip
## Bug Description

Layers panel is getting negative values for some of the positioning because it's using the Tooltip outside of story-editor. 

Use Tooltip from Story-Editor, and update the placement to bottom-end since we are at the left edge of the screen anyways.

### LINKED ISSUE #11303: Insert Media Popup not getting correct `isRTL` placement
## Bug Description

On the Insert Media Side Panel,  you click the + icon on an image and it opens a `Popup` to "insert the image" or add it as the background. When in RTL mode, the Popup is cutoff. Probably cause we are not sending `isRTL` to the Popup and thus not changing the placement. 

## Expected Behaviour

Popups shouldn't get cutoff, or layered underneath the admin sidebar. 
## Steps to Reproduce
Open Editor
Switch to RTL
Open Media panel
Click on the + icon that is shown over any image

## Screenshots
Screen Shot 2022-04-15 at 3 37 45 PM

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

BrookeGraham:
## Context
This is some follow up work related to the Layers Panel `Tooltip` getting negative valued x-offsets. Since the Layers Panel has some very nested absolute positioning the `getOffset` util thinks the `Tooltip` should be based off somewhere that the panel isn't. `Tooltip` offset updates, edge cases, etc where fixed in #11280. However, we still needed to address the `getOffset` to account for edge cases in `Popups` that were a work-around for those Tooltips.    

We had a special prop flag `resetXOffset`, to know when to look for the edge cases where this was an issue. Now that `Tooltip` is separate we can remove the `resetXOffset` flag and better handle any off screen issues directly in the `Popup` component. 

## Summary

`Popup` now updates the x-offset in cases where it would either collide with admin side nav, or otherwise get cutoff screen in x. 

## Relevant Technical Choices
Removed `resetXOffset` from the shared util `getOffset`. Both `Tooltip` and `Popup` use this function to grab offset values for portals. I wanted to keep this portion straightforward.  Since `Tooltip` and `Popup` have varied `placement` considerations I wanted to keep any special offset updates or edge case logic specific to each component. The `getOffset` will still give us `maxOffsetX` to prevent us from cutoff on the right edge, however,  `Popup` now takes `leftOffset` into consideration for RTL mode. `leftOffset` is also being utilized for LTR sidenav bar consideration. 

~~The bulk of these files was just adding `isRTL` and `leftOffset` to all uses of `Popup`.~~

New `usePopup` hook and `PopupProvider`. This provider now gives us access to `leftOffset`, `topOffset` and `isRTL` so we no longer need to pass these as props to `Tooltip` or `Popup` 🎉
 

## User-facing changes

Now that `Insertion Menu` is receiving `isRTL` the popup placement is correct in the RTL view as seen below. 
|Before|After|
|--|--|
|Screen Shot 2022-04-21 at 8 31 11 AM|Screen Shot 2022-04-21 at 8 29 54 AM|

## Testing Instructions

1. In RTL, open the Insert panel and hover over an image. 
2. Click the `+` icon  to open the `Insertion Menu`
3. The Popup should no longer be cutoff by the wp-admin bar
4. While we are here, let/s check for Popup regressions. 
5. Select / add text on the canvas and open the `Style` Panel
6. Open the color picker, (any version of it),  save styles, Font dropdown
7. We can also check the dropdowns in Document tab and the Publish Time Popup. 
8. Double check Dropdowns in the Dashboard and Dashboard settings as well.  

no

no

no

## Checklist

---

Fixes #11201 
Fixes #11303

### DIFF
--- a/packages/design-system/src/components/popup/index.js
+++ b/packages/design-system/src/components/popup/index.js
@@ -32,6 +32,7 @@ import PropTypes from 'prop-types';
  * Internal dependencies
  */
 import { noop } from '../../utils';
+import { usePopup } from '../../contexts';
 import { getTransforms, getOffset } from './utils';
 import { PLACEMENT, PopupContainer } from './constants';
 const DEFAULT_TOP_OFFSET = 0;
@@ -39,7 +40,6 @@ const DEFAULT_POPUP_Z_INDEX = 2;
 const DEFAULT_LEFT_OFFSET = 0;
 
 function Popup({
-  isRTL = false,
   anchor,
   dock,
   children,
@@ -49,11 +49,15 @@ function Popup({
   isOpen,
   fillWidth = false,
   refCallback = noop,
-  topOffset = DEFAULT_TOP_OFFSET,
-  leftOffset = DEFAULT_LEFT_OFFSET,
   zIndex = DEFAULT_POPUP_Z_INDEX,
   ignoreMaxOffsetY,
 }) {
+  const {
+    topOffset = DEFAULT_TOP_OFFSET,
+    leftOffset = DEFAULT_LEFT_OFFSET,
+    isRTL = false,
+  } = usePopup();
+
   const [popupState, setPopupState] = useState(null);
   const isMounted = useRef(false);
   const popup = useRef(null);
@@ -196,7 +200,6 @@ function Popup({
 }
 
 Popup.propTypes = {
-  isRTL: PropTypes.bool,
   anchor: PropTypes.shape({ current: PropTypes.instanceOf(Element) })
     .isRequired,
   dock: PropTypes.shape({ current: PropTypes.instanceOf(Element) }),
@@ -207,8 +210,6 @@ Popup.propTypes = {
   isOpen: PropTypes.bool,
   fillWidth: PropTypes.oneOfType([PropTypes.number, PropTypes.bool]),
   refCallback: PropTypes.func,
-  topOffset: PropTypes.number,
-  leftOffset: PropTypes.number,
   zIndex: PropTypes.number,
   ignoreMaxOffsetY: PropTypes.bool,
 };
--- a/packages/design-system/src/components/popup/index.js
+++ b/packages/design-system/src/components/popup/index.js
@@ -86,7 +86,7 @@ function Popup({
       const updatedXOffset = () => {
         // When in RTL the popup could render off the left side of the screen. If so, let's update the
         // offset to keep it inbounds.
-        if (x <= leftOffset) {
+        if (x < leftOffset) {
           switch (placement) {
             case PLACEMENT.BOTTOM_END:
             case PLACEMENT.TOP_END:
@@ -111,7 +111,7 @@ function Popup({
           }
         }
 
-        if (isRTL && right >= offset.bodyRight - leftOffset) {
+        if (isRTL && right > offset.bodyRight - leftOffset) {
           // maxOffset should keep us inbounds, expect in the case of RTL due to the admin-sidebar nav we could use another
           // switch case here to make offset more precise, however the math below will always return the popup fully in screen.
           return { ...offset, x: offset.bodyRight - width - leftOffset };
@@ -208,6 +208,7 @@ Popup.propTypes = {
   fillWidth: PropTypes.oneOfType([PropTypes.number, PropTypes.bool]),
   refCallback: PropTypes.func,
   topOffset: PropTypes.number,
+  leftOffset: PropTypes.number,
   zIndex: PropTypes.number,
   ignoreMaxOffsetY: PropTypes.bool,
 };
--- a/packages/design-system/src/components/popup/index.js
+++ b/packages/design-system/src/components/popup/index.js
@@ -34,8 +34,9 @@ import PropTypes from 'prop-types';
 import { noop } from '../../utils';
 import { getTransforms, getOffset } from './utils';
 import { PLACEMENT, PopupContainer } from './constants';
-const DEFAULT_TOPOFFSET = 0;
+const DEFAULT_TOP_OFFSET = 0;
 const DEFAULT_POPUP_Z_INDEX = 2;
+const DEFAULT_LEFT_OFFSET = 0;
 
 function Popup({
   isRTL = false,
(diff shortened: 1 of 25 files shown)


packages/design-system/src/components/popup/index.js
```diff
@@ -68,21 +69,61 @@ function Popup({
...
+            case PLACEMENT.RIGHT:
+              // These should only matter in the case of isRTL so we'll need the entire width of the popup as the offset
+              return { ...offset, x: width };
+            default:
+              return { ...offset, x: width / 2 + (isRTL ? 0 : leftOffset) };
+          }
+        }
+
+        if (isRTL && right > offset.bodyRight - leftOffset) {
+          // maxOffset should keep us inbounds, expect in the case of RTL due to the admin-sidebar nav we could use another
```
BrookeGraham: spelling `except`

packages/design-system/src/components/popup/utils/getOffset.js
```diff
@@ -134,12 +129,7 @@ export function getOffset({
   const { x: spacingH = 0, y: spacingV = 0 } = spacing || {};
 
   // Horizontal
-  const offsetX = () => {
-    if (resetXOffset && popupRect?.left <= 0) {
-      return isRTL ? width : 0;
-    }
```
BrookeGraham, line 140: Previously, this was to handle the color picker in the floating menu when in RTL. We couldn't directly update the offsets since `Tooltip` can sometimes come back with `popupRect?.left` being in the negative. This happen in the layers panel due to the nested absolute positioning. Now, Tooltip handles its own updated offset logic to deal account for this possibility.

BrittanyIRL (dismissed):
Makes sense to me ⭐

BrittanyIRL (dismissed):

felipebochehin87:
Open question on #11201.

packages/design-system/src/components/popup/utils/getOffset.js
```diff
@@ -152,7 +142,7 @@ export function getOffset({
 
   // Clamp values
   return {
-    x: Math.max(0, Math.min(offsetX(), maxOffsetX)),
+    x: Math.max(0, Math.min(offsetX, maxOffsetX)),
```
maxyinger, line 145: Heads up, we got this util to take care of this math. The util was probably created after the initial popup was made:
https://github.com/GoogleForCreators/web-stories-wp/blob/9404d909d2cda1afc17e00570944f72e9ff630a9/packages/units/src/range.js#L60-L62

packages/wp-story-editor/src/components/documentPane/publish/publishTime.js
```diff
@@ -133,6 +134,7 @@ function PublishTime() {
       </Row>
       <Popup
         topOffset={TOOLBAR_HEIGHT}
+        leftOffset={leftOffset}
```
maxyinger: So we're passing all of this through props cus we don't have access to the config provider in the Design System right?

I'm wondering if it maybe helps just to have a `PopupProvider` exported from the design system that takes care of these props. something we can use in the app like:
```js
import { PopupProvider } from '@googleforcreators/design-system';

function StoryEditor() {
  const { styleConstants: { leftOffset } = {} } = useConfig();
  // ...
  return (
    <PopupProvider value={{
        // pass whatever else we want here that we want from
        // providers in `story-editor` that we don't want to have 
        // to pass explicitly through props in every `Popup` Instance
        leftOffset,
        // ...
      }}>
      // ...
    </PopupProvider>
  );
}
```

then in the popup implementation we can just do something like this and not have to pass a prop to every instance:
```
function Popup(props) {
  const leftOffset = usePopupContext(({ leftOffset }) => leftOffset);
  // ...
}
```

maxyinger: could also use this approach to pass in things like `isRTL` & `topOffset`

maxyinger (dismissed):

packages/wp-story-editor/src/components/documentPane/publish/publishTime.js
BrookeGraham: Smart

packages/design-system/src/contexts/popup/popupProvider.js
```diff
@@ -0,0 +1,38 @@
...
+
+PopupProvider.propTypes = {
+  children: PropTypes.node,
+  value: PropTypes.shape({
+    isRTL: PropTypes.bool,
+    leftOffset: PropTypes.number,
+  }),
+};
+
+export default PopupProvider;
```
maxyinger:
```suggestion
import Context from './context';

export default Context.Provider;
```

packages/story-editor/src/components/form/color/colorInput.js
```diff
@@ -205,13 +206,14 @@ const ColorInput = forwardRef(function ColorInput(
...
   );
-  const { isRTL, styleConstants: { topOffset } = {} } = useConfig();
   const [dynamicPlacement, setDynamicPlacement] = useState(pickerPlacement);
 
   const {
     refs: { sidebar },
   } = useSidebar();
 
+  const { topOffset } = usePopup();
+
```
maxyinger: I would just pull `topOffset` off of `useConfig` since this can access it directly. Also prevents some bleeding of functionality into other things (makes `PopupProvider` solely part of the `Popup` implementation and not needed for anything else)

maxyinger (dismissed):
looks good!

packages/story-editor/src/components/form/color/colorInput.js
BrookeGraham: That were I was leaning too. Will do!

packages/design-system/src/components/popup/utils/getOffset.js
BrookeGraham: I decided against grabbing this from the units package since we don't already have that package as a dep for the design system package. Seemed like a lot to pull in for this one simple func. 🤷

maxyinger (approved):

mwritter (approved):
Looks good to me 👍

PR #11341, opened 2022-04-22
Design System: Popup and color picker props clean up

### LINKED ISSUE #11202: Design System: Dropdown Popup
## Feature Description

Creating a DropdownPopup will decrease the variety of general Popup usage. The Dropdown Search and DataList all have very similar use cases, feature, and overall feel for the Popup.

BrookeGraham:
## Context

Just a little more code quality props clean up for Popup. 

While we were going to pull the `Popup` component out of `Dropdown` we decided against that decision. `Dropdown's` use of the Popup, is pretty standard, and pulling it out would only lead to duplicated code. More so now that the logic that was previously for `Tooltip` has been removed from `Popup`. 

I used this ticket to removed props that were specific to Dropdown in the past and are no longer needed or used. I also removed a single use prop for the color picker while I was here. 

## Summary
We have a couple props that are handling one off use cases, or aren't used at all that could be handle more easily. 

## Relevant Technical Choices
Removed `invisible` and `fillHeight` props, they weren't needed. \

## User-facing changes
None. 

## Testing Instructions

1. Just regression on Popups, Dropdowns

no

no

no

## Checklist

---

Fixes #11202


sblinde (approved):
🐝

Makes sense given the context you've shared here.

maxyinger (approved):

swissspidy:
Does Felipe need to do some regression tests here? If so, would need to move this on the board accordingly

PR #11345, opened 2022-04-25
Shopping: Support custom CTA text

### LINKED ISSUE #11309: Shopping: Support custom CTA text
## Feature Description

As of very recently, the default CTA text for the shopping attachment ("Shop now") can be customized, just like for regular page attachments / outlinks.

See https://github.com/ampproject/amphtml/issues/38080

We now want to add support for this in the editor as well in the `PageAttachmentPanel`.

**To-do:**

* [ ] Add `cta-text` attribute to `amp-story-shopping-attachment` allowlist in `KSES`
	* [ ] Add test for it
* [ ] Add CTA text input field to `PageAttachmentPanel`
	* [ ] Maybe refactor this component to split it into two, as it is quite large already
	* [ ] Good time to add some JS unit tests for it as well
* [ ] Update `ProductOutput` component to add `cta-text` attribute 
	* [ ] Add test for it

## Additional Context

Existing CTA text input field for regular page attachments:

Screenshot 2022-04-19 at 16 58 06

spacedmonkey:
## Summary
 Allow for a custom call to action to be defined when a product is on canvas. 

## Testing Instructions

1. Enable shopping experiment. 
2. Active woocommerce and setup a test product. 
3. Create new story. 
4. Go to product tab, insert story. 
5. Click on page. 
6. Go to style tag. 
7. Change CTA text to something custom. 
8. See text change on button.  

## Checklist

---

Fixes #11309


spacedmonkey:
This PR is blocked upstream, as the CTA text is not rendering on the front end / story. But they change is solid and working for now.

spacedmonkey:
This PR is blocked upstream, as the CTA text is not rendering on the front end / story. But they change is solid and working for now.

packages/story-editor/src/components/canvas/displayLayer.js
```diff
@@ -192,7 +192,7 @@ function DisplayLayer() {
 
   const Overlay = useMemo(() => {
     if (hasProducts) {
-      return <ShoppingPageAttachment theme={pageAttachment.theme} />;
+      return <ShoppingPageAttachment pageAttachment={pageAttachment} />;
```
swissspidy: Why not 
```suggestion
      return <ShoppingPageAttachment {...pageAttachment} />;
```

packages/story-editor/src/components/canvas/shoppingPageAttachment/index.js
```diff
@@ -90,7 +90,8 @@ const TextWrapper = styled.span`
 const LIGHT_COLOR = '#FFFFFF';
 const DARK_COLOR = '#000000';
 
-function PageAttachment({ theme }) {
+function PageAttachment({ pageAttachment = {} }) {
+  const { ctaText = __('Shop Now', 'web-stories'), theme } = pageAttachment;
```
swissspidy: Related to my other suggestion to use `<ShoppingPageAttachment {...pageAttachment} />`

Instead of a single `pageAttachment` prop I think we can just keep separate `theme` and `ctaText` props. It's not like we're going to have tons of props here.

```suggestion
function PageAttachment({ ctaText = __('Shop Now', 'web-stories'), theme }) {
```

packages/story-editor/src/components/canvas/shoppingPageAttachment/index.js
```diff
@@ -100,17 +101,18 @@ function PageAttachment({ theme }) {
...
     </Wrapper>
   );
 }
 
 PageAttachment.propTypes = {
-  theme: PropTypes.string,
+  pageAttachment: PropTypes.shape({
+    ctaText: PropTypes.string,
+    theme: PropTypes.string,
+  }),
```
swissspidy: As per my other two suggestions for the props:

```suggestion
  ctaText: PropTypes.string,
  theme: PropTypes.string,
```

packages/story-editor/src/components/panels/design/pageAttachment/pageAttachment.js
```diff
@@ -301,25 +271,36 @@ function PageAttachmentPanel() {
...
-        clear
-        aria-label={__(
-          'Type an address to add a page attachment link',
-          'web-stories'
-        )}
-        hasError={hasError}
-        hint={hint}
-      />
-      {hasValidUrl && (
+      {hasProducts ? (
```
swissspidy, line 274: As per the original ticket description, this whole component/file is becoming rather long.
And now with these ternaries everywhere it's becoming a bit more difficult to see at a glance what UI options are available for each type of page attachment (regular links vs shopping attachment).

This is a good opportunity to consider extracting the components into separate files (and also a good opportunity to add tests while at it). WDYT?

Can also create a new ticket for that of course, but now seems a good opportunity to do it in one go.

packages/output/src/utils/shoppingAttachment.js
```diff
@@ -18,10 +18,15 @@
...
 
-function ShoppingAttachment({ products, theme }) {
+function ShoppingAttachment({
+  products,
+  theme,
+  ctaText = __('Shop Now', 'web-stories'),
+}) {
   return (
-    <amp-story-shopping-attachment theme={theme}>
+    <amp-story-shopping-attachment theme={theme} cta-text={ctaText}>
```
swissspidy: Let's not add a default value for `ctaText` here _and_ let's not add the `cta-text` attribute if the default `__('Shop Now', 'web-stories')` is provided.

Why?

If it's absent, AMP will provide the default "Shop New" text already anyway. Plus, the default text is translated into dozens of locales, which is quite handy.

packages/output/src/test/page.js
```diff
@@ -1505,6 +1505,10 @@ describe('Page output', () => {
         backgroundColor: { color: { r: 255, g: 255, b: 255 } },
         page: {
           id: 'bar',
+          pageAttachment: {
+            theme: 'light',
+            ctaText: 'Buy now',
+          },
```
swissspidy, line 1552: Let's move this to a new test, e.g. `it('should render shopping attachment with custom CTA text`)

Also a good opportunity to add a new `it('should produce valid output with shopping products and custom CTA text')` test for AMP validation.

Aside: the existing `should produce valid output with shopping products` test is currently skipped, but I think now it can be enabled as it should work.

packages/story-editor/src/components/panels/design/pageAttachment/pageAttachment.js
spacedmonkey: As you can see, I did refactor this component to make it much smaller. Lots of repeated code has been removed. How would think to refactor, break it into multiple compontents?

packages/output/src/utils/shoppingAttachment.js
spacedmonkey: Conditional attribute or empty attriubte?

swissspidy: Conditional, i.e. do not add attribute if the default text is provided (`cta-text={undefined}` works)

spacedmonkey: Fixed.

swissspidy (approved):

packages/story-editor/src/components/panels/design/pageAttachment/pageAttachment.js
swissspidy: >  break it into multiple compontents?

Yes, that's what I meant by "extracting the components into separate files"

PR #11378, opened 2022-04-27
Editor: Contributor Flow Submit Feedback

### LINKED ISSUE #10851: Contributor Flow Submit Feedback
## Feature Description

Right now, when a contributor submits their story for review there's no feedback dialog indicating that the story was indeed sent for review. Users who can publish get a confirmation dialog that their story was published. We should have a dialog for contributors when they submit their stories for review that tells them that they have indeed done so. 

https://user-images.githubusercontent.com/10720454/157312320-622ce519-1e68-4f95-a83c-9337279dfbc4.mp4

## Additional Context

This will involve updating some e2e tests.

Look to `PostPublishDialog` in the code base (search and find in `/packages`) for the publish version of this. 
We want to add a similar component that will show when a user who has no publish permissions (check `config` for permissions) submits a story for review.

Please also make a storybook component for this.

sblinde:
## Context

The purpose of this PR is to ensure that Contributors who do not have access to Publish new Stories can get affirmative feedback when their Story has been submitted for review (known as Pending in WP status).

## Summary

This PR does the following:

- Adds `isFreshlyPending` as a hook for future use within the Editor repo package
- Adds a confirmation modal specific to contributor workflow that confirms a Story has been submitted for review.

## Relevant Technical Choices

There was a hook that helped surface whether a story was freshly published, but in this case, we needed something that included the `pending` status that a Story goes into when submitted for review.

## To-do

No additional follow up needed.

## User-facing changes

Before: No visible feedback after submitting a story for review.
After:
Screen Shot 2022-04-27 at 1 52 23 PM

## Testing Instructions

1. Log into WordPress in a Contributor role
2. Create a new or edit an existing Story
3. Make a change and/or save the draft by hitting Submit for Review
4. You should see the modal above the first time you attempt to submit for review.

No.

No.

No.

## Checklist

- [X] I have added documentation where necessary

---

Fixes #10851


packages/e2e-tests/src/specs/editor/storyDetailsModal/contributorUser.js
```diff
@@ -62,7 +62,6 @@ describe('Story Details Modal - Contributor User', () => {
 
     await expect(page).not.toMatchElement('div[aria-label="Story details"]');
     await expect(page).not.toMatch('Story published.');
-    await expect(page).not.toMatchElement('button', { text: 'Dismiss' });
```
BrookeGraham, line 65: Should we check that the new modal shows now 🤔  perhaps in `.../specs/editor/pendingStories`?

packages/wp-story-editor/src/components/postReviewDialog/index.js
```diff
@@ -0,0 +1,62 @@
...
+    ({
+      state: {
+        capabilities: { hasPublishAction },
+        meta: { isFreshlyPending },
+      },
+    }) => ({
+      hasPublishAction,
+      isFreshlyPending,
+    })
+  );
```
BrookeGraham:
```suggestion
  const {
    state: {
      meta: { isFreshlyPending },
    },
  } = useStory();
```

BrookeGraham: No need for `hasPublishAction`. ... Is there 🤔

packages/wp-story-editor/src/components/postReviewDialog/index.js
```diff
@@ -0,0 +1,62 @@
...
+
+  const [isOpen, setIsOpen] = useState(false);
+
+  useEffect(() => setIsOpen(Boolean(isFreshlyPending)), [isFreshlyPending]);
```
BrookeGraham: `isFreshlyPending` is already a boolean.

packages/wp-story-editor/src/components/postReviewDialog/index.js
```diff
@@ -0,0 +1,62 @@
...
+
+  const onClose = useCallback(() => setIsOpen(false), []);
```
BrookeGraham, line 40: We should probably set `isFreshlyPending` to `false` once it has been submitted. Otherwise, if they make an edit and resubmit, the dialog won't appear a second time, even thought the changes have be auto saved.
review submitted

packages/wp-story-editor/src/components/postReviewDialog/stories/postReviewDialog.js
```diff
@@ -0,0 +1,46 @@
...
+ */
+import { StoryContext } from '@googleforcreators/story-editor';
+
+/**
+ * Internal dependencies
+ */
+import PostReviewDialog from '..';
+
+export default {
+  title: 'Stories Editor/Components/Dialog/Post-Review',
```
BrookeGraham: nit: .../Post-Review Submitted
Just to make it a little more clear.

packages/wp-story-editor/src/components/postReviewDialog/stories/postReviewDialog.js
```diff
@@ -0,0 +1,46 @@
...
+  component: PostReviewDialog,
+};
+
+const storyContext = {
+  state: {
+    meta: {
+      isFreshlyPublished: true,
```
BrookeGraham, line 35: This is currently not showing since `hasPublishAction` isn't defined.
Screen Shot 2022-04-27 at 4 03 16 PM

BrookeGraham (commented):
Just a couple questions / suggestions. Looking great though.

packages/e2e-tests/src/specs/editor/storyDetailsModal/contributorUser.js
sblinde: Oh, that's a good idea! i can sort of add this back in the opposite case, that we should expect it. Thanks!

packages/wp-story-editor/src/components/postReviewDialog/index.js
sblinde: You're right, I'll remove it!

packages/wp-story-editor/src/components/postReviewDialog/index.js
sblinde: Can do! Thanks!

packages/wp-story-editor/src/components/postReviewDialog/index.js
maxyinger: please still use a selector tho like you were originally

packages/wp-story-editor/src/components/postReviewDialog/index.js
maxyinger: Ahh I see, `PostPublishDialog` was doing the same practice and you were probably just pulling from there. yea, idk why they wrapped it in `Boolean(...)` in there, but I don't see the need for it

packages/wp-story-editor/src/components/postReviewDialog/index.js
maxyinger: @BrookeGraham I think the ticket asked for this to behave like the `PostPublishDialog` which only shows up when the story moved from not published to published. If we're mirroring that functionality, then I think it should only show up once.

Maybe the `submit for review` text should change to something like `update submission`. idk. regardless, I think this PR satisfies the requested UX in the ticket, and if design wants to alter the UX further they can in later tickets. Thoughts?

maxyinger: worth noting that you can trigger this dialog again if you take the story out of `submitted for review` status and back into a `draft`, then submit it again for review:

https://user-images.githubusercontent.com/35983235/165789320-b80d519a-d4c0-41b2-b911-ad0af0c3917c.mp4

packages/wp-story-editor/src/components/postReviewDialog/stories/postReviewDialog.js
maxyinger: updating that selector where you're pulling this from should fix this.

packages/wp-story-editor/src/components/postReviewDialog/index.js
maxyinger: updating this should fix that storybook error below

maxyinger (dismissed):
Mirrors the functionality & implementation of `PostPublishDialog`. nice work!

packages/wp-story-editor/src/components/postReviewDialog/index.js
BrookeGraham: Oh yeah, my bad. Mostly just meant we can remove `capabilities: { hasPublishAction },`

packages/wp-story-editor/src/components/postReviewDialog/index.js
BrookeGraham: Good point. Since the flow is slightly different from the regular publish flow I feel like having the modal show after updating made more sense. Especially so if the button's text did change from `submit for review` to something like `update submission`. Then it would mimic the regular flow a little more closely.

packages/wp-story-editor/src/components/postReviewDialog/index.js
BrookeGraham: @littlemilkstudio Can you remind me why using the selector pattern is the way we prefer to go?

maxyinger: yea absolutely. Basically `useContextSelector` only re-renders the component when the return value of the selector is shallowly equal between renders. so if you do something like this:
```js
const { state: { someStoryProperty } } = useStory();
```
this will re-render every time the story provider re-renders. 

However, if you do this:
```js
const { someStoryProperty } = useStory(({ state }) => ({ someStoryProperty: state.someStoryProperty }));
```
this will now only update when `someStoryProperty` updates.

Here's a more in depth [synopsis](https://github.com/GoogleForCreators/web-stories-wp/issues/2662#issuecomment-1011372651)

TLDR; it functions the same, but is more performant to use a selector

BrookeGraham: Ah, that is right! I knew it had something to do with re-rendering all the time verses only when `someStoryProperty` updates, similar to selectors in Redux. It had just been awhile since we talked about this, and I had forgotten the details. Thanks @littlemilkstudio you're the best.

sblinde: Super helpful context, thank you for this! And for Brooke asking the question too!

packages/wp-story-editor/src/components/postReviewDialog/index.js
sblinde: Yep! That's why, but you're right: no need to wrap in Boolean then!

spacedmonkey (dismissed):
I noticed something strange. The submit to review should be disabled until you make a change. Compare these two videos.

### Disable update button when nothing has changed. 
https://user-images.githubusercontent.com/237508/165955660-94e65fbd-3ee5-4687-8d60-b46e52415480.mov

### Not disabled button when nothing has changed. 
https://user-images.githubusercontent.com/237508/165955802-617c4a53-63b2-49f9-89a3-a5ef4327b29f.mov

sblinde:
> I noticed something strange. The submit to review should be disabled until you make a change. Compare these two videos.
> 
> ### Disable update button when nothing has changed.
>  Screen.Recording.2022-04-29.at.14.36.47.mov 
> ### Not disabled button when nothing has changed.
>  Screen.Recording.2022-04-29.at.14.38.15.mov

Thanks for the heads up. I'm not sure the code I've changed here would have any impact on these whatsoever. The functionality of the buttons didn't change, only a modal that appears in a certain flow, and the logic for that modal was added, nothing in the base of anything else should've been impacted. We might need to break this into a reported issue, I'd like to see if we can replicate this outside of this PR/branch.

swissspidy:
@spacedmonkey This is actually expected behavior and matches the flow in the regular WP block editor.

For contributors "Submit for review" is essentially "Save this current post/story as pending review", even when they have already done so. There's no "Update" button for contributors, they only have this one "Submit for review" button that is never disabled.

So this works as expected.


https://user-images.githubusercontent.com/841956/165958034-7f9f8ab7-b2e3-4716-85df-abf821fc93cc.mov

littlemilkstudio:
@spacedmonkey just checking back in here.

swissspidy (dismissed):

packages/story-editor/src/app/story/actions/useSaveStory.js
```diff
@@ -71,7 +72,7 @@ function useSaveStory({ storyId, pages, story, updateStory }) {
       const isStoryAlreadyPublished = ['publish', 'future', 'private'].includes(
         story.status
       );
-
+      const isStoryAlreadyPending = ['pending'].includes(story.status);
```
spacedmonkey:
```suggestion
      const isStoryAlreadyPending = 'pending' === story.status;
```

packages/story-editor/src/app/story/actions/useSaveStory.js
```diff
@@ -116,7 +117,9 @@ function useSaveStory({ storyId, pages, story, updateStory }) {
           const isStoryPublished = ['publish', 'future', 'private'].includes(
             data.status
           );
+          const isStoryPending = ['pending'].includes(data.status);
```
spacedmonkey:
```suggestion
          const isStoryPending = 'pending' === data.status;
```

spacedmonkey (dismissed):
Couple of nitpics, but seems to work.

sblinde:
@spacedmonkey Addressed your feedback, would you mind re-reviewing that change quick? This should be ready to merge after that approval. Thanks!

spacedmonkey (approved):

spacedmonkey:
@sblinde Approved. Thanks for actioning feedback.

PR #11384, opened 2022-04-28
Media Recording: add device selection

### LINKED ISSUE #11078: Media Recording: Add video/audio source selection
## Feature Description

We want to give the user the option to select the camera and microphone used for recording.

swissspidy:
## Context

In case you have multiple mics or cameras, we want to allow you choosing the right one

## Summary

This PR adds 2 drop-down menus to select microphone and camera as per the information received by the browser.

**Note: This just contains the functionality and is not pretty since design is subject to change**

## To-do

In the future:

* Update UI and user flow after getting new designs
* Improve error handling (also relies on design)

## User-facing changes

Not pretty, but gets the job done:

Screenshot 2022-04-28 at 12 04 54

## Testing Instructions

1. Start media recording
2. Switch microphone/camera
3. Camera/microphone used for recording should change accordingly

No

No

No

## Checklist

---

Fixes #11078


packages/story-editor/src/components/library/panes/media/local/mediaRecording/modal.js
```diff
@@ -64,6 +65,72 @@ function Modal({ isOpen, onClose }) {
...
+  const [mediaDevices, setMediaDevices] = useState([]);
+
+  const updateMediaDevices = useCallback(async () => {
+    try {
+      const devices = await navigator.mediaDevices.enumerateDevices();
+      setMediaDevices(
+        devices.filter((device) => device.kind !== 'audiooutput')
+      );
+    } catch (err) {
+      // eslint-disable-next-line no-console -- TODO: Figure out error handling.
```
timarney: Is there an issue for this or just going to leave as a todo?

spacedmonkey:
https://user-images.githubusercontent.com/237508/166665478-616d83a7-a7ea-4718-a1d4-b3a916eb14cf.mov

I am not seeing this work at all in firefox. Having issues getting it working in chrome as well.

timarney:
Tested microphone (plug / unplug) --- works great.

https://user-images.githubusercontent.com/62242/166665896-b6fd0cea-286c-4beb-9952-b64b99334966.mp4

spacedmonkey:
I am not see the dropdown at all in chrome. 
Screenshot 2022-05-04 at 11 36 55

spacedmonkey:
I would strongly recommend not allowing the changing of input while the video is recording. Changing of inputs should only be possible before the video. Is there anyway to short the permission check before the video starts record / when the dialog is openned for the first time?

spacedmonkey (dismissed):
I am not seeing video source changing working at all.

packages/story-editor/src/components/library/panes/media/local/mediaRecording/modal.js
swissspidy: Leaving as a todo for now since this depends on UX, which is still in flux.

swissspidy:
> I would strongly recommend not allowing the changing of input while the video is recording. Changing of inputs should only be possible before the video. Is there anyway to short the permission check before the video starts record / when the dialog is openned for the first time?

As per the PR description, this change only contains the bare minimum functionality as the whole UX for video recording, including permission checks, is subject to change.

spacedmonkey:
I noticed, that when I changed input, it triggers a render and makes a recorded video / audio clip restart. I am guessing this should not happen....

swissspidy:
Ah, I might need to clear the blob/stream when changing the input.

@spacedmonkey Does the latest version of this PR still display a webcam stream for you? As mentioned in our sync, it was somehow broken for me. So would be good to know whether it's only broken for me or not.

spacedmonkey:
@swissspidy I am seeing the video stream working. 

However, I am not seeing the video source changing work. I am also seeing some weird behaviours from once I have a record video. Changing of video sources does not work. It can make it hard to test.

swissspidy:
I will look into that.

Did you also run a `npm install` before testing the latest version? Asking because I updated a dependency.

spacedmonkey:
I updated and tested in different browsers. It works for me.

swissspidy:
Good news:

I can now see my video again and was able to fix a bunch of edge cases. Device selection now works much better.

Bad news:

Recording a video is not working for me, not even on `main`. If this happens not just for me, this must have regressed somewhere else and not here, so not really in scope for fixing in this PR.

swissspidy:
There's a new version of the `use-media-recorder` package that I'm going to try out.

packages/story-editor/src/components/library/panes/media/local/mediaRecording/modal.js
```diff
@@ -64,32 +68,188 @@ function Modal({ isOpen, onClose }) {
...
+    hasStream,
+    getMediaStream,
+    isImageCapture,
+    mediaBlob,
+    isRecording,
+    wasRecording,
+  ]);
+
+  // Try to get permissions as soon as the modal opens.
+  useEffect(() => {
```
timarney, line 202: This should help with the flow  👍 even for testing the permissions have been a problem.

timarney:
Tested the microphone + flow as I don't have a webcam setup yet (only built in).

Working as expected --- moving forward it would be nice to break out the device selection into a settings screen so the logic gets separated from the other code.

Will defer approval if someone else can test the webcam selection.

swissspidy:
> Working as expected --- moving forward it would be nice to break out the device selection into a settings screen so the logic gets separated from the other code.

AFAIK current UX mocks lean towards not having a separate settings screen to keep things simple. But component-wise we can definitely refactor things a bit later on.

spacedmonkey:
https://user-images.githubusercontent.com/237508/167872299-0c7a8d92-fc23-4fa7-a58f-890fe5db237b.mov

I am seeing video camera settings working. But I am seeing some weirdness, after the video recorded and reset type. 

I would recommend hiding the video / audio selection options, while the video is recording or is recorded. It will be extremely helpful for testing.

packages/story-editor/src/components/library/panes/media/local/mediaRecording/modal.js
```diff
@@ -201,7 +370,31 @@ function Modal({ isOpen, onClose }) {
         name={audioToggleId}
         checked={enableAudio}
         onChange={() => setEnableAudio((value) => !value)}
+        disabled={!audioInput}
       />
+
+      {videoInputs.length > 0 && (
```
spacedmonkey, line 376:
```suggestion
      {videoInputs.length > 0 && ! isRecording && (
```

packages/story-editor/src/components/library/panes/media/local/mediaRecording/modal.js
```diff
@@ -201,7 +370,31 @@ function Modal({ isOpen, onClose }) {
...
+          ariaLabel={__('Video Input', 'web-stories')}
+          placeholder={__('Select Video Input', 'web-stories')}
+          options={videoInputs}
+          onMenuItemClick={onChangeVideoInput}
+          selectedValue={videoInput}
+          disabled={!enableVideo}
+          popupZIndex={11}
+        />
+      )}
+      {audioInputs.length > 0 && (
```
spacedmonkey, line 387:
```suggestion
      {audioInputs.length > 0 && ! isRecording && (
```

packages/story-editor/src/components/library/panes/media/local/mediaRecording/modal.js
```diff
@@ -201,7 +370,31 @@ function Modal({ isOpen, onClose }) {
...
       />
+
+      {videoInputs.length > 0 && (
+        <DropDown
+          ariaLabel={__('Video Input', 'web-stories')}
+          placeholder={__('Select Video Input', 'web-stories')}
+          options={videoInputs}
+          onMenuItemClick={onChangeVideoInput}
+          selectedValue={videoInput}
+          disabled={!enableVideo}
```
spacedmonkey, line 383:
```suggestion
          disabled={!enableVideo || isRecording}
```
Or we could do this.

swissspidy:
> I would recommend hiding the video / audio selection options, while the video is recording or is recorded.

I understand the suggestion but it‘s out of scope for this ticket/PR because the whole UI will be revamped completely in the next sprint.

spacedmonkey:
PIP should be disabled in firefox. 

https://user-images.githubusercontent.com/237508/167881082-c597262b-6f45-463c-8090-e98e34e00148.mov

swissspidy:
Good catch. Both best fixed in separate tickets though 👍 Soubds unrelated to device selection.

spacedmonkey:
https://user-images.githubusercontent.com/237508/168076351-6be75381-692d-4f1b-a22c-bb6ba164e10a.mp4

Works in Linux :tada:

spacedmonkey (approved):

PR #11386, opened 2022-04-28
POC: Shopping: Have a central place to register shopping "vendors"

### LINKED ISSUE #11382: Shopping: Have a central place to register shopping "vendors"
## Feature Description

Have a central place in the PHP codebase to register / filter vendors, such as shopify. Values from this registry, should be passed to the dashboard. This would easily allow for third party developers to add their own vendor, using one filter.

spacedmonkey:
## Summary

Create a central class to register all types of shopping vendors. This store ( array ) of vendors can then be used in a number of places. 

## Testing Instructions

1.

## Checklist

---

Fixes #11382

### DIFF
--- a/includes/Admin/Dashboard.php
+++ b/includes/Admin/Dashboard.php
@@ -423,6 +423,7 @@ class Dashboard extends Service_Base {
 		}
 		$mime_types               = $this->types->get_allowed_mime_types();
 		$allowed_image_mime_types = $mime_types['image'];
+		$vendors                  = wp_list_pluck( $this->shopping_vendors->get_vendors(), 'label' );
 
 		$settings = [
 			'isRTL'                   => is_rtl(),
@@ -445,7 +446,7 @@ class Dashboard extends Service_Base {
 				'pages'          => '/wp/v2/pages/',
 				'publisherLogos' => '/web-stories/v1/publisher-logos/',
 			],
-			'vendors'                 => $this->shopping_vendors->get_vender_labels(),
+			'vendors'                 => $vendors,
 			'maxUpload'               => $max_upload_size,
 			'maxUploadFormatted'      => size_format( $max_upload_size ),
 			'capabilities'            => [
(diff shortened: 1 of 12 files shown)


includes/Shopping_Vendors.php
```diff
@@ -0,0 +1,86 @@
...
+		 */
+		$query = $this->injector->make( $this->vendors[ $name ]['class'] );
+
+		return $query;
+	}
+
+	/**
+	 * @return array
+	 */
+	public function get_vender_labels(): array {
```
swissspidy: Typo:

```suggestion
	public function get_vendor_labels(): array {
```

includes/Shopping_Vendors.php
```diff
@@ -0,0 +1,86 @@
...
+				'label' => __( 'Shopify', 'web-stories' ),
+				'class' => Shopify_Query::class,
+			],
+			'woocommerce' => [
+				'label' => __( 'WooCommerce', 'web-stories' ),
+				'class' => Woocommerce_Query::class,
+			],
+		];
+
+		$vendors = apply_filters( 'web_stories_shopping_vendors', $vendors );
```
swissspidy: Having this filter in the class constructor means it will potentially only be run once, no? That might lead to unexpected behavior.

It might be better to move all of this to `get_vendors()` and not use a `$vendors` class property at all.

includes/Admin/Dashboard.php
```diff
@@ -434,6 +445,7 @@ public function get_dashboard_settings(): array {
 				'pages'          => '/wp/v2/pages/',
 				'publisherLogos' => '/web-stories/v1/publisher-logos/',
 			],
+			'vendors'                 => $this->shopping_vendors->get_vender_labels(),
```
swissspidy: Perhaps we don't need a separate `get_vendor_labels` class? Could just use `wp_list_pluck` here. Makes things even simpler.

spacedmonkey:
@swissspidy Do you like this? Should I continue with it and merge it into query class.

PR #11387, opened 2022-04-28
Update to Jest 28

### LINKED ISSUE #10843: Update to Jest v28
## Task Description

Jest v28 is currently in alpha. One of the main new features is builtin support for the `--shard` parameter for distributed parallel test execution. See https://github.com/facebook/jest/pull/12546

That means we can remove our custom `jest-parallell-sequencer` package and its usage, and update CI to use this new parameter instead.

swissspidy:
## Context

Jest v28 has built-in support for sharding.

## Summary

Updates Jest to v28 and follows [the upgrade guide](https://jestjs.io/docs/upgrading-to-jest28) to update the test configuration.

## Relevant Technical Choices

* Removes our custom `jest-parallell-sequencer` package
* Adds `exports` resolver workarounds for two packages with known incompatibilities
* Removes `@wordpress/jest-console` from JS unit tests, as this more meant for e2e test context
* Updates one `ChecklistCategoryProvider` test to spy on `console.error` to keep the logs clean

## Testing Instructions

1.

## Checklist

---

Fixes #10843


packages/story-editor/src/components/checklist/countContext/test/checkCountContext.js
```diff
@@ -31,7 +33,17 @@ import {
   useRegisterCheck,
 } from '../checkCountContext';
 
+let consoleErrorFn;
+
 describe('ChecklistCategoryProvider', () => {
+  beforeEach(() => {
+    consoleErrorFn = jest.spyOn(console, 'error').mockImplementation(() => {});
```
timarney, line 40: Curious why this was added.

timarney (dismissed):

packages/story-editor/src/components/checklist/countContext/test/checkCountContext.js
swissspidy: The jsdom environment has a "fake" console that catches thrown errors, which it then passes on to Jest. So Jest then displays the actual console error in the test output. The error is of course expected in this test (using `toThrow()`), but we don't need to see it int the log.

Before that, I think the `@wordpress/jest-console` package was silently swallowing this error for some reason. But since I removed that, now in Jest 28 the message was properly surfaced.

So it's a combination of things that led to this.

packages/e2e-test-utils/src/publishPost.js
```diff
@@ -45,17 +45,24 @@ async function publishPost() {
...
   await page.click('.editor-post-publish-button');
 
   // Wait until the selector returns a truthy value.
-  await page.waitForFunction(() =>
-    wp.data.select('core/editor').getPermalink()
+  await page.waitForFunction(
+    () =>
+      wp.data.select('core/editor').getEditedPostAttribute('status') ===
+      'publish'
   );
```
spacedmonkey, line 61: Can you give some more details here?

spacedmonkey (approved):

packages/e2e-test-utils/src/publishPost.js
swissspidy: Previously this util just checked for `wp.data.select('core/editor').getPermalink()` and that it doesn't return `null`. If it returns a truthy value (e.g. the permalink string), we assumed the post was published.

But apparently nowadays `wp.data.select('core/editor').getPermalink()` always returns a string with the permalink, even if the post is still a draft.

We now wait for the post status to actually change and for the "View Post" link to appear, so that we are more certain that the post was actually published.

PR #11389, opened 2022-04-28
Design Panel: Document update logic

maxyinger:
## Context
Was doing some design panel performance exploration on #11390 and ran into some blockers with the current architecture. Didn't want to change any of the update logic architecture while moving off, so decided to document it. I know prometheus mentioned something similar here #6690 so hopefully this helps with that ticket.

## Summary
Added this chart to our docs. Hopefully helps others evaluate and assess the current system as we make changes here in the future:

```mermaid
flowchart TB
    subgraph story_reducer[story reducer]
        direction LR
        subgraph ACTION
            direction TB
            L[UPDATE_ELEMENTS] --> M[Story Updated]
        end
    end
    subgraph design_panels[StylePanels]
        subgraph design_panel_state[State]
            direction TB
            design_panel_state_el_updates((elementUpdates Array))
            design_panel_state_el_handlers((presubmitHandlers Array))
        end
        subgraph design_panel_callbacks[Callbacks]
            direction LR
            subgraph submit_handler[submit handler]
                subgraph SUBMIT [ ]
                    direction TB
                    E[onSubmit] --> F[[internalSubmit]]
                    F --> G[[apply presubmit element mutations on all staged elementUpdates]]
                    G --> H[[onSetProperties]]
                    H --> I[updateElementsById]
                end
            end
            subgraph submit_event[submit event]
                subgraph DOM EVENT FIRED [ ]
                    direction TB
                    K{Dispatch Event - 'submit'}
                end
            end
            subgraph push_update[pushUpdate called]
                subgraph INPUT ONCHANGE[ ]
                    direction TB
                    A[pushUpdate or pushUpdateForObject] --> |update, submit| B[[setElementUpdates - Add update to update queue]]
                    B -->|submit == false| C[elementUpdates - locally stored updates]
                    B --> |submit == true| D[submit]
                    D --> J[[setTimout to dispatch 'submit' event]]
                end
            end
        end
    end
       push_update .-> submit_event
       submit_event .-> submit_handler
       design_panels .-> story_reducer
```

## To-do
NA

## User-facing changes
NA

## Testing Instructions
NA

- [x] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1.

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [ ] I have tested this code to the best of my abilities
[...]

PR #11390, opened 2022-04-28
Performance: Design Panel Perf Exploration

### LINKED ISSUE #10072: Performance: Design Panel Empty Rerenders
Right now when an element is updated the entirety of the Design Panels re-renders. This task is to discover what's causing this in the current setup and propose a POC that eliminates these empty re-renders.

**Note**
This issue is currently blocked by the pending updates to simplify the design panel. Any current render optimizations may make assumptions about the current design panel architecture which is considered volatile until design simplifications are more complete.

maxyinger:
## Context
This code looked into the design panel performance.

## Summary
This PR contains the second of two separate failed approaches to increase perf in the design panel. Documenting both approaches here to help guide future approaches.

### Culling Unnecessary Re-renders From the Top Down (Original Approach)
This was the first approach I tried, and the code isn't documented in this PR. My goal here was to pass around and interact with the story data a little more wisely to avoid unnecessary re-renders. The main blocker I ran into here was the current architecture inhibiting this.

The main issues stem from 2 architectural choices that make the entire style pane deeply intertwined with selected element state:
1. Holding & staging element updates in local state before dispatching the updates to the top level story reducer. This is used primarily to leverage presubmit handlers, but might be subject to change with #6690 . I modeled out the update logic here in #11389
2. Having an abstract concept of a Design Panel. Since every input panel is abstracted on to be handled as a "design panel", we don't know how the design panel is going to consume the data. Because of this, we end up passing every design panel far more data than is essential for its function (ie `selectedElements` & `selectedElementAnimations`) even tho we rarely need all that info for the all inputs (ie color input only cares about color of current selection).

Altering the current architecture seemed out of scope for this ticket, and without doing that, I was unable to find any solutions that fixed performance issues by just improving how we pass around data.

### Culling Unnecessary Re-renders from the Bottom Up (Second Approach)
This is the approach outlined in the code changes in this PR. The basic idea is to reduce unnecessary re-renders internally in each panel by making each panel more or a wrapper component, inferring the data needed for the input in the wrapper, then making the panel's markup memoized and passing the inferred input data to the memoized input markup.

I tried out this approach on the alignment & sizePosition panels as a POC. The approach was successful in eliminating re-renders, but the eliminated re-renders didn't outweigh cost of the memoization. The memoization also added complexity to the code. Here are my findings:

#### Updating selection between copies of the same element (only position value changes)
**React Timing**
We saw an improvement here, but this doesn't provide the full picture. This isn't accounting for the cost of the memoization which in some cases actually made this slower, while in others it was faster, however basically averaged out to be about the same as before.

| main | this PR |
| ----- | ----- |
| ~11.5ms | ~6.5ms |
| Screen Shot 2022-04-28 at 12 35 43 PM | Screen Shot 2022-04-28 at 12 39 36 PM |

**Main Thread Timing**
Basically averages out to be about the same here and doesn't warrant the added technical complexity of making panels follow a wrapper & memoized controls model:

| main | this PR |
| ----- | ----- |
| ~44ms | ~44ms |
| Screen Shot 2022-04-28 at 12 46 44 PM |  Screen Shot 2022-04-28 at 12 43 58 PM |

### Summary
I think if we really want to see more substantial performance improvements here, it would be in altering the design panels more substantially in #6690 , and there are no quicker wins here.

That being said, this interaction is already under our 100ms budget [outlined here](https://github.com/GoogleForCreators/web-stories-wp/blob/782fd71d2f74752a62ea69fe1173f530b8f85664/docs/architecture.md?plain=1#L22-L23), so it may not be worth investing more time in until we add new features/regressions that push us over our allotted 100ms of response time (main is currently ~44ms in react development build and ~22ms in react prod build on my macbook pro rn).

I think this ticket, #10072 ,  was more of a discovery ticket than anything if you read the description, so I think this is enough to close it out. Will close the PR too, but just wanted to have the code up here to reference in case anybody wanted to see the technical implementation of some of the tactics used.

## Relevant Technical Choices
Described above

## To-do
NA

## User-facing changes
NA

## Testing Instructions
NA

1.

## Checklist

---

Fixes #10072

### DIFF
--- a/packages/react/src/index.js
+++ b/packages/react/src/index.js
@@ -28,6 +28,7 @@ export { default as usePrevious } from './usePrevious';
 export { default as useReduction } from './useReduction';
 export { default as useResizeEffect } from './useResizeEffect';
 export { default as renderToStaticMarkup } from './renderToStaticMarkup';
+export { default as useStableCallback } from './useStableCallback';
 export { default as useUnmount } from './useUnmount';
 export { default as useWhyDidYouUpdate } from './useWhyDidYouUpdate';
 export { default as useInitializedValue } from './useInitializedValue';
--- a/packages/react/src/useStableCallback.js
+++ b/packages/react/src/useStableCallback.js
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * External dependencies
+ */
+import { useCallback, useRef } from 'react';
+
+/**
+ * Takes a callback and returns an up to date version of the supplied
+ * callback, but maintains a stable reference.
+ *
+ * i.e.
+ * ```js
+ * const stableCallback = useStableCallback(callback);
+ * ```
+ *
+ * @param {Function} callback a callback composed in your components scope
+ * @return {Function} callback with a stable reference
+ */
+function useStableCallback(callback) {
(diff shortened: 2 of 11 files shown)


littlemilkstudio:
I think this ticket, #10072 , was more of a discovery ticket than anything if you read the description, so I think this is enough to close it out. Will close the PR too, but just wanted to have the code up here to reference in case anybody wanted to see the technical implementation of some of the tactics used.

Issue #11253, opened 2022-04-12
Shopping: Allow sorting products


swissspidy:
## Feature Description

* [**Figma**](https://www.figma.com/file/ZrO763yvMWJjYeAPsRnNfW/_Editor---Sprint-Feature-Level-Designs_2022?node-id=503%3A15921)

Add support to sort products by date, title, and price in ascending or descending order.

<img width="222" alt="Screenshot 2022-04-12 at 14 43 51" src="https://user-images.githubusercontent.com/841956/162965012-1e942112-7188-4d1e-b75f-b711ccc6f4b9.png">

* Add `orderby` and `order` support to the REST controller and the underlying product query logic (#11154)
* Add sort button/dropdown to the products sidebar tab

PR #11392, opened 2022-04-28
Shopping: Products Sidebar Tab

### LINKED ISSUE #11155: Shopping: Products Sidebar Tab
Establish new products sidebar tab (vs. the currently hijacked elements tab) to display products for adding them to the story.

Empty state:

Screenshot 2022-04-12 at 14 15 23

With search results:

Screenshot 2022-04-12 at 14 15 35

If a product is already on the current page there is a checkmark:

Screenshot 2022-04-12 at 14 31 19

When you hover over it, the checkmark changes to an X:

Screenshot 2022-04-12 at 14 31 33

Click on it to "reset" the product, i.e. it becomes an "empty" dot with no product data

## Acceptance Criteria

Products tab is updated with:

* New icon
* New title & description
* New search form
* New search results list with images, title, and price

Colors and measurements can be taken from Figma; just noting that it's not pixel perfect and we should use existing colors, fonts, etc. from the design system theme where there are differences.

### Out of scope for this ticket

* Dark mode toggle
* Sorting

timarney:
## Context

Establish new products sidebar tab 

## Summary

- Adds new shopping  icon 
- Added new title & description
- Added new search form
- Added new search results list with images, title, and price

## Testing Instructions

CI test issue fixes.

1. Ensure shopping is enabled
2. Visit the shopping tab
3. Find a product via the search box
4. Use the `+` button to add the product
5. Search again ... not the already added product has a ✅ icon
5. Hover over the checked product and remove it via the `X` button

No

No

No

## Checklist

---

Fixes #11155


packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -111,46 +129,72 @@ function ShoppingPane(props) {
...
+        <Row>
+          <HelperText>
+            {__(
+              'This will add products as a tappable dot on your story.',
+              'web-stories'
+            )}
+          </HelperText>
+        </Row>
+        <Row>
+          <Search
```
timarney, line 183: @swissspidy is the `<Search>` component the correct component to be using here?

Here's a screen of where this is at currently (super rough at the moment) -- was going to look into how to remove the "black" background + border for for the container but started thinking maybe this isn't the right component (the dropdown part...) should I just be rendering out list another way?

Screen Shot 2022-04-28 at 5 54 01 PM

timarney:
@swissspidy @spacedmonkey I'll pick this one back up after https://github.com/GoogleForCreators/web-stories-wp/pull/11359 there's still a few todos but this could use some initial comments / review.

packages/design-system/src/icons/index.js
```diff
@@ -132,3 +132,6 @@ export { default as TrashFilledSmall } from './trash_filled_small.svg';
 export { default as TurningLine } from './turning_line.svg';
 export { default as Union } from './union.svg';
 export { default as Video } from './video.svg';
+export { default as ProductCheck } from './product_check.svg';
```
swissspidy: Can't we use `checkmark.svg` on a green background with border radius? Then we can avoid this extra svg.

(Note: colors should be defined in the theme wherever possible and not hardcoded in CSS)

packages/design-system/src/icons/index.js
```diff
@@ -132,3 +132,6 @@ export { default as TrashFilledSmall } from './trash_filled_small.svg';
...
+export { default as ProductPlus } from './product_plus.svg';
```
swissspidy: Can we use `plus_filled.svg` instead?

packages/design-system/src/icons/index.js
```diff
@@ -132,3 +132,6 @@ export { default as TrashFilledSmall } from './trash_filled_small.svg';
...
+export { default as ProductRemove } from './product_remove.svg';
```
swissspidy: `cross.svg` on a red background with border radius? Then we can avoid this extra icon.

(Note: colors should be defined in the theme wherever possible and not hardcoded in CSS)

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -98,59 +188,122 @@ function ShoppingPane(props) {
...
+            menuStylesOverride={containerStyleOverrides}
+            placeholder={'search'}
```
swissspidy: Note that this should be uppercase Search and also translated.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -98,59 +188,122 @@ function ShoppingPane(props) {
...
-    onChange,
-    getOptionsByQuery: getProductsByQuery,
-    selectedId: product?.productId,
-    dropDownLabel: __('Product', 'web-stories'),
-    placeholder: isLoading ? __('Loading…', 'web-stories') : '',
-    disabled: isLoading ? true : isSaving,
-    primaryOptions: isLoading ? [] : initialOptions,
-    zIndex: 10,
-  };
+  const RenderItemOverride = forwardRef(
```
swissspidy: This is clever! Nice idea!

Unfortunately however, this won't give us the flexibility that we need, such as being able to display products even when this dropdown is closed, or being able to open a sorting dropdown next to it (#11253) without closing this dropdown with the products.

I'd recommend trying a more "classic" approach with a simple search form that queries products upon enter and then displays the list of found products below. This is more in line with other panels such as media library or page templates tab.

For keyboard navigation we can use `role="list"` and `role="listitem"` and use ArrowUp/ArrowDown to navigate between items. When on a single product & pressing Tab, the Plus/Delete icon is focused.

This works basically like in the media library or the custom fonts setting.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -98,59 +188,122 @@ function ShoppingPane(props) {
...
+      return (
+        // eslint-disable-next-line styled-components-a11y/click-events-have-key-events -- click events in the contained buttons.
+        <StyledListItem
+          ref={ref}
+          width={option.width}
+          active={isSelected}
+          {...rest}
+          onClick={null}
+        >
+          <StyledProductContainer>
```
swissspidy: For later, this whole product list item component would be a good candidate to extract into its own file and create a storybook story for.

timarney: Broke out the Product list and added a story

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -13,81 +13,192 @@
...
+    actions: { getProducts },
+  } = useAPI();
+
+  const { currentPage } = useStory(({ state: { currentPage } }) => ({
+    currentPage,
+  }));
+
+  useEffect(() => {
+    setOnPageProducts(getOnPageProducts(currentPage));
+  }, [currentPage]);
```
swissspidy: This can be simplified to get rid of this effect & `getOnPageProducts` by simply adding all the logic to the `useStory()` selector.

Something like this (untested & unformatted):

```js
const { currentPageProducts } = useStory(({ state: currentPage } }) => {
  return {
    currentPageProducts: currentPage.elements
      .filter(({type}) => type === 'product')
      .map(({ id, product }) => ({ elementId: id, productId: product?.productId });
  };
}));
```

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -13,81 +13,192 @@
...
+            onChange={onSearch}
+            placeholder={__('Search', 'web-stories')}
+            onKeyDown={handleInputKeyPress}
+            onFocus={handleFocus}
+            isOpen
+            clearId="clear-product-search"
+            handleClearInput={handleClearInput}
+          />
+        </Row>
+        {products && products?.length >= 1 ? (
```
swissspidy: Nit, since you're using optional chaining:

```suggestion
        {products?.length > 0 ? (
```

packages/story-editor/src/components/library/panes/shopping/util.js
```diff
@@ -0,0 +1,42 @@
...
+    }
+  });
+  return products;
+};
+
+export const isOnPage = (product, onPageProducts) => {
+  return onPageProducts.find((item) => item.productId === product?.productId);
+};
+
+export const formatPrice = ({ price, currency }) => {
```
swissspidy: In the `ProductList` component, let's grab the current locale via 

```js
const {
    locale: { locale },
  } = useConfig();
```

and pass it here as an argument, instead of hardcoding to `en-US`

packages/story-editor/src/components/library/panes/shopping/productList.js
```diff
@@ -0,0 +1,224 @@
...
+  }, [isMenuFocused, setCurrentFocusIndex]);
+
+  return (
+    // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions -- list handles arrow up and arrow down
+    <div role="list" onKeyDown={handleListNav}>
+      {products.map((product, index) => {
+        const imageSrc = product?.productImages[0]?.url || '';
+        const onPage = isOnPage(product, onPageProducts);
+        const handleClick =
+          onPage && onPage?.productId
```
swissspidy: Nit, since you are using optional chaining here:

```suggestion
          onPage?.productId
```

packages/story-editor/src/components/library/panes/shopping/productList.js
```diff
@@ -0,0 +1,224 @@
...
+              };
+
+        const handleFocus = (evt) => {
+          if (currentRowsRef.current[`row-${index}`] && evt.type === 'focus') {
+            setCurrentFocusIndex(index);
+          }
+        };
+
+        const ADD_PRODUCT_TEXT = sprintf(
+          /* translators: %s: add product title.*/
```
swissspidy:
```suggestion
          /* translators: %s: product title. */
```

packages/story-editor/src/components/library/panes/shopping/productList.js
```diff
@@ -0,0 +1,224 @@
...
+          __('Add %s', 'web-stories'),
+          product.productTitle
+        );
+
+        const REMOVE_PRODUCT_TEXT = sprintf(
+          /* translators: %s: remove product title.*/
```
swissspidy:
```suggestion
          /* translators: %s: product title. */
```

packages/story-editor/src/components/library/panes/shopping/productList.js
```diff
@@ -0,0 +1,224 @@
...
+                type={BUTTON_TYPES.TERTIARY}
+                size={BUTTON_SIZES.SMALL}
+                variant={BUTTON_VARIANTS.SQUARE}
+                onFocus={handleFocus}
+              >
+                {onPage ? onPageIcons : <Icons.PlusFilled />}
+              </Button>
+            </StyledActionsContainer>
+            {imageSrc ? (
+              <StyledImage alt={product?.productTitle} src={imageSrc} />
```
swissspidy: Need to add the `crossOrigin="anonymous"` attribute here, otherwise images do not load when video optimization (and thus cross-origin isolation) is turned on.

```suggestion
              <StyledImage
                alt={product?.productTitle}
                src={imageSrc}
                crossOrigin="anonymous"
              />
```

packages/story-editor/src/components/library/panes/shopping/productList.js
```diff
@@ -0,0 +1,224 @@
...
+
+        const REMOVE_PRODUCT_TEXT = sprintf(
+          /* translators: %s: remove product title.*/
+          __('Remove %s', 'web-stories'),
+          product.productTitle
+        );
+
+        return (
+          <StyledListItem
+            id={`row-${index}`}
```
swissspidy: Better to use the product ID here instead of the index

packages/story-editor/src/components/library/panes/shopping/productList.js
```diff
@@ -0,0 +1,224 @@
...
+const StyledImage = styled.img`
+  ${imgPlaceholder}
+`;
+
+const StyledDescription = styled.div`
+  padding-left: 12px;
+`;
+
+const StyledTitle = styled(Text).attrs({
+  size: THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.MEDIUM,
```
swissspidy: According to Figma the font size is suppose to be 14px, but `MEDIUM` is 16px.

```suggestion
  size: THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL,
```

packages/story-editor/src/components/library/panes/shopping/productList.js
```diff
@@ -0,0 +1,224 @@
...
+})`
+  color: ${({ theme }) => theme.colors.fg.primary};
+`;
+
+const StyledPrice = styled(Text)`
+  color: ${({ theme }) => theme.colors.fg.tertiary};
```
swissspidy: Should use `THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.X_SMALL` here.

The font size is supposed to be 12px according to Figma.

packages/story-editor/src/components/library/panes/shopping/productList.js
```diff
@@ -0,0 +1,224 @@
...
+  width: 41px;
+  height: 41px;
+`;
+
+const StyledImgPlaceHolder = styled.div`
+  ${imgPlaceholder}
+`;
+
+const StyledImage = styled.img`
+  ${imgPlaceholder}
```
swissspidy: Let's add `object-fit: cover;` here so images are not stretched.

packages/story-editor/src/components/library/panes/shopping/productList.js
```diff
@@ -0,0 +1,224 @@
...
+      setCurrentFocusIndex(0);
+    }
+  }, [isMenuFocused, setCurrentFocusIndex]);
+
+  return (
+    // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions -- list handles arrow up and arrow down
+    <div role="list" onKeyDown={handleListNav}>
+      {products.map((product, index) => {
+        const imageSrc = product?.productImages[0]?.url || '';
+        const onPage = isOnPage(product, onPageProducts);
```
swissspidy: `isOnPage` is a simple 1-liner that is a bit overkill to live in `util.js`.

I would just do `const isOnPage = onPageProducts.find((item) => item.productId === product?.productId);` here

packages/story-editor/src/components/library/panes/shopping/productList.js
```diff
@@ -0,0 +1,224 @@
...
+  useEffect(() => {
+    if (!isMenuFocused) {
+      setCurrentFocusIndex(0);
+    }
+  }, [isMenuFocused, setCurrentFocusIndex]);
+
+  return (
+    // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions -- list handles arrow up and arrow down
+    <div role="list" onKeyDown={handleListNav}>
+      {products.map((product, index) => {
```
swissspidy, line 67: Might be worth extracting the contents of this map callback into its own `Product` component to improve readability.

Then things like `handleClick` could be passed as props.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -13,81 +13,193 @@
...
+      try {
+        setProducts(await getProducts(value));
+      } catch (err) {
+        setProducts([]);
+      }
+    },
+    [getProducts]
+  );
+
+  const onSearch = useCallback(
```
swissspidy, line 101: Let's use `useDebouncedCallback` here, with a timeout of 300 should work.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -13,81 +13,193 @@
...
+      if (element) {
+        deleteElementById({ elementId: element.elementId });
+      }
+    },
+    [deleteElementById, onPageProducts]
+  );
+
+  const handleClearInput = useCallback(() => {
+    setSearchTerm('');
+  }, [setSearchTerm]);
```
swissspidy, line 182: We need to call `getProductsByQuery();` here too.

Otherwise, after I search for "ssgdsgdrgdrgaerdhgrdth" and then clear the input field, I am stuck at the "No products found" state.

Or actually, I don't even see the "No products found" message, so that's a bug:

Screenshot 2022-05-10 at 14 47 26

packages/story-editor/src/components/library/panes/shopping/util.js
swissspidy: We could even consider moving this function to the `i18n` package. Seems fitting.

packages/story-editor/src/components/library/panes/shopping/productList.js
timarney: I broke this up into a few components can move them back into just a `product component` but I think breaking them up seems a little clearer.

packages/story-editor/src/components/library/panes/shopping/productList.js
timarney: Will remove --- some of this code was pulled over from the custom fonts listbox .... in this case we don't really need it.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
timarney: What's the correct syntax for this scenario ?

```javascript 
const debouncedSearchChange = useDebouncedCallback(async (value) => {
    await trackEvent('search', {
      search_type: 'dashboard',
      search_term: value,
    });
    setKeyword(value);
  }, TEXT_INPUT_DEBOUNCE);
```

If I use something like this ☝️ I seem to only get back a SyntheticEvent which I can't seem to pull the value from.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
timarney: Updated.

packages/story-editor/src/components/library/panes/shopping/util.js
timarney: Made a small component inside the product component --- we can move it to i18n if that make more sense.

https://github.com/GoogleForCreators/web-stories-wp/pull/11392/files#diff-7c7bd61db1f996d77975c7bab7bfcaccc2a794a7f7a288e873bd1ec670e3d92aR43

packages/e2e-test-utils/src/shopping.js
```diff
@@ -29,14 +40,14 @@ export const insertProduct = async (product) => {
     expect(page).toClick('[aria-controls="library-pane-shopping"]'),
   ]);
 
-  // This clicks on the Product dropdown and waits for the dropdown menu to open.
-  await expect(page).toClick('[aria-label="Product"]');
-  await page.waitForSelector('[aria-label="Search"]');
-
-  await page.type('[aria-label="Search"]', product);
-  await page.keyboard.press('ArrowDown');
```
timarney, line 37: Tracked down an issue here where the search results are slightly delayed ... in turn the wrong product was getting selected (failing tests). 

Note: That was happening before the `debouncedProductsQuery` was added (just regular search without a built-in delay).

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
swissspidy: Don't have an example handy, but there's plenty of precedence in the code base. Prometheus pod might be able to help if needed.

packages/story-editor/src/components/floatingMenu/elements/karma/shopping.karma.js
```diff
@@ -14,36 +14,44 @@
...
 import { Fixture } from '../../../../karma';
 import { useStory } from '../../../../app/story';
 import { tabToCanvasFocusContainer } from '../../karma/utils';
 
+function isStoryEmpty() {
+  if (document.getElementById('emptystate-message')) {
+    return true;
+  }
+  return false;
+}
```
spacedmonkey: Can you explain what is happening here?

packages/story-editor/src/components/library/panes/shopping/useProductNavigation.js
```diff
@@ -0,0 +1,64 @@
...
+          Math.min(products.length - 1, index + 1)
+        );
+      }
+    },
+    [products]
+  );
+
+  return { handleListNav, currentRowsRef, setCurrentFocusIndex };
+}
+
```
spacedmonkey, line 63: Could be have prop types?

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -13,81 +13,204 @@
...
+            placeholder={__('Search', 'web-stories')}
+            onKeyDown={handleInputKeyPress}
+            onFocus={handleFocus}
+            isOpen
+            ariaClearLabel={__('Clear product search', 'web-stories')}
+            clearId="clear-product-search"
+            handleClearInput={handleClearInput}
+          />
+        </Row>
+        {products?.length > 0 ? (
```
spacedmonkey: I wonder if we should have a loading state here? To ensure that while we waiting for response, we don't see a no product found message.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -13,81 +13,204 @@
...
+function NoProductsFound({ searchTerm }) {
+  if (searchTerm?.length >= 1) {
+    return <HelperText>{__('No products found.', 'web-stories')}</HelperText>;
+  }
+  return null;
+}
+
+NoProductsFound.propTypes = {
+  searchTerm: PropTypes.string,
+};
```
spacedmonkey: Let's move this to another file. One component per file.

packages/story-editor/src/components/library/panes/shopping/productList.js
```diff
@@ -0,0 +1,109 @@
...
+            }
+          : () => {
+              onInsert && onInsert(product);
+            };
+
+        const handleFocus = (evt) => {
+          if (currentRowsRef.current[`row-${index}`] && evt.type === 'focus') {
+            setCurrentFocusIndex(index);
+          }
+        };
```
spacedmonkey: It is not the best thing for performance to define all these functions for each product. The functions seem pretty generic, can we do this outside of the loop?

packages/story-editor/src/components/library/panes/shopping/productImage.js
```diff
@@ -0,0 +1,53 @@
...
+
+const StyledImage = styled.img`
+  ${imgPlaceholder}
+  object-fit: cover;
+`;
+
+function ProductImage({ product }) {
+  const imageSrc = product?.productImages[0]?.url || '';
+  return imageSrc ? (
+    <StyledImage alt={product?.productTitle} src={imageSrc} />
```
spacedmonkey: @swissspidy Do you think we should check to see if these images need to be run through the cors proxy?

packages/story-editor/src/components/library/panes/shopping/productButton.js
```diff
@@ -0,0 +1,107 @@
...
+    /* translators: %s: remove product title.*/
+    __('Remove %s', 'web-stories'),
+    product?.productTitle
+  );
+
+  const onPageIcons = (
+    <>
+      <Icons.Checkmark className="check" /> <Icons.Cross className="remove" />
+    </>
+  );
```
spacedmonkey: We could define this outside the scope of this function. It seems like a const.

packages/story-editor/src/components/library/panes/shopping/product.js
```diff
@@ -0,0 +1,94 @@
...
+const StyledTitle = styled(Text).attrs({
+  size: THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.MEDIUM,
+})`
+  color: ${({ theme }) => theme.colors.fg.primary};
+`;
+
+const StyledPrice = styled(Text)`
+  color: ${({ theme }) => theme.colors.fg.tertiary};
+`;
+function ProductPrice({ price, currency }) {
```
spacedmonkey: This should be in it's own file.

packages/story-editor/src/components/library/panes/shopping/product.js
```diff
@@ -0,0 +1,94 @@
...
+          price={product?.productPrice}
+          currency={product?.productPriceCurrency}
+        />
+      </StyledDescription>
+    </>
+  );
+}
+
+Product.propTypes = {
+  product: PropTypes.object,
```
spacedmonkey: Feels required.

packages/story-editor/src/components/library/panes/shopping/product.js
```diff
@@ -0,0 +1,94 @@
...
+    locale: { locale },
+  } = useConfig();
+
+  if (price) {
+    return (
+      <StyledPrice>
+        {new Intl.NumberFormat(locale, {
+          style: 'currency',
+          currency: currency,
+        }).format(price)}
```
spacedmonkey: This feels like it should a function that lives in a util.

spacedmonkey: Maybe even a unit test for it.

packages/story-editor/src/components/floatingMenu/elements/karma/shopping.karma.js
spacedmonkey: Also add a comment.

spacedmonkey:
Should there be a min / max width on images. 
Screenshot 2022-05-12 at 14 45 16

packages/story-editor/src/components/library/panes/shopping/product.js
```diff
@@ -0,0 +1,94 @@
...
+  currency: PropTypes.string,
+};
+function Product({ product, handleClick, handleFocus, isOnPage }) {
+  return (
+    <>
+      <ProductButton
+        product={product}
+        isOnPage={isOnPage}
+        handleClick={handleClick}
+        handleFocus={handleFocus}
```
spacedmonkey: Why not onClick or onFocus?

packages/story-editor/src/components/library/panes/shopping/productButton.js
```diff
@@ -0,0 +1,107 @@
...
+    margin-right: 10px;
+    &:focus,
+    &:hover {
+      .remove {
+        display: block;
+      }
+      .check {
+        display: none;
+      }
+    }
```
spacedmonkey: I am not sure if we have done a lot of nested css like this. Doesn't really follow the pattern we have set down. Might want to copy other components.

packages/story-editor/src/components/library/panes/shopping/productImage.js
spacedmonkey:
```suggestion
    <StyledImage alt={product?.productTitle} src={imageSrc} loading="lazy" decoding="async" crossOrigin="anonymous" />
```

We need to ensure that all images have these attributes.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -13,81 +13,204 @@
...
-import ProductDropdown from './productDropdown';
+import ProductList from './productList';
 
 const HelperText = styled(Text).attrs({
   size: THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL,
 })`
   color: ${({ theme }) => theme.colors.fg.secondary};
 `;
+function NoProductsFound({ searchTerm }) {
+  if (searchTerm?.length >= 1) {
```
spacedmonkey: This logic does feel like it should live here.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -13,81 +13,204 @@
...
+  const speak = useLiveRegion('assertive');
+  const [searchTerm, setSearchTerm] = useState('');
+  const [isMenuFocused, setIsMenuFocused] = useState(false);
+  const [products, setProducts] = useState([]);
+  const {
+    actions: { getProducts },
+  } = useAPI();
+
+  const { currentPageProducts } = useStory(({ state: { currentPage } }) => {
+    return {
```
spacedmonkey: Do we need a return here?

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -13,81 +13,204 @@
...
+  const debouncedProductsQuery = useDebouncedCallback((value) => {
+    getProductsByQuery(value);
+  }, 300);
+
+  const onSearch = useCallback(
+    (evt) => {
+      const value = evt.target.value;
+      if (value !== searchTerm) {
+        setSearchTerm(value);
+        debouncedProductsQuery(value);
```
spacedmonkey: Wondering if search should be trigger by an useEffect? So this should only set search state and a debounced use effect should do the search. Might make the logic a little cleaner.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -13,81 +13,204 @@
...
+        product,
+      });
+
+      const PRODUCT_ADDED_TEXT = sprintf(
+        /* translators: %s: product title added to page.*/
+        __('%s added to page', 'web-stories'),
+        product.productTitle
+      );
+
+      speak(PRODUCT_ADDED_TEXT);
```
spacedmonkey, line 149: Nice.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -13,81 +13,204 @@
...
+    (product) => {
+      const element = currentPageProducts.find(
+        (item) => item.productId === product.productId
+      );
+      if (element) {
+        deleteElementById({ elementId: element.elementId });
+      }
+    },
+    [deleteElementById, currentPageProducts]
+  );
```
spacedmonkey: What if you have the same product multiple times on a single page? Say you tagged a photo, someone with earrings and tagged each ear.

packages/e2e-test-utils/src/shopping.js
```diff
@@ -18,6 +18,17 @@
...
 
+async function clearSearch() {
+  const hasText = await page.$eval(
+    '[aria-label="Product search"]',
+    (el) => el.value.length
+  );
+  const selector = '[aria-label="Clear product search"]';
+  if (hasText >= 1) {
+    await expect(page).toClick(selector);
+  }
```
spacedmonkey:
```suggestion
  if (hasText >= 1) {
    await expect(page).toClick('[aria-label="Clear product search"]');
  }
```

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -13,81 +13,204 @@
...
+  const handleInputKeyPress = useCallback((event) => {
+    const { key } = event;
+    if (key === 'ArrowDown' || key === 'Tab') {
+      setIsMenuFocused(true);
+    }
+  }, []);
+
+  const handleFocus = () => {
+    setIsMenuFocused(false);
+  };
```
spacedmonkey:
```suggestion
  const handleFocus = () => setIsMenuFocused(false);
```

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -13,81 +13,204 @@
...
+        setSearchTerm(value);
+        debouncedProductsQuery(value);
+      }
+    },
+    [searchTerm, setSearchTerm, debouncedProductsQuery]
+  );
+
+  useEffect(() => {
+    getProductsByQuery();
+  }, [getProductsByQuery]);
```
spacedmonkey:
```suggestion
  useEffect(() => getProductsByQuery(), [getProductsByQuery]);
```

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -13,81 +13,204 @@
...
+      } catch (err) {
+        setProducts([]);
+      }
+    },
+    [getProducts]
+  );
+
+  const debouncedProductsQuery = useDebouncedCallback((value) => {
+    getProductsByQuery(value);
+  }, 300);
```
spacedmonkey:
```suggestion
  const debouncedProductsQuery = useDebouncedCallback(getProductsByQuery, 300);
```

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
timarney: AFAIK you can only add product once on a page.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
timarney: I was able to simply this and just remove that component  

```javascript 
{products?.length > 0 ? (
          //...
        ) : (
          searchTerm?.length && <HelperText>{__('No products found.', 'web-stories')}</HelperText>
        )}
```

packages/story-editor/src/components/floatingMenu/elements/karma/shopping.karma.js
timarney: Added a comment --- 

Essentially this is a check to see if any products are on the canvas --- the `dots` don't have a `product` role or anything selectable to check based on dom elements.

For this test given a fresh story

- is the canvas empty (story empty)
- add a product --- (story not empty)
- remove the product (story empty again)

---- there's other checks for the state etc... but wanted something for the actual stage / canvas

packages/story-editor/src/components/library/panes/shopping/product.js
timarney: Thought about that but there's no `custom` code ...  we would only be test a built in function and seeing what happens when we pass good / bad values to it (which the schema should pick up).

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat

packages/story-editor/src/components/library/panes/shopping/productButton.js
timarney: Tried finding something else in the codebase where we swap icons on hover --- perhaps there's something but only found show / hide type of code.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
timarney: Updated and yes seems a little cleaner.

packages/story-editor/src/components/library/panes/shopping/useProductNavigation.js
timarney: Added

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
timarney: That said ... your right --- I can copy paste a dot 🤔 

With the current code it finds `one` of the dots and deletes it --- the icon still shows that that product exist ... clicking again removes the next dot etc... 

Might be okay as is?

timarney:
> Should there be a min / max width on images. Screenshot 2022-05-12 at 14 45 16

Added a min width + max width

packages/story-editor/src/components/library/panes/shopping/productList.js
timarney: Made some adjustments.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
timarney: Added a spinner for the loading state.

packages/story-editor/src/components/library/panes/shopping/productList.js
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  return (
+    // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions -- list handles arrow up and arrow down
+    <div role="list" onKeyDown={handleListNav}>
+      {products.map((product, index) => {
+        return (
+          <StyledListItem
+            role="listitem"
+            key={product?.productTitle}
```
swissspidy: `productTitle` is not guaranteed to be unique, better to use `productId` here.

packages/story-editor/src/components/library/panes/shopping/productButton.js
```diff
@@ -0,0 +1,108 @@
...
+`;
+
+const onPageIcons = (
+  <>
+    <Icons.Checkmark className="check" /> <Icons.Cross className="remove" />
+  </>
+);
+function ProductButton({ product, onClick, onFocus, isOnPage }) {
+  const ADD_PRODUCT_TEXT = sprintf(
+    /* translators: %s: add product title.*/
```
swissspidy:
```suggestion
    /* translators: %s: product title. */
```

packages/story-editor/src/components/library/panes/shopping/productButton.js
```diff
@@ -0,0 +1,108 @@
...
+    __('Add %s', 'web-stories'),
+    product?.productTitle
+  );
+
+  const REMOVE_PRODUCT_TEXT = sprintf(
+    /* translators: %s: remove product title.*/
```
swissspidy:
```suggestion
    /* translators: %s: product title. */
```

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -46,48 +51,160 @@ const HelperText = styled(Text).attrs({
...
+  const insertProduct = useCallback(
+    (product) => {
+      insertElement('product', {
+        width: 25,
+        height: 25,
+        product,
+      });
+
+      const PRODUCT_ADDED_TEXT = sprintf(
+        /* translators: %s: product title added to page.*/
```
swissspidy:
```suggestion
        /* translators: %s: product title. */
```

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -46,48 +51,160 @@ const HelperText = styled(Text).attrs({
...
+        __('%s added to page', 'web-stories'),
+        product.productTitle
+      );
+
+      speak(PRODUCT_ADDED_TEXT);
+    },
+    [insertElement, speak]
+  );
+
+  const deleteProduct = useCallback(
```
swissspidy, line 154: Can use `speak()` after removing a product too

packages/story-editor/src/components/library/panes/shopping/product.js
```diff
@@ -0,0 +1,67 @@
...
+import ProductImage from './productImage';
+import ProductButton from './productButton';
+import ProductPrice from './productPrice';
+
+const StyledDescription = styled.div`
+  padding-left: 12px;
+`;
+
+const StyledTitle = styled(Text).attrs({
+  size: THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.MEDIUM,
```
swissspidy: According to Figma the font size is supposed to be 14px, but `MEDIUM` is 16px.

```suggestion
  size: THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL,
```

packages/story-editor/src/components/library/panes/shopping/productPrice.js
```diff
@@ -0,0 +1,56 @@
...
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { Text } from '@googleforcreators/design-system';
+
+/**
+ * Internal dependencies
+ */
+import { useConfig } from '../../../../app/config';
+
+const StyledPrice = styled(Text)`
```
swissspidy: Should use `THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.X_SMALL` here.

The font size is supposed to be 12px according to Figma.

packages/story-editor/src/components/library/panes/shopping/productImage.js
swissspidy: Should not be needed here

packages/story-editor/src/components/library/panes/shopping/productList.js
timarney: updated

packages/story-editor/src/components/library/panes/shopping/productList.js
timarney: Now using

```
<StyledImage
      alt={product?.productTitle}
      src={imageSrc}
      loading="lazy"
      decoding="async"
      crossOrigin="anonymous"
    />
```

packages/story-editor/src/components/library/panes/shopping/productList.js
timarney: updated

timarney:
Running into issues with the test ---  waiting for loading etc... with the debounced calls

add-Hoodie with Zipper

add-Album

add-Sunglasses


=====


Pushed some updates here - https://github.com/GoogleForCreators/web-stories-wp/pull/11392/commits/88675409570d87350d65c3bbf71d957a86afb925 but still failing

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -46,48 +51,160 @@ const HelperText = styled(Text).attrs({
...
+    [searchTerm, setSearchTerm]
+  );
+
+  const debouncedProductsQuery = useDebouncedCallback(getProductsByQuery, 300);
+
+  useEffect(() => getProductsByQuery(), [getProductsByQuery]);
+  useEffect(
+    () => debouncedProductsQuery(searchTerm),
+    [searchTerm, debouncedProductsQuery]
+  );
```
spacedmonkey: Do we need both of these?

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -46,48 +51,160 @@ const HelperText = styled(Text).attrs({
...
+          <ProductList
+            isLoading={isLoading}
+            isMenuFocused={isMenuFocused}
+            onClick={onClick}
+            products={products}
+            onPageProducts={currentPageProducts}
+          />
+        ) : searchTerm?.length ? (
+          <HelperText>{__('No products found.', 'web-stories')}</HelperText>
+        ) : null}
```
spacedmonkey:
```suggestion
        {isLoading ? (
              <Loading>
                  <Spinner>
                   <CircularProgress size={24} />
               </Spinner>
              </Loading>
          )}
          {!isLoading && products?.length > 0 ? (
            <ProductList
                isMenuFocused={isMenuFocused}
                onClick={onClick}
                products={products}
                onPageProducts={currentPageProducts}
               />
         )};
         {!isLoading && products?.length === 0 & searchTerm?.length ? (
          <HelperText>{__('No products found.', 'web-stories')}</HelperText>
        )}
```

This is a little cleaner. Current block is a little confusing and nested.

spacedmonkey:
What is happening with this PR?

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
timarney: No -- kept the debounced one

timarney:
> What is happening with this PR?

Battling the e2e tests (at least in CI) --- not having much success :( 

We'll see if the latest commits help (maybe that extra useEffect) was causing issues

timarney:
@swissspidy @spacedmonkey 

Looks like this is now catching on the schema validation for the e2e test

```bash
received
      must have required property 'productUrl'
      must have required property 'productPriceCurrency'
      must NOT have additional properties, but found 'productCurrency'
```

spacedmonkey:
> Looks like this is now catching on the schema validation for the e2e test

Seems feel like valid error. We should fix them in another PR.

timarney:
> Seems feel like valid error. We should fix them in another PR.

Yes agreed

swissspidy:
Fixing this should take like 10 minutes, max. Fix it here or in another PR, whatever works best for y'all.

packages/e2e-tests/src/specs/editor/shopping/woocommerce.js
```diff
@@ -68,13 +66,14 @@ describe('Shopping', () => {
         await previewPage.close();
         const { items } = data;
 
-        expect(items).toHaveLength(3);
+        // expect(items).toHaveLength(3);
```
swissspidy: Let's remove commented out code

packages/story-editor/src/components/floatingMenu/elements/karma/shopping.karma.js
```diff
@@ -14,36 +14,45 @@
...
  */
 import { Fixture } from '../../../../karma';
 import { useStory } from '../../../../app/story';
 import { tabToCanvasFocusContainer } from '../../karma/utils';
 
+function isStoryEmpty() {
+  if (document.getElementById('emptystate-message')) {
+    return true;
+  }
+  return false;
```
swissspidy:
```suggestion
  return Boolean(document.getElementById('emptystate-message')));
```

packages/story-editor/src/components/library/panes/shopping/productImage.js
```diff
@@ -0,0 +1,61 @@
...
+import styled, { css } from 'styled-components';
+import PropTypes from 'prop-types';
+
+const imgPlaceholder = css`
+  display: block;
+  margin-right: 10px;
+  border-radius: 4px;
+  background-color: ${({ theme }) => theme.colors.interactiveBg.disable};
+  min-width: 41px;
+  max-width: 41px;
```
swissspidy: Why add a min/max width when there's a fixed width & height below anyway? It won't really have any effect.

```suggestion
```

packages/story-editor/src/components/library/panes/shopping/productImage.js
```diff
@@ -0,0 +1,61 @@
...
+const StyledImage = styled.img`
+  ${imgPlaceholder}
+  object-fit: cover;
+`;
+
+function ProductImage({ product }) {
+  const imageSrc = product?.productImages[0]?.url || '';
+  return imageSrc ? (
+    <StyledImage
+      alt={product?.productTitle}
```
swissspidy:
```suggestion
      alt={product?.productImages[0]?.alt}
```

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
swissspidy: This is okay as is.

We will handle preventing multiple products separately.

packages/story-editor/src/components/library/panes/shopping/stories/index.js
```diff
@@ -0,0 +1,141 @@
...
+      reviewCount: 0,
+      reviewUrl: 'http://localhost:10004/product/hoodie-with-logo/',
+    },
+    productDetails: 'This is a simple product.',
+    productUrl: 'http://localhost:10004/product/hoodie-with-logo/',
+  },
+];
+
+const configValue = {
+  locale: 'en_CA',
```
swissspidy:
```suggestion
  locale: 'en-CA',
```

Needs a dash, not an underscore

packages/story-editor/src/components/library/panes/shopping/productButton.js
```diff
@@ -0,0 +1,108 @@
...
+import PropTypes from 'prop-types';
+
+const baseIcon = css`
+  width: 16px;
+  height: 16px;
+  border-radius: 100%;
+  color: ${({ theme }) => theme.colors.interactiveBg.disable};
+`;
+
+const StyledActionsContainer = styled.div`
```
swissspidy: Related to https://github.com/GoogleForCreators/web-stories-wp/pull/11392/files#r871447762

Instead of using class names and nested CSS like this, the more styled-components compatible approach would be to use something like this:

```jsx
const Checkmark = styled(Icons.Checkmark)`
  width: 16px !important;
  height: 16px !important;
  border-radius: 100%;
  color: ${({ theme }) => theme.colors.interactiveBg.disable};
  background-color: ${({ theme }) => theme.colors.fg.positive};
`;

const Cross = styled(Icons.Cross)`
  width: 16px !important;
  height: 16px !important;
  border-radius: 100%;
  color: ${({ theme }) => theme.colors.interactiveBg.disable};
  background-color: ${({ theme }) => theme.colors.fg.negative};
`;

const StyledButton = styled(Button)`
  ${Cross} {
    display: none;
  }

  &:hover ${Checkmark}, &:focus-within ${Checkmark}, &:active ${Checkmark} {
    display: none;
  }

  &:hover ${Cross}, &:focus-within ${Cross}, &:active ${Cross} {
    display: block;
  }
`;

const StyledActionsContainer = styled.div`
  margin-right: 10px;
`;
```

And then further down:

```jsx
{isOnPage ? (
  <>
    <Checkmark />
    <Cross />
  </>
) : (
  <Icons.PlusFilled />
)}
```

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -46,48 +63,166 @@ const HelperText = styled(Text).attrs({
...
+        )}
+        {!isLoading && products?.length > 0 && (
+          <ProductList
+            isMenuFocused={isMenuFocused}
+            onClick={onClick}
+            products={products}
+            onPageProducts={currentPageProducts}
+          />
+        )}
+        {!isLoading && (products?.length === 0) & searchTerm?.length ? (
```
swissspidy: Is this supposed to be `&&` instead of `&`? Looks like an accidental use of a bitwise operator.

So I guess this was supposed to be `{!isLoading && products?.length === 0 && searchTerm?.length > 0 ? (...)`

Though I would remove that `searchTerm` check altogether, so that we also display "No products found" if there are no products at all in the store.


```suggestion
        {!isLoading && products?.length === 0 ? (
```

packages/story-editor/src/components/library/panes/shopping/product.js
```diff
@@ -0,0 +1,67 @@
...
+      </StyledDescription>
+    </>
+  );
+}
+
+Product.propTypes = {
+  product: PropTypes.object.isRequired,
+  onClick: PropTypes.func.isRequired,
+  onFocus: PropTypes.func,
+  isOnPage: PropTypes.object,
```
swissspidy: This is supposed to be a boolean. `PropTypes.bool`

packages/story-editor/src/components/library/panes/shopping/productList.js
```diff
@@ -0,0 +1,101 @@
...
+            tabIndex={-1}
+            ref={
+              (el) => (currentRowsRef.current[`row-${index}`] = el) // track the active row
+            }
+          >
+            <Product
+              product={product}
+              isOnPage={onPageProducts.find(
+                (item) => item.productId === product?.productId
+              )}
```
swissspidy: Since this is supposed to be a boolean, use `some()` instead of `find()`

packages/story-editor/src/components/library/panes/shopping/productButton.js
timarney: Nice -- updated

packages/story-editor/src/components/library/panes/shopping/productButton.js
timarney: Updated via @ swissspidy https://github.com/GoogleForCreators/web-stories-wp/pull/11392/files#diff-c888b39912da03829c9fd79046bf5915d4c0b42388924221325182c21b308055R38

packages/story-editor/src/components/library/panes/shopping/product.js
timarney: Update here
https://github.com/GoogleForCreators/web-stories-wp/blob/fix/11155/packages/story-editor/src/components/library/panes/shopping/productPrice.js#L28

packages/story-editor/src/components/library/panes/shopping/productPrice.js
timarney: Updated here:

https://github.com/GoogleForCreators/web-stories-wp/pull/11392/files#diff-78c8fc16b6c6f39ae876dbd1d30b2674df52e81874c83a53af3c559052a57e7dR29

packages/e2e-tests/src/specs/editor/shopping/woocommerce.js
timarney: Removed 

https://github.com/GoogleForCreators/web-stories-wp/blob/fix/11155/packages/e2e-tests/src/specs/editor/shopping/woocommerce.js#L69

packages/story-editor/src/components/library/panes/shopping/product.js
timarney: Added some extra guards for the variables 

https://github.com/GoogleForCreators/web-stories-wp/pull/11392/files#diff-78c8fc16b6c6f39ae876dbd1d30b2674df52e81874c83a53af3c559052a57e7dR39

Can break it out as need be in the future.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -46,48 +63,174 @@ const HelperText = styled(Text).attrs({
...
+          <ProductList
+            isMenuFocused={isMenuFocused}
+            onClick={onClick}
+            products={products}
+            onPageProducts={currentPageProducts}
+          />
+        )}
+        {!isLoading && products?.length === 0 ? (
+          <HelperText>{__('No products found.', 'web-stories')}</HelperText>
+        ) : null}
```
spacedmonkey:
```suggestion
        {!isLoading && products?.length === 0 && (
          <HelperText>{__('No products found.', 'web-stories')}</HelperText>
        )}
```

spacedmonkey:
https://user-images.githubusercontent.com/237508/168915580-ad4fce75-24db-47a5-b92f-6346adda213a.mov

Seeing the loading spinner jump in twice.

timarney:
> Screen.Recording.2022-05-17.at.22.46.52.mov 
> Seeing the loading spinner jump in twice.

Oh interesting --- will take a look wasn't seeing that using `local wp` but shows via Docker

swissspidy (dismissed):

packages/story-editor/src/components/library/panes/shopping/stories/index.js
```diff
@@ -0,0 +1,141 @@
...
+      reviewCount: 0,
+      reviewUrl: 'http://localhost:10004/product/hoodie-with-logo/',
+    },
+    productDetails: 'This is a simple product.',
+    productUrl: 'http://localhost:10004/product/hoodie-with-logo/',
+  },
+];
+
+const configValue = {
+  locale: 'en-CA',
```
spacedmonkey: CA?

packages/story-editor/src/components/library/panes/shopping/productPrice.js
```diff
@@ -0,0 +1,58 @@
...
+    locale: { locale },
+  } = useConfig();
+
+  if (price && currency && locale) {
+    return (
+      <StyledPrice>
+        {new Intl.NumberFormat(locale, {
+          style: 'currency',
+          currency: currency,
+        }).format(price)}
```
spacedmonkey, line 45: This should be moved to a function.

spacedmonkey: Maybe even in it's own hook.

swissspidy: I don't see the value/need to wrap this in its own function. It's just a simple function call. Perhaps you can elaborate on why this would make sense?

spacedmonkey: Unit tests?

swissspidy: Since we don't want to test the behavior of `Intl.NumberFormat` itself (since it's a Web API provided by the browser; we don't need to test how it works), the only thing we could test is that the function itself is called with the provided parameters. Which doesn't really provide much value, since it's a very simple function call.

packages/story-editor/src/components/library/panes/shopping/stories/index.js
timarney: en-US now

swissspidy:
@timarney With 8c834ada72a4ee83e98005e0405c609f9a2d3b25 there now is no "No products found" text anymore when there are no products in the store. That's why I suggested removing this yesterday.

Screenshot 2022-05-18 at 18 36 02

timarney:
> Before the products load in, I am seeing No products found. See should have some state for loaded as well as loading.

@spacedmonkey  -- Originally I was thinking about that text being tied to search functionality (latest commit) but there's the other case of API error or you actually don't have any products ... in which case thinking the search should go away as well (there's nothing to search) and probably the text should be different as some action needs to be taken -- setting up products?  Might be better to handle under an error handling PR?

swissspidy:
>  in which case thinking the search should go away as well (there's nothing to search) and probably the text should be different as some action needs to be taken -- setting up products? 

Ah, that's a good idea. 👍

spacedmonkey:
@timarney i have added a basic PR together that fixes the issue - https://github.com/GoogleForCreators/web-stories-wp/pull/11517

timarney:
@spacedmonkey -- updated with your suggestion.

Can cover the Error type of state in a fresh PR

https://github.com/GoogleForCreators/web-stories-wp/pull/11392#issuecomment-1130245540

swissspidy (approved):

spacedmonkey (approved):

PR #11396, opened 2022-04-29
Media: Fix CORS proxy regression on canvas

### LINKED ISSUE #11395: Hotlinked image not proxied on canvas
## Bug Description

The change in #10206 caused image URLs on the _canvas_ not to be proxied anymore, even though the image was correctly proxied in other areas (e.g. the layers panel)

## Expected Behaviour

Image hotlinking should work fine, including actually using the proxied URL everywhere

## Steps to Reproduce

1. Hotlink an image that needs a CORS proxy (for example https://wp.stories.google/e2e-tests/example.jpg)
2. See that insertion etc. works successfully, but that the image is not displayed correctly on the canvas

## Screenshots

Screenshot 2022-04-29 at 15 28 44

## Additional Context

- Plugin Version: 1.20.0
- WordPress Version: 5.9
- Operating System: macOS
- Browser: Chrome

swissspidy:
## Context

There was a small bug with image hotlinking where the CORS proxy was not used in all places where it should have been

## Summary

Fixes an issue with hotlinked images on the canvas not being run through the CORS proxy.

Adds e2e tests to prevent regressions.

## User-facing changes

Before:

Screenshot 2022-04-29 at 15 28 44

After:

Screenshot 2022-04-29 at 15 47 46

## Testing Instructions

1. Add the image `https://wp.stories.google/e2e-tests/example.jpg` in the editor via the "Insert by link" functionality
2. Verify that the image appears correctly on the canvas and also when previewing the story

No

No

No

## Checklist

---

Fixes #11395


packages/e2e-tests/src/plugins/web-stories-hotlink.php
```diff
@@ -31,16 +30,12 @@
 
 namespace Google\Web_Stories\E2E\Hotlink;
 
-$url       = content_url( '/e2e-assets/test.vtt' );
```
spacedmonkey: Can we try and use the changes found in https://github.com/GoogleForCreators/web-stories-wp/pull/11221

swissspidy: Personally I find that change with a getter function for a filter name that is only used once a bit overkill.

spacedmonkey: It is used twice, which is why it is a function 🤔 . This will be 3 times once this is merged.

packages/e2e-tests/src/specs/editor/media/hotlinking.js
```diff
@@ -0,0 +1,110 @@
...
+    // Ignore resource failing to load. This is only present because of the CORS error.
+    removeResourceErrorMessage = addAllowedErrorMessage(
+      'Failed to load resource'
+    );
+  });
+
+  afterAll(() => {
+    removeCORSErrorMessage();
+    removeResourceErrorMessage();
+  });
```
spacedmonkey, line 53: This is repeated code. Should we move to utils?

packages/e2e-tests/src/specs/editor/media/hotlinking.js
```diff
@@ -0,0 +1,110 @@
...
+import { createNewStory, takeSnapshot } from '@web-stories-wp/e2e-test-utils';
+/**
+ * Internal dependencies
+ */
+import { addAllowedErrorMessage } from '../../../config/bootstrap.js';
+
+const IMAGE_URL_LOCAL = `${process.env.WP_BASE_URL}/wp-content/e2e-assets/example-1.jpg`;
+const IMAGE_URL_CORS_PROXY = 'https://wp.stories.google/e2e-tests/example.jpg';
+
+describe('Media Hotlinking', () => {
```
spacedmonkey, line 33: Should we move the hotlinking found in background audio test to this line. Keep all the hotlink together.

spacedmonkey: Don't we need ` withPlugin('e2e-tests-hotlink');` here?

packages/e2e-tests/src/plugins/web-stories-hotlink.php
swissspidy: It is only used once _per URL_ though.

I don't see any issues with using strings like in this PR.

packages/e2e-tests/src/specs/editor/media/hotlinking.js
swissspidy: Oh, forgot to add it here. Good catch. No wonder the tests are currently failing.

Regarding the captions hotlinking, these make sense to keep with the other bg audio tests.

packages/e2e-tests/src/specs/editor/media/hotlinking.js
swissspidy: Do you mean `packages/e2e-tests/src/utils`? Hmm maybe, can't think of a good way right now.

Seems like a good task for a new ticket.

spacedmonkey: No a had a little function in `packages/e2e-tests/src/config/bootstrap.js` in mind. It should be very simple to put together.

packages/e2e-tests/src/plugins/web-stories-hotlink.php
spacedmonkey: I don't just see the put of repeated logic. So will you change my code when my PR is merged?

swissspidy: Well, the merge conflict will have to be resolved one way or another.

I don't have a strong preference. These are like two lines with simple strings that we never touch again in the future. Adding a function to get these two strings seemed to overkill for me. But I also don't wanna bikeshed this.

packages/e2e-tests/src/specs/editor/media/hotlinking.js
swissspidy: I wouldn't wanna further clutter `bootstrap.js` with this, it's already quite long as it is and the existing logic with the allowed error messages doesn't really fit in there anyway.

spacedmonkey: Okay, we can break that file up if you want....

swissspidy: Created #11406 for this

timarney (approved):

spacedmonkey (approved):

PR #11398, opened 2022-04-29
Output: Add flag to allow blob URLs, expose media app

ayushnirwal:
## Context

This PR adds a new flag 'allowBlobs' to the OutputStory component. This will allow third-party integrations to use this component to create previews with media files stored as blobs ( example [project](https://github.com/rtCamp/web-story-creation-tool/tree/try/published) ) .

Also, media API is exposed so that the project linked above can refresh the media library with the usage of action 'resetWithFetch' when a story with first-party media is imported.

## Testing Instructions

1.

## Checklist

---

Fixes #


swissspidy:
Thanks for opening this.

Sounds reasonable. Kinda like a "strict mode".

We could also think about just removing these blob URL checks completely, since this is also covered by the PHP AMP sanitization.

> Also, media API is exposed so that the project linked above can refresh the media library with the usage of action 'resetWithFetch' when a story with first-party media is imported.

Could you perhaps elaborate a bit on that? Curious to learn more about this use case.

ayushnirwal:
> Could you perhaps elaborate a bit on that? Curious to learn more about this use case.
This project [webstory-creation-tool](https://github.com/rtCamp/web-story-creation-tool/tree/try/published) is a standalone story editor which can be used to generate HTML for stories, stories and 1p media is stored in indexedDB ( hence blob URLs ).
This tool outputs ( exports ) a zip file with all the media items, HTML, some readme.txt, and a JSON file representing the story state. 
An exported story (zip file) can also be imported. If an imported story has 1p media it goes to indexedDB but without using 'resetwithfetch' media library doesn't reflect addition of new media.

ayushnirwal:
> We could also think about just removing these blob URL checks completely, since this is also covered by the PHP AMP sanitization.

So rather than `src = (flags.allowBlobs && isBlobURL(srcArg))? srcArg : "" `
we use `src = (flags.allowBlobs)? srcArg : "" ` ?

swissspidy:
I was thinking just `src = srcArg`.

But that was just a thought. Haven't considered all the implications of that, but it would certainly make things simpler. Something that can be done in its own ticket though I think.

The current PR looks fine as-is.

swissspidy (approved):

PR #11431, opened 2022-05-03
Publish: do not set visibility in story state

swissspidy:
## Context

`visibility` is only a field that we use internally, it's not something that exists in WP.

Storing this in the story provider is not needed. This can be in local state.

## Summary

Use `useState()` instead of the story provider context for keeping track of visibility.

## Testing Instructions

1.

## Checklist

---

Fixes #


spacedmonkey (approved):
Nice

PR #11435, opened 2022-05-04
E2E Tests: Run against WordPress 6.0 RC

### LINKED ISSUE #11373: E2E Tests: Run against WordPress 6.0 RC
## Task Description

We'd want to test against WordPress 5.8 as soon as the first RC is released, so we can track our progress on ensuring compatibility.

This should be in addition to running against `latest`.

Precedent: #8136

spacedmonkey:
## Testing Instructions

1.

## Checklist

---

Fixes #11373


swissspidy:
Getting this console warning in some of the tests:

```
wp.select( 'core/interface' ).isFeatureActive( scope, featureName ) is deprecated since version 6.0. Please use !! wp.select( 'core/preferences' ).isFeatureActive( scope, featureName )
```

I suppose this is coming from here:

https://github.com/GoogleForCreators/web-stories-wp/blob/4b912b8a570ee3ff3263a23226a436fc1181601d/packages/e2e-test-utils/src/visitBlockWidgetScreen.js#L36-L39

But I think the deprecation message is super incorrect. See https://github.com/WordPress/gutenberg/pull/39640#issuecomment-1117625713

`wp.data.select('core/preferences')` does not seem to work. So not clear to me how this can be addressed yet.

----

There's also this warning in the widgets editor, which is coming from WordPress/Gutenberg and not us:

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application.
```

See https://wordpress.slack.com/archives/C02QB2JS7/p1651684729609319, https://wordpress.org/support/topic/errors-on-widget-screen/#post-15614635 for where this was reported before.

Apparently related to this component in GB:

https://github.com/WordPress/gutenberg/blob/3ce2c79b229c719e381b0b5b404b4339c4c93a38/packages/widgets/src/blocks/legacy-widget/edit/index.js#L88-L183

**-> Solution:** We can **temporarily** ignore this console error in `bootstrap.js`.

swissspidy:
Apparently the correct selector is `wp.data.select('core/preferences').get('core/edit-widgets', 'welcomeGuide')`

So we can do this to fix this:

```js
const isWelcomeGuideActive = await page.evaluate(() => {
  return (
    // WordPress < 5.9
    wp.data
      .select('core/edit-widgets')
      ?.__unstableIsFeatureActive?.('welcomeGuide') ||
    // WordPress 5.9
    wp.data
      .select('core/interface')
      ?.isFeatureActive?.('core/edit-widgets', 'welcomeGuide') ||
    // WordPress 6.0
    Boolean(
      wp.data
        .select('core/preferences')
        .get('core/edit-widgets', 'welcomeGuide')
    )
  );
});
```

swissspidy:
Just realized that the order probably be needs the other way around, otherwise there probably still is a deprecation notice because all three functions are tried to be run. So my suggestion probably doesn't work as-is.

Haven't tested, but this might work:

```js
const isWelcomeGuideActive = await page.evaluate(() => {
  if (wp.data.select('core/preferences')) {
    return Boolean(
      wp.data
        .select('core/preferences')
        .get('core/edit-widgets', 'welcomeGuide')
    );
  }

  if (wp.data.select('core/interface')) {
    return wp.data
      .select('core/interface')
      ?.isFeatureActive?.('core/edit-widgets', 'welcomeGuide');
  }

  if (wp.data.select('core/edit-widgets')) {
    return wp.data
      .select('core/edit-widgets')
      ?.__unstableIsFeatureActive?.('welcomeGuide');
  }

  return false;
});
```

swissspidy:
Why 22008398ba30a84e0a35046f139ff3ffd7c4b3a1? IMHO that made it _less readable_ now, mixing an early return with the `||` logic below and not making it clear what each of them does.

spacedmonkey:
> Why https://github.com/GoogleForCreators/web-stories-wp/commit/22008398ba30a84e0a35046f139ff3ffd7c4b3a1? IMHO that made it less readable now, mixing an early return with the || logic below and not making it clear what each of them does.

Reverted

swissspidy (approved):

PR #11443, opened 2022-05-06
Blocks: Fix validation error for prefilled block

### LINKED ISSUE #11281: Prefilled Web Stories block causes block validation error
## Bug Description

I have noticed this issue twice, when I published my webstory, a pop-up will suggest creating a new post, when I click on add the block with the story will be invalid.

Screen Shot 2022-04-17 at 09 51 18

## Expected Behaviour

When the user includes a new post the story block should be valid.

## Steps to Reproduce

1. publish a new webstory
2. click on add a new post 
3. open the post with the story block

## Screenshots

block with error:
Screen Shot 2022-04-17 at 09 51 32

Console error:
Screen Shot 2022-04-17 at 09 51 57

## Additional Context

- Plugin Version: 1.19.0
- WordPress Version: 5.9.3
- Operating System: macOS
- Browser: Chrome 100

spacedmonkey:
## Summary

Ensure that there are no block validation. 

## Relevant Technical Choices

Add setter for source set and sizes to the story model. 
I also handled save depraction for the new mark of the story block. 

## Testing Instructions

1. publish a new webstory
2. click on add a new post
3. open the post with the story block

## Checklist

---

Fixes #11281


swissspidy:
Also a Jest test needs updating.

spacedmonkey:
> Also a Jest test needs updating.

Fixed.

swissspidy:
As per my previous comment, E2E tests need to be updated, after that this should be good to go.

timarney:
> As per my previous comment, E2E tests need to be updated, after that this should be good to go.

Outside of the failing test the code looks good.

spacedmonkey:
@timarney @swissspidy I don't know why the e2e tests are failing. I can see no message, either locally or ci. Any ideas on this one?

timarney:
I'll try locally

timarney:
Commented out the reporters in my local jest config
```
reporters: [['jest-silent-reporter', { useDots: true, showWarnings: true }]],
````

Which gives me back this.
Screen Shot 2022-05-10 at 5 56 51 AM

**Comment by spacedmonkey** [comment]
It doesn't give line number 🤔

**Comment by timarney** [comment]
Ya no line # ---   

It throws here

https://github.com/puppeteer/puppeteer/blob/1e82d980274a149eb37c5f338a74377c6692ae1c/src/common/JSHandle.ts#L522

File: packages/e2e-tests/src/specs/editor/publishingFlow.js [hunk 1]
```diff
@@ -44,18 +60,56 @@ describe.skip('Publishing Flow', () => {
...
+    });
+
+    const filename = await uploadFile('example-4.png');
+    uploadedFiles.push(filename);
+
+    await clickButton('button.media-button-select');
+
+    await page.waitForSelector('[alt="Preview image"]');
+    await expect(page).toMatchElement('[alt="Preview image"]');
+  }
```
swissspidy: Instead of uploading a new image, could we choose an existing image from the media library here to speed things up?

Could import `example-4.png` to the media library in `bin/local-env/install-wordpress.sh` too.

Aside: This would be a useful util for the `e2e-test-utils` package.

swissspidy:
After doing some debugging and research I stumbled upon https://stackoverflow.com/questions/68364356/jest-puppeteer-test-randomly-fails-without-output which was very helpful to find the culprit.

The `onClick` event handler for the "Add to new post" link in the post-publish dialog was not being passed the actual click event, thus the `trackClick` function (which expects an event) was throwing a `Cannot destructure property 'currentTarget' of undefined` `TypeError`.

That caused the test to silently fail.

It's odd that this was not caught/reported by the `@wordpress/jest-console` package — maybe it's broken? It should have caught that.

Luckily the custom Puppeteer environment as suggested by that SO answer helped catch this bug.

Added some extra hardening there now to prevent this from happening again in the future.

Also tried to make the e2e test itself a bit more robust.

Let's see if that helps 🤞

swissspidy:
I don't know why the remaining tests are marked as failed / errored, according to the logs they seem to have passed.

swissspidy (dismissed):

felipebochehin87:
@spacedmonkey @swissspidy I still see the error when I publish the Story and add it to a new post.

https://images.zenhubusercontent.com/235435637/dc6a9da6-3529-4480-b49e-bca572cbd181/validation_error.mp4

swissspidy:
@felipebochehin87 Thanks for testing. Looks like there's an issue when a story has no title.

@spacedmonkey Can you check this scenario? The `save()` function currently returns `null` if a story does not have a title. But the `Image` renderer does render an empty `<a>` link. Maybe the `save()` function should do that too?

If so, that needs a new block migration & an updated JS unit test for the `save()` function.

packages/e2e-tests/src/specs/editor/publishingFlow.js
spacedmonkey: Fixed

spacedmonkey:
@swissspidy This is another bug, unrelated to the one that I fixed. I have push up a fix and a e2e test for it. 

Basically the issue is that the default title, it empty and that makes errors. I think this wasn't an issue before because we didn't allow empty titles. Something changed upstream and broke this. 

I am really happy we are putting in tests here to stop this happening again.

timarney:
@spacedmonkey 

It would be helpful to update the test instructions for this one --- I ran into the title issue as well ...  https://github.com/GoogleForCreators/web-stories-wp/pull/11443#issuecomment-1122843386

Going to approve as that fix will be in another PR.

timarney (approved):

spacedmonkey:
@felipebochehin87 @timarney 

I am not longer seeing the title issue. 

https://user-images.githubusercontent.com/237508/168063341-aecdfc6e-b9af-4a36-9f53-8c3c93c06c70.mov

I believe the issue is now fixed.

felipebochehin87:
@swissspidy @spacedmonkey @timarney same here, the error is gone.

https://images.zenhubusercontent.com/235435637/5efb18f1-bb56-4570-b55d-80f74414b357/2022_05_12_09_27_46.mp4

PR #11448, opened 2022-05-06
Checklist: Warn about first page animations

### LINKED ISSUE #11320: Checklist: No Animation on First Page
## Feature Description

AMP doesn't allow animation on the first page of a Story. Currently, users can add animations to the first page of a Story in the editor. We want to acknowledge the animations so we're not stripping them from the first page (page order can change). It would be helpful for a user to know why the animation isn't playing and that it's not a bug (see https://github.com/GoogleForCreators/web-stories-wp/issues/9987).

This ticket is to add an issue to the checklist under the design panel telling the user that their animation won't play on the first page. It'll look something like the below (work with @swissspidy for proper verbiage and where (if present) the "Learn more" link should go). 

## Additional Context

This will be a single action no CTA checklist card variant. You can see all the checklist card variants in storybook (run `npm run storybook` and then go to story editor components checklistCard. It will look something like this: 
Screen Shot 2022-04-20 at 11 07 10 AM

```
  <ChecklistCard
          title="First Page Animations"
          footer={
            <DefaultFooterText>
              {'Animations are prevents on the first page of Stories to help with performance. '}
              <Link
                href="/demo"
                size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.X_SMALL}
              >
                {'Learn more'}
              </Link>
            </DefaultFooterText>
          }
        />
``` 

This issue will live in the design section of the checklist. 
Screen Shot 2022-04-20 at 11 18 46 AM

So the actual check will live here: `packages/story-editor/src/components/checklist/checks` 
and it will need to be imported and added here: `packages/wp-story-editor/src/components/checklist/design/index.js` 
This is because there are certain checks that are WordPress specific. 

A good example of another issue in the same style is here: `packages/story-editor/src/components/checklist/checks/storyAmpValidationErrors.js` 

Logic for this issue will live in the `checklist/checks/firstPageAnimation.js` (or whatever you end up calling it). You will need to check to see if the first page of a Story has animation. If it does then the check should render. 
You can find the relevant Story Page info through `useStory` - state holds `pages`. We want to know if pages[0] holds animations. There's enough examples of similar in the codebase to use the above to gain familiarity. 

Please include new unit tests for the check (follow the example of the other tests in that directory) and also include a new karma test (`packages/story-editor/src/components/checklist/karma`)

## Acceptance Criteria 
- New checklist design issue that renders when a first page of a Story has animation 
- Affiliated tests

mwritter:
## Context

Add a checklist item for 'First Page Animation' if there are animations on the first page to alert the user about the AMP spec not allowing animations on the first page

## Summary

Adds 'First Page Animation' check to the checklist

## Relevant Technical Choices

The AMP spec doesn't allow animations on the first page of stories for performance reasons, we need to alert the user about this so they know why the animations aren't playing on the first page of their story.

## To-do

UX question: https://github.com/googleforcreators/web-stories-wp/issues/11320#issuecomment-1118887914
## User-facing changes

Before/After

## Testing Instructions

1. Go to the Story Editor
2. Create a page with an animation on it - Go to any page other than the first page add an element (Text, Shape, etc) then apply an animation to it
3. Move the page with an animation on it (created in step two) to the first page, using the page carousel at the bottom or deleting all the other pages
4. Open the 'Checklist' in the bottom navigation to the right of the page carousel

On Staging:
- Notice there is not checklist item for 'First Page Animation'

On Branch:
- Notice there is now a checklist item for 'First Page Animations' alerting the user why the animations on the first page are not allowed/seen

No

No

No

## Checklist

---

Fixes #11320


packages/story-editor/src/components/checklist/checks/firstPageAnimation.js
```diff
@@ -0,0 +1,72 @@
...
+}
+
+const FirstPageAnimation = () => {
+  const isChecklistMounted = useIsChecklistMounted();
+  const animations = useStory(({ state: { pages } }) => pages[0]?.animations);
+
+  const isRendered = useMemo(
+    () => firstPageAnimation(animations),
+    [animations]
+  );
```
swissspidy:
```suggestion
  const hasFirstPageAnimations = useStory(({ state: { pages } }) => firstPageAnimation (pages[0]?.animations));

  const isRendered = hasFirstPageAnimations;
```

Not sure why all the checklist checks use `useMemo` so much, but wouldn't this be easier & cause less re-renders?

cc @littlemilkstudio in case you have some context on that

mwritter: @swissspidy - I agree, what you suggested was my first approach, the functionally definitely doesn't seem to need `useMemo` - but I was trying to follow the structure of the other `checks`.

mwritter: @swissspidy I dont believe `useMemo` is necessary in this `check` and possibly a few others - Ill remove it here to avoid extra re-renders 👍

cc @BrittanyIRL @BrookeGraham

sblinde (dismissed):
🐝

sblinde (approved):
Adding my approval back!

packages/story-editor/src/components/checklist/checks/firstPageAnimation.js
maxyinger: Yea I'm not sure either, but yea, the suggested change looks good to me. should cause less re-renders and the `firstPageAnimation` function isn't expensive to call either

PR #11454, opened 2022-05-09
Media: Skip video optimization based on criteria

### LINKED ISSUE #11405: Media: Only optimize videos not matching criteria
## Feature Description

Since the introduction of FFmpeg-powered video optimization we have this long-standing todo to only optimize videos meeting certain criteria:

https://github.com/GoogleForCreators/web-stories-wp/blob/b8d9c09d50d2ba30b169097a6576b5c656fda25f/packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js#L504-L507

This way we can avoid needlessly optimizing videos which are already fine in file size / format / dimensions.

So far I always thought there is no good library to use for this, but the other day I stumbled upon https://mediainfo.js.org/#mediainfojs which might be a very good fit. It supports lots of formats and provides all the required information about formats, codecs, file sizes, etc. We can also use it to improve detection of muted videos.

Things to look for:

* Dimensions exceeding 720x1280
* Not MP4 with H.264 codec or not WEBM with VP9 codec
* mp4 but not using h624 codec
* File size over recommended 4MB

If a video meets the criteria, we can skip transcoding and mark it as `isOptimized` directly.

swissspidy:
## Context

We want to avoid needlessly running video optimization on videos that are already in very good shape, for example when the creator has manually optimized videos before upload.

The aim is to improve overall user experience and performance during uploads.

## Relevant Technical Choices

Uses [mediainfo.js](https://github.com/buzz/mediainfo.js) to get detailed information about codecs, file size, frame rate, etc.

For the start, small videos (< 5MB), with either MP4 or WEBM type and max dimensions of 720x1280, will be automatically marked as optimized.

## User-facing changes

No visual changes per se, but when uploading smaller videos fitting the criteria they will not be optimized by default anymore.

## Testing Instructions

1. Enable video optimization
2. Upload small videos (< 5MB), with either MP4 or WEBM type and max dimensions of 720x1280
2. Notice that video is not automatically optimized, but uploaded straight away
3. Notice that the checklist does not mark the video as un-optimized
4. Verify that larger videos are still going through optimization
5. Disable video optimization
6. Verify that uploads still work. Smaller videos are marged as optimized, while larger videos and videos of other types (e.g. MOV) are rejected

No

No

No

## Checklist

---

Fixes #11405


packages/story-editor/src/app/media/utils/useMediaInfo.js
```diff
@@ -0,0 +1,181 @@
...
+
+      try {
+        // This will expose the window.MediaInfo global.
+        await loadScriptOnce(mediainfoUrl);
+
+        const mediaInfo = await window.MediaInfo({ format: 'JSON' });
+        const result = JSON.parse(
+          await mediaInfo.analyzeData(getSize, readChunk)
+        );
+
```
spacedmonkey, line 156: Could we use a service worker here?

packages/story-editor/src/components/form/mediaUploadButton.js
```diff
@@ -127,7 +121,10 @@ function MediaUploadButton({ onInsert, renderButton, buttonInsertText }) {
     (resource) => {
       try {
         if (isTranscodingEnabled && canTranscodeResource(resource)) {
-          if (transcodableMimeTypes.includes(resource.mimeType)) {
+          if (
+            !resource.isOptimized &&
+            TRANSCODABLE_MIME_TYPES.includes(resource.mimeType)
+          ) {
```
swissspidy, line 127: Explanation for this change:

`transcodableMimeTypes` did not include mime types that were allowed for upload, i.e. `.mp4` or `.webm`. So such files were never optimized.

With this change, `.mp4` or `.webm` files now also get passed to `optimizeVideo`, which then checks whether the file actually needs to be optimized/transcoded.

packages/story-editor/src/app/media/utils/useMediaInfo.js
swissspidy: I have not looked into using a web worker at all, so this would require some exploration and performance testing first.

At first glance I doubt there's any benefit in using a web worker here though. If this is not an extensive call, then the whole overhead of using a web worker could even make it slower.

packages/story-editor/src/app/media/utils/test/useMediaInfo.js
```diff
@@ -0,0 +1,96 @@
...
+
+const analyzeData = jest.fn();
+const mediaInfo = jest.fn(() =>
+  Promise.resolve({
+    analyzeData,
+    close: jest.fn(),
+  })
+);
+
+describe('useMediaInfo', () => {
```
swissspidy, line 216: Note to self: add tests for `isConsideredOptimized`

packages/story-editor/src/app/media/utils/test/useMediaInfo.js
```diff
@@ -0,0 +1,96 @@
...
+    document.documentElement.removeChild(mediaInfoScript);
+
+    delete window.MediaInfo;
+  });
+
+  afterEach(() => {
+    mediaInfo.mockClear();
+  });
+
+  describe('getFileInfo', () => {
```
swissspidy, line 248: Note to self: add more tests for the result normalization paths, covering more branches

packages/story-editor/src/app/media/utils/useMediaInfo.js
```diff
@@ -0,0 +1,199 @@
...
+    /**
+     * Returns information about a given media file.
+     *
+     * @param {File} file File object.
+     * @return {Promise<MediaInfo|null>} File info or null on error.
+     */
+    async (file) => {
+      const getSize = () => file.size;
+
+      const readChunk = (chunkSize, offset) =>
```
spacedmonkey, line 132: Can we reuse `createFileReader` function?

packages/story-editor/src/app/media/utils/useMediaInfo.js
```diff
@@ -0,0 +1,199 @@
...
+    const hasSmallDimensions =
+      fileInfo.width * fileInfo.height <=
+      MEDIA_VIDEO_DIMENSIONS_THRESHOLD.WIDTH *
+        MEDIA_VIDEO_DIMENSIONS_THRESHOLD.HEIGHT;
+
+    // Video is small enough and has an allowed mime type, upload straight away.
+    const result =
+      hasSmallFileSize &&
+      hasSmallDimensions &&
+      ['video/webm', 'video/mp4'].includes(fileInfo.mimeType);
```
spacedmonkey: Can't we use `allowedVideoMimeTypes` here?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -423,111 +425,142 @@ function useMediaUploadQueue() {
...
-            additionalData = {},
-            muteVideo,
-            trimData,
-          } = item;
-          if (ITEM_STATUS.PENDING !== itemState) {
+  const uploadWithTranscoding = useCallback(
+    async (item) => {
+      const { id, file, resource, muteVideo, trimData } = item;
+
+      const isVideo = file.type.startsWith('video/');
```
spacedmonkey: Could we use getTypeFromMime here?

packages/story-editor/src/components/form/mediaUploadButton.js
```diff
@@ -105,12 +105,6 @@ function MediaUploadButton({ onInsert, renderButton, buttonInsertText }) {
     isTranscodingEnabled,
   ]);
 
-  const transcodableMimeTypes = useMemo(() => {
-    return TRANSCODABLE_MIME_TYPES.filter(
-      (x) => !allowedVideoMimeTypes.includes(x)
-    );
```
spacedmonkey, line 111:
```
return [...TRANSCODABLE_MIME_TYPES, ...allowedVideoMimeTypes ];
```

Could we do this?

packages/story-editor/src/components/form/mediaUploadButton.js
spacedmonkey: Could we do this ? https://github.com/GoogleForCreators/web-stories-wp/pull/11454/files#r909766216

packages/story-editor/src/app/media/utils/useMediaInfo.js
swissspidy: Good question! I completely forgot about that util.

I don't think we can apply it here though, at least not straight away, since `createFileReader` does not pass on the load event (so we can't access `event.target.error`/`event.target.result`) and, most notably, `createFileReader` reads the whole file, but here we only need to read a chunk of it (`file.size()...`).

packages/story-editor/src/app/media/utils/useMediaInfo.js
swissspidy: It's something different.

`allowedVideoMimeTypes` is for generally checking which files can be uploaded to the editor. Typically only `video/webm` or `video/mp4`, but could also include other mime types like `video/ogg`, or whatever you configured on your site.

What we're interested in here is whether the video is to be considered  cross-browser compatible, e.g. can be played on all browsers. Safari does not support `video/ogg` for example. So even if `allowedVideoMimeTypes` contains `video/ogg`, we do not want to consider it "optimized".

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
spacedmonkey: Or
```
resource.type === 'video'
```

swissspidy: I guess theoretically yes, but I don't see any value in changing this to something like `const isVideo = getTypeFromMime(file.type) === 'video'`.

packages/story-editor/src/components/form/mediaUploadButton.js
swissspidy: I don't think so. `allowedVideoMimeTypes` could theoretically contain mime types that FFmpeg does not support (i.e. are not in `TRANSCODABLE_MIME_TYPES `). However, we only want to allow videos that can be transcoded if needed. Otherwise `optimizeVideo()` could error.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: > Or
> 
> ```
> resource.type === 'video'
> ```

`resource.type` could be `gif`, so that doesn't work.

packages/story-editor/src/app/media/utils/useMediaInfo.js
spacedmonkey: Should this function, look like this. 

```
const createFileReader = (file) => {
  const reader = new window.FileReader();
    reader.onload =  (event) => {
            if (event.target.error) {
              reject(event.target.error); 
           }
         resolve(reader);
    }
    reader.onerror = reject;
    reader.readAsArrayBuffer(file);
  });
};
```

swissspidy:
> I am unable to test this locally. I am getting a 404 error for the script.

Huh, that's strange! Looking into it right away.

packages/story-editor/src/app/media/utils/useMediaInfo.js
spacedmonkey: Why if the site does not support mp4? Which happens on a lot multisite installs.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -423,111 +425,142 @@ function useMediaUploadQueue() {
...
+
+      if (isVideo) {
+        const fileInfo = await getFileInfo(file);
```
spacedmonkey: We have data for duration, height and width here? Could we use this data to populate the resource?

packages/story-editor/src/app/media/utils/useMediaInfo.js
swissspidy: If a site does not support mp4, then it will never reach this point since you cannot even upload it.

And if you have existing mp4 files in your story and later forbid mp4 files than you have a whole lot of other problems that are not caused by this.

packages/story-editor/src/app/media/utils/useMediaInfo.js
swissspidy: That does not handle the chunking part though nor does it pass `event.target.result`.

I'm a bit wary of modifying `createFileReader` in this PR since it's also used in `getResourceFromLocalFile`, which is heavily used, so would require more extensive testing.

So I'd rather do this in a separate PR.

spacedmonkey: At the very least we should also have this line 

```
reader.onerror = reject;
``` 

Maybe add a todo, saying to use this function at a later date.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: Good question! I'd suspect we already have correct height and width here, but especially leveraging more accurate duration information makes sense. I'll add it.

spacedmonkey:
Seeing errors in safari. 
Screenshot 2022-06-30 at 12 37 50
Screenshot 2022-06-30 at 12 37 44

swissspidy:
The mediaelement-and-player script is something from WordPress, I don't see how that is related to this PR or why you'd be getting errors from there in the first place since we don't use MediaElement.js in the editor.

The console errors I will look into. It's simply Safari trying to load the source map for the JS file.

swissspidy:
Removed the source map reference, so those errors about the source map should be gone now (might need to clear the cache first)

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: So it's not as easy as I thought, as it would likely require adding a new reducer to update an item's resource's metadata like that. Seems best suited for its own ticket to explore.

spacedmonkey: Why do we need to do a call to `getFileInfo` here?  I did some testing and console.logging and between file and resource objects you have all this data. See the screenshot. 
Screenshot 2022-07-01 at 16 49 26


Could we not just do this. 
```
const fileInfo = {
    fileSize: file.size, 
    height: resource.height,
    width: resource.width,
    mimeType: resource.mimeType,
    duration: resource.length,
};
```

If we do this, we can save an expensive call to `getFileInfo`. If resource / file is not aviable, fallback to getFileInfo. We could even do a first check with resource data, to get height/width/file size/mime type. Then check other things later.

packages/story-editor/src/app/media/utils/useMediaInfo.js
```diff
@@ -0,0 +1,201 @@
...
+
```
spacedmonkey: So we don't check format, frameRate, audioCodec or file size here? Is this something for the future? Maybe add some todos here?

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -189,6 +191,18 @@ function useProcessMedia({
           return;
         }
 
+        const fileInfo = await getFileInfo(file);
```
spacedmonkey: Same question here- https://github.com/GoogleForCreators/web-stories-wp/pull/11454/files#r912080543

We have a lot of this information already in the resource.

swissspidy: Let me reply at https://github.com/GoogleForCreators/web-stories-wp/pull/11454/files#r912080543

packages/story-editor/src/app/media/utils/useMediaInfo.js
swissspidy: Let me reply at https://github.com/GoogleForCreators/web-stories-wp/pull/11454/files#r912080543

swissspidy:
> I am not sure where this check is happening atm.

Let me reply at https://github.com/GoogleForCreators/web-stories-wp/pull/11454/files#r912080543

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: To answer your question(s), for the start in this PR the check is very simple since we have to nail down the right criteria first. And for that we definitely need to take codecs and frame rate into account as well. As you've seen at #11405. And we only get that information via `getFileInfo`, which is the whole purpose of this ticket.

The exact criteria will follow in a separate PR.

spacedmonkey:
Screenshot 2022-07-01 at 17 11 22
Depending on which file, some / all of this data is also in the REST API.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
spacedmonkey: Can we do two checks? One for simple data, like height, width, duration, mime type and file size. If that check passes then do the call to file info?

swissspidy:
> Screenshot 2022-07-01 at 17 11 22
> 
> Depending on which file, some / all of this data is also in the REST API.

We need this data _before_ even uploading a file to the server

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: You mean like if it‘s a MOV file don‘t bother getting file info since it needs transcoding anyway?

spacedmonkey: > You mean like if it‘s a MOV file don‘t bother getting file info since it needs transcoding anyway?

Yes! Why load this library if we don't need to. Do a basic check for mime type, dimensions and length. If these fail check ( as in not mp4 etc ), then skill loading this library and just transcode. This is "preflight" check passes, then do the check with fileInfo and loading that library. Most transcoding happens, because the file type / size is wrong. We may never need to load this library.

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -189,6 +191,18 @@ function useProcessMedia({
           return;
         }
```
spacedmonkey, line 193: @swissspidy Here we have an uploaded attachment and the server would have this data. See https://github.com/GoogleForCreators/web-stories-wp/pull/11454#issuecomment-1172501960

swissspidy: For this case of optimizing an existing video, yes, the server would indeed likely have such information already.

We would need to make an HTTP request to fetch this information and then analyze it, falling back to `getFileInfo()`

A few potential issues I see here:

* Increased complexity due to having to maintain two ways for getting file info
* Unnecessary extra HTTP request in case the server does _not_ have the file info
* This server-side info is most likely only available in a WordPress context, not a decoupled context

spacedmonkey: I think we are okay to just use getFileInfo here. It is easier and the data is more consist and testable. Maybe sometime in the future. 

Another idea, would be do this at the php level, make a video as optimized, if it is the correct format, size, length etc.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: Yeah that makes sense.

The only downside is that we can't do this anymore:

https://github.com/GoogleForCreators/web-stories-wp/blob/cbd8e59cf3f85d59b48ac761d1bcef9ee5c46932/packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js#L438-L440

spacedmonkey: > The only downside is that we can't do this anymore:

Why not? Resource has isMuted property.

swissspidy: So in `getResourceFromLocalFile` we use `hasVideoGotAudio` to determine whether a video has audio or not.

This might not work for all types of videos, if the browser cannot properly play it. Think MOV, MKV, etc. It might mark a video as muted even though it _does_ have audio.

By using `getFileInfo` here we have more accurate audio information.

spacedmonkey: Could we not do this?

```
  if (isVideo) {
        if (isConsideredOptimizedResource(resource)) {
            const fileInfo = await getFileInfo(file);

           if (fileInfo) {
             // This more accurate audio information is also useful
            // if transcoding is not skipped, so we set this here.
            item.additionalData.isMuted = fileInfo.isMuted;

             if (isConsideredOptimized(fileInfo)) {
               // Do not override pre-existing mediaSource if provided,
               // for example by media recording.
               if (!item.additionalData.mediaSource) {
                 item.additionalData.mediaSource = 'video-optimization';
                }

                uploadItem(item);

                return;
               }
         } 
        }
      }
```

packages/story-editor/src/app/media/utils/useProcessMedia.js
spacedmonkey: > Another idea, would be do this at the php level, make a video as optimized, if it is the correct format, size, length etc.

Can should be done in a follow up ticket.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: Refactored this a bit in 8b65a5c235762b42e00e7ec61df0c89b0793aef2

At the cost of the more accurate audio information, but with short-circuit

packages/story-editor/src/app/media/utils/useMediaInfo.js
```diff
@@ -0,0 +1,249 @@
...
+     * @return {Promise<boolean>} Whether the file meets optimization criteria.
+     */
+    async (resource, file) => {
+      // This should never happen, but just in case.
+      if (resource.isOptimized) {
+        return true;
+      }
+
+      // Short-circuit for non-matching mime types.
+      if (resource.mimeType !== 'video/mp4') {
```
spacedmonkey:
```suggestion
      if (resource.mimeType !== MEDIA_TRANSCODED_MIME_TYPE) {
```

packages/story-editor/src/app/media/utils/useMediaInfo.js
```diff
@@ -0,0 +1,249 @@
...
+        fileInfo.fileSize < MEDIA_VIDEO_FILE_SIZE_THRESHOLD;
+
+      const hasSmallDimensions =
+        fileInfo.width * fileInfo.height <=
+        MEDIA_VIDEO_DIMENSIONS_THRESHOLD.WIDTH *
+          MEDIA_VIDEO_DIMENSIONS_THRESHOLD.HEIGHT;
+
+      // AVC is H.264.
+      const isSupportedMp4 =
+        fileInfo.mimeType === 'video/mp4' && fileInfo.videoCodec === 'avc';
```
spacedmonkey, line 251:
```suggestion
        fileInfo.mimeType === MEDIA_TRANSCODED_MIME_TYPE && fileInfo.videoCodec === 'avc';
```

packages/story-editor/src/app/media/utils/useMediaInfo.js
swissspidy: Hmm I think I am rather gonna add a new constant for this, otherwise it's a bit misleading.

Though really just a detail.

spacedmonkey:
Okay, so I uploaded a mp4. Waited for the transcode to complete. Then downloaded the transcoded version of the mp4. When I re-uploaded the file, it transcoded it again. Why would that happen? Should that not skip?

spacedmonkey:
More updates. A 45 second transcode video is 4.4MB, hitting the 4MB threshold. I would up this, to say 10MB

swissspidy:
> Okay, so I uploaded a mp4. Waited for the transcode to complete. Then downloaded the transcoded version of the mp4. When I re-uploaded the file, it transcoded it again. Why would that happen? Should that not skip?

I don't know. I will have to test this myself. Feel free to debug in `useMediaInfo` too.

> More updates. A 45 second transcode video is 4.4MB, hitting the 4MB threshold. I would up this, to say 10MB

Short answer: Our guideline is to keep video size < 4 MB. That's the threshold.

Long answer: We could theoretically think of the threshold to be more like "n MB x m seconds". With recommended video duration to be < 15s, that would be 4 / 15 = 0.2667 MB per second, so 12 MB for a 45 second video. However, then you'll upload a 12 MB file thinking it's fine, which it is not.

spacedmonkey:
> Long answer: We could theoretically think of the threshold to be more like "n MB x m seconds". With recommended video duration to be < 15s, that would be 4 / 15 = 0.2667 MB per second, so 12 MB for a 45 second video. However, then you'll upload a 12 MB file thinking it's fine, which it is not.

So check needs to take into account the length of the video. If we say 0.2667 * mediaInfo.duration < mediaInfo.filesize. Seems simple enough to compare against this.

packages/story-editor/src/app/media/utils/useMediaInfo.js
spacedmonkey: What is confusing about this, that is the expected mime type of a transcoded video. That is how I read that.

swissspidy:
Well yes, _theoretically_. However, then we would mark a 45 seconds 12 MB video as optimized, while it is clearly exceeding the recommended 4MB maximum, which is not good.

packages/story-editor/src/app/media/utils/useMediaInfo.js
```diff
@@ -0,0 +1,249 @@
...
+      }
+
+      const fileInfo = await getFileInfo(file);
+
+      if (!fileInfo) {
+        return false;
+      }
+
+      const hasSmallFileSize =
+        fileInfo.fileSize < MEDIA_VIDEO_FILE_SIZE_THRESHOLD;
```
spacedmonkey:
```suggestion
      const threshold = fileInfo.duration * 266666.667;
      const hasSmallFileSize = fileInfo.fileSize < threshold;
```

A bit of testing, this seems to work. For a 45 second video, it 11.1MB, which seems right.

swissspidy: See https://github.com/GoogleForCreators/web-stories-wp/pull/11454#issuecomment-1174038967

There are some disadvantages with this method.

packages/story-editor/src/app/media/utils/useMediaInfo.js
swissspidy: It is misleading once we add support for WEBM/VP9.

That's why I would create a new const for this use case.

spacedmonkey: >It is misleading once we add support for WEBM/VP9. 

Is that the plan?

swissspidy: Yeah once browser support is better. See the `todo` 10 lines above your comment.

spacedmonkey:
> Well yes, _theoretically_. However, then we would mark a 45 seconds 12 MB video as optimized, while it is clearly exceeding the recommended 4MB maximum, which is not good.

But what it the point of transcoding it at this point? The transcoded video was 4.4MB, retranscoding it, results in 4.4MB video file, as nothing changes. Making the transcode pointless. If 4MB is a hard limit, that once that limit is meet, either the video should be rejected somehow or quality / size should be reduced.

swissspidy:
There are several cases like that that need some more thinking/consideration, that's all.

packages/story-editor/src/app/media/utils/useMediaInfo.js
```diff
@@ -0,0 +1,249 @@
...
+    [mediainfoUrl]
+  );
+
+  const isConsideredOptimized = useCallback(
+    /**
+     * Determines whether a video file is to be considered optimized.
+     *
+     * Checks things like dimensions, file size, mime type, and codecs.
+     *
+     * @todo Allow WebM with VP9 once Safari catches up.
```
spacedmonkey, line 213: Are you sure that it doesn't already support it - https://caniuse.com/webm

swissspidy: Yes.

Check the footnotes. Safari on iOS currently only supports it for WebRTC. That's why the cells there aren't green.

spacedmonkey:
I am seeing some weirdness when uploading multiple files. I am seeing the same file uploading more than once in the media library. In some cases 10 times. 

https://user-images.githubusercontent.com/237508/177535346-68268c95-382c-4e83-8dc3-8e257ce8988b.mov

packages/story-editor/src/app/media/utils/useMediaInfo.js
```diff
@@ -0,0 +1,260 @@
...
+      // and < 4 MB in size. If uploading a longer video,
+      // it is only natural that it will exceed the size limit.
+      // Thus, we're instead checking for the average size per second.
+      // Example: 4MB for a 15s video, 12MB for a 45s long video.
+      // TODO: Revisit to avoid fallacy that we're OK with such large file sizes.
+      const hasSmallFileSize =
+        fileInfo.fileSize <
+        (MEDIA_VIDEO_FILE_SIZE_THRESHOLD /
+          MEDIA_RECOMMENDED_MAX_VIDEO_DURATION) *
+          fileInfo.duration;
```
spacedmonkey: Can't this be move to the short circuit above? We have duration from the resource and file size from the file.

packages/story-editor/src/app/media/utils/useMediaInfo.js
```diff
@@ -0,0 +1,260 @@
...
+
+      const hasSmallDimensions =
+        fileInfo.width * fileInfo.height <=
+        MEDIA_VIDEO_DIMENSIONS_THRESHOLD.WIDTH *
+          MEDIA_VIDEO_DIMENSIONS_THRESHOLD.HEIGHT;
```
spacedmonkey: Isn't this a repeat of the check above?

swissspidy: The check above only works for resources that are not placeholders, i.e. it won't work for MKV files for example.

This check here works for MKV files too.

packages/story-editor/src/app/media/utils/useMediaInfo.js
swissspidy: We don't have reliable duration information all the time, for example for MKV files or similar.

I might be able to add a _second_ check to the short-circuit above _if duration is available_.

I'll take a look.

spacedmonkey:
I am not able to drag an mkv into canvas. 



https://user-images.githubusercontent.com/237508/177539433-abe8e79d-489d-4eb4-911e-b95c4a065248.mov

swissspidy:
Does it work on main?

swissspidy:
@spacedmonkey still working on it as per our last discussion. I will re-request a review once it‘s ready.

packages/story-editor/src/components/library/panes/media/common/innerElement.js
```diff
@@ -45,6 +45,7 @@ const styledTiles = css`
   transition: 0.2s transform, 0.15s opacity;
   border-radius: 4px;
   opacity: 0;
+  object-fit: cover;
```
swissspidy, line 48: Needed because otherwise the elements can appear stretched:

Screenshot 2022-07-12 at 11 36 50

packages/story-editor/src/app/media/utils/useMediaUploadQueue/constants.js
```diff
@@ -15,6 +15,8 @@
  */
 export const ITEM_STATUS = {
   PENDING: 'PENDING',
+  PREPARING: 'PREPARING',
+  PENDING_TRANSCODING: 'PENDING_TRANSCODING',
```
swissspidy, line 19: Introduced these two new states to make the whole hook more maintainable and prevent items from being processed twice at a specific step (i.e. do not run mediainfo twice on the same video)

Best reviewed by looking at the effects in the hook and how items flow from pending -> preparing -> pending_transcoding -> transcoding -> uploading

spacedmonkey: Update this comment. https://github.com/GoogleForCreators/web-stories-wp/blob/690006d36e37b0972af1a4b8875d8227db97d688/packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js#L64

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -425,108 +436,121 @@ function useMediaUploadQueue() {
...
+    prepareForTranscoding,
+    isConsideredOptimized,
+    uploadItem,
     isTranscodingEnabled,
     canTranscodeFile,
+  ]);
+
+  // Transcode items sequentially.
+  useEffect(() => {
+    state.queue.forEach(
```
spacedmonkey, line 507: Why for each and not promise all?

packages/story-editor/src/components/library/panes/media/common/innerElement.js
spacedmonkey: Is this related to this changed?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/constants.js
spacedmonkey: What about adding these new states to here

https://github.com/GoogleForCreators/web-stories-wp/blob/dba641ba6c13e3313ad3313fd0c165ca99ea6eaa/packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js#L733-L744

https://github.com/GoogleForCreators/web-stories-wp/blob/dba641ba6c13e3313ad3313fd0c165ca99ea6eaa/packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js#L644-L651

https://github.com/GoogleForCreators/web-stories-wp/blob/dba641ba6c13e3313ad3313fd0c165ca99ea6eaa/packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js#L588-L596

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -425,108 +436,121 @@ function useMediaUploadQueue() {
...
-            state: itemState,
-            resource,
-            additionalData = {},
-            muteVideo,
-            trimData,
-          } = item;
-          if (ITEM_STATUS.PENDING !== itemState) {
-            return;
-          }
+    state.queue.forEach(
```
spacedmonkey, line 439: Should we change all of these to forEach?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -425,108 +436,121 @@ function useMediaUploadQueue() {
...
-          if (
-            resource.type === 'video' &&
-            resource.isMuted !== null &&
-            additionalData?.isMuted === undefined
-          ) {
-            additionalData.isMuted = resource.isMuted;
-          }
+        // Changing item state so that an item is never processed twice
+        // in this effect.
+        prepareItem({ id });
```
spacedmonkey, line 455: So pending and prepare are basically the same?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: Because we don't need to await promises and we don't care about return values.

In such cases `.forEach` makes more sense than `.map`

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: I only changed it where it makes sense.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: No.

A pending item is one that has not been acted on yet. This kicks off the uploading process for the item, i.e. gets the ball rolling.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -425,108 +436,121 @@ function useMediaUploadQueue() {
...
-            isAnimatedGif(await file.arrayBuffer());
+        if (isVideo) {
+          // TODO: Consider always using getFileInfo() to have more accurate audio information.
 
-          const needsTranscoding =
-            isTranscodingEnabled &&
-            (isGifThatNeedsTranscoding || canTranscodeFile(file));
+          if (
+            item.additionalData.mediaSource !== 'recording' &&
+            (await isConsideredOptimized(resource, file))
```
spacedmonkey, line 464: For resources objects, like mpg / avi, it is a shame we can not update missing fields like mute, length etc.

swissspidy: That's what the TODO there is for a few lines above.

I had this before but reverted it because of the preference for short-circuit.

spacedmonkey: How about something like this 
```
const newResource = populateInfor(resource, file);
if( !resource.isOptimized ) {
```
Then return `isConsideredOptimized` a resource object and populate missing fields. 

Should be simple enough to change up the code as well.

packages/story-editor/src/components/library/panes/media/common/innerElement.js
swissspidy: I just noticed it while testing this one

spacedmonkey: This should be handled in another PR, as it is unrelated.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: That is quite the course reversal again. IIUC last time we discussed that it can be wasteful to run MediaInfo on every resource, even if it provides useful information and that we would start with this simple approach of just checking optimization status. That's why I added the TODO following that discussion so that we could come back to this later.

packages/story-editor/src/components/library/panes/media/common/innerElement.js
swissspidy: I'll leave it in as it's just a one line change, not worth creating another PR for it.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
spacedmonkey: No it isn't just changing params really. As it may not be clear to you what I meant, I put together a branch. See https://github.com/GoogleForCreators/web-stories-wp/commit/843009560a5e459e0eef075a9f439437b1824b97

Basically, change to return `isConsideredOptimized`. That needs to done is update some tests and then update the media queue to use the updated resource. Could be a very simple change and makes the code more effective.

spacedmonkey: After looking at this again, I think it might not really be worth it. The only benefit here, is adding height / width / length / isMuted to files that need to be transcoded. But these are the most useless and only displayed in the media library. It might worth adding to the code base but done in the follow up PR.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
spacedmonkey: Okay, let's add some comment to explain them. It isn't clear to me why this change was made and future versions of us will thank us for the comment 😉

spacedmonkey:
I am seeing the poster generation hanging. 



https://user-images.githubusercontent.com/237508/178968311-4d4d8aec-5537-498b-9cc6-1e8d2e2c0078.mov



I believe this is happening because poster generation is trying to happen on blob url. Not sure why this is happening.

spacedmonkey:
> I believe this is happening because poster generation is trying to happen on blob url. Not sure why this is happening.

This seems to be happening on main as well.

spacedmonkey (approved):
This is a good first start. I think we have more to do here in future PRs.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
swissspidy: Ah 💡 Now I understand it better! Thanks for explaining it to me, it's much appreciated.

Agreed that it would be nice to leverage the fileInfo when it's already there, but that it would also be a rather low impact change.

Let's revisit this again in the future as you suggested 👍

felipebochehin87:
@spacedmonkey @swissspidy I still see the "Video optimization" message when I upload a  video that is a mp4 file, 3.97MB, 720x1280. The video that I am uploading is the second one. The first is to show that the message is still being displayed.

https://images.zenhubusercontent.com/235435637/d65fd64d-f708-4987-a113-d099b06948d2/video_opt_message.mp4



https://images.zenhubusercontent.com/235435637/b54ad3c3-e372-4f5a-822b-9fd6a6a3898a/3_97_mb_720_1280.mp4

swissspidy:
@felipebochehin87 Good catch! There was some incorrect comparison being used ("<" instead of "<="). However, your video also has a frame rate of only 20 FPS, whereas the recommendation is 24+ FPS, which is why your video is still being optimized. So that is still expected. Apologies for not updating the instructions accordingly.

felipebochehin87:
Oh thanks @swissspidy! I'll retest right now :)

felipebochehin87:
@swissspidy I uploaded the video below and also got the message:

This video is 1280x720, 120 FPS, 2.12MB. Am I missing anything else?

https://images.zenhubusercontent.com/235435637/43def0d2-6cb2-4650-a18e-f1cd5b913cb3/2022_07_18_09_11_15.mp4

swissspidy:
@felipebochehin87 Please try again now :-) My fix wasn't uploaded to the QA environment yet when I wrote my comment. But now it's updated there and the video is uploading without optimization.

PR #11429, opened 2022-05-03
Added remaining layer lock handling

barklund:
## Context

This adds a bunch of extra features related to locked layers. Hopefully everything to make the feature fully functional.

See todo-list for details.

## Technical choices
Some simplifications and assumptions for technical reasons:
* Multi-selection is not possible if any element is locked. This makes multi-selection transformations easier to handle (they're always allowed because no element in selection can be locked)
* You cannot right-click a locked element directly, but once selected, you can. You can however no longer "click through" a locked element once selected.

Both of the above scenarios deviate from Figma's handling of locked elements, but both of the above solutions were much easier to implement than to copy how Figma does them.

## To-do

* [x] When dragging media, do not highlight locked media and shape as potential drop target nor allow media to be dropped "into" locked elements
* [x] When hovering a locked element, do not display a slight blue border
* [x] Do not allow locked elements to be clicked, unless already selected (can be selected through right-click menu or layer panel)
* [x] Filter out locked elements from lasso selection
* [x] Filter out locked elements from shift-selection in layer panel
* [x] When meta-selecting a non-locked layer via layer panel, if old selection was a locked layer, only select new element
* [x] When meta-selecting a locked layer via layer panel, forget old selection and only select the locked layer
* [x] When shift-selecting a non-locked layer via canvas, if old selection was a locked layer, only select new element
* [x] Do not show links when hovering locked elements
* [x] When a locked element is selected, do not allow any transformation of the element using moveable (hide all handles and use default cursors)
* [x] When a locked media or shape element is selected, allow right-click to open context menu
* [x] When a locked non-media/non-shape element is selected, allow right-click to open context menu
* [x] When a locked element is selected, do not allow movement via keyboard arrows
* [x] When focusing a locked element on canvas, make lock status part of element accessible name
* [x] When editing a locked media element, do not allow resizing the frame
* [x] When focusing a locked element, announce lock state as part of element name for screen-readers.
* [x] When a locked text element is selected, do not allow edit mode using enter key, any key, or click inside
* [x] Add unit tests for changes to story reducer
* [x] Add karma tests for changes to layer selection
* [x] Add karma tests for changes to lasso selection
* [x] Add karma tests for changes to canvas selection

## User-facing changes

_TODO_

## Testing Instructions

This PR can be tested by following these steps:

1. Add multiple elements to canvas of different types
2. Stack some on top of each other
3. Lock some layers, especially some of the top ones
4. Observe that layer and element selection makes sense both using canvas, lasso, and layer panel.

[...]

PR #11477, opened 2022-05-12
Editor: Layer renaming

### LINKED ISSUE #9248: Nameable Layers
## Summary

Discussed and requested by design in conjunction with #8983 but isolated to make sure details are attended to accordingly.

Needs some discovery and syncing cross pods before this is dev ready. My understanding is this would be adding functionality that adds a new property getting saved to every story layer that would let them have a readable name separate from their id, etc. 

## References

https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor-Sprint-Feature-Level-Designs_?node-id=9935%3A421370
Screen Shot 2021-10-04 at 1 10 00 PM

## Acceptance Criteria

- [ ] You can rename the layer by double-clicking on it in the Layers Panel, using Canvas RCM->Rename layer or Layers Panel context menu.
- [ ] In edit mode: Enter key and blur event will save changes, Escape key will cancel them.

## Other notes 

Naming layers is going to impact keyboard users because right now the layers panel is hidden for keyboards as all functionality is mirrored in keyboard shortcuts. The accessibility of the layers panel will need to be reconsidered ahead of this functionality getting added to the layers panel.

marianakyrkosh:
## Context

Editing layers names

## Summary

Editing of the layers names can be enabled by:
- double clicking on the layers names;
- clicking Rename Layer in the RCM of the element;
- clicking Rename Layer in the RCM of the layers menu;

A name is saved by clicking Enter, clicking outside of the layer button (Blur).
The editing is canceled without saving a new name buy clicking Escape.

## Relevant Technical Choices

layerName property was added to the element object.

## To-do

1. Several components can be extracted from the layers.js file to separate files.
2. Keyboard events handler functions can be extracted to utils so they can be reused.

## User-facing changes

image
image
image

## Testing Instructions

Renaming successful scenario:
1. Double click on the Name -> an input appears -> type a new name -> click Enter -> the layer has a new name.
2. Right click on the element -> a menu opens -> click Rename Layer -> an input appears -> type a new name -> click Enter -> the layer has a new name.
3. Right click on a layer button in the Layers menu -> a menu opens -> click Rename Layer -> an input appears -> type a new name -> click Enter -> the layer has a new name.
4. Repeat 1 - 4, but instead of clicking Enter, click outside of the layer button (Blur).

Renaming canceled scenario:
5. Repeat 1 - 4, but click Escape button instead of Enter -> the layer name should not be updated.

Note: when an input field is visible, layer action icons should be hidden.

## Checklist

---

Fixes #9248


packages/story-editor/src/components/panels/design/layer/layer.js
```diff
@@ -264,6 +308,11 @@ function preventReorder(e) {
 }
 
 function Layer({ element }) {
+  const [layerName, setLayerName] = useState(
+    element.layerName
+      ? element.layerName
+      : __('Renaming a layer.', 'web-stories')
```
barklund: This has to use the current "default" name that is returned by `LayerContent`. Currently this returns a component, but in my PR in #11429 it has been changed to a function, `getLayerText`, returning a simple string.

I have split the relevant part of the code into a new separate PR in #11485. Feel free to merge that one as soon as it is approved and then you can merge main in to update this part here.

packages/story-editor/src/components/panels/design/layer/layer.js
```diff
@@ -302,42 +351,93 @@ function Layer({ element }) {
...
+            <LayerIcon
+              element={element}
+              getProxiedUrl={getProxiedUrl}
+              currentPageBackgroundColor={currentPageBackgroundColor}
+            />
+          </LayerIconWrapper>
+          <LayerInputDescription>
+            <LayerInput
+              tabIndex={-1}
+              aria-label={__('Renaming a layer.', 'web-stories')}
```
barklund: I think the label should be `"Layer name"`, as that is what you're editing in the input.

packages/story-editor/src/components/panels/design/layer/useLayerSelection.js
```diff
@@ -44,10 +47,27 @@ function useLayerSelection(layer) {
         shiftKey: evt.shiftKey,
       });
 
+      if (evt.detail === 2 && !isLocked && !isBackground) {
```
barklund: The "detail === 2" part here definitely needs a comment (that it's a double click).

Alternatively, we have a `useDoubleClick` that we use elsewhere for these events.

barklund: Also, I think that locked layers can still be renamed. I don't see why not at least. Figma allows this, so I think we should just do that too.

packages/story-editor/src/components/panels/design/layer/useLayerSelection.js
```diff
@@ -44,10 +47,27 @@ function useLayerSelection(layer) {
...
+        updateElementById({
+          elementId: elementId,
+          properties: { isEditable: true },
```
barklund: This property doesn't belong here at all. The elements in the story are persisted to the database and changes add an entry into the history.

Saving this information directly on the element has a number of issues:

- It adds history entries
- It is stored to the database
- It causes needless re-renders
- It allows multiple layers to be renamed at the same time

See this for some issues, where I have multiple layers with an input field:
Screen Shot 2022-05-12 at 21 26 40

I understand that this information needs to be available in both the canvas and the layer panel, so it has to be stored somewhere "high" in the component tree, but it should not be in the story provider directly. Perhaps it could be a new property in the canvas provider, with state`renamableLayer` and actions `setRenamableLayer` and `clearRenamableLayer` and then `null` means no layer is renamable, and some string means that the layer with the ID matching that string is renamable?

packages/story-editor/src/components/panels/design/layer/layer.js
```diff
@@ -25,8 +25,9 @@ import {
   BUTTON_TYPES,
   Icons,
   themeHelpers,
+  Input,
 } from '@googleforcreators/design-system';
-import { useRef, memo } from '@googleforcreators/react';
+import { useRef, memo, useState } from '@googleforcreators/react';
```
barklund, line 32: This file is definitely getting too long. I also have additional edits in #114290, so if you have some extra time, it would be refactor this somewhat, alternatively just create a P2 follow-up ticket to refactor this component.

packages/story-editor/src/app/rightClickMenu/items/layerName.js
```diff
@@ -0,0 +1,61 @@
...
+function LayerName() {
+  const updateSelectedElements = useStory(
+    (ctx) => ctx.actions.updateSelectedElements
+  );
+  const toggleLayerNaming = useCallback(
+    () =>
+      updateSelectedElements({
+        properties: (oldElement) => ({
+          ...oldElement,
+          isEditable: oldElement.isLocked ? false : true,
```
barklund: This should again not go through the story provider, but some other provider. Also, why is the editable property set to the opposite of the locked property? And why not just use `!`?

barklund: Oh yeah, and we need the property, wherever it ends up existing, to be called something like `renamable` or `changeLayerName` or something like that. `editable` is too broad a term, that we also use for other things, so it can get confusing (we have an edit-mode for elements, but it is completely unrelated to this property).

barklund: Finally, the input field needs to be focused, and the layer panel needs to be opened if not already opened, when you right click to rename an element.

This should be done after the refactoring to a property on (e.g.) the canvas provider and then has to be done in two places:

1. In the footer layer button in `components/footer/layers`, the `isOpen` property should be set to true in an effect, if `renamableLayer` is non-null. That would force-open the dialog.
2. In the layer in `components/panels/design/layer/layer.js`, the input field should be focused in an effect, if `renamableLayer` is set to the specific layer. That would force-focus the input and it would work even if the layer panel was originally hidden.

packages/story-editor/src/components/panels/design/layer/layer.js
barklund: Come to think of it, the whole layer panel should even live inside `panels/design/` at all anymore. It's hardly a panel (doesn't really use any panel bits, except for an unpadded `PanelContent`, which could be removed), and it's definitely not related to the design panel at all.

So, follow-up ticket to refactor this component as well as move the whole layer dialog to a fresh location (probably as `components/layers` or something like that).

packages/story-editor/src/components/panels/design/layer/layer.js
barklund: Actually, the layer name, if provided, should be returned by all those `getLayerText` functions in all the different elements. It's not very DRY to go around adding this to 5 different functions.

Actually a better way to do it would be to create a single `getLayerName` function in the elements package, that would first look at `element.layerName` and if not set, use the layer name for the element type. This could also factor in the background layer name too. That's a much cleaner way to do it actually! I will amend #11485 to do that instead!

That would result in the custom layer name as entered by the user would also be used in the right-click select menu and as the aria-label for the element on the canvas (neither of which works currently).

See below, where the layer name is not used in the right-click menu:
Screen Shot 2022-05-12 at 21 43 05

barklund: I did this slight refactor in 57c7b283bc.

barklund (dismissed):
This is really good, but just has a few issues.

- The current layer name should be extracted from the element type, but this work is already done in #11485, so feel free to merge that in once merged to main (@merapi will probably do that first thing Friday)
- The state of whether a layer is renamable should not exist in the story provider, but in some other location. Also, only one layer should ever be renamable.
- If a layer is renamed, the input should be focused (and the layer dialog should be open to do that). It should never be possible to have the rename input unfocused at all.

I have suggested fixes for all these inline below. I will tentatively approve pending the implementation of the above changes.

packages/story-editor/src/components/panels/design/layer/layer.js
marianakyrkosh: It's 'Renaming a layer.' in the Figma  design (the link to the design is in the ticket).

packages/story-editor/src/components/panels/design/layer/useLayerSelection.js
merapi: +1 for `renamableLayer` (id of the layer currently being renamed). CanvasProvider can do the job (spans canvas and layers panel).

packages/story-editor/src/components/panels/design/layer/layer.js
marianakyrkosh: Done

packages/story-editor/src/components/panels/design/layer/useLayerSelection.js
marianakyrkosh: Done

packages/story-editor/src/components/panels/design/layer/layer.js
merapi: `Renaming a layer.` is just an example layer name in Figma, right? the input itself holds the layer name, so `Layer name` seems to be a good label for it.

packages/story-editor/src/components/panels/design/layer/useLayerSelection.js
merapi: @barklund WDYT about renaming the background layer? (`!isBackground` is now removed). Should we ask UX?

packages/story-editor/src/components/panels/design/layer/useLayerSelection.js
marianakyrkosh: Moved code to the canvasProvider and fixed multiple renamable layers bug.

packages/story-editor/src/app/rightClickMenu/items/layerName.js
marianakyrkosh: Done

merapi (changes requested):
I'm reporting what I see in the current version:

1. After entering the edit mode (via double click) you cannot select a part of the text, instead it cancels the edit mode.

2. Go into edit mode, type something and cancel the edit mode (press Escape). The layer name is reverted to the old name - good, but when you enter the edit mode again you see the old (canceled) value.

3. image
This^ should look like it looked before (and in Figma). Those extra icons are unnecessary (I deleted my layer by accident when selecting text).

4. If the Background layer name cannot be changed then you shouldn't be able to enter the edit mode on that layer.

packages/story-editor/src/components/panels/design/layer/layer.js
barklund: * Added new issue #11576

merapi (dismissed):
This looks good now 👍🏻 

Nit: handling empty layer name.
Right now if you add layer `media/unsplash:gl8BgGhvVdg`, rename it to `layer1`, then rename it to empty string, it will become `media/unsplash:gl8BgGhvVdg`, but it should probably become `layer1`? Then if you enter the edit mode again - it will show an empty input (should always be the layer name visible before entering the edit mode?).

And a thing to consider: if we remove the `margin-left: 2px`, then the layer name will not shift when you enter the edit mode - it looks better IMO (even if it's 2px more over the icon).

merapi (approved):

PR #11490, opened 2022-05-16
Avoid setting numeric story slugs

swissspidy:
## Context

There was an edge case issue with saving stories without titles, resulting in console errors due to an error response from the REST API.

Previously, if the story doesn't have a title, the `updateSlug` util (which is called when clicking on the title input and then out of it again) set the current story ID (which is a number) as the story slug.

That is not correct, because the story slug must be a string. Also, there's no need to set the story ID as the slug anyway; it can just continue to be an empty string.

## Summary

Fixes `updateSlug` by ensuring that the slug is an empty string and not using the story ID at all.

Not sure why this was done this way, especially on blur. But alas.

## Testing Instructions

1. Create a new story
2. Click on title input field
3. Click somewhere else, add some content to the story
4. Save the story
5. Verify that there are no console errors

No

No

No

## Checklist

---

Fixes #

### DIFF
--- a/packages/story-editor/src/components/header/title.js
+++ b/packages/story-editor/src/components/header/title.js
@@ -26,7 +26,6 @@ import { themeHelpers } from '@googleforcreators/design-system';
  * Internal dependencies
  */
 import { useStory } from '../../app/story';
-import { useConfig } from '../../app/config';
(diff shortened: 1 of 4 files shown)


spacedmonkey (approved):
Seems to work well.

PR #11506, opened 2022-05-17
Admin UI issues

### LINKED ISSUE #5076: UI difficulties when admin menu is long
## Bug Description

Splitting this out from #4517

There are issues when the admin menu excessively long (e.g. when lots of plugins are active)

## Expected Behaviour

The editor should be usable even if the admin menu is excessively long or sometimes also when the screen is smaller

## Steps to Reproduce

1. Install many plugins to make admin menu super long **or** just make your window smaller
1. Observe behavior

## Screenshots

Screen Shot 2020-10-05 at 11 02 54 AM

## Additional Context

- Plugin Version: 1.1.0
- Operating System: macOS
- Browser: Chrome 86

---

mwritter:
## Context

Allow the user to have lots of plugins and/or work with a small screen and not have a blank space under the editor.

## Summary

Sets the wp content css style to span the entire page, also sets `ignoreMaxOffsetY` to `true` for all tooltips so that the tooltips stay in place on scroll (now that the wp content has fixed position it shouldn't scroll, the admin plugin menu scrolls)

## Relevant Technical Choices

Setting the wp content `position` to `fixed` allows the content to span the entire page, while still maintaining the correct functionality for the admin plugin menu.  The tooltips `ignoreMaxOffsetY` was set to `true` so that the tooltips do not scroll with the admin plugin menu

## User-facing changes

before/after
https://user-images.githubusercontent.com/14057928/168895399-10e338bd-5702-4e19-92f4-478fc4841175.mp4

## Testing Instructions

1. Go to the Story Editor as admin
2. Adjust the window height where some of the admin plugins can't be seen

On Staging:
- Notice when you scroll the canvas there is a blank section under the editor and the admin plugin menu spans the full screen

On Branch:
- Notice when you scroll the canvas stays fixed and the admin plugin menu scrolls

Notes: this change adjusts the position of the wp content which could cause UI discrepancies - the tooltips needed to be adjusted to display correctly, there may be other UI elements that need tweaking, please test popups and tooltips and scroll functionality.  

No

No

No

## Checklist

---

Fixes #5076

### DIFF
--- a/packages/design-system/src/components/tooltip/index.js
+++ b/packages/design-system/src/components/tooltip/index.js
@@ -100,10 +100,6 @@ let lastVisibleDelayedTooltip = null;
  * @param {string} props.className Classname.
  * @param {string} props.isDelayed If this tooltip is to be displayed instantly on hover (default) or by a short delay.
  * @param {number} props.popupZIndexOverride If present, passes an override for z-index to popup
- * @param {boolean} props.ignoreMaxOffsetY  Defaults to false. Sometimes, we want the popup to respect the y value
- * as perceived by the page because of scroll. This is really only true of dropDowns that
- * exist beyond the initial page scroll. Because the editor is a fixed view this only
- * comes up in peripheral pages (dashboard, settings).
  * @return {import('react').Component} BaseTooltip element
  */
 
@@ -120,7 +116,6 @@ function BaseTooltip({
   // to the whole canvas
   className = null,
   popupZIndexOverride,
-  ignoreMaxOffsetY = false,
   styleOverride,
   ...props
 }) {
@@ -396,7 +391,6 @@ const BaseTooltipPropTypes = {
   className: PropTypes.string,
   isDelayed: PropTypes.bool,
   popupZIndexOverride: PropTypes.number,
-  ignoreMaxOffsetY: PropTypes.bool,
 };
 BaseTooltip.propTypes = BaseTooltipPropTypes;
 
--- a/packages/design-system/src/components/tooltip/index.js
+++ b/packages/design-system/src/components/tooltip/index.js
@@ -166,11 +166,11 @@ function BaseTooltip({
             anchor: forceAnchorRef || anchorRef,
             popup,
             isRTL,
-            ignoreMaxOffsetY,
+            ignoreMaxOffsetY: true,
           })
         : {},
     });
-  }, [dynamicPlacement, spacing, forceAnchorRef, isRTL, ignoreMaxOffsetY]);
+  }, [dynamicPlacement, spacing, forceAnchorRef, isRTL]);
 
   // When near the edge of the viewport we want to force the tooltip to a new placement as to not
   // cutoff the contents of the tooltip.
--- a/packages/wp-story-editor/src/style.css
+++ b/packages/wp-story-editor/src/style.css
@@ -28,7 +28,7 @@
 
 body.js.edit-story #wpcontent {
   position: fixed;
-  width: calc(100% - 36px);
+  width: 100%;
 }
 
 body.js.edit-story #wpcontent,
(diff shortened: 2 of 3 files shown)


mwritter:
@BrookeGraham - I noticed you did some work with the tootltips / popups recently, with the css fix I implemented the tooltips were scrolling with the page/admin menu while the story editor now is `fixed` and doesn't move with the tooltip. So I hardcoded the tooltips to just `ignoreMaxOffsetY` so they stay in place on scroll.  Are there any glaring issues that come to mind with this fix? Any insight would be greatly appreciated thanks!

BrookeGraham:
I can take a look on Thursday. I’ll be out Wednesday.

packages/wp-story-editor/src/style.css
```diff
@@ -26,6 +26,11 @@
   display: none;
 }
 
+body.js.edit-story #wpcontent {
+  position: fixed;
+  width: calc(100% - 36px);
```
BrookeGraham: This `width` won't be correct when the admin menu nav is expanded since `MENU_WIDTH` is 160 px whereas `MENU_FOLDED_WIDTH` is 36 px

packages/design-system/src/components/tooltip/index.js
```diff
@@ -166,11 +166,11 @@ function BaseTooltip({
             anchor: forceAnchorRef || anchorRef,
             popup,
             isRTL,
-            ignoreMaxOffsetY,
+            ignoreMaxOffsetY: true,
```
BrookeGraham, line 164: If you explicitly set this to `true`, you won't need to pull it in as a prop any longer, which should be fine for tooltip.  I don't think it gets used anywhere. It was added for `Dropdowns`, which use `Popup`, once we separated `Tooltip` from `Popup` it was no longer need as a prop for `Tooltip`.  The `getOffset` util is still shared between the two which is why it was getting sent down. So long story short, this should be 💯  We had plans to pull separate out `Dropdown` and `Popup` as well, but decided it wasn't really worth it.

mwritter: Okay cool, Thanks for looking into that and for the insight! Ill remove this prop 👍

packages/wp-story-editor/src/style.css
mwritter: Okay gotcha Ill look into that

sblinde (dismissed):
🐝

I think the proposed solution here is solid, but I'd love to see it tested by someone who has been closer to the project for longer than our team, with special attention to dropdown menus and tooltips and other parts of the editor.

BrookeGraham:
The issue I see here is that when the admin menu is expanded horizontally, we cut off the right edge of the editor. 
Kapture 2022-05-23 at 11 31 35

mwritter:
> The issue I see here is that when the admin menu is expanded horizontally, we cut off the right edge of the editor. Kapture 2022-05-23 at 11 31 35 [ ](https://user-images.githubusercontent.com/1820266/169875408-6124d943-a4d6-48c3-b9d2-4ca801d32b9a.gif) Kapture 2022-05-23 at 11 31 35 [ ](https://user-images.githubusercontent.com/1820266/169875408-6124d943-a4d6-48c3-b9d2-4ca801d32b9a.gif)

Thanks for finding that! I didn't know that was a collapse/expand icon.. 😅

sblinde (approved):
Adding my approval back

packages/wp-story-editor/src/style.css
```diff
@@ -26,6 +26,15 @@
   display: none;
 }
 
+body.js.edit-story #wpcontent {
+  position: fixed;
+  width: calc(100% - 160px);
+}
+
+body.js.edit-story.folded #adminmenuback {
+  top: 32px;
```
BrookeGraham, line 35: should `body.js.edit-story.folded` be `top: 32` or `right: 32` 🤔

mwritter: #adminmenuback is the div behind the #adminmenuwrap (the menu bar that holds the plugins), which makes the #adminmenuwrap look like its the full length of the page.  #adminmenuback covers up the #wpadminbar (menu bar at the top of the screen) with this fix so I had to move #adminmenuback down by `32px` (the height of #wpadminbar) so that it doesn't cover up #wpadminbar. 😅 it is a little confusing.

PR #11508, opened 2022-05-17
Animations: Disallow negative Duration or Delay values

### LINKED ISSUE #11439: Animations: Editor crashes when using negative duration values
## Bug Description

The editor crashes when trying to set a negative number as the animation duration.

## Expected Behaviour

No crash.

Negative duration values are not allowed, so we can just revert to the number `0` in that case.

This what we already do for the delay.

## Steps to Reproduce

1. Add new element to page
1. Add animation
1. Set animation duration to a negative number
1. See editor crash

## Screenshots

Screenshot 2022-05-05 at 12 34 47

## Additional Context

- Plugin Version: 1.20.0
- WordPress Version: 5.9
- Operating System: macOS
- Browser: Chrome

mflorida:
## Context

The UI would crash when a user entered a negative value for one of the animation properties.

## Summary

Changes to the `sanitizeTimings` function as well as the `onChange` handler were made to reset negative values set by the user to `0`.

## Relevant Technical Choices

Negative values are caught and reset to `0` before those values can be applied to the Duration or Delay parameters.

## To-do

None

## User-facing changes

Negative values can be entered, but will be reset to `0` to avoid a UI crash.

**Before**
https://user-images.githubusercontent.com/4173034/168925062-015dc285-1a72-4edb-92db-7735269cb242.mov

**After**
https://user-images.githubusercontent.com/4173034/168925040-79df4209-b77d-4a8e-8a3a-3e203f282e8d.mov

## Testing Instructions

- For a new or existing story, add animation to a page element (on a page that allows animations)
- Enter a negative value into the `Duration` or `Delay` field
- Value should reset to `0` on blur or <key>Enter</key>

1. Clone this branch
2. Startup a local dev server
3. Follow testing instructions above

No

No

No

## Checklist

---

Fixes #11439

### DIFF
--- a/packages/animation/src/parts/simpleAnimation.js
+++ b/packages/animation/src/parts/simpleAnimation.js
@@ -26,9 +26,11 @@ import { AnimationOutput, WithAnimation } from '../outputs';
 import getInitialStyleFromKeyframes from '../utils/getInitialStyleFromKeyframes';
 import { AMPAnimationProps } from './types';
 
-export const sanitizeTimings = (timings) => ({
-  ...timings,
-  easing: timings.easing || 'linear',
+export const sanitizeTimings = ({ easing, duration, delay, ...other }) => ({
+  ...other,
+  easing: easing || 'linear',
+  duration: !isNaN(Number(duration)) ? Math.max(duration, 0) : 0,
+  delay: !isNaN(Number(delay)) ? Math.max(delay, 0) : 0,
 });
 
 function SimpleAnimation(
--- a/packages/story-editor/src/components/panels/design/animation/effectPanel.js
+++ b/packages/story-editor/src/components/panels/design/animation/effectPanel.js
@@ -126,7 +126,7 @@ function EffectPanel({
         effectConfig={config}
         field={field}
         onChange={(value, submitArg) =>
-          handleInputChange({ [field]: value }, submitArg)
+          handleInputChange({ [field]: Math.max(value, 0) }, submitArg)
         }
         disabledOptions={disabledTypeOptionsMap[type]?.options || []}
         disabled={disabled}
--- a/packages/animation/src/parts/test/index.js
+++ b/packages/animation/src/parts/test/index.js
@@ -17,6 +17,7 @@
  * Internal dependencies
  */
 import { ANIMATION_TYPES } from '../../constants';
+import { sanitizeTimings } from '../simpleAnimation';
 import { AnimationPart, throughput } from '..';
 
 describe('AnimationPart', () => {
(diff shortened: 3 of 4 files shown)


packages/animation/src/parts/simpleAnimation.js
```diff
@@ -26,9 +26,11 @@ import { AnimationOutput, WithAnimation } from '../outputs';
 import getInitialStyleFromKeyframes from '../utils/getInitialStyleFromKeyframes';
 import { AMPAnimationProps } from './types';
 
-export const sanitizeTimings = (timings) => ({
-  ...timings,
-  easing: timings.easing || 'linear',
+export const sanitizeTimings = ({ easing, duration, delay, ...other }) => ({
+  ...other,
+  easing: easing || 'linear',
+  duration: !isNaN(Number(duration)) ? Math.max(duration, 0) : 0,
```
sblinde, line 32: Wow, I really like this

sblinde (dismissed):
🐝

Nicely done

swissspidy (dismissed):

packages/story-editor/src/components/panels/design/animation/effectPanel.js
```diff
@@ -126,7 +126,7 @@ function EffectPanel({
         effectConfig={config}
         field={field}
         onChange={(value, submitArg) =>
-          handleInputChange({ [field]: value }, submitArg)
+          handleInputChange({ [field]: Math.max(value, 0) }, submitArg)
```
mwritter, line 129: The effectInput looks like it has a defualt min property set to 0, and it is a numericInput, which has a validation process already in place in useNumericInput.js parseInput 🤔 I think the underlying issue here is that we're overriding that validation here. It might be a better fix for all EffectInputs if we update effectInput.js to use the correctly validated value instead.

sblinde (approved):
🐝

We can move forward with this fix.

packages/story-editor/src/components/panels/design/animation/effectPanel.js
sblinde: I think this can tackle what you've suggested: https://github.com/GoogleForCreators/web-stories-wp/issues/11530

I think it's worth investigating further. This fix does fix the issue at hand, but we're coming into a rabbit hole sort of situation as we dig further, so I think for code quality sake, it warrants investigation when we can prioritize doing so.

packages/animation/src/parts/simpleAnimation.js
mflorida: The general fix was suggested by @swissspidy. 👍

PR #11510, opened 2022-05-17
Media: Refactor hotlinking dialog to be more reusable

### LINKED ISSUE #11143: Media: Refactor hotlinking dialog to be more reusable
## Feature Description

Now the hotlinking of assets being supported in multiple places, see https://github.com/GoogleForCreators/web-stories-wp/issues/10888 https://github.com/GoogleForCreators/web-stories-wp/issues/10515, we the existing hotlinking dialog, should be refactor to make reusable in more contexts. 

At the same time, we should add some unit tests and storybook integration.

spacedmonkey:
## Summary

Refactor share code into share component.

## Testing Instructions

1.

## Checklist

---

Fixes #11143


swissspidy:
The two `useInsert` hooks still have some shared code.

I feel like it should be possible to move more logic (such as the `getHotlinkInfo()` part) to `packages/story-editor/src/components/hotlinkModal/hotlinkModal.js`

Here's an attempt: #11522

packages/story-editor/src/components/hotlinkModal/useHotlinkModal.js
```diff
@@ -0,0 +1,146 @@
...
+
+    setIsInserting(true);
+
+    try {
+      const hotlinkInfo = await getHotlinkInfo(link);
+      if (!allowedFileTypes.includes(hotlinkInfo?.ext)) {
+        const error = new Error(__('Invalid link.', 'web-stories'));
+        error.code = 'invalid_ext';
+        throw error;
+      }
```
swissspidy: Is this a new check? I don't see a similar file type check on `main` 🤔 

It's confusing to throw an error here that is caught in the same block anyway—basically a `goto`.

Plus, I don't think this should call `onError` (and thus `trackError`)

So would just do something simple like this:

```suggestion
      if (!allowedFileTypes.includes(hotlinkInfo?.ext)) {
        setErrorMsg(__('Invalid link.', 'web-stories'));
        return;
      }
```

packages/story-editor/src/components/hotlinkModal/useHotlinkModal.js
```diff
@@ -0,0 +1,146 @@
...
+import {
+  getHotlinkDescription,
+  isValidUrlForHotlinking,
+  getErrorMessage,
+} from './utils';
+
+function useHotlinkModal({ onSelect, onClose, onError, allowedFileTypes }) {
+  const [isInserting, setIsInserting] = useState(false);
+  const [link, setLink] = useState('');
+  const [errorMsg, setErrorMsg] = useState(false);
```
swissspidy:
```suggestion
  const [errorMsg, setErrorMsg] = useState(null);
```

This helps avoid type warnings (for example in PhpStorm but theoretically also in TypeScript) when later setting strings here when the initial value is a boolean.

I know this might have been different on `main`, but it's a good opportunity to change this.

packages/story-editor/src/components/hotlinkModal/useHotlinkModal.js
```diff
@@ -0,0 +1,146 @@
...
+      if (!allowedFileTypes.includes(hotlinkInfo?.ext)) {
+        const error = new Error(__('Invalid link.', 'web-stories'));
+        error.code = 'invalid_ext';
+        throw error;
+      }
+      const needsProxy = await checkResourceAccess(link);
+
+      await onSelect({ link, hotlinkInfo, needsProxy });
+      setLink('');
+      setErrorMsg(false);
```
swissspidy:
```suggestion
      setErrorMsg(null);
```

Related to my other comment about types.

I know this might have been different on `main`, but it's a good opportunity to change this.

packages/story-editor/src/components/hotlinkModal/utils.js
```diff
@@ -41,6 +41,7 @@ export function getErrorMessage(code, description) {
     case 'rest_invalid_param':
     case 'rest_invalid_url':
       return __('Invalid link.', 'web-stories');
+    case 'invalid_ext':
```
swissspidy: Related to my other suggestion, let's undo this.

packages/story-editor/src/components/hotlinkModal/useHotlinkModal.js
spacedmonkey: This is a new fix. In main, I was able to add a link to mp3 for image / video hotlinking with an error showing.

spacedmonkey: Would we not work to track this as an error?

packages/story-editor/src/components/hotlinkModal/useHotlinkModal.js
spacedmonkey: Fixed.

packages/story-editor/src/components/hotlinkModal/useHotlinkModal.js
swissspidy: Not interested in tracking this as an error for now

spacedmonkey: Fixed.

swissspidy (approved):

swissspidy:
Always nice when you're able to delete code like this 💪 

Merging this now to unblock the other PRs.

PR #11525, opened 2022-05-19
Hotlinking: Do not use CORS proxy for captions and background audio

### LINKED ISSUE #11479: Media: Require CORS for captions hotlinking
## Bug Description

Subtitles are not displayed on the frontend when the VTT file does not have the required CORS headers set.

## Expected Behaviour

Handling of caption files with missing CORS header needs to be improved.

First, the parent `<amp-video>` needs `crossorigin="anonymous"` for it to work ([see docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/track#attr-src)).
Second, if the VTT file doesn’t have CORS headers it won’t work.

Since we can't use the CORS proxy on the frontend, we must not use the CORS proxy for captions in the editor, as it gives the false impression of any VTT file working. Basically give an error message straight away.

## Steps to Reproduce

1. Start a new story
2. Click on the background
3. Click on Style tab
4. Click on "Link to audio file" and paste https://stories-qa-wordpress-amp.pantheonsite.io/wp-content/uploads/2022/02/file_example_MP3_1MG.mp3
5. Click on "Use audio file"
6. Click on "Link to caption file" and paste https://wp.stories.google/stories/how-to-optimize-videos-for-web-stories/tracks/wst024_page2_my_captions.vtt
7. Click on "Use caption"
8. Click on preview story button - the audio will play, but the subtitles won't be displayed

## Screenshots

https://images.zenhubusercontent.com/235435637/e8191c4e-1967-4674-b69d-0a59a48a56ee/no_subtitles.mp4

## Additional Context

- Plugin Version: 1.21.0-alpha
- WordPress Version: 5.9
- Operating System: macOS
- Browser: Chrome

swissspidy:
## Context

Video tracks require proper CORS headers being set, otherwise they won't load on the frontend. So our CORS proxy workaround is useless.

## Summary

This PR changes the hotlinking modal so that it fails when there is a CORS issue when trying to insert VTT files or audio files.

## User-facing changes

No visual change initially in the dialog:
Screenshot 2022-05-19 at 14 20 41

But then the new error message:

Screenshot 2022-05-19 at 14 25 13

## Testing Instructions

1. Try to insert a file like https://api.ardmediathek.de/subtitle-format-service/webvtt/urn:ard:subtitle:74cfd50da83320c6.vtt for background audio or video captions.
2. Verify that it works as expected and the captions are visible when viewing the story (ignore any styling differences that may appear)
3. Try to insert a file like https://wp.stories.google/stories/how-to-optimize-videos-for-web-stories/tracks/wst024_page2_my_captions.vtt for captions and see an error message about CORS

No

No

No

## Checklist

---

Fixes #11479


packages/output/src/utils/backgroundAudio.js
```diff
@@ -43,6 +43,7 @@ function BackgroundAudio({ backgroundAudio, id }) {
         autoPlay="autoplay"
         layout="nodisplay"
         poster=""
+        crossorigin="anonymous"
```
spacedmonkey: We could optionally put this attribute on if needed?

packages/story-editor/src/components/panels/shared/captionsPanelContent.js
```diff
@@ -184,6 +184,7 @@ function CaptionsPanelContent({
                 allowedFileTypes={allowedFileTypes}
                 insertText={__('Use caption', 'web-stories')}
                 insertingText={__('Selecting caption', 'web-stories')}
+                canUseProxy={false}
```
spacedmonkey, line 205: What is the point of the property if it is always false?

packages/output/src/utils/backgroundAudio.js
swissspidy: I thought about it, but there's no harm in always adding it. Avoids having to add extra complexity for that.

packages/story-editor/src/components/panels/shared/captionsPanelContent.js
swissspidy: It default's to `true` in the component. But for captions it's disabled.

spacedmonkey:
I think we should try and get https://github.com/GoogleForCreators/web-stories-wp/pull/11510 merged first. It will make porting the change easier.

packages/story-editor/src/components/panels/shared/hotlinkModal.js
```diff
@@ -40,12 +50,39 @@ import {
...
+const DOCS_URL =
+  'https://wp.stories.google/docs/troubleshooting/common-issues/';
+
 const InputWrapper = styled.form`
   margin: 16px 4px;
   width: 470px;
   height: 100px;
 `;
 
+function CORSMessage() {
```
spacedmonkey: Does this make sense to be in it's own file?

packages/story-editor/src/components/panels/shared/hotlinkModal.js
```diff
@@ -125,6 +163,12 @@ function HotlinkModal({
       const hotlinkInfo = await getHotlinkInfo(link);
       const shouldProxy = await checkResourceAccess(link);
 
+      if (shouldProxy && !canUseProxy) {
```
spacedmonkey: In my PR, I do something similar. But instead of returning early, I throw a error. See https://github.com/GoogleForCreators/web-stories-wp/pull/11510/files#diff-63f158a92c1b323c663fb4b9fcf8a3d81ed7097c0dde851f6bec2a4d6812631fR90

Could we do something similar.

packages/story-editor/src/components/panels/shared/hotlinkModal.js
```diff
@@ -125,6 +163,12 @@ function HotlinkModal({
...
+        setErrorMsg(<CORSMessage />);
```
spacedmonkey: Other error messages are in this handled in this function. https://github.com/GoogleForCreators/web-stories-wp/blob/256639d8367d95e165e38ca5e733ae0d847ff1f4/packages/story-editor/src/components/library/panes/media/local/hotlink/utils.js#L39-L55

packages/story-editor/src/components/panels/shared/hotlinkModal.js
swissspidy: The difference is that throwing an error like in your PR would mean calling `onError()` (and thus `trackError()`), which I don't want for this case, which is why this only sets the message but nothing else.

spacedmonkey: Right, that is a good point. Might be worth a comment, that not tracking the error is by design here.

packages/story-editor/src/components/panels/shared/hotlinkModal.js
swissspidy: I don't see a reason to move this to a separate file. It's specific to this file and not used elsewhere. Plus it's super simple and basically a simple string.

packages/story-editor/src/components/hotlinkModal/useHotlinkModal.js
```diff
@@ -89,11 +125,16 @@ function useHotlinkModal({ onSelect, onClose, onError, allowedFileTypes }) {
       const hotlinkInfo = await getHotlinkInfo(link);
       if (!allowedFileTypes.includes(hotlinkInfo?.ext)) {
         setErrorMsg(__('Invalid link.', 'web-stories'));
-        setIsInserting(false);
```
spacedmonkey, line 92: Not needed because of the finally?

packages/story-editor/src/components/hotlinkModal/useHotlinkModal.js
```diff
@@ -31,7 +34,40 @@ import {
   getErrorMessage,
 } from './utils';
 
-function useHotlinkModal({ onSelect, onClose, onError, allowedFileTypes }) {
+function CORSMessage() {
```
spacedmonkey: Having a component in a file for a hook, now feels very strange. For tidiness sake ( not reusable ), I think this should be in it's own file. There should even be a storybook integration that way as well.

packages/story-editor/src/components/hotlinkModal/useHotlinkModal.js
swissspidy: Correct

packages/story-editor/src/components/hotlinkModal/useHotlinkModal.js
swissspidy: This is not a storybook-worthy component. It's simply a string with an HTML tag in it, which is why it uses `TranslateWithMarkup`. I can move it to `utils` if that makes it more tidy.

spacedmonkey (dismissed):

spacedmonkey (approved):

PR #11526, opened 2022-05-19
Dashboard filter by taxonomy

### LINKED ISSUE #11436: Dashboard: Add Filter by Taxonomy
## Summary

This issue will add a new filter option to the Dashboard to filter Stories, based on hierarchical taxonomy. 

Screen Shot 2022-04-20 at 11 42 53 AM

## References

UX: Still in progress, but this issue's implementation can be started prior to having fleshed out comps for this. We can mimic the experience described below.

For an example of what this could look like, please see `All Stories` in the default WordPress view for what this means. The functionality between these two spaces should be the same: a dropdown, similar to the Author in the screenshot seen above.

## Alternatives Considered

No alternatives.

This should not impact performance.

We could capture usage of this dropdown via telemetry.

mwritter:
## Context

Add filter by story category to the dashboard.

## Summary

Give the user the ability to filter by story category.  Similar to the author filter.

## Relevant Technical Choices

Mirrors author filter on the dashboard.

## User-facing changes

before/after
https://user-images.githubusercontent.com/14057928/169389100-4fc7bd45-1174-4b4b-9640-0aa0e0a795e9.mp4

## Testing Instructions

1. Go to story dashboard

On Staging:
- Notice there is no 'Category' filter

On Branch:
- Notice there is a 'Category' filter
2. Make sure there are stories with 'categories' assigned to them
3. Test filter correctly filters stories
4. Test that the search within the filter dropdown is working correctly as well (to search for category to filter by)

No

No

No

## Checklist

---

Fixes #11436


packages/dashboard/src/app/views/myStories/index.js
```diff
@@ -129,14 +133,45 @@ function MyStories() {
...
+        }
+
+        const categoryData = data.map(({ id, name }) => ({
+          id,
+          name,
+        }));
+        setQueriedCategories((current) => {
+          const existingIds = current.map(({ id }) => id);
+          const newCategories = categoryData.filter(
+            (category) => !existingIds.includes(category.id)
```
sblinde: Linter called this out, but probably just need to rename category to something else here potentially

packages/dashboard/src/app/views/myStories/karma/myStories.karma.js
```diff
@@ -493,6 +493,41 @@ describe('CUJ: Creator can view their stories in grid view', () => {
         expect(authorEl).toBeDefined();
       }
     });
+
+    it('should filter by category', async () => {
+      let { stories } = await getStoriesState();
```
sblinde: Same with this linter warning, `const` would be better here

includes/Admin/Dashboard.php
```diff
@@ -433,6 +433,7 @@ public function get_dashboard_settings(): array {
 				'settings'       => '/web-stories/v1/settings/',
 				'pages'          => '/wp/v2/pages/',
 				'publisherLogos' => '/web-stories/v1/publisher-logos/',
+				'categories'     => '/web-stories/v1/web_story_category/',
```
mflorida: Since `web_story_category` is part of a URL path, it should be hyphenated `web-story-category` to follow patterns currently in use (and standard to WordPress).

mflorida (commented):
The only thing that jumps out at me is the `web_story_category` URL path should be hyphenated. Looks lovely otherwise.

includes/Admin/Dashboard.php
mflorida: This could probably be shortened to just `categories` since `web-stories` is already part of the path.

mwritter: 🤔 @spacedmonkey @mflorida it seems that `/web-stories/v1/web_story_category/` is the registered url for getting story categories, replacing the underscores with dashes will cause a `404`.  Im not sure if we want to make that change here.

mflorida: Ah ok - that’s an existing endpoint.

spacedmonkey:
So this only works for web stories categories? What about other taxonomies? @swissspidy should we support this functionality here?

packages/dashboard/src/app/views/shared/bodyViewOptions.js
```diff
@@ -91,6 +103,22 @@ export default function BodyViewOptions({
...
+                hasSearch
+                hasDropDownBorder
+                searchResultsLabel={__('Search results', 'web-stories')}
+                aria-label={__('Filter stories by category', 'web-stories')}
+                onChange={category.toggleFilterId}
+                getOptionsByQuery={queryCategoriesBySearch}
+                selectedId={category.filterId}
+                placeholder={__('Categories', 'web-stories')}
+                primaryOptions={category.queriedCategories}
+                options={category.queriedCategories}
```
spacedmonkey: This labels should come from here
https://github.com/GoogleForCreators/web-stories-wp/blob/85420b48d5aca0f5aae5d1452528f7d4f1fed302/includes/Taxonomy/Category_Taxonomy.php#L73-L95

spacedmonkey:
Instead of all the hardcoded values in this PR, I think that we should load the taxonomies endpoint and get values like that. 
Screenshot 2022-05-20 at 14 05 15

mwritter:
> So this only works for web stories categories? What about other taxonomies? @swissspidy should we support this functionality here?

@spacedmonkey Thank you for looking into this, and I appreciate the insight!  Ill look into use those labels, and using the actual taxonomies endpoint.  And yes currently this only supports filtering by story categories.  I believe the decision was made in one of our Design Sync meetings, @sblinde @agingoldseco.

sblinde:
> > So this only works for web stories categories? What about other taxonomies? @swissspidy should we support this functionality here?
> 
> @spacedmonkey Thank you for looking into this, and I appreciate the insight! Ill look into use those labels, and using the actual taxonomies endpoint. And yes currently this only supports filtering by story categories. I believe the decision was made in one of our Design Sync meetings, @sblinde @agingoldseco.

This is correct, and was some of the context shared prior to our team on the original issue linked ([see here](https://github.com/googleforcreators/web-stories-wp/issues/11287#issuecomment-1103717624)). Within our design meeting, we suggested we could see about further enhancing to include a second dropdown with tags in the future, but the scope of this story is to build similar functionality to the `All Categories` dropdown that exists within `Posts` in WordPress. Trying to merge the ideas of both tags and categories together (or to support custom taxonomies, which I found also is a thing that people could use if they had a plugin to do so) would significantly increase the scope of this work.

packages/dashboard/src/app/views/shared/bodyViewOptions.js
sblinde: Indeed, I think the `All Categories` one is probably the right fit for this (versus my original `Categories` in my first commit).

includes/Admin/Dashboard.php
swissspidy: Ideally we do not hardcode `web_story_category` here, but just pass `'taxonomies'     => '/web-stories/v1/taxonomies/',`

Then the dashboard can ping this endpoint to get a list of all hierarchical taxonomies and then display a dropdown for those.

So that this is not hardcoded to just this single taxonomy.

swissspidy:
To further clarify: Ideally this works for all _hierarchical_ taxonomies that exist, just like on the "All Stories" page, and is not hardcoded to this single `web-story-category` taxonomy.

Just like the story-editor does this, essentially.

sblinde:
> To further clarify: Ideally this works for all _hierarchical_ taxonomies that exist, just like on the "All Stories" page, and is not hardcoded to this single `web-story-category` taxonomy.
> 
> Just like the story-editor does this, essentially. _Just not for flat taxonomies like tags._

@swissspidy Thanks for the comment, that's helpful context. 

So if I'm understanding correctly:

Categories = Hierarchical 
Tags = Not hierarchical
Custom Taxonomy = Could, or could not be, but we expect custom taxonomies that are hierarchical to be supported, and looks like based on the documentation that they should have `hierarchical = true || false` as a property from the API ([here](https://developer.wordpress.org/reference/functions/get_taxonomies/))

Is this correct?

swissspidy:
Yes, precisely. This way we can build this in a more taxonomy-agnostic way. Apologies for not making this clear from the beginning.

Our pod is happy to provide more details about taxonomies and how they work. Feel free to reach out.

The `getTaxonomies` callback in `wp-story-editor` and in general the taxonomy handling there and in `story-editor` is a good way to dive into it.

Example response from the taxonomies endpoint: https://stories-qa-wordpress-amp.pantheonsite.io/wp-json/web-stories/v1/taxonomies?type=web-story

The interesting part there (and in `getTaxonomies`) is the usage of `_links['wp:items']`, through which you'll get the `/wp-json/web-stories/v1/web_story_category` API URL.

After that it's gonna be very similar to what you already have here. Just with an array of taxonomies instead of just one.

Hope that makes sense :)

sblinde:
Thanks @swissspidy! Appreciate the feedback and knowledge!

sblinde:
Converting this PR to a draft status as we continue to work on it, not ready for review at this time but want to save the comments left because they're helpful as we continue to make commits.

includes/Admin/Dashboard.php
```diff
@@ -433,6 +433,7 @@ public function get_dashboard_settings(): array {
 				'settings'       => '/web-stories/v1/settings/',
 				'pages'          => '/wp/v2/pages/',
 				'publisherLogos' => '/web-stories/v1/publisher-logos/',
+				'taxonomies'     => '/web-stories/v1/taxonomies/',
```
spacedmonkey, line 460: Let's also preload this api, by adding to this area. 
https://github.com/GoogleForCreators/web-stories-wp/blob/3d0a8a1db3c3e4b6120219b183c73ab9d744885f/includes/Admin/Dashboard.php#L302

PR #11527, opened 2022-05-20
Editor: Freeze story data

### LINKED ISSUE #11363: Saving story output unexpectedly modifies story data
## Bug Description

I just noticed this while debugging something related to a [support topic](https://wordpress.org/support/topic/h1-tags-not-working-showing/):

When you save a story, it unexpectedly modifies element objects since they are passed by reference.

`OutputPage` directly modifies `page.elements` here

https://github.com/GoogleForCreators/web-stories-wp/blob/6d32ea3cd648abcc262aa1ce26b6f360575bec3b/packages/output/src/page.js#L80-L94

Because of the way `tagName = 'auto'` is retrieved from `element` (which in theory should never exist), this also has the unfortunate side effect of "hardcoding" the tag name in story data. So if an element is saved as an h1 once, it can never be saved as an h2 or so. It will always stay an h1.

This is a simple possible solution to fix this `OutputPage`:

```js
const regularElements = otherElements.map((element) => {
  // Avoid edits to the original element object.
  const newElement = { ...element };
  const { id: elementId, type } = newElement;

  if ('text' === type) {
    newElement.tagName = tagNamesMap.get(elementId);
  }

  // Remove invalid links.
  // TODO: this should come from the pre-publish checklist in the future.
  if (pageAttachment?.url && isElementBelowLimit(element)) {
    delete newElement.link;
  }

  return newElement;
});
```

**Alternatives considered**

* `getStoryPropsToSave` in `packages/output` (or `useAutoSave` and `useSaveStory` in `packages/story-editor`) could use `clone-deep` before passing on the `story` and `pages` data.
* Maybe something else?

## Expected Behaviour

Saving a story should not modify story data, ever.

We should fix this and add some robust JS unit tests (in the output package) and Karma tests to prevent regressions.

We should also add a **data migration** to remove excess `tagName` properties from element objects.

## Steps to Reproduce

There are two easy ways to reproduce this:

**Headings**

1. Open the story dev tools
2. Add a new "Title 1" element
3. See in the story dev tools that it was added
4. Save the story
5. In the dev tools, see that the element now suddenly has a `"tagName": "h1"` property
6. Decrease font size of the element to something super small
7. Preview the story and see that it still is a h1, even if it should not be.

**Links**

1. Add a page attachment
8. Add a new element
9. Add a link to the element
10. Move the element to the lower part of the page (so it triggers a page attachment warning)
11. Save story
12. See that the link was irrevocably removed from the element

## Additional Context

- Plugin Version: 1.20.0-alpha
- WordPress Version: 5.9
- Operating System: macOS
- Browser: Chrome 100

barklund:
## Context

This PR freezes the story data in the story provider. This is done to make sure that no component ever directly manipulates story data, but always only updates it through the reducer.

I did find 3 locations, that violated this principle, so they have all been fixed. But the lack of strict mode means, that we cannot be sure there aren't other attempts at updating the state that fail.

This might mean, that there are things that are silently failing now after introducing this. I did add all types of elements and go through all their options on both the design panel and design menu, but I might still not have caught every possible thing that can happen. And something might even have failed as I went through and I just didn't notice.

@swissspidy, would it be possible to run the editor scripts in strict mode? I tried manually prepending `"use strict"` to the scripts in `assets/js,` but it didn't seem to do anything.

@sblinde, I'm tagging you in here, as this is a potentially huge change in the editor, but hopefully not. I just want a lot of experienced eyes on this.

## To-do

- [x] Freeze story data state
- [x] Fix impure story data update in overlay conversion
- [x] Fix impure story data update in link display
- [x] Fix impure story data update in load story
- [ ] Find any other attempts at updating state and fix them

## User-facing changes

_None_

## Testing Instructions

Even though nothing should be changed by this PR, something might be broken. However, if something is broken, an alert like this should pop up in the editor: Screenshot 2022-05-20 at 10 52 10

1. Please do "everything" in the editor that involves inserting, removing, or updating elements on canvas and see if anything is broken in the editor.

## Checklist

---

Fixes #11363


packages/story-editor/src/app/story/useStoryReducer/useStoryReducer.js
```diff
@@ -82,10 +82,27 @@ function useStoryReducer(partial) {
...
+      o[prop] !== null &&
+      (typeof o[prop] === 'object' || typeof o[prop] === 'function') &&
+      !Object.isFrozen(o[prop])
+    ) {
+      deepFreeze(o[prop]);
+    }
+  });
+
+  return o;
+}
```
barklund, line 106: This is copied in (and slightly modified due to lint rules) from [the `deepFreeze` library](https://github.com/substack/deep-freeze/blob/master/index.js). It seemed like such a small feature that it didn't make a lot of sense to add it as a dependency, but I'm okay with doing that if that's consensus. If we don't do that, this function should be moved to an util somewhere and have some tests added.

packages/story-editor/src/components/panels/design/filter/convertOverlay.js
```diff
@@ -31,9 +31,9 @@ function convertToSolid(currentOverlay, currentType) {
       // converting from any gradient to SOLID
       // use "opaque'st" stop as primary color, but fix alpha at .3
       const { stops } = currentOverlay;
-      const stopsByAlpha = stops.sort(
-        (x, y) => (y.color.a ?? 1) - (x.color.a ?? 1)
-      );
+      const stopsByAlpha = stops
+        .slice()
+        .sort((x, y) => (y.color.a ?? 1) - (x.color.a ?? 1));
```
barklund, line 36: `sort()` sorts inline, so we need slice. It was never really a problem because the old array was immediately de-referenced, but still, it was impure.

packages/output/src/page.js
```diff
@@ -78,19 +78,29 @@ function OutputPage({
...
+    // Invalid links must be removed
     // TODO: this should come from the pre-publish checklist in the future.
-    if (pageAttachment?.url && isElementBelowLimit(element)) {
-      delete element.link;
+    const hasIllegalLink = pageAttachment?.url && isElementBelowLimit(element);
+    const requiresChange = needsTagName || hasIllegalLink;
+
+    // If neither change needed, return original
+    if (!requiresChange) {
+      return element;
```
swissspidy: Would there be harm in always returning `newElement`?

packages/output/src/page.js
```diff
@@ -78,19 +78,29 @@ function OutputPage({
   );
 
   const regularElements = otherElements.map((element) => {
```
swissspidy: Regarding `OutputPage`, I was wondering whether a small regression test for `getStoryMarkup` would be useful, to verify that the passed `pages` array is not being modified in any way.

barklund:
> @barklund I assume you ask because modifying a frozen object throws an error in strict mode. All scripts should already be in strict mode (every webpack module has its own `"use strict";`) and you'll get these errors. So this seems to work as expected.

@swissspidy, ah, that's perfect. I just tested it in the console by copying the reference to one of the objects in the application, and that didn't throw any error, but I guess that's because the console is not in strict mode. That was quite 🤦  of me!

So I think we're good here? Can someone approve or do you want more time to test things out: @swissspidy, @sblinde, @merapi?

packages/output/src/page.js
barklund: I didn't see a reason to make a shallow clone of the object if it wasn't necessary. But no, there's no problem with it.

packages/output/src/page.js
barklund: I felt that freezing the entire object made such a test irrelevant, as it is now impossible to modify any part of the object.

sblinde (dismissed):
👏

Based on my knowledge of the bug, and testing the first scenario listed in the original issue (https://github.com/googleforcreators/web-stories-wp/issues/11363) in the QA environment, I think this PR covers the fix, but agreed that it'd be nice to move into a util in time w/ tests.

packages/output/src/page.js
swissspidy: True, when generating the story output via the story editor when saving a story. But if using the `output` package in some other way then this is not the case

barklund: Very true. I can add a test for that specifically. Freezing the object and testing that it doesn't throw an error would be a decent test, no?

swissspidy: I think so, yeah

merapi (dismissed):
Looks like a good universal fix 👍🏻 

Merged main and resolved conflicts in migrations (review was auto-dismissed by it).

swissspidy (dismissed):
The two reported bugs (links being removed and headings being "hardcoded") are resolved by this

sblinde (dismissed):
Adding my approval back

swissspidy:
Seeing some Karma errors:

>	Unhandled promise rejection: TypeError: Cannot assign to read only property 'colors' of object '#<Object>'

https://github.com/GoogleForCreators/web-stories-wp/runs/6564559817?check_suite_focus=true#step:7:55

merapi (approved):

sblinde (approved):

PR #11552, opened 2022-05-23
Shopping: Allow sorting products 

### LINKED ISSUE #11253: Shopping: Allow sorting products

timarney:
## Context

Add support to sort products by date, title, and price in ascending or descending order.

## Summary

- Adds sort by dropdown to the product sidebar and updates the search queries to allow sorting.

## Testing Instructions

- Keyboard handling for dropdown options doesn't seem to be working.

1. Enable shopping + a shopping vendor
2. Visit the shopping tab
3. Select a sort order from `sort` dropdown

* Note: the dropdown doesn't match the designs (using the default dropdown)
Screen Shot 2022-05-24 at 5 18 34 AM

4. Ensure the products are sorted as per the sort order.

- New query params orderby & order

No

No

## Checklist

---

Fixes #11253


includes/Shopping/Shopify_Query.php
```diff
@@ -297,4 +304,35 @@ public function get_search( string $search_term ) {
...
+	 * @since 1.21.0
+	 *
+	 * @param string $sort_by sort condition for product query.
+	 * @return array
+	 */
+	protected function parse_sort_by( string $sort_by ): array {
+		
+		switch ( $sort_by ) {
+			case 'a-z':
+				$order = 'TITLE|false';
```
spacedmonkey: This should be an array.

spacedmonkey:
```suggestion
				$order = array( 'sort_key' => 'TITLE', 'reverse' => false );
```

includes/Shopping/Shopify_Query.php
```diff
@@ -297,4 +304,35 @@ public function get_search( string $search_term ) {
...
+
+	/**
+	 * Parse sort type into array for product query
+	 *
+	 * @since 1.21.0
+	 *
+	 * @param string $sort_by sort condition for product query.
+	 * @return array
+	 */
+	protected function parse_sort_by( string $sort_by ): array {
```
spacedmonkey: If this is shared logic, we should maybe create a abstance class that both these classes use.

includes/Shopping/Shopify_Query.php
timarney: It gets split into an array at the end ... wasn't sure if repeating array keys for all the cases would be better or not.

includes/Shopping/Shopify_Query.php
timarney: The resulting array keys (sortKey + reverse) and fields(TITLE etc..)  ends up being different (same cases for the switch) ---  but yes as far as the duplication (would be nice to avoid for maintenance)

spacedmonkey: Should be part of the interface?

includes/Shopping/Shopify_Query.php
spacedmonkey: I think it makes it cleaner this way.

timarney: I'll update to just use an array

includes/Shopping/Shopify_Query.php
swissspidy: @timarney @spacedmonkey Instead of passing values like `a-z` to the REST API (and thus to the Product Query), let's pass separate `orderby` and `orderby` values everywhere.

So while you will select "lowest price" in the dropdown, this will send `orderby=price, order=asc` to the REST API

This makes this much simpler and does not couple the PHP logic to the UI changes.

So for example if we ever update the UI to add a separate dropdown for choosing ascending/descending, we do not have to update the PHP code.

swissspidy: Interfaces are for public methods, not protected/private methods that are specific to each implementation.

But I propose not doing this at all, see https://github.com/GoogleForCreators/web-stories-wp/pull/11552/files#r880339767

includes/Interfaces/Product_Query.php
```diff
@@ -39,7 +39,8 @@ interface Product_Query {
 	 * @since 1.21.0
 	 *
 	 * @param string $search_term Search term.
+	 * @param string $sort_by sort order for query.
 	 * @return Product[]|WP_Error
 	 */
-	public function get_search( string $search_term );
+	public function get_search( string $search_term, string $sort_by);
```
swissspidy: As per https://github.com/GoogleForCreators/web-stories-wp/pull/11552/files#r880339767, this becomes much easier with separate `$orderby` and `$order` params

includes/REST_API/Products_Controller.php
```diff
@@ -163,8 +163,15 @@ public function get_items( $request ) {
...
-		$search_term  = ! empty( $request['search'] ) ? $request['search'] : '';
-		$query_result = $query->get_search( $search_term );
+		$search_term = ! empty( $request['search'] ) ? $request['search'] : '';
+		
+		/**
+		 * Request context.
+		 *
+		 * @var string $sort_by
+		 */
+		$sort_by      = ! empty( $request['sort_by'] ) ? $request['sort_by'] : '';
```
swissspidy: As per https://github.com/GoogleForCreators/web-stories-wp/pull/11552/files#r880339767, it becomes much easier with separate `order` and `orderby` params.

Also don't forget to declare new params in `get_collection_params()`

Something like 


```php
/**
 * Retrieves the query params for the products collection.
 *
 * @since 1.21.0
 *
 * @return array Collection parameters.
 */
public function get_collection_params(): array {
	$query_params = parent::get_collection_params();

	$query_params['order'] = [
		'description' => __( 'Order sort attribute ascending or descending.', 'web-stories' ),
		'type' => 'string',
		'default' => 'desc',
		'enum' => ['asc', 'desc'],
	];

	$query_params['orderby'] = [
		'description' => __( 'Sort collection by product attribute.', 'web-stories' ),
		'type' => 'string',
		'default' => 'date',
		'enum' => [
			'date',
			'price',
			'title',
		],
	];

	return $query_params;
}
```

includes/Shopping/Shopify_Query.php
timarney: Makes sense.

includes/REST_API/Products_Controller.php
spacedmonkey: Massive plus one here. Looking at this, not is the cleaner to understand, but is inline with other endpoints.

includes/Shopping/Shopify_Query.php
timarney: Pushed up refactored code with orderby and order as separate params.

includes/REST_API/Products_Controller.php
timarney: Added.

tests/phpunit/integration/tests/Shopping/Shopify_Query.php
```diff
@@ -288,4 +290,46 @@ public function test_get_search_empty_search_response(): void {
...
 	}
+
+	/**
+	 * @covers ::fetch_remote_products
+	 * @covers ::get_search
+	 * @covers ::get_products_query
+	 * @covers ::execute_query
+	 * @covers ::parse_sort_by
+	 */
+	public function test_get_search_sort_by_query(): void {
```
swissspidy: Just a quick observation: this looks like a great opportunity to use a data provider to test all these combinations.

tests/phpunit/integration/tests/Shopping/Shopify_Query.php
```diff
@@ -288,4 +290,46 @@ public function test_get_search_empty_search_response(): void {
...
+		$this->assertNotWPError( $actual );
+		$this->assertStringContainsString( 'sortKey: PRICE', $this->request_body );
+		$this->assertStringContainsString( 'reverse: false', $this->request_body );
+		
+		$actual = $this->instance->get_search( 'some search term', 'price', 'desc' );
+		$this->assertNotWPError( $actual );
+		$this->assertStringContainsString( 'sortKey: PRICE', $this->request_body );
+		$this->assertStringContainsString( 'reverse: true', $this->request_body );
+
+		remove_filter( 'pre_http_request', [ $this, 'mock_response_no_results' ] );
```
swissspidy, line 335: I'd recommend removing the filter before doing any assertions.

This way, if any assertions fail, the filter is still cleaned up. Otherwise it won't happen.

For this particular test this will be easier to do when using a data provider.

includes/Shopping/Shopify_Query.php
```diff
@@ -157,8 +157,8 @@ protected function execute_query( string $query ) {
 	 */
 	protected function get_products_query( string $search_term, string $orderby, string $order ): string {
 		$search_string = empty( $search_term ) ? '*' : '*' . $search_term . '*';
-		$orderby       = $orderby === 'date' ? 'CREATED_AT' : strtoupper( $orderby );
-		$order         = $order === 'asc' ? 'false' : 'true';
+		$orderby       = 'date' === $orderby ? 'CREATED_AT' : strtoupper( $orderby );
+		$order         = 'asc' === $order ? 'false' : 'true';
```
swissspidy: Maybe name this `$reverse` for clarity?

includes/Shopping/Shopify_Query.php
```diff
@@ -151,16 +151,17 @@ protected function execute_query( string $query ) {
...
+	 * @param string $order Order sort attribute ascending or descending.
 	 * @return string The assembled GraphQL query.
 	 */
-	protected function get_products_query( string $search_term ): string {
+	protected function get_products_query( string $search_term, string $orderby, string $order ): string {
 		$search_string = empty( $search_term ) ? '*' : '*' . $search_term . '*';
-
-		// TODO(#11154): Support different sortKeys.
-		// Maybe use "available_for_sale:true AND " query to only show items in stock.
+		$orderby       = 'date' === $orderby ? 'CREATED_AT' : strtoupper( $orderby );
```
swissspidy: Maybe rename to `$sortkey` for clarity?

includes/Shopping/Shopify_Query.php
```diff
@@ -198,9 +199,12 @@ protected function get_products_query( string $search_term ): string {
 	 * @since 1.21.0
 	 *
 	 * @param string $search_term Search term to filter products by.
+	 * @param string $orderby sort field for query.
+	 * @param string $order ASC or DESC.
 	 * @return array|WP_Error Response data or error object on failure.
 	 */
-	protected function fetch_remote_products( string $search_term ) {
+	protected function fetch_remote_products( string $search_term, string $orderby = '', string $order = '' ) {
```
swissspidy:
```suggestion
	protected function fetch_remote_products( string $search_term, string $orderby, string $order ) {
```

No need for default values here since `get_search()` below always passes something

includes/Shopping/Shopify_Query.php
```diff
@@ -242,17 +247,21 @@ protected function fetch_remote_products( string $search_term ) {
...
 
+	
 	/**
 	 * Get products by search term.
 	 *
 	 * @since 1.21.0
 	 *
 	 * @param string $search_term Search term.
+	 * @param string $orderby sort field for query.
+	 * @param string $order ASC or DESC.
```
swissspidy: Nit: It's lowercase asc and desc since that's what `get_products_query` uses for comparison.

```suggestion
	 * @param string $orderby Sort retrieved products by parameter. Default 'date'.
	 * @param string $order   Whether to order products in ascending or descending order.
	 *                        Accepts 'asc' (ascending) or 'desc' (descending). Default 'desc'.
```

includes/Shopping/Shopify_Query.php
```diff
@@ -198,9 +199,12 @@ protected function get_products_query( string $search_term ): string {
 	 * @since 1.21.0
 	 *
 	 * @param string $search_term Search term to filter products by.
+	 * @param string $orderby sort field for query.
+	 * @param string $order ASC or DESC.
```
swissspidy:
```suggestion
	 * @param string $orderby Sort retrieved products by parameter.
	 * @param string $order   Whether to order products in ascending or descending order.
	 *                        Accepts 'asc' (ascending) or 'desc' (descending).
```

packages/story-editor/src/components/library/panes/shopping/productSort.js
```diff
@@ -0,0 +1,85 @@
...
+const StyledContainer = styled.div`
+  margin-left: 12px;
+`;
+
+function SortDropdown({ onChange, sortId }) {
+  const isShoppingIntegrationEnabled = useFeature('shoppingIntegration');
+
+  if (!isShoppingIntegrationEnabled) {
+    return null;
+  }
```
swissspidy: This is not needed because the `ShoppingPane` component already checks for this

packages/story-editor/src/components/library/panes/shopping/productSort.js
```diff
@@ -0,0 +1,85 @@
...
+    <StyledContainer>
+      <Datalist.DropDown
+        dropDownLabel={__('Sort', 'web-stories')}
+        onChange={onChange}
+        selectedId={sortId}
+        options={options.map((option) => ({
+          id: `${option.orderby}-${option.order}`,
+          ...option,
+        }))}
+        aria-label={__('Sort by', 'web-stories')}
```
swissspidy:
```suggestion
        aria-label={__('Product sort options', 'web-stories')}
```

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -68,6 +69,12 @@ function ShoppingPane(props) {
   const [loaded, setLoaded] = useState(false);
   const [isLoading, setIsLoading] = useState(false);
   const [searchTerm, setSearchTerm] = useState('');
+  const [orderby, setOrderby] = useState('date');
+  const [order, setOrder] = useState('desc');
+  const onSortBy = (option) => {
+    setOrderby(option?.orderby);
+    setOrder(option?.order);
```
swissspidy: The use of optional chaining here indicates that `option` can be `null`. Is this ever the case?

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -212,6 +219,7 @@ function ShoppingPane(props) {
             clearId="clear-product-search"
             handleClearInput={handleClearInput}
           />
+          <ProductSort onChange={onSortBy} sortId={`${orderby}-${order}`} />
```
swissspidy: Not related to this specific line but let's expand the existing Karma tests to also cover product sorting

swissspidy (commented):
Just a few nitpicks here and there, overall looks good.

Karma test coverage would be nice.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
timarney: I'll remove --- the only case would be an input error in the source code.

includes/REST_API/Products_Controller.php
```diff
@@ -163,8 +163,24 @@ public function get_items( $request ) {
...
-		$search_term  = ! empty( $request['search'] ) ? $request['search'] : '';
-		$query_result = $query->get_search( $search_term );
+		$search_term = ! empty( $request['search'] ) ? $request['search'] : '';
+		
+		/**
+		 * Request context.
+		 *
+		 * @var string $orderby
+		 */
+		$orderby = ! empty( $request['orderby'] ) ? $request['orderby'] : '';
```
spacedmonkey:
```suggestion
		$orderby = ! empty( $request['orderby'] ) ? $request['orderby'] : 'date';
```

includes/REST_API/Products_Controller.php
```diff
@@ -163,8 +163,24 @@ public function get_items( $request ) {
...
+
+		/**
+		 * Request context.
+		 *
+		 * @var string $order
+		 */
+		$order = ! empty( $request['order'] ) ? $request['order'] : '';
```
spacedmonkey:
```suggestion
		$order = ! empty( $request['order'] ) ? $request['order'] : 'desc';
```

tests/phpunit/integration/includes/Shopping/Mock_Vendor_Error.php
```diff
@@ -32,9 +32,11 @@ class Mock_Vendor_Error implements Product_Query {
 	 * @since 1.21.0
 	 *
 	 * @param string $search_term Search term.
+	 * @param string $orderby Sort collection by product attribute.
+	 * @param string $order Order sort attribute ascending or descending.
 	 * @return Product[]|WP_Error
 	 */
-	public function get_search( string $search_term ) {
+	public function get_search( string $search_term, string $orderby, $order ) {
```
spacedmonkey:
```suggestion
	public function get_search( string $search_term, string $orderby, string $order ) {
```

tests/phpunit/integration/tests/Shopping/Shopify_Query.php
timarney: Okay

I pushed up data provider --- to iterate on (as I don't think I have it setup properly) i.e. passing args[0], args[1] etc..

What's the correct setup for a test like this? 


https://github.com/GoogleForCreators/web-stories-wp/pull/11552/commits/b59acac94159affd39ed4865b7cde281bf730449#diff-93f38efcb6c6fff109a2e65269ca80cc58d42daa995500825396f0544be3f83cR334

includes/Shopping/Shopify_Query.php
```diff
@@ -242,17 +248,21 @@ protected function fetch_remote_products( string $search_term ) {
...
 	 * @param string $search_term Search term.
+	 * @param string $orderby sort field for query.
+	 * @param string $order ASC or DESC.
 	 * @return Product[]|WP_Error
 	 */
-	public function get_search( string $search_term ) {
-		$result = $this->fetch_remote_products( $search_term );
-
+	public function get_search( string $search_term, string $orderby = 'date', string $order = 'desc' ) {
+
```
spacedmonkey:
```suggestion
```

includes/Shopping/Shopify_Query.php
```diff
@@ -198,9 +199,13 @@ protected function get_products_query( string $search_term ): string {
 	 * @since 1.21.0
 	 *
 	 * @param string $search_term Search term to filter products by.
+	 * @param string $orderby Sort retrieved products by parameter.
+	 * @param string $order   Whether to order products in ascending or descending order.
+     *                        Accepts 'asc' (ascending) or 'desc' (descending).
```
swissspidy:
```suggestion
	 *                        Accepts 'asc' (ascending) or 'desc' (descending).
```

My fault for mixing tabs and spaces in my previous suggestion

includes/Shopping/WooCommerce_Query.php
```diff
@@ -76,12 +78,17 @@ public function get_search( string $search_term ) {
...
-				'order'   => 'DESC',
 				's'       => $search_term,
+				'orderby' => $orderby,
+				'order'   => $order,
 			]
 		);
 
+		if ( 'price' === $orderby ) {
+			// @todo this only orders products based on wc_get_products previously called
+			$products = wc_products_array_orderby( $products, 'price', $order );
```
spacedmonkey: @swissspidy What if there are more than 100 products that meet the search term. Like you have 1000 products with the name t-shirt. Wouldn't this only sort the top 100?

spacedmonkey (dismissed):
I have some questions about how pricing sorting works. I would also like to understand the plan for pagination as well. 

Also a little confused by the PR doesn't match the figma. 

There are also lints errors that need fixing.

includes/Shopping/WooCommerce_Query.php
swissspidy: Uhm, yeah, I guess. Seems like that's the only way WooCommerce provides out of the box.

I think on Monday we discussed using meta queries as a workaround

Does something like https://github.com/woocommerce/woocommerce/wiki/wc_get_products-and-WC_Product_Query#adding-custom-parameter-support work? In combination with `'orderby' => 'meta_value',`

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
timarney: Added a test here - https://github.com/GoogleForCreators/web-stories-wp/pull/11552/commits/8008da617a5586b831239e437fdb4b1217639656#diff-fc3f39bc6c4a8d0d2df87c0807cb1ac0c536bc846029b6a67c42a2e9ba74f84bR125

includes/Shopping/WooCommerce_Query.php
timarney: @spacedmonkey your correct per the todo note 

> this only orders products based on wc_get_products previously called


Looks like maybe we can do something with https://woocommerce.wp-a2z.org/oik_api/get_catalog_ordering_args

```php
case 'price':
        $callback = 'DESC' === $order ? 'order_by_price_desc_post_clauses' : 'order_by_price_asc_post_clauses';
        [add_filter](https://core.wp-a2z.org/oik_api/add_filter)( ['posts_clauses'](https://woocommerce.wp-a2z.org/oik_hook/posts_clauses/), array( [$this](https://woocommerce.wp-a2z.org/oik_api/$this::WC_Query), [$callback](https://woocommerce.wp-a2z.org/oik_api/$this::WC_Query) ) );
        break;
```

Also found 

https://stackoverflow.com/questions/47056744/woocommerce-order-by-price-not-working-correctly

```php
/**
 * Customize ordering by price
 */
add_filter('woocommerce_get_catalog_ordering_args', function ($args) {
    $orderby_value = isset($_GET['orderby']) ? wc_clean($_GET['orderby']) : apply_filters('woocommerce_default_catalog_orderby', get_option('woocommerce_default_catalog_orderby'));

    if ('price' == $orderby_value) {
        $args['orderby'] = 'meta_value_num';
        $args['order'] = 'ASC';
        $args['meta_key'] = '_price';
    }

    if ('price-desc' == $orderby_value) {
        $args['orderby'] = 'meta_value_num';
        $args['order'] = 'DESC';
        $args['meta_key'] = '_price';
    }

    return $args;
});

```

tests/phpunit/integration/tests/Shopping/Shopify_Query.php
timarney: @spacedmonkey fixed here - https://github.com/GoogleForCreators/web-stories-wp/pull/11575/files#diff-93f38efcb6c6fff109a2e65269ca80cc58d42daa995500825396f0544be3f83cR335

includes/Shopping/WooCommerce_Query.php
timarney: Updated with this commit --- seems to work 

https://github.com/GoogleForCreators/web-stories-wp/pull/11552/commits/d8fba81ed96fd3525cbfc5c74ef164ff7679c613

spacedmonkey:
Screenshot 2022-05-27 at 11 57 36

I am not seeing the price sorting working here.

timarney:
> I am not seeing the price sorting working here.

@spacedmonkey re: https://github.com/GoogleForCreators/web-stories-wp/pull/11552#issuecomment-1139511799

Locally see video seems to be working ...

Curious if you manually add a meta_key for price do you see "proper" results?

```php
$wc_args  = $wc_query->get_catalog_ordering_args( $orderby, strtoupper( $order ) );
$wc_args['meta_key'] = '_price'; // try adding this
```

My local without manually setting a meta key
https://user-images.githubusercontent.com/62242/170830920-5f8d9aaa-5cab-4b8a-9ba8-702c73d5b9cb.mp4

spacedmonkey:
@timarney Still not seeing it working. 
Screenshot 2022-05-30 at 10 37 58

swissspidy:
@spacedmonkey I think you need to regenerate product lookup table data after importing those sample products.

Go to WooCommerce > Status > Tools and click on "Product lookup tables"

After that the sorting worked fine for me.

swissspidy (dismissed):

swissspidy (dismissed):

includes/Shopping/Shopify_Query.php
```diff
@@ -151,16 +151,17 @@ protected function execute_query( string $query ) {
...
 		$search_string = empty( $search_term ) ? '*' : '*' . $search_term . '*';
-
-		// TODO(#11154): Support different sortKeys.
-		// Maybe use "available_for_sale:true AND " query to only show items in stock.
+		$sortkey       = 'date' === $orderby ? 'CREATED_AT' : strtoupper( $orderby );
+		$reverse       = 'asc' === $order ? 'false' : 'true';
 		return <<<QUERY
 {
-  products(first: 100, sortKey: CREATED_AT, query: "title:$search_string") {
+  products(first: 100, sortKey: $sortkey, reverse: $reverse, query: "title:$search_string") {
```
spacedmonkey, line 164: I am not sure this is correct. Instead of using `reverse`, shouldn't this is last?

https://shopify.dev/api/usage/pagination-graphql#backward-pagination

timarney: I think `reverse` is the right property to use for `ASC` / `DESC` type of queries
see - https://youtu.be/S37WsC8GzSA?t=352

swissspidy: @spacedmonkey That documentation is for pagination, while this PR is for sorting.

swissspidy:
@timarney FYI there's now a merge conflict after fixing the Karma tests in the other PR.

swissspidy (approved):

spacedmonkey (approved):

spacedmonkey:
Tested after calling product lookup index. Seems to work well.

PR #11575, opened 2022-05-26
Shopping: Support pagination in API 

### LINKED ISSUE #11592: Shopping: Support pagination in API
## Feature Description

Related: #11575

Since WooCommerce and Shopify use different means of pagination, we'd need to support both in our `Product_Query` interface.

**WooCommerce**

Supports `posts_per_page`/`limit` and `page` parameters. Also a `paginate` param.

See https://github.com/woocommerce/woocommerce/wiki/wc_get_products-and-WC_Product_Query

**Shopify**

Shopify uses cursor-based pagination, see https://shopify.dev/api/usage/pagination-graphql for all the details

We'd need to expose `hasNextPage` and `endCursor` from the `pageInfo` field in the API response. The `endCursor` then is what we use for `after` to fetch 2nd, 3rd, etc. pages.

### LINKED ISSUE #11593: Shopping: Infinite Scroll Pagination
## Feature Description

Adding pagination support on the API side is being handled in #11592

This ticket is about implementing pagination for the shopping tab in the editor using infinite scroll.

We already have a few different infinite scroll implementations in the editor and the dashboard, so ideally we can reuse an existing one.

While this is somewhat blocked by #11592, I think some parts can be worked on already now, and tested with Karma tests (see also #11314 #11590)

## Developer Notes

- Enable WooCommerce Plugin

spacedmonkey:
## Summary

Respect per page param for products api. Adding this later might be harder, so let's do this now. 

## Testing Instructions

1.

## Checklist

---

Fixes #11592


timarney:
Overall I like the idea but wondering if without pagination if users will understand the query uses the per page param.

swissspidy:
Let's implement this by also supporting offsets at the same time, since both require changes to method signatures and stuff. I will create a separate ticket for tackling this, with details. Let's hold off on this until then.

swissspidy:
See #11593 (adding pagination support in the API) and #11592 (adding pagination to the UI). Let's implement this there.

includes/Shopping/Shopify_Query.php
```diff
@@ -214,55 +225,103 @@ protected function fetch_remote_products( string $search_term, string $orderby,
...
+				 * @var ShopifyGraphQLResponse $result
+				 */
+				$result        = json_decode( $body, true );
+				$has_next_page = $result['data']['products']['pageInfo']['hasNextPage'];
+				if ( ! $has_next_page ) {
+					return new WP_Error( 'rest_no_page', __( 'Error fetching products', 'web-stories' ), [ 'status' => 404 ] );
+				}
+				$after = (string) $result['data']['products']['pageInfo']['endCursor'];
+			}
+		}
```
swissspidy: This causes quite a lot of requests to the Shopify API.

In the editor, when paginating through the products responses, for every page N it would make N API requests to Shopify. Example: when you reach page 10 in the editor, it will call `/web-stories/v1/products?page=10&per_page=20`, which then leads to `execute_query()` being called 10 times here.

1. Page 1: 1 request
2. Page 2: 2 requests
3. Page 3: 3 requests
4. and so on

When reaching page 10, there will have been 55 API requests in a very short time. This way users will very quickly run into [API rate limits](https://shopify.dev/api/usage/rate-limits#time-based-limits).

What would it take for our API to support cursor-based pagination as well?

Then, when paginating in the editor we would just call `/web-stories/v1/products?per_page=20&after=abc`, `/web-stories/v1/products?per_page=20&after=def`, etc. This will result in way fewer API requests.

For WooCommerce, nothing would change.

spacedmonkey: I think that supporting cursor based pagination breaks the genericness we have at the moment. I think we need talk about this change, as it is fundamental.

swissspidy: Yeah it would definitely mean updating the current genericness to account for both types of pagination.

I'd imagine the editor would make calls like `/web-stories/v1/products?per_page=20&page=2&after=abc` and we'd pass both `$page` and `$after` to the `Product_Query`.

WooCommerce would just take `$page` and ignore `$after`, and Shopify would ignore `$page` instead and use `$after`.

That's why I am curious to see how much work that would be.

----

Regardless, it looks like `get_search()` needs to return not just an array of `Product`s, but also pagination information (`hasNextPage` and `endCursor` for Shopify`, `max_num_pages`, `total` for WooCommerce)

Then we need to add this information as REST API response headers (e.g. `X-WP-Total`, `X-WP-TotalPages`, `X-WP-EndCursor`)

WooCommerce supports this with the `'paginate' => true` flag, see https://github.com/woocommerce/woocommerce/wiki/wc_get_products-and-WC_Product_Query

```php
// Get products with extra info about the results.
$args = array(
    'paginate' => true,
);
$results = wc_get_products( $args );
echo $results->total . ' products found\n';
echo 'Page 1 of ' . $results->max_num_pages . '\n';
echo 'First product id is: ' . $results->products[0]->get_id() . '\n';
```

spacedmonkey: I have refactored to move the caching. I think we don't need to cursor. The current implemention, means when you hit, page 1 and the page 2, it will reuse page 1, meaning no more requests.

swissspidy: Okay, I'll take a look at it.

The next problem is that in the editor we still need to know if there are more pages to fetch. Which means we still need to send some API response headers so that we can then implement pagination in the editor.

includes/REST_API/Products_Controller.php
```diff
@@ -164,34 +166,56 @@ public function get_items( $request ) {
...
 
-		return rest_ensure_response( $products );
+		/**
+		 * Response object.
+		 *
+		 * @var WP_REST_Response $response
+		 */
+		$response = rest_ensure_response( $products );
+
+		$response->header( 'X-WP-HAS-NEXT-PAGE', (string) $query_result['has_next_page'] );
```
swissspidy:
```suggestion
		$response->header( 'X-WP-HasNextPage', (string) $query_result['has_next_page'] );
```

This is more in line with other headers (e.g. `X-WP-TotalPages`, `X-WP-Upload-Attachment-ID`, `X-WP-Nonce`)

tests/phpunit/integration/includes/Shopping/Mock_Vendor.php
```diff
@@ -52,7 +54,7 @@ public function get_search( string $search_term, string $orderby, string $order
 				]
 			);
 		}
-
-		return $products;
+		$has_next_page = 1;
```
swissspidy: Shouldn't that be a boolean `true`?

Ditto for `Mock_Vendor_Invalid`

includes/Shopping/Shopify_Query.php
```diff
@@ -151,17 +151,22 @@ protected function execute_query( string $query ) {
...
 		$search_string = empty( $search_term ) ? '*' : '*' . $search_term . '*';
 		$sortkey       = 'date' === $orderby ? 'CREATED_AT' : strtoupper( $orderby );
 		$reverse       = 'asc' === $order ? 'false' : 'true';
+		$after         = empty( $after ) ? '' : $after;
+		$after_query   = empty( $after ) ? '' : " after: \"$after\",";
+
 		return <<<QUERY
 {
-  products(first: 100, sortKey: $sortkey, reverse: $reverse, query: "title:$search_string") {
+  products(first: $per_page, $after_query sortKey: $sortkey, reverse: $reverse, query: "title:$search_string") {
```
swissspidy: Adding `$after_query` like this is a bit error prone, hard to read and not in line with how the other args are added.

This should work as well:

```suggestion
  products(first: $per_page, after: $after, sortKey: $sortkey, reverse: $reverse, query: "title:$search_string") {
```

And above:

```php
$after         = empty( $after ) ? null : $after;
```

includes/Shopping/Shopify_Query.php
```diff
@@ -151,17 +151,22 @@ protected function execute_query( string $query ) {
 	 * @since 1.21.0
 	 *
 	 * @param string $search_term Search term to filter products by.
+	 * @param string $after   After node.
+	 * @param int    $per_page   Limit query.
 	 * @param string $orderby Sort collection by product attribute.
 	 * @param string $order Order sort attribute ascending or descending.
 	 * @return string The assembled GraphQL query.
```
swissspidy: Nit: indentation

swissspidy:
I think we're on to something here! Nice work so far 🚀

includes/Shopping/Shopify_Query.php
spacedmonkey: No, I tried this and gave me a graphQL error.

swissspidy: I tried this in the GraphiQL explorer at https://shopify.dev/custom-storefronts/tools/graphiql-storefront-api and it does definitely work. Can you please test this again?

Proof:

Screenshot 2022-06-01 at 00 18 40

spacedmonkey: Passing after an string of null worked. Empty string did not work.

includes/Shopping/Shopify_Query.php
```diff
@@ -270,26 +277,63 @@ protected function fetch_remote_products( string $search_term, string $orderby,
...
+	 * @param int    $per_page    Number of products to be fetched.
+	 * @param string $orderby     Sort retrieved products by parameter.
+	 * @param string $order       Whether to order products in ascending or descending order.
+	 *                            Accepts 'asc' (ascending) or 'desc' (descending).
+	 * @return array|WP_Error Response data or error object on failure.
+	 */
+	protected function fetch_remote_products( string $search_term, int $page, int $per_page, string $orderby, string $order ) {
+		$after = '';
+		if ( $page > 1 ) {
+			for ( $i = 1; $i < $page; $i ++ ) {
```
timarney, line 313: Can you add a note about what this loop is doing?

swissspidy:
Note that there are some lint errors here still.

includes/Shopping/Shopify_Query.php
```diff
@@ -270,26 +276,64 @@ protected function fetch_remote_products( string $search_term, string $orderby,
...
+				if ( ! $has_next_page ) {
+					return new WP_Error( 'rest_no_page', __( 'Error fetching products from Shopify.', 'web-stories' ), [ 'status' => 404 ] );
+				}
+				$after = (string) $result['data']['products']['pageInfo']['endCursor'];
+			}
+		}
+
+		$result = $this->get_remote_products( $search_term, $after, $per_page, $orderby, $order );
+
+		return $result;
```
spacedmonkey:
```suggestion
		return $this->get_remote_products( $search_term, $after, $per_page, $orderby, $order );
```

includes/Shopping/Shopify_Query.php
```diff
@@ -270,26 +275,80 @@ protected function fetch_remote_products( string $search_term, string $orderby,
 		return $result;
 	}
 
-	
+	/**
+	 * Get cache key for properties.
+	 *
+	 * @since 1.21.0
```
swissspidy: Nit: we're now in the 1.22.0 cycle

spacedmonkey:
```suggestion
	 * @since 1.22.0
```

includes/Shopping/Shopify_Query.php
```diff
@@ -270,26 +275,80 @@ protected function fetch_remote_products( string $search_term, string $orderby,
...
+	protected function get_cache_key( $search_term, $after, $per_page, $orderby, $order ): string {
+		$cache_args = (string) wp_json_encode( compact( 'search_term', 'after', 'per_page', 'orderby', 'order' ) );
+
+		return 'web_stories_shopify_data_' . md5( $cache_args );
+	}
+
+	/**
+	 * Remotely fetches all products from the store.
+	 *
+	 * @since 1.21.0
```
spacedmonkey:
```suggestion
	 * @since 1.22.0
```

includes/Shopping/Shopify_Query.php
spacedmonkey: Fixed.

swissspidy (approved):

PR #11577, opened 2022-05-26
Shopping: Prevent duplicate products on a page 

### LINKED ISSUE #11521: Shopping: Prevent duplicate products on a page
## Feature Description

Right now the idea is that a single product can only appear on a page once.

Currently, it's possible to end up with multiple products by duplicating or pasting elements.

How can we avoid that?

Possible options:

* Do not add element and display a snackbar message
* Add the element but without the product attached

Not sure where's the best place to do this.

Relevant code:

* `useCanvasCopyPaste`: handles keyboard cop/paste
* `useCanvasKeys`: handles element duplication via `mod+d`.
* `duplicateElementsById` reducer
* `useAddPastedElements` hook

spacedmonkey:
## Context

It is makes for invalid story by having duplicate products on the same page. 

## Relevant Technical Choices

Tweak the reducer to not allow for new element to be created / duplicate with the same id. 

## Testing Instructions

1. Create a story. 
2. Add a product. 
3. Select product element. 
4. Copy element cmd + c
5. Deselect product
6. paste element cmd + v. 
7. Product element should **not** be duplicated. 

Right click should not be able to duplicate product.
Layer panel  should not be able to duplicate product.

## Checklist

---

Fixes #11521


spacedmonkey:
Love some feedback from @barklund and @merapi as if this is the best way to do this.

timarney:
Gave this a quick test --- seems to work well.

merapi:
Is there anything extra I need to do to enable shopping? I did it a while ago like this: install WooCommerce, add product, enable the experiment - it worked, for the frontend I needed one additional step, but this time I'm getting 
 /v1/products 400/500 in the editor.

swissspidy:
@merapi You need to select WooCommerce on the Web Stories settings page.

spacedmonkey:
@timarney Why are they your changes in this PR?

merapi (dismissed):
LG ✅ 

We should also remove the "Duplicate Layer" icon from the Layers Panel (not sure if here).

image

spacedmonkey:
@merapi I have made a tweak to hide this button. Can you take a look.

merapi (approved):
Still LG ✅ 😄

spacedmonkey:
Awaiting a review by @swissspidy

timarney (approved):

packages/story-editor/src/app/story/useStoryReducer/reducers/duplicateElementsById.js
```diff
@@ -55,6 +59,13 @@ function duplicateElementsById(state, { elementIds }) {
       return;
     }
 
+    if (
+      elementToDuplicate.type === 'product' &&
+      currentPageProductIds.includes(elementToDuplicate?.product?.productId)
```
swissspidy: This check is not needed, no?

Because this is for duplicating an existing product, this product will **always** be in `currentPageProductIds`

So we don't need `currentPageProductIds` here at all.

swissspidy (commented):
#11521 mentioned displaying a snackbar message for duplicate products. I added this now in #11649, targeting this branch here.

After that, this should be good to go.

packages/story-editor/src/app/story/useStoryReducer/reducers/duplicateElementsById.js
spacedmonkey: Removed.

spacedmonkey:
I have also fixed a couple of places that had hard coded references to product element type.

swissspidy (approved):

packages/story-editor/src/app/canvas/useCanvasKeys.js
```diff
@@ -227,10 +238,32 @@ function useCanvasKeys(ref) {
...
+        product?.productId &&
+        currentPageProductIds.includes(product.productId)
+      ) {
+        showSnackbar({
+          message: sprintf(
+            /* translators: %s: product title. */
+            __('Product "%s" already exists on the page.', 'web-stories'),
+            product.productTitle
+          ),
+          timeout: 1000,
```
swissspidy: Whoops, I meant to remove that in my other PR.

```suggestion
```

PR #11578, opened 2022-05-26
Editor: Correctly initialize new story publish date to "immediately"

### LINKED ISSUE #11403: Stories are not setting their publish date correctly.
## Bug Description

In testing we've discovered that stories that have previously been saved as a draft, will retain this date and time as the published date, no matter when the story is actually published. This is inaccurate and is causing problems for clients who may be editing stories for extended period of time. 

## Expected Behaviour

The publish date time should reflect the actual date & time when the story was succesfully published, not when the story first began life as a draft.

example URL - https://stories-new-wordpress-amp.pantheonsite.io/web-stories/teststory4/ 

## Steps to Reproduce

1. Create a story
2. Wait a day
3. Publish the story
4. note that the publish date / time is the day previous

## Screenshots

1. Unpublished draft, view of the document publish time, which is incorrectly indicated a date in the past:
Screen Shot 2022-04-29 at 3 08 47 PM
2 Same story - prepublish modal also showing the publish date/time set to when the story was first saved as a draft 

Screen Shot 2022-04-29 at 3 09 09 PM

3. Post publish - Structured data check reveals the old date is retained
Screen Shot 2022-04-29 at 3 09 55 PM
4. Post publish - Published story in dashboard with old date displayed as the published date/time - 
Screen Shot 2022-04-29 at 3 10 10 PM

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

barklund:
## Context

This fixes a bug where you create a new story, save it and then reload it, and it will then have a story publish date identical to the original creation time.

## User-facing changes

A new story will have publish-time set to "Immediately" until published or explicitly set to a specific date and time.

### New empty story
1. Create a new empty story
2. Add an element to the page
3. Save the story
4. Reload the entire browser window
5. Open the document panel
6. Observe that the publish time and date is set to "immediately"

### New story from template
1. Create a new empty story from an existing template
2. Move or edit any element on a page
3. Save the story
4. Reload the entire browser window
5. Open the document panel
6. Observe that the publish time and date is set to "immediately"

## Checklist

---

Fixes #11403


packages/story-editor/src/app/story/effects/useLoadStory.js
```diff
@@ -56,6 +56,8 @@ function loadStory(storyId, post, restore, clearHistory) {
     terms,
   } = post;
 
+  const date = origDate === modified ? null : origDate;
```
barklund: @swissspidy, the issue here is, that when I go to create a new story, the already preloaded story data (loaded in from the cached REST response in the document) has a fixed date of "now" rather than no date at all (`null`). This is just an auto-draft created of the new empty story, but it still contains initial data including a publish date.

The above code assumes that only such a brand new story auto-draft has `date` and `modified` set to the same value and assumes the publish date to be `null` in such a case and this works. The application will then save `null` and `null` will correctly be loaded back in once the application reloads.

A better solution here would be for the original auto-draft to simply have a publish date of `null` in the first place, but I'm not sure where to fix that.

If that is not achievable, we might still want to look for a better way to test for "brand new empty story" than the above check. Technically a just published story might have modified and publish dates set to the same timestamp (though that doesn't appear to happen under normal circumstances), and this snippet above would then unset the publish date.

So, my thoughts for the proper fix in priority:

1. Make sure an initial auto-draft of a new empty story has `date` set to `null` rather than current timestamp,
2. If that doesn't work, replace the above check with a foolproof way to determine if a story is brand new,
3. If that is not possible, go with what I have above.

Remember that all of the above should also work for a new story created from a template. My fix above does do that, but we must make sure that any alternative (and better) fix also does that.

swissspidy: Thanks for digging into this. Not sure if there are other implications when setting the `date` to `null` in this case. But I gave it a try in #11598 if you wanna test it.

IMHO it does paper over the underlying issue that we send the `date` field when saving a story, even though we haven't touched it (same goes for all other fields of the story). Compare this to Gutenberg where only the actually changed fields are sent to the REST API, not all fields. Which is why it does not suffer this same problem.

This now just works because we set the date to `null` for drafts and then send `'date' => null` when saving a story, which then means it won't get updated in the db.

swissspidy: So #11598 only works if also doing it for `draft` and not just `auto-draft`, but that seems to break the list view on the dashboard, which expects proper dates.

barklund: @swissspidy, what's your recommendation here? I will completely refer to you when it comes to data integrity. In my list above, should we go with option 1, 2, or 3? And if 2, how would you recommend we then go about this?

barklund: If we do look into your original idea of only sending changed data to the API, I don't see how that fixes the issue, that the client would load a non-null timestamp in and then display the publish date as that timestamp (which is part of the issue). Sure, we wouldn't be re-sending the data to the backend, but we would still be displaying the wrong date in the client, no?

merapi (approved):
I tested it a while ago and it worked fine for the issue in question.
What are the next steps here?

packages/story-editor/src/app/story/effects/useLoadStory.js
barklund: @swissspidy, fresh ping on this issue now that you're back.

swissspidy: You're right, we might still be displaying the wrong date in the client.

In that case, based on https://github.com/WordPress/gutenberg/blob/9580b45e6e18dd06076af9f7e1ea66babee22bf5/packages/editor/src/store/selectors.js#L756-L786, we should be able to go with (2), like this:

```suggestion
  const date =
    ['draft', 'auto-draft', 'pending'].includes(status) &&
    (origDate === modified || !date)
      ? null
      : origDate;
```

That seems to work fine in my testing right now.

WDYT?

barklund: Sounds great, let's go with that!

barklund: Implemented in 93612f128bd656921127367ecace9fa7107bd576.

swissspidy (approved):

PR #11583, opened 2022-05-27
Shopping: Settings: Test API Connection

### LINKED ISSUE #11267: Shopping: Settings: Test API Connection
## Feature Description

For the Shopify settings we want to offer a "Test connection" button

**Note:** Exact wording & UX is still TBD, but we should be able to implement an MVP solution for this already without it.

Here's what I had in mind:

1. Add a "Test connection" button below the two input fields
    * If Shopify host or API key is missing or if there are unsaved changes to them, the button is greyed out
2. Clicking on the button will:
    * Change the button text to "Testing connection…" (and `speak()` that as well)
    * Maybe display a spinner before/after the button
    * Call the existing REST API endpoint for fetching products
3. If the REST API response is successful:
    * Display a "Connection successful" message (and `speak()` that as well)
4. If the REST API returns an error related to authentication:
    * Display a "Connection failed. Please check your credentials." error message or something (and `speak()` that as well)
    * If the error is not clearly related to authentication (e.g. a 404, 500) then we should have a different error message. Maybe simply "Connection failed. Please try again." for now

timarney:
## Context

For the Shopify settings we want to offer a "Test connection" button

See: https://github.com/GoogleForCreators/web-stories-wp/issues/11267#issue-1204361046 for implementation details. 

Depends on https://github.com/GoogleForCreators/web-stories-wp/pull/11573

## Summary

Adds test connection button -- shows success or error depending credentials

This PR handles what's outlined in the issue we'll need a follow-up PR see todo

## To-do

Handle caching issue see: https://github.com/GoogleForCreators/web-stories-wp/pull/11583#issuecomment-1144652008 ... can create a follow-up PR

## Testing Instructions

1.

No

No

No

## Checklist

---

Fixes #11267


packages/wp-dashboard/src/components/editorSettings/shopping/test/shopping.js
```diff
@@ -182,4 +205,31 @@ describe('Editor Settings: Shopping <Shopping />', function () {
...
+      </ConfigProvider>
+    );
+
+    const input = screen.getByTestId('shopify-test-connection');
+    expect(input).toBeInTheDocument();
+
+    // eslint-disable-next-line testing-library/no-unnecessary-act, require-await
+    await act(async () => {
+      fireEvent.click(input);
+    });
```
swissspidy: If the linter says the `act()` is unnecessary — have you tried not adding it? 🙂 

All Testing Library utils are already wrapped in act. Most of the time, if you're seeing an act warning, it's not just something to be silenced, but it's actually telling you that something unexpected is happening in your test.

```suggestion
    fireEvent.click(input);
```

packages/wp-dashboard/src/components/editorSettings/shopping/shopify/index.js
```diff
@@ -150,6 +182,35 @@ function ShopifySettings({
...
+    const hasHost = hostInputError === '' && host !== '';
+    const hasToken = accessTokenInputError === '' && accessToken !== '';
+    return hasHost && hasToken && !isTestingConnection ? true : false;
+  }, [
+    host,
+    accessToken,
+    accessTokenInputError,
+    hostInputError,
+    isTestingConnection,
+  ]);
```
swissspidy: This does not need to be a callback, since it's just called directly anyway. `useMemo` would be more accurate, but also overkill. It's just a simple `const`.

I'd simplify to something like

```js
const canTestConnection = host && !hostInputError && accessToken && !accessTokenInputError && !isTestingConnection;
```

packages/wp-dashboard/src/components/editorSettings/shopping/test/shopping.js
timarney: Yes -- tried not adding ... was throwing errors about updating state.

> Warning: An update to ShopifySettings inside a test was not wrapped in act(...) + outputs a bunch more

swissspidy: You can properly await for this update (caused by the API test callback being fired after the click) this by using `waitFor()`

```js
fireEvent.click(input);

await waitFor(() => {
  expect(screen.getByTestId('api-status')).toHaveTextContent(
    'Connection successful'
  );
});
```

timarney:
@swissspidy 

Do we want to use the more detailed error message coming back from the API for this now?



> `Invalid Shopify API credentials provided` vs `Connection failed. Please try again.`

```
{code: "rest_invalid_credentials", message: "Invalid Shopify API credentials provided.",…}
code: "rest_invalid_credentials"
data: {status: 401}
message: "Invalid Shopify API credentials provided."
```

timarney:
We'll always want a fresh non cached query when checking the connection i.e. if the token gets switched between pressing the test connection button (the product query is the same).

@spacedmonkey --- to prevent caching for this test connection scenario how were you suggesting I do that? 

Can we pass a header or should I add a parameter?

spacedmonkey:
@timarney I am not sure we should reuse the search here. I think we should maybe have endpoint endpoint, called `/web-stories/v1/products/test`. This would return a response like this 

```
{ success: true }
```
Or a WP_Error on error. 

Then we can easier control this. Thoughts?

timarney:
> @timarney I am not sure we should reuse the search here. I think we should maybe have endpoint endpoint, called `/web-stories/v1/products/test`. This would return a response like this
> 
> ```
> { success: true }
> ```
> 
> Or a WP_Error on error.
> 
> Then we can easier control this. Thoughts?

I think a new endpoint makes sense in the future but given this ends up being only for the Shopify provider and only the test connection button use case ... seems like more maintenance overhead.

spacedmonkey:
> I think a new endpoint makes sense in the future but given this ends up being only for the Shopify provider and only the test connection button use case ... seems like more maintenance overhead. Also we'll still need to query something on the Shopify end ya?

We could also have a status / test endpoint for woo as well. See this logic. 

https://github.com/GoogleForCreators/web-stories-wp/blob/aade27ebe43692e6b6fc7ec2730582f81399781c/includes/Shopping/WooCommerce_Query.php#L65-L73

This a simple endpoint like this

```
	public function get_status( $request ) {
                // TODO move logic to it's own method. 
		/**
		 * Shopping provider.
		 *
		 * @var string $shopping_provider
		 */
		$shopping_provider = $this->settings->get_setting( Settings::SETTING_NAME_SHOPPING_PROVIDER );
		$query             = $this->shopping_vendors->get_vendor_class( $shopping_provider );

		if ( 'none' === $shopping_provider ) {
			return new WP_Error( 'rest_shopping_provider', __( 'No shopping provider set up.', 'web-stories' ), [ 'status' => 400 ] );
		}
			
		if ( ! $query ) {
			return new WP_Error( 'rest_shopping_provider_not_found', __( 'Unable to find shopping integration.', 'web-stories' ), [ 'status' => 400 ] );
		}
		
		$status = $query->get_status();
		if ( is_wp_error( $status ) ) {
			return $status;
		}
                $response = [
                     'success' => $status
                ];

		return rest_ensure_response( $response );
	}
```

```
register_rest_route(
			$this->namespace,
			'/' . $this->rest_base .'/status',
			[
				[
					'methods'             => WP_REST_Server::READABLE,
					'callback'            => [ $this, 'get_status' ],
					'permission_callback' => [ $this, 'get_items_permissions_check' ],
				]
			]
		);
```

timarney:
@spacedmonkey 

For the Shopify query we could also need to move the data logic etc...

All easy enough to do but I think I'll tackle outside of this PR --- keep this one more on the UI end.

```php
if ( isset( $result['errors'] ) ) {
			$wp_error = new WP_Error();
			
			foreach ( $result['errors'] as $error ) {
				$error_code = $error['extensions']['code'];
				// https://shopify.dev/api/storefront#status_and_error_codes.
				switch ( $error_code ) {
					case 'THROTTLED':
						$wp_error->add( 'rest_throttled', __( 'Shopify API rate limit exceeded. Try again later.', 'web-stories' ), [ 'status' => 429 ] );
						break;
					case 'ACCESS_DENIED':
						$wp_error->add( 'rest_invalid_credentials', __( 'Invalid Shopify API credentials provided.', 'web-stories' ), [ 'status' => 401 ] );
						break;
						
						
						... etc
```

swissspidy:
A new status test endpoint might be nice in the future, but out of scope for this ticket

packages/wp-dashboard/src/api/shopping.js
```diff
@@ -0,0 +1,40 @@
...
+ * Get products.
+ *
+ * @param {Object} config Configuation object.
+ * @param {string} search Search term.
+ * @return {Promise} Request promise.
+ */
+export function getProducts(config, search) {
+  return apiFetch({
+    path: addQueryArgs(config.api.products, {
+      per_page: 100,
```
swissspidy: For performance reasons we could reduce this to just 1.

packages/wp-dashboard/src/components/editorSettings/shopping/shopify/index.js
```diff
@@ -244,6 +300,37 @@ function ShopifySettings({
...
         </TranslateWithMarkup>
       </TextInputHelperText>
+
+      <TestConnectionButton
+        onFocus={() => setTestConnectionStatus('')}
+        type={BUTTON_TYPES.SECONDARY}
+        size={BUTTON_SIZES.SMALL}
+        disabled={!canTestConnection}
+        onClick={onTestConnection}
+        data-testid="shopify-test-connection"
```
swissspidy: Instead of using `data-testid` here and in the test, it would be better to use `getByRole('button', { name: 'Test connection' })`

`getByTestId ` [should only be the last resort](https://testing-library.com/docs/queries/bytestid) when `getByRole` & co. don't work.

swissspidy (commented):
Just a couple of nits, otherwise looks good

swissspidy (approved):

PR #11616, opened 2022-05-31
Shopping: Prevent inserting products without images

### LINKED ISSUE #11591: Shopping: Prevent inserting products without images
## Feature Description

In the shopping tab in the editor, we want to disable the Insertion button for products without images.

The button should just appear disabled but still be focusable (for a11y). (There's precedence for this in the code base, see also `aria-disabled`)

The insertion button should have a tooltip that says "Products without images cannot be added"

Ideally that's also somehow mentioned via `aria-label` (for a11y)

## Developer Notes

- Enable WooCommerce Plugin

sblinde:
## Context

This PR disables products within the product list in shopping from being added if there is no product image, as well as an informative tooltip that explains why they cannot be added.

## Relevant Technical Choices

`aria-disabled` was added here as an alternative to `disabled` which allows the user to focus on the item with keyboard navigation/screen reader instead of removing the button entirely from their view.

## User-facing changes

Products without images are disabled from being added to the Story, as seen in this screenshot.
Screen Shot 2022-05-31 at 4 55 33 PM

## Testing Instructions

1. Create a new item in a tool (like WooCommerce) and make sure it has no image.
2. With the Shopping experiment enabled, open a Story.
3. Navigate to the Shopping tab and see that the button should be disabled.
4. The button should allow for you to tab-navigate to it, but cannot be added
5. The button should also show a tooltip on hover and have an aria-label with the same message as the tooltip.
6. The button, when disabled, should not allow you to add it to the Story (either by clicking or by keyboard).

To test this running locally, make sure the following is true:
- Enable WooCommerce Plugin
- In Stories > Settings, make sure WooCommerce is selected as your Shopping Provider.

No.

No.

No.

## Checklist

---

Fixes #11591


packages/story-editor/src/components/library/panes/shopping/productButton.js
```diff
@@ -60,41 +64,78 @@ const StyledButton = styled(Button)`
...
     __('Remove %s', 'web-stories'),
     product?.productTitle
   );
 
+  const DISABLED_PRODUCT_TEXT = __(
+    'Products without images cannot be added.',
+    'web-stories'
+  );
+
+  const ariaLabel = () => {
```
swissspidy: I suppose it's a preference thing but instead of a function we could use `let ariaLabel = ...` and some if statements. I don't think we typically use functions for that elsewhere.

packages/story-editor/src/components/library/panes/shopping/productButton.js
```diff
@@ -60,41 +64,78 @@ const StyledButton = styled(Button)`
...
+};
+
 function ProductButton({ product, onClick, onFocus, isOnPage }) {
   const ADD_PRODUCT_TEXT = sprintf(
     /* translators: %s: product title. */
     __('Add %s', 'web-stories'),
     product?.productTitle
   );
 
+  const hasImage = product?.productImages && product?.productImages?.length;
```
swissspidy:
```suggestion
  const hasImage = product?.productImages?.length;
```

Same effect

packages/story-editor/src/components/library/panes/shopping/productButton.js
```diff
@@ -60,41 +64,78 @@ const StyledButton = styled(Button)`
 const StyledActionsContainer = styled.div`
   margin-right: 10px;
 `;
+const TooltipWrapper = ({ hasImage, children }) => {
```
swissspidy, line 74: 👍

packages/story-editor/src/components/library/panes/shopping/productButton.js
sblinde: Totally, can do that! Thanks for the suggestion

packages/story-editor/src/components/library/panes/shopping/productButton.js
sblinde: Can do!

packages/story-editor/src/components/floatingMenu/elements/karma/shopping.karma.js
```diff
@@ -67,6 +67,74 @@ describe('Shopping integration', () => {
     return storyContext.state.selectedElements[0];
   };
 
+  describe('Shopping tab', () => {
+    it('should handle product search add and remove', async () => {
+      const productTitle = 'Hoodie';
+      await focusProductSearchInput();
+      expect(isStoryEmpty()).toEqual(true);
```
sblinde: Fixing this quick, main branch removed a function that was used in these tests.

packages/story-editor/src/karma/fixture/db/getProductsResponse.js
```diff
@@ -75,7 +75,17 @@ export default [
     productBrand: '',
     productPrice: 18,
     productPriceCurrency: 'USD',
-    productImages: [],
+    productImages: [
```
sblinde, line 78: Product images had to be added for the previously written tests to pass.

mwritter (dismissed):
Looks good 👍

packages/story-editor/src/components/library/panes/shopping/productButton.js
```diff
@@ -60,41 +64,77 @@ const StyledButton = styled(Button)`
...
 
+  const hasImage = Boolean(product?.productImages?.length);
   const REMOVE_PRODUCT_TEXT = sprintf(
     /* translators: %s: product title. */
     __('Remove %s', 'web-stories'),
     product?.productTitle
   );
 
+  const DISABLED_PRODUCT_TEXT = __(
+    'Products without images cannot be added.',
```
timarney: Wondering if the product name should be added here.

sblinde:
> Testing via the keyboard I was able to tab to a product without an image, press enter -> insert the product.
> 
> Screen Shot 2022-06-06 at 1 04 42 PM

Thanks! I'll take a look, thanks for catching this.

packages/story-editor/src/components/library/panes/shopping/productButton.js
sblinde: I'm open to that (and initially built it with the product name) but went by what was worded in #11591 for the tooltip for consistency, but it's easy enough to change if we'd like to.

packages/story-editor/src/components/library/panes/shopping/productButton.js
```diff
@@ -60,41 +64,88 @@ const StyledButton = styled(Button)`
...
-        ) : (
-          <Icons.PlusFilled />
-        )}
-      </StyledButton>
-    </StyledActionsContainer>
+    <TooltipWrapper hasImage={hasImage}>
+      <StyledActionsContainer>
+        <StyledButton
+          aria-label={ariaLabel}
+          aria-disabled={hasImage ? false : true}
```
barklund:
```suggestion
          aria-disabled={!hasImage}
```

packages/story-editor/src/components/library/panes/shopping/productButton.js
```diff
@@ -60,41 +64,88 @@ const StyledButton = styled(Button)`
 const StyledActionsContainer = styled.div`
   margin-right: 10px;
 `;
+const TooltipWrapper = ({ hasImage, children }) => {
+  return !hasImage ? (
+    <Tooltip
+      hasTail
+      title={__('Products without images cannot be added.', 'web-stories')}
```
barklund: This string is duplicated - could we not reuse the `ariaLabel` variable here?

barklund (approved):
Looks good to me on the frontend part, just a few nits.

packages/story-editor/src/components/library/panes/shopping/productButton.js
sblinde: Yep, can do!

packages/story-editor/src/components/floatingMenu/elements/karma/shopping.karma.js
```diff
@@ -67,6 +67,70 @@ describe('Shopping integration', () => {
     return storyContext.state.selectedElements[0];
   };
 
+  describe('Shopping tab', () => {
```
swissspidy: We've set up `packages/story-editor/src/components/library/panes/shopping/karma/shopping.karma.js` in the meantime for all the shopping karma tests for the stuff in the sidebar panel.

You might wanna move these tests here to that file.

packages/story-editor/src/components/library/panes/shopping/productButton.js
```diff
@@ -60,41 +64,86 @@ const StyledButton = styled(Button)`
...
+  let ariaLabel = '';
+  if (!hasImage) {
+    ariaLabel = DISABLED_PRODUCT_TEXT;
+  } else if (isOnPage) {
+    ariaLabel = REMOVE_PRODUCT_TEXT;
+  } else {
+    ariaLabel = ADD_PRODUCT_TEXT;
+  }
+
+  const noop = () => {};
```
swissspidy: Nit: can this move outside the component or use `import { noop } from '../../../../utils/noop';`

packages/story-editor/src/components/library/panes/shopping/productButton.js
```diff
@@ -60,41 +64,86 @@ const StyledButton = styled(Button)`
...
+          // we have to utilize `noop` for the onClick
+          // as `disabled` will disable the ability to focus
+          // and tab order does not accomplish the same
+          onClick={
+            hasImage
+              ? () => {
+                  onClick(product, isOnPage);
+                }
+              : noop
+          }
```
swissspidy, line 129: Do we need `noop` here though? We're using `aria-disabled`, not `disabled`.

In my quick testing just doing `onClick={() => {
            onClick(product, isOnPage);
          }}` worked fine and I could still focus the element using the keyboard

swissspidy (commented):
Just a couple of nits, otherwise this exactly what is needed 🚀

packages/story-editor/src/components/library/panes/shopping/productButton.js
sblinde: We do need something to disable the functionality if we're only using `aria-disabled`. Per Tim's testing (see his comment above) and my own testing after, with `aria-disabled`, you can still hit enter on an element that should be disabled and add it to the page. 

`disabled` would disable this from happening but then removes it from keyboard focus, and `aria-disabled` is wholly preferred over utilizing both (or utilizing just `disabled`).

Also seems to gel with what I've read about here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled 

> ARIA attributes only provide semantics. To semantically disable an element without removing it from focus order or from the accessibility tree, set aria-disabled="true" on the element. You should include JavaScript to disable the functionality of the element while also changing the appearance of the element so sighted users know it is disabled.

packages/story-editor/src/components/library/panes/shopping/productButton.js
sblinde: Sure can!

packages/story-editor/src/components/floatingMenu/elements/karma/shopping.karma.js
sblinde: Excellent! Can do, thanks

mwritter (approved):
Looks good

swissspidy (approved):

PR #11618, opened 2022-06-01
Layer Masks

### LINKED ISSUE #9246: Layer Masks
## Summary
Instead of advanced masking (like in Photoshop) we will have a set of QoL improvements to our drop target mechanism (drop an image into shape to create a masked image).

## Acceptance Criteria

Simple masking: Re-use the current shape drop target mechanism.
- [ ] When you select an image and shape, Canvas RCM should have ~"Use Shape as a Mask" option (Layers Panel context menu should have that also? [RCM doesn't work on multi-selection in the LP so it's not possible now]).
- [ ] It should be somehow represented in the Layers Panel.
- [ ] You should be able to remove the mask (icon in the Layers Panel? - we are waiting for the design and RCM - can be "Remove Mask" below "Duplicate Element")

No

No

## Other notes 

Layer masking is going to impact keyboard users because right now the layers panel is hidden for keyboards as all functionality is mirrored in keyboard shortcuts. The accessibility of the layers panel will need to be reconsidered ahead of this functionality getting added to the layers panel.

timarney:
## Context

Adds use shape as mask functionality. 

## Summary

- Adds Right click option on Image (add / remove mask) when 1 shape and 1 image are selected
- Adds remove mask option to layers panel when a shape mask has been applied.

https://www.youtube.com/watch?v=4IsADJhje-I

## Testing Instructions

Add mask
1. Add image + shape to the stage
2. Right click on the image
3. Add mask to image

Remove Mask From RCM
1. Right click on image 
2. Remove mask

Remove Mask via Layers Panel

1. On the layers panel select the remove mask icon and remove the mask

No

No

No

## Checklist

---

Fixes #9246


merapi:
Should we disable the option or not show it when not applicable? right now when you select two images it will show "Use Shape as a Mask" disabled option - not really useful IMO, we can ask UX to confirm this.

Right now you can only drop an image into a shape, not the other way around - we didn't plan to implement that here, we should probably ask UX if that will make sense and try to estimate the effort (1 day vs 1 week effort) based on that two factors product owner should decide whatever to implement it now, in the future or never - that's the ideal scenario IMO.

For now, we can do whatever is clear and do a follow-up after the UX/product decisions.

timarney:
> Should we disable the option or not show it when not applicable? 

The other option I was thinking of was to show a SnackBar message when too many items or not the correct type of items are selected.

merapi:
> > Should we disable the option or not show it when not applicable?
> 
> The other option I was thinking of was to show a SnackBar message when too many items or not the correct type of items are selected.

That might be a little too noisy - you select two images and a shape because you wanted to group them via RCM and a shape/mask message will appear? Another option is to show a tooltip over a disabled option so you know why it's disabled. UX/design has a voice here, so if they don't propose it - we will not do it.

packages/story-editor/src/app/rightClickMenu/menus/multipleElementsMenu.js
```diff
@@ -33,6 +34,8 @@ function MultipleElementsMenu() {
     handleGroupSelectedElements,
     handleUngroupSelectedElements,
   } = useElementActions();
+  const { addShapeMask, shapeMaskElements } = useShapeMask();
+  const { shape, image } = shapeMaskElements();
```
barklund: `getShapeMaskElements()` perhaps? Or `find...`?

packages/story-editor/src/components/canvas/karma/rightClickMenu.karma.js
```diff
@@ -1352,4 +1364,58 @@ describe('Right Click Menu integration', () => {
...
+      // select shape and image targets
+      await clickOnTarget(imageFrame);
+      await clickOnTarget(shapeFrame, 'Shift');
+
+      // multiple elements should be selected
+      const { selectedElements } = await fixture.renderHook(() =>
+        useStory(({ state }) => ({
+          selectedElements: state.selectedElements,
+        }))
+      );
```
barklund: You can just select only the attribute, `useStory` doesn't have to return an object:

```js
      const selectedElements = await fixture.renderHook(() =>
        useStory(({ state }) => state.selectedElements)
      );
```

packages/story-editor/src/components/panels/design/layer/layer.js
```diff
@@ -267,11 +268,11 @@ const LayerAction = styled(Button).attrs({
...
-   * all of our Icons right now have an embedded padding,
-   * however the new designs just disregard this embedded
-   * padding, so to accommodate, we'll make the icon its
-   * intended size and manually center it within the button.
-   */
+    * all of our Icons right now have an embedded padding,
+    * however the new designs just disregard this embedded
+    * padding, so to accommodate, we'll make the icon its
+    * intended size and manually center it within the button.
+    */
```
barklund: This doesn't look right? We don't have any formatter rules for this in place, but I don't think this indent should be there - it looks off?

Screen Shot 2022-06-08 at 14 00 08

packages/story-editor/src/components/panels/design/layer/layer.js
```diff
@@ -350,6 +351,15 @@ function Layer({ element }) {
     })
   );
 
+  const { hasShapeMask, removeShapeMask } = useShapeMask();
+  const hasMask = hasShapeMask(element);
+  const removeMaskTitle = __('Remove Mask', 'web-stories');
+
+  let maskId = null;
+  if (hasMask) {
+    maskId = `mask-${element?.mask?.type}-${element.id}-frame`;
```
barklund: We should move this id generation to a central location. If someone changes the id where the mask is created, this would break.

packages/story-editor/src/components/panels/design/layer/layer.js
```diff
@@ -420,11 +430,19 @@ function Layer({ element }) {
...
-              getProxiedUrl={getProxiedUrl}
-              currentPageBackgroundColor={currentPageBackgroundColor}
-            />
+            <div
+              style={{
+                ...(hasMask
+                  ? { position: 'relative', clipPath: `url(#${maskId})` }
+                  : {}),
+              }}
+            >
```
barklund: I'd rather that we did this with a styled-component element, and then optionally added that in an extra component:

```js
const MaskedIconWrapper = styled.div`
  position: relative;
  clipPath: url(#${({ maskId }) => maskId});
`;
...
function IconWrapper({ hasMask, maskId, children }) {
  if (!hasMask) {
    return children;
  }
  return (
    <MaskedIconWrapper maskId={maskId}>
      {children}
    </MaskedIconWrapper>
  );
}
...
  <LayerIconWrapper>
    <IconWrapper hasMask={hasMask} maskId={maskId}>
      <LayerIcon ... />
    </IconWrapper>
  </LayerIconWrapper>
```

It just seems a bit cleaner to me. You could potentially even move resolution of `hasMask` and `maskId` into this component just pass `element` to it.

packages/story-editor/src/components/panels/design/layer/layer.js
```diff
@@ -350,6 +351,15 @@ function Layer({ element }) {
     })
   );
 
+  const { hasShapeMask, removeShapeMask } = useShapeMask();
```
barklund: Hm, there's a potential performance issue here. Because `useShapeMask` depends on `selectedElements`, it will re-run every time selected elements are updated, which means every layer in the layer panel will re-render every time selection changes.

However, the list of selected elements is only used for the `shapeMaskElements` function in that hook, which isn't used at all inside the layer, so all the layers re-render for nothing.

I think you need to split this hook (`useShapeMask`) into two. Create a new hook that re-renders every time selection changes which defines which of the two selected elements is the shape, and which is the image. This new hook (maybe `useShapeMaskElements`) would only be used in the right-click menu (which won't be open while selection changes anyway). The function `addShapeMask` would probably go in this hook too.

Then this old hook is now "static" in terms of selection change, and can be used in both the layer panel and the right-click menu. And then you could actually refactor this `useShapeMask` hook to take the currently selected element as an argument, which it needs internally for a number of functions, and you now don't have to pass that anymore. It can even directly return the value of `hasShapeMask` as a boolean rather than as a function.

packages/story-editor/src/utils/useShapeMask.js
```diff
@@ -0,0 +1,115 @@
...
+        x,
+        y,
+        width,
+        height,
+        scale,
+        rotationAngle,
+        mask: {
+          type: element.mask.type,
+        },
+      });
```
barklund: This can be compacted a bit via our generic util, `objectPick`:

```js
const SHAPE_PROPS = ['x', 'y', 'width', 'height', 'scale', 'rotationAngle'];
...
      const shapeProps = objectPick(element, SHAPE_PROPS);
      insertElement('shape', {
        ...shapeProps,
        mask: {
          type: element.mask.type,
        },
      });
```

barklund (approved):
In general this looks really good. There's a bunch of nits, but one issue might be consequential, so a small refactor is required for the `useShapeMask` hook.

I'm not too sure how I feel about the layer icon being the image in a mask, but without actually masking the image in the same was as on the canvas (you might have zoomed and moved the image center inside the mask), so it's a bit misleading, but I know UX lead this, so I guess it's okay.

packages/story-editor/src/components/panels/design/layer/layer.js
timarney: Good call --- that cleans things up nicely especially not need to pass the element in multiple places.

packages/story-editor/src/app/rightClickMenu/menus/multipleElementsMenu.js
```diff
@@ -34,8 +34,8 @@ function MultipleElementsMenu() {
     handleGroupSelectedElements,
     handleUngroupSelectedElements,
   } = useElementActions();
-  const { addShapeMask, shapeMaskElements } = useShapeMask();
-  const { shape, image } = shapeMaskElements();
+  const { addShapeMask, getShapeMaskElements } = useShapeMaskElements();
+  const { shape, image } = getShapeMaskElements();
```
barklund: Hmm, looking at this again, this component doesn't actually need a reference to the shape or the image. This could instead just be something like:

```js
  const { canMergeIntoMask, mergeIntoMask } = useShapeMaskElements();
  ...
        {canMergeIntoMask && (
        <ContextMenuComponents.MenuButton
          onClick={mergeIntoMask}
```

That'll clean everything up a bit more here.

barklund (approved):
This looks really good. Just had another idea for a small optimization, but it's mostly just a nit now.

packages/story-editor/src/app/rightClickMenu/menus/multipleElementsMenu.js
timarney: Nice - ya I'll make that update tomorrow A.M.

timarney: Updated

packages/story-editor/src/components/panels/design/layer/layer.js
timarney: Can add here or in a follow up PR --- where might be a good spot?

merapi: Some utility function in e.g. `story-editor/src/utils/generateMaskId.js`?

image

merapi (approved):
👌🏻 

Nit: I think the Unmask option should be hidden not disabled (it's unnecessary on every single layer).

packages/story-editor/src/components/panels/design/layer/layer.js
timarney: @merapi @barklund 

I ended up adding it to the mask package https://github.com/GoogleForCreators/web-stories-wp/pull/11618/commits/3bc21715e7b01b7491d47e1be2f665ec8e23d649 to avoid reaching out of that package.

For this PR I updated frame.js + the `shapeMaskWrapper.js` component.

timarney:
> Nit: I think the Unmask option should be hidden not disabled (it's unnecessary on every single layer).

Fixed.

PR #11624, opened 2022-06-02
Shopping: Add structured data 

### LINKED ISSUE #11153: Shopping: Add structured data
For stories containing shoppable products we want to generate [product structured data](https://developers.google.com/search/docs/advanced/structured-data/product). Since full product information is available in the editor, ideally we somehow save/prepare the structured data during output.

There are several ways to approach this, so we need to figure out what makes most sense.

**Important:** this should be built in a way so that it can be integrated with plugins like Yoast SEO, so that the structured data printed by Yoast (or some other plugin) also includes the products from the story.

**Option 1**

One possible solution off the top of my head:

Right now in the plugin we do generate some structured data in the `Discovery` class only, not in the editor. It takes some information from the `Story` model to assemble this data.

We could expand the `Story` model to read products data from the `post_content_filtered` JSON blob so we can use it in `Discovery` to print schema data.

**Potential issue:** in PHP we would have to parse the JSON, go through all products, etc. to find that required information. Plus the product elements might not contain all information we need for schema data.
To mitigate this, in the editor, we could save a duplicate list of products as part of the JSON blob with this data.

**Option 2**

**Alternative:** we could generate some minimum `<script type="application/ld+json">` markup in the `output` package. But then in PHP we would have to parse it for reuse so that we can extend it and make it available to other plugins.

**Option 3**

Store the product data also in post meta when saving a story (in addition to the JSON story data).

While this adds some redundancy, it avoids having to parse the `post_content_filtered` JSON blob, which is good.

spacedmonkey:
## Summary

Copy product data to post meta and render structured data on the front end. 

## Testing Instructions

1. Create a story with a products. 
2. Save story. 
3. View front end. 
4. Look at markup and see structured data. 

## Checklist

---

Fixes #11153


includes/Plugin.php
```diff
@@ -107,6 +107,7 @@ class Plugin extends ServiceBasedPlugin {
 		'story_post_type'              => Story_Post_Type::class,
 		'story_shortcode'              => Shortcode\Stories_Shortcode::class,
 		'svg'                          => Media\SVG::class,
+		'product_meta'                 => Shopping\Product_Meta::class,
```
swissspidy: Also need to update `.phpstorm.meta.php`

swissspidy:
Just looking at this again with fresh eyes and wondering whether this can be simplified.

In theory we don't need all these changes to the reducers as we don't need to keep this products information in state. We just need to send it when saving the story.

For that, we just need to call `getAllProducts(pages)` within `getStoryPropsToSave()`

Can you give that a try? Or am I missing something?

packages/story-editor/src/app/story/useStoryReducer/reducers/utils.js
```diff
@@ -160,3 +160,17 @@ export function exclusion(left = [], right = []) {
   const leftJoinKeys = left.map(({ id }) => id);
   return rightSet.filter(({ id }) => !leftJoinKeys.includes(id));
 }
+
+export function getAllProducts(pages) {
+  const products = pages
+    .map((page) =>
+      page?.elements
```
swissspidy: Nit: optional chaining is not necessary here, `page` should never be `null` here.

packages/story-editor/src/app/story/useStoryReducer/reducers/test/utils.js
```diff
@@ -229,3 +230,51 @@ describe('exclusion', () => {
...
+  });
+
+  it('should return a product', () => {
+    expect(
+      getAllProducts([
+        {
+          elements: [
+            {
+              type: 'product',
+              product: { id: 'b' },
```
swissspidy: Nit: products have a `productId` property, not `id`. It would be nice if the tests would reflect that.

packages/story-editor/src/app/story/useStoryReducer/reducers/utils.js
```diff
@@ -160,3 +160,17 @@ export function exclusion(left = [], right = []) {
...
+        ?.filter(({ type }) => type === 'product')
+        .map(({ product }) => product)
+    )
+    .flat()
+    .filter(Boolean);
+
+  const list = products.map((product) => JSON.stringify(product));
+  return Array.from(new Set(list)).map((product) => JSON.parse(product));
```
swissspidy: `JSON.stringify` and `JSON.parse` are rather expensive calls when all that's needed here is to de-duplicate the list of products / get a list of unique products.

Given the existing data structure (products within elements within pages) and the existing use of `Array.prototype.map` and `Array.prototype.flat` etc. above, a more straightforward solution would be to use a map / reduce pattern here, maybe with a simple `for` loop to remove duplicates or by using `new Map()` which has very similar performance than a loop, avoiding too many iterations and needless JSON conversion.

Something like this maybe:

```js
export function getAllProducts(pages) {
  const allProducts = pages.reduce((acc, page) => {
    acc.push(
      ...page.elements
        .filter(({ type }) => type === 'product')
        .map(({ product }) => product)
        .filter(Boolean)
    );
    return acc;
  }, []);

  return [
    ...new Map(
      allProducts.map((product) => [product.productId, product])
    ).values(),
  ];
}
```

spacedmonkey:
> Can you give that a try? Or am I missing something?

So I had an idea of how we could use the products data. Here. 

https://github.com/GoogleForCreators/web-stories-wp/blob/51988430aea139ed1b484844f8110501a7e4b782/packages/story-editor/src/components/library/panes/shopping/shoppingPane.js#L135

https://github.com/GoogleForCreators/web-stories-wp/blob/9db395da8c1c0447326c171363de03851977fb2f/packages/story-editor/src/components/library/panes/shopping/productDropdown.js#L39-L44

We could populate this data with all the products on the story. 

But this may be overkill...

swissspidy:
> We could populate this data with all the products on the story.

In these places we can always derive the list of products from the existing state if needed.

includes/Discovery.php
```diff
@@ -231,6 +245,56 @@ protected function get_schemaorg_metadata(): array {
...
+					'ratingValue' => $product['aggregateRating']['ratingValue'] ?? 0,
+					'reviewCount' => $product['aggregateRating']['reviewCount'] ?? 0,
+					'url'         => $product['aggregateRating']['reviewUrl'] ?? '',
+				];
+			}
+			$product_data[] = $data;
+		}
+
+		return [
+			'products' => [
```
swissspidy:
```suggestion
			'mainEntity' => [
```

Otherwise it does not pass the schema markup validator.

includes/Discovery.php
```diff
@@ -231,6 +245,56 @@ protected function get_schemaorg_metadata(): array {
...
+		}
+		$product_data = [];
+		foreach ( $products as $product ) {
+			$data = [
+				'@type'     => 'Product',
+				'brand'     => $product['productBrand'] ?? '',
+				'productID' => $product['productId'] ?? '',
+				'url'       => $product['productUrl'] ?? '',
+				'name'      => $product['productTitle'] ?? '',
+			];
```
swissspidy: Just realized we can also add the product price here.

```suggestion
			$data = [
				'@type'     => 'Product',
				'brand'     => $product['productBrand'] ?? '',
				'productID' => $product['productId'] ?? '',
				'url'       => $product['productUrl'] ?? '',
				'name'      => $product['productTitle'] ?? '',
				'offers'    => [
					[
						'@type'         => 'Offer',
						'price'         => $product['productPrice'] ?? 0,
						'priceCurrency' => $product['productPriceCurrency'] ?? '',
					],
				],
			];
```

includes/Discovery.php
```diff
@@ -231,6 +245,56 @@ protected function get_schemaorg_metadata(): array {
...
+			if ( isset( $product['productImages'] ) && $product['productImages'] ) {
+				$data['image'] = $product['productImages'][0]['url'];
+			}
+			if ( isset( $product['aggregateRating'] ) && $product['aggregateRating'] ) {
```
swissspidy:
```suggestion
			if ( ! empty( $product['aggregateRating']['reviewCount'] ) ) {
```

Schema.org validation complains if there's an `AggregateRating` but `reviewCount` is zero.

This way, we don't add this block if there are no reviews.

packages/story-editor/src/app/story/useStoryReducer/reducers/utils.js
spacedmonkey: This was copied and pasted from elsewhere. 

https://github.com/GoogleForCreators/web-stories-wp/blob/e00e50da7a02d956b993f4038ca1c15d46eda3ef/packages/story-editor/src/utils/getUniquePresets.js#L16-L19

I have refactored it now, to be a little cleaner and easily to now what is going on.

packages/story-editor/src/app/story/utils/test/getAllProduct.js
```diff
@@ -0,0 +1,67 @@
...
+
+describe('getAllProducts', () => {
+  it('should return empty', () => {
+    expect(getAllProducts([])).toStrictEqual([]);
+  });
+
+  it('should return a product', () => {
+    expect(
+      getAllProducts([
+        {
```
timarney, line 48: Maybe add some data that isn't a valid product as well to ensure it filters.

spacedmonkey: Why would the product be invalid here? If it was valid, what could we do about this?

timarney: Could be a different test --- just something to test the `type` is filtering out non products 

   ```elements.forEach(({ product, type }) => {```

spacedmonkey: Done.

includes/Shopping/Product_Meta.php
```diff
@@ -0,0 +1,129 @@
...
+	 * @param int $story_id ID of story.
+	 * @return array
+	 */
+	public function get_products( int $story_id ): array {
+		/**
+		 * Product data.
+		 *
+		 * @var array|false $products
+		 */
+		$products = get_post_meta( $story_id, $this::PRODUCTS_POST_META_KEY, true );
```
spacedmonkey:
```suggestion
		$products = get_post_meta( $story_id, self::PRODUCTS_POST_META_KEY, true );
```

includes/Discovery.php
```diff
@@ -231,6 +248,57 @@ protected function get_schemaorg_metadata(): array {
...
+				'productID' => $product['productId'] ?? '',
+				'url'       => $product['productUrl'] ?? '',
+				'name'      => $product['productTitle'] ?? '',
+				'offers'    => [
+					[
+						'@type'         => 'Offer',
+						'price'         => $product['productPrice'] ?? 0,
+						'priceCurrency' => $product['productPriceCurrency'] ?? '',
+					],
+				],
```
swissspidy: Looks like we can also add the `description` here with the `productDetails` field

```suggestion
				'@type'            => 'Product',
				'brand'            => $product['productBrand'] ?? '',
				'productID'        => $product['productId'] ?? '',
				'url'              => $product['productUrl'] ?? '',
				'name'             => $product['productTitle'] ?? '',
				'description'      => $product['productDetails'] ?? '',
				'offers'           => [
					[
						'@type'         => 'Offer',
						'price'         => $product['productPrice'] ?? 0,
						'priceCurrency' => $product['productPriceCurrency'] ?? '',
					],
				],
```

After that this should be good to go!

spacedmonkey: Fixed.

swissspidy (approved):

Issue #11620, opened 2022-06-01
Dashboard: Consolidating the FilterProvider to include Author Filter


sblinde:
## Feature Description
As part of the feature for Dashboard: Filter by Taxonomy (#11436), our team had to make a larger Filter Provider that held information specific to Taxonomies in order to properly create filters. However, the logic between the Filter by Author work and the Filter by Taxonomy work is separate, in two different places. 

This body of work is to unite the two into the larger FilterProvider, so that as we continue onto more filter work in the future (see: #11586), we have one space that holds the logic for each filter, based on the logic of each type of filter that's needed.

This story will block the work needed to accomplish #11586, as it would be better to improve upon the filter provider the team has built before moving on to adding more filters within the Dashboard space.

This tackles both Code Quality and an Enhancement, supporting further filter work, and improving upon the work delivered by #11436.

## Acceptance Criteria

- Both the Taxonomies filters and the Author filters will exist under one `FilterProvider`
- Adding another filter in the future should be made easier by this work: this work should allow one to extend the filters as needed for future work

## Alternatives Considered

The alternative would be moving forward to add more filters, which would add more separation between filters (which is what exists today with the last body of work to support Filtering by Taxonomies). It would be ideal if we could accomplish this body of tech debt work prior to that work.

swissspidy:
Appreciate you all keeping an eye on code quality 👍 🚀

swissspidy:
Another thing I mentioned at https://github.com/GoogleForCreators/web-stories-wp/pull/11625#discussion_r892222772 is to show the taxonomies & author filters also on the list view

Might be something we can do as part of this ticket as well.

mwritter:
https://user-images.githubusercontent.com/14057928/173854259-6aaf2224-b3db-4c49-a324-bdeda906db05.mp4
Currently we're making a request on every key stroke to search within the 'datalist' - will fix here

PR #11625, opened 2022-06-02
Dashboard: support filtering stories by taxonomy

### LINKED ISSUE #11436: Dashboard: Add Filter by Taxonomy

mwritter:
## Context

Add programmatic taxonomy filters to the dashboard.

## Summary

Give the user the ability to filter by story taxonomies.  Similar to the author filter.

## Relevant Technical Choices

There can be multiple taxonomy filters, and a programmatic solution is needed to build out these filters.  The solution used here lends itself to encompass all dashboard filters state into one place making the filtering process simpler later on.

## To-do

Move all dashboard story filters into the new provider to simplify dashboard story filtering

## User-facing changes

before/after
https://user-images.githubusercontent.com/14057928/171685426-33d5664b-0485-4567-8773-d45d6aae5bf8.mp4

## Testing Instructions

1. Go to story dashboard

On Staging:
- Notice there are no taxonomy filters

On Branch:
- Notice there are taxonomy filters (categories, verticals)
2. Make sure there are stories with taxonomies assigned to them
3. Test filters correctly filters stories
4. Test that the search within the filter dropdown is working correctly as well

No

No

No

## Checklist

---

Fixes #11436


packages/dashboard/src/app/views/shared/bodyViewOptions.js
```diff
@@ -91,6 +98,36 @@ export default function BodyViewOptions({
...
         </Text>
         <ControlsContainer>
+          {layoutStyle === VIEW_STYLE.GRID && filters?.length
+            ? filters.map((filter) => (
+                <StorySortDropdownContainer key={filter.key}>
+                  <StyledDatalist
+                    hasSearch
+                    hasDropDownBorder
+                    searchResultsLabel={__(
+                      filter.labels.searchItems,
```
mwritter: @spacedmonkey @swissspidy - These filters have to be created programmatically, it seems like `i18n-no-variables` needs these value to be hardcoded though.  Do you have any suggestions on how we should handle these?  (Remove i18n function calls here?) - https://github.com/GoogleForCreators/web-stories-wp/pull/11526#issuecomment-1132880329

sblinde: I think you could use something like this (permalink is to my PR of an example within the Shopping feature, but should be a helpful example):
https://github.com/GoogleForCreators/web-stories-wp/blob/7ac2f5196e00204a7f85a7324b106beaea5ed6a3/packages/story-editor/src/components/library/panes/shopping/productButton.js#L86-L90

As far as which line to use, I'd love feedback on that. I'm not sure if searchItems or all-items is the preferred title for the dropdown, sounds like searchItems perhaps.

sblinde:
Gave this an initial review and it looks solid to me once the i8n changes are accommodated (and feedback for which value we should use in the dropdown). Glad to give my approval after that, but because I contributed heavily to this work, would be ideal to have someone outside of the two of us review this as well. Nice work!

sblinde (dismissed):
👏

packages/dashboard/src/app/views/shared/bodyViewOptions.js
mwritter: @sblinde @swissspidy - We're getting the exact wording we need from the backend but `i18n` linter `i18n-no-placeholders-only` doesn't like when we use `__('%s', 'web-stories', filter.label.searchTerm)`.

sblinde: @mwritter I think there's two options here:
1. We leave things the way they are in this PR (but add an override to the linter in this case)
2. Or, we could do something like `Search %s` and the label we use instead is `filter.label.all-items`.

Both aren't the most ideal option, unfortunately, but the second one might allow for more translation opportunities.

packages/dashboard/src/app/views/shared/bodyViewOptions.js
```diff
@@ -91,6 +98,30 @@ export default function BodyViewOptions({
           <TranslateWithMarkup>{resultsLabel}</TranslateWithMarkup>
         </Text>
         <ControlsContainer>
+          {layoutStyle === VIEW_STYLE.GRID && filters?.length
```
swissspidy, line 99: Sorta related: wondering why we don't currently support filtering on list view. Maybe worth adding a ticket?

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
```diff
@@ -0,0 +1,151 @@
...
+};
+
+/**
+ * Hook used for taxonomy filters logic
+ * Initilizes the taxonomy filters data
+ *
+ * @return {Array} [taxonomies, queryTaxonomyTerm] initial taxonomy filters data and a function to query the terms
+ */
+
+function useTaxonomyFilters() {
```
swissspidy:
```suggestion
/**
 * Hook used for taxonomy filters logic
 * Initializes the taxonomy filters data
 *
 * @return {Array} [taxonomies, queryTaxonomyTerm] initial taxonomy filters data and a function to query the terms
 */
function useTaxonomyFilters() {
```

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
```diff
@@ -0,0 +1,151 @@
...
+      },
+    }) => ({
+      getTaxonomies,
+      getTaxonomyTerm,
+    })
+  );
+
+  /**
+   * Query individual taxonomy data
+   * This is needed to initilize the primaryOptions and needed to search and set queriedOptions
```
swissspidy:
```suggestion
   * This is needed to initialize the primaryOptions and needed to search and set queriedOptions
```

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
```diff
@@ -0,0 +1,151 @@
...
+   */
+  const queryTaxonomies = useCallback(async () => {
+    const data = await getTaxonomies();
+    const promises = [];
+    const hierarchicalTaxonomies = data.filter(({ hierarchical }) =>
+      Boolean(hierarchical)
+    );
+
+    for (const taxonomy of hierarchicalTaxonomies) {
+      // initilize the data with an empty array
```
swissspidy:
```suggestion
      // initialize the data with an empty array
```

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
```diff
@@ -0,0 +1,151 @@
...
+   * This should only be needed once get all the taxonomies and set individual term data
+   *
+   * @see queryTaxonomyTerm
+   */
+  const queryTaxonomies = useCallback(async () => {
+    const data = await getTaxonomies();
+    const promises = [];
+    const hierarchicalTaxonomies = data.filter(({ hierarchical }) =>
+      Boolean(hierarchical)
+    );
```
swissspidy: @spacedmonkey WDYT about adding support for filtering hierarchical taxonomies in `Stories_Taxonomies_Controller::get_collection_params()`?

Then we could fetch `/wp-json/web-stories/v1/taxonomies?type=web-story&hierarchical=true` directly and don't need to do the filtering client-side.

Would be a nice follow-up code quality ticket.

includes/Admin/Dashboard.php
```diff
@@ -303,6 +303,7 @@ public function load_stories_dashboard(): void {
 			'/web-stories/v1/settings/',
 			'/web-stories/v1/publisher-logos/',
 			'/web-stories/v1/users/me/',
+			'/web-stories/v1/taxonomies/',
```
swissspidy:
```suggestion
			'/web-stories/v1/taxonomies/?' . build_query(
				[
					'type'    => $this->story_post_type->get_slug(),
					'context' => 'edit',
				]
			),
```

The query args here need to match the query args used in `getTaxonomies()` in `packages/wp-dashboard/src/api/taxonomies.js`

This way, we can save 1 HTTP request when loading the dashboard.

To verify this chang, open the dev tools and see that there is no HTTP request to `/web-stories/v1/taxonomies/` when opening the dashboard

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
```diff
@@ -0,0 +1,151 @@
...
+
+export const getFilterPlaceholder = (filter) => {
+  return filter.labels?.allItems || filter.labels?.name || filter.restBase;
+};
+
+export const getFilterAriaLabel = (filter) => {
+  const ariaLable =
+    filter.labels?.singularName || filter.labels?.name || filter.restBase;
+  return `Filter stories by ${ariaLable}`;
+};
```
swissspidy: Note that all strings need to use `__()` so they can be translated ⚠️  Here with `sprintf` and a `translators:` comment for the placeholder.

However, here we don't need a custom label & function for it. We can just use the existing `filter.labels.filterByItem` label.

See my below suggestions that make these functions here obsolete.

---

Aside: Typo `ariaLable` -> `ariaLabel`

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
```diff
@@ -0,0 +1,151 @@
...
+    return taxonomies.map((taxonomy) => ({
+      key: taxonomy.restBase,
+      restPath: taxonomy.restPath,
+      labels: taxonomy.labels,
+      filterId: null,
+      primaryOptions: taxonomy.data,
+      queriedOptions: taxonomy.data,
+      query: queryTaxonomyTerm,
+      placeholder: getFilterPlaceholder(taxonomy),
+      ariaLabel: getFilterAriaLabel(taxonomy),
```
swissspidy:
```suggestion
      ariaLabel: taxonomy.labels.filterByItem,
```

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
```diff
@@ -0,0 +1,151 @@
...
+  const initializeTaxonomyFilters = useCallback(() => {
+    return taxonomies.map((taxonomy) => ({
+      key: taxonomy.restBase,
+      restPath: taxonomy.restPath,
+      labels: taxonomy.labels,
+      filterId: null,
+      primaryOptions: taxonomy.data,
+      queriedOptions: taxonomy.data,
+      query: queryTaxonomyTerm,
+      placeholder: getFilterPlaceholder(taxonomy),
```
swissspidy:
```suggestion
      placeholder: taxonomy.labels.allItems,
```

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
```diff
@@ -0,0 +1,151 @@
...
+      ariaLabel: getFilterAriaLabel(taxonomy),
+      searchResultsLabel: getFilterSearchResultsLabel(taxonomy),
```
swissspidy:
```suggestion
      searchResultsLabel: taxonomy.labels.searchItems,
```

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
```diff
@@ -0,0 +1,151 @@
...
+export const getFilterAriaLabel = (filter) => {
+  const ariaLable =
+    filter.labels?.singularName || filter.labels?.name || filter.restBase;
+  return `Filter stories by ${ariaLable}`;
+};
+
+export const getFilterSearchResultsLabel = (filter) => {
+  return (
+    filter.labels?.searchItems ||
+    (filter.labels?.name ? `Search ${filter.labels.name}` : 'Search Taxonomy')
```
swissspidy: Ditto here about missing `__()` calls, but again, see my below suggestions that make this function here obsolete.

packages/wp-dashboard/src/api/taxonomies.js
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Get a taxonomy term.
+ *
+ * @param {Object} config configuration object.
+ * @param {string} endpoint absolute url to interact with taxonomy
+ * @param {Object} args Additional args.
+ * @return {Promise} Term promise.
+ */
+
+export function getTaxonomyTerm(config, endpoint, args = {}) {
```
swissspidy:
```suggestion
/**
 * Get a taxonomy terms.
 *
 * @param {Object} config configuration object.
 * @param {string} endpoint absolute url to interact with taxonomy
 * @param {Object} args Additional args.
 * @return {Promise} List of taxonomy terms.
 */

export function getTaxonomyTerms(config, endpoint, args = {}) {
```

This returns a list of all terms, no?

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
```diff
@@ -0,0 +1,151 @@
...
+
+  /**
+   * Query individual taxonomy data
+   * This is needed to initilize the primaryOptions and needed to search and set queriedOptions
+   *
+   * @param {Object} taxonomy object holding taxonomy data, including restBase and restPath used to fetch individual taxonomy terms
+   * @param {string|undefined} search string use to query taxonomy by name
+   * @return {Array} taxonomy terms
+   */
+  const queryTaxonomyTerm = useCallback(
```
swissspidy: Also use plural here?

swissspidy (commented):
Overall this is looking pretty good so far and seems to work well.

Just a few minor notes here and there that should also address your i18n questions.

packages/dashboard/src/app/views/myStories/index.js
```diff
@@ -78,8 +79,16 @@ function MyStories() {
...
   );
-  const { apiCallbacks, canViewDefaultTemplates } = useConfig();
+  const { filtersLoading, getFiltersObject } = useFilters(
+    ({ state: { filtersLoading }, actions: { getFiltersObject } }) => ({
+      filtersLoading,
+      getFiltersObject,
+    })
+  );
+
+  const _filters = JSON.stringify(getFiltersObject());
```
swissspidy: What's the reason for using `JSON.stringify` and `JSON.parse()` for the filters? They're not usually cheap calls, so if we can avoid them, that would be nice.

packages/dashboard/src/app/views/myStories/filters/provider.js
```diff
@@ -0,0 +1,132 @@
...
+   * Returns a object where the keys are the filter keys
+   * and the values are the filterId
+   *
+   * @return {Object}
+   */
+  const getFiltersObject = useCallback(() => {
+    const filterObj = {};
+    for (const filter of state.filters) {
+      const { key, filterId } = filter;
+      filterObj[key] = filterId;
```
swissspidy:
```suggestion
      if (filterId) {
        filterObj[key] = filterId;
      }
```

This avoids `null` Object values, which is important as otherwise these `null` values get passed all the way down to `fetchStories()`, breaking our REST API preloading.

To verify this change, open devtools and verify there's no HTTP request to `wp-json/web-stories/v1/web-story` when opening the dashboard.

packages/dashboard/src/app/views/shared/bodyViewOptions.js
mwritter: For sure!

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
mwritter: 🤦‍♂️

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
mwritter: I totally agree, I originally was trying to use `__()` and `sprintf` - https://github.com/GoogleForCreators/web-stories-wp/pull/11625/commits/5e63aef8a1855b804f4caa052b8b4a94515674f8#:~:text=export%20const%20getFilterPlaceholder,.key%3B - but the `i18n` linter does not allow single placeholders - https://github.com/GoogleForCreators/web-stories-wp/pull/11625#discussion_r888240652 and  https://github.com/WordPress/gutenberg/blob/trunk/packages/eslint-plugin/docs/rules/i18n-no-placeholders-only.md - and we're getting the exact wording from the backend that we want to use here.  Should we ignore the linting warning or is there a better way to handle this?

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
mwritter: 🤔 can we assume these 'labels' are going to be on every taxonomy?

swissspidy: Yes 👍

packages/dashboard/src/app/views/myStories/index.js
mwritter: This was done because the `filters` in the filters provider was causing an infinite loop as a dependency on the `useEffect` that was calling `fetchStories` but I do see away around that 👍

includes/Admin/Dashboard.php
spacedmonkey: You will also need to change the preloading as well.

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
spacedmonkey: Also you `sprintf` here.

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
```diff
@@ -0,0 +1,151 @@
...
+        taxonomyApi: { getTaxonomies, getTaxonomyTerm },
+      },
+    }) => ({
+      getTaxonomies,
+      getTaxonomyTerm,
+    })
+  );
+
+  /**
+   * Query individual taxonomy data
```
spacedmonkey:
```suggestion
   * Query individual taxonomy data.
```

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
```diff
@@ -0,0 +1,151 @@
...
+   * This is needed to initilize the primaryOptions and needed to search and set queriedOptions
+   *
+   * @param {Object} taxonomy object holding taxonomy data, including restBase and restPath used to fetch individual taxonomy terms
+   * @param {string|undefined} search string use to query taxonomy by name
```
spacedmonkey:
```suggestion
   * @param {string} search string use to query taxonomy by name
```

includes/Admin/Dashboard.php
swissspidy: This _is_ the preloading section.

This suggestion fixes the preloading.

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
spacedmonkey: 💯

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
spacedmonkey: I just tried to prototype this. It would not be simple to make this change on the REST API.

spacedmonkey: There are no functions to query by attributes and then get objects back. It would have to do manual searches for this prooperty. I dont see this being that big of a problem. This lookup is very quick.

mwritter: Also, It would be convenient to be able to fetch all the taxonomies with their terms instead of having to make multiple calls to the backend here, though Im not sure how difficult that would be to implement in.

packages/wp-dashboard/src/api/taxonomies.js
```diff
@@ -0,0 +1,63 @@
...
+ * @return {Promise} Taxonomies promise.
+ */
+
+export async function getTaxonomies(config) {
+  const result = await apiFetch({
+    path: addQueryArgs(config.api.taxonomies, {
+      type: 'web-story',
+      context: 'edit',
+    }),
+  });
```
spacedmonkey: This request needs to be preloaded.

mwritter: I believe this is addressed here https://github.com/GoogleForCreators/web-stories-wp/pull/11625#discussion_r892237701 right

spacedmonkey:
Screenshot 2022-06-08 at 15 33 05

If you filter by term that has no stories, then it show an odd message. Some kind of error message would be useful here. 

https://user-images.githubusercontent.com/237508/172643669-19e73a27-a3bd-49b4-af57-dc90476fadc2.mov

The filters load in much later. Let's ensure that these requests are preloaded and loading in gracefully.

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
spacedmonkey: I did some research into this for now. There is no scalable way of doing this. If some taxonomies have 100+ terms, then that single request would result in far to many database queries.

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
```diff
@@ -0,0 +1,151 @@
...
+   * @return {Array} taxonomy terms
+   */
+  const queryTaxonomyTerm = useCallback(
+    async (taxonomy, search) => {
+      const { restBase, restPath } = taxonomy;
+      const terms = await getTaxonomyTerm(restPath, { search });
```
spacedmonkey:
```suggestion
      const terms = await getTaxonomyTerm(restPath, { search, per_page: -1 });
```

Let's get all terms here, no little to 100.

mwritter:
> If you filter by term that has no stories, then it show an odd message. Some kind of error message would be useful here.

This seems to be the default screen/message when there are no stories returned based on the filters (see filter by 'author').  A new message could be shown if that is needed.

mwritter:
> The filters load in much later. Let's ensure that these requests are preloaded and loading in gracefully.

👍 Ill look into that!

packages/wp-dashboard/src/api/taxonomies.js
mwritter: Yes, plural terminology does work better here 👍

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
swissspidy: > There are no functions to query by attributes and then get objects back. It would have to do manual searches for this prowebperty. I dont see this being that big of a problem. This lookup is very quick.

There are. `get_taxonomies()` uses `wp_filter_object_list()` for that.

I'll whip up a quick PR to explain.

swissspidy: See #11652

swissspidy: > I did some research into this for now. There is no scalable way of doing this. If some taxonomies have 100+ terms, then that single request would result in far to many database queries.

+1

It's also not very RESTful

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
swissspidy: You would need to use `__()` only when more text is involved. If it's just a variable and nothing else, you don't need to wrap it in `__()`. That's what the lint rule guards against.

If you have text like "Do something with %s" -> wrap it in `__()`

If you only have a variable like `fooBar` -> just use as-is

swissspidy:
>  Let's ensure that these requests are preloaded and loading in gracefully.

Do you mean preloading the requests for all taxonomy terms endpoints (e.g. /wp-json/web-stories/v1/web-story-tag` , and not just the `/wp-json/web-stories/v1/taxonomies` endpoint?

> A new message could be shown if that is needed, if so we should add that information to the ticket.

A new ticket might be easiest if it's too complicated.

Otherwise should be able to display `Sorry, we couldn't find any results.` like we do when searching for stories

sblinde:
> > Let's ensure that these requests are preloaded and loading in gracefully.
> 
> Do you mean preloading the requests for all taxonomy terms endpoints (e.g. /wp-json/web-stories/v1/web-story-tag`, and not just the`/wp-json/web-stories/v1/taxonomies` endpoint?
> 
> > A new message could be shown if that is needed, if so we should add that information to the ticket.
> 
> A new ticket might be easiest if it's too complicated.
> 
> Otherwise should be able to display `Sorry, we couldn't find any results.` like we do when searching for stories

@swissspidy We could do this in #11620, actually, might be a good place for that enhancement while we consolidate the Filters together.

spacedmonkey:
Instead of no matches found, should it should the label 
Screenshot 2022-06-09 at 11 58 07

There are labels for no terms and no terms found. 
Screenshot 2022-06-09 at 12 00 51

spacedmonkey:
Somewhat unrelated to this change, but worth noting. Seeing clipping of gs and long category names very hard to read. 
Screenshot 2022-06-09 at 12 02 40

Maybe just making the dropdowns a little wider might do the trick.

packages/wp-dashboard/src/api/taxonomies.js
swissspidy: Correct.

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
```diff
@@ -0,0 +1,152 @@
...
+      labels: taxonomy.labels,
+      filterId: null,
+      primaryOptions: taxonomy.data,
+      queriedOptions: taxonomy.data,
+      query: queryTaxonomyTerms,
+      placeholder: sprintf(
+        /* translators: All Taxonomies label*/
+        '%s',
+        taxonomy.labels.allItems
+      ),
```
swissspidy, line 128:
```suggestion
      placeholder: taxonomy.labels.allItems,
```

There's no need to use `sprintf` here.

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
```diff
@@ -0,0 +1,152 @@
...
+      ariaLabel: sprintf(
+        /* translators: Filter by Taxonomy label*/
+        '%s',
+        taxonomy.labels.filterByItem
+      ),
+      searchResultsLabel: sprintf(
+        /* translators: Search Taxonomy label*/
+        '%s',
+        taxonomy.labels.searchItems
+      ),
```
swissspidy:
```suggestion
      placeholder: taxonomy.labels.allItems,
      ariaLabel: taxonomy.labels.filterByItem,
      searchResultsLabel: taxonomy.labels.searchItems,
```

There's no need to use `sprintf` here. It doesn't make any sense to use that if there's nothing but a placeholder in the string. Just use the variable directly.

Translator comments are only needed when actually using `sprintf` in combination with `__()`

spacedmonkey:
These labels seem wrong. 
Screenshot 2022-06-09 at 12 18 25
The place holder here, should be Search category. 

The title here should be search results and not Search category. 
Screenshot 2022-06-09 at 12 18 41

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
swissspidy: Usually yes, but here we remove the code.

mwritter:
> These labels seem wrong

I agree, I did notice this.  These filters were modeled from the other filter on this page - 'author' which uses the generic 'Search' and 'No matches found' labels. But I do believe it would be better as you mentioned to use the appropriate labels instead of generic ones.  Ill update the taxonomy filters and the author filter to use the more specific labels here.

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
mwritter: Thanks for all the insight 😅 Im a little new to the `i18n` stuff.  So we want to use `i18n` with these labels correct?  We can return just the placeholders here, and are you suggesting use `sprintf()` in `bodyViewOptions` instead?

swissspidy: No. These are already translated server-side. No need to do anything else.

No placeholders, no `sprintf`.

What I shared in the suggestion above is all that's needed.

mwritter: Oh okay, awesome thank you for clarifying that for me!

mwritter:
>The place holder here, should be Search category.

@swissspidy @spacedmonkey

These filters are a fixed width which will cause the placeholder to be truncated for these labels. Do we mind if they are truncated?  If we need to change the width to accommodate the placeholders this ticket is going to be bigger lift and will most likely cause this feature to be pushed to the next sprint. The other labels seem to work fine to me though.


Could this fall under https://github.com/GoogleForCreators/web-stories-wp/issues/11620 ?

mwritter:
>>Let's ensure that these requests are preloaded and loading in gracefully.

>Do you mean preloading the requests for all taxonomy terms endpoints (e.g. /wp-json/web-stories/v1/web-story-tag , and not just the /wp-json/web-stories/v1/taxonomies endpoint?

🤔 @spacedmonkey - also is it possible to preload these requests?

swissspidy:
> These filters are a fixed width which will cause the placeholder to be truncated for these labels. Do we mind if they are truncated? If we need to change the width to accommodate the placeholders this ticket is going to be bigger lift and will most likely cause this feature to be pushed to the next sprint. The other labels seem to work fine to me though.

It should be OK for now, but can we at least add `text-overflow: ellipsis;` for now?

We can definitely improve things later as part of #11620?

> also is it possible to preload these requests?

It should be. I'm sure @spacedmonkey can submit a small PR against your branch to enable this. No need for you to dig into PHP more than necessary :)

mwritter:
> can we at least add `text-overflow: ellipsis;` ?

For sure, good idea 👍

> We can definitely improve things later as part of #11620?

> > also is it possible to preload these requests?
> It should be. I'm sure @spacedmonkey can submit a small PR against your branch to enable this. No need for you to dig into PHP more than necessary :) 
🙏

spacedmonkey:
> It should be. I'm sure @spacedmonkey can submit a small PR against your branch to enable this. No need for you to dig into PHP more than necessary :)

I don't think preloading all these requests is the best idea. Instead of waiting until REST API request completes, can we render the dropdown with no options and load the options in an effect, that re-renders the dropdown? This way the dropdown would render and the options will pop in when they are ready?

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
```diff
@@ -79,13 +78,10 @@ function useTaxonomyFilters() {
    * @see queryTaxonomyTerms
    */
   const queryTaxonomies = useCallback(async () => {
-    const data = await getTaxonomies();
+    const data = await getTaxonomies({ hierarchical: true });
```
swissspidy, line 106: 🎉

sblinde:
@mwritter attempted to get the preloading working for fetching the **taxonomies data only**, and from what I've seen working with him on this, the delay loading the additional taxonomies-related filters still happens, despite setting his work up to preload the taxonomies, using the suggestions made by @spacedmonkey for adding it to the PHP file.

When he adds static taxonomies data to debug this, there is no flash, indicating that the UI attempts to load the Dashboard prior to the preloading being completed for taxonomies. There may be more to what's needed beyond our scope of knowledge of the PHP part of this repo to support taxonomies being preloaded. To note as well, this is with the absence of any specific data about the actual terms associated with the taxonomies.

So I think, given this being the end of the sprint tomorrow, the complexity of this work, and the need for this to be QA'd, here's some options that seem viable to me for moving forward:

1. We move forward with what @mwritter has in this PR as part of our next release, and next sprint we dig deeper into why preloading is not working appropriately for **taxonomies ONLY** (not the individual terms) as part of the greater epic of #11663. This means the short delay in loading the dropdowns will exist with this body of work until the following release.
2. We do not move this filter by taxonomy work along as part of this upcoming release, and we halt on delivering the filter by taxonomy until the following release (where this, plus the #11663 work can be delivered).

I also know that there's been some discussion of whether preloading should be done for the terms associated with taxonomies. I don't think that's necessary to support preloading all of the terms here: we can fetch them after we get taxonomies. However, preloading terminologies, if we want to avoid the delay in the dropdowns (especially because this is user-defined and user controlled custom taxonomies support), will need to happen to avoid the visual delay in seeing the taxonomies dropdown. 

To note also: I can replicate it more easily with throttling my browser to `Slow 3G`, but it's harder to catch with a faster connection as far as I can replicate.

Happy to talk about this when everyone's on early tomorrow eastern time too. I just want to make sure we're appropriately tackling the feedback here.

cc: @swissspidy @spacedmonkey @mwritter

includes/Admin/Dashboard.php
```diff
@@ -303,6 +303,12 @@ public function load_stories_dashboard(): void {
 			'/web-stories/v1/settings/',
 			'/web-stories/v1/publisher-logos/',
 			'/web-stories/v1/users/me/',
+			'/web-stories/v1/taxonomies/?' . build_query(
+				[
+					'type'    => $this->story_post_type->get_slug(),
+					'context' => 'edit',
```
spacedmonkey:
```suggestion
					'context' => 'edit',
					'hierarchical' => 'true'
```

Need to add this now we passing this param in the quest to ensure that is correctly preloaded.

swissspidy: Heads up that the arrows need to be aligned, otherwise there's a lint error.

packages/design-system/src/components/datalist/searchInput.js
```diff
@@ -88,6 +88,17 @@ const Input = styled.input.attrs({
...
+    text-overflow: ellipsis;
+  }
+  &::-webkit-input-placeholder {
+    /* Chrome/Opera/Safari */
+    text-overflow: ellipsis;
+  }
+  &::-moz-placeholder {
+    /* Firefox 19+ */
+    text-overflow: ellipsis;
+  }
```
swissspidy, line 101: This does not seem to be working, text is cut off without ellipsis. Can you check this again?

These might also need `white-space: nowrap; overflow: hidden;`

swissspidy:
@sblinde In short, let's proceed.

But let me just take a step back, ignore preloading, PHP, etc. to describe the issue at hand.

When the dashboard loads, the authors dropdown is already visible, just with a "No authors found" message until all authors have been loaded:

Screenshot 2022-06-10 at 13 47 00

On the other hand, the taxonomy dropdowns are only displayed after all the _terms_ have been loaded (`getTaxonomyTerms`), even though the _taxonomies_ themselves (`getTaxonomies`) have been loaded.

**Note:** just to clarify, the `getTaxonomies` call is in fact already preloaded, so there's no HTTP request we're waiting for there.

https://user-images.githubusercontent.com/841956/173059337-7df7e943-9d81-4364-9650-52da60a757fe.mov

The question is:

Is it possible to display the dropdowns before all the terms have been loaded, just with a "No xyz found" message?

From what I can see, yes, it is possible.

The issue is in the `queryTaxonomies` callback within `useTaxonomyFilters()`:
https://github.com/GoogleForCreators/web-stories-wp/blob/305dc9d7787956a3c5985c256a1e605335e8e3b9/packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js#L74-L104

`setTaxonomies()` is only ever called after **all the terms** have been loaded.

A quick hack locally fixed this (though the terms loading seemed to be broken afterwards):

```diff
  const queryTaxonomies = useCallback(async () => {
    const data = await getTaxonomies({ hierarchical: true });
+   setTaxonomies(data);
    const promises = [];
```

Proof:


https://user-images.githubusercontent.com/841956/173061416-2965053b-d3db-4923-90e5-ec1d0d89a643.mov



It seems to me that `queryTaxonomies` can also be optimized a bit to not wait for all terms to be loaded before updating the state. Instead, the state can be updated multiple times, after each successful loading.

packages/design-system/src/components/datalist/searchInput.js
mwritter: I can look into that.  It should be working in Chrome when the `input` doesn't have focus but when an `input` is in focus Chrome disregards the `text-overflow`.  Firefox seems to allow it and I haven't tested it on Safari.

sblinde (approved):
👏👏👏👏

Well done. Gave the code a quick check again, and did a quick QA on the qa instance. Everything at first glance is working well so far (w/ just categories and categories + custom taxonomies).

packages/design-system/src/components/datalist/searchInput.js
mwritter: It looks like theres no way to do this for focused inputs in Chrome. But as I mentioned it works when the search is not in focus for Chrome, and it works on Firefox for both in focus and not.

https://user-images.githubusercontent.com/14057928/173126206-d5d1a210-6c8f-438a-9266-9a7d32ef87aa.mp4

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
```diff
@@ -0,0 +1,154 @@
...
+    [getTaxonomyTerms]
+  );
+
+  /**
+   * Query all the taxonomies.
+   * This should only be needed once get all the taxonomies and set individual term data.
+   *
+   * @see queryTaxonomyTerms
+   */
+  const queryTaxonomies = useCallback(async () => {
```
swissspidy, line 105: The whole logic with the termPromises ref etc. is a bit hard to read/follow. I suggest creating a follow-up ticket to improve readability a bit.

spacedmonkey:
Screenshot 2022-06-13 at 11 05 56
What should happen in if there are lots of taxonomies registered? How should it stack / clip?

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
```diff
@@ -0,0 +1,154 @@
...
+   * Maps the fetched taxonomy data back to parent taxonomy.
+   *
+   * @return {void}
+   */
+  const _setTaxonomiesTermData = useCallback(async () => {
+    const fetched = await Promise.all(termPromises.current);
+    const taxonomyObject = {};
+    for (const arr of fetched) {
+      // grab the first elements 'taxonomy' which should map to the parents 'slug'
+      const key = arr.at(0)?.taxonomy;
```
spacedmonkey:
```suggestion
      const key = arr.[0]?.taxonomy;
```

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
```diff
@@ -0,0 +1,154 @@
...
+    }
+    setTaxonomies((current) => {
+      return current.map((c) => {
+        const data = taxonomyObject[c.slug] || [];
+        return {
+          ...c,
+          data,
+        };
+      });
+    });
```
spacedmonkey:
```suggestion
    setTaxonomies((current) => current.map((c) => {
          ...c,
          data: taxonomyObject[c.slug] || [],
        }
    });
```

Arrow function do no need returns right? 

Also, current, isnt a great name, how about `taxonomies`?

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
swissspidy: What's wrong with using [`at()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at)?

packages/design-system/src/components/datalist/searchInput.js
```diff
@@ -107,6 +118,7 @@ const SearchInput = forwardRef(function SearchInput(
     value,
     onChange,
     focusFontListFirstOption = noop,
+    placeholder,
```
spacedmonkey:
```suggestion
    placeholder = __('Search', 'web-stories'),
```

packages/design-system/src/components/datalist/searchInput.js
```diff
@@ -133,7 +145,7 @@ const SearchInput = forwardRef(function SearchInput(
         aria-expanded={isExpanded}
         value={value}
         onKeyDown={handleKeyPress}
-        placeholder={__('Search', 'web-stories')}
+        placeholder={placeholder || __('Search', 'web-stories')}
```
spacedmonkey:
```suggestion
        placeholder={placeholder}
```

spacedmonkey:
Created a little PR to ensure that only taxnomies that show_ui flag is set to true - https://github.com/GoogleForCreators/web-stories-wp/pull/11678

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
spacedmonkey: Personal preference, I guess.

mwritter:
> What should happen in if there are lots of taxonomies registered? How should it stack / clip?

I believe we need some design input on this one @sblinde.

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
mwritter: I can update this to use array notation instead of using the array function to be more consistent with the code base.

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
mwritter: `taxonomies` is already declared in the upper scope, I can rename this to be more readable.

sblinde:
> > What should happen in if there are lots of taxonomies registered? How should it stack / clip?
> 
> I believe we need some design input on this one @sblinde.

@mwritter Agreed! I'll make sure it's part of the improvements epic ✅

packages/dashboard/src/app/views/myStories/filters/taxonomy/useTaxonomyFilters.js
mwritter: 🤔 I agree, the logic is a bit hard to read, I see a way to make a small change that'll simplify it 😅

PR #11627, opened 2022-06-06
Floating menu: fix visibility and positioning

### LINKED ISSUE #11377: Floating toolbar is sometimes invisible to users, but still exists
## Bug Description

We have experienced a number of occasions where the floating toolbar is not presented to user on element selection, but it's still on page, as some of the controls within can still be interacted with.

Here is video of me encountering this yesterday
https://user-images.githubusercontent.com/58738530/165594360-3c2e7f9d-f3ad-4d06-bd0e-bb6d4ce377b0.mp4

## Expected Behaviour

Floating toolbar should be presented whenever it's expected, should be visible, and should be usable.

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

merapi:
## Summary

This PR fixes several of the reported bugs regarding a disappearing or invisible but clickable floating menu.

The reports were without detailed steps to replicate, here's what I found:

1. Issues around the positioning of the FM (easily reproducible on small canvases):

https://user-images.githubusercontent.com/21036224/172141539-6bd45cb8-8067-49f3-b6c9-6dcd4807b70e.mov

FM being somewhere here:
Screenshot 2022-06-06 at 11 57 31

2. Issue when element is dragged out of canvas:

https://user-images.githubusercontent.com/21036224/172142121-6ca3d407-1da3-4364-8829-7ef8e8a75351.mov

3. FM is hidden but still clickable (because of opacity use), example when resizing text but happens also in other cases:

https://user-images.githubusercontent.com/21036224/172142586-ba612526-67d6-4b4d-a88f-6d9506387af1.mov

4. Other issues that are not easily reproducible.

I think most of the issues with invisible FM came from `isAnythingTransforming` being a little buggy - it needs to be reset manually after every kind of transformation, thus I added the general fix onMouseDown (didn't saw any downsides).

## Testing Instructions

1. Test 3 cases from the Summary.
2. Try to break the FM in any other way.

## Checklist

---

Fixes #11377

### DIFF
--- a/packages/design-system/src/components/contextMenu/styled.js
+++ b/packages/design-system/src/components/contextMenu/styled.js
@@ -24,12 +24,9 @@ import styled from 'styled-components';
 export const POPOVER_Z_INDEX = 9991;
 
 export const Popover = styled.div`
-  --translate-x: calc(var(--delta-x, 0) * 1px);
-  --translate-y: calc(var(--delta-y, 0) * 1px);
   display: ${({ isOpen }) => (isOpen ? 'block' : 'none')};
   position: ${({ isInline }) => (isInline ? 'relative' : 'absolute')};
   z-index: ${({ popoverZIndex }) => popoverZIndex};
-  transform: translate(var(--translate-x), var(--translate-y));
 `;
 Popover.defaultProps = {
   popoverZIndex: POPOVER_Z_INDEX,
--- a/packages/design-system/src/components/contextMenu/styled.js
+++ b/packages/design-system/src/components/contextMenu/styled.js
@@ -17,7 +17,6 @@
  * External dependencies
  */
 import PropTypes from 'prop-types';
-import { useCallback } from '@googleforcreators/react';
 import styled from 'styled-components';
 
 // z-index needs to be higher than the wordpress toolbar z-index: 9989.
@@ -52,38 +51,7 @@ export const Shadow = styled.div`
 `;
 
 export function SmartPopover(props) {
-  const setRef = useCallback(
-    (node) => {
-      if (!node) {
-        return;
-      }
-
-      const boundingBox = node.getBoundingClientRect();
-      const max = {
-        x: window.innerWidth - boundingBox.width,
-        y: window.innerHeight - boundingBox.height,
-      };
-
-      // This is modeling the behavior of chrome:
-      // - menu becomes right justified if not enough space horizontally
-      // - menu sticks to bottom if not enough space vertically
-      const horizontalEdgeCondition = props.isRTL
-        ? boundingBox.x < 0
-        : max.x < boundingBox.x;
-      const horizontalEdgeTransform = props.isRTL
-        ? boundingBox.width
-        : -boundingBox.width;
-      const delta = {
-        x: horizontalEdgeCondition ? horizontalEdgeTransform : 0,
-        y: Math.min(0, max.y - boundingBox.y),
-      };
-      node.style.setProperty('--delta-x', delta.x);
-      node.style.setProperty('--delta-y', delta.y);
-    },
-    [props.isRTL]
-  );
-
-  return props.isOpen ? <Popover ref={setRef} {...props} /> : null;
+  return props.isOpen ? <Popover {...props} /> : null;
 }
 
 SmartPopover.propTypes = {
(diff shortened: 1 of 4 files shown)


packages/design-system/src/components/contextMenu/styled.js
```diff
@@ -52,38 +51,7 @@ export const Shadow = styled.div`
...
-        y: Math.min(0, max.y - boundingBox.y),
-      };
-      node.style.setProperty('--delta-x', delta.x);
-      node.style.setProperty('--delta-y', delta.y);
-    },
-    [props.isRTL]
-  );
-
-  return props.isOpen ? <Popover ref={setRef} {...props} /> : null;
+  return props.isOpen ? <Popover {...props} /> : null;
```
merapi, line 51: This caused bug 1. and seems unnecessary, "Chrome behavior" is already done somewhere else (in `updatePosition`). From my testing removing it doesn't cause any issue elsewhere.

packages/story-editor/src/components/canvas/frameElement.js
```diff
@@ -242,10 +242,15 @@ function FrameElement({ id }) {
...
 
+  const { clearTransforms } = useTransform((state) => ({
+    clearTransforms: state.actions.clearTransforms,
+  }));
+
   const handleMouseDown = useCallback(
     (evt) => {
       if (!isSelected) {
         handleSelectElement(id, evt);
+        clearTransforms();
```
merapi, line 253: This should remove all(?) issues with `isAnythingTransforming`, clearly when you do a mouse down - nothing is transforming yet - when you start dragging then we set the transforming flag.

packages/story-editor/src/components/floatingMenu/menu.js
```diff
@@ -48,7 +48,8 @@ const MenuWrapper = styled.section`
   display: flex;
   position: absolute;
   z-index: ${Z_INDEX_FLOATING_MENU};
-  opacity: ${({ visuallyHidden }) => (visuallyHidden ? 0 : 1)};
+  visibility: ${({ visuallyHidden }) =>
+    visuallyHidden ? 'hidden' : 'visible'};
```
merapi, line 52: Fixes all issues with invisible FM but clickable buttons.

packages/design-system/src/components/contextMenu/styled.js
timarney: Wondering if the `SmartPopover` should just be removed --- it's just a Popover with props now.

merapi: We still have that `props.isOpen ?` part where we return `null` instead of just hiding the component visually via `isOpen` but it's still in the tree.

timarney (approved):

packages/design-system/src/components/contextMenu/styled.js
```diff
@@ -52,38 +48,7 @@ export const Shadow = styled.div`
 `;
 
 export function SmartPopover(props) {
```
barklund, line 50: Why are props not deconstructed here like we always do? We could just do `{ isOpen, ...rest }` - does the `Popover` ever need the `isOpen` property?

barklund (approved):
This seems to work really well.

I was able to cause the original issue to re-appear in this PR with a text field and jumping in and out of edit mode, but only once and I can't re-create it no matter how much I try, so not sure if it was actually a bug or maybe I pressed the X icon 🤷 . However, feel free to see if there might still be an issue with text edit-mode somewhere?

packages/design-system/src/components/contextMenu/styled.js
merapi: I just saw that `Popover` is exported and used somewhere else (with `isOpen`) so I didn't want to change it, thus we cannot do this (because `isOpen` is needed):
image

barklund: But `isOpen` will always be true when we use this `SmartPopover`, so it would just be:

```js
function SmartPopover({ isOpen, ...rest }) {
  return isOpen ? <Popover isOpen {...rest} /> : null;
}
```

Alternatively, you could set `isOpen` to default `true` in `Popover` (it already has a `.defaultProps` config).

merapi: Sure, it's just a detail at this point (and 101/94 chars vs 87 like we have right now 😉 ).

barklund: I just don't like having this one component not use prop destructuring like we do everywhere else. It's important to me, okay? 😄

PR #11628, opened 2022-06-06
Canvas: Support selecting element by clicking on the border

### LINKED ISSUE #5325: Allow selecting element by clicking on the outer border

timarney:
## Context

## Summary
Adjusts calculations for get `box` [`getBox()`] to accommodate for border.

Without border click
https://www.youtube.com/watch?v=D40BMjkiyEg

With border click
https://www.youtube.com/watch?v=11K02cL_8eY

## Testing Instructions

1.

No

No

No

## Checklist

---

Fixes #5325


timarney:
Tested with text as well.
Screen Shot 2022-06-06 at 9 28 05 AM

merapi:
Looks promising, found just two things to fix:

1. In shapes, the bounding box is in the middle of the border (more visible for big borders, I use 40):
image

2. Text resizing
Selected mode - ok:
image
Edit-mode - not sure if it's ok or not:
image
Started resizing (just 1px) - instant jump to big font:
image

packages/story-editor/src/components/canvas/frameElement.js
```diff
@@ -80,7 +80,7 @@ const Wrapper = styled.div`
   &:focus,
   &:active {
     outline-color: ${({ theme, hasMask }) =>
-      hasMask ? 'transparent' : theme.colors.border.selection};
+    hasMask ? 'transparent' : theme.colors.border.selection};
```
barklund: Hmm, you seem to have your browser misconfigured. Prettier definitely wants that extra white-space here, so it should be reverted.

barklund (changes requested):
This PR is missing implementing this one line from the original PR:

> Additionally, in discussion with George, we should include the border to the selection line as well.

However, you will find that this single line has a lot of consequences. It means that when you add border, the element in essence grows and that leads to many other changes. The size and position of the element as displayed and updated in the design panel updates. The element when dragged now snaps according to its outer border shape, not inner border shape. And other elements now snap to this one using its outer border shape, not inner. Especially this snapping stuff might be a bit tricky, which is why this is a pretty huge ticket to take on. And there are probably other consequences too, that I haven't considered.

timarney:
https://github.com/GoogleForCreators/web-stories-wp/pull/11628#pullrequestreview-999648851

☝️ Noting comment above as it outlines what areas are impacted by any PR trying to fix this issue.


Adding a note just as I dig around.

Looking at this again from `main`

Depending on the type of shape looks like the border is applied either fully to the outside i.e. for rectangles vs other shapes it's in the middle ... 

Screen Shot 2022-06-09 at 10 32 09 AM



Screen Shot 2022-06-09 at 11 08 30 AM

timarney:
Going to close this as I don't think this approach is going to work.

PR #11652, opened 2022-06-08
REST API: support filtering taxonomies by hierarchical attribute

swissspidy:
## Context

Based on the discussion at #11625 / https://github.com/GoogleForCreators/web-stories-wp/pull/11625#discussion_r892228319

## Summary

Adds support for filtering hierarchical and flat taxonomies:

* `/wp-json/web-stories/v1/taxonomies`: shows all taxonomies
* `/wp-json/web-stories/v1/taxonomies?hierarchical=true`: shows only hierarchical taxonomies
* `/wp-json/web-stories/v1/taxonomies?hierarchical=flat`: shows only flat taxonomies

## Testing Instructions

1.

## Checklist

---

Fixes #

### DIFF
--- a/includes/REST_API/Stories_Taxonomies_Controller.php
+++ b/includes/REST_API/Stories_Taxonomies_Controller.php
@@ -99,9 +99,16 @@ class Stories_Taxonomies_Controller extends WP_REST_Taxonomies_Controller implem
 		$registered = $this->get_collection_params();
 
 		if ( isset( $registered['type'] ) && ! empty( $request['type'] ) ) {
-			$taxonomies = get_object_taxonomies( $request['type'], 'objects' );
+			/**
+			 * Object type.
+			 *
+			 * @var string Object type.
+			 */
+			$type = $request['type'];
+
+			$taxonomies = get_object_taxonomies( $type, 'objects' );
 		} else {
-			$taxonomies = get_taxonomies( '', 'objects' );
+			$taxonomies = get_taxonomies( [], 'objects' );
 		}
 
 		if ( isset( $registered['hierarchical'], $request['hierarchical'] ) ) {
@@ -110,6 +117,11 @@ class Stories_Taxonomies_Controller extends WP_REST_Taxonomies_Controller implem
 
 		$data = [];
 
+		/**
+		 * Taxonomy.
+		 *
+		 * @var WP_Taxonomy $value
+		 */
 		foreach ( $taxonomies as $tax_type => $value ) {
 			if ( empty( $value->show_in_rest ) || ( 'edit' === $request['context'] && ! current_user_can( $value->cap->assign_terms ) ) ) {
 				continue;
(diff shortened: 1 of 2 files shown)


includes/REST_API/Stories_Taxonomies_Controller.php
```diff
@@ -83,6 +84,73 @@ public function prepare_item_for_response( $taxonomy, $request ): WP_REST_Respon
...
+	 */
+	public function get_collection_params(): array {
+		$query_params = parent::get_collection_params();
+
+		$query_params['per_page']['default'] = 100;
+
+		$query_params['hierarchical'] = [
+			'description' => __( 'Whether to show only hierarchical taxonomies.', 'web-stories' ),
+			'type'        => 'boolean',
+			'default'     => false,
```
spacedmonkey: Let's not default this.

swissspidy: Right, otherwise it just displays flat terms by default. Definitely wanna remove that. Good catch!

includes/REST_API/Stories_Taxonomies_Controller.php
```diff
@@ -83,6 +84,73 @@ public function prepare_item_for_response( $taxonomy, $request ): WP_REST_Respon
...
+		$data = [];
+
+		foreach ( $taxonomies as $tax_type => $value ) {
+			if ( empty( $value->show_in_rest ) || ( 'edit' === $request['context'] && ! current_user_can( $value->cap->assign_terms ) ) ) {
+				continue;
+			}
+
+			$tax               = $this->prepare_item_for_response( $value, $request );
+			$tax               = $this->prepare_response_for_collection( $tax );
+			$data[ $tax_type ] = $tax;
```
spacedmonkey, line 132: I hate the shape of this response. If we are doing this, can we change the shape of this.

spacedmonkey:
```suggestion
			$data[] = $tax;
```

swissspidy: This is just copied from core and the expected response format

AFAIK we also rely on this in the editor, so I would not wanna change this.

Screenshot 2022-06-08 at 18 05 05

spacedmonkey: Never mind

spacedmonkey:
This is exactly the same as what I was playing around with locally. I didn't want to have to override this core method like this, but if you want to do this, then I am happy with this change. 

Add a unit test and I will approve this change.

spacedmonkey:
@swissspidy This just need php lints to pass and small unit tests and this is good to go.

spacedmonkey:
Sharing for visibility https://github.com/GoogleForCreators/web-stories-wp/issues/11436#issuecomment-1150986600

spacedmonkey (approved):

mwritter (approved):
👍

PR #11656, opened 2022-06-09
Shopping: Improve product icon.

spacedmonkey:
## Summary

Improve layer icon for product elements. Use product image if there is one. 

## User-facing changes

Screenshot 2022-06-09 at 10 44 58

## Testing Instructions

1.

## Checklist

---

Fixes #


packages/element-library/src/product/icon.js
```diff
@@ -17,68 +17,21 @@
...
-  );
+/**
+ * Internal dependencies
+ */
+import VisibleImage from '../shared/visibleImage';
+
+function ProductLayerIcon({ element: { product } }) {
+  const productImage = product?.productImages?.[0];
+  const { url, alt } = productImage;
+  if (!url) {
```
swissspidy: I am getting this error for a product without any images:

```
TypeError: Cannot destructure property 'url' of 'productImage' as it is undefined.

    at ProductLayerIcon
```

This fixes it:

```suggestion
  const productImage = product?.productImages?.[0] || {};
  const { url, alt } = productImage;
  if (!url) {
```

packages/element-library/src/product/icon.js
```diff
@@ -17,68 +17,21 @@
...
+    return <Icons.Shopping width={21} height={21} title="" />;
```
swissspidy: Just noting that the icon is quite small:

Screenshot 2022-06-09 at 12 11 54

Though I guess it's negligible because we'll prevent actually adding products without images (See #11616)

So no action needed here for now.

swissspidy:
```suggestion
    return <Icons.Shopping width={21} height={21} aria-hidden />;
```

Better to use `aria-hidden` instead of an empty title to remove it from the accessibility tree

swissspidy (approved):

PR #11667, opened 2022-06-10
History: Reduce number of actions causing history pollution 

### LINKED ISSUE #10289: History: Investigate actions still causing history pollution
## Feature Description
There are a few actions that still cause history pollution. It would be good to investigate those and ensure that if user does an action, just one Undo would revert it. Currently, some actions require multiple clicks of "Undo".

Known actions:
- Uploading Video directly onto a Page
- Removing audio from a video
- Trimming a video

There might be more.

## Additional Context
See [GH conversation here.](https://github.com/google/web-stories-wp/pull/10240#issuecomment-1016894611)

timarney:
## Context

Currently, some actions require multiple clicks of "Undo" because of history pollution (unnecessary entries). 

See: https://github.com/GoogleForCreators/web-stories-wp/issues/10289

## Summary

This PR adds an ignore for `resource.isOptimized` and skips entries that have a `blob` url for an element.

## To-do

This PR needs testing across the various scenarios see testing Instructions

- Demo video with fix  https://www.youtube.com/watch?v=QMy05XaCDjg

## Testing Instructions

👉  Note what's happening on current main
Actions take multiple undos to revert and sometimes end up in a broken state
- https://www.youtube.com/watch?v=9VpjT2PbjDw
- https://github.com/GoogleForCreators/web-stories-wp/issues/10289#issuecomment-1148664264

<hr>

 **1) Uploading Video directly onto a Page**
Upload a video onto the stage --- undo the upload .... video should be removed from the stage with a single undo. 

**2) Trim a video**
Note the original video length.  Trim video to x amount. Undo the trim -- the video length should now be reverted to the original length with a single undo.

**3) Remove audio from a video**
Remove the audio, undo --- the undo should restore the audio with a single undo.

No

No

No

## Checklist

---

Fixes #10289

### DIFF
--- a/packages/story-editor/src/app/story/effects/useHistoryEntry.js
+++ b/packages/story-editor/src/app/story/effects/useHistoryEntry.js
@@ -25,7 +25,7 @@ import { useEffect, useRef } from '@googleforcreators/react';
  */
 import { useHistory } from '../../history';
 import deleteNestedKeys from '../utils/deleteNestedKeys';
-import skipBlobUrls from '../utils/skipBlobUrls';
+import pageContainsBlobUrl from '../utils/pageContainsBlobUrl';
 
 // Changes to these properties of elements do not create a new history entry
 // if only one (or multiple) of these properties change but nothing else changes.
@@ -112,7 +112,7 @@ function useHistoryEntry({ story, current, pages, selection, capabilities }) {
           JSON.stringify(adjustedPages) === JSON.stringify(adjustedEntryPages);
       }
 
-      if (skipBlobUrls(pages)) {
+      if (pageContainsBlobUrl(pages)) {
         skipAddingEntry = true;
       }
     }
--- a/packages/story-editor/src/app/story/effects/useHistoryEntry.js
+++ b/packages/story-editor/src/app/story/effects/useHistoryEntry.js
@@ -36,6 +36,7 @@ const ELEMENT_PROPS_TO_IGNORE = [
   'resource.isMuted',
   'resource.posterId',
   'resource.poster',
+  'resource.isOptimized',
   'resource.font.metrics',
   'resource.font.weights',
   'resource.font.variants',
@@ -103,6 +104,16 @@ function useHistoryEntry({ story, current, pages, selection, capabilities }) {
         skipAddingEntry =
           JSON.stringify(adjustedPages) === JSON.stringify(adjustedEntryPages);
       }
+
+      // skip entries that have a blob url
+      // https://github.com/GoogleForCreators/web-stories-wp/issues/10289
+      pages.map((page) => {
+        page.elements.forEach((element) => {
+          if (element?.resource?.src.includes('blob:')) {
+            skipAddingEntry = true;
+          }
+        });
+      });
     }
 
     if (!skipAddingEntry) {
--- a/packages/story-editor/src/app/story/effects/useHistoryEntry.js
+++ b/packages/story-editor/src/app/story/effects/useHistoryEntry.js
@@ -37,12 +37,18 @@ const ELEMENT_PROPS_TO_IGNORE = [
   'resource.isMuted',
   'resource.posterId',
   'resource.poster',
-  'resource.isOptimized',
   'resource.font.metrics',
   'resource.font.weights',
   'resource.font.variants',
   'resource.font.fallbacks',
   'resource.font.styles',
+  'resource.isOptimized',
+  'resource.length',
+  'resource.lengthFormatted',
+  'resource.trimData.original',
+  'resource.trimData.start',
+  'resource.trimData.end',
+  'resource.creationDate',
 ];
 
 // Record any change to core variables in history (history will know if it's a replay)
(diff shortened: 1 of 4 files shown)


packages/story-editor/src/app/story/effects/useHistoryEntry.js
```diff
@@ -103,6 +104,16 @@ function useHistoryEntry({ story, current, pages, selection, capabilities }) {
         skipAddingEntry =
           JSON.stringify(adjustedPages) === JSON.stringify(adjustedEntryPages);
       }
+
+      // skip entries that have a blob url
+      // https://github.com/GoogleForCreators/web-stories-wp/issues/10289
+      pages.map((page) => {
+        page.elements.forEach((element) => {
+          if (element?.resource?.src.includes('blob:')) {
```
swissspidy: Note that there's a reusable `isBlobURL` helper in the `media` package that can be used for this.

packages/story-editor/src/app/story/effects/useHistoryEntry.js
```diff
@@ -41,6 +42,13 @@ const ELEMENT_PROPS_TO_IGNORE = [
   'resource.font.variants',
   'resource.font.fallbacks',
   'resource.font.styles',
+  'resource.isOptimized',
+  'resource.length',
```
timarney, line 46: 'resource.length' etc.. ignored for trim action .... from initial testing I think we still get 1 entry here (which we need) when the src changes.

packages/story-editor/src/app/story/effects/useHistoryEntry.js
```diff
@@ -103,6 +111,16 @@ function useHistoryEntry({ story, current, pages, selection, capabilities }) {
...
+
+      // skip entries that have a blob url
+      // https://github.com/GoogleForCreators/web-stories-wp/issues/10289
+      pages.map((page) => {
+        page.elements.forEach((element) => {
+          if (isBlobURL(element?.resource?.src)) {
+            skipAddingEntry = true;
+          }
+        });
+      });
```
barklund: I'm not a big fan of this code residing here, but I guess it's okay for now.

I feel like the history provider shouldn't have all this domain-specific functionality (I know it had some before as well) and it would be better if it had a way of "registering" exception functions, and we'd then create the exceptions in the story provider where they belong. But I guess that's a bit too much work for this PR.

barklund (approved):
Not loving the direction where this is going with the increased interconnectedness, but it seems to solve the issue so it's okay for now.

But I'd like it if we had a follow-up issue to disconnect the history provider from the domain of web stories. We could even move it to a separate package.

packages/story-editor/src/app/story/effects/useHistoryEntry.js
```diff
@@ -103,6 +111,16 @@ function useHistoryEntry({ story, current, pages, selection, capabilities }) {
         skipAddingEntry =
           JSON.stringify(adjustedPages) === JSON.stringify(adjustedEntryPages);
       }
+
+      // skip entries that have a blob url
+      // https://github.com/GoogleForCreators/web-stories-wp/issues/10289
+      pages.map((page) => {
+        page.elements.forEach((element) => {
+          if (isBlobURL(element?.resource?.src)) {
```
swissspidy: `resource.poster` can potentially also be a blob URL, would be good to account for that as well.

packages/story-editor/src/app/story/effects/useHistoryEntry.js
timarney: Makes sense --- can look at that in a follow up PR.  For now might move this out to more of a util function.

packages/story-editor/src/app/story/effects/useHistoryEntry.js
timarney: Added a check for that -- it's also ignored https://github.com/GoogleForCreators/web-stories-wp/blob/c77193a33c18aa43bfe614fa364df17a60911243/packages/story-editor/src/app/story/effects/useHistoryEntry.js#L39

packages/story-editor/src/app/story/utils/skipBlobUrls.js
```diff
@@ -0,0 +1,39 @@
...
+  // skip entries that have a blob url
+  // https://github.com/GoogleForCreators/web-stories-wp/issues/10289
+  let skip = false;
+  pages.map((page) => {
+    page.elements.forEach((element) => {
+      if (
+        isBlobURL(element?.resource?.src) ||
+        isBlobURL(element?.resource?.poster)
+      ) {
+        skip = true;
```
swissspidy: Just `return true` early here and save a few iterations.

Then you don't even need the `skip` variable, since at the end of the function you can just `return false`

packages/story-editor/src/app/story/utils/pageContainsBlobUrl.js
```diff
@@ -0,0 +1,38 @@
...
+        isBlobURL(element?.resource?.src) ||
+        isBlobURL(element?.resource?.poster)
+      ) {
+        return true;
+      }
+    });
+  });
+
+  return result.includes(true);
+}
```
swissspidy: You'll probably want this (needs indentation fix)

```suggestion
  // skip entries that have a blob url
  // https://github.com/GoogleForCreators/web-stories-wp/issues/10289
  return pages.some((page) => 
   page.elements.some((element) => 
     (
        isBlobURL(element?.resource?.src) ||
        isBlobURL(element?.resource?.poster)
      )
    });
  });
}
```

swissspidy:
Let's add some solid QA instructions for Felipe, and then a new ticket for Morten's suggestion 👍

timarney:
@swissspidy @barklund I updated the test instructions and added https://github.com/GoogleForCreators/web-stories-wp/issues/11698 .

Issue #9643, opened 2021-11-05
Multipage Canvas POC


miina:
## Feature Description

Edit: See https://github.com/GoogleForCreators/web-stories-wp/issues/9643#issuecomment-1064304530 for the latest update. Design TBD.

Important: UX prefers the feature to be released once it's fully ready, there won't be originally considered MVP etc. even though we'll most likely implement it in multiple steps. Once we know better, we can convert this issue into an epic.

See [Figma](https://www.figma.com/file/dc9xrVlRhQq4KTE9WQzU1R/_Editor---Sprint-Feature-Level-Designs_?node-id=10548%3A453241)

Note: this feature has been blocked for some time due to potential hit on performance. It has been brought up again by UX as high priority so we should take another look into this to find a way how to implement it.

Some options for unblocking this:
- Improve the Page performance (we should do this anyway, also see #8255). EDIT: Note that various performance improvements have been done by Pea pod meanwhile, some of those also influence the Page rendering time.
- Perhaps we can use the images generated anyway for the page carousel thumbnails. The images are tiny though, and when loading the first time, these are not generated yet, so this needs some looking into. EDIT: See #9749 

<img src="https://user-images.githubusercontent.com/637548/173100843-3114a49a-c01e-428c-b42d-f3d7f549dc57.png" />

miina:
UX brought this issue up as quite important and requested.
We should take another look at this, most likely the issue will be done in two steps:
- Displaying non-interactive pages (able to navigate by clicking but not able to drag anything to the next page etc)
- Displaying partially interactive pages: possible to drag an element from the active page to prev-next pages.

First we need to fix the thumbnail generation to only generate one image per page, we could potentially reuse those for the prev/next pages.

miina:
Update from design sync on the 9th of March

The ideal option from Amy:

Make the entire story interactive and side-to-side scrollable

Implications:
- Users will be able to see how objects line up from canvas to canvas, and potentially have point guides in place for exactness
- Users can copy, paste, and drag objects between all canvases
- Users can edit any slide anywhere on screen by scrolling to it and putting it wherever they feel comfortable editing it, even if partial view
- Scrolling would work somewhat like Canva
- There will be no “main canvas” indicators in the workspace, but perhaps an indicator of the canvas you’re working on at any given time on click or nav (for accessibility?)
- Each canvas would have a canvas menu, but quick menu doesn’t show up till you click on it

bmattb:
[...]

PR #11669, opened 2022-06-10
POC: Editor: Multipage Canvas

barklund:
## Context

This adds a preview of the pages before and after the current ones to either side of the canvas, if there is enough space. It adds as many pages, as there is room for.

## To-do

- [x] Which size should they pages be and with what margin?
- [x] Should the pages be active in any way? Can you click them? Can you drag stuff to them? Or from them?

## User-facing changes

| Window height | Screenshot |
|-|-|
| Narrow | Screen Shot 2022-06-10 at 17 32 42 |
| Medium | Screen Shot 2022-06-10 at 17 31 52 |
| Tall | Screen Shot 2022-06-10 at 17 32 15 |

It even works in RTL:
Screen Shot 2022-06-10 at 17 40 17

## Testing Instructions

0. Enable _"Context pages"_ feature flag
1. Add multiple pages to a story (preferably with some content
2. Switch page a few times
3. Observe the previous and future pages in the margins of the current page
4. Try different browser sizes to see the changes

## Checklist

---

Partially addresses #9643


timarney:
Nice -- overall :)

Seeing some issues with previous and next screen previews --- likely separate from this PR as the "thumbnail" view has same issue. 
Screen Shot 2022-06-10 at 1 33 19 PM

sblinde (approved):
🐝 

Code looks solid. Do agree: there are some pre-existing issues w/ the thumbnail view of pages where the `Fit` button overlaps the arrows when there's a lot of pages, but I know that's outside of this scope and body of work.

swissspidy:
Are the pages basically all just screenshots? What is the performance impact of this? Can we measure it?

It might be best to put this behind a feature flag until that is more clear.

barklund:
> Are the pages basically all just screenshots? What is the performance impact of this? Can we measure it?

They are all instances of `PagePreview`, so they're complex HTML until they've been snapshotted by the snapshot logic that we have in place. There might very well be a performance impact of this, as we previously only snapshot tiny thumbnails in the carousel, but now snapshot these much larger canvases. It will have an impact on both computation time for generating these snapshots as well as memory used for storing them. I have not putting any tracking of that in at this time.

> It might be best to put this behind a feature flag until that is more clear.

That sounds like a good idea!

merapi:
I'm trying to make it work on QA so the design pod can review it, right now there is no build ZIP.

packages/story-editor/src/components/canvas/extraPages.js
```diff
@@ -0,0 +1,161 @@
...
+        listWidth={listWidth}
+        extraPageHeight={extraPageHeight}
+      >
+        {pagesToShow.map((pageNum) => (
+          <ExtraPage key={pageNum} extraPageWidth={extraPageWidth}>
+            <ExtraPagePreview
+              page={pages[pageNum]}
+              onClick={clickPage(pages[pageNum].id)}
+              aria-label={sprintf(
+                /* translators: %s: Page number */
```
swissspidy:
```suggestion
                /* translators: %s: page number. */
```

To address the i18n lint warning

> Warning: The string "Go to page %s" has 2 different translator comments.

merapi:
I can reproduce the bug from 00:13 locally and on QA.

https://user-images.githubusercontent.com/21036224/175024042-4ec2c652-2610-414e-89d4-7af134c137ce.mov

Other than that we are waiting for the design pod review (page size/margins etc).

barklund:
> I can reproduce the bug from 00:13 locally and on QA.

Whoa, that's a weird bug! I'll investigate, thanks for catching it!

barklund:
Approved by UX:
- Size, padding, opacity, and hover looks good

To do after UX review:
- [ ] Last page never shows (some off-by-one error somewhere)
- [ ] Resize all page previews to also show non-safe space (includes page carousel and grid view, should probably be new ticket)
- [ ] Don't show hover state when dragging from canvas to neighbor page
- [ ] Don't allow neighbor pages to be draggable
- [ ] Measure impact on low-end devices
- [ ] Fade pages even more further away, show max 4 either direction

Issue #11576, opened 2022-05-26
Refactor layer panel


barklund:
## Feature Description

The layer panel should be refactored and moved to a more appropriate place in the folder structure.

Goals of this issue include:

- Make sure the `layer/layer.js` file is much smaller. Currently it sits at 500+ lines and is quite impossible to work with. There are also several bits of duplicated logic in there.
- Move the whole layer folder to a more appropriate location. The layer panel is no longer associated with the design panel nor the sidebar, so it should be decoupled from these in the folder structure.

PR #11676, opened 2022-06-13
Layer naming/grouping/locking integration

### LINKED ISSUE #11566: Layer Grouping&Locking integration
## Feature Description

Discussed with Amy:
You can only lock/unlock the whole group (folder). Layers inherit the lock state from the group.

### LINKED ISSUE #11567: Layer Naming&Grouping integration

merapi:
## To-do

- [x] Fix layout when renaming a group
- [x] Fix DND layers in/out of a group

## Testing Instructions

1.

## Checklist

---

Fixes #11566
Fixes #11567


packages/story-editor/src/app/rightClickMenu/hooks/useElementActions.js
```diff
@@ -110,7 +121,30 @@ const useElementActions = () => {
...
+    for (const [index, element] of selectedElements.reverse().entries()) {
+      const position = elements.findIndex(({ id }) => id === element?.id);
+      if (position !== elementFromGroupWithHighestPosition) {
+        const newPosition = elementFromGroupWithHighestPosition - index;
+        arrangeElement({
+          elementId: element.id,
+          position: newPosition,
+        });
+      }
+    }
```
barklund, line 141: Doesn't this cause multiple history entries, or are they actually grouped because we only update history on the next js frame in a useEffect?

Also, I feel this should go in the story reducer directly (maybe as `arrangeElements`), then we can just invoke a single method here with all the ID's and then we can also better test this functionality.

Actually, thinking about it, what if we moved all this functionality to the story reducer - `groupElements`, `ungroupElements`, `renameGroup` - would that not make things easier?

barklund: Oh, you did partially move it to the reducer, I see now. Maybe we could move the elements logic in there too - it wouldn't be too hard to do. `addGroup` could always assume that you are grouping the selected elements and only those. `deleteGroup` could similarly also delete the groups referenced by the selected elements and only those.

packages/story-editor/src/app/story/useStoryReducer/reducers/addGroup.js
```diff
@@ -0,0 +1,54 @@
...
+    ...state.pages.slice(0, pageIndex),
+    newPage,
+    ...state.pages.slice(pageIndex + 1),
+  ];
+
+  return {
+    ...state,
+    pages: newPages,
+  };
+}
```
barklund: Having played with [`immer`](https://github.com/immerjs/immer) in a hobby-project recently, I really don't like all this work anymore. This entire snippet could be simplified to just:

```js
function addGroup(state, { groupId, name }) {
  return produce(state, draft => {
    const page = draft.pages.find(({ id }) => id === draft.current);
    if (!page.groups) page.groups = {};
    page.groups[groupId] = { name, isLocked: false };
  });
}
```

That's sooooo much cleaner. We should create a code quality ticket about replacing the story reducer with immer (we already have great tests for it, so should be easy to verify that everything works).

But that's also a story for another day.

packages/story-editor/src/components/panels/design/layer/layer.js
```diff
@@ -364,8 +363,13 @@ function Layer({ element }) {
   });
 
   const layerId = `layer-${element.id}`;
+  const isNested = element.groupId;
 
-  const lockTitle = element.isLocked
+  const lockTitle = isNested
+    ? element.isLocked
+      ? __('Group is Locked', 'web-stories')
+      : __('Group is Unlocked', 'web-stories')
```
barklund, line 381: This seems a bit misleading - there's a locked icon, but the tooltip says "Group is Unlocked":

Screen Shot 2022-06-13 at 12 04 49

Maybe these layers inside a group should use the regular lock/unlock icons, but just have them at .5 opacity to indicate they cannot be individually toggled instead of using this filled lock icon?

packages/story-editor/src/components/panels/design/layer/reorderableGroup.js
```diff
@@ -0,0 +1,62 @@
...
+import { LAYER_HEIGHT } from './constants';
+import Group from './group';
+
+const LayerSeparator = styled(ReorderableSeparator)`
+  height: ${LAYER_HEIGHT}px;
+  margin: -${LAYER_HEIGHT / 2}px 0;
+  padding: ${LAYER_HEIGHT / 2}px 0;
+`;
+
+const ReorderableGroup = memo(function ReorderableLayer({
```
barklund:
```suggestion
const ReorderableGroup = memo(function ReorderableGroup({
```

packages/story-editor/src/components/panels/design/layer/useGroupSelection.js
```diff
@@ -0,0 +1,66 @@
...
+      toggleLayer({
+        elementId: groupLayers[0].id,
+        metaKey: evt.metaKey,
+        shiftKey: evt.shiftKey,
+        withLinked: true,
+      });
+
+      const isDoubleClick = evt.detail === 2;
+      if (isDoubleClick) {
+        setRenamableLayer({ elementId: groupId });
```
barklund, line 58: Uh, not a big fan of this overloading of `elementId` to mean `groupId` under some circumstances. Could we either just straight use `groupId` or revert to just "id" and then have a boolean next to it indicating group or element? But this can be followup ticket too.

packages/story-editor/src/components/panels/design/layer/group.js
```diff
@@ -0,0 +1,512 @@
+/*
```
barklund, line 1: Woah, a brand new component with 512 lines? I assume this has a ton of overlaps with the regular layer component, but it's left as-is because we already have #11576 to deal with it?

packages/story-editor/src/components/panels/design/layer/group.js
```diff
@@ -0,0 +1,512 @@
...
+  };
+
+  return (
+    <LayerContainer>
+      {isRenameable && isLayerNamingEnabled ? (
+        <LayerInputWrapper>
+          <GroupIconsWrapper>
+            <Icons.ChevronDown />
+            <Icons.Group />
+          </GroupIconsWrapper>
```
barklund: The layout here is a bit tight. The group name needs some spacing to the icon, IMHO:

Screen Shot 2022-06-13 at 12 31 58

barklund (approved):
In general, this works really well. There's a bunch of stuff I'd like to have cleaned up, but we can do that in followups post-release (or at least post-bash).

packages/story-editor/src/components/panels/design/layer/group.js
merapi: Fixed by @timarney 
image

timarney:
@merapi 

For renaming I think there's a bug.

- Add a layer (layer has default name)
- Rename a layer ('my layer')
- Undo the rename (layer has default name)
- 2x click to rename again --- editable text now shows previous rename  ('my layer')

merapi:
> @merapi
> 
> For renaming I think there's a bug.
> 
> * Add a layer (layer has default name)
> * Rename a layer ('my layer')
> * Undo the rename (layer has default name)
> * 2x click to rename again --- editable text now shows previous rename  ('my layer')

@timarney Good find, I can reproduce it on the main branch. We should either create a bug ticket or fix it on the spot (if looks easy).

packages/story-editor/src/app/rightClickMenu/menus/groupMenu.js
```diff
@@ -0,0 +1,63 @@
...
+    handleDuplicateSelectedElements,
+    //handleGroupSelectedElements,
+    handleUngroupSelectedElements,
+  } = useElementActions();
+  const isLayerGroupingEnabled = useFeature('layerGrouping');
+  const { selectedElements } = useStory(({ state }) => ({
+    selectedElements: state.selectedElements,
+  }));
+  const isGroupSelected = selectedElements.some((el) => el.groupId);
+  // const isOnlyGroupSelected = selectedElements.every(
```
timarney, line 41: Should be removed

packages/story-editor/src/app/rightClickMenu/menus/groupMenu.js
```diff
@@ -0,0 +1,63 @@
...
+ * Internal dependencies
+ */
+import { RIGHT_CLICK_MENU_LABELS } from '../constants';
+import { useElementActions } from '../hooks';
+import { useStory } from '../..';
+
+function GroupMenu() {
+  const {
+    handleDuplicateSelectedElements,
+    //handleGroupSelectedElements,
```
timarney, line 33: Should be removed

timarney:
> > @merapi
> > For renaming I think there's a bug.
> > 
> > * Add a layer (layer has default name)
> > * Rename a layer ('my layer')
> > * Undo the rename (layer has default name)
> > * 2x click to rename again --- editable text now shows previous rename  ('my layer')
> 
> @timarney Good find, I can reproduce it on the main branch. We should either create a bug ticket or fix it on the spot (if looks easy, this one might not be super easy/one-liner).

Added https://github.com/GoogleForCreators/web-stories-wp/issues/11686

timarney (approved):
Approving - just the lint + commented code cleanup outstanding.

packages/story-editor/src/components/panels/design/layer/group.js
```diff
@@ -157,7 +157,36 @@ const GroupIconsWrapper = styled.div`
   display: flex;
   align-items: center;
   justify-content: flex-start;
-  min-width: 36px;
+  width: 48px;
```
timarney, line 160: See: https://github.com/GoogleForCreators/web-stories-wp/issues/11713#issuecomment-1156604524

packages/story-editor/src/utils/generateGroupName.js
```diff
@@ -0,0 +1,54 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * External dependencies
+ */
+import { __, sprintf } from '@googleforcreators/i18n';
+
+export function getNextGroupNumber(groups) {
```
timarney, line 22: Added to cover https://github.com/GoogleForCreators/web-stories-wp/issues/11714

packages/story-editor/src/app/rightClickMenu/hooks/useElementActions.js
merapi: Note about `addGroup/updateGroupById/deleteGroupById` - they only handle the `groups` object, so you first need to create a group (generate new UUID) and then add elements to that group (assign groupId). We use `addGroup` in other places where it's not strictly connected to grouping selected elements, e.g. when pasting a group into a new story - we first need to create/modify the `groups` object.

packages/story-editor/src/components/panels/design/layer/group.js
merapi: We can for sure unify the styles part, but if you compare those two files now, the functionality is quite different so there would be a lot of branching in one file.

PR #11677, opened 2022-06-13
On-Canvas Media Recording

### LINKED ISSUE #11077: Media Recording: Feature Detection
## Feature Description

As part of the media recording feature we need more robust feature detection.

All our supported browsers support `MediaRecorder`, but the list of supported mime types is quite different.

At first glance:

* Safari on macOS: only supports `audio/mp4` and `video/mp4`, with codecs such as VP8, VP9, AV1, H.264, H.265, OPUS
* Firefox on macOS: only supports `audio/webm`, `video/webm` and `audio/ogg`, with codecs such as VP8, VP9, AV1, H.264, H.265, OPUS
* Chrome on macOS: only supports `audio/webm`, `video/webm` and `video/x-matroska`, with codecs such as VP8, VP9, AV1, H.264, H.265, OPUS  
  See https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/fast/mediarecorder/MediaRecorder-isTypeSupported.html / https://developer.chrome.com/blog/mediarecorder/ for details

We should use [`MediaRecorder.isTypeSupported`](https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder/isTypeSupported) to determine which mime types to use. Perhaps ideally with the `H.264` codec, but not specifying the codec should also work.

## Additional Context

https://cconcolato.github.io/media-mime-support/
https://stackoverflow.com/a/64656254/1185292

### LINKED ISSUE #11269: GIF Recording: Initial Implementation
## Feature Description

This is very similar to the existing video and image recording POC that we have.

The goal for this feature is to upload and treat the recorded video as a "GIF" (aka a looping, muted video).

A lot of the functionality for this already exists in the editor, so it's mostly a matter of implementing the UI and putting the pieces together.

The MVP solution here could be to display a checkbox/toggle/switch next to the preview after stopping recording. Maybe something like this:

Screenshot 2022-04-14 at 14 32 09

**Note:** exact wording and design still TBD

When choosing GIF, the preview video is muted & looping. Then, when clicking on "Insert", we need to:

1. Insert the video as a "GIF" element to the canvas
2. Upload the video to WP and mark it as a GIF and as muted

This is where we need to connect the dots.

The current upload flow is basically:

Media recording dialog -> `useUploadWithPreview` -> `useUploadMedia` (calls `getResourceFromLocalFile()`) -> `useMediaUploadQueue`

A good start seems to be to pass `additionalData = { mediaSource: 'gif-conversion', isMuted: true } to `uploadWithPreview()`.

Then, we need to ensure the element inserted by `onUploadStart` / `insertElement` has the correct type too.

So ideally we can somehow override the resource type after calling `getResourceFromLocalFile`.

Haven't looked closely into it, so needs some more digging.

Needless to say that tests would be important for this change.

### LINKED ISSUE #11542: Media Recording: Move on canvas
## Feature Description

The initial media recording prototype was built in a dialog. We now want to bring the media recording capabilities directly to the canvas. It will somewhat resemble the video trimmer feature.

There is **a lot** going on in the current Figma designs as depicted below, showing plenty of functionality that won't be in V1 of this feature.

As a start, we want to port over the existing media recording capability that we have to the canvas.

In essence that means:

1. Clicking on the recording button in the sidebar will enable "recording mode" on the canvas
2. In recording mode, a start/stop/insert/retry button is displayed below the canvas
3. A timer is displayed above the canvas, counting the seconds since recording started / how long the video already is
4. A menu on the right side of the canvas (which should like the quick actions menu) allows a few actions:
	1. "X" will cancel recording and end recording mode 
	2. A gear icon button that opens a dialog for changing the camera/microphone
	3. If easily possible: a trim button that allows trimming the recorded video before it is uploaded to the server

(We can update this section as progress is made on the UX side)

Previous Slack discussion: https://xwp.slack.com/archives/C0135V8LYJE/p1652884370376849

Current design on Figma: https://www.figma.com/file/ZrO763yvMWJjYeAPsRnNfW/_Editor---Sprint-Feature-Level-Designs_2022?node-id=661%3A24394

Current mock (still WIP)
Screenshot 2022-05-18 at 15 31 33

### LINKED ISSUE #11568: Unable to replay videos in the editor once you have changed page
## Bug Description

Videos in the editor can't be replayed after being added to a page if you have then moved away from that page, back to another page and then back to the original once more. When first adding a video to a page you can play and stop the video by clicking on it. 

If you then navigate to another page or create a new page and come back to that page you can no longer play the video. If you hover over the video you just get the "**Stop**" icon which when clicked does nothing.

**Recording of this experience:**

https://user-images.githubusercontent.com/91896826/170304152-cb57c538-a3b9-440d-967e-3fb87ad4bcb7.mp4

This isn't a great experience for the user of the editor as the only other way to view these videos again is in the preview mode. When editing they should be able to replay the videos.

This was brought to our attention in the support forum in this topic: https://wordpress.org/support/topic/cant-get-individual-slides-to-play/#post-15678082

## Expected Behaviour

You should be able to navigate away from a page with a video in the editor and back to it and still be able to replay that video to see how if runs.

## Steps to Reproduce

1. Create a new story and add a video to page 2. You can stop and replay the video while you remain on this page.
2. Add a new page or move to an existing page.
3. Go back to page 2 where you added the video and attempt to play the video by clicking on it.

## Screenshots

Recording attached above.

## Additional Context

- Plugin Version: Tested on 1.21.0 & 1.20.0 and recreated on both.
- Browser: Recreated the issue on Chrome, FireFox and Edge.
- Console errors: Not seeing any console errors in dev tools when reproducing the issue.

### LINKED ISSUE #11671: Media Recording: Telemetry
## Feature Description

Tracking events such as:

* Opening the recording dialog
* Cancelling/closing the dialog
* Changing webcam / microphone
* Muting
* Retry
* Image capture
* Inserting -> passing length
* Toggling gif/video mode

swissspidy:
## Context

The initial media recording prototype was built in a dialog. We now want to bring the media recording capabilities directly to the canvas. It will somewhat resemble the video trimmer feature.

## Summary

This PR completely overhauls the media recording flow, which is now displayed right on the canvas area.

It also touches a lot of other areas, most changes of which are explained in the next section.

## Relevant Technical Choices

**Media Source**

Adds a new `recording` media source term to the media source taxonomy so that we can later easily identify all images and videos stemming from media recording this way.

Hat tip to Jonny for the suggestion!

**Context Menu**

In `packages/design-system/src/components/contextMenu/components/icon.js`, uses `display: flex` for `IconWrapper` to allow using icons which are smaller than 32x32 while keeping them centered within the wrapper.

`flex: 1;` is specifically needed for Safari

**Quick Actions**

Adds new quick action menu items that are only shown in media recording mode. These actions are:

* Close (to exit media recording mode)
* Options (to open the settings modal to change the webcam and microphone)
* Mute (to enable/disable audio recording)

**Fixing `postProcessingResource` (poster generation, blurhash retrieval, base color detection**

During testing I noticed that, after inserting a recorded video/image, several HTTP calls for updating base color and blurhash were made for the same image. Ideally only 1 request is made.

This was because of two reasons:

First, in some places, `updateBaseColor` was called by directly importing `useDetectBaseColor`, circumventing all the logic in the media provider which checks whether a resource has already been processed.

This was fixed by moving the processing/processed checks directly to the hook as well as by only exposing it through the media provider.

So if you wanna use `updateBaseColor`, you should get it via `useLocalMedia()`

Second, after successful upload, we call the `onUploadSuccess` callback (which typically calls `postProcessingResource`) twice—once for the final resource ID and once for the resource with `previousResourceId` (the one that was temporarily on canvas until we got the final ID from the database). This caused base color and BlurHash generation to run twice for a given resource, because the `resource.id` was identical in this case.

This was addressed by wrapping `postProcessingResourceCalls in `if (id === resource.id) { postProcessingResource(resource); }`.

See 0f2fa36ff650b097acde5d35fcc2a2f93834f311 for this change in particular.

**Upload video as GIF**

Until now, we've only treated videos from the WP media library as GIFs if they were converted using FFmpeg and had the `gif-conversion` media source attached.

But now we need to flag videos as GIFs also when toggling this option in the media recording flow.

To make this work, this PR adds a new `web_stories_is_gif` meta key (in the `Is_Meta` class)

`useUploadMedia` is also updated to convert video resources to GIF resources if needed.

`getResourceFromAttachment` is updated to check for this post meta value in addition to the media source.

**Require Video Optimization**

Originally I hoped that we could just use whatever video format we get by the browser for media recording. However, none of the browsers seem to support a single codec that is understood by all of them.

To address this, media recording requires video optimization to be enabled, so that we can convert all recorded videos to MP4 before upload.

Note: technically video optimization is not needed for image capture. So maybe in the future we can only require it for video recording but not for image capture.

**Insert element as background**

Updates `useInsertElement` to support adding an element as the background element.

## To-do

In the future we'll want to support video trimming capabilities as part of the editing flow, though that will require quite a bit of refactoring of the video trimmer component, which currently expects to be used for actual story elements.

See #11685

## User-facing changes

[Demo Video](https://drive.google.com/file/d/17SvvWMU7XXG8lREkQffEQSj_TazFXw6c/view?usp=sharing)

## Testing Instructions

**Video Play/Pause Button**

1. Create a new story and add a video to page
2. You should be able to play/pause it using the button.
3. Add a new page or move to an existing page.
4. Go back to page 2 where you added the video and attempt to play the video by clicking on it.
5. You should be able to play/pause it using the button

**Experiment disabled**

Verify that media recording logic is not available when the media recording experiment is disabled

**Video optimization disabled**

Verify that media recording logic is not available when video optimization is disabled in the settings

**General flow**

1. Click on the "Record Video" icon to follow the media recording flow.
5. Notable things to test:
	* Cross-browser support
	* Usage of external microphone/camera
	* Error handling (e.g. when not granting permissions)
	* Muting button
	* GIF mode: videos should loop, have no audio and should be displayed in the media library without duration info
	* Image capture
	* When you close the media recording UI, the webcam should no longer be active (i.e. on a MacBook there should be no more green indicator light next to the camera)
	* When you switch camera/microphone from the settings dialog and close it, the change should take effect
	* Recorded videos should not trigger the video optimization checklist item

**Worth noting**

* In some browsers like Chrome you might see cameras and microphones twice. This is **expected**, since 1 device refers to your operating system's default device.

No

Tracking various new events related to media recording (e.g. changing settings, inserting recorded media, etc.)

No

## Checklist

---

Fixes #11542
Fixes #11269
Fixes #11568
Fixes #11671
Fixes #11077


packages/design-system/src/components/contextMenu/components/icon.js
```diff
@@ -26,6 +26,13 @@ import { VisuallyHidden } from '../../visuallyHidden';
 const IconWrapper = styled.div`
   width: 32px;
   height: 32px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+
+  > * {
+    flex: 1;
+  }
```
swissspidy, line 35: This allows using icons that are smaller than 32x32 while keeping them centered within `IconWrapper`.

packages/story-editor/src/components/canvas/layout.js
```diff
@@ -140,6 +141,15 @@ const PageAreaContainer = styled(Area).attrs({
     `}
 `;
 
+const PageTitleContainer = styled(Area).attrs({
```
swissspidy, line 144: This new grid area is used for the duration indicator during video recording

spacedmonkey:
Had a quick play with this and it looks amazing!

Can I recommend one thing? Add a record option to the right click menu.
Screenshot 2022-06-13 at 12 08 39

packages/story-editor/src/components/mediaRecording/stories/permissionsDialog.js
```diff
@@ -0,0 +1,41 @@
...
+  },
+  argTypes: {
+    adNetwork: {
+      options: [],
+      control: 'select',
+    },
+  },
+};
+
+// export const _default = (args) => {
```
timarney: Should remove this.

timarney:
1) On a laptop where I had previously installed [camo](https://reincubate.com/camo/) 

The camo option was available in the dropdown --- I selected it (knowing it wouldn't be available) .... it prompted to install (I canceled) and ended up in a state where the camera "green light" is on but seeing these screens.

Screen Shot 2022-06-15 at 6 07 00 AM


Exited the dialog ... webcam still on.
Screen Shot 2022-06-15 at 6 07 50 AM

timarney:
Dropdown is getting cut off.

Screen Shot 2022-06-15 at 6 04 29 AM

swissspidy:
> The camo option was available in the dropdown --- I selected it (knowing it wouldn't be available) .... it prompted to install (I canceled) and ended up in a state where the camera "green light" is on but seeing these screens.

Don't think we can really do anything about that. We get the list of available media devices from the browser and then try to use those. But I'll see if I can reproduce.

> Exited the dialog ... webcam still on.

Does this also happen when using another camera that is actually available?

> Dropdown is getting cut off.

I don't think it's cut off, but the dropdown items don't have any labels. Odd. I'll try to reproduce...

packages/story-editor/src/components/mediaRecording/footer.js
```diff
@@ -0,0 +1,253 @@
...
+    setCountdown: actions.setCountdown,
+    toggleRecordingMode: actions.toggleRecordingMode,
+    getMediaStream: actions.getMediaStream,
+    resetState: actions.resetState,
+  }));
+
+  const uploadWithPreview = useUploadWithPreview();
+
+  const onRetry = useCallback(() => {
+    resetState();
```
timarney, line 160: Noting:
When using "image" mode.

I was expecting when I pressed retry I would remain in "image" mode ... with a fresh countdown vs being taken back to the record video / capture image option screen.

swissspidy: What if you decided that you don't want to take a photo anymore but instead record a video?

timarney: True - I can exit but that's quite a few more clicks to get back to either type of record mode vs going back to the options buttons.  Thinking more of a user doing this multiple times trying to get the image they want.  

I think what you have works works for now.

timarney:
> Don't think we can really do anything about that. We get the list of available media devices from the browser and then try to use those. But I'll see if I can reproduce.

Yes that's fine just noting the details.

--- Yes when using an available device exiting still leaves the camera on until I do a browser refresh.

packages/story-editor/src/app/media/utils/getResourceFromMedia3p.js
```diff
@@ -294,7 +292,7 @@ function getVideoResourceFromMedia3p(m) {
     height: videoUrls.full.height,
     poster: m.imageUrls[0].url,
     length,
-    lengthFormatted: formatVideoLength(length),
+    lengthFormatted: getVideoLengthDisplay(length),
```
spacedmonkey, line 295: 👍

spacedmonkey:
https://user-images.githubusercontent.com/237508/173822594-d84fc3cb-7a46-4b9f-ae84-4d03a31114da.mov

The length of the video is wrong on upload preview.

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
```diff
@@ -0,0 +1,261 @@
...
+                      muted
+                      disablePictureInPicture
+                    />
+                  )}
+                </VideoWrapper>
+              </>
+            )}
+            {isImageCapture && (
+              <VideoWrapper>
+                <Photo
```
spacedmonkey: We could add decoding="async" here.

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
```diff
@@ -0,0 +1,261 @@
...
+    try {
+      setMediaBlobUrl(createBlob(blob));
+    } catch (e) {
+      // Do nothing.
+    }
+
+    const imageFile = blobToFile(
+      blob,
+      `image-capture-${format(new Date(), 'Y-m-d-H-i')}.jpeg`,
+      'image/jpeg'
```
spacedmonkey: We could use `MEDIA_POSTER_IMAGE_MIME_TYPE` here.

packages/story-editor/src/components/mediaRecording/provider.js
```diff
@@ -0,0 +1,327 @@
...
+      return;
+    }
+
+    try {
+      const devices = await navigator.mediaDevices.enumerateDevices();
+      setMediaDevices(
+        devices.filter((device) => device.kind !== 'audiooutput')
+      );
+    } catch (err) {
+      // Do nothing for now.
```
spacedmonkey: No error like a snackbar or error tracking.

packages/story-editor/src/components/mediaRecording/provider.js
```diff
@@ -0,0 +1,327 @@
...
+  const onStop = useCallback((blob) => {
+    try {
+      setMediaBlobUrl(createBlob(blob));
+    } catch (e) {
+      // Do nothing.
+    }
+    const f = blobToFile(
+      blob,
+      `webcam-capture-${format(new Date(), 'Y-m-d-H-i')}.mp4`,
+      'video/mp4'
```
spacedmonkey:
```suggestion
      MEDIA_TRANSCODED_MIME_TYPE
```

packages/story-editor/src/components/mediaRecording/provider.js
```diff
@@ -0,0 +1,327 @@
...
+
+  const onStop = useCallback((blob) => {
+    try {
+      setMediaBlobUrl(createBlob(blob));
+    } catch (e) {
+      // Do nothing.
+    }
+    const f = blobToFile(
+      blob,
+      `webcam-capture-${format(new Date(), 'Y-m-d-H-i')}.mp4`,
```
spacedmonkey:
```suggestion
      `webcam-capture-${format(new Date(), 'Y-m-d-H-i')}.${MEDIA_TRANSCODED_FILE_TYPE}`,
```

spacedmonkey:
Do we need to generate metadata for these videos like this - 
https://github.com/GoogleForCreators/web-stories-wp/blob/e00e50da7a02d956b993f4038ca1c15d46eda3ef/packages/story-editor/src/app/media/utils/getResourceFromUrl.js#L75-L88

So that preview looks better on upload.

spacedmonkey:
I am seeing airpods twice on this list. 
Screenshot 2022-06-15 at 13 44 43

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
swissspidy: It‘s not a poster image though, it‘s a different kind of image file that just happens to be a JPEG as well. Using this const would be misleading, which is why I didn‘t use it.

spacedmonkey:
In safari I am not seeing the gif logo. 
Screenshot 2022-06-15 at 13 56 22

spacedmonkey:
Screenshot 2022-06-15 at 13 58 49 (2)
Picture in picture icon showing in firefox.

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
spacedmonkey: Define a new const then?

spacedmonkey:
Screenshot 2022-06-15 at 14 10 41
Screenshot 2022-06-15 at 14 09 58
Screenshot 2022-06-15 at 14 09 52

Do we really want to render the native video controls here? They look very different in different browser. Safari for example has option to airplay the video, chrome to download the video and firefox has PIP options. 

We don't have controls like this on other videos do we?

spacedmonkey:
This functionality even works on iPad OS. 

https://user-images.githubusercontent.com/237508/173839548-52d1f17d-6c2d-4439-97e3-21f57dc2ff79.mov

packages/story-editor/src/components/mediaRecording/provider.js
```diff
@@ -0,0 +1,327 @@
...
+    setIsGif(false);
+    setMediaBlobUrl(null);
+    setCountdown(0);
+
+    // clearMediaStream() is supposed to do this, but it doesn't have an effect.
+    // Probably because liveStream is a new MediaStream instance.
+    // Anyway, this stops the camera/mic from being used.
+    if (liveStream) {
+      liveStream.getTracks().forEach((track) => {
+        track.stop();
```
spacedmonkey: Should this have a catch?

mwritter:
> Do we really want to render the native video controls here? They look very different in different browser. Safari for example has option to airplay the video, chrome to download the video and firefox has PIP options.

You could use `html` attributes on the `video` element to remove the extra controls, each browser my render them differently but the 'allowed' controls could all be the same.

packages/story-editor/src/components/mediaRecording/provider.js
swissspidy: No, since [`MediaStreamTrack.stop()`](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/stop) does not return a Promise nor does it throw errors.

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
swissspidy: It probably doesn't make any difference because it's basically an existing blob from the recording, but sure.

packages/story-editor/src/components/mediaRecording/provider.js
swissspidy: Correct.

packages/story-editor/src/components/mediaRecording/provider.js
spacedmonkey: Sorry, I should have lookup this api up first. 🤦

packages/story-editor/src/components/mediaRecording/footer.js
```diff
@@ -0,0 +1,271 @@
...
+    args.resource = {
+      ...resource,
+      alt: __('Camera Capture', 'web-stories'),
+    };
+
+    if (file.type.startsWith('video')) {
+      args.additionalData.isMuted = isMuted;
+
+      if (isGif) {
+        args.additionalData.mediaSource = 'gif-conversion';
```
spacedmonkey: @swissspidy Should we be marking video recording. So we could find them later? A "My recordings" section or something?

swissspidy: Good idea!

Though then I have to figure out another way to mark a recorded GIF as a GIF so that `getResourceFromAttachment()` returns a GIF resource accordingly. Any ideas?

spacedmonkey:
https://user-images.githubusercontent.com/237508/174047241-7003971e-d780-46bb-b88a-3eab092d9dc3.mov

I am noticing the recording icon is still showing in the tabs and I am seeing the light on the web cam still on, long after recording is finished.

packages/story-editor/src/components/mediaRecording/footer.js
```diff
@@ -0,0 +1,271 @@
...
+
+  return (
+    <StyledFooter showOverflow>
+      {hasMediaToInsert && (
+        <>
+          <RetryButton onClick={onRetry}>
+            {__('Retry', 'web-stories')}
+          </RetryButton>
+          <InsertButton onClick={onInsert}>
+            {__('Insert', 'web-stories')}
```
spacedmonkey: In my testing, the insert can take 1 second to kick in. I think we should disable this button after this is clicked, to ensure that users do not click it more than once.

packages/story-editor/src/components/mediaRecording/playPauseButton.js
```diff
@@ -0,0 +1,159 @@
+/*
```
spacedmonkey, line 1: This file appears to be a copy and paste from the video element. Is there anyway to move this component to be spared?

spacedmonkey: We also have a play / pause button here 
https://github.com/GoogleForCreators/web-stories-wp/blob/d8b28a203a9874c890d99aa1ad24935a48a6a8b1/packages/story-editor/src/components/panels/shared/media/audioPlayer.js#L173-L181

I wonder these could all use the same compontent.

packages/story-editor/src/components/mediaRecording/provider.js
```diff
@@ -0,0 +1,341 @@
...
+  }, [duration, stopRecording]);
+
+  const resetStream = useCallback(() => {
+    // clearMediaStream() is supposed to do this, but it doesn't have an effect.
+    // Probably because liveStream is a new MediaStream instance.
+    // Anyway, this stops the camera/mic from being used.
+    if (liveStream) {
+      liveStream.getTracks().forEach((track) => {
+        track.stop();
+      });
```
spacedmonkey:
```suggestion
      liveStream.getTracks().forEach((track) => track.stop());
```

packages/story-editor/src/components/mediaRecording/playPauseButton.js
swissspidy: > I am noticing the recording icon is still showing in the tabs and I am seeing the light on the web cam still on, long after recording is finished.

The component used for the video element is much more complex, it can't be reused here

As for the play button itself, maybe something for the future to move that to the design system or something.

swissspidy:
> I am noticing the recording icon is still showing in the tabs and I am seeing the light on the web cam still on, long after recording is finished.

Tim flagged the same thing further above as well. I am still looking into it.

packages/story-editor/src/components/mediaRecording/provider.js
```diff
@@ -0,0 +1,341 @@
...
+    resetState();
+  }, [resetState]);
+
+  const toggleSettings = useCallback(() => {
+    setIsSettingsOpen((state) => !state);
+  }, []);
+
+  const toggleIsGif = useCallback(() => {
+    setIsGif((state) => !state);
+  }, []);
```
spacedmonkey: Is there much value in wrapping in `useCallback`

swissspidy: No. I'll remove these. Some leftovers from development.

packages/story-editor/src/components/mediaRecording/footer.js
spacedmonkey: Nothing jumps out at me. But a refactor like should be in another PR.

spacedmonkey:
Other than the nitpics and issue I have already flagged. This is looking good 👍

packages/story-editor/src/components/mediaRecording/provider.js
swissspidy: Actually, in my testing it helps to avoid unnecessary re-renders

packages/story-editor/src/components/mediaRecording/footer.js
swissspidy: Recording mode is closed within the `onInsert()` callback. The only thing in there that could slow things down is the async `getResourceFromLocalFile()` call.

In my local testing so far that's always been super quick though, even for long videos.

Could you perhaps do some simple time tracking to see how fast/slow it is on your device?

e.g.

```js
// in onInsert
const before = window.performance.now();

// ...
// await getResourceFromLocalFile(file)
// ...

const after = window.performance.now();
console.log(after - before);
```

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
```diff
@@ -0,0 +1,228 @@
...
+
+  const isMuted = !hasAudio || isGif;
+
+  const captureImage = useCallback(async () => {
+    if (!streamRef.current) {
+      return;
+    }
+
+    const blob = await getImageFromVideo(streamRef.current);
+    try {
```
spacedmonkey:
```suggestion
    let blob;
    try {
       blob = await getImageFromVideo(streamRef.current);
```

Why not wrap this in try catch?

swissspidy: I'll need to think more about the error handling there in general, but ya probably needed

packages/story-editor/src/components/canvas/useInsertElement.js
```diff
@@ -39,9 +39,13 @@ function createElementForCanvas(type, props) {
...
-  const { addElement } = useStory(({ actions }) => ({
-    addElement: actions.addElement,
-  }));
+  const { addElement, combineElements, backgroundElementId } = useStory(
+    (state) => ({
+      addElement: state.actions.addElement,
+      combineElements: state.actions.combineElements,
+      backgroundElementId: state.state.currentPage?.elements?.[0]?.id,
+    })
+  );
```
spacedmonkey:
```suggestion
  const { addElement, combineElements, backgroundElementId } = useStory(
    ({actions, state}) => ({
      addElement: actions.addElement,
      combineElements: actions.combineElements,
      backgroundElementId: state.currentPage?.elements?.[0]?.id,
    })
  );
```

packages/story-editor/src/components/mediaRecording/provider.js
```diff
@@ -0,0 +1,329 @@
...
+    // and then calling enumerateDevices() will trigger another permission prompt.
+    // TODO: Figure out a good way to work around that.
+    navigator.mediaDevices.addEventListener('devicechange', updateMediaDevices);
+
+    return () => {
+      navigator.mediaDevices.removeEventListener(
+        'devicechange',
+        updateMediaDevices
+      );
+    };
```
spacedmonkey, line 177:
```suggestion
    return () => navigator.mediaDevices.removeEventListener( 'devicechange', updateMediaDevices );
```

packages/story-editor/src/components/mediaRecording/settingsModal.js
```diff
@@ -0,0 +1,181 @@
...
+    clearMediaBlob: actions.clearMediaBlob,
+    toggleSettings: actions.toggleSettings,
+  }));
+
+  const [localVideoInput, setLocalVideoInput] = useState(videoInput);
+  const [localAudioInput, setLocalAudioInput] = useState(audioInput);
+
+  useEffect(() => {
+    setLocalVideoInput(videoInput);
+  }, [videoInput]);
```
spacedmonkey:
```suggestion
  useEffect(() => setLocalVideoInput(videoInput), [videoInput]);
```

packages/story-editor/src/components/mediaRecording/settingsModal.js
```diff
@@ -0,0 +1,181 @@
...
+
+  useEffect(() => {
+    setLocalAudioInput(audioInput);
+  }, [audioInput]);
```
spacedmonkey:
```suggestion
  useEffect(() => setLocalAudioInput(audioInput), [audioInput]);
```

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
```diff
@@ -0,0 +1,228 @@
...
+                    <>
+                      {/* eslint-disable-next-line jsx-a11y/media-has-caption -- We don't have tracks for this. */}
+                      <Video
+                        ref={videoRef}
+                        src={mediaBlobUrl}
+                        autoPlay
+                        muted={isMuted}
+                        loop={isGif}
+                        disablePictureInPicture
+                        tabIndex={0}
```
spacedmonkey: I wonder if we should have shared attributes like this. 
```
const videoAttributes = {
   disablePictureInPicture: true,
   autoPlay: true,
   tabIndex: 0, 
   muted: isMuted,
}
```

```
<Video ref={streamRef}  {...videoAttributes} />
```

packages/story-editor/src/components/mediaRecording/footer.js
spacedmonkey: I did some profiling. 

21 second video - 489.49000000208616
2 minute video - 639.6449999958277
2 minute video - 2004.3800000026822 - 4x cpu slowdown. 
1 minute video - 2695.420000001788 - 6x cpu slowdown. 

To me it is noticeable on a 2.3 GHz 8-Core Intel Core i9.

spacedmonkey: https://user-images.githubusercontent.com/237508/174269014-eda3e2b2-ebd3-44da-ad8a-8e66ab2f8b4d.mov

This takes around 3 second for the insert button to disappear. My worry here is that a user will click it twice.

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
spacedmonkey: We could just wrap everything in a try catch maybe. I don't see why it not, error handling is good.

packages/story-editor/src/components/mediaRecording/provider.js
swissspidy: With Prettier this will just become 

```diff
-    return () => {
+    return () =>
      navigator.mediaDevices.removeEventListener(
        'devicechange',
        updateMediaDevices
      );
-    };
```

And thus not adding any value

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
swissspidy: There is more to that than just wrapping in a `try catch`. I am also thinking about user messaging etc. I will revisit this section.

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
swissspidy: I don't see the need for this. These props are just for this one particular component, not shared with others, and not passed down further in the tree. Moving them away from here to a separate `const` would make it harder to follow.

packages/story-editor/src/components/mediaRecording/footer.js
swissspidy: Thanks for sharing these numbers and the video 👍 Very helpful

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
spacedmonkey: Why are the attributes difference between components? Like tabindex. 

Could these be one component with conditional attributes.

swissspidy: One `<Video>` is for displaying your webcam stream before and during recording (the one with the `streamRef`)

The other one (with the `videoRef`) displays the final video after recording and has different props depending on whether it's a GIF or not.
It needs a `tabIndex` so that it can receive keyboard focus, which allows you to still play/pause the video even without any actual video controls being displayed. Otherwise, videos without controls are not interactive.

These two don't really share much as they serve different purposes.

I'm happy to move the only two shared props (`autoPlay` and `disablePictureInPicture`) to the base styled component, but that's the only optimization I can think of here without making the code less readable.

packages/story-editor/src/app/media/local/actions.js
```diff
@@ -91,42 +91,6 @@ export const removeAudioProcessing =
     });
   };
 
-export const setBaseColorProcessing =
```
spacedmonkey, line 94: Why change this? Is it related to this PR?

swissspidy: I noticed that base color and blurhash for a single recorded video/image were generated and uploaded up to 3 times. I‘ve tweaked the logic as part of this PR.
Now it only happens once, as expected.

Will clean up the code & update the PR description accordingly soon.

spacedmonkey: Do we need to do the same for poster images?

swissspidy: The poster image was only generated once, all looked good there. But will check again.

packages/story-editor/src/components/library/panes/media/local/mediaRecording/mediaRecording.js
```diff
@@ -54,13 +56,28 @@ function MediaRecording() {
...
+  const { isTranscodingEnabled } = useFFmpeg();
+
   const onClick = useCallback(() => {
     trackEvent('media_recording_toggled', {
       status: isInRecordingMode ? 'closed' : 'open',
     });
     toggleRecordingMode();
   }, [isInRecordingMode, toggleRecordingMode]);
 
+  if (!enableMediaRecording) {
```
spacedmonkey, line 69:
```suggestion
  if (!enableMediaRecording || !isTranscodingEnabled) {
```

swissspidy: This check is specifically done a few lines down in order to a) make removing the feature flag code in the future easier b) provide a contextual explanatory inline comment before the check.

spacedmonkey: We don't do this in other places. See
https://github.com/GoogleForCreators/web-stories-wp/blob/c7450e9d5da803f970a1931ab83b273dcc908b07/packages/story-editor/src/components/panels/design/captions/captions.js#L141
https://github.com/GoogleForCreators/web-stories-wp/blob/c7450e9d5da803f970a1931ab83b273dcc908b07/packages/story-editor/src/components/panels/design/pageBackgroundAudio/backgroundAudio.js#L64

swissspidy: Which is not nice but also not the end of the world

spacedmonkey:
While the video is processing, I am seeing 0:00 as timestamp. 
Screenshot 2022-06-20 at 16 48 42

spacedmonkey:
RTL
Screenshot 2022-06-20 at 16 57 06

spacedmonkey:
Are we okay with buttons looking like this with longer text?
Screenshot 2022-06-20 at 16 59 31

swissspidy:
> While the video is processing, I am seeing 0:00 as timestamp.

Fixed in 870eebc382a325bb61d7f8d43c74eae3cd6bee87 👍

Improved the hardening in this area once again

> RTL

Fixed in 00d0bb55c5b77cf98a295a4db1e5544b4c770947 👍 

> Are we okay with buttons looking like this with longer text?

Yes, I don't see anything wrong there. It's fine for the buttons to exceed the canvas width

spacedmonkey (approved):
Welcome done on getting this across the line 🎉

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
```diff
@@ -0,0 +1,247 @@
...
+
+    try {
+      blob = await getImageFromVideo(streamRef.current);
+      setMediaBlobUrl(createBlob(blob));
+    } catch (e) {
+      trackError('media_recording_capture', e.message);
+
+      showSnackbar({
+        message: __(
+          ' There was an error taking a photo. Please try a gain.',
```
timarney:
```suggestion
          ' There was an error taking a photo. Please try again.',
```

timarney (approved):

PR #11678, opened 2022-06-13
Dashboard: Filter taxonomies by `show_ui` value

spacedmonkey:
## Summary

Only show filters that have the show_ui flag set to true. 

## Testing Instructions

1.

## Checklist

---

Fixes #


sblinde (approved):
👏

Awesome, thanks for this. My PHP is rusty for sure but based on what knowledge I have, this makes sense to me.

includes/REST_API/Stories_Taxonomies_Controller.php
```diff
@@ -159,6 +161,11 @@ public function get_collection_params(): array {
 			'type'        => 'boolean',
 		];
 
+		$query_params['show_ui'] = [
+			'description' => __( 'Whether to show only show in ui taxonomies.', 'web-stories' ),
```
swissspidy:
```suggestion
			'description' => __( 'Whether to show only show taxonomies that allow a UI in the admin.', 'web-stories' ),
```

Based on the docblocks in `WP_Taxonomy`, improving readability a bit.

mwritter (approved):
👍

tests/phpunit/integration/tests/REST_API/Stories_Taxonomies_Controller.php
```diff
@@ -95,6 +95,27 @@ public function test_get_items_hierarchical_false(): void {
...
+		$request->set_param( 'type', 'attachment' );
+		$response = rest_get_server()->dispatch( $request );
+
+		$this->assertFalse( $response->is_error() );
+		$data = $response->get_data();
+		$this->assertNotEmpty( $data );
+		$this->assertCount( 1, $data );
+		$this->assertArrayHasKey( 'web_story_media_source', $data );
+		$this->assertArrayHasKey( 'name', $data['web_story_media_source'] );
+		$this->assertSame( 'Source', $data['web_story_media_source']['name'] );
```
swissspidy: While the feature works and the test passes, the test does not actually verify that the taxonomy has `show_ui` set to `false`

Assertions like this would be more robbust:

```suggestion
		$this->assertFalse( $response->is_error() );
		$data = $response->get_data();
		$this->assertNotEmpty( $data );
		foreach( array_keys( $data ) as $taxonomy ) {
			$tax = get_taxonomy( $taxonomy );
			$this->assertInstanceOf(\WP_Taxonomy::class, $tax);
			$this->assertFalse($tax->show_ui);
		}
```

In addition to that, it would be great to have tests for the `show_ui=>true` case as well. Compare this to the tests for the `hierarchical` option, where all three cases are covered:

* `test_get_items`: covers default case (all taxonomies included)
* `test_get_items_hierarchical_false`
* `test_get_items_hierarchical_true`

It should be easy to copy these tests and adapt them for `show_ui`

spacedmonkey: Refactored the test to be a little cleaner.

swissspidy: I take it you see little benefit in adding tests for the other cases (default, false, true)?

spacedmonkey: > I take it you see little benefit in adding tests for the other cases (default, false, true)?

Correct, but I can add it if you want.

swissspidy (approved):

tests/phpunit/integration/tests/REST_API/Stories_Taxonomies_Controller.php
spacedmonkey: I have updated tests to make them a lot cleaner.

tests/phpunit/integration/tests/REST_API/Stories_Taxonomies_Controller.php
```diff
@@ -77,42 +90,56 @@ public function test_get_items(): void {
...
+		foreach ( $data as $tax ) {
+			$this->assertArrayHasKey( 'hierarchical', $tax );
+			$this->assertSame( $hierarchical, $tax['hierarchical'] );
 		}
 	}
+
+	/**
+	 * @return array
+	 */
+	public function data_show_ui(){
```
swissspidy: @spacedmonkey FYI there are some lint warnings now here.

spacedmonkey: Fixed.

spacedmonkey:
Side note, we could also use this here in the editor. 

https://github.com/GoogleForCreators/web-stories-wp/blob/27c547c73de98ae361f340cc025122ece0441b9f/packages/story-editor/src/components/panels/document/taxonomies/taxonomies.js#L52

swissspidy:
Wanna include that change here in this PR as well?

spacedmonkey:
> Wanna include that change here in this PR as well?
Added in https://github.com/GoogleForCreators/web-stories-wp/pull/11678/commits/d2ed5a49cedacc61d04015ea080935690655b22e

PR #11683, opened 2022-06-13
Shopping: Infinite Scroll Pagination

### LINKED ISSUE #11593: Shopping: Infinite Scroll Pagination

spacedmonkey:
## Summary

Handle pagination in the shopping panel. Reuse dashboard infinite scroll component. 

## User-facing changes

https://user-images.githubusercontent.com/237508/173386049-6a9aa486-cc7f-4f4b-b9d5-19533326269d.mov

## Testing Instructions

1. Active shopping with shopify. 
2. Go to editor. 
3. Go to product tab. 
4. Scroll to bottom. 
5. See more products load in. 

## Checklist

---

Fixes #11593


spacedmonkey:
Developers testing this PR, I recommend changing this line to 10. 

https://github.com/GoogleForCreators/web-stories-wp/blob/9c739fa071dd2785c81887cd7528bf556ddb4f83/packages/wp-story-editor/src/api/shopping.js#L39

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -236,20 +265,31 @@ function ShoppingPane(props) {
...
-          <Loading>
-            <Spinner>
-              <CircularProgress size={24} />
-            </Spinner>
-          </Loading>
+        {!loaded && isLoading && (
+          <LoadingContainer>
+            <LoadingSpinner animationSize={50} circleSize={6} />
+          </LoadingContainer>
         )}
```
swissspidy: Now there is no spinner anymore when searching for products or sorting them. Let's ensure that we keep displaying a spinner in these cases.

spacedmonkey: It doesn't have a spinner as it does not need one anymore.

swissspidy: Why don't we _need_ a spinner anymore when searching?

The spinner was a useful indicator for the user to know that something is happening.

spacedmonkey: Added back.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -269,20 +292,31 @@ function ShoppingPane(props) {
...
-        {isLoading && (
-          <Loading>
-            <Spinner>
-              <CircularProgress size={24} />
-            </Spinner>
-          </Loading>
+        {!loaded && isLoading && (
+          <LoadingContainer>
+            <LoadingSpinner animationSize={50} circleSize={6} />
```
swissspidy: 50 is quite large. Let's go with half of this for now, 25.

Here and also in `InfiniteScroller`. We can pass `animationSize` as a prop there.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -235,9 +245,22 @@ function ShoppingPane(props) {
...
+    setPage(1);
+    setLoaded(false);
+    debouncedProductsQuery('', 1, orderby, order);
   }, [debouncedProductsQuery, order, orderby]);
 
+  const onLoadMore = useCallback(() => {
+    if (canLoadMore) {
+      const newPage = page + 1;
+      setPage(newPage);
+      debouncedProductsQuery(searchTerm, newPage, orderby, order);
```
swissspidy: Let's use `searchProducts` here instead. Otherwise this will trigger a tracking event, which we don't need for infinite scrolling.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -73,53 +68,72 @@ function ShoppingPane(props) {
...
-    (value = '', sortBy, sortOrder) => {
-      searchProducts(value, sortBy, sortOrder);
+    (value = '', _page, sortBy, sortOrder) => {
+      searchProducts(value, _page, sortBy, sortOrder);
       trackEvent('search', {
         search_type: 'products',
         search_term: value,
         search_order: sortOrder,
         search_orderby: sortBy,
+        search_page: _page,
```
swissspidy: Good idea but I don't think we're interested in tracking individual pages in infinite scroll for now.

swissspidy (commented):
Just a couple of nits, otherwise looks good. Working well.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
spacedmonkey:
```suggestion
```

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
spacedmonkey:
```suggestion
      searchProducts(searchTerm, newPage, orderby, order);
```

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -235,9 +245,22 @@ function ShoppingPane(props) {
...
+    debouncedProductsQuery('', 1, orderby, order);
   }, [debouncedProductsQuery, order, orderby]);
 
+  const onLoadMore = useCallback(() => {
+    if (canLoadMore) {
+      const newPage = page + 1;
+      setPage(newPage);
+      debouncedProductsQuery(searchTerm, newPage, orderby, order);
+    }
+  }, [canLoadMore, debouncedProductsQuery, order, orderby, page, searchTerm]);
```
spacedmonkey:
```suggestion
  }, [canLoadMore, searchProducts, order, orderby, page, searchTerm]);
```

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
spacedmonkey: Screenshot 2022-06-14 at 10 22 16

We have bigger elsewhere.

swissspidy: Hmm but not the same size 50 either. I'll create a separate ticket to unify these then. Easiest.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -235,9 +244,22 @@ function ShoppingPane(props) {
 
   const handleClearInput = useCallback(() => {
     setSearchTerm('');
-    debouncedProductsQuery('', orderby, order);
+    setPage(1);
+    setLoaded(false);
+    debouncedProductsQuery('', 1, orderby, order);
```
timarney, line 251: Maybe move `setLoaded` above `setPage` here or use to keep passing the page number closer together ... not hard to spot but could get missed.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -73,48 +68,67 @@ function ShoppingPane(props) {
   const speak = useLiveRegion('assertive');
   const [loaded, setLoaded] = useState(false);
   const [isLoading, setIsLoading] = useState(false);
```
spacedmonkey: @swissspidy There is now a lot of state here. Wonder if we should refactor this later to be a little cleaner.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
spacedmonkey: Moved the setPage all together here.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
spacedmonkey: I added the props in. I had to change both size and circles. So I pass an object instead. Seems to work well.

packages/design-system/src/components/infiniteScroller/index.js
```diff
@@ -176,7 +180,7 @@ const InfiniteScroller = ({
 
     return (
       <LoadingContainer>
-        <LoadingSpinner animationSize={50} circleSize={6} />
+        <LoadingSpinner {...loadingSpinnerProps} />
```
timarney, line 183: Might want to give this a little more space under the search input.

Screen Shot 2022-06-14 at 5 50 37 AM

spacedmonkey:
Wondering that now, we have this functionality, if we should lower the per page value, to say 40/50. We don't need 100 items and might mean the REST API requests are a little faster.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -269,20 +290,32 @@ function ShoppingPane(props) {
...
+            <ProductList
+              isMenuFocused={isMenuFocused}
+              onClick={onClick}
+              products={products}
+              onPageProducts={currentPageProducts}
+            />
+            <InfiniteScroller
+              allDataLoadedMessage={allDataLoadedMessage}
+              allDataLoadedAriaMessage={__(
+                'Products are loaded',
```
timarney, line 310: Testing with VoiceOver... on the initial load it seems "Products are loaded" get announced.

1) When no results turn up ... that doesn't seem to get announced.

Screen Shot 2022-06-14 at 5 56 45 AM

2) I also get Loading more products announced when the search is already done.

Screen Shot 2022-06-14 at 5 56 09 AM

spacedmonkey: Odds are these are issues with the upstream compontent and not in my implementation. Let's note these and fix them in another issue.

swissspidy:
> Wondering that now, we have this functionality, if we should lower the per page value, to say 40/50. We don't need 100 items and might mean the REST API requests are a little faster.

Yeah makes sense. 50 seems like a good start for now.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
swissspidy: Doesn't seem like too much right now. Maybe something for the future when we add more stuff here—can keep an eye on it.

packages/design-system/src/components/infiniteScroller/index.js
spacedmonkey: @timarney Can you pull down the latest changes. 
Screenshot 2022-06-14 at 11 06 16

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
swissspidy: Can you create that issue?

Looks like `InfiniteScroller` should probably also use `useLiveRegion` for the announcements instead of the custom `AriaOnlyAlert`.

packages/design-system/src/components/infiniteScroller/index.js
timarney: That better.

packages/design-system/src/components/infiniteScroller/index.js
```diff
@@ -176,7 +180,7 @@ const InfiniteScroller = ({
...
       </LoadingContainer>
     );
   }, [allDataLoadedMessage, canLoadMore, loadState]);
```
swissspidy:
```suggestion
  }, [allDataLoadedMessage, canLoadMore, loadState, loadingSpinnerProps]);
```

Though maybe need to run Prettier on that

spacedmonkey: Fixed.

timarney (approved):

PR #11700, opened 2022-06-14
Undoing a layer rename reverts to rename when going back to edit mode.

### LINKED ISSUE #11686: Undoing a layer rename reverts to rename when going back to edit mode.
## Bug Description

After renaming a layer and undoing the layer name the layer name uses the wrong value when re-entering edit mode.

See:
https://github.com/GoogleForCreators/web-stories-wp/pull/11676#issuecomment-1153807461

## Expected Behaviour

When editing the layer name it contains the current value for the layer.

## Steps to Reproduce

- Add a layer (layer has default name)
- Rename a layer ('my layer')
- Undo the rename (layer has default name)
- 2x click to rename again --- editable text now shows previous rename ('my layer')

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

timarney:
## Context

After renaming a layer and undoing the layer name the layer name uses the wrong value when re-entering edit mode.

See:
https://github.com/GoogleForCreators/web-stories-wp/pull/11676#issuecomment-1153807461

## Summary

I think the issue here is the useState + initial value

> the initial value matters only the first time a component renders. After that, it will preserve the state between renders.

https://dev.to/cronokirby/react-pitfalls-usestate-initialization-39be

This PR adds a useEffect to ensure the "initial value" element.layerName gets updates

## Testing Instructions

1. Enable the layer renaming feature
2. Add a new element to the stage
3. Open the layers panel
4. Rename the layer
5. Undo the rename 
6. Enter edit mode to rename the layer again --- the value should match the current layer name

No

No

No

## Checklist

---

Fixes #11686


packages/story-editor/src/components/panels/design/layer/layer.js
```diff
@@ -352,6 +352,10 @@ function Layer({ element }) {
     })
   );
 
+  useEffect(() => {
+    setNewLayerName(layerName);
+  }, [layerName]);
```
barklund, line 357: With this, I think we can also remove  almost all the other places, where `setNewLayerName` is invoked, as most of them are workarounds to exactly this thing missing.

barklund (approved):
Works great!

packages/story-editor/src/components/panels/design/layer/layer.js
timarney: Okay so 

I think we'll still need the `handleChange` and `handleKeyDown`

But we can update `updateLayerName` to be this code below ya?

```javascript
const updateLayerName = () => {
    setRenamableLayer(null);
    const trimmedLayerName = newLayerName.trim();
    if (trimmedLayerName) {
      updateElementById({
        elementId: element.id,
        properties: { layerName: trimmedLayerName },
      });
    }
  };
```

timarney: Pushed an update for that 

https://github.com/GoogleForCreators/web-stories-wp/pull/11700/commits/c1680cdb393ce3094b03553aed31499fb3db82d3

merapi:
Oh, this should go to QA first, but since it's a tiny fix and all of us 3 already tested it - I suggest leaving/accepting it as is.

PR #11712, opened 2022-06-15
Improve PHP type hints and casting

swissspidy:
## Context

Originally I only wanted to update the `ampproject/amp-wp` Composer dependency since a new version of the AMP plugin was released.

It turned out to be quite the rabbit hole because I started using PHP 8.1 locally and couldn't run the typical PHPCS and PHPStan workflows anymore on this version, so I ended up upgrading them, which in turn resulted in a ton of linter warnings that had to be addressed.

## Summary

One of the main improvements this includes is specifying more precisely the contents of arrays.

**Examples**

Bad:

```php
@param array $args WP_Query arguments.
```

Better:

```php
@param array<string, mixed> $args WP_Query arguments.
```

Bad:

```php
@return array|false Lock data or false.
```

Better:

```php
@return array{time?: int, user?: int}|false Lock data or false.
```

Bad:

```php
$alt = get_post_meta( $image_id, '_wp_attachment_image_alt', true );
```

Better:

```php
/**
 * Alt text.
 *
 * @var string $alt
 */
$alt = get_post_meta( $image_id, '_wp_attachment_image_alt', true );
```

Useful links to learn more about PHPStan and PHPStan types:

https://phpstan.org/blog/solving-phpstan-no-value-type-specified-in-iterable-type
https://phpstan.org/writing-php-code/phpdoc-types

## Relevant Technical Choices

Uses `phpstan-type`, `phpstan-var`, `phpstan-param` etc. directives to improve type hints for the more complex data types across the code base

Temporarily updates `szepeviktor/phpstan-wordpress` to `dev-master` because there hasn't been a release yet for the latest updates.

Also adds `ReturnTypeWillChange` attributes where needed to suppress deprecation notices on PHP 8.1

## User-facing changes

N/A

## Testing Instructions

1.

No

No

No

## Checklist

---

Fixes #


includes/compat/Web_Stories_Compatibility.php
```diff
@@ -352,7 +352,7 @@ public function get_php_version() {
 	 * @return array
 	 */
 	public function get_extensions() {
-		return (array) $this->extensions;
+		return $this->extensions;
```
spacedmonkey, line 355: ?

includes/compat/Web_Stories_Compatibility.php
```diff
@@ -365,7 +365,7 @@ public function get_extensions() {
 	 * @return array
 	 */
 	public function get_required_files() {
-		return (array) $this->required_files;
+		return $this->required_files;
```
spacedmonkey, line 368: ?

includes/compat/Web_Stories_Compatibility.php
swissspidy: It‘s known to be an array so there‘s no need for the type cast.

I ended up addressing hundreds of linter warnings, this is one of them.

But please note that everything here is still in progress, which is why I converted this to a draft again.

includes/compat/Web_Stories_Compatibility.php
swissspidy: See above

includes/Admin/Admin.php
```diff
@@ -114,15 +114,20 @@ public function admin_body_class( $class ) {
 	 * @since 1.0.0
 	 *
 	 * @param string|mixed $content Default post content.
-	 * @param \WP_Post     $post    Post object.
+	 * @param WP_Post      $post    Post object.
 	 * @return string|mixed Pre-filled post content if applicable, or the default content otherwise.
 	 */
-	public function prefill_post_content( $content, $post ) {
+	public function prefill_post_content( $content, WP_Post $post ) {
```
spacedmonkey, line 120: This hook into a filter, so can ensure that this is a WP_Post object here?

swissspidy: Yes, because only the first argument is filterable, the second one provided by WP never changes the type. It's always a `WP_Post`.

Please note that this PR is still in progress. I somehow ended up doing way more than what I originally anticipated. I might revert much of it or even close this again.

timarney:
Reviewed the code + test install + small click around --- working as expected.

spacedmonkey (approved):

timarney (approved):

PR #11724, opened 2022-06-15
Shopping: Hide disallowed products from the product dropdown

### LINKED ISSUE #11706: Shopping: Hide disallowed products from the product dropdown
## Feature Description

It is currently possible to end up adding duplicate products or ones without an image to a page, since we don't currently enforce that in the product dropdown (as used in the Style tab and the floating menu)

So we need to enforce the same rules for duplicates & products without images also for the dropdown.

Since the dropdown does not have way to disable items or mark them as disallowed, the easiest way to achieve this is by adjusting `getProductsByQuery` in `ProductDropdown` to filter out all products from the result set that are not allowed.

Good opportunity to add Karma tests to verify the change.

To reproduce:

1. Add product to the page
2. Using the dropdown in the floating menu, choose a product without an image or one that's already on the page

spacedmonkey:
## Summary

Do not allow products to re-selected or products without images.

## Testing Instructions

#### Duplicate products
1. Create a new story. 
2. Add a couple of products onto canvas. 
3. Select product. 
4. See dropdown. See that other products that are on canvas are not in dropdown. 

#### Product without image.
1. Create a new story. 
2. Add a couple of products onto canvas. 
3. Select product. 
4. See dropdown. See that other products without an image, do not appear in dropdown. 

## Checklist

---

Fixes #11706

### DIFF
--- a/packages/story-editor/src/components/library/panes/shopping/karma/shopping.karma.js
+++ b/packages/story-editor/src/components/library/panes/shopping/karma/shopping.karma.js
@@ -28,6 +28,8 @@ function isStoryEmpty() {
   return Boolean(document.getElementById('emptystate-message'));
 }
 
+const dropDownSelector = '[aria-label="Design menu"] [aria-label="Product"]';
+
 describe('Shopping integration', () => {
   let fixture;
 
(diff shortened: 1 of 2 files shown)


swissspidy:
As per the ticket description, this is a good opportunity to extend the Karma tests in `packages/story-editor/src/components/floatingMenu/elements/karma/shopping.karma.js` to verify the change.

swissspidy (approved):
Works well!

timarney (approved):

PR #11727, opened 2022-06-16
Shopping: Limit number of products per page 

### LINKED ISSUE #11254: Shopping: Limit number of products per page
## Feature Description

The exact solution is still TBD, but we want to limit the number of products per page in some way.

For example by displaying a snackbar message when the limit (currently 6?) is reached:

Screenshot 2022-04-12 at 14 49 26

spacedmonkey:
## Summary

Only allow 6 products on a page. 

## Relevant Technical Choices

Added some unit tests. 
Improved storybook. 
Add karma test. 

## User-facing changes

https://user-images.githubusercontent.com/237508/174128271-593b0fee-1fcd-49b3-9136-132fcc7dc71c.mov

## Testing Instructions

#### Test 7 products insert. 
1. Create a story. 
2. Go to product tab. 
3. Insert 7 products. 
4. 7th product should not insert and message shown. 

#### Test copy and paste
1. Create a story. 
2. Go to product tab. 
3. Insert 5 products. 
4. Select and copy product elements. 
5. Create new page. 
6. Paste elements. 
7. Message should be shown and ensure that no more than 6 products are shown. 

## Checklist

---

Fixes #11254


packages/story-editor/src/app/canvas/useCanvasCopyPaste.js
```diff
@@ -103,6 +103,19 @@ function useCanvasGlobalKeys() {
...
+      );
+
+      if (
+        currentPageProductIds.length >= 6 ||
+        newProductElements.length + currentPageProductIds.length > 6
+      ) {
+        showSnackbar({
+          message: __('Only 6 items can be added per page.', 'web-stories'),
+        });
+      }
```
swissspidy: Let's make the number 6 a const in `packages/story-editor/src/constants/index.js`, e.g. `MAX_PRODUCTS_PER_PAGE` or something. Then we only have to update this value once in the codebase if we wanna change it. Plus we avoid magic numbers.

Related: Should use `_n()` here + `sprintf()` for the message.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
```diff
@@ -243,9 +243,19 @@ function ShoppingPane(props) {
   );
 
   const onClick = useCallback(
```
swissspidy: On the ticket I left a comment last week that we should essentially do the same as #11616 — disable the insert button completely when the limit has been reached.

Then there is no need for the snackbar here.

The tooltip text in `packages/story-editor/src/components/library/panes/shopping/productButton.js` can be adjusted to mention the limit.

The existing Karma tests in `packages/story-editor/src/components/library/panes/shopping/karma/shopping.karma.js` can be amended to test this scenario

Also: Ditto here regarding usage of `_n()` and `sprintf()`

packages/story-editor/src/components/library/panes/shopping/productList.js
```diff
@@ -76,8 +78,9 @@ function ProductList({
           >
             <Product
               product={product}
+              canAddMore={canAddMore}
               isOnPage={onPageProducts.some(
-                (item) => item.product.productId === product?.productId
+                (item) => item?.product?.productId === product?.productId
```
spacedmonkey:
```suggestion
                (item) => item.product.productId === product?.productId
```

swissspidy:
@spacedmonkey Just got clarification/confirmation from Andrew now that we should completely prevent pasting/duplicating **all** products if pasting/duplicating would exceed the limit . Is this feasible to adapt?

packages/story-editor/src/app/canvas/useCanvasCopyPaste.js
spacedmonkey: Fixed.

packages/story-editor/src/components/library/panes/shopping/shoppingPane.js
spacedmonkey: Fixed.

spacedmonkey:
> @spacedmonkey Just got clarification/confirmation from Andrew now that we should completely prevent pasting/duplicating **all** products if pasting/duplicating would exceed the limit . Is this feasible to adapt?

Yeah, I have refactored the logic so it now, gets rid of products if limit is met.

packages/story-editor/src/app/canvas/useCanvasCopyPaste.js
```diff
@@ -103,6 +104,29 @@ function useCanvasGlobalKeys() {
...
+      );
+
+      if (
+        currentPageProductIds.length >= MAX_PRODUCTS_PER_PAGE ||
+        newProductElements.length + currentPageProductIds.length >
+          MAX_PRODUCTS_PER_PAGE
+      ) {
+        showSnackbar({
+          message: sprintf(
+            /* translators: %s: max number of products. */
```
timarney:
```suggestion
            /* translators: %d: max number of products. */
```

packages/story-editor/src/components/library/panes/shopping/productButton.js
```diff
@@ -66,27 +67,47 @@ const StyledActionsContainer = styled.div`
   margin-right: 10px;
 `;
 
-const DISABLED_PRODUCT_TEXT = __(
+const DISABLED_PRODUCT_TEXT = sprintf(
+  /* translators: %s: max number of products. */
```
timarney:
```suggestion
  /* translators: %d: max number of products. */
```

packages/story-editor/src/app/canvas/useCanvasCopyPaste.js
```diff
@@ -103,6 +104,29 @@ function useCanvasGlobalKeys() {
...
+            _n(
+              'Only %d item can be added per page.',
+              'Only %d items cans be added per page.',
```
swissspidy:
```suggestion
              'Only %d items can be added per page.',
```

packages/story-editor/src/components/library/panes/shopping/karma/shopping.karma.js
```diff
@@ -122,6 +134,22 @@ describe('Shopping integration', () => {
       await expect(productButton.getAttribute('aria-disabled')).toBe('true');
     });
 
+    it('should disable button if product more than 6 product', async () => {
```
swissspidy:
```suggestion
    it('should disable button if there are more than 6 products on the page', async () => {
```

packages/story-editor/src/components/library/panes/shopping/productButton.js
```diff
@@ -100,28 +121,32 @@ function ProductButton({ product, onClick, onFocus, isOnPage }) {
     product?.productTitle
   );
 
+  const disabled = !isOnPage && !canAddMore;
```
swissspidy, line 124:
```suggestion
  const disabled = (!isOnPage && !canAddMore) || !hasImage;
```

Makes the `hasImage && !disabled` and `!hasImage || disabled` checks below easier to maintain as there's only 1 variable. WDYT?

swissspidy (commented):
Just a couple of nits, otherwise looks good & works well! 🚀

timarney (approved):

packages/story-editor/src/components/library/panes/shopping/productButton.js
spacedmonkey: I don't think that is going to work. There are cases were we need to know disabled and HasImage separately. I also find that a little less readable.

packages/story-editor/src/app/canvas/useCanvasCopyPaste.js
```diff
@@ -103,6 +104,29 @@ function useCanvasGlobalKeys() {
...
+            _n(
+              'Only %d item can be added per page.',
+              'Only %d items can be added per page.',
+              MAX_PRODUCTS_PER_PAGE,
+              'web-stories'
+            ),
+            MAX_PRODUCTS_PER_PAGE
+          ),
+        });
+      }
```
swissspidy: Let's return early here to avoid double snackbar notifications:

```suggestion

      return addPastedElements(elements, animations);
      }
```

Screenshot 2022-06-17 at 16 00 39

swissspidy (commented):
Just noticed that this will cause double snackbar notifications, which is a bit noisy and confusing. If we can fix that, then this should be good to go.

packages/story-editor/src/app/story/useStoryReducer/test/addElements.js
```diff
@@ -108,4 +108,177 @@ describe('addElements', () => {
     });
     expect(result.selection).toStrictEqual(['123']);
   });
+
+  it('should only add a product once', () => {
```
swissspidy, line 112: Related: #11729

packages/story-editor/src/app/canvas/useCanvasCopyPaste.js
```diff
@@ -103,20 +104,48 @@ function useCanvasGlobalKeys() {
...
+              showSnackbar({
+                message: sprintf(
+                  /* translators: %s: product title. */
+                  __('Product "%s" already exists on the page.', 'web-stories'),
+                  productTitle
+                ),
+                thumbnail: productImages && {
+                  src: productImages[0].url,
+                  alt: productImages[0].alt,
+                },
```
spacedmonkey: This is just nice to have, but added the product image to the snackbar message.

packages/story-editor/src/app/canvas/useCanvasCopyPaste.js
spacedmonkey: Good catch. Pushed a fix and improved logic.

packages/story-editor/src/app/canvas/useCanvasCopyPaste.js
swissspidy: Wanna do the same change in `packages/story-editor/src/app/canvas/useCanvasKeys.js`?

swissspidy: I know we currently disallow products without images, but if this ever changes and `productImages` is an empty array here, this will silently fail.

This adds some hardening:

```suggestion
                thumbnail: productImages?.[0]?.url && {
                  src: productImages[0].url,
                  alt: productImages[0].alt,
                },
```

spacedmonkey: Done.

swissspidy (approved):

PR #11728, opened 2022-06-16
Shopping: Reset pulse animation when adding products 

### LINKED ISSUE #11705: Shopping: Reset pulse animation when adding products
## Feature Description

From bug bash:

> When you add multiple products, they pulse, but out of phase and it’s really bugging me that it is out of phase!

We might be able to solve this by passing the number of sibling elements as a prop to `<Display>` in `packages/story-editor/src/components/canvas/displayElement.js` and then in `ProductDisplay` use that as the `key` prop on `ShoppingTagDot` to force React to re-render all product dots when the number of products changes. This will cause the animations to restart.

timarney:
## Context

From bug bash:

> When you add multiple products, they pulse, but out of phase and it’s really bugging me that it is out of phase!

## Summary

Uses sibling count for ShoppingTagDot

See: https://github.com/GoogleForCreators/web-stories-wp/issues/11705#issue-1271399486

## Testing Instructions

Testing

1. Add multiple products to the stage -- note the dots pulse at the same time
2. Add a new dot -- note the dots pulse at the same time
3. Remove a dot -- note the dots pulse at the same time

No

No

No

## Checklist

---

Fixes #11705


packages/story-editor/src/components/canvas/displayLayer.js
```diff
@@ -56,13 +56,19 @@ const DisplayPageArea = styled(PageArea)`
 `;
 
 function DisplayPage({ pageElements, editingElement }) {
+  const siblingCount = pageElements.length - 1;
```
swissspidy: What I had in mind was for `siblingCount` to be the number of sibling elements _with the same type_. This way, we do not needlessly re-render products if another type of element is added/removed.

timarney: Just thinking `ProductDisplay` will be the only one at the moment that needs that prop ...

Would it be better to pass `elements` and then let ProductDisplay figure out it's siblings by type?

```javascript
ProductDisplay({ element, elements }) {

 const siblingCount =elements.filter(({ type }) => type === ELEMENT_TYPES.PRODUCT )
 
 ...
```

<hr>

Or would siblingCount be an object with a count for each type of element?

<hr>

swissspidy: If you pass `elements` to the component, the component will re-render eeevery time `elements` changes, which is way too often.

What I meant in my comment above is basically moving `const siblingCount = ....` to within the `.map()` callback a few lines below.

```diff
if (editingElement === element.id) {
  return null;
}
+
+const siblingCount = pageElements.filter(
+  ({ type }) => type === element.type)
+  .length - 1;

return (
-  <DisplayElement key={element.id} element={element} isAnimatable />
+  <DisplayElement
+    key={element.id}
+    element={element}
+    isAnimatable
+    siblingCount={siblingCount}
+  />
```

Then the consuming `ProductDisplay` component will only re-render when _products_ are added/removed.

The other elements won't use this prop, so they won't be affected.

swissspidy (approved):
Works well. I suppose the flicker one sees when adding/removing a product (and thus causing a re-render to all other products) is inevitable

spacedmonkey (approved):

swissspidy:
> Plugin builds for [e3ec5e4](https://github.com/GoogleForCreators/web-stories-wp/commit/e3ec5e4901c6074cb84c9ba1a819c0912fd2ee43) are ready 🛎️!
> 
> * Download [development build](https://storage.googleapis.com/web-stories-wp-github-artifacts/refs/pull/11728/merge/web-stories-dev.zip?099246ef1184f7f4f1596eb140137307f4589a4a)
> * Download [production build](https://storage.googleapis.com/web-stories-wp-github-artifacts/refs/pull/11728/merge/web-stories.zip?099246ef1184f7f4f1596eb140137307f4589a4a)

@felipebochehin87 Now this is ready for testing!

PR #11730, opened 2022-06-16
Editor: Fixed right-click in layer panel

### LINKED ISSUE #11717: Layers Panel: Support RCM on multi selection
## Feature Description

Currently, we don't support RCM for multiple layers selected, it will show a single item RCM (the one you clicked last).
It will be useful for grouping/locking multiple elements from the Layers Panel.

barklund:
Essentially just had to make sure that the reorderable only looks for regular left-clicks and ignores other mouse button clicks.

## Testing Instructions

1. Select multiple elements on stage in any possible way
2. Right-click (without pressing shift or meta at the same time) one of the layers in the layer panel and observe, that all the layers remain selected.

## Checklist

---

Fixes #11717

### DIFF
--- a/packages/story-editor/src/components/reorderable/useReordering.js
+++ b/packages/story-editor/src/components/reorderable/useReordering.js
@@ -36,9 +36,10 @@ function useReordering(onPositionChange, numChildren) {
   const handleStartReordering = useCallback(
     ({ position: currentPos, onStartReordering = () => {} }) =>
       (evt) => {
-        // Only allow reordering with non-modified click on non-background element.
+        // Only allow reordering with non-modified left-click on non-background element.
         // Modified (shift+ or meta+) clicks are for selection.
-        if (!evt.shiftKey && !evt.metaKey) {
+        // Right-clicks (button===2) are for context menu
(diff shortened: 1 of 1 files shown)


packages/story-editor/src/components/reorderable/useReordering.js
```diff
@@ -36,9 +36,10 @@ function useReordering(onPositionChange, numChildren) {
   const handleStartReordering = useCallback(
     ({ position: currentPos, onStartReordering = () => {} }) =>
       (evt) => {
-        // Only allow reordering with non-modified click on non-background element.
```
barklund, line 39: The "non-background element" part was left from before the _reorderable_ was generalized out of the layer panel, so this comment does not belong here anymore.

timarney:
Should the functionality work for `Control-click` on a mac to display the RCM?

Tested with a mouse and works as expected.

merapi (approved):
I tested it on my integration branch too - works great 🎉 
I think it's safe to say that this doesn't need any additional QA now and can be merged, either way, another round of QA will be conducted on the integration branch on Monday.

timarney (approved):
Re-tested --- works great.

PR #11736, opened 2022-06-17
Dashboard: FiltersProvider optimizations incorporate Authors filter

### LINKED ISSUE #11620: Dashboard: Consolidating the FilterProvider to include Author Filter

mwritter:
## Context

Consolidate dashboard filters into one provider.  Optimize FiltersProvider and incorporate the authors filter.

## Summary

Adds author filtering functionality to the FiltersProvider. Refactor existing use of authors filter. Update `datalist` component.

## Relevant Technical Choices

The filters now allow for caching and postponing primary option fetching with `getPrimaryOptions`.  This allows other components, like the `datalist`, to build UI with known initial data and update the `options` once they've been fetched or pulled from a cache.

## Testing Instructions

1. Go to Story Dashboard
2. Use filters
 - All taxonomy filters, and the Authors filter have changed
3. Check the Story Editor for any DropDown issues
 - The only one I noticed that was affected is Story Editor -> Document -> Publishing -> Author

No

No

No

## Checklist

---

Fixes #11620


packages/dashboard/src/app/views/myStories/filters/reducer.js
```diff
@@ -51,18 +51,31 @@ const reducer = (state, { type, payload = {} }) => {
       const filters = [...state.filters];
       filters[idx] = { ...filter, ...value };
 
+      // update filtersObject
+      const filtersObject = {};
+      for (const f of filters) {
```
sblinde: Nit: I'd prefer a more descriptive name than `f` for this, if `filter` isn't declared in the above scope, I feel like that would work.

sblinde: Nit: I'd prefer a more descriptive name than `f` for this, if `filter` isn't declared in the above scope, I feel like that would work.

packages/dashboard/src/app/views/myStories/filters/reducer.js
```diff
@@ -51,18 +51,31 @@ const reducer = (state, { type, payload = {} }) => {
...
       return {
         ...state,
         filters,
+        filtersObject,
       };
     }
 
     case types.REGISTER_FILTERS: {
       const { value } = payload;
+      const currentFilters = state.filters.map((f) => f.key);
```
sblinde: Nit: Same here w/ `f` and `v` if at all possible.

sblinde: Nit: Same here w/ `f` and `v` if at all possible.

packages/dashboard/src/app/views/myStories/filters/test/useAuthorFilter.js
```diff
@@ -0,0 +1,76 @@
...
+  it('should set authors to return value of getAuthors', async () => {
+    const { result } = renderHook(() => useAuthorFilter());
+    // flush promise queue
+    await act(() => Promise.resolve());
+
+    expect(result.current.authors).toHaveLength(1);
+  });
+
+  describe('initializeAuthorFilter', () => {
+    it('should initilize author filter data', async () => {
```
sblinde: Nit: `initialize`

sblinde, line 55: Nit: `initialize`

mwritter: my finger refuses to hit the 'a' every time😆

packages/dashboard/src/app/views/myStories/filters/reducer.js
mflorida: You could still use `filter` since it'll be scoped to the body of the `for` loop (and if the outer `filter` isn't used in the loop), but could cause confusion. I say use a capital `F`, with authority!

sblinde (approved):
🐝

Looked at it in QA env, seems solid! Working really well as far as I can see. Well done!

sblinde (approved):

packages/dashboard/src/app/views/myStories/filters/reducer.js
```diff
@@ -35,34 +35,49 @@ const reducer = (state, { type, payload = {} }) => {
   switch (type) {
     case types.UPDATE_FILTER: {
       const { key, value } = payload;
-      const filter = state.filters.find((f) => f.key === key);
+      const _filter = state.filters.find((filter) => filter.key === key);
```
spacedmonkey, line 38: Why change this?

mwritter: This was updated in accordance with the other comments by @sblinde on `filters/reducer` to be more readable. https://github.com/GoogleForCreators/web-stories-wp/pull/11736#discussion_r900405598, https://github.com/GoogleForCreators/web-stories-wp/pull/11736#discussion_r900405907.  Also the variable name `filter` is used around line 56 now - so one of them had to change.

sblinde:
@swissspidy @spacedmonkey Looks like QA passed and we seem to be good to go here. This PR has some other PRs that are dependent on it, so if either of you get a chance tomorrow during the day for you both to take another look, let us know and we can merge this in and get the others moving as well. Thanks for the feedback!

swissspidy (approved):
🚢  it

(after resolving the merge conflict)

packages/dashboard/src/app/views/myStories/filters/reducer.js
spacedmonkey: I don't feel like adding an underscore here makes it more readable. I would leave it as it was and just change line 56.

sblinde: @spacedmonkey My feedback was about not using letters alone as arguments within a function and to use a more descriptive word. It's not the change of adding an underscore that was suggested, but it is a way to tackle the change I suggested so that `filter` isn't declared as upper scope and so the argument is more descriptive than `f`.

sblinde: I understand your suggestion as well, but just want to note that it's important to me that we leave our code readable and understood by whoever may pick up on it in the future, so it's feedback I will always give to my team when I'm able to.

Issue #11631, opened 2022-06-06
The right arrow in the page carousel is sometimes hidden and unavailable


bmattb:
## Bug Description

If a user has a large enough number of pages in thier story to warrant the presentation of the left and right navigation arrows, sometimes the right arrow is hidden under other app elements and not able to be selected.

In the image below you can see that the right arrow is hidden under the Fit button and cannot be used.

![Screen Shot 2022-06-06 at 1.01.37 PM.png](https://images.zenhubusercontent.com/5e8b274ee5c3c7dfddf4be62/108cab2c-d26c-48b0-bd7b-957ba2069022)

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

mflorida:
This may be resolved with PR #11737 - the width of the carousel was reduced to accommodate the wider update left and right button groups. The spacing between the thumbnails may need adjusting.

miina:
Closing as not able to repro currently.

PR #11737, opened 2022-06-17
Workspace: Footer button updates and related refactoring

### LINKED ISSUE #8086: Make Carousel Grid Section (more) Responsive
## Feature Description

Carousel area is getting pretty crowded and needs some love and attention. We've added in a help center and moved the pre publish checklist to be toggle buttons. 

For the time being, these buttons are stuck at their smaller breakpoint size (which is supposed to be at 700px screen size).

To make this truly responsive for translations and screen sizes it's a bigger lift than just adding a button for the prepublish checklist which is the case for release 1.9. 

This will need some finesse with design and eng to make sure that the carousel grid section resizes appropriately at breakpoints and that the grid calculation that's getting done to handle the column size isn't run excessively and updates correctly. 

## Additional Context

Tagging Prometheus and Pea Pod because either of our crews can do this and it's good info for us all to have. 

See this thread: https://xwp.slack.com/archives/C01ESBB8ZN1/p1624570880176300
---

### LINKED ISSUE #10891: Workspace: Updated Footer Toggle Button Style
## Feature Description

This came up when redesigning the layers panel and moving it to the footer to be a toggle button instead of living in the inspector pane. 

A new toggle button style was introduced with it. Ideally, the toggle buttons used in the footer are uniform. 

| New | Existing | 
| --- | --- | 
| Screen Shot 2022-03-11 at 1 43 26 PM | Screen Shot 2022-03-11 at 1 44 18 PM |

The main differences here are the background color and circle for the number badge and a chevron.

## Additional Context

Let's figure out how to unify these before working on them. 

Related work #10847

mflorida:
## Context

The buttons in the footer of the editor were to be updated with a 'badge' style used for the notification count. The Layers button also needed a disclosure (chevron) icon added to indicate to users that this was a menu-type UI element.

## Summary

Styling for the buttons was updated, disclosure icon added to the Layers button, and carousel width was adjusted to prevent interference with wider buttons.

## Relevant Technical Choices

- Footer toggle button component was refactored to more robustly handle child components/elements/text
- Style rules for icons were moved to the toggle button component since they shared the same styling
- `hasMenuList` prop was added to the `ToggleButton` component for conditional rendering of the chevron disclosure icon
- Styling of the Zoom button was updated to match the other footer buttons (it uses a different core component with different styling)
- Adding the 'badge' styling and disclosure icon to the Layers button increased the widths of the left and right areas in the footer, so the carousel grid sizing was adjusted to accommodate this. This resulted in a narrower 'window' to show page thumbnails, mainly affecting stories with a large number of pages and narrower browser viewports.
- General cleanup of sizing and spacing in footer button contents

## To-do

- Consider different layout of footer buttons to allow a wider carousel view.
- The `Disclosure` and `Badge` components should be extracted for re-use.
- Hard-coded values may need to be updated with constants defined in the design system.

## User-facing changes

Before (1440px viewport)
Screen Shot 2022-06-17 at 12 54 45 PM

After (1440px viewport)
Screen Shot 2022-06-17 at 12 57 48 PM

After (1920px viewport)
Screen Shot 2022-06-17 at 1 11 38 PM

## Testing Instructions

- Ensure functionality of footer buttons is retained
- Ensure consistent styling of footer buttons
- Check other `ToggleButton` component instances in the UI to ensure proper styling

No.

No.

No.

## Checklist

---

Fixes #10891 
Related to #8086 (review suggested)
Related to #11631 (review suggested)


swissspidy:
Screenshot 2022-06-20 at 14 29 31

Tip: Reference one issue number per line

Words like "Fixes" and "resolve" are special keywords on GitHub that will auto-close the referenced issues. _However_, if you write `Resolve #123 #and #456`, only 123 will get closed, not 456.

So you will have to write:

```
Fixes #123
Fixes #456
Fixes $789
```

packages/design-system/src/components/dropDown/select/components.js
```diff
@@ -91,27 +91,30 @@ SelectButton.propTypes = {
...
+    width: auto;
+    color: ${theme.colors.fg.secondary};
+    transition: transform 100ms;
 
-    ${isOpen &&
-    css`
+    &.open {
       transform: rotate(180deg);
-    `}
+    }
```
barklund: We don't normally use class names with S-C.

We would do this using either a property:

```js
const DisclosureIcon = styled(ChevronDownSmall)(
  ({ theme }) => css`
  ...
  ${({ isOpen}) => isOpen && 'transform: rotate(.5turn);'}
`;
```

Or two different elements:

```js
const DisclosureIconOpen = styled(ChevronDownSmall)...;
const DisclosureIconClosed= styled(DisclosureIconOpen)`
  transform: rotate(.5turn);
`;
```

Or just an inline style directly on the element:

```js
<DisclosureIcon style={isOpen ? {transform: 'rotate(.5turn)'} : {}} />
```

Or the new fourth option we've been using in a few places: CSS Variable:


```js
const DisclosureIcon = styled(ChevronDownSmall)(
  ({ theme }) => css`
  ...
  --chevron-rotation: 0;
  transform: rotate(var(--chevron-rotation));'}
`;
...
<DisclosureIcon style={isOpen ? {'--chevron-rotation': '.5turn'} : {}} />
```
I think we should stick to one of those and not start mixing in a new style like this.  I think all of the above can be used for this particular instance. The first method would cause a new style to be injected into the DOM the first time a button is opened, but after that, it would reuse existing classes.

I can see how this class name based approach would make sense though, as it is pretty clean and fast, it is just not the way we do this normally across all the existing components.

packages/design-system/src/components/dropDown/select/components.js
```diff
@@ -91,27 +91,30 @@ SelectButton.propTypes = {
...
   `
 );
-ChevronWrap.propTypes = {
+
+export const Disclosure = ({ className = '', isOpen = false, ...other }) => (
+  <DisclosureIcon
+    className={`${className} ${isOpen ? 'open' : ''}`.trim()}
```
barklund: If we do decide to start using class names in this way, I think it makes sense to use the tiny utility library [`classnames`](https://www.npmjs.com/package/classnames) to clean things like this up:

```js
import cx from 'classnames';
...
  <DisclosureIcon className={cx(className, {open: isOpen})} {...other} />
```

It is probably worth a discussion.

packages/story-editor/src/components/footer/zoomSelector/zoomSelector.js
```diff
@@ -104,6 +128,7 @@ function ZoomSelector() {
       selectedValue={zoomSetting}
       popupFillWidth={false}
       selectButtonStylesOverride={selectButtonCSS}
+      className={'zoom-selector-dropdown'}
```
swissspidy: Nit:

```suggestion
      className="zoom-selector-dropdown"
```

packages/design-system/src/components/dropDown/select/components.js
barklund: Looking at this component again, does it ever receive the `className` property? Is it used as `styled(Disclosure)` somewhere I'm not aware of?

packages/story-editor/src/components/toggleButton/index.js
```diff
@@ -35,57 +39,135 @@ import { forwardRef } from '@googleforcreators/react';
...
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    color: ${theme.colors.fg.primary};
+    background-color: ${theme.colors.bg.tertiary};
+    border-radius: calc(${resolveSizeUnits($size)} / 2);
+    font-size: ${resolveSizeUnits($fontSize)};
+    line-height: 0;
+  `}
 `;
```
swissspidy: Could you perhaps elaborate on `resolveSizeUnits` and why we need it? I don't think I fully grasp it

Aside: the function name is confusing for what it does.

packages/story-editor/src/components/toggleButton/index.js
```diff
@@ -35,57 +39,135 @@ import { forwardRef } from '@googleforcreators/react';
...
+
+        .badge {
+          margin: 0 8px;
+        }
+      `
+    }
+
+    .badge.open,
+    &:hover .badge,
+    &:focus .badge {
```
barklund: Again, there's a lot of classes added here. Is this the best way to do this?

packages/design-system/src/components/dropDown/select/components.js
```diff
@@ -91,27 +91,30 @@ SelectButton.propTypes = {
...
+    {...other}
+  />
+);
```
swissspidy: Could you perhaps elaborate on this change to use an `open` class name instead of the existing way of just checking the prop? Seems unnecessary.

We rarely use `className` elsewhere since most of the things can be done with the functionality provided by styled-components and this seems like no different.

packages/story-editor/src/components/toggleButton/index.js
```diff
@@ -35,57 +39,135 @@ import { forwardRef } from '@googleforcreators/react';
...
+    color: ${theme.colors.fg.secondary};
+    transition: transform 100ms;
+
+    &.open {
+      transform: rotate(180deg);
+    }
+  `
+);
+
+// TODO: Extract 'Disclosure' to its own component
```
barklund: Yes please, this appears to be duplicated code.

barklund (commented):
This looks great and both the resulting UX in the editor as well as the resulting code in each component is very clean. I just have some concerns about how we're using conditional styles, as this seems to use a new method that we're not using elsewhere in the codebase.

packages/design-system/src/components/dropDown/select/components.js
mflorida: Noted. I will update as suggested.

packages/design-system/src/components/dropDown/select/components.js
mflorida: Using a CSS custom property is a good option - it's a cleaner approach to using variables within the styled components pattern.

packages/design-system/src/components/dropDown/select/components.js
mflorida: To also respond to comments from @swissspidy, using `className` to alter appearance based on a condition is not a common pattern in the codebase, and I can make changes based on that feedback. 

I do, however, think using a meaningful `className` value (that's not used for styling) to help identify components in the rendered output can be useful for development and tracking down bugs more easily.

packages/story-editor/src/components/toggleButton/index.js
mflorida: Old habits… It can be cleaner, but deviates from patterns used elsewhere.

packages/story-editor/src/components/footer/zoomSelector/zoomSelector.js
mflorida: This is the default pattern used in my IDE, but will modify to use double-quotes.

Can requiring double-quotes be added as an eslint rule?

packages/design-system/src/components/dropDown/select/components.js
swissspidy: > I do, however, think using a meaningful `className` value (that's not used for styling) to help identify components in the rendered output can be useful for development and tracking down bugs more easily.

We do use `babel-plugin-styled-components` ([repo](https://github.com/styled-components/babel-plugin-styled-components), [docs](https://styled-components.com/docs/tooling#babel-plugin)) in the project to automatically have more meaningful and deterministic class names based on the component name. For example, here I know this is the output of the `StyledStoryDisplayContent` component:

Screenshot 2022-06-20 at 19 50 54

 If that is not sufficient for development and debugging, I am all ears for improving that experience and happy to discuss that in a dedicated forum. Though this should be done with the whole code base in mind and not selectively for a handful of components.

packages/story-editor/src/components/toggleButton/index.js
mflorida: This allows use of non-pixel values to be used. If the argument is numeric (either string or number), the value will be returned with `px` appended, where a value like `100%` will be returned as-is. This is similar to React's behavior of appending `px` to numbers used as values of `style` objects used for inline styling. In the `CountBadge` component, it allows use of `calc` with non-px values.

swissspidy: Is there an example here where this is used, i.e. something like `100%` (or another non-numeric arg) is being passed?

Jus trying to better understand this since this introduces a new pattern to the code base.

packages/design-system/src/components/dropDown/select/components.js
mflorida: I understand and agree with your point about consistency. This may be more of a concern related to the patterns required to use Styled Components than anything specific to this project.

A lot of the classNames generated by Styled Components can be redundant and hard to track up the component tree. For example, classNames like `components__Container-sc-xxxxxx` and `components__CardWrapper-sc-xxxxxx` are generic and lose meaning because they're repeatedly redefined in components to accommodate the patterns required for composition using the Styled Components library.

Going up the component tree offers a little context, but navigating the codebase to find specific instances of general-use components isn't obvious (maybe just to developers new to the codebase and patterns used, myself included).

I hope that makes sense.

packages/story-editor/src/components/toggleButton/index.js
mflorida: 🤔 I'll do some soul searching (aka try to remember my original intention).

packages/design-system/src/components/dropDown/select/components.js
swissspidy: It does, thanks a lot for sharing your feedback & experience! It's much appreciated.

Definitely agree that there's room for improvement in this area, espcially for generic class names like "Container" or "Wrapper". Those are good examples of names that we see across the code base.

I've now opened #11749 to further discuss this and brainstorm on ideas.

packages/design-system/src/components/dropDown/select/components.js
mflorida: Creation of a separate `Disclosure` component resolves this.

packages/design-system/src/components/dropDown/select/components.js
mflorida: Resolved with creation of a separate `Disclosure` component.

packages/story-editor/src/components/toggleButton/index.js
mflorida: Done

packages/design-system/src/components/dropDown/select/components.js
mflorida: I think more thoughtful use (or non-use) of 'wrapper' components could go a long way towards more readable and maintainable code. 🤓

mflorida: > Looking at this component again, does it ever receive the `className` property? Is it used as `styled(Disclosure)` somewhere I'm not aware of?

By modifying className dynamically, styling can be defined in a parent component, eliminating the need to call `styled(Disclosure)` and passing props to define styling. If a parent already has access to a prop used for styling, it doesn't need to be passed on to affected children. Hopefully I explained that properly. I can share a code snippet to show what I mean. One possible issue that comes to mind with this approach would be unintentional styling of child components that could have the className value applied. I think the natural CSS cascade would handle this, but there could be specificity issues that would interfere. 🤔

packages/story-editor/src/components/toggleButton/index.js
mflorida: Updated to use only 'px' values.

packages/story-editor/src/components/toggleButton/index.js
mflorida: @barklund In extracting this component, I applied negative margins to the icon `svg` because the actual graphic has a lot of empty space around it. For regular icons, it makes sense to keep the margin as-is, but I feel like spacing of a little disclosure chevron should only occupy as much space as the graphic itself.

The menus on the dashboard were affected by the styling change, but IMO, the padding should be applied to the menu, not the contents.

image

Thoughts?

mflorida:
> Screenshot 2022-06-20 at 14 29 31
> 
> Tip: Reference one issue number per line
> 
> Words like "Fixes" and "resolve" are special keywords on GitHub that will auto-close the referenced issues. _However_, if you write `Resolve #123 #and #456`, only 123 will get closed, not 456.
> 
> So you will have to write:
> 
> ```
> Fixes #123
> Fixes #456
> Fixes $789
> ```

I used 'Relates to' because there should probably be a review involved.

packages/design-system/src/components/dropDown/select/components.js
barklund: > > Looking at this component again, does it ever receive the `className` property? Is it used as `styled(Disclosure)` somewhere I'm not aware of?
> 
> By modifying className dynamically, styling can be defined in a parent component, eliminating the need to call `styled(Disclosure)` and passing props to define styling. If a parent already has access to a prop used for styling, it doesn't need to be passed on to affected children. Hopefully I explained that properly. I can share a code snippet to show what I mean. One possible issue that comes to mind with this approach would be unintentional styling of child components that could have the className value applied. I think the natural CSS cascade would handle this, but there could be specificity issues that would interfere. 🤔

Yeah I know how it works. but [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) dictates that we should only add functionality we actually need. And if this component never needs restyling, it's just useless code. That's my point, sorry if that wasn't clear.

mflorida:
### NOTE

There may be styling issues in menus with 'chevron' icons, which are fixed in #11770 (No PR for that yet).

packages/design-system/src/components/disclosure/index.js
```diff
@@ -0,0 +1,73 @@
...
+ * @return {*} Disclosure icon component
+ */
+export const Disclosure = ({
+  direction = 'down',
+  duration = 0,
+  isOpen = false,
+  rotation,
+  ...rest
+}) => {
+  const [whenClosed, whenOpen] = rotation ?? rotate[direction];
```
barklund, line 54: Won't this fail if rotation is unset? Not sure what purpose `rotation ??` then serves? Maybe `rotate.down` could be default?

barklund (approved):
This looks really good and greatly simplifies the code. Great effort!

packages/design-system/src/components/disclosure/index.js
barklund: Ah, now I get it. You either specify a `direction` (which defaults to `"down"`) or a manual `rotation` tuple. Is the latter even relevant?

packages/design-system/src/components/disclosure/index.js
```diff
@@ -0,0 +1,73 @@
...
+
+  const iconStyle = css`
+    height: ${THEME_CONSTANTS.ICON_SIZE}px;
+    width: auto;
+    margin: 0 -10px;
+    color: ${({ theme }) => theme.colors.fg.secondary};
+    transition: transform ${duration};
+    transform: rotate(${isOpen ? whenOpen : whenClosed}deg);
+  `;
```
barklund, line 63: Can we move all the static parts to a class? This is way more inline CSS than necessary I feel. Only the transition and transform really need to be inline. And the transition could be a dynamic part of the class, as it won't change in the lifetime of each disclosure instance, I think?

packages/design-system/src/components/disclosure/index.js
mflorida: I did it this way in case you wanted a 90 degree rotation, or any other arbitrary start and end values, but then I set ‘left’ and ‘right’ directions to point down when open, so the rotation prop may not be necessary after all. It could always be added in the future if necessary.

packages/design-system/src/components/disclosure/index.js
mflorida: This is passed into the `css` prop which is a prop specific to styled components and will be extracted into a class. Pretty slick feature of that library.

packages/design-system/src/components/disclosure/index.js
mflorida: Calling out my YAGNI tendencies are appreciated. 🙂

packages/design-system/src/components/dropDown/select/components.js
mflorida: I totally agree, it’s just a matter of where component-specific styling is defined - wrapped in `styled(AwesomeComponent)` or defined in a parent component with a selector to target `AwesomeComponent`. This is used elsewhere in the codebase, but with element or pseudo selectors.

PR #11738, opened 2022-06-17
Dashboard: Add No Results Message To Filters 

### LINKED ISSUE #11664: Dashboard: Add No Results Message To Filters
## Feature Description

In the case in which a filter is used and no results are found, the UI should give feedback to the user that there were no results found for the filter. 

## Acceptance Criteria

- When no results are found for an active filter search on the Dashboard, the UI should indicate `Sorry, we couldn't find any results.`

## Additional Context

This feedback was a part of https://github.com/GoogleForCreators/web-stories-wp/pull/11625#issuecomment-1150104862

Additionally, it may be easier to accomplish this logic once #11620 has been completed.

sblinde:
## Context

As part of the work slated to improve Dashboard filtering, this adds a consistent "Sorry, we couldn't find any results." when a user has searched and used author and/or taxonomy filters, or keyboard search.

✅  Important note: This work relies on the changes in `feature/11620_FiltersProvider_add_authors_filter` as a branch to verify this works as intended. When #11620 has been merged, this can point towards the `main` branch instead of that feature branch.

## Summary

Note: Updated screenshots as of June 22nd, 2022
**Example of keyword search returning no results:**
Screen Shot 2022-06-22 at 10 10 19 AM

**Example of filters search returning no results:**
Screen Shot 2022-06-22 at 10 10 27 AM

**Example of no stories to display with no filters and no keywords enabled:**
Screen Shot 2022-06-22 at 10 10 34 AM

## Relevant Technical Choices

This imports the `filtersObjects` data and uses it to determine whether a filter has been enabled. If the `filterObjects` is empty, we haven't utilized any of the `FilterProvider` filters. If it contains items, and there are no stories to display, we can safely assume that the `No results` message should display.

The same logic is used for the `Explore Templates` button, which should only display when there are no stories created at all.

## To-do

Nothing.

## User-facing changes

This makes the `no results found` message consistent, regardless of whether a keyword search was used, or if multiple filters have been used.

## Testing Instructions

1. Navigate to the Dashboard
2. Use the filters and/or keyword search to return a result that has no matches
3. See the "Sorry, we couldn't find any results." message specific to whether a keyword was used
4. The "Explore Templates" button should only display if there are no stories to view

No.

No.

No.

## Checklist

- [] I have verified accessibility to the best of my abilities ([docs](https://github.com/googleforcreators/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [X] I have added documentation where necessary

---

Fixes #11664


packages/dashboard/src/app/views/myStories/content/index.js
```diff
@@ -86,15 +87,8 @@ function Content({
...
-                ? sprintf(
-                    /* translators: %s: search term. */
-                    __(
-                      'Sorry, we couldn\'t find any results matching "%s"',
-                      'web-stories'
-                    ),
-                    search.keyword
-                  )
+              {search?.keyword || Boolean(Object.keys(filtersObject))
+                ? __("Sorry, we couldn't find any results.", 'web-stories')
```
spacedmonkey: I think it is better we keep on the old message.

swissspidy: IIRC my original suggestion was:

If there is a search term: Use existing message `Sorry, we couldn't find any results matching "%s"`
If there is a filter: Use `Sorry, we couldn't find any results`

Note: no period at the end of either message

sblinde: Sounds good! I can implement that per the specific cases. Thanks!

packages/dashboard/src/app/views/myStories/content/index.js
```diff
@@ -88,15 +89,8 @@ function Content({
...
```
spacedmonkey:
```
const noResultsMessage = search?.keyword ? sprintf(
                    /* translators: %s: search term. */
                    __(
                      'Sorry, we couldn\'t find any results matching "%s"',
                      'web-stories'
                    ),
                    search.keyword
                  ) : __("Sorry, we couldn't find any results.", 'web-stories');
```

Then 
```
{search?.keyword || Boolean(Object.keys(filtersObject)) ? noResultsMessage : __('Start telling Stories.', 'web-stories')}
```

packages/dashboard/src/app/views/myStories/content/index.js
sblinde: See: https://github.com/GoogleForCreators/web-stories-wp/pull/11738/commits/96e74e155faf9ff15d99f24a4862b381e680192c

packages/dashboard/src/app/views/myStories/content/index.js
sblinde: Saw this after I started refactoring this, but this should address the logic required as well as the button logic (the button that suggests looking at templates in the case of no stories being made yet) https://github.com/GoogleForCreators/web-stories-wp/pull/11738/commits/96e74e155faf9ff15d99f24a4862b381e680192c

sblinde:
@spacedmonkey @swissspidy Feedback addressed, also changed the button logic as a part of the last commit. There are updated screenshots as well in the description (and additional QA notes for the button logic). Thanks!

packages/dashboard/src/app/views/myStories/content/index.js
sblinde: If it's better to just have the idea you suggested instead of rendering the text the way I did it in the commit, let me know, it's a quick refactor to do.

packages/dashboard/src/app/views/myStories/content/index.js
```diff
@@ -45,10 +45,29 @@ import {
...
+NoAvailableContent.propTypes = {
+  filtersObject: PropTypes.object,
+  keyword: PropTypes.string,
+};
+
 function Content({
   allPagesFetched,
   canViewDefaultTemplates,
   filter,
+  filtersObject = {},
```
mwritter, line 70: 🤔 You could pull this straight from the `FiltersProvider` instead of passing props - but not totally necessary

mwritter (approved):

packages/dashboard/src/app/views/myStories/content/index.js
```diff
@@ -18,7 +18,7 @@
  * External dependencies
  */
 import PropTypes from 'prop-types';
-import { __, sprintf } from '@googleforcreators/i18n';
+import { sprintf, __ } from '@googleforcreators/i18n';
```
spacedmonkey, line 21: ?

packages/dashboard/src/app/views/myStories/content/index.js
sblinde: Let me see, that could be easy to do

packages/dashboard/src/app/views/myStories/content/index.js
sblinde: As part of the feedback given, I had to use `sprintf` again, and I just put it in a different order when I imported it to address feedback vs. the original code... that's all.

spacedmonkey (approved):

sblinde:
This should be ready for QA, but it still pointed to the branch it depends on. Merging (and pointing to `main`) will be done once https://github.com/GoogleForCreators/web-stories-wp/issues/11620 has been merged!

packages/dashboard/src/app/views/myStories/content/index.js
spacedmonkey: So just revert that change, leave in it's original order.

swissspidy: IDK, the order doesn't matter and the change is already done. No big deal.

PR #11739, opened 2022-06-20
Use native eyedropper API if available

### LINKED ISSUE #9478: Use native eyedropper API if available
## Feature Description

The EyeDropper API enables developers to use a browser-supplied eyedropper in the construction of custom color pickers.

Chrome has started shipping this new API in v95. See https://www.chromestatus.com/feature/6304275594477568

It would be nice if we could use that one if available, so we don't have to use html-to-image for it.

Demo:

https://twitter.com/tomayac/status/1432969851811794944
https://codepen.io/matanbobi/pen/GREoZvo

timarney:
## Context

The EyeDropper API enables developers to use a browser-supplied eyedropper in the construction of custom color pickers.

If available use the native browser EyeDropper in place of the current `html-to-image` EyeDropper. 

|  `html-to-image` EyeDropper   |  EyeDropper API (native)  |  
|---|---|
| Screen Shot 2022-06-20 at 6 28 39 AM  | Screen Shot 2022-06-20 at 6 30 40 AM  Demo 👉  https://www.youtube.com/watch?v=V5n7pgiQc9I | 

## Summary

Adds code to check if the EyeDropper API is supported.  In which case it will be used when opening the EyeDropper. 

There's currently support in Chrome and Edge.

https://caniuse.com/mdn-api_eyedropper

## Testing Instructions

1. Create a new story 
2. Add some text and an image to the stage (to sample colors)
3. Use the eye dropper tool to update the color
4.  Select the background layer 
5. Select the style panel
6. Update the background color via the "Page Background" section

No

No

No

## Checklist

---

Fixes #9478


packages/story-editor/src/components/eyedropper/useEyeDropperApi.js
```diff
@@ -0,0 +1,70 @@
...
+    c = '0x' + c.join('');
+    return {
+      r: (c >> 16) & 255,
+      g: (c >> 8) & 255,
+      b: c & 255,
+      a: 1,
+    };
+  }
+  throw new Error(`Unable to convert to hex ${hex}`);
+}
```
swissspidy: Let's use `getSolidFromHex` instead of some code copied from CodeSandbox

a) because of licenses, b) because we already have code for this

packages/story-editor/src/components/colorPicker/basicColorPicker.js
```diff
@@ -187,8 +193,12 @@ function BasicColorPicker({
...
               size={BUTTON_SIZES.SMALL}
               aria-label={__('Pick a color from canvas', 'web-stories')}
-              onClick={initEyedropper()}
-              onPointerEnter={initEyedropper(false)}
+              onClick={
+                isEyeDropperApiSupported ? openEyeDropper : initEyedropper()
+              }
+              onPointerEnter={
+                isEyeDropperApiSupported ? noop : initEyedropper(false)
+              }
```
barklund: Could we not abstract this into a single hook? So this just calls `handleDropperInitialize` and `handleDropperClick` respectively? That would reduce duplicated code a bit too.

packages/story-editor/src/components/eyedropper/useEyeDropperApi.js
```diff
@@ -0,0 +1,51 @@
...
+ */
+/**
+ * External dependencies
+ */
+import { useEffect, useCallback, useRef } from 'react';
+import { getSolidFromHex } from '@googleforcreators/patterns';
+
+function useEyeDropperApi({ onChange }) {
+  const isEyeDropperApiSupported =
+    (typeof window !== 'undefined') & ('EyeDropper' in window);
```
barklund: `&&`, I presume?

packages/story-editor/src/components/eyedropper/useEyeDropperApi.js
```diff
@@ -0,0 +1,51 @@
...
+
+  const eyeDropper = useRef(false);
```
barklund: It's better to initialize it to `null` to make the type somewhat consistent.

packages/story-editor/src/components/eyedropper/useEyeDropperApi.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    try {
+      const { sRGBHex } = await eyeDropper.current.open();
+      onChange(getSolidFromHex(sRGBHex.substring(1)).color);
+    } catch (e) {
+      //eslint-disable-next-line no-console -- Surface error for debugging.
+      console.log(e.message);
+    }
+  }, [eyeDropper, isEyeDropperApiSupported, onChange]);
```
barklund: If `onChange` is a dependency (which it should be) it should be memoized in the parent component using it via `useCallback`. Currently `onChange` is re-defined on every render, so this whole function runs on every render making the memoization here useless.

packages/story-editor/src/components/colorPicker/basicColorPicker.js
```diff
@@ -127,6 +129,10 @@ function BasicColorPicker({
     onChange: (newColor) => handleColorChange({ color: newColor }),
   });
 
+  const { isEyeDropperApiSupported, openEyeDropper } = useEyeDropperApi({
+    onChange: (newColor) => handleColorChange({ color: newColor }),
```
barklund: Yeah, if the two hooks both just invoked their callbacks with `{color}`, both components could be simplified a lot. But I'd still like to see a new third hook in between the component and the two underlying hooks abstracting this out another level.

timarney:
Karma test for the EyeDropper on this PR ...

After a bunch of debugging I tried creating a standalone HTML page and a test using Puppeteer  --- 

See:

https://www.youtube.com/watch?v=gDtnM01dkhU

Looks like the EyeDropper doesn't track with the programatic click i.e. 

```
await page.mouse.click(rect.x + offset.x, rect.y + offset.y);
```

Same for Karma from what I can tell.

packages/story-editor/src/components/colorPicker/karma/eyedropper.karma.js
```diff
@@ -30,6 +30,12 @@ describe('Eyedropper', () => {
 
   beforeEach(async () => {
     fixture = new Fixture();
+
+    // removed to run test against htmlToImage EyeDropper vs API
+    // see issue testing with native API here
+    // https://github.com/GoogleForCreators/web-stories-wp/pull/11739#issuecomment-1162367409
+    delete window.EyeDropper;
```
timarney, line 51: Nothing this ☝️ 

https://www.youtube.com/watch?v=gDtnM01dkhU

barklund (commented):
This functionality is awesome. And truth be told, it does work a bit better than our own implementation 🙈.

I just have some nits about structure - I think the hooks can be better abstracted for a cleaner API.

packages/story-editor/src/components/colorPicker/karma/eyedropper.karma.js
swissspidy: Probably needs a cleanup in `afterEach`? (Or better in `beforeAll`/`afterAll`)

packages/story-editor/src/components/colorPicker/basicColorPicker.js
timarney: Okay I made some updates to just use the callback that's already being passed in https://github.com/GoogleForCreators/web-stories-wp/pull/11739/commits/754904d0bd2077f0941ff0457f0168aab4dacc2e I'll refactor some more to move to either a third hook or combined hook to simplify / remove the duplication.

merapi (commented):
I think we should utilize `isEyedropperActive` when using this new API to make the view less obstructed.

New API (zoom circle cannot be captured):
Screenshot 2022-06-22 at 14 58 16

Old Eyedropper (same action, border color change):
image

timarney:
@merapi I updated this to now use `isEyedropperActive` --- there ends up being a bunch of code on `eyedropperLayer.js` that isn't needed when the API is available.  So you think we should also do some early returns to avoid things like the on MouseMove callbacks?

https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/story-editor/src/components/canvas/eyedropperLayer.js#L216

merapi:
`isEyedropperActive` updates 👍🏻 

> there ends up being a bunch of code on eyedropperLayer.js that isn't needed when the API is available. So you think we should also do some early returns to avoid things like the on MouseMove callbacks?

Hm, don't we want to just do this in `eyedropperLayer.js`?:
```js
  ...
  useGlobalKeyDownEffect('esc', closeEyedropper);

  if (isEyeDropperApiSupported) {
    return null;
  }
  ...
```

packages/story-editor/src/components/canvas/eyedropperLayer.js
```diff
@@ -272,17 +275,21 @@ function EyedropperLayer() {
...
-                width={MAGNIFIER_SIZE}
-                height={MAGNIFIER_SIZE}
-              />
-            </Circle>
-            <ColorInfo ref={magnifierColor} />
-          </Magnifier>
+          {!isEyeDropperApiSupported && (
+            <CanvasImage ref={imgRef} src={img} alt="" />
+          )}
+          {!isEyeDropperApiSupported && (
```
barklund: Surely these can be combined into one condition?

barklund (approved):
Looks good and works great!

packages/story-editor/src/components/canvas/eyedropperLayer.js
timarney: 👍 

https://github.com/GoogleForCreators/web-stories-wp/pull/11739/commits/3332f842ee7ce381ee91e27ebd53de94012916a4

merapi: Nit: We can even do an early return if `isEyeDropperApiSupported ` is `true` like suggested [here](https://github.com/GoogleForCreators/web-stories-wp/pull/11739#issuecomment-1164329214) and skip this part, `eyedropperLayer` is not needed at all if we have the native API.

merapi (approved):
Tiny nit to consider but otherwise 👍🏻

packages/story-editor/src/components/canvas/eyedropperLayer.js
timarney: Updated:

https://github.com/GoogleForCreators/web-stories-wp/pull/11739/commits/a155bb7c0a75302e30111aaada746451d9fa3d74

timarney:
@merapi code is updated with that early return now.

timarney:
QA
https://github.com/GoogleForCreators/web-stories-wp/issues/9478#issuecomment-1169908016

PR #4310, opened 2020-09-02
Makes retrieving media from local api bypass preloaded cache when WP Upload modal closed.

ezxyz:
## Summary

the preloader is intercepting the request `web-stories/v1/media?context=edit&per_page=100&page=1&_web_stories_envelope=true`
Instead of using the request that is cached we should modify it by appending cache_bust=true when called via after upload media

## Relevant Technical Choices
using a cache_bust parameter for the api call from resetWithFetch (triggered onClose for the WP Media modal).

## User-facing changes

Local media pane refreshes properly when new media is uploaded via the wp media modal. (see #4109)

## Testing Instructions

Upload via local media pane "upload button". Close the modal. The pane should have the new media displayed. 

---

Fixes #4109

Issue #4109, opened 2020-08-20
Media library not refreshing after uploading via WP media modal


csossi:
## Bug Description

A newly uploaded image is appearing immediately in "Images" media category, but not in "All" category

## Expected Behaviour

Expected newly uploaded image to appear under both "Images" and "All"

## Steps to Reproduce

- create new story
- select "Upload" from media panel
- browse for and upload an image
- insert this new image into page

- note that image isn't appearing under "All" category (even though it IS appearing in "Images" category

## Screenshots

Newly uploaded image is on the canvas here, but isn't in media "all" section:
![image](https://user-images.githubusercontent.com/16945664/90705690-8936da80-e261-11ea-9472-0a412049589d.png)

However, you can see it here when viewing just "Images":
![image](https://user-images.githubusercontent.com/16945664/90705757-abc8f380-e261-11ea-8bd1-666302ddefeb.png)

(Note: If I navigate away from editor and return, then the image WILL appear under "All")

## Additional Context

- Plugin Version: Testing in PR#4012
- Operating System: Win10
- Browser: Chrome

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

swissspidy:
Seems like a duplicate of #1201 — either one should be closed. Need to check with media pod though.

swissspidy:
Oh, and not a stable blocker IMO

bmattb:
@o-fernandez  - can you confirm that this is not a blocker to Beta 2 release today

o-fernandez:
Agree, not a blocker for Beta. But let's take a look at it for Stable. @diegovar would it be possible to TAL at this?

swissspidy:
Estimate seems a bit high. Perhaps @spacedmonkey could be of assistance here regarding the media modal?

ezxyz:
See: https://xwp.slack.com/archives/CCCAUAH6F/p1598962509042200 for discussion.

TLDR: This is possibly due to a PHP issue, but would like someone who knows PHP to confirm.

ezxyz:
So the conclusion from the discussion is that it is because the initial request is being cached so we just need to add a parameter like `cache_bust=true` to the api request for 'All" media.

PR #11759, opened 2022-06-20
Media: Remove manual cache busting

### LINKED ISSUE #4312: Media: Remove manual cache busting
## Task Description

In the editor we currently preload media items using the `wp.apiFetch` preloading middleware.

It is basically works as a cache that can’t be invalidated. 

So if `/path/to/load/media` is preloaded, it's not possible to request `/path/to/load/media` again to fetch new data as it will always use the cached data.

This causes issues like #4109 and requires solutions like #4310.

One possible alternative solution mentioned by @jhtin is to make the initial call look different from the subsequent calls. Then you preload the first call that looks different and then every subsequent call isn’t cached. Basically the opposite of what #4310 dos.

Essentially, what we need is a way to **pre-populate media state upon page load**, ideally one that is not so tightly coupled to WordPress, i.e. avoiding `wp.apiFetch` preloading middleware

timarney:
## Context

> We've added manual cache busting in the past because we had to support older WordPress versions.

> Now that we are bumping the WP version requirement to WP 5.7+ we should be able to finally remove this manual cache busting.

See: https://github.com/GoogleForCreators/web-stories-wp/issues/4312#issuecomment-1160536112

## Summary

Removes cache bust parameter where it has been applied 

## Testing Instructions

1.  Upload via local media pane "upload button". Close the modal. The pane should have the new media displayed (#4310)

2. #4109
- create new story
- select "Upload" from media panel
- browse for and upload an image
- insert this new image into page 
- ensure the new media shows under the "all" category and the "images" category for the media panel (no page refresh)

No

No

No

## Checklist

---

Fixes #4312


swissspidy:
Let's get the testing instructions from #4310 / #4109 in here to make sure there are really no regressions.

We might even be able to write an e2e test for this (media library refreshing after upload through media modal), WDYT?

timarney:
> We might even be able to write an e2e test for this (media library refreshing after upload through media modal), WDYT?

Likely ... that said looks like we're skipping some media tests already ya?

```javascript
it.skip('should insert an image by clicking on it', 
```

https://github.com/googleforcreators/web-stories-wp/issues/7107#issuecomment-817666571

spacedmonkey (approved):

swissspidy (approved):

PR #11763, opened 2022-06-20
Editor: Optimize story reducers using immer library

### LINKED ISSUE #11723: Try immer in the story reducer
## Feature Description

The story reducer has a lot of logic for making updates to the current story. This is implemented with manually de- and restructuring objects and arrays to avoid mutating objects. However, this creates a lot of overhead. Using immer would make this a lot simpler, as it allows mutating, but returns a new object as a result.

The process could potentially include making sure that all the current functionality is correctly tested and only then updated.

We should try it out in the story reducer and see how much it simplifies the code and how much it impacts performance (if any).

barklund:
## Context

This converts the story reducers to using immer to make them easier to read and write (purely DX).

This uses all the new and extended unit tests for the reducers to make sure the conversion is done correctly.

## User-facing changes
 _None_.

## Testing Instructions

## Checklist

---

Fixes #11723


packages/story-editor/src/app/story/useStoryReducer/reducers/utils.js
```diff
@@ -87,9 +87,13 @@ export function updateElementWithUpdater(element, properties) {
...
-    return element;
+    return null;
   }
-  return { ...element, ...allowedProperties };
+  if (allowedProperties.animation) {
+    return allowedProperties.animation;
+  }
+  Object.assign(element, allowedProperties);
+  return null;
 }
```
barklund, line 97: Some of the remaining functions in this file could also be updated so that they work on the passed objects rather than return immutable clones.

mflorida:
This looks amazing! Definitely improves readability.

swissspidy:
Although I am not really familiar with `immer`, at first glance the change looks good, definitely makes things more readable.

How are your impressions so far? Would it be difficult to also adopt `immer` for our other reducers in the code base? Seems like a valuable thing for consistency.

PS. there are some merge conflicts now

packages/story-editor/src/app/story/useStoryReducer/reducers/arrangeElement.js
```diff
@@ -86,38 +89,21 @@ function arrangeElement(state, { elementId, position, groupId = false }) {
...
-  newPageElements[currentPosition] = {
-    ...newPageElements[currentPosition],
-  };
-  if (groupId !== false) {
-    newPageElements[currentPosition].groupId = groupId;
+  // Update group id on current element
+  if (groupId) {
+    page.elements[currentPosition].groupId = groupId;
+  } else {
+    delete page.elements[currentPosition].groupId;
```
merapi: Right now the `groupId` can be either a `string` = actual groupId (move into the group), `null/undefined` = remove the groupId (move out of the group) or `false` = don't update the id (move within the group/non-group).

` * @param {string|boolean} payload.groupId New group id (when moving in the Layers Panel), false to skip updating.`

packages/story-editor/src/app/story/useStoryReducer/reducers/combineElements.js
```diff
@@ -53,115 +54,87 @@ import { removeAnimationsWithElementIds } from './utils';
...
+    };
+
+    // Elements are now
+    page.elements = page.elements
+      // Remove first element if combining from existing id
+      .filter(({ id }) => id !== firstId)
+      // Update reference to second element
+      .map((el) => (el.id === secondId ? newElement : el));
+
+    // First element should always be the image getting applied to
```
timarney: These comments are really helpful - to give some back story into what is happening.

💯

barklund:
> How are your impressions so far? Would it be difficult to also adopt `immer` for our other reducers in the code base? Seems like a valuable thing for consistency.

I have used Immer before in hobby projects, and it really makes this kind of data wrangling much easier and cleaner. It is a very stable library and works really well at scale too.

The only potential "concern" I can think of in deploying it elsewhere is that Immer automatically deep-freezes the object when created, which might cause unintended bugs. Ideally our objects should be immutable in other places too, but if not, it might cause errors. We can fix this by either fixing our mutations or by disabling auto-freeze.

sblinde:
I've never used `immer` before but really liking how clean this looks, I'll have to check it out myself a bit further, but really seems like a great method to not mutate data, which I'm all for!

merapi (approved):
Looks good and should be safe to merge thanks to the good test coverage 💯 👍🏻

PR #11768, opened 2022-06-21
Unit tests: download AMP validator file beforehand

swissspidy:
## Summary

This should make running AMP validation JS unit tests more robust by first downloading the AMP Validator file locally in a temp folder so it's readily available for all tests.

## Testing Instructions

1. Running for example `npm run test:js packages/output/src/test/` multiple times locally should never error or time out.

No

No

No

## Checklist

---

Fixes #


.github/workflows/tests-unit-js.yml
```diff
@@ -66,10 +65,18 @@ jobs:
         env:
           PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: true
 
+      - name: Download AMP validator JS
+        id: amp_validator
+        run: |
+          cd $RUNNER_TEMP && curl -O -f -s -S https://cdn.ampproject.org/v0/validator_wasm.js
+          AMP_VALIDATOR_FILE="${RUNNER_TEMP}/validator_wasm.js"
+          echo "::set-output name=validator_file::$AMP_VALIDATOR_FILE"
```
spacedmonkey, line 73: This is a 7.2mb file. Anyway we can cache this?

spacedmonkey:
I am still seeing this error. 

```
  ● Text Element output › AMP validation › should produce valid AMP output when setting text letter-spacing

    Expected value   undefined
    Received:
      undefined

    Message:
      Assertion failed: WebAssembly is uninitialized

    Difference:

    Compared values have no visual difference.
        at new module$contents$goog$asserts_
```

In all the amp validation tests. 
Screenshot 2022-06-27 at 15 00 45

timarney (approved):
Ran multiple times not seeing any issues locally

swissspidy:
@spacedmonkey What's the full command you're using to run tests locally?

If you're using PhpStorm/IntelliJ to run the tests, you're most likely running `node_modules/.bin/jest ...` directly, in which case you will not benefit from the change here (which only works when using `npm run test:js`)

In this case it's up to you to change your test config to either use `npm run test:js` or manually execute the `npm run pretest:js` script in the "Before launch" section, which will then cache the validator script for you.

Like so:

Screenshot 2022-06-27 at 20 10 30

.github/workflows/tests-unit-js.yml
swissspidy: You mean between runs or ..?

I don't see an issue with downloading this anew for every CI run, ensures we're always using the latest version.

We could perhaps cache it with a very short expiration time or something, but that wouldn't really achieve anything.

spacedmonkey:
@swissspidy Still getting the editor locally. 
Screenshot 2022-06-29 at 17 17 23

Screenshot 2022-06-29 at 17 28 01

PR #11787, opened 2022-06-23
Dashboard: Adjust Filter Dropdown Width

### LINKED ISSUE #11665: Dashboard: Adjust Filter Dropdown Width
## Feature Description

As a part of our Filters improvement and enhancement work, we should have the filters that are on the grid view of the Dashboard also be visible in the List view.

## Acceptance Criteria

- As a part of ensuring the list view is functional, let's be sure that the width of the dropdown elements in both the List and Grid view are appropriately sized.
- Dropdowns should also scale within the space in the top of the dashboard appropriately. See: https://github.com/GoogleForCreators/web-stories-wp/pull/11625#issuecomment-1153728314

## Out of Scope

- ~All filters that exist within the Dashboard Grid View should also exist as filters for the list view of the Dashboard~ Moved to https://github.com/GoogleForCreators/web-stories-wp/issues/11767
- ~This should be built in a way that adding filters in the future should work in both the List and Grid views of the Dashboard.~ Moved to https://github.com/GoogleForCreators/web-stories-wp/issues/11767

## Additional Context

This feedback was requested as a part of https://github.com/GoogleForCreators/web-stories-wp/pull/11625#discussion_r892222772

It may be easier to accomplish this after the core work is done with #11620

mwritter:
## Summary
Allow for multiple filters to be displayed on the Dashboard without overflowing.  Also allow for filter labels to expand the size of the filter, and drop down section based on the filter name length, option name length and the search placeholder length.

## Relevant Technical Choices
Used arbitrary max width for filters, options, and placeholder text to keep the filters reasonable size.

## User-facing changes

Before:
Before Filters

After:
After Filters

## Testing Instructions

1. Go to Dashboard
2. Make sure you have Multiple taxonomies 
- I used 'CPT UI' plugin to make a few filters with long text and then used the Story Editor to add taxonomy options
- Make sure to set 'Hierarchical' to 'true' for the filter you create with 'CPT UI'
3. Click on the 'Switch to list view' icon

On Staging:
- Notice filters can overflow the visible space, make the user have to scroll horizontally to  see other filters
- Notice there filters are not on the list view

On Branch:
- Notice filters widths vary, filters do not overflow visible space they stack in a grid formation.
- Notice the filters are on the list view

No

No

No

## Checklist

---

Fixes #11665


packages/dashboard/src/app/views/shared/bodyViewOptions.js
```diff
@@ -119,6 +128,10 @@ export default function BodyViewOptions({
                     noMatchesFoundLabel={filter.noMatchesFoundLabel}
                     searchPlaceholder={filter.searchPlaceholder}
                     offsetOverride
+                    maxWidth={FILTER_MAX_WIDTH}
```
mflorida, line 118: Maybe it's too late to bring this up, but do this list items require a max-width to not look funky? It's acceptable behavior for custom-styled drop-downs to expand to fit the width of the options. I've seen both fixed and expanding styles in the wild.

packages/dashboard/src/app/views/shared/bodyViewOptions.js
```diff
@@ -36,10 +36,14 @@ import { StandardViewContentGutter, ViewStyleBar } from '../../../components';
 import { DROPDOWN_TYPES, VIEW_STYLE } from '../../../constants';
 import useFilters from '../myStories/filters/useFilters';
 
+const FILTER_MAX_WIDTH = 350;
```
mflorida, line 39: Can this match the width of the sorting menu? In the example 'after' image, that's the widest one besides the super-wide menu.

mflorida:
The 'Viewing all XX stories' text seems super tight and awkward. Could that be shifted _above_ the filter menus so it can be displayed in a single line? It would more sense contextually as well.

packages/dashboard/src/app/views/shared/bodyViewOptions.js
mwritter: The `maxWidth` prop here is technically only used to style to `Popup` components width.  There is some logic added to the `Popup` component that now depends on this prop to keep these popups from being absurdly to big (ie if the user wants an 'option' thats a 200 word paragraph).  This way any where that uses the 'datalist' can now specify a `maxWidth` for the popup.  But if you have a suggestion on how make it look cleaner I can defiantly look into it.

mwritter:
> > > The 'Viewing all XX stories' text seems super tight and awkward. Could that be shifted _above_ the filter menus so it can be displayed in a single line? It would more sense contextually as well.
> > 
> > 
> > This is how the it currently works on `main`. I think we'd need a designers input before changing that up.
> 
> In the ‘original’ design ...

https://user-images.githubusercontent.com/14057928/175558126-4e06bd00-ff10-47c9-9e9f-8ede86770f0f.mp4

sblinde (approved):
🐝

Saw your demo of this during the design meeting too, well done here!

PR #11799, opened 2022-06-24
Editor: Replace deep clone dependency with native structured clone 

### LINKED ISSUE #11505: Editor: Replace deep clone dependency with native structured clone
## Feature Description

[`structuredClone`](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) is now available in all browsers. We currently use [`clone-deep`](https://github.com/GoogleForCreators/web-stories-wp/blob/8786821f787193c0b4127654f42269fbec8e9afd/packages/story-editor/package.json#L67) for this purpose, but [only in a single place](https://github.com/GoogleForCreators/web-stories-wp/blob/8786821f787193c0b4127654f42269fbec8e9afd/packages/story-editor/src/app/story/effects/useHistoryEntry.js#L56) as far as I can tell.

We should remove this dependency and use the browser native alternative.

timarney:
## Context

[`structuredClone`](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) is now available in all browsers. We currently use `clone-deep` for this purpose.  

> We should remove this dependency and use the browser native alternative.

## Summary

Replaces clone-deep with structuredClone

## Testing Instructions

1.

No

No

No

## Checklist

---

Fixes #11505


barklund (approved):
Looks good!

Have you checked for any cloning via JSON (i.e. `JSON.parse(JSON.stringify(object))`) in the source code? I have a feeling we might be doing that, but not sure.

spacedmonkey:
Just checking this is not related 

https://github.com/GoogleForCreators/web-stories-wp/blob/e00e50da7a02d956b993f4038ca1c15d46eda3ef/packages/story-editor/src/utils/getUniquePresets.js#L16-L19

merapi:
Regarding the https://github.com/GoogleForCreators/web-stories-wp/issues/11505#issuecomment-1129730015

@swissspidy you commented on May 18 and the second Safari version with `structuredClone` was released on May 17 so technically we were already good back then, but the first version that supported it was released not that long ago on March 14 so the question is - should we include the polyfill for those users? Chrome/Edge/Opera supports it since early Feb so any device with an older browser will not run the editor. Maybe we can check our analytics data?

timarney:
@merapi 

> Regarding the [#11505 (comment)](https://github.com/GoogleForCreators/web-stories-wp/issues/11505#issuecomment-1129730015)
> 
> @swissspidy you commented on May 18 and the second Safari version with `structuredClone` was released on May 17 so technically we were already good back then, but the first version that supported it was released not that long ago on March 14 so the question is - should we include the polyfill for those users? Chrome/Edge/Opera supports it since early Feb so any device with an older browser will not run the editor. Maybe we can check our analytics data?

I think the polyfill is automatic from the browser config for whatever browsers need it

https://github.com/GoogleForCreators/web-stories-wp/issues/11505#issuecomment-1129730015

swissspidy:
I think what Marcin is saying is that with our current browserslist config the polyfill would not be included anymore, since the addition to browsers is rather new but there have been >2 new releases since then, yet not all users might be on these latest two versions.

So in accordance with [our browser support policy](https://github.com/GoogleForCreators/web-stories-wp/blob/main/docs/browser-support.md), the question is whether we'd want/need to force our polyfill to be included so that our editor still works for users on Safari 15.3 or older. So the ask is to check analytics to see how our Safari usage looks like to inform such a decision.

I'll happily check our analytics.

From what I can see at first glance, the polyfill should still be included because of the "last 2 Opera versions" directive, and structuredClone was only added in the latest Opera version v84.

I haven't checked the built code though to confirm.

---

Aside: it looks like `structuredClone` is also only available in Node 17+, yet we're still on Node 16 at the moment. So if you're trying to write Jest tests for `useHistoryEntry` (which we actually might want to consider, given the whole `skipAddingEntry` logic), they would fail because of that.

merapi:
> Of course needs verification that the polyfill is indeed included as expected.

I'm testing that and something is wrong here, even `last 10 Safari versions` don't make it work on older Safari, maybe this `structuredClone` is too new or I'm making some mistake.

swissspidy:
Does changing this line:

https://github.com/GoogleForCreators/web-stories-wp/blob/5efe289d2b446aea1d645d140a46743eed1bbe93/babel.config.cjs#L35

to `corejs: '3.23'` change things for you?

I just read in the docs that apparently `corejs: 3` is interpreted as 3.0, and in 3.0 that polyfill was not included yet.

When I do this and run `npm run dev` I do see references to `core-js/modules/web.structured-clone.js` in the generated JS files. I just don't have an older Safari around to verify it.

merapi:
> Does changing this line:
> 
> https://github.com/GoogleForCreators/web-stories-wp/blob/5efe289d2b446aea1d645d140a46743eed1bbe93/babel.config.cjs#L35
> 
> to `corejs: '3.23'` change things for you?
> 
> I just read [in the docs](https://babeljs.io/docs/en/babel-preset-env#corejs) that apparently `corejs: 3` is interpreted as 3.0, and in 3.0 that polyfill was not included yet.
> 
> When I do this and run `npm run dev` I do see references to `core-js/modules/web.structured-clone.js` in the generated JS files. I just don't have an older Safari around to verify it.
> 
> With `corejs: 3` I do **not** see these references. So it does make a difference.

It works! 🎉 
Great find Pascal!
TIL, I was playing with `core-js` options and didn't see this:

image

```diff
- Who puts "Warning" in a grayed-out color instead of big bold red? 😄 
```

merapi (approved):
Also tested on QA env on older Safari - ✅

PR #11802, opened 2022-06-24
When removing a layer from a group, it can appear still inside the group

### LINKED ISSUE #11785: When removing a layer from a group, it can appear still inside the group
## Bug Description

If you remove a middle layer from a 3+ element group using the RCM option "Remove layer from group", it is still z-indexed "inside" the group.

## Expected Behaviour

When removing a layer from a group, it should be position either above or below all the other group elements.

## Steps to Reproduce

1. Add 3 elements to a group
2. In the layer panel, select the middle element, right-click and select "Remove layer from group"
3. Observe that the list of layers is now in an invalid state.

## Screenshots

layer-remove-from-group-bug

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

timarney:
## Context

> If you remove a middle layer from a 3+ element group using the RCM option "Remove layer from group", it is still z-indexed "inside" the group.

See: https://github.com/GoogleForCreators/web-stories-wp/issues/11785

## Summary

Updates the layer ungroup code to move selected layer above it's previous group

Demo: https://www.youtube.com/watch?v=Rx3u0mteWDw

## Testing Instructions

1. Add 3 elements to a group
2. In the layer panel, select the middle element, right-click and select "Remove layer from group"
3. Observe that the layer selected moves to the spot above the group the item was in.

No

No

No

## Checklist

---

Fixes #11785


packages/story-editor/src/components/panels/design/layer/layerList.js
```diff
@@ -69,7 +69,7 @@ const ReorderableLayer = memo(function ReorderableLayer({
     <Fragment key={id}>
       <LayerSeparator
         groupId={element.groupId}
-        nestedOffset={nestedOffset}
+        nestedOffset={Boolean(nestedOffset)}
```
timarney, line 72: Unrelated to this PR but was making noise in the console - prop is expected to be a Boolean

barklund (commented):
I feel like all of this logic belongs inside the reducer. We could create a new action, `removeElementFromGroup` or just `ungroupElement`. In that way we can make it much cleaner because it can access the entire state seamlessly - and we can test it exhaustively.

timarney:
> I feel like all of this logic belongs inside the reducer. We could create a new action, `removeElementFromGroup` or just `ungroupElement`. In that way we can make it much cleaner because it can access the entire state seamlessly - and we can test it exhaustively.

Sounds good - I moved this back to draft while I make the updates.

merapi:
It looks like a regression from #11763 was missed (when grouping elements):
image

In the old version we created the `groups` key:
image

miina:
One finding from testing: when removing a layer from the group, it's always positioned above the group at first, however, when changing selection, it can jump below the group, for example:
position

What I did:
- created a group of 4 elements
- removed the last item from the group, it was added above the group
- Clicked on the remaining of the group in the Layer panel to select the group
- Now the last item jumped to below the group

packages/story-editor/src/app/story/useStoryReducer/test/addGroup.js
```diff
@@ -79,6 +79,26 @@ describe('addGroup', () => {
     });
   });
 
+  it('should add group even if there are no groups', () => {
```
miina: Nit:
Is this for testing it works when `groups` is not set yet? Since in the previous test there are also no groups but the `groups` prop is set, just empty. Perhaps this then:
```suggestion
  it('should add group even if `groups` is not set, () => {
```

miina (approved):
Works well, LGTM!

Found one issue when testing, left it as a separate comment.

packages/story-editor/src/app/story/useStoryReducer/test/addGroup.js
barklund: True, the text could be more explicit. Or the tests more . Then empty `groups` object is a bit rare, so either it should be there and be non-empty or not be there at all.

barklund:
> > One finding from testing: when removing a layer from the group, it's always positioned above the group at first, however, when changing selection, it can jump below the group
> 
> Can't repro?:

I think what happens is a different bug. As you click the group to highlight the first layer in the group, you accidentally move the group to the position it is already in, which somehow moves it to the top. Which definitely shouldn't happen, but that's a bug in moving groups, not in the ungrouping action. Please correct me if I'm wrong...

group-move-bug

miina:
> > > One finding from testing: when removing a layer from the group, it's always positioned above the group at first, however, when changing selection, it can jump below the group
> > 
> > 
> > Can't repro?:
> 
> I think what happens is a different bug. As you click the group to highlight the first layer in the group, you accidentally move the group to the position it is already in, which somehow moves it to the top. Which definitely shouldn't happen, but that's a bug in moving groups, not in the ungrouping action. Please correct me if I'm wrong...

Hm, I'm not certain, I didn't see the blue line when it happened locally, however, I can also repro it with accidental dragging as well. Since neither of you can repro it then we can just ignore for now.

PR #11828, opened 2022-06-28
Workspace: Update disclosure menu components

### LINKED ISSUE #11770: Workspace: Fix styling of menu-like components that use a 'chevron' icon
## Bug Description

As part of the work done for #10891, a new `Disclosure` component was created to normalize styling and functionality of the footer buttons. These changes affected styling other components in the app that used the chevron icon, and styling should be updated to restore proper appearance. (Use of the new `Disclosure` component will resolve this, as well as reflect the open/closed state of the menu and its contents.

## Expected Behaviour

All menu-like components that use the chevron icon should have the same appearance and behavior.

## Steps to Reproduce

Build from `fix/10891-footer-toggle-button-style`. Spacing is off in the font selector in the pop-up bar for editing text and the media type selector in the 'Insert' pane. Open/closed state for the publish time menu, as well as the text alignment and text color menus in the text editing pop-up bar.

This is dependent on changes in #10891, which has yet to be merged to `main`.

## Screenshots

Spacing is off:
Screen Shot 2022-06-21 at 1 47 41 PM 
Screen Shot 2022-06-21 at 1 47 11 PM

Does not reflect open/closed state:
Screen Shot 2022-06-21 at 1 50 20 PM 
Screen Shot 2022-06-21 at 1 49 36 PM 
Screen Shot 2022-06-21 at 1 48 52 PM

## Additional Context

Depends on #10891 

- Plugin Version: 1.22.0-alpha.0
- WordPress Version: 6.0
- Operating System: macOS
- Browser: Chrome 102.0.5005.115

mflorida:
## Context

All menu-like components that use the chevron icon should have the same appearance and behavior, and use the same component where possible.

## Summary

Update instances of the 'chevron' icon in clickable elements used to show/hide content or menus with the `Disclosure` component and resolve related styling issues.

## Relevant Technical Choices

Refactored disclosure-type UI elements to use the `Disclosure` component for consistent appearance and implementation.

## To-do

- Address any unresolved styling issues
- Replace other disclosure-type UI elements as needed

## User-facing changes

Before - icons do not properly reflect open/closed state

Screen Shot 2022-06-27 at 8 32 05 PM

Screen Shot 2022-06-27 at 8 33 04 PM

Screen Shot 2022-06-27 at 8 34 14 PM

After - icons properly reflect open/closed state

Screen Shot 2022-06-27 at 8 34 48 PM

Screen Shot 2022-06-27 at 8 35 16 PM

Screen Shot 2022-06-27 at 8 36 05 PM

## Testing Instructions

Ensure 'disclosure' chevron icons rotate to reflect the open/closed state properly.

No

No

No

## Checklist

---

Fixes #11770 

NOTE: #11737 must be merged first, with this immediately following


packages/design-system/src/components/disclosure/index.js
```diff
@@ -25,6 +25,7 @@ import PropTypes from 'prop-types';
 import { THEME_CONSTANTS } from '../../theme';
 import { ChevronDownSmall } from '../../icons';
 
+// TODO: Account for RTL layout for 'left' and 'right'?
```
sblinde: Note: I'm still reviewing the code here and haven't looked at it to do a quick QA, but do you see any issues with how this works/looks before and after enabling RTL?

packages/design-system/src/components/disclosure/index.js
```diff
@@ -39,25 +40,29 @@ const rotate = {
  *
  * @param {Object} [props] All props.
  * @param {string} [props.direction] Initial direction of chevron icon
+ * @param {boolean} [props.disabled] Should the icon have a 'disabled' appearance?
```
sblinde, line 42: I'm not sure whether we need to expand for this functionality for now, unless you're actively seeing the `disabled` or `aria-disabled` prop in any of the components you've updated here. So I think this could likely be removed for now.

sblinde (commented):
Looks pretty good, I only found one issue when doing a quick QA pass:

For the RTL layout, if you add an item of text in the editor, and go to use the floating menu, the alignment of the disclosure icon is odd with the color picker. All the others look good so far, as far as I can see.

Screen Shot 2022-06-28 at 11 38 35 AM

packages/design-system/src/components/disclosure/index.js
mflorida: I did see it elsewhere, that's what prompted the addition here. It was added just for appearance, but if kept could also prevent rotation of the icon. I'd assume a 'disabled' menu or button would _likely_ not set `isOpen` to `true`?

packages/design-system/src/components/disclosure/index.js
mflorida: There are no `left` or `right` instances of this specific component, but there _are_ left/right facing disclosure icons used in the app that _could_ use it.

I'm not sure if RTL layout would alter the left/right relation of the icon to the content being shown, thus the `?` in the comment.

mflorida:
> Looks pretty good, I only found one issue when doing a quick QA pass:
> 
> For the RTL layout, if you add an item of text in the editor, and go to use the floating menu, the alignment of the disclosure icon is odd with the color picker. All the others look good so far, as far as I can see.
> 
> Screen Shot 2022-06-28 at 11 38 35 AM

The color picker needs a major refactor - it's pieced together and I built on top of that. I tried to untangle how it's implemented, but there are a bunch of nested components to dig through…

How does it function on the `new` staging server? Or on a different branch of the `qa` server?

packages/design-system/src/components/disclosure/index.js
mflorida: Here are the checklist items in LTR vs RTL layout. IMO, the RTL icons should point to the left. 🤔

LTR:
Screen Shot 2022-06-28 at 11 09 08 AM

RTL:
Screen Shot 2022-06-28 at 11 08 48 AM

But these do not currently use the `Disclosure` component.

packages/design-system/src/components/disclosure/index.js
mflorida: A quick search for `theme.colors.fg.disable` in the source shows a lot of components have a 'disabled' state. Saying that, if the `disabled` prop stays for the `Disclosure` component, the instances where it's used should be updated if its parent supports a 'disabled' state.

packages/design-system/src/components/disclosure/index.js
mflorida: Thinking further, this component should not handle RTL layout.

sblinde:
> Looking at the floating menu before updating the `Disclosure` component (below), the spacing is way off. So the changes in this PR fixes these and makes the color picker better (but still needs a refactor).
> 
> image

Yeah, it could totally be pre-existing which you've evidenced, that I didn't check before my comment, but just something to note.

sblinde (approved):

packages/design-system/src/components/disclosure/index.js
sblinde: Gotcha! Then I think it makes sense to have it, awesome. Thanks.

packages/design-system/src/components/disclosure/index.js
sblinde: I can agree with your assessment here. I think the component that utilizes `Disclosure` would be the space where RTL/LTR would be handled.

PR #11837, opened 2022-06-29
Shortpixel Image Optimizer Incompatibility

### LINKED ISSUE #10797: Shortpixel Image Optimizer Incompatibility
## Bug Description

Plugins like Shortpixel are used for image optimizations within WordPress.

We upload an image for every custom saved page template. These are only used in the editor and should not be optimized by Shortpixel.

Otherwise, in combination with the bug where multiple images are being generated (#10787), this is pretty wasteful.

## Expected Behaviour

Shortpixel should not optimize page template images

## Steps to Reproduce

1. Install [Shortpixel Image Optimizer plugin ](https://wordpress.org/plugins/shortpixel-image-optimiser/)
2. Go to Web Stories > Explore Templates > Use (any) template
3. Go to Page template panel
4. Save page template
5. In Shortpixel stats, see that it cost a credit to optimize the page template image

## Additional Context
Here is the original report: https://wordpress.org/support/topic/duplicate-custom-template-images-in-uploads-folder

When this plugin is disabled, there are no erroneous images cloned. 
We will gather more details about the specific settings in Shortpixel and share them here when available. 

- Plugin Version: 1.18.0
- WordPress Version:
- Operating System:
- Browser:

timarney:
## Context

Plugins like Shortpixel are used for image optimizations within WordPress.

We upload an image for every custom saved page template. These are only used in the editor and should not be optimized by Shortpixel.

## Summary

Adds a filter per

https://github.com/GoogleForCreators/web-stories-wp/issues/10797#issuecomment-1164154035

To check for a page template image and prevent the optimization

## Testing Instructions

Via #10797

1. Install [Shortpixel Image Optimizer plugin](https://wordpress.org/plugins/shortpixel-
2. image-optimiser/)
3. Go to Web Stories > Explore Templates > Use (any) template
4. Go to Page template panel
5. Save page template
6 In Shortpixel stats, see that it cost a credit to optimize the page template image

> Shortpixel should not optimize page template images

Note: Another way to test is to add a media image with `web-stories-page-template` in the name. 

Press the `Optimize now button` -- watch the Network tab (ajax request) + response
Screen Shot 2022-06-29 at 4 58 48 PM

No

No

No

## Checklist

---

Fixes #10797


timarney:
Example payload
Screen Shot 2022-06-28 at 7 39 20 PM

.phpstorm.meta.php
```diff
@@ -30,6 +30,7 @@
 			'integrations.cfi'             => \Google\Web_Stories\Integrations\Conditional_Featured_Image::class,
 			'integrations.sitekit'         => \Google\Web_Stories\Integrations\Site_Kit::class,
 			'integrations.themes_support'  => \Google\Web_Stories\Integrations\Core_Themes_Support::class,
+			'integrations.shortpixel'         => \Google\Web_Stories\Integrations\ShortPixel::class,
```
swissspidy: Nit: indentation

includes/Integrations/ShortPixel.php
```diff
@@ -0,0 +1,60 @@
...
+	 * Ensures page template urls bypass optimisation.
+	 *
+	 * @since 1.23.0
+	 *
+	 * @param string[] $urls  Urls that will be sent to optimisation.
+	 * @param int      $post_id Post ID.
+	 * @return string[] The filtered Urls.
+	 */
+	public function image_urls( $urls, $post_id ): array {
+		if ( str_contains( $urls[0], 'web-stories-page-template' ) ) {
```
swissspidy: On the ticket I originally mentioned doing a term query to get the media source, but given that this is the only use case for now `str_contains` is definitely faster 👍 

What if the page template image is not the first item in the list though? Does that happen? Seems like `array_filter` is more fitting to catch such a case (if it exists).

includes/Integrations/ShortPixel.php
```diff
@@ -0,0 +1,60 @@
...
+	/**
+	 * Ensures page template urls bypass optimisation.
+	 *
+	 * @since 1.23.0
+	 *
+	 * @param string[] $urls  Urls that will be sent to optimisation.
+	 * @param int      $post_id Post ID.
+	 * @return string[] The filtered Urls.
+	 */
+	public function image_urls( $urls, $post_id ): array {
```
swissspidy: If we don't use the `$post_id` param we can just remove it.

includes/Integrations/ShortPixel.php
spacedmonkey: Str_contains is either PHP 8.0+ or WP 5.9 - https://github.com/WordPress/wordpress-develop/commit/3cc8f1237a737ccf895682029afacafd7305d420

We support 5.7 and 5.8

tests/phpunit/integration/tests/Integrations/ShortPixel.php
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	/**
+	 * @covers ::image_urls
+	 */
+	public function test_image_urls(): void {
+		$short_pixel = new \Google\Web_Stories\Integrations\ShortPixel();
+		$urls        = [ 'http://localhost:8899/wp-content/uploads/2022/03/web-stories-page-template-768.jpg' ];
+		$result      = $short_pixel->image_urls( $urls, 10 );
+		$this->assertSame( $result, [] );
```
spacedmonkey:
```suggestion
		$this->assertEqualSets( $result, [] );
```

tests/phpunit/integration/tests/Integrations/ShortPixel.php
```diff
@@ -0,0 +1,55 @@
...
+
+	/**
+	 * @covers ::image_urls
+	 */
+	public function test_image_urls_non_page_template_image(): void {
+		$short_pixel = new \Google\Web_Stories\Integrations\ShortPixel();
+		/* given a non "page-template" image the urls should pass thru */
+		$urls   = [ 'http://localhost:8899/wp-content/uploads/2022/03/example-750.jpg' ];
+		$result = $short_pixel->image_urls( $urls, 10 );
+		$this->assertSame( $result, $urls );
```
spacedmonkey:
```suggestion
		$this->assertEqualSets( $result, $urls );
```

tests/phpunit/integration/tests/Integrations/ShortPixel.php
swissspidy: Note that order of params should be the other way around. `$this->assertSame( $expected, $actual )` / `$this->assertEqualSets( [], $result );`

For comparing an empty array `assertSame` seems suited well though

tests/phpunit/integration/tests/Integrations/ShortPixel.php
swissspidy: Note that either way the order of params should be the other way around

includes/Integrations/ShortPixel.php
swissspidy: Curious why tests aren't failing though 🤔

timarney: I'll adjust the code (just in case)

From what I see .... the urls contain are same at different sizes.

Also read that the post_id isn't always available so opted to check based on the URL string

tests/phpunit/integration/tests/Integrations/ShortPixel.php
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+namespace Google\Web_Stories\Tests\Integration\Integrations;
+
+use Google\Web_Stories\Tests\Integration\TestCase;
+
+/**
+ * @coversDefaultClass \Google\Web_Stories\Integrations\ShortPixel
+ */
+class ShortPixel extends TestCase {
```
spacedmonkey:
```suggestion
class ShortPixel extends DependencyInjectedTestCase {

	/**
	 * @var \Google\Web_Stories\Integrations\ShortPixel
	 */
	private $instance;

	public function set_up(): void {
		parent::set_up();

		$this->instance = $this->injector->make( \Google\Web_Stories\Integrations\ShortPixel::class );
	}
```

tests/phpunit/integration/tests/Integrations/ShortPixel.php
```diff
@@ -0,0 +1,55 @@
...
+	/**
+	 * @covers ::register
+	 */
+	public function test_register(): void {
+		$short_pixel = new \Google\Web_Stories\Integrations\ShortPixel();
+		$short_pixel->register();
+		$this->assertSame( 10, has_filter( 'shortpixel_image_urls', [ $short_pixel, 'image_urls' ] ) );
```
spacedmonkey:
```suggestion
		$this->instance->register();
		$this->assertSame( 10, has_filter( 'shortpixel_image_urls', [ $this->instance, 'image_urls' ] ) );
```

tests/phpunit/integration/tests/Integrations/ShortPixel.php
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	/**
+	 * @covers ::image_urls
+	 */
+	public function test_image_urls(): void {
+		$short_pixel = new \Google\Web_Stories\Integrations\ShortPixel();
+		$urls        = [ 'http://localhost:8899/wp-content/uploads/2022/03/web-stories-page-template-768.jpg' ];
+		$result      = $short_pixel->image_urls( $urls, 10 );
```
spacedmonkey:
```suggestion
		$urls        = [ 'http://localhost:8899/wp-content/uploads/2022/03/web-stories-page-template-768.jpg' ];
		$result      = $this->instance->image_urls( $urls, 10 );
```

tests/phpunit/integration/tests/Integrations/ShortPixel.php
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	/**
+	 * @covers ::image_urls
+	 */
+	public function test_image_urls_non_page_template_image(): void {
+		$short_pixel = new \Google\Web_Stories\Integrations\ShortPixel();
+		/* given a non "page-template" image the urls should pass thru */
+		$urls   = [ 'http://localhost:8899/wp-content/uploads/2022/03/example-750.jpg' ];
+		$result = $short_pixel->image_urls( $urls, 10 );
```
spacedmonkey:
```suggestion
		/* given a non "page-template" image the urls should pass thru */
		$urls   = [ 'http://localhost:8899/wp-content/uploads/2022/03/example-750.jpg' ];
		$result = $this->instance->image_urls( $urls, 10 );
```

tests/phpunit/integration/tests/Integrations/ShortPixel.php
```diff
@@ -0,0 +1,55 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Web_Stories\Tests\Integration\Integrations;
+
+use Google\Web_Stories\Tests\Integration\TestCase;
```
spacedmonkey:
```suggestion
use Google\Web_Stories\Tests\Integration\DependencyInjectedTestCase;
```

includes/Integrations/ShortPixel.php
```diff
@@ -0,0 +1,59 @@
...
+	/**
+	 * Ensures page template urls bypass optimisation.
+	 *
+	 * @since 1.23.0
+	 *
+	 * @param string[] $urls  Urls that will be sent to optimisation.
+	 * @return string[] The filtered Urls.
+	 */
+	public function image_urls( $urls ): array {
+		if ( false !== strpos( $urls[0], 'web-stories-page-template' ) ) {
```
spacedmonkey:
```suggestion
        if( empty( $urls ) ) {
             return $urls;
        }
		if ( false !== strpos( $urls[0], 'web-stories-page-template' ) ) {
```

includes/Integrations/ShortPixel.php
```diff
@@ -0,0 +1,71 @@
...
+	 *
+	 * @param string[] $urls  Urls that will be sent to optimisation.
+	 * @return string[] The filtered Urls.
+	 */
+	public function image_urls( $urls ): array {
+		if ( empty( $urls ) ) {
+			 return $urls;
+		}
+
+		$match = array_filter(
```
spacedmonkey: Could we just use an in_array?

spacedmonkey (changes requested):

includes/Integrations/ShortPixel.php
timarney: I think we need to match on a partial string so I don't think  in_array will work

spacedmonkey (approved):

includes/Integrations/ShortPixel.php
```diff
@@ -0,0 +1,71 @@
...
+			static function( $url ) {
+				return false !== strpos( $url, 'web-stories-page-template' );
+			}
+		);
+
+		if ( $match ) {
+			return [];
+		}
+
+		return $urls;
```
swissspidy: `array_filter` returns the filtered array, so can't we just do `return array_filter( ....)`?

Right now this is filtering the array and then if the array is not empty, return an empty array, otherwise return it unchanged.

includes/Integrations/ShortPixel.php
```diff
@@ -0,0 +1,71 @@
...
+	 *
+	 * @since 1.23.0
+	 *
+	 * @param string[] $urls  Urls that will be sent to optimisation.
+	 * @return string[] The filtered Urls.
+	 */
+	public function image_urls( $urls ): array {
+		if ( empty( $urls ) ) {
+			return $urls;
+		}
```
swissspidy: This is not really needed, as `array_filter` on an empty array won't do anything. We can just remove it. Reduces complexity and another case to test against (right now there's no test for this scenario anyway)

includes/Integrations/ShortPixel.php
```diff
@@ -0,0 +1,61 @@
...
+ */
+class ShortPixel extends Service_Base {
+
+	/**
+	 * Runs on instantiation.
+	 *
+	 * @since 1.23.0
+	 */
+	public function register(): void {
+		add_filter( 'shortpixel_image_urls', [ $this, 'image_urls' ], 10, 1 );
```
swissspidy: Nit:

```suggestion
		add_filter( 'shortpixel_image_urls', [ $this, 'image_urls' ] );
```

includes/Integrations/ShortPixel.php
timarney: Updated --- 

The filter should now keep and return non page-template urls.

Added / updated the tests as well.

swissspidy (approved):

PR #11849, opened 2022-06-30
Canvas: Show base color / blurhash while resource is loading

### LINKED ISSUE #10258: Canvas: Show base color /  blurhash while resource is loading
## Feature Description

Context https://github.com/google/web-stories-wp/pull/10253#issuecomment-1016411650. 

Show base color and blurhash in canvas while a video/image/gif resource is loading as a placeholder. 

See https://github.com/google/web-stories-wp/pull/9941 https://github.com/google/web-stories-wp/pull/9953 https://github.com/google/web-stories-wp/issues/9647

spacedmonkey:
## Summary

Show base color / blurhash based placeholder for element on canvas. 

## User-facing changes

https://user-images.githubusercontent.com/237508/176723782-2d057043-a230-4249-aeac-50ed8662646c.mov

## Testing Instructions

1. Create story. 
2. Insert a number of video / image elements. 
3. Save story as draft. 
4. Go to media library. 
5. Delete video / image used in elements. 
6. Edit draft story.
7. See placeholder. 
8. Resized elements, see placeholder change size as well. 

## Checklist

---

Fixes #10258


swissspidy:
Seems to work well already!

spacedmonkey:
https://user-images.githubusercontent.com/237508/176920957-d112ddbc-8a84-4b65-98ff-2a376c911906.mov

As noted above, here is the issue in safari. It is none issue, but it make testing this PR a little harder.

packages/story-editor/src/components/canvas/displayElement.js
```diff
@@ -240,6 +241,7 @@ function DisplayElement({
             isCurrentResourceUploading={isCurrentResourceUploading}
             maybeEnqueueFontStyle={maybeEnqueueFontStyle}
             siblingCount={siblingCount}
+            renderResourcePlaceholder={renderResourcePlaceholder}
```
miina, line 244: Since this seems to be used only in the `element-library` package and is only for the media elements, perhaps the util and the usage could be moved to the `element-library` directly to the corresponding elements' display logic? Seems like something that could belong directly to the element/media display. 

Or is it here because `renderResourcePlaceholder` uses the `react-blurhash` library, and it shouldn't be imported from the `element-library` package?
Thoughts?

swissspidy: > Or is it here because `renderResourcePlaceholder` uses the `react-blurhash` library, and it shouldn't be imported from the `element-library` package?

Correct. See discussion at https://github.com/GoogleForCreators/web-stories-wp/issues/10258#issuecomment-1171408347

packages/element-library/src/media/display.js
```diff
@@ -123,6 +124,9 @@ function MediaDisplay({
       mask={mask}
       showPlaceholder={showPlaceholder}
     >
+      {showPlaceholder &&
```
miina, line 127: Q: Are the `showPlaceholder` and `resourcePlaceholder` both needed for the media display?
It looks like in case of `showPlaceholder`, the following CSS is applied to media: `background-color: #C4C4C4;`. Wondering it it's still needed now?

Perhaps we could just use `renderResourcePlaceholder?.(resource)` here then.

spacedmonkey: If the resource does not have base color or blurhash, nothing is rendered and it fallbacks to #C4C4C4 background, so still needed.

miina (approved):
LGTM!

(Looks like e2e tests didn't run, not sure if that's expected or should be restarted)

merapi (approved):
Looks good, just one thing to fix:

`width` and `height` passed to `BlurhashContainer` is taken from the original image (big one) so the canvas blurhash looks different from the thumbnail blurhash.

I think we can remove the `width` and `height` props from the `BlurhashContainer` as they are optional and the parent will take care of the rest, or use `width=100%, height=auto`.

Example (1st load - current, 2nd - fixed):

https://user-images.githubusercontent.com/21036224/177538881-27c831e6-778b-446f-8511-030e262e3d08.mp4

spacedmonkey:
> width and height passed to BlurhashContainer is taken from the original image (big one) so the canvas blurhash looks different from the thumbnail blurhash.

@merapi The height and width are required here. Depending on these, the blurhash looks different. But you are correct that the height and width of the resource are the wrong. I have tweaked the logic to use the size of the element, which improves performance and looks a lot better.

merapi:
> @merapi The height and width are required here. Depending on these, the blurhash looks different. But you are correct that the height and width of the resource are the wrong. I have tweaked the logic to use the size of the element, which improves performance and looks a lot better.

Looks like the docs are not updated? it defaults to `auto`.
https://github.com/woltapp/react-blurhash/blob/master/src/Blurhash.tsx#L8

It is a bit misleading but width/height is only used in div style and what actually impacts performance is the resolution props:
```jsx
<div
  {...rest}
  style={{ display: 'inline-block', height, width, ...style, position: 'relative' }}
>
  <BlurhashCanvas
    hash={hash}
    height={resolutionY}
    width={resolutionX}
    punch={punch}
    style={canvasStyle}
  />
</div>
```
Either solution will make it look correct 👍🏻

merapi:
> The height and width are used to generate the canvas. See https://github.com/woltapp/react-blurhash/blob/master/src/BlurhashCanvas.tsx

Nope, we are not using `BlurhashCanvas`, we use `Blurhash`.

spacedmonkey:
> Nope, we are not using `BlurhashCanvas`, we use `Blurhash`.

But blurhash does blurhashcanvas - https://github.com/woltapp/react-blurhash/blob/458b44880845085066e5640bdc5a24a51f077002/src/Blurhash.tsx#L53-L59

merapi:
> > Nope, we are not using `BlurhashCanvas`, we use `Blurhash`.
> 
> But blurhash does blurhashcanvas - https://github.com/woltapp/react-blurhash/blob/458b44880845085066e5640bdc5a24a51f077002/src/Blurhash.tsx#L53-L59

Correct, but look how it passes the props.

image

spacedmonkey:
>Correct, but look how it passes the props. 

Right, I have passed x and y props.

merapi:
> > Correct, but look how it passes the props.
> 
> Right, I have passed x and y props.

That is unnecessary (actually degrading the performance).

> The X-axis resolution in which the decoded image will be rendered at. Recommended min. 32px. Large sizes (>128px) will greatly decrease rendering performance. (Default: 32)

Right now my test image is rendered with resolution=336, it takes ~62ms to render, while with recommended/default resolution=32 it takes ~1ms.

I have a little tool for performance testing and there is no point to go above >32 in resolution because there is no visual difference, it just takes longer to render (we encode the image with [components=4](https://github.com/GoogleForCreators/web-stories-wp/blob/d2b8c3058456f086adfe5a40b8b01cd45b2f7197/packages/story-editor/src/utils/getBlurHashFromImage.js#L56) so there is not much data to work with).

image

So the best settings are: resolution=32 (default), width/height can be undefined/auto or actual box dimensions - no difference.

spacedmonkey:
> So the best settings are: resolution=32 (default), width/height can be undefined/auto or actual box dimensions - no difference.

Height and width have been removed.

timarney (approved):

PR #11853, opened 2022-07-01
Canvas: Video poster not rendering in Safari

### LINKED ISSUE #10167: Canvas: Video poster not rendering in Safari

spacedmonkey:
## Summary
While working on #11849, I found a quick fix for this issue. Setting background image for poster seems to work well here.  

## Testing Instructions

This issues an issue in safari. Test in all browser but with a special focus safari. 
1. Create story. 
2. Insert video onto canvas. 
3. Add a custom video poster ( both upload file and link to file ). 
4. See poster render. 
5. Create another page, change to new page. 
6. Go back to original page. See poster. 

## Checklist

---

Fixes #10167


packages/element-library/src/video/trim.js
```diff
@@ -30,14 +30,10 @@ import { StoryPropTypes, getTransformFlip } from '@googleforcreators/elements';
...
-import { getBackgroundStyle, videoWithScale } from './util';
+import { getBackgroundStyle, Video as _Video } from './util';
 import Captions from './captions';
 
-const Video = styled.video`
-  position: absolute;
-  max-width: initial;
-  max-height: initial;
-  ${videoWithScale}
+const Video = styled(_Video)`
```
swissspidy: We don't typically use underscore suffixes or prefixes.

We usually do `const StyledVideo = styled(Video)`

Note: when using `StyledVideo`, also need to update the `jsx-a11y/media-has-caption` config here to add it to the list:
https://github.com/GoogleForCreators/web-stories-wp/blob/5b29ef015ddc679812c0c7dcb00422a07738b520/.eslintrc#L91-L95

packages/element-library/src/gif/display.js
```diff
@@ -27,24 +26,7 @@ import { StoryPropTypes } from '@googleforcreators/elements';
...
-const Image = styled.img`
-  position: absolute;
-  max-height: initial;
-  object-fit: contain;
-  width: ${({ width }) => `${width}px`};
-  left: ${({ offsetX }) => `${-offsetX}px`};
-  top: ${({ offsetY }) => `${-offsetY}px`};
-  max-width: ${({ isBackground }) => (isBackground ? 'initial' : null)};
-`;
+import { getBackgroundStyle, Video, VideoImage } from '../video/util';
```
swissspidy: Since these are shared between the gif and the video element, this is a good opportunity to move them to `media/util.js` instead.

packages/element-library/src/video/util.js
```diff
@@ -34,3 +33,19 @@ export const getBackgroundStyle = () => {
     maxWidth: 'initial',
   };
 };
+
+export const Video = styled.video`
+  position: absolute;
+  max-width: initial;
+  max-height: initial;
+  height: ${({ height }) => `${height}px`};
+  background-image: ${({ poster }) => `${poster}`};
```
swissspidy: IIUC this is the main change you've added?

Shouldn't this be like this though?

```suggestion
  background-image: ${({ poster }) => `url("${poster}")`};
```

packages/element-library/src/gif/display.js
```diff
@@ -27,24 +26,7 @@ import { StoryPropTypes } from '@googleforcreators/elements';
...
-const Image = styled.img`
-  position: absolute;
-  max-height: initial;
-  object-fit: contain;
-  width: ${({ width }) => `${width}px`};
-  left: ${({ offsetX }) => `${-offsetX}px`};
-  top: ${({ offsetY }) => `${-offsetY}px`};
-  max-width: ${({ isBackground }) => (isBackground ? 'initial' : null)};
-`;
+import { getBackgroundStyle, Video, VideoImage } from '../media/util';
```
miina, line 29: Nit: Hm, should `Video` and `VideoImage` be in a constants file instead? Doesn't sound like these are utilities.
Although I can see that it was done so already before as well so feel free to ignore.

miina (approved):
Works great.

In other browsers now the poster is shown + the background image is displayed as well, I suppose that's ok.

packages/element-library/src/gif/display.js
spacedmonkey: I don't feel strongly either way.

timarney (approved):

spacedmonkey:
@swissspidy I have tested this with hotlinked poster and it works. Can you review on your end?

swissspidy:
Not seeing any issues either. Though also having troubles to reproduce the original issue in the first place anyway.

Let's send this to QA

felipebochehin87:
Safari (video below). Also checked on Firefox and Safari. I couldn't reproduce the issue locally, nor in [QA](https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/), but I checked the functionality and the fix didn't cause any negative impact.

[Uploading... Screen Recording 2022-07-12 at 06.24.00.mov]

PR #11884, opened 2022-07-07
Fix layer re-arrangement inside groups using shortcuts

### LINKED ISSUE #11784: Layer rearrangement using keyboard shortcuts with groups is buggy
## Bug Description

If you rearrange a layer with keyboard shortcuts, grouped layers are no longer guaranteed to be consecutive.

Note that <kbd>cmd</kbd>+<kbd>shift</kbd>+<kbd>up</kbd> and  <kbd>cmd</kbd>+<kbd>shift</kbd>+<kbd>down</kbd> are also possible and have related but less severe issues.

## Expected Behaviour

* Grouped layers should always be consecutive.
* When a layer is moved up/down using keyboard shortcuts, it should either:
  * Be moved only within its level in the layer hierarchy
    * So would move completely above or below another group if _ungrouped_ or
    * Only move inside its group if _grouped_
    * If we go with this option, we could adopt <kbd>cmd</kbd>+<kbd>left</kbd> and  <kbd>cmd</kbd>+<kbd>right</kbd> as the way to move elements in and out of groups (beware of RTL)
  * Be moved freely, but change groups as needed

#### Move ungrouped layer "inside" a group
1. Add 3 elements
2. Group 2 of them
3. Move the last element up and down in the layer stack using <kbd>cmd</kbd>+<kbd>up</kbd> and  <kbd>cmd</kbd>+<kbd>down</kbd>
4. Observe that the ungrouped layer can be positioned inside a group

#### Move layer from one group to another without actually doing so.

1. Add 4 elements
2. Group them 2 and 2
3. In the layer panel, select the _top-most_ layer in the _bottom-most_ group
4. Move the layer up one step by pressing <kbd>cmd</kbd>+<kbd>up</kbd>
5. Observe that the moved layer is now moved out of its own group to the top level, but appears to be inside the other group

## Screenshots
layer-arrange-bug

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

miina:
## Context
Layer arrangement using shortcuts got broken after adding the groups feature. This PR aims to fix it.

## Summary
Allows moving a layer up/down considering groups. A layer is moved inside/outside of the groups depending on the position and grouping.

If the layer is the first item of the group, and UP is pressed, the layer is moved outside of that group, the position stays the same. If the layer is the last item of the group and DOWN is pressed, the layer is moved outside of the group, the position stays the same.

The same logic is also used for moving a layer into a group.

For other cases, the logic remains as currently is.

## User-facing changes
Groups are now added / removed from layers when using shortcuts.

## Testing Instructions

1. Add 3 elements and group the first 2 elements.
2. Now select the ungrouped element.
3. Use <kbd>cmd</kbd>+<kbd>up</kbd> and <kbd>cmd</kbd>+<kbd>down</kbd> to change the position of the layer.
4. Look at the Layer Panel at the same time -- verify that the selected layer is moving as expected into a group and out of the group when moving it up and down in the layers list.

groups

No

No

No

## Checklist

---

Fixes #11784


packages/story-editor/src/app/canvas/utils/test/getLayerArrangementProps.js
```diff
@@ -0,0 +1,122 @@
...
+
+  it('should get the layer group and position correctly when moving an element into a group', () => {
+    const elements = [
+      { id: 'a', isBackground: true },
+      { id: 'b' },
+      { id: 'c', groupId: 'g1' },
+      { id: 'd' },
+      { id: 'e', groupId: 'g2' },
+    ];
+    // Position stays the same but the layer is moved out of the group.
```
barklund:
```suggestion
    // Position stays the same but the layer is moved into either group.
```

merapi:
> Looks like the shortcuts don't support rearranging groups currently at all, couldn't see an issue for it, should we add? cc @merapi

Yeah, I think we can also add a few more things that are missing:
- DND rearranging multiple layers (select 2 layers and DND them in the Layers Panel - doesn't work, you can only DND groups or single layers).
- Same as above but via keyboard.
- Keyboard rearrangement (of any kind) is not possible in the Layers Panel, you need to select the layer on Canvas - feels buggy.
It can be one ticket: Layers rearrangement improvements, WDYT?

merapi (approved):
Works great 👍🏻 

(I merged main for that addGroup fix)

packages/story-editor/src/app/canvas/utils/getLayerArrangementProps.js
```diff
@@ -0,0 +1,83 @@
...
+    const isFirstInGroup = elements[currentPosition + 1]?.groupId !== groupId;
+    if (
+      (isLastInGroup && position === LAYER_DIRECTIONS.BACKWARD) ||
+      (isFirstInGroup && position === LAYER_DIRECTIONS.FORWARD)
+    ) {
+      return {
+        position: currentPosition,
+        groupId: null,
+      };
+    }
```
barklund, line 55: Note here, that if the element is both first and last and will be moved out of the group, the group should also be removed, which currently doesn't happen. If you move the last element out of a group, the group object persists in the groups object on the page, but it cannot be used anymore because no element references it.

We could add this to a follow-up, but I think we could just add a `removeGroup: isFirstInGroup && isLastInGroup` and then in `useCanvasKeys` check for that prop. Or we could update the underlying `arrangeElement` reducer to always check if a group is left empty and then delete it - that would probably be the proper way to do it actually.

barklund (approved):
This works extremely well. Excellent job here - it's a much cleaner approach than we originally discussed in the ticket. You kind of merged both ideas of the navigation here - with a great result to follow!

The caveats mentioned by @merapi are significant though. Especially the one about selecting a layer from the layer panel and not being able to rearrange with keyboard directly, but having to focus the canvas element in order for key events to be registered.

miina:
> > Looks like the shortcuts don't support rearranging groups currently at all, couldn't see an issue for it, should we add? cc @merapi
> 
> Yeah, I think we can also add a few more things that are missing:
> 
> * DND rearranging multiple layers (select 2 layers and DND them in the Layers Panel - doesn't work, you can only DND groups or single layers).
> * Same as above but via keyboard.
> * Keyboard rearrangement (of any kind) is not possible in the Layers Panel, you need to select the layer on Canvas - feels buggy.
>   It can be one ticket: Layers rearrangement improvements, WDYT?

Sounds good, I'll create a follow-up issue for these.

packages/story-editor/src/app/canvas/utils/getLayerArrangementProps.js
miina: Updated the reducer in [9c1f855](https://github.com/GoogleForCreators/web-stories-wp/pull/11884/commits/9c1f8555b5e9d484f5efc94af63799d413c5f86c), looks like generally the groups were not being cleaned up elsewhere either so didn't see any other code that needed adjusting. Let me know if I missed any place.

miina:
> > Looks like the shortcuts don't support rearranging groups currently at all, couldn't see an issue for it, should we add? cc @merapi
> 
> Yeah, I think we can also add a few more things that are missing:
> 
> * DND rearranging multiple layers (select 2 layers and DND them in the Layers Panel - doesn't work, you can only DND groups or single layers).
> * Same as above but via keyboard.
> * Keyboard rearrangement (of any kind) is not possible in the Layers Panel, you need to select the layer on Canvas - feels buggy.
>   It can be one ticket: Layers rearrangement improvements, WDYT?

Follow-up: #11906

PR #11886, opened 2022-07-07
Improving styled-components class names for debugging

### LINKED ISSUE #11749: Improving styled-components class names for debugging
## Task Description

Re-sharing feedback from https://github.com/GoogleForCreators/web-stories-wp/pull/11737#discussion_r901929974:

> I understand and agree with your point about consistency. This may be more of a concern related to the patterns required to use Styled Components than anything specific to this project.
> 
> A lot of the classNames generated by Styled Components can be redundant and hard to track up the component tree. For example, classNames like `components__Container-sc-xxxxxx` and `components__CardWrapper-sc-xxxxxx` are generic and lose meaning because they're repeatedly redefined in components to accommodate the patterns required for composition using the Styled Components library.
> 
> Going up the component tree offers a little context, but navigating the codebase to find specific instances of general-use components isn't obvious (maybe just to developers new to the codebase and patterns used, myself included).

### Additional Context

Related links:

https://styled-components.com/docs/tooling#better-debugging

spacedmonkey:
## Summary

Add the following to the babel config file 

```
{
  "plugins": [
    [
      "babel-plugin-styled-components",
      {
        "meaninglessFileNames": ["index", "styles", "components"]
      }
    ]
  ]
}
```

## Testing Instructions

1.

## Checklist

---

Fixes #11749

### DIFF
--- a/babel.config.cjs
+++ b/babel.config.cjs
@@ -47,9 +47,10 @@ module.exports = function (api) {
     plugins: [
       '@wordpress/babel-plugin-import-jsx-pragma',
       [
-        'babel-plugin-styled-components', {
-          'meaninglessFileNames': ['index', 'styles', 'components']
-        }
+        'babel-plugin-styled-components',
+        {
+          meaninglessFileNames: ['index', 'styles', 'components'],
+        },
       ],
     ],
     sourceMaps: true,
--- a/babel.config.cjs
+++ b/babel.config.cjs
@@ -46,7 +46,11 @@ module.exports = function (api) {
     ],
     plugins: [
       '@wordpress/babel-plugin-import-jsx-pragma',
-      'babel-plugin-styled-components',
+      [
+        'babel-plugin-styled-components', {
+          'meaninglessFileNames': ['index', 'styles', 'components']
+        }
+      ],
     ],
     sourceMaps: true,
     env: {


mflorida (commented):
What's the result in the output of adding these config parameters? Does it just clean up the generated classNames?

In general, I've found that adding meaningful className values to components to be super useful for both debugging and targeting child components for styling. It can reduce the amount of prop drilling necessary to conditionally style a child component based on state available to its parent.

swissspidy:
See https://styled-components.com/docs/tooling#control-which-file-names-are-meaningless for how this should affect output.

The change here is a first step at improving the auto-generated class names to make debugging easier.

As per https://github.com/GoogleForCreators/web-stories-wp/issues/11749#issuecomment-1169687938, I hesitate to just manually add class names, as it has downsides as well.

As for styling child components, we should rather reference the _components_ instead of class names, see example at https://styled-components.com/docs/advanced#referring-to-other-components

Let's continue discussing class names and stuff in the ticket and merge this as a first step towards that goal.

Hope that makes sense.

swissspidy (approved):

mflorida:
> https://styled-components.com/docs/advanced#referring-to-other-components

This is a nice pattern. 

One potential downside is that it only works with child components in the current scope, which may require importing from a separate module. Whether or not that's a good thing is subjective, but seems like a somewhat heavy-handed approach _just_ for styling. 

Using a reference is good for navigating through components in your IDE and avoiding specificity conflicts, but relying _solely_ on reference (making sure components are exported/imported and using only auto-generated classNames) can make state-based styling of deeply nested children more difficult, as well as making debugging in the browser a bit trickier.

Ultimately, whatever can reduce the number of 'containers' or 'wrappers' would be a _huge_ win, whether that's through styling by reference or className, or by simplifying the structure and more carefully defining styles.

mflorida (commented):
Babel config change presents essentially zero risk but huge potential reward in improving compiled output.

Issue #11206, opened 2022-04-06
Background audio: Allow users to drag audio onto canvas to set background audio


spacedmonkey:
## Feature Description

Allow users to set background audio by dragging an audio file ( like an MP3 ) onto canvas. On placement on canvas, this would upload the audio file and set the background audio for you. 

This brings background audio inline with page background, that allows you to set the background media via drag into canvas.

PR #11887, opened 2022-07-08
Audio Recording

### LINKED ISSUE #11079: Audio Capture: Initial implementation
## Feature Description

Enable support for audio-only recording, uploading the audio clip to WordPress and then setting it as page background audio.

**[Figma](https://www.figma.com/file/ZrO763yvMWJjYeAPsRnNfW/_Editor---Sprint-Feature-Level-Designs_2022?node-id=661%3A24394)**

While there are no designs for this yet, I'd imagine the flow to be somewhat like this:

1. When opening the media recording mode, by default we'll turn on the webcam and microphone. You see yourself on the canvas.
2. In the quick actions menu you can then disable either the camera or mic (but not both).
3. When you turn off the camera, instead of seeing yourself on the canvas we display an equalizer to visualize the sound
4. When you start recording, you continue to see the visualization
5. When you stop recording, we display a simple audio player to play/pause the audio
6. There is an "Insert background audio" button at the bottom
7. When you click on the button, we:
	1. Upload the audio file to WordPress via `uploadFile` from `useUploader`
	2. Take the resulting file URL and add it as page background audio
	3. Use `setHighlights` from `useHighlights` to focus the page background audio panel in the "Style" tab, so that the user knows where the audio file was added

## Additional Context

* `Stories_Media_Controller` needs updating to allow uploading audio files to WP

merapi:
## Summary

This PR provides audio-only recording, uploading the audio clip to WordPress and then setting it as page background audio (with highlight).

The flow is as follows:
1. When opening the media recording mode, by default we'll turn on the webcam and microphone. You see yourself on the canvas.
2. In the quick actions menu you can then disable either the camera or mic (but not both).
3. When you turn off the camera, instead of seeing yourself on the canvas we display an equalizer to visualize the sound
4. When you start recording, you continue to see the visualization
5. When you stop recording, we display a simple audio player to play/pause the audio
6. There is an "Insert background audio" button at the bottom
7. When you click on the button, we:
	1. Upload the audio file to WordPress via `uploadFile` from `useUploader`
	2. Take the resulting file URL and add it as page background audio
	3. Use `setHighlights` from `useHighlights` to focus the page background audio panel in the "Style" tab, so that the user knows where the audio file was added

## Relevant Technical Choices

Used file type: `audio/ogg`.

## To-do

- [ ] I made the video off icon myself, so not sure if it's fine.
- [x] Tests

## Testing Instructions

1. Follow the flow and check if it works as expected.

## Checklist

---

Fixes #11079


packages/story-editor/src/components/mediaRecording/provider.js
```diff
@@ -134,8 +139,8 @@ function MediaRecordingProvider({ children }) {
     // granted permission, so it's not easy to detect.
     // TODO: Figure out how to retry without microphone if possible.
     mediaStreamConstraints: {
-      audio: audioInput && hasAudio ? { deviceId: audioInput } : true,
-      video: videoInput ? { deviceId: videoInput } : true,
+      audio: audioInput && hasAudio ? { deviceId: audioInput } : false,
+      video: videoInput && hasVideo ? { deviceId: videoInput } : false,
```
swissspidy: We need to keep the default at `true`, otherwise you can end up with `audio: false, video: false,`, which will throw an error as you need to request at least one of them.

Also, if you're a brand new user (or when you reset the site's permissions at `chrome://settings/content`), the browser won't ask for microphone or camera permission and right now we don't have a way to re-grant permission.

That means you can end up in a state where you will have camera permission but not microphone permission, with no way out of there.

packages/story-editor/src/components/mediaRecording/footer.js
```diff
@@ -276,6 +282,37 @@ function Footer({ captureImage, videoRef }) {
...
+  } = useUploader();
+
+  const { updateStory } = useStory(({ actions: { updateStory } }) => ({
+    updateStory,
+  }));
+  const { setHighlights } = useHighlights(({ setHighlights }) => ({
+    setHighlights,
+  }));
+
+  const onAudioInsert = async () => {
```
swissspidy, line 305: I suppose this does _not_ need `useCallback` because it's passed directly to a button element and not another React component?

swissspidy (commented):
This is super cool! Really nice work 🚀 

Regarding the file type, that's a good topic, as it is relevant for both recording and playback.

For _playback_, Safari does not support ogg. MP3 seems to be the only format universally supported by all browsers.

For _recording_, according to `MediaRecorder.isTypeSupported()` browser support is like follows:

* Chrome: `audio/webm`
* Firefox: `audio/ogg`, `audio/webm`
* Safari: `audio/mp4`

For this reason I don't think we can just assume `audio/ogg` for the file type either.

It should be possible to run the file through FFmpeg in `onAudioInsert` before uploading, adding a new function to our existing `useFFmpeg` hook for it.

If you wanna take a stab at this, that would be excellent, otherwise I am happy to take a look too since I am familiar with that part of the code base.

packages/story-editor/src/components/mediaRecording/footer.js
merapi: Yep, this just goes to the button via styled Button, can't see `useCallback` benefits here.

packages/story-editor/src/components/mediaRecording/provider.js
merapi: 👍🏻  changed, but I will double-check how it behaves in a fresh browser session.

swissspidy: If there are issues, perhaps we can rewrite like this:

```js
      audio: hasAudio ? (audioInput ? { deviceId: audioInput } : true) : false,
      video: hasVideo ? (videoInput ? { deviceId: videoInput } : true) : false,
```

packages/story-editor/src/app/media/utils/useFFmpeg.js
```diff
@@ -457,6 +457,59 @@ function useFFmpeg() {
...
+          '-i',
+          file.name,
+          // Output filename. MUST be different from input filename.
+          tempFileName
+        );
+
+        const data = ffmpeg.FS('readFile', tempFileName);
+
+        return new blobToFile(
+          new Blob([data.buffer], { type: 'audio/mpeg' }),
```
spacedmonkey, line 490: Can we use `AUDIO_MIME_TYPE`?

packages/story-editor/src/app/media/utils/useFFmpeg.js
```diff
@@ -457,6 +457,59 @@ function useFFmpeg() {
...
+    async (file) => {
+      //eslint-disable-next-line @wordpress/no-unused-vars-before-return -- False positive because of the finally().
+      const trackTiming = getTimeTracker('load_mp3_conversion');
+
+      let ffmpeg;
+
+      try {
+        ffmpeg = await getFFmpegInstance(file);
+
+        const tempFileName = uuidv4() + '.mp3';
```
spacedmonkey, line 476: Can we use `AUDIO_FILE_TYPE`?

packages/story-editor/src/components/mediaRecording/provider.js
```diff
@@ -56,7 +58,10 @@ function MediaRecordingProvider({ children }) {
 
   const [isGif, setIsGif] = useState(false);
 
+  const [hasVideo, setHasVideo] = useState(true);
   const [hasAudio, setHasAudio] = useState(true);
+  const FILE_TYPE = hasVideo ? VIDEO_FILE_TYPE : AUDIO_FILE_TYPE;
```
spacedmonkey: This line should be moved to line 112, meaning FILE_TYPE, MIME_TYPE do not need to be passed in the hook.

packages/story-editor/src/components/mediaRecording/provider.js
```diff
@@ -106,12 +111,15 @@ function MediaRecordingProvider({ children }) {
       }
       const f = blobToFile(
         blob,
-        `webcam-capture-${format(new Date(), 'Y-m-d-H-i')}.${VIDEO_FILE_TYPE}`,
-        VIDEO_MIME_TYPE
+        `${hasVideo ? 'webcam' : 'audio'}-capture-${format(
+          new Date(),
+          'Y-m-d-H-i'
+        )}.${FILE_TYPE}`,
+        MIME_TYPE
```
spacedmonkey: This is starting to feel a little hard to read.

packages/story-editor/src/components/mediaRecording/footer.js
```diff
@@ -276,6 +283,40 @@ function Footer({ captureImage, videoRef }) {
...
+  }));
+  const { setHighlights } = useHighlights(({ setHighlights }) => ({
+    setHighlights,
+  }));
+
+  const { convertToMp3 } = useFFmpeg();
+
+  const onAudioInsert = async () => {
+    setIsInserting(true);
+    speak(__('Inserting…', 'web-stories'));
```
spacedmonkey, line 307: Are we inserting here? Are we not setting background audio.

packages/story-editor/src/components/mediaRecording/footer.js
```diff
@@ -287,10 +328,15 @@ function Footer({ captureImage, videoRef }) {
...
+          <InsertButton
+            onClick={hasVideo ? onInsert : onAudioInsert}
+            disabled={isInserting}
+          >
             {isInserting
               ? __('Inserting…', 'web-stories')
-              : __('Insert', 'web-stories')}
+              : hasVideo
+              ? __('Insert', 'web-stories')
+              : __('Insert background audio', 'web-stories')}
```
spacedmonkey, line 351: So it should be 
`Set background audio` and `Settings background audio`?

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
```diff
@@ -202,26 +205,40 @@ function MediaRecordingLayer() {
...
+                            muted={isMuted}
+                            loop={isGif}
+                            tabIndex={0}
+                          />
+                          {!isGif && <PlayPauseButton videoRef={videoRef} />}
+                        </>
+                      ) : (
+                        <>
+                          {/* eslint-disable-next-line jsx-a11y/media-has-caption -- No captions wanted/needed here. */}
+                          <audio controls="controls" src={mediaBlobUrl} />
```
spacedmonkey, line 228: See this comment https://github.com/GoogleForCreators/web-stories-wp/pull/11677#issuecomment-1156455571. 

We do not show the built in players for video / audio, as they look very different in different browsers. I would recommending show the visualizer and reuse play / pause icon that is used in video.

packages/story-editor/src/app/uploader/useUploader.js
```diff
@@ -53,8 +54,14 @@ function useUploader() {
...
       ...allowedVectorMimeTypes,
       ...allowedVideoMimeTypes,
+      ...allowedAudioMimeTypes,
     ],
-    [allowedImageMimeTypes, allowedVectorMimeTypes, allowedVideoMimeTypes]
+    [
+      allowedImageMimeTypes,
+      allowedVectorMimeTypes,
+      allowedVideoMimeTypes,
+      allowedAudioMimeTypes,
```
spacedmonkey: Screenshot 2022-07-11 at 14 06 14

Allowing this, means that this message is now incorrect. It is not correctly possible to drap an audio file into canvas. See https://github.com/GoogleForCreators/web-stories-wp/issues/11206

packages/story-editor/src/app/media/utils/useFFmpeg.js
```diff
@@ -457,6 +457,59 @@ function useFFmpeg() {
...
+          // Output filename. MUST be different from input filename.
+          tempFileName
+        );
+
+        const data = ffmpeg.FS('readFile', tempFileName);
+
+        return new blobToFile(
+          new Blob([data.buffer], { type: 'audio/mpeg' }),
+          outputFileName,
+          'audio/mpeg'
```
spacedmonkey, line 492: https://github.com/GoogleForCreators/web-stories-wp/pull/11887/files#r917885212

packages/story-editor/src/components/mediaRecording/audio.js
```diff
@@ -0,0 +1,119 @@
...
+  <AudioWrapper>
+    <AudioAnalyser source={liveStream} />
+  </AudioWrapper>
+);
+
+Audio.propTypes = {
+  liveStream: PropTypes.object,
+};
+
+const AudioAnalyser = (props) => {
```
spacedmonkey:
```suggestion
const AudioAnalyser = ({ source }) => {
```
Nit pick, but props isn't very clear here.

packages/story-editor/src/app/uploader/useUploader.js
spacedmonkey: This also means unit tests are breaking.

spacedmonkey:
Screenshot 2022-07-11 at 13 54 35
Screenshot 2022-07-11 at 14 17 22

Should these be changed to Record video or audio?

spacedmonkey:
@swissspidy I believe this ticket would easier for this ticket to get merged, if https://github.com/GoogleForCreators/web-stories-wp/issues/11206 is actioned first. We do not currently upload audio files via the REST API and validation does not support it. Without doing this work for, then this is going to be an issue. https://github.com/GoogleForCreators/web-stories-wp/pull/11887#discussion_r917908903

packages/story-editor/src/components/mediaRecording/footer.js
swissspidy: Inserting as background audio.

The current wording works for me. We can do a holistic review of all media recording labels separately if needed.

packages/story-editor/src/app/uploader/useUploader.js
swissspidy: How else should we solve this then?

Seems like it warrants its own ticket.

swissspidy:
#11206 is not a priority.

packages/story-editor/src/app/uploader/useUploader.js
spacedmonkey: The issue is that if you add audio mime types to the allowed list, then it will appear on this message. It means you can drag an mp3 into canvas. A quick work around would be have some kind of override for this one case and fix in https://github.com/GoogleForCreators/web-stories-wp/issues/11206. Or just do work on https://github.com/GoogleForCreators/web-stories-wp/issues/11206

spacedmonkey: @swissspidy @merapi I put together a little workaround for this issue - https://github.com/GoogleForCreators/web-stories-wp/pull/11892

See what you think.

packages/story-editor/src/components/mediaRecording/footer.js
spacedmonkey: Nothing is being inserting. That is bad english. Set is the correct wording here.

packages/story-editor/src/components/mediaRecording/audio.js
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * External dependencies
+ */
+import { useEffect, useRef, useState } from '@googleforcreators/react';
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+
+const AudioWrapper = styled.div`
+  background: #26292a;
```
spacedmonkey: Should this value be hardcoded like this?

spacedmonkey: Could we just use brand.gray.80?

packages/story-editor/src/app/media/utils/useFFmpeg.js
merapi: This is a `convertToMp3` function so we don't want to use `AUDIO_MIME_TYPE` here - it might change in the future to superior OGG (as soon as Safari starts fully supporting it).

packages/story-editor/src/app/media/utils/useFFmpeg.js
merapi: Ditto https://github.com/GoogleForCreators/web-stories-wp/pull/11887#discussion_r918123364

packages/story-editor/src/app/media/utils/useFFmpeg.js
spacedmonkey: Okay, elsewhere we have use const for mime and type ext. We should do the same thing.

packages/story-editor/src/components/mediaRecording/footer.js
merapi: We are converting the file, uploading (inserting) the file onto the server and setting the audio bg property 🤔 
Setting background audio is just a fraction of the time this label is visible, we are mainly uploading/converting, I guess the upload part is the longest, so maybe `Uploading...` label would be the best? the user would know that it's network-dependent action and can take time. UX should handle this? holistic review +1.

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
merapi: Note: visualizer will display just the slice of the audio, not the whole timeline with scrubbing abilities - that would be actually useful but I think we might introduce the audio trimmer in the near future? @swissspidy

IMO for a final file preview, the default audio player with a scrubber is far more useful than a slice visualizer without a scrubber. Only trimmer would be superior.

spacedmonkey: We do not have scrubbing for video. So for consistency, let's not have it here. I think once trimming is introduced, then we can add scrubbing.

packages/story-editor/src/components/mediaRecording/provider.js
merapi: I think we are okay here with true/true, the difference I saw was the size of the output file from the device, when `audio=true, video=true (just hidden)` on 10s recording was `1987172` bytes, while `audio=true, video=false` was `64623` bytes, but when converted to mp3 both ended up with `~82605` bytes.

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
swissspidy: I'm fine with having the default audio player here with a scrubber & time display, until we come up with something better in the future.

packages/story-editor/src/app/media/utils/useFFmpeg.js
merapi: But we export/reuse them or they might change, mp3 file type will not change nor do we need to explain the meaning behind the value (like we do for numbers), here it is only for this function so using 
```js
const MP3_FILE_TYPE = 'mp3'
+
..."." + MP3_FILE_TYPE`
```
 doesn't bring any value over
```js
...".mp3"
```
in a `convertToMp3` function.
In this case it's just too much, what's more, if you make a variable out of it you "encourage" people to change it - if they change it it will break the function, it should be hardcoded.
We export `AUDIO_FILE_TYPE` because it makes sense, it might/will change/be reused.

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
merapi: Trimming for video is introduced as we speak.

packages/story-editor/src/components/mediaRecording/audio.js
merapi: Yeah, I came up with a better naming 👍🏻

merapi:
> Should these be changed to Record video or audio?

Media panel: renamed to `Record Video or Background Audio`.
Empty state: not sure.

packages/story-editor/src/components/mediaRecording/footer.js
```diff
@@ -301,18 +357,25 @@ function Footer({ captureImage, videoRef }) {
...
+                  <Icons.PhotoCamera width={24} height={24} aria-hidden />
+                  {__('Take a photo', 'web-stories')}
+                </CaptureButton>
+              </>
+            ) : (
               <RecordingButton onClick={onStart}>
-                <Icons.Camera width={24} height={24} aria-hidden />
-                {__('Record Video', 'web-stories')}
+                <Icons.Mic width={24} height={24} aria-hidden />
+                {__('Record Audio', 'web-stories')}
```
miina, line 378: Hm, audio can be recorded only when the user first disables the video from the menu bar? A bit hard to discover but might be just me :)

miina (approved):
I don't have a lot of context to this feature in terms of UX decisions / existing video recording but the audio recording addition looks good and works great! 🎶

packages/story-editor/src/components/mediaRecording/footer.js
merapi: We can do `[ Record Audio (Only) ]` button, WDYT @swissspidy ?

packages/story-editor/src/components/mediaRecording/footer.js
```diff
@@ -276,6 +287,46 @@ function Footer({ captureImage, videoRef }) {
...
+    const mp3File = await convertToMp3(file);
+    const resource = await uploadFile(
+      mp3File,
+      {
+        mediaSource: 'recording',
+      },
+      allowedAudioMimeTypes
+    );
+    const backgroundAudio = {
+      resource,
```
spacedmonkey: Background audio is not a normal resource here. BackgroundAudioPropType only has some fields. We should pick out just these fields.

packages/story-editor/src/components/mediaRecording/footer.js
swissspidy: Let's handle UX/UI improvements like that in a follow-up ticket, shouldn't hold off this PR.

packages/story-editor/src/app/uploader/useUploader.js
```diff
@@ -80,12 +68,19 @@ function useUploader() {
...
+   * @param {boolean} args.isFileTooLarge Whether file is too large for consumer.
+   * @param {Array} args.overrideAllowedMimeTypes Array of override allowed mime types.
    */
   const validateFileForUpload = useCallback(
-    (file, canTranscodeFile, isFileTooLarge) => {
+    ({
+      file,
+      canTranscodeFile,
+      isFileTooLarge,
+      overrideAllowedMimeTypes = allowedMimeTypes,
```
swissspidy, line 82: Nit: I'd probably rename this to make less confusing

```suggestion
      allowedMimeTypes = defaultAllowedMimeTypes,
```

packages/story-editor/src/components/library/panes/media/local/mediaRecording/mediaRecording.js
```diff
@@ -78,7 +78,7 @@ function MediaRecording() {
     return null;
   }
 
-  const label = __('Record Video', 'web-stories');
+  const label = __('Record Video or Background Audio', 'web-stories');
```
swissspidy:
```suggestion
  const label = __('Record Video/Audio', 'web-stories');
```

packages/story-editor/src/components/mediaRecording/footer.js
swissspidy: Current wording is OK for now

swissspidy (approved):
Overall looks good 👍

packages/story-editor/src/app/uploader/useUploader.js
merapi: `allowedMimeTypes` is already declared in the upper scope so we would have to rename quite a few variables.

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
```diff
@@ -87,6 +87,16 @@ const MicOff = styled(Icons.MicOff).attrs({
...
 
+const Video = styled(Icons.Camera).attrs({
+  width: 24,
+  height: 24,
+})``;
+
+const VideoOff = styled(Icons.CameraOff).attrs({
+  width: 24,
+  height: 24,
+})``;
```
spacedmonkey: Nitpic. Could we do this?
```suggestion
const videoIconSizes = {
  width: 24,
  height: 24,
};
const Video = styled(Icons.Camera).attrs(videoIconSizes)``;
const VideoOff = styled(Icons.CameraOff).attrs(videoIconSizes)``;
```

packages/story-editor/src/app/uploader/useUploader.js
spacedmonkey: This is why I did it this way. But I am not against changing it.

spacedmonkey (changes requested):
Couple of nitpics and one thing to action.

packages/story-editor/src/components/mediaRecording/footer.js
merapi: That should set us for years to come 😉  https://github.com/GoogleForCreators/web-stories-wp/pull/11887/commits/8016893d0d20c4d981ec2de73edf95d0d1a6eec6

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
merapi: Did it for all of them 👍🏻  https://github.com/GoogleForCreators/web-stories-wp/pull/11887/commits/a466dba336076c205a769fc227c8c3eb0d8455ef

spacedmonkey (approved):

Issue #11402, opened 2022-04-29
Editor: Add Semantic Headings to the right click menu of Text Elements


bmattb:
## Feature Description

Now that automatic semantic headings are applied to text elements within stories (see https://github.com/googleforcreators/web-stories-wp/issues/5326), users are requesting a way to be able to view what that value is or will be on publish so adjustments can be made beforehand.

This feature will allow a user to select the text within a page and see and change the text element within the sidebar.

Many thanks to @swissspidy for the implementation idea, and to @aaskedall for the UX guidance.

## Acceptance Criteria

- This should be under a new experimental feature flag.
  -  The semantic headings experiment flag is on the branch titled: `feature/11788-experiment-semantic-headings`... Please branch off of this when working on this feature.
  - For an example of how to use it with a feature, see this PR: https://github.com/GoogleForCreators/web-stories-wp/pull/11626/files
- Within the right click menu of a text element on the page, there should be a new item called Heading Level
- This Heading Level item should have a submenu that show the following properties (along with the one currently selected):
	- `Automatic (xyz)` (stored as `auto`)
	- `Heading 1` (stored as `h1`)
	- `Heading 2` (stored as `h2`)
	- `Heading 3` (stored as `h3`)
	- `Paragraph` (stored as `p`)
- When the menu item is changed to another option in the menu, it should update the text element's `tagName property` 

## Developer Notes

Notes from @swissspidy  on implementation:
`xyz` is replaced with whatever `getTextElementTagNames()` returns for the element.

When changing this option, the text element's `tagName` property is updated.

`getTextElementTagNames()` is updated to take this `tagName` property into account.

## Alternatives Considered

In addition to this space, this functionality will also be accomplished within the sidebar of the Editor.

bmattb:
Andrew to provide UX guidance on this one

sblinde:
Our plan is to include this in the upcoming release, so this story will be pulled in to the sprint once we have UX (which I hope to talk about on Tuesday!)

felipebochehin87:
Issue found. Tested against https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/, using PR #12062.

1. Create a new story
2. Add two "Title 1" to the canvas
3. Switch their position - the top one is set to H1, the other is set to H2 - OK
4. Right click on the lower "Title 1" and change the heading to anything but H1 or H2
5. Click out of the element
[...]

PR #11888, opened 2022-07-08
Editor: Add Semantic Headings panel to Style tab

### LINKED ISSUE #11788: Editor: Add Semantic Headings to the Editor Sidebar
## Feature Description

This feature will allow a user to select the text within a page and see and change the text element within the sidebar.

Many thanks to @swissspidy for the implementation idea, and to @aaskedall for the UX guidance.

## Acceptance Criteria

- This should be under a new experimental feature flag.
  -  The semantic headings experiment flag is on the branch titled: `feature/11788-experiment-semantic-headings`... Please branch off of this when working on this feature.
  - For an example of how to use it with a feature, see this PR: https://github.com/GoogleForCreators/web-stories-wp/pull/11626/files
- Within the Accessibility area of the Editor when text is selected, there should be a new Dropdown
- This dropdown should be titled "Heading Level" and should show the following properties:
	- `Automatic (xyz)` (stored as `auto`)
	- `Heading 1` (stored as `h1`)
	- `Heading 2` (stored as `h2`)
	- `Heading 3` (stored as `h3`)
	- `Paragraph` (stored as `p`)
- When the dropdown item is changed to another option in the dropdown, it should update the text element's `tagName property` 

Example of implementation from @swissspidy (approved by @aaskedall):
Screenshot 2022-05-30 at 16 19 17

## Developer Notes

Notes from @swissspidy  on implementation:
`xyz` is replaced with whatever `getTextElementTagNames()` returns for the element.

When changing this option, the text element's `tagName` property is updated.

`getTextElementTagNames()` is updated to take this `tagName` property into account.

Below the dropdown we can add a helper text `<a>Learn more</a> about using headings in Web Stories.`, where we can link to the docs

## Alternatives Considered

In addition to this space, this functionality will also be accomplished within a right-click menu on the text element.

### LINKED ISSUE #11916: Semantic Heading indicator "Learn More" text is too large
## Bug Description

originally reported by @sblinde 
Font size for “Learn More” needs to be smaller
image2

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

sblinde:
## Context

This PR allows a user to select one or multiple text elements within the canvas and either view or change the semantic elements used to render text.

## Summary

This PR adds a new experiment. When this experiment is enabled, a user can change the semantic heading of an element of text within the Editor. This does NOT alter the size or style of the text, only the semantic element within the output HTML (such as an `h1` or `h3`, etc).

## User-facing changes

Screen Shot 2022-07-08 at 1 21 23 PM

https://user-images.githubusercontent.com/841956/179558369-af622a63-f470-4ca7-9cb7-3e8803f45b5f.mov

## Testing Instructions

1. Enable the `Semantic Headings` experiment in Stories > Experiments.
2. Navigate to a Story within the Editor.
3. Add multiple text elements with variying font sizes.
4. You should be able to see in the Style section of the sidebar, underneath a new Accessibility tab, the type of semantic heading.
5. You should be able to change the semantic heading, and publish your changes.
6. After this is done, the element within the dev tools should reflect the semantic heading change.
7. You should also be able to do this for multiple text elements at once (if the text elements have mismatching types when initially selected, i.e. a Heading 1 and a Heading 2, you should see "Mixed" as the current value in the dropdown).

Things to confirm:

* If multiple items that are Automatic are selected, it should show 'Automatic (Mixed)'
* If two items are selected, and one was set automatic and the other was not, it should be indicated that it is Mixed.
* When there are two large elements with equal font sizes are set to "Automatic", and you move one of them higher up on the canvas, it should become the h1, and if you move it down, it will become the h2 (the dropdown should update accordingly)

No

No

No

## Checklist

---

Fixes #11788
Fixes #11916


packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
```diff
@@ -0,0 +1,167 @@
...
+      css={highlight && styles.FLASH}
+      onAnimationEnd={() => resetHighlight()}
+      name="textAccessibility"
+      title={__('Accessibility', 'web-stories')}
+      isPersistable={!highlight}
+    >
+      <Row>
+        <DropDown
+          ref={(node) => {
+            if (node && highlight?.focus && highlight?.showEffect) {
```
miina: Q: I don't currently see a way in the editor to actually highlight the panel, is this something that will be done in a follow-up issue?

sblinde: @miina This is existing logic from another panel used in this space.

I do see a highlight around the object and I am able to navigate with it using keyboard navigation, as well as the other panels within this space. Are you not seeing this work within the QA instance?

Screen Shot 2022-07-11 at 12 49 17 PM
Screen Shot 2022-07-11 at 12 49 39 PM

packages/output/src/utils/getTextElementTagNames.js
```diff
@@ -37,9 +38,24 @@ function getTextElementTagNames(elements) {
...
        * @return {Map<string, string>} Tag names map.
        */
-      (tagNamesMap, { id, fontSize, content }) => {
+      (tagNamesMap, { id, fontSize, content, tagName }) => {
+        // if a new tagName is added,
+        // add it to the element
+        if (newTag) {
+          tagNamesMap.set(id, newTag);
+          return tagNamesMap;
+        }
```
swissspidy: This is a bit confusing since it sets `newTag` on _every_ element. Is it really intended this way?

includes/Experiments.php
```diff
@@ -369,6 +369,17 @@ public function get_experiments(): array {
...
 			],
+			/**
+			 * Author: @sblinde
+			 * Issue: #5326
+			 * Creation date: 2022-06-29
+			 */
+			[
+				'name'        => 'showSemanticHeadings',
+				'label'       => __( 'Semantic headings', 'web-stories' ),
+				'description' => __( 'Display semantic heading tags', 'web-stories' ),
```
swissspidy: Slightly misleading because we already do use semantic heading tags; this is just about exposing a UI for them.

Maybe something like this:

```suggestion
				'description' => __( 'Display UI for changing semantic heading tags', 'web-stories' ),
```

Not a big deal though since this will be a rather short-lived feature flag

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
```diff
@@ -0,0 +1,167 @@
...
+          dropDownLabel={__('Heading Level', 'web-stories')}
+        />
+      </Row>
+      <Text size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.X_SMALL}>
+        <TranslateWithMarkup
+          mapping={{
+            a: (
+              <a
+                href={__(
+                  'https://github.com/GoogleForCreators/web-stories-wp/tree/main/docs',
```
swissspidy: I will provide you with the correct link ASAP.

In the meantime, note that this is the link to the developer-facing docs. User docs are all at `https://wp.stories.google/docs/`, so you can use that for now.

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
```diff
@@ -0,0 +1,167 @@
...
+          }
+          onMenuItemClick={handleChange}
+          dropDownLabel={__('Heading Level', 'web-stories')}
+        />
+      </Row>
+      <Text size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.X_SMALL}>
+        <TranslateWithMarkup
+          mapping={{
+            a: (
+              <a
```
swissspidy: Let's add an `onClick` handler with `trackClick(evt, 'click_headings_docs');` to this link

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
```diff
@@ -0,0 +1,167 @@
...
+    // Update the selected text elements with the
+    // value chosen in the text accessibility dropdown
+    const newTags = getTextElementTagNames(selectedTextElements, value);
+    setCurrentTags(Array.from(newTags.values()));
+  };
+  // Check if tags match for selected item in dropdown
+  // If they don't match, we want to show 'Mixed'
+  // However, if they match, show the matching value
+  const currentValue =
+    new Set(currentTags || []).size === 1 ? currentTags[0] : '((MULTIPLE))';
```
swissspidy: Let's use the available const for that.

```suggestion
    new Set(currentTags || []).size === 1 ? currentTags[0] : MULTIPLE_VALUE;
```

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
```diff
@@ -0,0 +1,167 @@
...
+];
+
+function TextAccessibilityPanel({ selectedElements, pushUpdate }) {
+  // Feature flagging for Semantic Headings
+  const showSemanticHeadings = useFeature('showSemanticHeadings');
+  const [selectedTextElements, setSelectedTextElements] = useState([]);
+  const [currentTags, setCurrentTags] = useState([]);
+  useEffect(() => {
+    // We only want to impact elements if all selected elements are of type text
+    const textElements = selectedElements.filter(({ type }) => 'text' === type);
```
swissspidy: Let's move this out of the effect so that the effect only runs when `textElements` changes.

Theoretically it should not be needed though since this panel is only displayed for text elements anyway, so I guess this filter can just as well be removed

packages/output/src/utils/getTextElementTagNames.js
```diff
@@ -37,9 +38,24 @@ function getTextElementTagNames(elements) {
...
        * @param {number} element.fontSize Font size.
        * @param {string} element.content Text content.
+       * @param {string} element.tagName Existing tagName
        * @return {Map<string, string>} Tag names map.
        */
-      (tagNamesMap, { id, fontSize, content }) => {
+      (tagNamesMap, { id, fontSize, content, tagName }) => {
+        // if a new tagName is added,
+        // add it to the element
+        if (newTag) {
```
swissspidy: This function is sometimes called with `newTag = 'auto'`, and we don't wanna set `auto` here.

packages/output/src/utils/getTextElementTagNames.js
```diff
@@ -37,9 +38,24 @@ function getTextElementTagNames(elements) {
...
+          tagNamesMap.set(id, newTag);
+          return tagNamesMap;
+        }
+
+        // if a tag already exists,
+        // utilize the one already part of the element
+        if (tagName) {
```
swissspidy: If a user chooses "Automatic" from the dropdown, we set `tagName = 'auto'`, on the element. We don't wanna actually set that as a tag name though.

```suggestion
        if (tagName && tagName !== 'auto') {
```

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
```diff
@@ -0,0 +1,167 @@
...
+    value: 'h3',
+  },
+  {
+    label: __('Paragraph', 'web-stories'),
+    value: 'p',
+  },
+  {
+    label: __('Automatic', 'web-stories'),
+    value: 'auto',
+  },
```
swissspidy: Let's move this to the beginning of the list since this is suppose to be the default behavior.

swissspidy (commented):
Overall looks good, it's getting close!

See also #11905 which directly maps to some of my comments.

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
miina: @sblinde Highlighting is a bit different feature, it allows highlighting effect, for example when linking to a panel from somewhere else in the editor. Example: when selecting an empty page and clicking on the "Change background color" quick action, the Background panel is opened and highlighted

highlighting

Will this panel be linked to from somewhere else? If not then there's no need for the highlighting logic here either.

sblinde: Aha! Thanks for the explanation. And I think you're right: that's not an intended usage as I'm aware of, so I can strip that logic out.

packages/output/src/utils/getTextElementTagNames.js
sblinde: It only will if a second argument with a tag name is added. The one time it's called with an additional argument is in the usage of the component when setting a new tag that's a change (and, would be used also in the right click menu for text once that feature is built).

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: Excellent! Thank you!

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: Can do!

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: Will do!

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: Looking on it, I think you're right, none of this should show if they aren't text. I can do that!

packages/output/src/utils/getTextElementTagNames.js
sblinde: Aha! Okay, I wasn't aware. And I saw the PR you opened with suggested changes to this logic, thank you!

packages/output/src/utils/getTextElementTagNames.js
sblinde: And I see now that you have explained further in this PR that there are places where auto is set, so your change in your PR would be the best approach for this given that information!

sblinde:
Going to merge in @swissspidy 's suggestions from https://github.com/GoogleForCreators/web-stories-wp/pull/11905 and continue to address feedback after that initial merge. Many thanks so far for the feedback!

sblinde:
> > Also what's with this horizontal scrollbar? It doesn't seem to happen on `main`
> > Screenshot 2022-07-11 at 23 06 51
> 
> Thanks for catching this, I'll check it out!

Took a look, it's pre-existing and not a part of this work. It's actually when you don't have any saved styles, so when you do have saved styles, you don't see the scroll bar. You can check it out on the `new` staging URL environment to see it.

Screen Shot 2022-07-12 at 1 24 05 PM

swissspidy:
^ @merapi can Prometheus pod look into that bug? Thx!

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: Removed it with this commit, thanks! https://github.com/GoogleForCreators/web-stories-wp/pull/11888/commits/0406524707b2ba47c41ac958eaccac4567c1d98d

packages/output/src/utils/getTextElementTagNames.js
sblinde: I added a test as well for this logic, given that we don't want to use `auto` if passed. https://github.com/GoogleForCreators/web-stories-wp/pull/11888/commits/0406524707b2ba47c41ac958eaccac4567c1d98d

includes/Experiments.php
sblinde: Awesome! Did just that, thank you! https://github.com/GoogleForCreators/web-stories-wp/pull/11888/commits/0406524707b2ba47c41ac958eaccac4567c1d98d

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: Added: https://github.com/GoogleForCreators/web-stories-wp/pull/11888/commits/0406524707b2ba47c41ac958eaccac4567c1d98d

Just made it dashes instead of underscores based on some other usage I saw in the application, but if there's a pattern for naming I'm missing and underscore is preferred, let me know.

packages/output/src/utils/getTextElementTagNames.js
sblinde: Addressed with a test in https://github.com/GoogleForCreators/web-stories-wp/pull/11888/commits/0406524707b2ba47c41ac958eaccac4567c1d98d

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: Updated in: https://github.com/GoogleForCreators/web-stories-wp/pull/11888/commits/0406524707b2ba47c41ac958eaccac4567c1d98d

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: Was changed as part of merging in https://github.com/GoogleForCreators/web-stories-wp/pull/11888/commits/6dc71c5d492413303d5c952fe5ccfef660dcb476

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: Done in https://github.com/GoogleForCreators/web-stories-wp/pull/11888/commits/6dc71c5d492413303d5c952fe5ccfef660dcb476

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: https://github.com/GoogleForCreators/web-stories-wp/pull/11888/commits/6dc71c5d492413303d5c952fe5ccfef660dcb476

mwritter (approved):
Looks good to me

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
```diff
@@ -0,0 +1,168 @@
...
+      </Row>
+      <Text size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.X_SMALL}>
+        <TranslateWithMarkup
+          mapping={{
+            a: (
+              <a
+                href={__('https://wp.stories.google/docs', 'web-stories')}
+                rel="noreferrer"
+                target="_blank"
+                aria-label={__('Learn more', 'web-stories')}
```
swissspidy: This should not be needed.

`aria-label` just overrides the link text in the a11y tree, and here the link text and the `aria-label` are identical. So we can just remove the `aria-label`.

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
swissspidy: @sblinde We want to track clicks on the "Learn more" link, not clicks on the dropdown. Sorry for the confusion.

And all underscores please. `click_headings_docs`. We don't use dashes for event names.

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
```diff
@@ -0,0 +1,168 @@
...
+  }, [selectedElements, setSelectedTextElements]);
+
+  if (!showSemanticHeadings) {
+    return null;
+  }
+
+  // Map all types of tag names in the selected elements
+  // and then convert to an Array for usage
+  const handleChange = (evt, value) => {
+    trackClick(evt, 'click-headings-docs');
```
swissspidy: As per my other comment, this should be removed. We want to track clicks on the "Learn more" link.

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
```diff
@@ -0,0 +1,168 @@
...
+    selectedElements.map(() => {
+      return pushUpdate({ tagName: value });
+    });
```
swissspidy: This should use `.forEach` instead, since a) we don't intend modify `selectedElements` here and b) `pushUpdate` does not return anything, so this is a bit misleading and potentially error prone.

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
```diff
@@ -0,0 +1,168 @@
...
+    setCurrentTags(Array.from(textTags.values()));
+  }, [selectedElements, setSelectedTextElements]);
+
+  if (!showSemanticHeadings) {
+    return null;
+  }
+
+  // Map all types of tag names in the selected elements
+  // and then convert to an Array for usage
+  const handleChange = (evt, value) => {
```
swissspidy: Worth using `useCallback`?

packages/output/src/utils/getTextElementTagNames.js
```diff
@@ -37,9 +38,24 @@ function getTextElementTagNames(elements) {
...
        * @return {Map<string, string>} Tag names map.
        */
-      (tagNamesMap, { id, fontSize, content }) => {
+      (tagNamesMap, { id, fontSize, content, tagName }) => {
+        // if a new tagName is added,
+        // add it to the element
+        if (newTag && newTag !== 'auto') {
+          tagNamesMap.set(id, newTag);
+          return tagNamesMap;
+        }
```
swissspidy: I'm still a bit confused by that behavior, since it just forces every single element to have the same tagName. Is that we want?

If I am reading the `handleChange` code in the panel correctly, we want a behavior that is more like this:

> When selecting tag name X from the dropdown, call `getTextElementTagNames()` but force tag name X for the selected element(s)
> This way, if I for example select h1 from the dropdown, we can force this element to be an h1 and preventing other elements (which might be earlier in the list) to be the h1

WDYT?

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
```diff
@@ -0,0 +1,168 @@
...
+                href={__('https://wp.stories.google/docs', 'web-stories')}
+                rel="noreferrer"
+                target="_blank"
+                aria-label={__('Learn more', 'web-stories')}
+              />
+            ),
+          }}
+        >
+          {__(
+            '<a>Learn more</a> about using headings in Stories.',
```
swissspidy:
```suggestion
            '<a>Learn more</a> about using headings in stories.',
```

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
```diff
@@ -0,0 +1,168 @@
...
+          onMenuItemClick={handleChange}
+          dropDownLabel={__('Heading Level', 'web-stories')}
+        />
+      </Row>
+      <Text size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.X_SMALL}>
+        <TranslateWithMarkup
+          mapping={{
+            a: (
+              <a
+                href={__('https://wp.stories.google/docs', 'web-stories')}
```
swissspidy:
```suggestion
                href={__('https://wp.stories.google/docs/seo/', 'web-stories')}
```

This will be the final URL.

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: Ah, that totally makes sense. Can do, and will change the underscores, thanks.

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: Could do

packages/output/src/utils/getTextElementTagNames.js
sblinde: So, if we're using this function within this component, we only want to alter the selected elements, and that's what's called with the change with the newTag. If one element is selected, that's the only element we send. If someone selects five text elements and hits the dropdown for a new option, then we make the call and change each element, all five that are selected. This allows a user to update in bulk.

With this use case, we don't send anything other than the elements that a user has selected in the editor. We're not changing every single element, only the ones we've selected, and the ones the user has clicked a dropdown option for. So, if there's two text elements on a page (say, a Paragraph and a Heading 1) and the user selects only the Paragraph and changes it to a Heading 2, then they have a Heading 2 (former Paragraph) and the Heading 1 elements. The change is contained to the elements one has selected.

Does that help make a bit more sense?

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: Awesome! Thank you!

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: Thank you!

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: Addressed in https://github.com/GoogleForCreators/web-stories-wp/pull/11888/commits/705ece200af422cea010cd4b6b17fe5843cddee3

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: Removed and addressed in https://github.com/GoogleForCreators/web-stories-wp/pull/11888/commits/705ece200af422cea010cd4b6b17fe5843cddee3

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: Removed, thank you! https://github.com/GoogleForCreators/web-stories-wp/pull/11888/commits/705ece200af422cea010cd4b6b17fe5843cddee3

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: Indeed, sounds good to me. Changed to `forEach` https://github.com/GoogleForCreators/web-stories-wp/pull/11888/commits/705ece200af422cea010cd4b6b17fe5843cddee3

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: https://github.com/GoogleForCreators/web-stories-wp/pull/11888/commits/705ece200af422cea010cd4b6b17fe5843cddee3

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: https://github.com/GoogleForCreators/web-stories-wp/pull/11888/commits/705ece200af422cea010cd4b6b17fe5843cddee3

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
sblinde: URL updated https://github.com/GoogleForCreators/web-stories-wp/pull/11888/commits/705ece200af422cea010cd4b6b17fe5843cddee3

packages/output/src/utils/getTextElementTagNames.js
swissspidy: > With this use case, we don't send anything other than the elements that a user has selected in the editor.

Right! I didn't realize it only calls the function only on selected elements 🤦 

It does make more sense now, but the issue is that we have to always call the function on all text elements on a page.

Otherwise the behavior is not correct.

**Example:**

* Element A, font size 100
* Element B, font size 60

When previewing the story, this would result in element A being the h1, B being h2.

Now, if you only select element B, the dropdown will incorrectly say "Automatic (Heading 1)"

Screenshot 2022-07-14 at 15 51 22

Likewise, if you select element A and change the dropdown to "Heading 2", element B will become the Heading 1.

To ensure that, we need to call the function _on all text elements_ but force element A to be an h2

sblinde: @swissspidy Thanks for explaining that, I see what you mean with this. This only has the knowledge of the selected elements, not all elements. Hmmm. Let me see what I can do with this.

sblinde:
Bugs reported as part of bug bash:
1. If multiple items that are Automatic are selected, it shows 'Automatic ()' instead of 'Automatic (Mixed)'
2. If two items are selected, and one was set automatic and the other was not, it should be indicated that it is Mixed.
3. Font size for the Learn More button needs to be smaller, should double check that.

sblinde:
Latest push was an end of the day push of code changes, note that this PR needs a little bit more work to support the idea of Automatic (although, may be difficult to accomplish given defining headings in the story provider, can't determine what's automatic when everything now has a tag and `auto` is overridden`). 

Tests may be temporarily broken and functionality may not be perfect until I'm back on tomorrow to finish up writing out that logic.

sblinde:
I think the last refactor of this brings some things to mind that should be considered with building this going forward:

The last refactor tackled a few of the things we talked about: it added the tag name to the elements at the top level within the StoryProvider. That way, the calculated "tagName" is based on the algorithmic logic within `getTextElementTagNames` and inclusive of both the selected tag as well as the other tags on the page being viewed/edited itself. So there doesn't need to be a component state anymore for the tag names or the elements.

However, this change introduces a fork in the road in terms of what we move forward to build past this:

### Solution 1
**We continue to support and allow for someone to make the tag name "Automatic" within the dropdown:**
1. I we use the `getTextElementTagNames` functionality, the way it's built right now with the latest changes, will change anything that is set to an `auto` tagName to display as the appropriate text. So we should only get `h1`, `h2`, `h3`, `p`.  But there is no way to indicate in the UI that a `tagName` has been assigned automatically by the `getTextElementTagNames` function, or to know that a user set it.
2. If we do support "Automatic", we may need a property on the element to indicate that it has been set automatically and not in actual data. Meaning, if the algorithm chose for a text element to be `h1`, we could indicate that it was automatically selected in the UI with something like `automaticAssignedTagName: true` or a property like that. This could be set in the `getTextElementTagNames` function, and any element that does not have an actual `tagName` as part of its data would be `automaticAssignedTagName: false`. This is a way we could accomplish allowing someone to automatically assign the tag name in the dropdown, and also show the user that the semantic heading was chosen automatically.

### Solution 2
**We no longer support `auto` and users can choose only the predefined options.**
1. The change made to this code will not display any text element as `auto`, because of the way `getTextElementTagNames`  works (`auto` will be overridden by other logic). `Automatic` is removed from the dropdown.
2. We show a user that they either have an `h1`, `h2`, `h3`, or `p` element, and we allow them to change to any of these, but automatically setting it with the dropdown is not a feature.

I think that's the path this code needs to take in order to finish this feature, and because the most ideal solution to add functionality (keeping automatic) means adding a property to the element that wasn't previously defined, I leave it in the hands of the teams involved to decide.


### To Do:
1. Depending on the solution at hand, we'll need to accomodate either removing `auto` support in the dropdown, or add the property within the `getTextElementTagNames` to support a user knowing whether the text is automatic.
2. A test will need to be updated for the addition of tag names to the storyProvider.
3. Depending on the path decided for the rest of the work, some logic remains to be built for what displays in the dropdown if automatic is supported.


cc: @swissspidy

sblinde:
Due to the large refactor to include tagNames as part of the story provider, going to move this back to `In Progress` so it can be QA'd after the proposed changes were made.

swissspidy (approved):

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
```diff
@@ -0,0 +1,186 @@
...
+  }, []);
+
+  const handleChange = useCallback(
+    (evt, value) => {
+      setTagNamesOverrides(
+        new Map(selectedElements.map(({ id }) => [id, value]))
+      );
+
+      selectedElements.forEach(() => {
+        pushUpdate({ tagName: value });
```
miina: This might need `true` flag here as well to actually push the update. You can see it by having a clean history state and changing the tag -- there's no history entry created. The `pushUpdate` system needs updating (or removing), it's a bit out of date, there's an issue for it, too: #6690

packages/story-editor/src/components/panels/design/textAccessibility/test/textAccessibility.js
```diff
@@ -0,0 +1,114 @@
...
+
+  it('should call pushUpdate when dropdown has been updated', () => {
+    const { pushUpdate } = arrange([defaultElement]);
+    const input = screen.getByTestId('text-accessibility-dropdown');
+    fireEvent.click(input);
+    const option2 = screen.queryAllByRole('option')[1];
+    fireEvent.click(option2);
+    expect(pushUpdate).toHaveBeenCalledTimes(1);
+  });
+  it('should call pushUpdate three times when multiple items have been updated', () => {
```
miina, line 95: Very nitty nit: Empty line missing before the test 😅

packages/output/src/utils/getTextElementTagNames.js
```diff
@@ -37,9 +37,17 @@ function getTextElementTagNames(elements) {
...
        * @param {number} element.fontSize Font size.
        * @param {string} element.content Text content.
+       * @param {string} element.tagName Existing tagName
        * @return {Map<string, string>} Tag names map.
        */
-      (tagNamesMap, { id, fontSize, content }) => {
+      (tagNamesMap, { id, fontSize, content, tagName }) => {
+        // if a tag already exists,
+        // utilize the one already part of the element
+        if (tagName && tagName !== 'auto') {
```
miina, line 46: Would it make sense to create a constant for `auto`?

miina (approved):
Works well!

One finding when testing -- it seems that something changes visually in the editor as well for text at some point -- not even sure what it's related to since this PR doesn't really change the display.

gif:
headings

What I did:
1. Add a Title 1 and a Paragraph
2. Select both and set Title 3
3. Save the story
4. Click Undo/Redo to see the text visually changing

packages/output/src/utils/getTextElementTagNames.js
swissspidy: Seems a bit overkill to me right now. Maybe when we need it in more places, e.g. when tackling #11402

swissspidy:
> One finding when testing -- it seems that something changes visually in the editor as well for text at some point -- not even sure what it's related to since this PR doesn't really change the display.

Hmm not seeing that on my end in Chrome:

https://user-images.githubusercontent.com/841956/179708904-75e40671-1bfb-4fba-91e0-f58df3d1d448.mov

miina:
> > One finding when testing -- it seems that something changes visually in the editor as well for text at some point -- not even sure what it's related to since this PR doesn't really change the display.
> 
> Hmm not seeing that on my end in Chrome:
> 
>  Screen.Recording.2022-07-19.at.10.49.01.mov

Did you try with these steps, too?

1. Add a Title 1 and a Paragraph
2. Select both and set Title 3
3. Save the story
4. Click Undo/Redo to see the text visually changing

I'm able to reproduce it with these steps but not reliably. Suspecting it has to do with the fonts loading and not with this PR anyway, we can probably ignore for now.

merapi:
> Did you try with these steps, too?

I can't repro.

PR #11893, opened 2022-07-11
Fix auto-setting rotated background

### LINKED ISSUE #10519: Layer selection incorrectly detects rotated elements under the cursor

miina:
## Context
Previously, rotation was not properly considered when auto-setting background, causing setting background even if the media was only partially covering the background area.

## Summary
This PR re-uses the logic implemented in #11163 to calculate properly if the element is truly covering the background area fully before setting it as a background.

## User-facing changes
Background is not set when there's a part that's not covered:

autobg

## Testing Instructions

1. Add a media and rotate it
2. Now resize it so that it would almost cover the background area but not fully, e.g. leave one corner uncovered
3. Verify the media is not auto-set as background
4. Now resize it to cover fully and verify it's now auto-set.

## Checklist

---

Fixes #10519


barklund (approved):
Works just like it says! And excellent job of generalizing the relevant bits of code here!

miina:
Seeing a test fail, doesn't fail locally: https://github.com/GoogleForCreators/web-stories-wp/runs/7393124649?check_suite_focus=true#step:7:638

Not sure if this is a known flaky test or for some reason just doesn't fail locally, seems unrelated test to the changes as well 🤷

PR #11905, opened 2022-07-11
Headings: Suggested edits to TextAccessibilityPanel

swissspidy:
This is a companion PR for my review on #11888

### DIFF
--- a/packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
+++ b/packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
@@ -18,7 +18,7 @@
  * External dependencies
  */
 import PropTypes from 'prop-types';
-import { __, TranslateWithMarkup } from '@googleforcreators/i18n';
+import { __, sprintf, TranslateWithMarkup } from '@googleforcreators/i18n';
 import {
   DropDown,
   Text,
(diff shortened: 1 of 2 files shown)


packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
```diff
@@ -98,11 +84,49 @@ function TextAccessibilityPanel({ selectedElements, pushUpdate }) {
...
+      value: 'h2',
+    },
+    {
+      label: HEADING_LEVELS.h3,
+      value: 'h3',
+    },
+    {
+      label: HEADING_LEVELS.p,
+      value: 'p',
+    },
```
barklund, line 127: This could just be:

```js
    ...Object.entries(HEADING_LEVELS).map(([value, label]) => ({ value, label })),
```

Not the easiest to read, but it would work and would not require updates if we were ever to add more element types 🤷‍♂️ .

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
```diff
@@ -98,11 +84,49 @@ function TextAccessibilityPanel({ selectedElements, pushUpdate }) {
...
+  const selectedValue = 'auto' === tagName ? 'auto' : currentValue;
+
+  const options = [
+    {
+      label:
+        'auto' === selectedValue
+          ? sprintf(
+              /* translators: %s: heading level. */
+              __('Automatic (%s)', 'web-stories'),
+              HEADING_LEVELS[currentValue]
```
barklund, line 107: Uh, this is nice. I like that!

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
```diff
@@ -98,11 +84,49 @@ function TextAccessibilityPanel({ selectedElements, pushUpdate }) {
...
   // Check if tags match for selected item in dropdown
   // If they don't match, we want to show 'Mixed'
   // However, if they match, show the matching value
   const currentValue =
-    new Set(currentTags || []).size === 1 ? currentTags[0] : '((MULTIPLE))';
+    new Set(currentTags || []).size === 1 ? currentTags[0] : MULTIPLE_VALUE;
+
+  const isIndeterminate = MULTIPLE_VALUE === currentValue;
+
+  const tagName = getCommonValue(selectedElements, 'tagName', 'auto');
```
barklund, line 96: Doesn't this do the same thing as the unique set operation above? They seem to overlap a bit? Or is the `currentValue` based on the automatically calculated tags and `tagName` is based on the predefined tag stored on the element?

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
```diff
@@ -32,61 +32,47 @@ import { useState, useEffect } from '@googleforcreators/react';
...
+  h2: __('Heading 2', 'web-stories'),
+  h3: __('Heading 3', 'web-stories'),
+  p: __('Paragraph', 'web-stories'),
+};
 
 function TextAccessibilityPanel({ selectedElements, pushUpdate }) {
   // Feature flagging for Semantic Headings
   const showSemanticHeadings = useFeature('showSemanticHeadings');
   const [selectedTextElements, setSelectedTextElements] = useState([]);
   const [currentTags, setCurrentTags] = useState([]);
```
barklund, line 52: I don't 100% understand why these local states are necessary at all looking closer into this component? What state is required, that isn't already stored in the story reducer?

barklund (commented):
A few nits

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
swissspidy: > Or is the `currentValue` based on the automatically calculated tags and `tagName` is based on the predefined tag stored on the element?

yeah

packages/story-editor/src/components/panels/design/textAccessibility/textAccessibility.js
swissspidy: @barklund Please note that this PR just adds a few suggestions to the original PR. I recommend reviewing #11888 instead

sblinde: > 

Indeed, this is only an addition to that PR, but I can briefly explain:
Local state is needed because tags and selected elements, by default, are not combined in the story reducer. We also need to separately know how many tags they are, and whether they match, to support multiple text items being selected and proper display text (if a H1 and an H2 are selected, we say 'mixed' for the heading levels. if two H2s are selected, we show that both elements are 'Heading 2'). For that reason, both selected elements with their tags, as well as the original list of tags, are added to component state.

sblinde:
This branch was merged into https://github.com/GoogleForCreators/web-stories-wp/pull/11888 to continue to work on feedback

PR #11908, opened 2022-07-12
Adding trimmer functionality to media recording

### LINKED ISSUE #11685: Media Recording: Add Video Trimming
## Feature Description

**[Figma](https://www.figma.com/file/ZrO763yvMWJjYeAPsRnNfW/_Editor---Sprint-Feature-Level-Designs_2022?node-id=661%3A24394)**

Screenshot 2022-06-13 at 18 15 58

For v2 of the media recording feature we'd want to include video trimming on canvas, so that users can trim the video before it is uploaded to the site.

Involves some refactoring of the video trimming UI and calling `trimVideo()` before inserting the video element to the page.

### LINKED ISSUE #11922: Play button is unavailable to users when trimming a gif
## Bug Description

[Jonny] If you switch to gif and try to trim the video, there is no play icon. This can make it hard to trim the video, as you can not pause a looping video. 

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

barklund:
## Context

This refactors both the existing trimmer and the media recording system to be able to merge the two concepts.

Due to the necessary extra provider, the media recording has been split out from the media recording layer. This will probably cause conflicts with the audio work.

## To-do

- [x] Re-trimming while still in recording mode is a bit broken
- [x] It is still buggy - sometimes the reference to `videoNode` is stale and things fail.
- [x] After trimming, the video still plays in the full duration in the preview, but the trim data has been recorded
- [ ] The preview on-canvas while the video is being trimmed and uploaded is also not trimmed, but shows the full duration
- [ ] The generated thumbnail is taken from the start of the full video, not the start of the trimmed video

## User-facing changes

_TBD_

## Testing Instructions

1. Start a media recording
2. Record yourself
3. Click the scissors to trim
4. Play with trimming, cancelling, re-trimming, inserting, etc.

Please note known bugs above.

## Checklist

---

Fixes #11685
Fixes #11922


merapi:
Might be connected to that stale `videoNode`, if I trim before the preview playback finishes I get this:

image

image

image

If I wait it seems ok?

Play(/pause) in preview (without touching trimming) doesn't work.

barklund:
The remaining issues are both very straight-forward to understand and test around, but also pretty annoying to fix. Well, the first one with the preview in the media recorder is fairly easy, I can copy the code from the trimmer to start and stop the video at the right frame offsets, but manipulating the video in the preview on canvas seems a lot harder (and even more code to duplicate), and the frame snapshot too.

I'm considering doing only the first issue in this version and filing the other two tickets as followup. The latter at least seem better suited for WP pod, as this thing is in their domain.

But please, @merapi and @miina, check it out as-is both the result and the code and give me a review. I think this is solid for bug bash at least.

Finally, we could also file a follow-up bug to deal with the duration only coming after the video has played through once making trimming not work until then. We could use the duration (in seconds only though) stored inside the recorder provider until a more precise time is known. We could probably also extract the exact time from the blob somewhere and use that value. Or if we can't do either of those, we could at least disable the trim button until it's available after the first play-through.

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
```diff
@@ -168,80 +101,47 @@ function MediaRecordingLayer() {
...
-                </VideoWrapper>
-              )}
-            </Wrapper>
-            <ProgressBar />
-            <Countdown />
+            <MediaRecording />
           </DisplayPageArea>
-          <Footer captureImage={debouncedCaptureImage} videoRef={videoRef} />
+          <StyledFooter showOverflow>
+            {isTrimming ? <VideoTrimmer /> : <Footer />}
```
miina, line 142: Maybe we should display a loading indicator here until we have the duration when the trim icon is clicked?

packages/story-editor/src/components/videoTrim/useVideoNode.js
```diff
@@ -89,18 +105,42 @@ function useVideoNode(videoData) {
...
     }
-    videoNode.addEventListener('timeupdate', onTimeUpdate);
-    videoNode.addEventListener('loadedmetadata', onLoadedMetadata);
+
+    // We might already have metadata, if so invoke the event handler directly.
+    // Note that on first playback before completion, duration is Infinite, so
+    // so we need to wait until it isn't anymore.
+    if (
+      !isNaN(videoNode.duration) &&
+      isFinite(videoNode.duration) &&
```
swissspidy: Just realized that `isFinite` also checks for `NaN` so the first check is redundant.

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
barklund: Yeah, we should definitely do something here. I have some ideas for how to improve, but it's a based on technical limitations so hard to solve well in UX.

packages/story-editor/src/components/mediaRecording/footer.js
```diff
@@ -231,17 +218,21 @@ function Footer({ captureImage, videoRef }) {
...
       args.additionalData.isGif = isGif;
 
-      // Get length from the video blob on screen and fall back
-      // to the duration in state if needed since it's not super reliable.
-      const { length, lengthFormatted } = getVideoLength(videoRef.current);
-      args.resource.length = length || duration;
-      args.resource.lengthFormatted = length
-        ? lengthFormatted
-        : getVideoLengthDisplay(duration);
+      args.resource.length = duration;
```
spacedmonkey, line 237: We should round this to a whole number.

packages/story-editor/src/components/mediaRecording/footer.js
```diff
@@ -231,17 +218,21 @@ function Footer({ captureImage, videoRef }) {
...
+      args.resource.lengthFormatted = getVideoLengthDisplay(duration);
```
spacedmonkey, line 238: See https://github.com/GoogleForCreators/web-stories-wp/pull/11908/files#r922077299

packages/story-editor/src/components/videoTrim/useVideoNode.js
```diff
@@ -68,17 +68,33 @@ function useVideoNode(videoData) {
...
+
   useEffect(() => {
     if (!videoNode || !videoData) {
       return undefined;
     }
 
     function restart(at) {
       videoNode.currentTime = at / 1000;
-      videoNode.play();
+      videoNode.play().catch(() => {});
```
spacedmonkey, line 90: 👍

packages/story-editor/src/components/videoTrim/recordingProvider.js
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright 2020 Google LLC
```
spacedmonkey:
```suggestion
 * Copyright 2022 Google LLC
```

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
```diff
@@ -168,80 +101,47 @@ function MediaRecordingLayer() {
...
+            width: FULLBLEED_RATIO * 480,
+            height: 480,
+            scale: 1,
+            focalX: 50,
+            focalY: 50,
+            flip: {},
+          },
+          resource: {
+            src: mediaBlobUrl,
+            length: duration,
```
spacedmonkey, line 120: Length formated?

packages/story-editor/src/components/mediaRecording/footer.js
barklund: This is rounded, AFAIK. The original value here comes from the recording counter (which just counts seconds), and [when trimming, the value is updated to the `ceil()` of the trim interval](https://github.com/GoogleForCreators/web-stories-wp/pull/11908/files#diff-cc3a2d938726f362ee058f86a5ce158b8b62d8fa84a3d5abacc96460f980c63aR29-R31).

packages/story-editor/src/components/canvas/mediaRecordingLayer.js
barklund: This is only to fool the `generateVideoStrip` function into thinking, that it's a "real" resource. [That function only extracts these two values](https://github.com/GoogleForCreators/web-stories-wp/blob/e00e50da7a02d956b993f4038ca1c15d46eda3ef/packages/media/src/generateVideoStrip.js#L43).

packages/story-editor/src/components/videoTrim/recordingProvider.js
barklund: Fixed in da2b3d2664

spacedmonkey:
Sharing comment for visibility. https://github.com/GoogleForCreators/web-stories-wp/pull/11887#discussion_r918355684

swissspidy:
> Sharing comment for visibility. [#11887 (comment)](https://github.com/GoogleForCreators/web-stories-wp/pull/11887#discussion_r918355684)

This PR only covers trimming for videos, not for audio files.

In the future we may consider adding an audio trimmer though, but that requires UX and prioritization.

packages/story-editor/src/components/mediaRecording/footer.js
```diff
@@ -244,17 +231,20 @@ function Footer({ captureImage, videoRef }) {
...
       args.additionalData.isGif = isGif;
 
-      // Get length from the video blob on screen and fall back
-      // to the duration in state if needed since it's not super reliable.
-      const { length, lengthFormatted } = getVideoLength(videoRef.current);
-      args.resource.length = length || duration;
-      args.resource.lengthFormatted = length
-        ? lengthFormatted
-        : getVideoLengthDisplay(duration);
+      args.resource.length = duration;
```
swissspidy, line 237: @barklund This originally used `getVideoLength(videoRef.current)` whenever possible because it was more accurate. Is that not possible anymore?

packages/story-editor/src/components/mediaRecording/mediaRecording.js
```diff
@@ -0,0 +1,164 @@
...
+  const hasPlayButton = !isGif || isTrimming;
+
+  return (
+    <>
+      <Wrapper>
+        {!isImageCapture && (
+          <>
+            {mediaBlobUrl && hasVideo && (
+              <VideoMode value={!isGif} onChange={onToggleVideoMode} />
+            )}
```
swissspidy: We should probably not render the video/gif toggle during trimming, as it's quite confusing.

During trimming the video is looping anyway and the play button is also displayed.

The toggle can still be used after trimming.

packages/story-editor/src/components/mediaRecording/footer.js
barklund: Only when there's no trim data. When there is, the video element length doesn't actually reflect the final video length (the trim offsets do). But I can revert so it will use the former if there's no trim.

packages/story-editor/src/components/mediaRecording/mediaRecording.js
barklund: Makes sense, I'll remove it.

felipebochehin87:
Trimming works for gifs and video;
Play/Stop buttons displayed for both videos and gifs;
When I trim/re-trim a video or a gif and preview the story, only the trimmed part of it is reproduced while on preview story screen;
Same as above for publishing story.


https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/post.php?post=19102&action=edit#page=%25224bf235a6-a904-4a4c-8e36-b5c6a039ecaa%2522

swissspidy:
@barklund Looks good, but there's a merge conflict

swissspidy:
@spacedmonkey your merge causes lint errors, can you fix those?

spacedmonkey:
> @spacedmonkey your merge causes lint errors, can you fix those?

Done

spacedmonkey:
https://user-images.githubusercontent.com/237508/180219796-9093f3bf-50dd-4619-a0f3-cd6203e6c9ee.mov

On a 40 second video, it takes nearly 40 seconds for the trim option to show up.

packages/story-editor/src/components/mediaRecording/playPauseButton.js
```diff
@@ -93,12 +93,15 @@ function PlayPauseButton({ videoRef }) {
...
     const onVideoPause = () => setIsPlaying(false);
+    const onTimeUpdate = ({ target: { paused } }) => setIsPlaying(!paused);
 
     videoNode.addEventListener('play', onVideoPlay);
     videoNode.addEventListener('pause', onVideoPause);
+    videoNode.addEventListener('timeupdate', onTimeUpdate);
     return () => {
       videoNode.removeEventListener('play', onVideoPlay);
       videoNode.removeEventListener('pause', onVideoPause);
+      videoNode.removeEventListener('paustimeupdatee', onTimeUpdate);
```
spacedmonkey:
```suggestion
      videoNode.removeEventListener('timeupdate', onTimeUpdate);
```

Doesn't this be like this?

barklund: Yeah, my bad, fixed in af91a08.

merapi (approved):
👍🏻 

We will deal with the remaining items here https://github.com/GoogleForCreators/web-stories-wp/issues/11988

PR #11911, opened 2022-07-13
Canvas: Fix height and width for media element's BlurHash container

spacedmonkey:
## Summary

Fix issue blurhash being too big on videos smaller than 128px by 128px. 

## Testing Instructions

1.

## Checklist

---

Fixes #

### DIFF
--- a/packages/story-editor/src/components/canvas/renderResourcePlaceholder.js
+++ b/packages/story-editor/src/components/canvas/renderResourcePlaceholder.js
@@ -24,10 +24,8 @@ const placeholderStyles = css`
   position: absolute !important;
   top: 0;
   left: 0;
-  min-height: 100%;
-  min-width: 100%;
-  max-height: 100%;
-  max-width: 100%;
+  height: 100%;
+  width: 100%;
 `;
 
 const BlurhashContainer = styled(Blurhash)`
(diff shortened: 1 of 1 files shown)


packages/story-editor/src/components/canvas/renderResourcePlaceholder.js
```diff
@@ -26,6 +26,8 @@ const placeholderStyles = css`
   left: 0;
   min-height: 100%;
   min-width: 100%;
+  max-height: 100%;
+  max-width: 100%;
```
miina: Would setting `width; 100%;` and `height: 100%` also work? Instead of using max and min both.

merapi: min and max could be removed if we do:
```jsx
return <BlurhashContainer height="100%" width="100%" hash={blurHash} punch={1} />
```
otherwise, we cannot override the width/height from Blurhash without `!important`.

But because `placeholderStyles` is also used in the `BaseColorContainer` we need to leave something here too (width/height=100%?).

spacedmonkey: It is all the same to me. Changed to use height / width 100%.

miina (approved):
LGTM!

merapi (approved):
LG 👍🏻 

image

PR #11913, opened 2022-07-14
Shopping: Support drag & drop from library

### LINKED ISSUE #11708: Shopping: Support drag & drop from library
## Feature Description

For consistency it would be nice to eventually support dragging products from the library to the canvas.

timarney:
## Context

Add support for dragging products from the library to the canvas.

## Summary

Adds draggable product image

## To-do

Handle dragging product that already exists on the canvas

## Testing Instructions

1. Visit shopping tab
2. Drag product image from shopping panel to the canvas
3. The image upon drop should insert the product that was selected
4. Products without images or that are already on the page should not be draggable (in other words: if the "Plus" icon button is disabled, so should be dragging)

No

No

No

## Checklist

---

Fixes #11708


packages/story-editor/src/components/library/panes/shopping/productList.js
```diff
@@ -80,7 +80,7 @@ function ProductList({
               product={product}
               canAddMore={canAddMore}
               isOnPage={onPageProducts.some(
-                (item) => item.product.productId === product?.productId
+                (item) => item.product?.productId === product?.productId
```
swissspidy: Why was this needed?

packages/story-editor/src/components/library/panes/shopping/productImage.js
```diff
@@ -57,4 +75,31 @@ ProductImage.propTypes = {
   product: PropTypes.object.isRequired,
 };
 
-export default ProductImage;
+function DraggableProductImage({ product, isOnPage }) {
+  const src = product?.productImages[0]?.url || '';
```
swissspidy, line 80: A product without an image should not be draggable.

a) because we do not want to allow products without images on a page and b) it looks broken

Screenshot 2022-07-15 at 14 00 47

packages/story-editor/src/components/library/panes/shared/libraryMoveable.js
```diff
@@ -261,6 +261,12 @@ function LibraryMoveable({
           offsetX: editorToDataX(x - pageX, pageSize.width),
           offsetY: editorToDataY(y - pageY, pageSize.height),
         });
+      } else if (type === 'product') {
+        insertElement(type, {
+          ...elementProps,
+          width: 25,
+          height: 25,
```
swissspidy: Also need to add `x` and `y` here so that the product is inserted in the spot where you drop it.

swissspidy (changes requested):

packages/story-editor/src/components/library/panes/shopping/productList.js
timarney: Will re-check --- was throwing an error but possibly that was prior code.

timarney: Removed --- issue with earlier code.

packages/story-editor/src/components/library/panes/shared/libraryMoveable.js
timarney: Will fix

packages/story-editor/src/components/library/panes/shared/libraryMoveable.js
```diff
@@ -261,6 +261,14 @@ function LibraryMoveable({
...
```
miina: I'm not very familiar with products but do these always have fixed 25 width and height? Perhaps there's a constant somewhere for the fixed size we can use instead?

packages/story-editor/src/components/library/panes/shopping/productImage.js
```diff
@@ -32,11 +37,24 @@ const StyledImgPlaceHolder = styled.div`
...
+const ProductImageWrap = styled.div`
+  position: relative;
+`;
+
 const StyledImage = styled.img`
   ${imgPlaceholder}
   object-fit: cover;
 `;
 
+const StyledImageClone = styled.img`
```
miina, line 50: Maybe it would make sense to instead of showing the image, showing the actual CTA button that's displayed on the page?
This would:
- Display the user what will actually be added on the page
- Fix snapping while dragging (the other elements snap to the dragged image, however, that's not actually added so it isn't very useful)
- Fix being able to drag a product outside of the page. Generally, when an element is dragged outside of the page, it is deleted, or not added - in case of dragging a new element from the library. However, this is calculated based on the element that's being dragged. If we're displaying the image instead of the button that's actually added to the page, it's possible to drag a product so that it's created a bit outside of the page, for example:
dragging

timarney: @swissspidy let me know how you think we should proceed based on these comments.

packages/story-editor/src/components/library/panes/shared/libraryMoveable.js
timarney: Matches the width and height used here https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/story-editor/src/components/library/panes/shopping/shoppingPane.js#L202

miina: Would be nice to have a constant for it if it's used in more than one file.

packages/story-editor/src/components/library/panes/shopping/productImage.js
miina: (One alternative would be displaying the image but 25x25 instead)

timarney: Updated to use the 25x25 image for now.

packages/story-editor/src/components/library/panes/shared/libraryMoveable.js
timarney: Sounds good - I added a constants file ---

https://github.com/GoogleForCreators/web-stories-wp/pull/11913/commits/53c811d9a457f8a73f2b3b5ab35e7a3b5a704257#diff-63211c3c2db20e5f634ea3022912ae3a8d30741d1e517604240a647055058b4b

spacedmonkey:
Screenshot 2022-07-20 at 15 14 32
Seeing this error when I drag into canvas.

packages/story-editor/src/components/library/panes/shopping/productImage.js
swissspidy: Using a 25x25 image should be fine for now. We can revisit later if needed

timarney:
Added a conditional to fix the `style` issue seems to only happen using local WP.

https://github.com/GoogleForCreators/web-stories-wp/pull/11913/commits/75ba201890485481bc5a04ef7ccf83699f93b8fe

spacedmonkey (approved):
Seems to work well.

packages/story-editor/src/components/library/panes/shopping/productImage.js
```diff
@@ -57,4 +76,31 @@ ProductImage.propTypes = {
...
+          cloneElement={StyledImageClone}
+          cloneProps={{
+            loading: 'lazy',
+            decoding: 'async',
+            crossOrigin: 'anonymous',
+            src,
+          }}
+        />
+      )}
+    </ProductImageWrap>
```
swissspidy: Products should not be draggable if there are already 6 products on the page

Just like the "Add" button is disabled when that happens.

In other words: When I can't add a product by clicking, I should also not be able to add one by dragging.

swissspidy (changes requested):
Products are still draggable even when the insertion button is disabled because the limit of 6 products has been reached

Screenshot 2022-07-21 at 13 13 43

packages/story-editor/src/components/library/panes/shopping/productImage.js
timarney: Added the canAddMore check

spacedmonkey:
Should the tooltip appear on when hover over the image when you try and drag it when you are not allowed?
Screenshot 2022-07-21 at 12 56 19

swissspidy:
We can keep the tooltip as-is, only on the button.

spacedmonkey (approved):

packages/story-editor/src/components/library/panes/shopping/product.js
```diff
@@ -46,7 +46,11 @@ function Product({ product, onClick, onFocus, isOnPage, canAddMore }) {
         onClick={onClick}
         onFocus={onFocus}
       />
-      <ProductImage product={product} />
+      <ProductImage
+        product={product}
+        isOnPage={isOnPage}
+        canAddMore={canAddMore}
```
swissspidy: Nit: it would be more readable to just have one prop, e.g. `isEnabled` or `canDrag` or something

swissspidy (approved):

miina (approved):

miina:
Would be nice to add a Karma test as well for it if possible.
(It'll need a slight delay for Moveable/clone to load, for example:) https://github.com/GoogleForCreators/web-stories-wp/blob/b98f74e9820633b41d8fc03fe1f06bdbf7ab43dd/packages/story-editor/src/components/library/karma/shapes/shapes.karma.js#L66

timarney:
Test was added here : https://github.com/GoogleForCreators/web-stories-wp/pull/11913/files#diff-fcd139b38af8fc60a75b51efebd73d31b67f0d19e1c799c037432b70f80ebcc9R244

PR #11924, opened 2022-07-15
Add grouping to templates

### LINKED ISSUE #11697: Templates: Make use of layer naming/grouping/locking
## Feature Description

With the new layers naming/grouping/locking features there's an opportunity to improve UX when using them.

Here's an example of the "All About Cars" template, where grouping layers would make it way more user friendly:

Screenshot 2022-06-14 at 00 46 49

At first glance it seems like every template page with more than 10 layers on it is a good candidate for manual review and potentially grouping the layers.

Then, the groups could have more meaningful names to make them easier to distinguish.

mwritter:
## Context

Add grouping, group names, and layer locking to templates
## Summary

Adds grouping, group names, and layer locking to templates
## Testing Instructions

1. Go to Explore Templates
2. Click though the templates
- Notice most templates (that seem to need it, based on amount of layers - about 10 layers or more) now utilize layer grouping and locking.

No

No

No
## Checklist

---

Fixes #11697


packages/templates/src/raw/fresh-and-bright/template.json
```diff
@@ -1623,7 +1683,29 @@
...
-      "pageTemplateType": "editorial"
+      "pageTemplateType": "editorial",
+      "groups": {
+        "0f035801-d8bb-4a82-ae00-0cba807b6ae6": {
+          "name": "Action",
+          "isLocked": false,
+          "isCollapsed": true
+        },
+        "235f4576-d429-41e8-9823-543901200926": {
+          "name": "Action",
```
miina, line 1694: I believe Pea Pod was mainly working on templates and not sure if there's anyone to ask but perhaps @swissspidy might know:
Looks like in this template there are various cases of buttons created by a text and a rectangle behind it. Any idea if there's a reason why a text with background color isn't used here instead to have better centering and just one element? 🤔
 It's grouped in this PR, however, having a text with background might make more sense?
Screenshot 2022-07-25 at 18 45 40

swissspidy: The templates were originally created by another agency.

It's possible that they were not aware of the background color feature or that it didn't fit the needs for some reason.

Agreed that having a text with background makes more sense in cases like this 👍

miina: Added a follow-up issue since it'd take quite some time to fix it everywhere in this PR: #12032

miina:
@merapi @barklund I've reviewed the changes of this PR and added some fixes and additions, this should be ready for review now.

merapi (approved):
After spending 10 minutes reviewing 27k lines of mostly auto-generated code I have enough 😜 LG 👍🏻

PR #11925, opened 2022-07-18
Settings: Add checkbox for data removal upon uninstall

### LINKED ISSUE #8453: Settings: Add checkbox for data removal upon uninstall
## Feature Description

Continuing discussion at https://github.com/google/web-stories-wp/pull/2534#discussion_r441679994

In #2534 we've added an `uninstall.php` file to remove any plugin data when uninstalling the plugin in WordPress.

That data removal is off by default though and needs to be enabled by filtering `web_stories_erase_data_on_uninstall`.

The idea was to turn this into an actual UI option on the settings page.

## Additional Context

Interestingly, plugins like WooCommerce or Jetpack don't have a UI option for this and instead require setting a constant in `wp-config.php` to enable data removal.

This begs the question whether we really want to add yet another option to the settings page or keep it a developer-only option (perhaps also using a constant instead of a filter).

---

timarney:
## Context

Data removal is off by default though and needs to be enabled by filtering web_stories_erase_data_on_uninstall we need to add a UI to allow users to toggle this option.

## Summary

Adds checkbox to the story settings page.

Screen Shot 2022-07-22 at 7 02 46 AM

## To-do

Update text as needed.

## Testing Instructions

1. Visit settings
2. Toggle data removal setting on / off

## Checklist

---

Fixes #8453


includes/Admin/Admin.php
```diff
@@ -76,6 +76,7 @@ public function register(): void {
 		add_filter( 'default_content', [ $this, 'prefill_post_content' ], 10, 2 );
 		add_filter( 'default_title', [ $this, 'prefill_post_title' ] );
 		add_filter( 'display_media_states', [ $this, 'media_states' ], 10, 2 );
+		add_filter( 'web_stories_erase_data_on_uninstall', [ $this, 'data_removal' ], 10, 2 );
```
swissspidy: Instead of filtering `web_stories_erase_data_on_uninstall`, we should just call `get_option( Settings:: SETTING_NAME_DATA_REMOVAL )` in `uninstall.php` directly.

This is because the `Admin` class here is only loaded in the admin, but a plugin can also be uninstall via WP-CLI for example.

Note: Using `get_option` should be fine, but could theoretically also use the Services injector, though seems a bit overkill. Plus not sure if it's properly available in CLI context as well, so needs some testing. Hence the suggestion to just use `get_option`.

cc @spacedmonkey

uninstall.php
```diff
@@ -35,7 +35,8 @@
  *
  * @param bool $erase Whether to erase data. Default false.
  */
-$erase = (bool) apply_filters( 'web_stories_erase_data_on_uninstall', false );
+
+$erase = (bool) get_option( \Google\Web_Stories\Settings::SETTING_NAME_DATA_REMOVAL );
```
swissspidy: We still want to wrap this in a filter.

```php
$erase = (bool) get_option( \Google\Web_Stories\Settings::SETTING_NAME_DATA_REMOVAL );

/**
 * Filters whether data should be erased when uninstalling the plugin.
 *
 * @since 1.23.0
 *
 * @param bool $erase Whether to erase data. Default false.
 */
$erase = (bool) apply_filters( 'web_stories_erase_data_on_uninstall', $erase );
```

swissspidy: @timarney Now the filter docblock is duplicated in this file. Please check again.

And my bad for changing the version number. It should of course stay `@since 1.0.0`

timarney: Right -- missed that ... updated now.

packages/wp-dashboard/src/components/editorSettings/dataRemoval/index.js
```diff
@@ -0,0 +1,75 @@
...
+}) {
+  const onChange = useCallback(
+    () => updateSettings({ dataRemoval: !isEnabled }),
+    [updateSettings, isEnabled]
+  );
+
+  return (
+    <SettingForm>
+      <div>
+        <SettingHeading>{__('Data removal', 'web-stories')}</SettingHeading>
```
swissspidy:
```suggestion
        <SettingHeading>{__('Plugin Uninstall', 'web-stories')}</SettingHeading>
```

packages/wp-dashboard/src/components/editorSettings/dataRemoval/index.js
```diff
@@ -0,0 +1,75 @@
...
+            onChange={onChange}
+            checked={Boolean(isEnabled)}
+          />
+          <CheckboxLabelText
+            size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL}
+            aria-checked={Boolean(isEnabled)}
+            forwardedAs="span"
+          >
+            {__(
+              'Remove data when uninstalling the Web Story plugin.',
```
swissspidy:
```suggestion
              'Remove all data including stories and saved templates when uninstalling the Web Stories plugin.',
```

uninstall.php
swissspidy: It's _still_ duplicated

Screenshot 2022-07-19 at 12 48 03

swissspidy (changes requested):

uninstall.php
swissspidy: The docblock needs to be on the line directly preceding the `apply_filters` call

See https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#4-hooks-actions-and-filters

spacedmonkey:
I tried testing this and doesnt seem to work. On local, I used WP CLI to uninstall the plugin. 
Screenshot 2022-07-20 at 12 47 45

spacedmonkey:
Screenshot 2022-07-20 at 12 55 53

swissspidy:
This might work (untested!):

```php
// Autoloader for dependencies.
if ( file_exists( WEBSTORIES_PLUGIN_DIR_PATH . '/third-party/vendor/scoper-autoload.php' ) ) {
	require WEBSTORIES_PLUGIN_DIR_PATH . '/third-party/vendor/scoper-autoload.php';
}

// Autoloader for plugin itself.
if ( file_exists( WEBSTORIES_PLUGIN_DIR_PATH . '/includes/vendor/autoload.php' ) ) {
	require WEBSTORIES_PLUGIN_DIR_PATH . '/includes/vendor/autoload.php';
}

// Bootstrap the plugin by registering all services.
// Bails early if services are already registered.
// TODO: Needed?
PluginFactory::create()->register();
```

spacedmonkey:
A simplier work around was to add the plugin file, like this.

```
require_once 'web-stories.php';
```

swissspidy:
> A simplier work around was to add the plugin file, like this.
> 
> ```
> require_once 'web-stories.php';
> ```

Won't that cause constants to be defined twice? Also unnecessarily runs compatibility checks etc. So I would not wanna do this.

uninstall.php
```diff
@@ -28,14 +28,17 @@
 	return;
 }
 
+
```
spacedmonkey, line 62:
```suggestion

if ( ! defined( 'WEBSTORIES_PLUGIN_DIR_PATH' ) ) {
	define( 'WEBSTORIES_PLUGIN_DIR_PATH', __DIR__ );
}

// Autoloader for dependencies.
if ( file_exists( WEBSTORIES_PLUGIN_DIR_PATH . '/third-party/vendor/scoper-autoload.php' ) ) {
	require WEBSTORIES_PLUGIN_DIR_PATH . '/third-party/vendor/scoper-autoload.php';
}

// Autoloader for plugin itself.
if ( file_exists( WEBSTORIES_PLUGIN_DIR_PATH . '/includes/vendor/autoload.php' ) ) {
	require WEBSTORIES_PLUGIN_DIR_PATH . '/includes/vendor/autoload.php';
}


```

In my testing this fixed the issue of uninstalling the plugin in WP CLI.

spacedmonkey (dismissed):
Seems to work well in my testing.

swissspidy (dismissed):

swissspidy:
@felipebochehin87 For QA on this PR, please do not do it on the QA site. Instead, create a new local site (e.g. in localwp.com), install the production build from https://github.com/GoogleForCreators/web-stories-wp/pull/11925#issuecomment-1188887081, create some stories, and see if the stories are properly deleted on uninstall. So when you install the plugin again there should be no stories anymore.

Hope that makes sense.

felipebochehin87:
It does @swissspidy thanks a lot for the info!

felipebochehin87:
@swissspidy @spacedmonkey @timarney please let me know if I am doing something wrong:

1. Create a local WP website
2. Install and activate the prod plugin
3. Create some stories (I already had some)
4. Deactivate and delete the prod plugin
5. Install and activate the dev plugin

I follow this steps, but when I install again the plugin, I see the old stories.

https://images.zenhubusercontent.com/235435637/435e706a-73f9-4cc2-a4cb-3a701f921ea5/2022_07_21_19_56_04.mp4

timarney:
@felipebochehin87  --- I added a screenshot in the summary above.

felipebochehin87:
Thanks @timarney @swissspidy 

I still see some weird things - steps 3 and 4.

1. Installed the prod build

2. Checked the checkbox

3. Deleted the plugin - I click on deactivate and on delete, then "Deleting..." remains on screen. The plugin is removed from the left menu, but not from the list of plugins, and the label remains displaying "Deleting...". If I refresh the page it goes back to "Delete" and when I click on Delete, then Web Stories is removed from the list of plugins

4. Installed the dev plugin - stories are still there

swissspidy:
Indeed I see that too.

Screenshot 2022-07-22 at 14 40 23


@timarney Can you check this again please?

You might need to add a `PluginFactory::create()->register();` call after all like originally suggested, but not haven't tested.

swissspidy:
For reference, this is the error that happens:

```
Fatal error: Uncaught Google\Web_Stories\Exception\InvalidService: The service ID "injector" is not recognized and cannot be retrieved. in /Users/pascalb/Local Sites/qa11925/app/public/wp-content/plugins/web-stories/includes/Exception/InvalidService.php:75
Stack trace:
#0 /Users/pascalb/Local Sites/qa11925/app/public/wp-content/plugins/web-stories/includes/Infrastructure/ServiceContainer/SimpleServiceContainer.php(51): Google\Web_Stories\Exception\InvalidService::from_service_id('injector')
#1 /Users/pascalb/Local Sites/qa11925/app/public/wp-content/plugins/web-stories/includes/Services.php(102): Google\Web_Stories\Infrastructure\ServiceContainer\SimpleServiceContainer->get('injector')
#2 /Users/pascalb/Local Sites/qa11925/app/public/wp-content/plugins/web-stories/includes/uninstall.php(178): Google\Web_Stories\Services::get_injector()
#3 /Users/pascalb/Local Sites/qa11925/app/public/wp-content/plugins/web-stories/includes/uninstall.php(224): Google\Web_Stories\delete_terms()
#4 /Users/pascalb/Local Sites/qa11925/app/public/wp-content/plugins/web-stories/uninstall.php(84): Google\Web_Stories\delete_site()
#5 /Users/pascalb/Local Sites/qa11925/app/public/wp-admin/includes/plugin.php(1241): include_once('/Users/pascalb/...')
```

With the following change the plugin is initialized properly for uninstall and there are no longer any fatal errors:

```php
if ( ! defined( 'WP_UNINSTALL_PLUGIN' ) ) {
	return;
}

if ( ! defined( 'WEBSTORIES_VERSION' ) ) {
	define( 'WEBSTORIES_VERSION', '1' );
}

if ( ! defined( 'WEBSTORIES_DEV_MODE' ) ) {
	define( 'WEBSTORIES_DEV_MODE', false );
}

if ( ! defined( 'WEBSTORIES_PLUGIN_DIR_FILE' ) ) {
	define( 'WEBSTORIES_PLUGIN_FILE', __DIR__ . '/web-stories.php' );
}

if ( ! defined( 'WEBSTORIES_PLUGIN_DIR_PATH' ) ) {
	define( 'WEBSTORIES_PLUGIN_DIR_PATH', __DIR__ );
}

if ( ! defined( 'WEBSTORIES_PLUGIN_DIR_URL' ) ) {
	define( 'WEBSTORIES_PLUGIN_DIR_URL', plugin_dir_url( WEBSTORIES_PLUGIN_FILE ) );
}

// Autoloader for dependencies.
if ( file_exists( WEBSTORIES_PLUGIN_DIR_PATH . '/third-party/vendor/scoper-autoload.php' ) ) {
	require WEBSTORIES_PLUGIN_DIR_PATH . '/third-party/vendor/scoper-autoload.php';
}

// Autoloader for plugin itself.
if ( file_exists( WEBSTORIES_PLUGIN_DIR_PATH . '/includes/vendor/autoload.php' ) ) {
	require WEBSTORIES_PLUGIN_DIR_PATH . '/includes/vendor/autoload.php';
}

\Google\Web_Stories\PluginFactory::create()->register();
```

**That said**, it feels super overkill to me. We don't need the services and the injector just to get some taxonomy slugs in `delete_terms()`. It would be easier to just use class constants there.

spacedmonkey:
I didnt see these error locally

timarney:
Updated and tested with changes from https://github.com/GoogleForCreators/web-stories-wp/pull/11925#issuecomment-1192568091

swissspidy:
It works 🎉 🎉 

Only found two unrelated bugs with the uninstall code.

**First**, draft stories are not deleted. To fix this, we need to add `'post_status' => 'any'` here:

https://github.com/GoogleForCreators/web-stories-wp/blob/7109307f62821a0b3a166335718b9371ba8d21f7/includes/uninstall.php#L154-L162

**Second**, the `{$taxonomy}_children` option is not deleted for the `web_story_category` taxonomy. To fix this, we need to call `clean_taxonomy_cache()` here:

https://github.com/GoogleForCreators/web-stories-wp/blob/7109307f62821a0b3a166335718b9371ba8d21f7/includes/uninstall.php#L183-L187

Can do it in a loop, i.e.

```php
foreach ( $taxonomies as $taxonomy ) {
	clean_taxonomy_cache( $taxonomy );
}
```

Can you add this in this PR?

Then it should be good to go!

includes/uninstall.php
```diff
@@ -154,6 +154,7 @@ function delete_posts(): void {
 		[
 			'fields'           => 'ids',
 			'suppress_filters' => false,
+			'post_status' => 'any',
```
swissspidy: Needs proper arrow alignment.

Tip: run `npx husky install` to set up git pre-commit hooks that catch this kinda stuff before pushing.

timarney: I seem to run into this depending on which PHP version I'm running doesn't catch on PHP 8.x .... fine when using 7.x

swissspidy: I see. PHPCS PHP 8.x compatbility is not fully released yet unfortunately

swissspidy (approved):

swissspidy:
@felipebochehin87 Now it should work! 🤞

felipebochehin87:
1. Created a local WP website
2. Installed and activated the prod plugin
3. Created some stories (I already had some)
4. Deactivated and deleted the prod plugin
5. Installed and activated the dev plugin
6. Visited the Dashboard - 0 stories returned, redirected to the Templates screen

PR #11939, opened 2022-07-18
Fix background audio setting & mic-only device support

### LINKED ISSUE #11920: User cannot initiate Audio Self Recording if they don't have a webcam
## Bug Description

When someone wants to record an audio file on a computer that does not have a webcam, they get a message that says “Error: Could not initialize recording. Reason: No Camera Found.” [Seen on Windows 11, Chrome, on two machines that do not have web cameras (desktop, laptop w/o webcam)] - Reporter: Sam
Of note: When you do get this error, the buttons below that say “Record Video” and “Take a photo” look to still be enabled, but nothing happens when you click on them.

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

### LINKED ISSUE #11921: Media Recording: Audio is incorrectly inserted on story instead of page
## Bug Description

Inserting an Audio recording defaults to inserting that audio as the background audio for the entire story, which is different from what was expected.

## Expected Behaviour

The self recorded Audio should be inserted into the page only, not the entire story. Users should be able to self-record different audio for each story page.

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

merapi:
## Summary

This PR fixes two issues with Background Audio:
#11921
#11920

Mic-only support:
If you have a device without a built-in camera please test this PR. You can also attach the camera and check edge cases - I did a bunch of testing with or without resetting the permissions at different stages of the process.
The worst thing that can happen is you start a fresh session (after perms reset) and grant audio-only permission and then attach the camera - you will have to reload the page for the camera permission popup.

## Testing Instructions

1. You need a device without a built-in camera.
2. Check if the audio recording works.

## Checklist

---

Fixes #11921
Fixes #11920

### DIFF
--- a/packages/story-editor/src/components/mediaRecording/footer.js
+++ b/packages/story-editor/src/components/mediaRecording/footer.js
@@ -293,9 +293,17 @@ function Footer({ captureImage, videoRef }) {
     actions: { uploadFile },
   } = useUploader();
 
-  const { updateStory } = useStory(({ actions: { updateStory } }) => ({
-    updateStory,
-  }));
+  const { updateCurrentPageProperties, currentPage, setSelectedElementsById } =
+    useStory(
+      ({
+        state: { currentPage },
+        actions: { updateCurrentPageProperties, setSelectedElementsById },
+      }) => ({
+        updateCurrentPageProperties,
+        setSelectedElementsById,
+        currentPage,
+      })
+    );
   const { setHighlights } = useHighlights(({ setHighlights }) => ({
     setHighlights,
   }));
@@ -318,12 +326,18 @@ function Footer({ captureImage, videoRef }) {
       resource: objectPick(resource, Object.keys(BackgroundAudioPropTypeShape)),
     };
 
+    const backgroundElement = currentPage?.elements.find(
+      ({ isBackground }) => isBackground
+    );
+    setSelectedElementsById({ elementIds: [backgroundElement.id] });
     setHighlights({
-      highlight: states.BACKGROUND_AUDIO,
+      highlight: states.PAGE_BACKGROUND_AUDIO,
     });
 
-    updateStory({
-      properties: { backgroundAudio },
+    updateCurrentPageProperties({
+      properties: {
+        backgroundAudio,
+      },
     });
     setIsInserting(false);
     toggleRecordingMode();
@@ -348,7 +362,7 @@ function Footer({ captureImage, videoRef }) {
               ? __('Inserting…', 'web-stories')
               : hasVideo
               ? __('Insert', 'web-stories')
-              : __('Insert background audio', 'web-stories')}
+              : __('Insert page background audio', 'web-stories')}
           </InsertButton>
         </>
       )}
--- a/packages/story-editor/src/components/mediaRecording/footer.js
+++ b/packages/story-editor/src/components/mediaRecording/footer.js
@@ -293,17 +293,23 @@ function Footer({ captureImage, videoRef }) {
     actions: { uploadFile },
   } = useUploader();
 
-  const { updateCurrentPageProperties, currentPage, setSelectedElementsById } =
-    useStory(
-      ({
-        state: { currentPage },
-        actions: { updateCurrentPageProperties, setSelectedElementsById },
-      }) => ({
-        updateCurrentPageProperties,
-        setSelectedElementsById,
-        currentPage,
-      })
-    );
+  const {
+    updateCurrentPageProperties,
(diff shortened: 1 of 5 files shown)


packages/story-editor/src/components/mediaRecording/provider.js
```diff
@@ -143,12 +143,23 @@ function MediaRecordingProvider({ children }) {
     // granted permission, so it's not easy to detect.
     // TODO: Figure out how to retry without microphone if possible.
     mediaStreamConstraints: {
-      audio: audioInput && hasAudio ? { deviceId: audioInput } : true,
-      video: videoInput && hasVideo ? { deviceId: videoInput } : true,
+      audio: hasAudio ? (audioInput ? { deviceId: audioInput } : true) : false,
+      video: hasVideo ? (videoInput ? { deviceId: videoInput } : true) : false,
```
merapi, line 147: For the mic-only device we need to request a device with `video: false` - we do that when one of the errors occurs.

spacedmonkey (approved):
Seems to work well.

packages/story-editor/src/components/mediaRecording/footer.js
```diff
@@ -318,12 +326,18 @@ function Footer({ captureImage, videoRef }) {
       resource: objectPick(resource, Object.keys(BackgroundAudioPropTypeShape)),
     };
 
+    const backgroundElement = currentPage?.elements.find(
+      ({ isBackground }) => isBackground
+    );
```
swissspidy: Let's move this to the `useStory` selector so that it already returns the `backgroundElementId` directly and not just the whole `currentPage` object

packages/story-editor/src/components/mediaRecording/provider.js
```diff
@@ -143,12 +143,23 @@ function MediaRecordingProvider({ children }) {
...
     },
     onStop,
   });
 
+  useEffect(() => {
+    if (
+      error?.name === 'NotFoundError' ||
+      error?.name === 'NotReadableError' ||
+      error?.name === 'OverConstrainedError'
```
swissspidy, line 156: IIRC in some browsers this error can occur even when the user has not yet granted permission, so I hope this does not cause any issue in some edge cases. Just noting it here for visibility.

swissspidy (commented):
Unfortunately I don't think I have a way to test this, but the change looks reasonable.

Just one nit

packages/story-editor/src/components/mediaRecording/footer.js
merapi: https://github.com/GoogleForCreators/web-stories-wp/pull/11939/commits/5beefd2fd0cd2be00e0c81c05f23fdf2579f4a8a

merapi:
This is what it looks like when you don't have a camera, only a mic (tested on W10 Edge/FF/Opera).
`NotFoundError` is thrown because it cannot find a device with `video/audio=true` then we set `hasVideo=false` so it requsts `video=false, audio=true` device and we display Audio Recording UI.

While the message is correct (No camera found) maybe we can add something about the audio-only option if they grant mic permissions?

image

packages/story-editor/src/components/mediaRecording/provider.js
merapi: I haven't seen it anywhere, for me, it always looks like this:
https://github.com/GoogleForCreators/web-stories-wp/pull/11939#issuecomment-1188305242

packages/story-editor/src/components/mediaRecording/provider.js
```diff
@@ -143,12 +143,23 @@ function MediaRecordingProvider({ children }) {
...
+    ) {
+      setHasVideo(false);
```
miina, line 158: Q: Should this influence the "Enable Video" quick-action as well? Can't test it unfortunately but looks like the button would still allow turning video on again (even though I suppose it'd give the error again and thus basically do nothing). Thoughts?

miina (approved):
LGTM! Also tried it out by setting the error in code directly, seems to work well like that.

packages/story-editor/src/components/mediaRecording/provider.js
merapi: In the next line we do `setVideoInput(null);` which hides the enable/disable video icon.

felipebochehin87:
Audio is added only to the page that is selected. If I clone this page, audio is cloned. If I add a new page, no audio is added to it. And if I have multiple pages and I record an audio, only the selected page has the audio linked to it.

PR #11940, opened 2022-07-18
Include border into the element selection line

### LINKED ISSUE #5325: Allow selecting element by clicking on the outer border

miina:
## Context
Originally only an "outer" border was implemented outside of the selection line.
It was discussed later that the border should also be included in the selection line, so the user can snap by the border and also select an element by clicking on the border.

## Summary
This PR adjusts the frame, edit, display mode to include the border into the selection line.

## Relevant Technical Choices
- Note that the border width does not change the width and height of the element directly as originally intended since the border is in pixels and not responsive.
- Non-rectangular shapes have a different border logic and are already displaying the frame somewhere in the middle of the border -- these will not have the selection line perfectly including the border now either. This change mainly applies to rectangular elements / shapes.
- Edit mode stays the same for media -- border is not included in the edit mode.

## To-do

- [x] Fix display with unlocked border
- [x] Fix text edit with unlocked border
- [x] Fix resizing multiple elements
- [x] Refactor
- [x] Fix tests
- [x] Add new test for selecting by clicking on border

## User-facing changes
text-edit
border-resizing
border-display

## Testing Instructions

1. Add rectangular elements with border (media, text, shape)
2. Increase and decrease the border -- verify that border is now included in the selection line.
3. Verify it's possible to drag from the border
4. Verify it's possible to select an element by clicking on the border
5. Verify that resizing still works and isn't jumpy / changing the element size unexpectedly.
6. Verify that border is also included in the selection in text edit mode.
7. Verify that media edit mode does not include border (is as before)
8. Verify that output is not influenced by this change.
9. Verify that when adding border to an element, the content of the element stays in the same place.

## Checklist

---

Fixes #5325


merapi:
> Note that the border width does not change the width and height of the element directly as originally intended since the border is in pixels and not responsive.

A side effect of that(?) is the alignment for those elements is trickier, we should probably consult that with UX but Andrew is out until Wednesday.

(elements in both groups are aligned to the left)
image

miina:
> > Note that the border width does not change the width and height of the element directly as originally intended since the border is in pixels and not responsive.
> 
> A side effect of that(?) is the alignment for those elements is trickier, we should probably consult that with UX but Andrew is out until Wednesday.
> 

This was actually like this before as well -- since the border width isn't included to the width/height currently either. Since we're using non-responsive CSS border but the elements are responsive, using the border to align is not really possible and is as it was before. We could create a follow-up issue if this needs changing.

merapi:
> This was actually like this before as well

Oh, that's good then - no change in behavior 👍🏻

merapi (approved):
LG, @barklund PTAL and let's move it to QA because there is a lot to check.

miina:
Moved it to QA already in case something time-consuming comes up but would be still good to get a review from @barklund, too.

packages/story-editor/src/components/canvas/frameElement.js
```diff
@@ -326,7 +326,7 @@ function FrameElement({ id }) {
         <Wrapper
           ref={combinedFocusGroupRef}
           data-element-id={id}
-          {...box}
+          {...boxWithBorder}
```
barklund, line 329: Right-clicking on the border does not include the bordered element in the select dropdown:
Screen Shot 2022-07-28 at 16 06 16

packages/story-editor/src/components/canvas/frameElement.js
```diff
@@ -192,12 +192,12 @@ function FrameElement({ id }) {
...
   const isLinkActive = isAnythingTransforming && !isLocked;
 
-  const getBox = useUnits(({ actions }) => actions.getBox);
+  const getBoxWithBorder = useUnits(({ actions }) => actions.getBoxWithBorder);
 
   useLayoutEffect(() => {
     setNodeForElement(id, elementRef.current);
   }, [id, setNodeForElement]);
-  const box = getBox(element);
+  const boxWithBorder = getBoxWithBorder(element);
```
barklund, line 200: Lasso-selecting on the canvas and including the border of an element but only the border, does not select that element:
border-lasso

packages/story-editor/src/components/canvas/multiSelectionMoveable/useResize.js
```diff
@@ -53,13 +56,17 @@ function useResize({ onGroupEventEnd, targetList, setTransformStyle, frames }) {
...
         newHeight = dataToEditorY(updates.height);
       }
-      target.style.width = `${newWidth}px`;
-      target.style.height = `${newHeight}px`;
+
+      // We add the border size back for the target display.
+      const frameWidth = newWidth + left + right;
+      const frameHeight = newHeight + top + bottom;
+      target.style.width = `${frameWidth}px`;
+      target.style.height = `${frameHeight}px`;
```
barklund, line 82: Somewhere around here there's a bug when resizing a bordered element as part of multi-select. You can make it negative size!
negative-size

barklund (commented):
Generally looks great, but found a few edge-cases with bugs.

packages/story-editor/src/components/canvas/frameElement.js
miina: @barklund This should be fixed now -- the lasso has been updated to use `useElementPolygon` instead, seems to work great and made things a bit simpler: https://github.com/GoogleForCreators/web-stories-wp/pull/11940/commits/43e0f5f46fa7a4571b3bd9c19d98c2ab5be74179

miina: lasso

packages/story-editor/src/components/canvas/multiSelectionMoveable/useResize.js
miina: This was a general bug with the multi-selection moveable -- it didn't consider the min width/height set for each element. Fixed now.

packages/story-editor/src/components/canvas/frameElement.js
miina: Fixed now within `useElementPolygon`. [d1245a7](https://github.com/GoogleForCreators/web-stories-wp/pull/11940/commits/d1245a7dfe82ed6bbfeb0ff00c4b11247799ac37)

Issue #9964, opened 2021-12-10
Media: Filtering media does not work with uploading assets


spacedmonkey:
## Bug Description

If you filter the media library by image, videos that are uploading still displayed. 

## Steps to Reproduce

1. Open editor
2. Upload video. 
3. Filter by image. 

## Screenshots

https://user-images.githubusercontent.com/237508/145632762-da883e22-021b-4a44-b000-92c85523ef74.mov

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

swissspidy:
IIRC that was a deliberate design decision to always show uploading items like that.

Likewise, if you are already filtering by videos and upload an image, it should still be visible.

spacedmonkey:
Reopening as this ticket now valid again. When filtering, uploaded media is lost. See video. 



https://user-images.githubusercontent.com/237508/179958667-0df3c40f-76f5-412e-90b9-bd13848ec247.mov

The issue is this line 
https://github.com/GoogleForCreators/web-stories-wp/blob/77bc376ee9e406bf69e9534fc9d2b72617ae0b48/packages/story-editor/src/app/media/local/reducers/setMediaType.js#L27

Sadly this was missed as part of https://github.com/GoogleForCreators/web-stories-wp/pull/9785. 

There is no easy way to work out if a resource is loading, other than to use `isCurrentResourceUploading`. Which might a problem.

swissspidy:
@spacedmonkey Separate reducers and state for uploading items is redundant since we already have this information in our `useMediaUploadQueue` hook. So this can be simplified a bit.

See https://github.com/GoogleForCreators/web-stories-wp/compare/try/9964-prepend-items?expand=1 for a potential solution.

spacedmonkey:
> See https://github.com/GoogleForCreators/web-stories-wp/compare/try/9964-prepend-items?expand=1 for a potential solution.

Looks interesting and like not replicating state in two places. It work noting, in my testing, it did not work. Once the upload finished, the new resource disappeared. Simply moving `prependMedia` to when the media is finished here. 

https://github.com/GoogleForCreators/web-stories-wp/blob/b44192a83fc26ff6c6fd68c4f924ed3db44b8871/packages/story-editor/src/app/media/useUploadMedia.js#L192-L196

Should we good to go.

PR #11941, opened 2022-07-18
Media: Processing items missing after opening media dialog

### LINKED ISSUE #9963: Media: Processing items missing after opening media dialog
## Bug Description

While a uploading a media asset, if you open the media dialog and close it, the uploading item disappears. 

## Steps to Reproduce

1. Open editor. 
2. Upload a video. 
3. Open media dialog. 
4. Close media dialog. 
5. See media is missing. 

## Screenshots

https://user-images.githubusercontent.com/237508/145632298-5237e4cf-696e-47a5-9695-45515470ae42.mov

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

spacedmonkey:
## Summary

Add `onUpload` and `onDelete` callbacks to media upload component. This has a number of key benefits. This allow for media  added and removed from within the editor. For example, if an image is delete from the dialog, it is removed from the media library and on the canvas. 

I have also taken this opportunity to use this new `onDelete` to improve background audio and captions. If these files are deleted within the dialog, then background audio and caption are reset / removed. 

## User-facing changes

https://user-images.githubusercontent.com/237508/179557163-02d60356-78a7-4eb7-ac6e-9083b1fcca83.mov

## Testing Instructions

1. Create a story. 
2. Open media dialog. 
3. Upload image. 
4. Insert into canvas. 
5. See new image into media library. 
---
1. Create a story. 
2. Open media dialog. 
3. Insert image into canvas. 
4. Open media dialog. 
5. Select image ( the one used before ). 
6. Delete image 
7. Close dialog. 
8. See image removed from canvas and in media library.
---
1. Create a story. 
2. Select page
3. Go to style tab. 
4. Upload audio file.
5. Select file. 
6. Click replace file. 
7. Delete image ( the one used before ). 
8. Close dialog. 
9. See background audio is reset. 

## Checklist

---

Fixes #9963


packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
```diff
@@ -108,14 +115,16 @@ export const MediaPicker = ({ render, ...props }) => {
     ({
       state: { canTranscodeResource },
       actions: {
-        resetWithFetch,
```
swissspidy, line 113: Can we now completely remove the `resetWithFetch` function from `packages/story-editor/src/app/media/local/useContextValueProvider.js`?

packages/story-editor/src/components/form/mediaUploadButton.js
```diff
@@ -157,11 +164,29 @@ function MediaUploadButton({ onInsert, renderButton, buttonInsertText }) {
...
+        media: [resource],
+      });
+    },
+    [prependMedia]
+  );
+
+  const onDelete = useCallback(
+    (resource) => {
+      deleteMediaElement({ id: resource?.id });
+      deleteElementsByResourceId({ id: resource?.id });
```
swissspidy: Why the optional chaining, can `resource` really be `null` here?

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
```diff
@@ -217,12 +226,31 @@ export const MediaPicker = ({ render, ...props }) => {
...
+        media: [resource],
+      });
+    },
+    [prependMedia]
+  );
+
+  const onDelete = useCallback(
+    (resource) => {
+      deleteMediaElement({ id: resource?.id });
+      deleteElementsByResourceId({ id: resource?.id });
```
swissspidy: Why the optional chaining, can `resource` really be `null` here?

packages/story-editor/src/components/panels/shared/media/captionsPanelContent.js
```diff
@@ -104,6 +104,12 @@ function CaptionsPanelContent({
   );
 
   const onError = useCallback((err) => trackError(eventName, err?.message), []);
+  const onDeleteTrack = useCallback(
+    (deletedResource) => {
+      handleRemoveTrack(deletedResource?.id);
```
swissspidy: Why the optional chaining, can `deletedResource ` really be `null` here?

packages/wp-story-editor/src/components/mediaUpload/mediaPicker/useMediaPicker.js
```diff
@@ -92,7 +99,58 @@ function useMediaPicker({
...
+        if (window.confirm(window.wp.media.view.l10n.warnDelete)) {
+          this.model.destroy({
+            wait: true,
+            error: function () {
+              window.alert(window.wp.media.view.l10n.errorDeleting);
+            },
+          });
+
+          this.moveFocus();
+        }
```
swissspidy: Where is this copied from? Instead of copying, can we just call the original `deleteAttachment` function or something? Ditto for `trashAttachment` below

spacedmonkey: Fixed.

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
spacedmonkey: Removed.

packages/wp-story-editor/src/components/mediaUpload/mediaPicker/useMediaPicker.js
```diff
@@ -92,7 +99,35 @@ function useMediaPicker({
...
+          // eslint-disable-next-line prefer-rest-params -- expected.
+          arguments
+        );
+
+        if (onDelete) {
+          onDelete(getResourceFromMediaPicker(this.model.attributes));
+        }
+      },
+    });
+    window.wp.media.view.Attachment.Details = wsDetails;
```
swissspidy, line 136: Probably want to do a cleanup here:

```suggestion
    window.wp.media.view.Attachment.Details = wsDetails;
    
    return () => {
      window.wp.media.view.Attachment.Details = currentDetails;
    };
```

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
```diff
@@ -217,12 +226,31 @@ export const MediaPicker = ({ render, ...props }) => {
       showSnackbar,
     ]
   );
+
+  const onUpload = useCallback(
+    (resource) => {
+      prependMedia({
+        media: [resource],
+      });
```
swissspidy: What if I filtered my media library by just images or videos instead of "All Types"? Then `prependMedia()` causes a video to be prepended to my list of images like this:

Screenshot 2022-07-19 at 13 10 01

I see two options here:

1. We wrap this call in a check for `mediaType`
2. We add a `mediaType` check directly to the `prependMedia` reducer

Option (2) seems more robust, in case we forgot to do (1) in some places. Plus we can easily add tests.

WDYT?

packages/story-editor/src/components/form/mediaUploadButton.js
```diff
@@ -157,11 +164,29 @@ function MediaUploadButton({ onInsert, renderButton, buttonInsertText }) {
...
+    [prependMedia]
+  );
+
+  const onDelete = useCallback(
+    (resource) => {
+      deleteMediaElement({ id: resource.id });
+      deleteElementsByResourceId({ id: resource.id });
+    },
+    [deleteElementsByResourceId, deleteMediaElement]
+  );
```
swissspidy: Since this is identical to the code in `useQuickActions.js`, and overall the pattern is quite repetitive, I am wondering whether it would be easiest to move (parts of) this to a shared component.

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
spacedmonkey: I am seeing this issue in main. So if a drag a video into canvas when I am filtering image, I see it prepended to media library. I think We should add this to `prependMedia` reducer but in maybe into another PR, as it is another issue,

packages/story-editor/src/components/form/mediaUploadButton.js
spacedmonkey: Fixed.

packages/story-editor/src/components/publishModal/content/storyPreview.js
```diff
@@ -298,6 +311,7 @@ const StoryPreview = () => {
                 {(hasUploadMediaAction || enablePosterHotlinking) && (
                   <StyledMedia
                     onChange={handleChangePoster}
+                    onDelete={onDelete}
```
miina: Just curious, wasn't deleting purposefully not added to the publishing modal? Since it's required for publishing.

packages/story-editor/src/components/panels/design/videoAccessibility/videoAccessibility.js
```diff
@@ -171,6 +180,7 @@ function VideoAccessibilityPanel({ selectedElements, pushUpdate }) {
           value={poster}
           cropParams={cropParams}
           onChange={handleChangePoster}
+          onDelete={onDelete}
```
miina: Here the Delete is essentially the same as Reset now, right? Since it's not actually fully deleted for the user, there's always a default. Or in case of Delete, the image is actually deleted from the library as well, in case of Reset not?

miina: ^ I wouldn't think as a user that "Delete" here would mean deleting from the library as well 🤔 I'd assume that the poster would just be deleted from that field.

spacedmonkey: @miina I don't think you understand what delete means in this context. If a user clicks delete in this dialog, it is completely deleted, file and all. 

Screenshot 2022-07-20 at 09 26 58

In that context, reseting makes sense here, as the file the user selected no longer exists.

packages/story-editor/src/components/publishModal/content/storyPreview.js
spacedmonkey: https://github.com/GoogleForCreators/web-stories-wp/pull/11941#discussion_r925324309

packages/story-editor/src/components/form/mediaUploadButton.js
spacedmonkey: Fixed. Reused compontent.

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
spacedmonkey: This took a lot longer to track down than I thought. I have a fix here, see what you https://github.com/GoogleForCreators/web-stories-wp/pull/11941/commits/dbfeb95d7c03162512628453214e9916c5a31af5

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
```diff
@@ -180,51 +107,11 @@ export const MediaPicker = ({ render, ...props }) => {
...
-      postProcessingResource,
-      transcodableMimeTypes,
-      optimizeVideo,
-      optimizeGif,
-      showSnackbar,
-    ]
-  );
   return (
-    <MediaUpload
+    <MediaUploadButton
```
swissspidy, line 113: Neat, always nice to see code removals like this. And it works without any issues?

packages/story-editor/src/app/media/local/reducers/setMediaType.js
```diff
@@ -24,7 +28,7 @@ function setMediaType(state, { mediaType }) {
   }
   return {
     ...INITIAL_STATE,
-    media: state.media.filter(({ local }) => local), // This filter allows remove temporary file returned on upload
+    media: state.media.filter(({ src }) => isBlobURL(src)), // This filter allows remove temporary file returned on upload
```
swissspidy: Not sure I really understand the purpose of this filter here, even in it's previous form 🤔 

Doesn't that _keep_ temporary files instead of _removing_ them?

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
swissspidy: > I am seeing this issue in main. So if a drag a video into canvas when I am filtering image, I see it prepended to media library. 

Right. Just found #9964 because of this, indicating that we did that on purpose.

dbfeb95d7c03162512628453214e9916c5a31af5 means that you'll still see the uploading items when you change the filter again, right?

That couldn't be done in the reducer, so it kinda makes sense to put it there.

On the other hand, it's weird that we have to filter the media elements at that stage. Would be better to put this into `useContextValueProvider` directly, not have it in the component.

But yes, let's do that in a separate PR so we can refine & test it better.

packages/story-editor/src/app/media/local/reducers/setMediaType.js
spacedmonkey: Before, we had a local property on the resource object. This broke a while back. This is a check to filter out all "local" resources. We can no longer do this as state is not managed here anymore.

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
spacedmonkey: I have reverted https://github.com/GoogleForCreators/web-stories-wp/commit/dbfeb95d7c03162512628453214e9916c5a31af5 let's handle this in https://github.com/GoogleForCreators/web-stories-wp/issues/9964.

packages/story-editor/src/app/media/local/reducers/setMediaType.js
swissspidy: Yeah but to filter out local resources you'd do `state.media.filter(({ local }) => ! local)` and not `state.media.filter(({ local }) => local)` 

Again, this would _keep_ the items, not _remove_ them, which seems odd.

Am I missing something?

packages/story-editor/src/components/panels/design/videoAccessibility/videoAccessibility.js
swissspidy: Is it worth renaming `onDelete` to something else to make it less confusing?

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
spacedmonkey: Seems to.

packages/story-editor/src/components/panels/design/videoAccessibility/videoAccessibility.js
spacedmonkey: How about onDeleteMedia and onUploadMedia?

packages/story-editor/src/app/highlights/quickActions/useQuickActions.js
spacedmonkey: Left my notes on this here https://github.com/GoogleForCreators/web-stories-wp/issues/9964#issuecomment-1190102837

packages/story-editor/src/components/panels/design/videoAccessibility/videoAccessibility.js
miina: Yep, that is more clear when reading the code.

spacedmonkey: Changed.

packages/wp-story-editor/src/components/mediaUpload/mediaPicker/useMediaPicker.js
```diff
@@ -75,6 +84,44 @@ function useMediaPicker({
...
+  const deleteMedia = useCallback(
+    (resource) => {
+      deleteMediaElement({ id: resource.id });
+      deleteElementsByResourceId({ id: resource.id });
+      if (onDeleteMedia) {
+        onDeleteMedia(resource);
+      }
+    },
+    [deleteElementsByResourceId, deleteMediaElement, onDeleteMedia]
+  );
```
swissspidy: This should not be in the wp-story-editor package, since actions like prepending media and deleting media elements is something that also needs to happen in a decoupled/standalone context.

packages/story-editor/src/components/mediaUpload/mediaUpload.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+  );
+
+  const uploadMedia = useCallback(
+    (resource) => {
+      if (['image', 'video', 'gif'].includes(resource.type)) {
+        prependMedia({
+          media: [resource],
+        });
+      }
```
spacedmonkey, line 64: We could call `postProcessingResource` to trigger post processing resource early.

swissspidy: Doesn't seem necessary because it will be called right away already due to this effect:

https://github.com/GoogleForCreators/web-stories-wp/blob/22cf54d030e407f469350f21273f6d1c223aa99b/packages/story-editor/src/app/media/local/useContextValueProvider.js#L288-L292

spacedmonkey:
So just an update here. 

I did some playing around with this. I basically adding some state, to add an array of new resources added and then when the dialog is completed only add new resources on dialog close. Something like this. 

```
  useEffect(() => {
    if (doProcess && media) {
      console.log(media);
      prependMedia({
        media,
      });
      setDoProcess(false);
      setMedia([]);
    }
  }, [doProcess, media, prependMedia]);

  const onClose = useCallback(() => {
    if (onCloseCallback) {
      onCloseCallback();
    }
    setDoProcess(true);
  }, [onCloseCallback]);

  const uploadMedia = useCallback(
    (resource) => {
      if (['image', 'video', 'gif'].includes(resource.type)) {
        setMedia((state) => [resource, ...state]);
      }
      if (onUploadMedia) {
        onUploadMedia(resource);
      }
    },
    [onUploadMedia]
  );
```

However, I got into state based problems. The above code worked well on the first run, add these callbacks are not rehooked, then on close i was getting an empty array. I am not sure I was around this. If anyone else has any thoughtst that would be great.

swissspidy:
Quick summary of my chat with Jonny about this yesterday:

UX-wise instantly adding the media to the library would be preferable. But that currently causes issues with race conditions. Example scenario:

1. Open media modal
2. Upload file
3. File is added to the editor media library in the background -> this triggers things like blurhash processing, etc. due to this effect:  
  https://github.com/GoogleForCreators/web-stories-wp/blob/621f5c3412ae692e3881057583b7f6fb57485c42/packages/story-editor/src/app/media/local/useContextValueProvider.js#L278-L282
4. In the media modal, click on "Insert" media -> the `onSelect`/`onInsert` callback calls `postProcessingResource`
    https://github.com/GoogleForCreators/web-stories-wp/blob/621f5c3412ae692e3881057583b7f6fb57485c42/packages/story-editor/src/components/form/mediaUploadButton.js#L135
5. The problem: at this point the `postProcessingResource` callback will be stale and is thus not "aware" that the processing already happened in step 3. This causes everything to run twice.

We discussed that we could perhaps remove the `postProcessingResource` call and instead trigger this via an effect (similar to the existing effect).

AIs:

* Pascal to take a closer look into this to see what kind of options we have here.

spacedmonkey:
@swissspidy I think I have found a solution to this problem using a ref. See https://github.com/GoogleForCreators/web-stories-wp/pull/11941/commits/2d34582d4b0140d9cdf1ea4fab1c1750101b4d81. 

Take a look at the commit and see what I am doing here.

swissspidy:
The change is a bit hard to read/follow, i's not obvious what it does, can you elaborate a bit? 

As per my above comment I plan on taking a closer look at this problem to better understand it. But that probably won't be today anymore I'm afraid

spacedmonkey:
@swissspidy That is happening in last commit. 

The instance of wp.media in now stored / cached in a ref. This means, that a new instance is not created on every usage, keeping state. 

On adding / remove of media in the dialog, an array ( queue ) of resources is saved. On closing the dialog, resources are added to the media library all at once. Stopping race conditions, where adding resource in the background results in base color/blurhash/poster is generated in the background and not updated image/video element later inserted into canvas.

packages/story-editor/src/components/mediaUpload/mediaUpload.js
```diff
@@ -0,0 +1,129 @@
...
+    },
+    [featuredMedia, updateStory]
+  );
+
+  const deleteMedia = useCallback(
+    (resource) => {
+      deleteMediaElement({ id: resource.id });
+      deleteElementsByResourceId({ id: resource.id });
+      setNewResources((state) => state.filter(({ id }) => id !== resource.id));
+      resetFeatureMedia(resource.id);
```
timarney:
```suggestion
      const id = resource.id;
      deleteMediaElement({ id });
      deleteElementsByResourceId({ id });
      setNewResources((state) => state.filter(({ newId }) => newId !== id));
      resetFeatureMedia(id);
```

swissspidy:
> On adding / remove of media in the dialog, an array ( queue ) of resources is saved. On closing the dialog, resources are added to the media library all at once.

As per my previous comments and the previous discussion, I would really prefer feedback straightaway, **no** queue.

spacedmonkey:
@swissspidy @timarney is one of you going to take this PR on? Or will wait until I get back.  

I think it is close to mergable but I haven't got review yet.

timarney:
> Pascal to take a closer look into this to see what kind of options we have here.

@swissspidy are you still going to [look at options](https://github.com/GoogleForCreators/web-stories-wp/pull/11941#issuecomment-1192331381) for this one?

swissspidy:
Yes I still intend to look into it. Haven't found time yet to fully dig into it though, but it's not urgent anyways

swissspidy:
I went down quite the rabbit whole trying to analyze this one and coming up with a good solution. While it feels like we're 90% there, it's clear now that this needs some more thinking and tinkering to nail down the remaining 10%. And given the impact of this change, also tests.

Some preliminary findings in no particular order:

* Storing the `wp.media` instance in a ref had negative side effects like stale data. For example when when deleting an item in the modal and closing & reopening it, you’d still see the deleted item there.  
* Adding items to the library (either directly or in a queue) that were uploaded to the media modal is not always wanted.  
  * For example MOV files should not be added to the library since they require transcoding
  * It also causes issues for animated GIFs, where the GIF is added to the library, and then we convert it to an MP4 and you briefly see 2 items in the library instead of just 1.  
      * This could be solved by some kind of in-place update of the GIF element in the media library

Related: #12123

spacedmonkey:
This PR is now out of date. Closing and re-openning if I have a better idea.

PR #11966, opened 2022-07-20
Code Quality: remove `incrementalSearchDebounceMedia` experiment

### LINKED ISSUE #11942: Code Quality: remove `incrementalSearchDebounceMedia` experiment
## Task Description

This has been lingering around for too long. Let's remove the code for this experiment.

timarney:
## Context

 Removes `incrementalSearchDebounceMedia` experiment.

## Summary

Removes incremental and delay props from search

## Testing Instructions

1.

## Checklist

---

Fixes #11942


packages/story-editor/src/flags.js
```diff
@@ -20,9 +20,6 @@
  *
  * @enum {string}
  */
-const Flags = {
```
swissspidy, line 23: I think we can remove this whole file now

packages/story-editor/src/components/library/common/searchInput.js
```diff
@@ -71,24 +66,24 @@ function WrappedSearchInput({
 
   const submitValue = useCallback(
     (value) => {
-      if (incremental && delayMs) {
+      if (delayMs) {
         changeSearchTermDebounced();
```
swissspidy: Just removing the `incremental` check  completely changes the meaning of this.

We also need to remove `delayMs` and `changeSearchTermDebounced` completely, because they are only relevant for incremental search.

Basically, we can just remove `submitValue` and call `onSearch` directly

packages/story-editor/src/components/library/common/searchInput.js
```diff
@@ -58,50 +49,32 @@ function WrappedSearchInput({
...
-      if (incremental && delayMs) {
-        changeSearchTermDebounced();
-      } else {
-        onSearch(value);
-      }
-    },
-    [changeSearchTermDebounced, onSearch, delayMs, incremental]
-  );
-
   const onChange = useCallback(
```
timarney, line 56: @swissspidy 

Should we remove the onChange as well and let onClear or "" + Enter --- clear the search ?

If yes I should be able to remove the `localValue` as well.

timarney: We'll keep the current functionality --- onClear or 'backspace' clear  will reset the search.

timarney:
Using the prod build --- looks like it related to something with Story data deletion.

Without adding a story:

Screen Shot 2022-07-22 at 9 02 25 AM


After adding a story:
Screen Shot 2022-07-22 at 8 59 26 AM

swissspidy:
@timarney wrong PR?

swissspidy (approved):

packages/story-editor/src/components/library/common/searchInput.js
```diff
@@ -58,50 +49,32 @@ function WrappedSearchInput({
   placeholder,
   onSearch,
   disabled = false,
-  incremental = false,
-  delayMs = 500,
 }) {
   // Local state so that we can debounce triggering searches.
   const [localValue, setLocalValue] = useState(initialValue);
```
miina, line 54: Is saving the `localValue` still necessary? Couldn't we just take the current value of the input for `onSubmit` instead of keeping it in the state with `onChange`?

timarney: I think we would end up in a state where the search input field doesn't update as the value is passed via the inputValue prop.  Could make changes to https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/design-system/src/components/search/input/index.js#L80 but feels like that's outside of the scope of this PR.

miina (approved):

PR #11983, opened 2022-07-21
Fix Karma tests

merapi:
## Summary

I'm testing if my commit will fix all the issues.

So far fixed:
`should allow adding new text style from multi-selection`:
With higher thumbnails, the Canvas became a little bit smaller and because the Paragraph is only 21px high it was not selected by the 20,20 click.
Initially wanted to reuse our other `selectTarget` function that clicks in the center of the bounding box, but in this case it is also failing because of the rotation handle:
Screenshot 2022-07-21 at 15 25 14

| Old Canvas | New Canvas | New Old Canvas ;] |
|-----|-----|-----|
|  Screenshot 2022-07-21 at 15 14 37   |  Screenshot 2022-07-21 at 15 15 08   | Screenshot 2022-07-22 at 11 08 09 |

## Testing Instructions

1.

## Checklist

---

Fixes #

### DIFF
--- a/packages/story-editor/src/components/canvas/layout.js
+++ b/packages/story-editor/src/components/canvas/layout.js
@@ -85,7 +85,7 @@ const LayerGrid = styled.section`
   */
   grid:
     'h h h h h h h' ${HEADER_HEIGHT}px
-    '. . . t . . .' minmax(16px, 1fr)
+    '. . . t . . .' minmax(10px, 1fr)
     'p b . c . f n' var(--viewport-height-px)
     '. . . . . . .' 1fr
     'w w w w w w w' ${({ footerHeight }) => footerHeight}px
--- a/packages/story-editor/src/components/canvas/layout.js
+++ b/packages/story-editor/src/components/canvas/layout.js
@@ -85,7 +85,7 @@ const LayerGrid = styled.section`
   */
   grid:
     'h h h h h h h' ${HEADER_HEIGHT}px
-    '. . . t . . .' minmax(10px, 1fr)
+    '. . . t . . .' minmax(${HEADER_GAP}px, 1fr)
     'p b . c . f n' var(--viewport-height-px)
     '. . . . . . .' 1fr
     'w w w w w w w' ${({ footerHeight }) => footerHeight}px
--- a/packages/story-editor/src/constants/index.js
+++ b/packages/story-editor/src/constants/index.js
@@ -21,7 +21,7 @@ export * from './performanceTrackingEvents';
 
 export const ADMIN_TOOLBAR_HEIGHT = 32;
 export const HEADER_HEIGHT = 64;
-export const HEADER_GAP = 16;
+export const HEADER_GAP = 10;
 export const CANVAS_MIN_WIDTH = 570;
 export const SIDEBAR_WIDTH = 360;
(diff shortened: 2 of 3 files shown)


packages/story-editor/src/constants/index.js
```diff
@@ -21,7 +21,7 @@ export * from './performanceTrackingEvents';
 
 export const ADMIN_TOOLBAR_HEIGHT = 32;
 export const HEADER_HEIGHT = 64;
-export const HEADER_GAP = 16;
+export const HEADER_GAP = 10;
```
miina, line 24: Hm, is this reverting any of the changes requested by design / does this have any visual impact that needs to be followed up?

miina (approved):
Tests are passing!

packages/story-editor/src/constants/index.js
merapi: It looks like there are more places that are sensitive to the actual canvas size on the screen (my first fix didn't fix all of them), so I thought it would be better to get the old size back by reducing the whitespace by a tiny bit and tested that here, I see no cons to that solution and we get the old/bigger canvas (why make it smaller?) and the whitespace around the canvas is still big IMO (we could reduce it even more, but we would have to adjust the tests). WDYT @swissspidy ?

packages/story-editor/src/components/canvas/layout.js
```diff
@@ -85,7 +85,7 @@ const LayerGrid = styled.section`
   */
   grid:
     'h h h h h h h' ${HEADER_HEIGHT}px
-    '. . . t . . .' minmax(16px, 1fr)
+    '. . . t . . .' minmax(10px, 1fr)
```
swissspidy: Should we use `HEADER_GAP` here?

packages/story-editor/src/constants/index.js
swissspidy: Do you perhaps have a before/after screenshot for this change?

merapi: Added to the table^ @swissspidy

swissspidy: Thanks! The difference is indeed minimal. Seems fine to me 👍

merapi:
I'm getting a lot of:
```
[percy] Encountered an error uploading snapshot: CUJ: Text Sets (Text and Shape Combinations): Using Text Sets should allow user to drag and drop text set onto page: Text set dragged
[percy] Error: Cannot create snapshot, the build is already finalized. If this happens frequently in a parallel test suite, you may want to use a "finalize all" step in CI: https://docs.percy.io/docs/parallel-test-suites#using-finalize-all
```

swissspidy:
IIRC that happens when re-running the same job, because the nonce we use never changes. I can whip up a quick PR that might address this.

merapi:
> IIRC that happens when re-running the same job, because the nonce we use never changes. I can whip up a quick PR that might address this.

Right, I'm re-running the failed job, but I did in the past and it worked fine now 6th attempt with those failures 🤔

swissspidy:
But aren't there still some failing Karma tests? 🤔

merapi:
@swissspidy This is Karma 1/4:
`[percy] Error: Can only finalize pending builds
Error: Process completed with exit code 1.`

PR #12021, opened 2022-07-26
Code Quality: Remove hotlinking feature flag code 

### LINKED ISSUE #11997: Code Quality: Remove hotlinking feature flag code
## Task Description

This feature has been enabled by default in v1.23.0. Time to remove all code around the feature flag check.

Three feature flags:

* `videoPosterHotlinking`
* `linkIconHotlinking`
* `posterHotlinking`

timarney:
## Context

This feature has been enabled by default in v1.23.0. 

## Summary

Removes feature flags for:

`videoPosterHotlinking`
`linkIconHotlinking`
`posterHotlinking`

## Testing Instructions

1. Verify that hotlinking for story poster, video poster, and link icon is available and enabled and works for both contributors and admins
2. Other functionality should be left unchanged.

No

No

No

## Checklist

---

Fixes #11997


packages/wp-story-editor/src/components/documentPane/publish/publish.js
```diff
@@ -366,7 +363,7 @@ function PublishPanel({ nameOverride }) {
                 ariaLabel={__('Poster image', 'web-stories')}
                 onChangeErrorText={posterErrorMessage}
                 imgProps={featuredMedia}
-                canUpload={hasUploadMediaAction || enablePosterHotlinking}
+                canUpload={hasUploadMediaAction}
```
swissspidy: Isn't that a change in semantics?

Previously: `canUpload={hasUploadMediaAction || enablePosterHotlinking}`

But now we allow poster hotlinking by default (basically making `enablePosterHotlinking` always `true`), so technically this should be `canUpload={true}`, no?

packages/story-editor/src/components/publishModal/content/storyPreview.js
```diff
@@ -319,7 +316,7 @@ const StoryPreview = () => {
                     ariaLabel={__('Poster image', 'web-stories')}
                     onChangeErrorText={posterErrorMessage}
                     imgProps={featuredMedia}
-                    canUpload={hasUploadMediaAction || enablePosterHotlinking}
+                    canUpload={hasUploadMediaAction}
```
swissspidy: Same here, shouldn't this be `canUpload={true}`?

packages/story-editor/src/components/publishModal/content/storyPreview.js
```diff
@@ -295,7 +292,7 @@ const StoryPreview = () => {
                     data-testid="story_preview_logo"
                   />
                 )}
-                {(hasUploadMediaAction || enablePosterHotlinking) && (
+                {hasUploadMediaAction && (
```
swissspidy: Same here, with poster hotlinking always allowed we can remove this entire condition here, no?

packages/wp-story-editor/src/components/documentPane/publish/publish.js
timarney: Good point -- updated.

swissspidy:
Looks like some tests are failing now. Haven't checked closely, perhaps they need updating.

timarney:
> Looks like some tests are failing now. Haven't checked closely, perhaps they need updating.


```
expected document not to contain element, found <button data-testid="media-upload-button">Media Upload Button!</button> instead

      172 |     const uploadMediaButton = screen.queryByText('Media Upload Button!');
      173 |
    > 174 |     expect(uploadMediaButton).not.toBeInTheDocument();
          |                                   ^
      175 |   });
```

Looking into it.

timarney:
Tested with a user who doesn't have upload permissions  --- 

Screen Shot 2022-07-27 at 6 23 19 AM

=====

timarney:
Erroring on the `MediaInput`

> Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

    Check the render method of `ForwardRef(MediaInput)`.

includes/Experiments.php
```diff
@@ -289,42 +289,6 @@ public function get_experiments(): array {
...
-				'group'       => 'editor',
-				'default'     => true,
-			],
-			/**
-			 * Author: @spacedmonkey
-			 * Issue: #11777
-			 * Creation date: 2022-06-24
-			 */
-			[
-				'name'        => 'linkIconHotlinking',
```
miina, line 310: I can still see `videoPosterHotlinking` and `linkIconHotlinking` appearing in some Karma tests when doing a search in the whole codebase.

packages/story-editor/src/components/publishModal/content/storyPreview.js
miina: Not sure if you discussed this somewhere else that the comment was marked resolved, but since now `enablePosterHotlinking` is always `true`, wouldn't we always display this and can remove `hasUploadMediaAction` check as well?

timarney:
> Not sure if you discussed this somewhere else that the comment was marked resolved, but since now enablePosterHotlinking is always true, wouldn't we always display this and can remove hasUploadMediaAction check as well?

It was flagged for removal but removing seems to be causing other issues -- test failures + this error https://github.com/GoogleForCreators/web-stories-wp/pull/12021#issuecomment-1196559746 thinking there's something deeper into one of the child components around the `can user upload` perms.

miina:
> [#12021 (comment)](https://github.com/GoogleForCreators/web-stories-wp/pull/12021#discussion_r937569971)
> 
> > Not sure if you discussed this somewhere else that the comment was marked resolved, but since now enablePosterHotlinking is always true, wouldn't we always display this and can remove hasUploadMediaAction check as well?
> 
> It was flagged for removal but removing seems to be causing other issues -- test failures + this error [#12021 (comment)](https://github.com/GoogleForCreators/web-stories-wp/pull/12021#issuecomment-1196559746) thinking there's something deeper into one of the child components around the `can user upload` perms.

Hm, okay. I suppose we can leave as is then. Just by reading the code I assumed that the whole component could always displayed but `canUpload` should only be `true` in case of `hasUploadMediaAction`.

timarney:
> Hm, okay. I suppose we can leave as is then. Just by reading the code I assumed that the whole component could always displayed but canUpload should only be true in case of hasUploadMediaAction.

I'll take another look --- see if I can spot anything

swissspidy:
@timarney That `hasUploadMediaAction` must go away. Otherwise contributors cannot update the story poster in the pre-publish dialog. Notice the missing button here:

Screenshot 2022-08-09 at 11 10 10


Quickly addressed it in 96029c07d382d2a2214b9d3cfa0f298eb5a3af14 but there might be more tests that need updating.

timarney:
@swissspidy any idea why removing that causes issues with the MediaInput (at least in the tests).
 
 ```javascript
 detail: Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
  
  Check the render method of `ForwardRef(MediaInput)`.
  ```

swissspidy:
That test renders the pre-publish modal without defining `MediaUpload` in the `configValue` in its `arrange()` function. See how the `storyPreview.js` test does it:

https://github.com/GoogleForCreators/web-stories-wp/blob/a6b3cb6047426b680db026413ca69588b13e99d0/packages/story-editor/src/components/publishModal/test/storyPreview.js#L60-L68

Now that we removed this condition, the `MediaUpload` component is always rendered by this test. But because it's `undefined` there's an error.

swissspidy (approved):

miina (approved):

PR #12022, opened 2022-07-26
Editor: Changed video recording trim to happen before insertion

### LINKED ISSUE #11988: Media Recording - Trimming: Better user experience
## Feature Description

Current version:
When you click the Insert button after trimming - we insert the full-length video on the Canvas and the trimmed video will appear in the Media panel after the processing is done.

We want to change it so that you wait on the trimming UI (with a proper message) until the processing is done and we insert the trimmed video. It should also be easier to generate a correct thumbnail for the trimmed video that way?

So it will fix the issues from https://github.com/GoogleForCreators/web-stories-wp/pull/11908:
- [x] The preview on-canvas while the video is being trimmed and uploaded is also not trimmed, but shows the full duration
- [x] The generated thumbnail is taken from the start of the full video, not the start of the trimmed video

barklund:
## Context

This moves the actual trimming of the video recording up so it happens as soon as the user presses "trim", and not only when they press "insert".

## Todo

- [x] Make it work!
- [x] Fix bug around re-trimming ~(partially fixed, see comment below)~
- [x] Fix UX around waiting for trim to complete (with a progress bar perhaps?)

## User-facing changes
This flow now happens when trimming a new recording - the "cancel trim" button in the footer of course also works.
trimmy

## Testing Instructions

Steps to validate:
1. Start a video recording and record 10+ seconds of video
2. Click the trim button to start trimming the recording
3. Move any trim offset and press trim
4. Observe that trim happens instantly with a dialog informing you of this
5. Wait until trim completes (the shorter the faster)
6. Observe that the displayed video is only the trimmed bit
7. Click the trim button again
8. Observe that the original video is available to trim from, with the previously trimmed bit "preselected"
9. Abort trimming
10. Observe that the original trimmed video still displays correctly
11. Press insert
12. Observe that the trimmed video is uploaded and inserted to the page
13. Observe that the trimmed video has the correct duration and thumbnail

## Checklist

---

Fixes #11988


swissspidy:
Thanks for working on this 👍 

I originally had something simpler in mind without a dialog, but of course I didn't now the technical constraints around this.

Would be good to discuss this with @aaskedall to get his thoughts on this.

packages/story-editor/src/components/mediaRecording/useTrim.js
```diff
@@ -17,30 +17,46 @@
...
+
+function useTrim({ setDuration, onTrimmed, file }) {
   const [trimData, setTrimData] = useState({ start: 0, end: null });
   const [isTrimming, setIsTrimming] = useState(false);
+  const [isProcessingTrim, setIsProcessingTrim] = useState(false);
+  const { trimVideo } = useFFmpeg();
   const onTrim = useCallback(
-    (newTrimData) => {
+    async (newTrimData) => {
+      setIsTrimming(false);
```
miina: Hm, should this be set to `false` after the trimming has actually happened, after `newTrimData` as it was before? It seems that it's set to false right away `onTrim` currently?

packages/story-editor/src/components/mediaRecording/useTrim.js
```diff
@@ -17,30 +17,46 @@
...
       if (newTrimData) {
         setTrimData(newTrimData);
-        const millis = newTrimData.end - newTrimData.start;
-        const seconds = Math.ceil(millis / 1000);
-        setDuration(seconds);
+        setIsProcessingTrim(true);
```
miina, line 40: What's the difference between `processingTrim` and `isTrimming`? Does `processingTrim` include waiting until the playback has finished and `isTrimming` doesn't? Would be great to have a comment for readability to differentiate these two.

barklund: `isTrimming` means that the trimming interface is currently displayed and the user is adjusting trimming.
`isProcessingTrim` means that the "trimming in progress" dialog is currently displayed and the user is waiting. But yes, the first of these should probably be renamed to something more clear.

packages/story-editor/src/components/mediaRecording/useTrim.js
barklund: Yes, the trimming interface is hidden right away and the "trimming in progress" dialog is displayed instead.

barklund:
There is still an annoying bug with re-trimming an already trimmed video (while still in the recording mode). Sometimes the original blob URL doesn't work anymore, like it has been garbage collected (or `window.URL.revokeObjectURL` has been called), but I don't see why that should happen?

We need to fix this bug, but it's very inconsistent, so it's pretty hard to track down.

merapi:
We need to reset the last trimming session data, otherwise, you can end up at a dead end (no start/end markers):

image

The close button should probably work as a cancel button, otherwise, if you don't re-enter the trimming screen your video will be lost.

merapi (approved):
We have the known re-trimming bug and some edge cases to handle - otherwise looks good 👍🏻

packages/story-editor/src/components/mediaRecording/provider.js
```diff
@@ -233,12 +254,14 @@ function MediaRecordingProvider({ children }) {
   const previousBlobUrl = usePrevious(mediaBlobUrl);
 
   useEffect(() => {
-    if (previousBlobUrl && previousBlobUrl !== mediaBlobUrl) {
+    if (
+      previousBlobUrl &&
+      previousBlobUrl !== mediaBlobUrl &&
+      previousBlobUrl !== originalMediaBlobUrl
+    ) {
```
barklund, line 261: [I'm an idiot](https://github.com/GoogleForCreators/web-stories-wp/pull/12022#issuecomment-1204470892)!

> like it has been garbage collected (or window.URL.revokeObjectURL has been called), but I don't see why that should happen

Maybe because it has been revoked right here? 🤦  Well, I guess this fixes it though

packages/story-editor/src/components/mediaRecording/useTrim.js
```diff
@@ -17,30 +17,67 @@
...
     },
-    [setDuration]
+    [file, onTrimmed, setDuration, trimVideo]
   );
-  const startTrim = useCallback(() => setIsTrimming(true), []);
+  const startTrim = useCallback(() => setIsAdjustingTrim(true), []);
   const resetTrim = useCallback(() => {
-    setIsTrimming(false);
+    setIsAdjustingTrim(false);
     setTrimData({ start: 0, end: null });
```
barklund, line 59: I didn't actually reset the trim info in the reset function, so that's why it didn't work correctly.

PR #12044, opened 2022-07-28
Fix font loading

### LINKED ISSUE #12043: Some fonts are not loaded/displayed and have the wrong position
## Bug Description

Issue 1)
The font is not loaded/the wrong font is displayed.

Issue 2)
The position/bounding box is wrong.

Original report:
Text is not correctly displayed and takes more room than assigned,some examples:
Screenshot 2022-07-26 at 11 08 52
Screenshot 2022-07-26 at 11 07 32
Screenshot 2022-07-26 at 11 04 53
Screenshot 2022-07-26 at 11 02 31

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

merapi:
## Context

This PR addresses issue 1) (#12043).

Before:
image

After: 
image

## Testing Instructions

1.

## Checklist

---

Fixes #12043


swissspidy:
Can we add tests for this to prevent regressions?

merapi:
> Can we add tests for this to prevent regressions?

For font loading verification a good candidate can be the Percy diff because the bounding box doesn't change, so we can only detect the regression visually or by checking the font requests.
Unfortunately, we don't always check every Percy diff, so maybe we can do a visual comparison in Karma test for that?
I did a quick test with the Puppeteer `page.screenshot` function (integrated with our Karma Fixture) - it can return a Buffer so we can:
1) Compare that to other predefined Buffer or image (screenshot of an element with Teko font).
2) Check specific coords and their color (which we know will break if the font is not loaded).

WDYT @swissspidy ?

miina:
@merapi Since @swissspidy is out this week and you'll be out next week when he is back, maybe we can merge the fix and create a follow-up issue for the test? WDYT?

miina (approved):

swissspidy:
I‘m wary of introducing another type of visual regression tests.

Could we perhaps check for font stylesheets being loaded correctly?

barklund (approved):
Can we just check the height and width of the text field for the test perhaps?

merapi:
> Can we just check the height and width of the text field for the test perhaps?

Based on my testing - no:
> the bounding box doesn't change

---

> Could we perhaps check for font stylesheets being loaded correctly?

I will look at it in a follow-up, doesn't look trivial.

PR #12060, opened 2022-08-02
E2E Tests: Ensure WP updates to latest minor

timarney:
## Context

When running E2E tests we want to use the latest minor version of WordPress 5.7

## Summary

Updates WP install to force minor 

```
wp core update --minor --force --quiet
```

vs the existing

```
wp core update --force --quiet
```

## Testing Instructions

1. CI / E2E tests

no

no

no

## Checklist

---

Fixes #


.github/workflows/tests-e2e.yml
```diff
@@ -136,7 +136,7 @@ jobs:
       matrix:
         # TODO: add back Firefox once support is more mature.
         browser: ['chrome']
-        wp: ['5.7']
+        wp: ['5.7.1']
```
swissspidy: Interesting. Our Docker install script is supposed to install the latest version, but seems like it doesn't:

https://github.com/GoogleForCreators/web-stories-wp/runs/7628435294?check_suite_focus=true#step:8:674

Never realized that.

Instead of specifying 5.7.x here, let's update this code instead:

https://github.com/GoogleForCreators/web-stories-wp/blob/3041c6cacb733be3500e266815c9fadae21f260c/bin/local-env/install-wordpress.sh#L59-L63

Something like this:

```bash
# Potentially update WordPress
if [ "$WP_VERSION" == "latest" ]; then
	echo -e $(status_message "Updating WordPress to the latest major")
	wp core update --force --quiet
else
  echo -e $(status_message "Updating WordPress to the latest minor")
  wp core update --minor --force --quiet
fi
```

Then this change here should not be needed

timarney:
Screen Shot 2022-08-10 at 5 57 13 PM

bin/local-env/install-wordpress.sh
```diff
@@ -56,10 +56,13 @@ fi
...
+# Potentially update WordPress
 if [ "$WP_VERSION" == "latest" ]; then
-	# Potentially update WordPress
-	echo -e $(status_message "Updating WordPress")
+	echo -e $(status_message "Updating WordPress to the latest major")
 	wp core update --force --quiet
+else
+  echo -e $(status_message "Updating WordPress to the latest minor")
+  wp core update --minor --force --quiet
 fi
```
swissspidy, line 66: Let's merge this change already separately; it makes sense to do

timarney: @swissspidy I've repurposed this PR to only contain that update now.

swissspidy (approved):

Issue #11906, opened 2022-07-12
Layers rearrangement improvements


miina:
## Feature Description
There a few bugs / missing features in the layers arrangement currently that should be implemented:

- [ ] Drag and drop for multiple layers: (select 2 layers and drag-drop them in the Layers Panel - doesn't work, you can only drag-drop groups or single layers).
- [ ] Same as above but via keyboard
- [ ] Keyboard rearrangement (of any kind) is not possible in the Layers Panel currently, the user first needs to select the layer on canvas. It should be possible to rearrange when entering the Layers Panel via keyboard
- [ ] It should be possible to rearrange groups using <kbd>cmd</kbd>+<kbd>up</kbd> and <kbd>cmd</kbd>+<kbd>down</kbd>

## Additional Context
See https://github.com/GoogleForCreators/web-stories-wp/pull/11884#issuecomment-1180259836 for original discussion

PR #12066, opened 2022-08-03
Code: Layer panel refactor

### LINKED ISSUE #11576: Refactor layer panel

barklund:
## Context

This refactors the code in the layer panel.

Note that with this PR the following naming is attempted:

- Element - an element in the story
- Group - a group of elements in the story
- Layer - a row in the layer panel, which can represent either an element or a group.

So no more _layer=element_ stuff.

## To-do

- [x] Refactor layer list
- [x] Combine layer and group components and generalize where applicable
- [ ] Add tests where relevant

## User-facing changes

_None_

## Testing Instructions

1. Check that layer panel works as before.

## Checklist

---

Fixes #11576


packages/story-editor/src/components/panels/layer/constants.js
```diff
@@ -16,3 +16,5 @@
 
 export const LAYER_HEIGHT = 35;
 export const NESTED_PX = 30;
+
+export const nestedOffsetCalcFunc = (evt) => evt.offsetY < LAYER_HEIGHT / 1.7;
```
miina, line 20: **🥜 nitpick:**
Should this maybe go to an utils file since it doesn't seem to be a constant? 😄

packages/story-editor/src/components/panels/layer/useLayers.js
```diff
@@ -14,30 +14,107 @@
...
+            name: element.groupId,
+          };
+          return [groupLayer, elementLayer];
+        })
+        // Flatten the nested arrays
+        .flat()
+    );
+  }, [elementLayers]);
+
+  const handlePositionChange = useCallback(
```
miina: **❓question:**
I know it was brought over from another file and is unchanged, but does `handlePositionChange` here apply only when dragging a layer? Or could this be reapplied for keyboard events, too? Thinking that if it's just for dragging then would be good to specify the name for clarity (since we have #11906 for the keyboard handling).

miina (commented):
Generally looks great, nice refactor!

Looks like the count of layers is missing now (see the Percy screenshots for example, it shows the diff well, verified on local, too).

packages/story-editor/src/components/panels/layer/useLayers.js
barklund: Renamed it - it is specifically for dragging only to handle the partial overlap and figure out what the new position is going to be.

packages/story-editor/src/components/panels/layer/constants.js
barklund: Naa, that's too much trouble IMO. It's a constant function! 😂

packages/story-editor/src/components/footer/layers/layers.js
```diff
@@ -42,7 +42,9 @@ const StyledNavigationWrapper = styled(NavigationWrapper)`
 `;
 
 function Layers() {
-  const layersLength = useLayers().length;
+  const nunElements = useStory(
+    ({ state }) => state.currentPage?.elements?.length
+  );
```
barklund: @miina, well-spotted - I didn't realize this used the same hook before. Well, now that layers mean all the rows in the layer panel (both elements and groups), I feel it makes sense for this to just show the number of elements rather than the full number of layers, or?

packages/story-editor/src/components/footer/layers/layers.js
```diff
@@ -42,7 +42,9 @@ const StyledNavigationWrapper = styled(NavigationWrapper)`
 `;
 
 function Layers() {
-  const layersLength = useLayers().length;
+  const nunElements = useStory(
```
miina: Nun elements? 😆 

Edit:

```suggestion
  const numElements = useStory(
```

packages/story-editor/src/components/footer/layers/layers.js
miina: Makes sense to me!

miina (approved):
`nun` -> `num` and LGTM! ⛪

packages/story-editor/src/components/footer/layers/layers.js
barklund: 🙏 ⛪

barklund: (fixed in 974f60f161)

Issue #1402, opened 2020-04-30
Enable autosave for published stories


o-fernandez:
## Feature Description

Currently, per #221 we disable autosave for published stories. The proposed change here is: 

- When user makes changes after they open the published story, we autosave to a second file (doesn't overwrite published post) every 60 seconds (same interval). 
- User can push changes live by triggering "publish/update" button
- User can trigger "Discard changes/revert to published version" option if they want to ignore any changes made after the last publish

The proposed UX changes are (need finalization before ticket moves to eng queue): 

- Add a "discard changes" or "revert to published" button/option in the UI for published posts with changes
- May want some other way to show in WP (in the editor, stories dashboard, etc.) whether a published story has unpublished changes in draft. 
- We may want to change the buttons nomenclature, we currently closely follow the WP article/page editor and replace "save draft" with "switch to draft" for publish stories, but in this scenario that wouldn't make as much sense since users would still be able to keep a working draft. In that case we think that having the 3 main buttons continue to be "save as draft", "preview" and "Update/publish changes" could be good, and we'd want to move the "unpublish/switch to draft" option and add "discard changes/revert to published" to some "other options" menu (maybe a drop down arrow next to update/publish). 

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

swissspidy:
A brief summary of how autosaves/revisions work in WordPress:

(I recommend trying it yourself to get a better feeling for it)

1. Create a new post (as `post` post type)
2. The WIP post is periodically saved to both session storage and the database (if online)  
    * If the post is still a draft, this just means updating the currently edited post.
    * If the post is already published, this will create a new autosave (`revision` post type, referencing the `post`)
  Note: See #3842 for details on local autosaving
3. If someone else is taking over editing from you and locks you out of the editor (see #6055 for story locking), the same applies. Your changes will be saved so you don't lose anything.

Let's say you edit a post, save it, and then make some changes again.

If you later open the editor, you might get a message like this (depending on whether the changes were saved to session storage or the database:

![Screenshot 2021-12-07 at 23 08 25](https://user-images.githubusercontent.com/841956/145122489-43355406-1018-424d-a5cc-e7f6cce6a0dc.png)
![Screenshot 2021-12-07 at 22 11 55](https://user-images.githubusercontent.com/841956/145122502-490e59b3-e40c-4cdd-add5-45da117fe5f3.png)

[...]

PR #12096, opened 2022-08-11
Autosaves: Set up initial infrastructure 

### LINKED ISSUE #12094: Autosaves: Set up initial infrastructure
## Feature Description

There are few things we need to get going.

* Add feature flag (`improvedAutosaves` or something)
* In `AutoSaveHandler`, use `autoSave` instead of `saveStory` if flag is enabled

timarney:
## Context

* Add feature flag (`improvedAutosaves` or something)
* In `AutoSaveHandler`, use `autoSave` instead of `saveStory` if flag is enabled

## Testing Instructions

1. Visit experiments 
2. Note that an option for Improved Autosaves exists
3. Ensure saving a still works

no

no

no

## Checklist

---

Fixes #12094


packages/story-editor/src/components/autoSaveHandler/index.js
```diff
@@ -26,30 +27,34 @@ import { useConfig, useHistory, useStory } from '../../app';
...
       status,
+      autoSave,
       saveStory,
     })
   );
   const isUploading = useIsUploadingToStory();
 
   const isDraft = 'draft' === status || !status;
 
+  const save = improvedAutosaves ? autoSave : saveStory;
```
miina, line 51: Hm, would it still be `saveStory` for a draft always? Or is this handled in the backend?

timarney: Looks like there's a isDraft check here

https://github.com/GoogleForCreators/web-stories-wp/blob/339cb6598fc1b3d3b0b99b33c373c159c8f8f1bd/packages/story-editor/src/components/autoSaveHandler/index.js#L60

I haven't dug into the full auto save flow yet

miina: As far as I understand, for drafts, the story itself gets overwritten by the autosave (`saveStory`), however, in case of published stories, an autosave (`revision`) gets created instead and the story itself will not get overwritten. So I'm wondering if it should be `autoSave` only if `improvedAutoSaves && !isDraft`( That's actually something that could also be done in #1402 )?

Shouldn't this PR also include the backend logic for making sure `autoSave` is creating a revision with all the necessary data? Or is the backend logic already in place for that when `autoSave` is used here?

swissspidy: I don't think we need any backend changes here. We already have a `Stories_Autosaves_Controller` and WordPress chooses between autosaves and revisions automatically depending on if it's a draft or not:

https://github.com/WordPress/wordpress-develop/blob/37368555a5251e49b748bf91407a24fad6d32e9a/src/wp-includes/rest-api/endpoints/class-wp-rest-autosaves-controller.php#L223-L230

But should be easy to verify :-)

timarney: > In AutoSaveHandler, use autoSave instead of saveStory if flag is enabled

Seeing the autosave call 

Screen Shot 2022-08-17 at 5 18 06 PM

miina: Alright, I wasn't sure that autosaving already saves all the data we need for stories, but that's great if it does :)

swissspidy (approved):

miina (approved):

Issue #1715, opened 2020-05-11
Adopting TypeScript


swissspidy:
## Feature Description

Possibly adopting TypeScript for our code base has been brought up a few times in the past.

A step in that direction was made at some point by switching our JSDoc parsing to use TypeScript, allowing for more expressive type documentation. See #2960.

That's a pretty common thing to do and is something Gutenberg has been doing as well, for example. Gutenberg even creates `d.ts` files based on the JSDoc annotations. That's neat because it improves DX without actually having to write TypeScript code. That path has its limits, however.

Actually authoring code in TypeScript is of course a whole different topic. 

My proposal is not to just convert our hundreds of thousands of lines of code to TypeScript. However, TypeScript could be _selectively_ used for some _parts_ of the code base.

In particular, I think it would be really helpful for _low-level code_ that does not require frequent maintenance. For instance:

* Media-related code (`Resource`, `createResource`)
* Code related to the story/elements data structure, which is the heart of the story editor after all
* Small packages like `tracking`, `i18n` or `url` that change _rarely_ but are used _often_

Some considerations off the top of my head:

* Parsing TypeScript can be slower. We need to ensure that builds and tests are not significantly slower.
* Not everyone on the team is familiar with TypeScript

Other benefits:

* More alignment with Google-internal tooling.

## Overview

<table>
<tr>
	<td><b>Package
	<td><b>Status
	<td><b>Issue / PR
<tr>
	<td><code>date
	<td> ✅
	<td>#12224
<tr>
	<td><code>dom
	<td> ✅
	<td>#12218
<tr>
	<td><code>i18n
	<td> ✅
	<td>#12198
<tr>
	<td><code>media
	<td> ✅
	<td>#12127
<tr>
	<td><code>moveable
	<td>✅
	<td>#12169
<tr>
	<td><code>patterns
	<td> ✅
	<td>#12200
<tr>
	<td><code>react
	<td> ✅
	<td>#12127
<tr>
	<td><code>stickers
	<td>✅
	<td>#12204
<tr>
	<td><code>transform
	<td> ✅ 
	<td> #12202
<tr>
	<td><code>tracking
	<td> ✅ 
	<td> #12085 / #10783
<tr>
	<td><code>units
	<td> ✅
	<td>#12127
<tr>
	<td><code>url
	<td> ✅
	<td>#12212
</table>

## Additional Context

Gutenberg went through a similar process. These posts are a great resource on that:

[...]

PR #12127, opened 2022-08-17
TypeScript Conversion: media, units, react packages

### LINKED ISSUE #12087: TypeScript: Convert `media` package
There's already some POC here that can be leveraged.

The main benefit is the typing around the `Resource` type that we use all over the editor.

But there are also half-baked resource-like objects for things like background audio and captions that are worth creating proper types for.

swissspidy:
## Context

This is part of our ongoing effort to convert (parts of) our codebase to TypeScript. See #1715 for details.

## Summary

This marks the first step of migrating the `media` package to TypeScript, unblocking better typing usage in all consumer packages relying on it.

Due to my struggles with `Could not find a declaration file for module ...` errors, I also ended up converting the `units` and `react` packages (which are dependencies) to React.

Figured I'll keep everything in this one PR since it seemed so intertwined.

Success story:

* Important bug uncovered thanks to TypeScript, see https://github.com/GoogleForCreators/web-stories-wp/pull/12127#discussion_r949984838

## Relevant Technical Choices

* Added custom typings for the `shallow-equal` module
* Added custom typings for globals like `WEB_STORIES_ENV`
* Remove custom `useReduction` fork in favor of the `use-reduction` package directly.
  Adapts Jest config to transpile `use-reduction` package because it's using ESM
* Removed `ResizeObserver` polyfill. Ran into https://github.com/que-etc/resize-observer-polyfill/issues/83 and figured we probably don't need this anymore anyway due to improved browser support.

## To-do

* It would be nice to eventually import a better `Element` type from a shared package (`elements` package?)
* Be able to move custom typings to the top-level `typings` folder
* Figure out better solution to address `Could not find a declaration file for module ...` errors so we can better convert packages one at a time.
* Figure out better typings for `UnitsProvider` and `useUnits` and `useContextSelector`
* Add some documentation for contributors

## User-facing changes

N/A

## Testing Instructions

1.

No

No

No

## Checklist

---

Fixes #12087


packages/media/src/getCanvasBlob.ts
```diff
@@ -14,11 +14,11 @@
...
  */
 
-function getCanvasBlob(canvasEl) {
+function getCanvasBlob(canvasEl: HTMLCanvasElement): Promise<Blob | null> {
   return new Promise(function (resolve) {
     canvasEl.toBlob((blob) => {
-      resolve(blob, 'image/jpeg');
-    });
+      resolve(blob);
+    }, 'image/jpeg');
```
swissspidy, line 21: We've never spotted this bug before

packages/media/src/types.ts
```diff
@@ -200,6 +201,161 @@ export { ResourcePropTypes };
...
+  displayName: string;
+  /** The author's profile or website. */
+  url: string;
+}
+
+export interface Attribution {
+  /** The optional author of the media object. */
+  author?: AttributionAuthor;
+  /** The optional url to register the media usage. */
+  registerUsageUrl: string;
```
timarney: Need to make this optional ya?

registerUsageUrl?: string;

timarney:
Looking really solid

packages/media/src/types.ts
swissspidy: Yup indeed!

Now wondering whether `author` and `registerUsageUrl` should be split up into its own interface...

packages/media/src/getResourceSize.ts
```diff
@@ -19,6 +19,14 @@
  */
 import { FULLBLEED_HEIGHT, PAGE_WIDTH } from '@googleforcreators/units';
```
swissspidy, line 20: So this `Could not find a declaration file for module '@googleforcreators/units'.` error has been nagging me for a while now and can't figure out why it's not working 🤔

packages/units/src/dimensions.ts
```diff
@@ -152,7 +184,11 @@ export function getBox(
  * @return {{x:number, y:number, width:number, height:number, rotationAngle:number}} The
  * "box" in the editor space.
  */
-export function getBoxWithBorder(element, pageWidth, pageHeight) {
+export function getBoxWithBorder(
+  element: Element,
+  pageWidth: number,
+  pageHeight: number
+) {
```
miina: **💬 suggestion:**

This should probably also return `ElementBox`?

packages/units/src/dimensions.ts
```diff
@@ -109,14 +113,42 @@ export function editorToDataX(x, pageWidth, withRounding = true) {
  * @param {boolean} [withRounding=true] Whether the dataPixels rounding should occur.
  * @return {number} The value in the "data" space.
  */
-export function editorToDataY(y, pageHeight, withRounding = true) {
+export function editorToDataY(
+  y: number,
+  pageHeight: number,
+  withRounding = true
+) {
```
miina: Here, too, return type?

packages/media/src/types.ts
```diff
@@ -200,6 +201,161 @@ export { ResourcePropTypes };
...
+  /** BlurHash. */
+  blurHash: string;
+  /** Whether the resource externally hosted. */
+  isExternal: boolean;
+  /** Whether the resource is a placeholder. */
+  isPlaceholder: boolean;
+  /** Whether the resource needs a CORS proxy. */
+  needsProxy: boolean;
+  /** Resource creation date. */
+  creationDate: string;
```
miina: **💡 thought:**
Should we also use `readonly` properties? E.g. `creationDate` sounds like one that shouldn't be rewritten.

(Edit: I'm seeing now elsewhere occasional usage of `readonly` as well)

packages/media/src/types.ts
```diff
@@ -200,6 +201,161 @@ export { ResourcePropTypes };
...
+  /** The resource's poster ID. */
+  posterId?: string;
+  /** Length in seconds. */
+  length: number;
+  /** The formatted length, e.g. "01:17". */
+  lengthFormatted: string;
+  /** Whether the resource has already been optimized. */
+  isOptimized: boolean;
+  /** Whether the resource is muted. */
+  isMuted: boolean;
```
miina: **❓question:**
Wouldn't `isMuted` and other video-only props also be optional? 🤔

packages/media/src/hasVideoGotAudio.ts
```diff
@@ -14,15 +14,28 @@
...
  * Determines whether a video element has audio tracks.
  *
  * @param {HTMLVideoElement} video Video element.
  * @return {boolean} Whether the video has audio or not.
  */
-function hasVideoGotAudio(video) {
+function hasVideoGotAudio(video: HTMLVideoElement): boolean {
   return (
-    video.mozHasAudio ||
+    (video.mozHasAudio as boolean) ||
```
miina: **❓question:**
Does this need `as boolean` even though this prop's type is configured as `boolean` above? I thought it would be required only if TypeScript wouldn't know what the type is. Thoughts?

packages/media/src/getTypeFromMime.ts
```diff
@@ -14,14 +14,19 @@
...
- * @param {string|null} mimeType Mime type.
- * @return {string|null} Element type.
+ * @param {string} mimeType Mime type.
+ * @return {string} Element type.
  */
-const getTypeFromMime = (mimeType) => {
-  return mimeType?.split('/')[0] || null;
-};
+function getTypeFromMime(mimeType: string): ResourceType {
+  return mimeType?.split('/')[0] as ResourceType;
```
miina: **💬 suggestion:**

Maybe remove the `?` as well if the string is required?

packages/media/src/getCanvasBlob.ts
miina, line 21: **👏 praise:**

Nice catch!

miina (commented):
Looking great!

packages/units/src/dimensions.ts
swissspidy: Yes, but the return type is inferred from the function body itself, so it's technically redundant.

packages/media/src/hasVideoGotAudio.ts
swissspidy: Indeed

packages/media/src/types.ts
swissspidy: Probably yes, but that's easier to decide when going through the media-related code in the editor to see which ones can actually be readonly.

packages/media/src/getCanvasBlob.ts
swissspidy: Looks like this actually fixes #12132 too!

packages/units/src/getBoundRect.ts
```diff
@@ -130,7 +143,13 @@ export function getCorners(angle, x, y, width, height) {
   };
 }
 
-function getCorner(pivotX, pivotY, cornerX, cornerY, angle) {
```
miina, line 133: **💬 suggestion:**
Looks like the utils in this file don't have return types, maybe add? Or is it purposefully so?

packages/react/src/useContextSelector.ts
```diff
@@ -53,8 +61,7 @@ function useContextSelector(context, selector, equalityFn = shallowEqual) {
   };
 
   // Update the selector fn to memoize the selected value by [equalityFn].
-  const patchedSelector = equalityFn ? equalityFnCallback : selector;
```
miina, line 56: **💡 thought:**

Hm, if it would be called with `equalityFn = null`, would it fail now?

packages/media/src/types.ts
```diff
@@ -200,6 +201,166 @@ export { ResourcePropTypes };
...
+  /** Resource sizes */
+  sizes: { [key: string]: ResourceSize };
+  /** Resource author a ttribution */
+  attribution?: Attribution;
+  /** The resource's poster. */
+  poster?: string;
+  /** The resource's poster ID. */
+  posterId?: string;
+  /** Length in seconds. */
+  length: number;
```
miina: **🥜 nitpick:**

I suppose `length` and `lengthFormatted` should also be optional since images don't have this?

packages/media/src/getResourceSize.ts
miina: Curious, what was the issue?

miina (approved):
Looks good assuming the failing Karma tests are just flaky (re-running). Lots of files to go through so would be great if someone else would review as well.

packages/media/src/getResourceSize.ts
swissspidy: That's the thing, I couldn't figure it out properly. Probably just some tsconfig misconfiguration somewhere, basically it did not like importing from this non-TS package. That's why I ended up just converting the `units` package to TypeScript 🙃 

For next time it would be great to resolve though. If you have any ideas, PLMK

packages/units/src/getBoundRect.ts
swissspidy: They're implicit.

Right now we do not enforce explicit return types, though [there's a lint rule for that](https://typescript-eslint.io/rules/explicit-function-return-type/).

packages/react/src/useContextSelector.ts
swissspidy: Theoretically yes, though you'd probably get a type error. Plus I couldn't find any usage like that in our codebase.

timarney (approved):

PR #12129, opened 2022-08-18
AutoSaving: offline local autosaving (sessionStorage)

### LINKED ISSUE #3842: Autosaving while offline (sessionStorage)
## Feature Description

When the user is editing a story while being offline, can we store autosaves in sessionStorage like Gutenberg does?

Then we could display messages like "The backup of this post in your browser is different from the version below. Restore backup"

## Additional Context

Related: #3841 (Add message when Autosave exists)

Original Gutenberg implementation: https://github.com/WordPress/gutenberg/pull/16490

Code is currently here: https://github.com/WordPress/gutenberg/blob/7d880708c6281c185256b8e7d3dffe178da33d09/packages/editor/src/components/local-autosave-monitor/README.md

---

miina:
## Summary
This PR implements local autosaving using sessionStorage. Every X seconds, a local autosave is created. When the story is loaded, there's a check for local autosave, and if it exists, the user is prompted to restore it (or dismiss it).

## To-do

- [x] Check how to configure local autosave interval
- [x] Tests

## User-facing changes
Screenshot 2022-08-18 at 11 22 21

## Testing Instructions

1. Make sure autosaving locally is configured before testing (Default interval is 15 seconds, no need to configure separately)
2. Add a new story, make some changes and save
3. Now make a new change and wait for 15 seconds (or whatever the autosaving interval is configure to be)
4. Reload the story without saving
5. Now a dialogue with prompting to restore the local backup should pop up.
6. Verify that clicking "Restore backup" actually restores the previous unsaved change from step 3. 
7. Verify that clicking "Dismiss" does nothing.
8. Verify that reloading the story after Restoring or Dismissing doesn't show the dialogue anymore (the local autosave should have been deleted after either of the action.
9. Now make changes again and wait for 15 seconds (or whatever the autosaving interval is configured to be)
10. Save the story and reload
11. Verify that no dialogue is displayed (since the saved state is the same as the local baclup.

Adds tracking for error of saving / getting / deleting sessionStorage data.

## Checklist

---

Fixes #3842


includes/Admin/Editor.php
```diff
@@ -370,6 +370,7 @@ public function get_editor_settings(): array {
 
 		$settings = [
 			'autoSaveInterval'        => \defined( 'AUTOSAVE_INTERVAL' ) ? AUTOSAVE_INTERVAL : null,
+			'localAutoSaveInterval'   => \defined( 'LOCAL_AUTOSAVE_INTERVAL' ) ? LOCAL_AUTOSAVE_INTERVAL : null,
```
miina: @swissspidy Any thoughts on how it would be the best to define the local autosave interval? Added it here for now but maybe it should come from somewhere else, e.g. defined when initing the app. Thoughts?

packages/story-editor/src/app/story/effects/useLoadStory.js
```diff
@@ -99,6 +99,8 @@ function loadStory(storyId, post, restore, clearHistory) {
     storyId: storyId,
     title,
     status: statusFormat,
+    // @todo Temporary, check for an alternative.
+    isNew: status === 'auto-draft',
```
miina, line 103: In Gutenberg, the check for `isNew` happens by checking the status for `auto-draft`, in our editor, however, `auto-draft` is overwritten to `draft`. I suppose it makes it easier to just handle the `draft` status in the code, however, now it's not possible to check if the post is new by that. Added `isNew` here for now, wondering if there's a better alternative 🤔

includes/Admin/Editor.php
swissspidy: I'd just hardcode it to a number. WP does that as well.

It can be changed via the `web_stories_editor_settings` filter below.

```suggestion
			'localAutoSaveInterval'   => 15,
```

packages/story-editor/src/utils/localAutoSave.js
```diff
@@ -0,0 +1,37 @@
...
+    JSON.stringify({
+      story,
+      pages,
+    })
+  );
+}
+
+export function deleteLocalAutosave(storyId, isNew) {
+  window.sessionStorage.removeItem(storyKey(storyId, isNew));
+}
```
swissspidy: Worth doing this similar to our `localStore` API in `packages/design-system/src/utils/localStore.js`?

Note the error handling & error tracking.

miina: Moved to design-system in [6959b94](https://github.com/GoogleForCreators/web-stories-wp/pull/12129/commits/6959b94307f65d45164614ca92c960c7ab154c50), followed the same structure for consistency, assuming it might be used for more than one case.

packages/story-editor/src/components/localAutoSaveHandler/localAutoSave.js
```diff
@@ -0,0 +1,169 @@
...
+  }, [setBackup, isNew, storyId]);
+
+  if (!backup) {
+    return null;
+  }
+
+  const restoreBackup = () => {
+    restoreLocalAutoSave();
+    sessionStore.deleteItemByKey(getSessionStorageKey(storyId, isNew));
+    setBackup(false);
```
barklund: `false`? Is `backup` a boolean flag or the actual backup value? If the latter, shouldn't this be `null`?

packages/story-editor/src/components/localAutoSaveHandler/localAutoSave.js
```diff
@@ -0,0 +1,169 @@
...
+    <Dialog
+      isOpen
+      secondaryText={__('Dismiss', 'web-stories')}
+      onClose={onClose}
+      actions={
+        <>
+          <Button
+            type={BUTTON_TYPES.TERTIARY}
+            size={BUTTON_SIZES.SMALL}
+            onClick={() => onClose()}
```
barklund: Surely, this can just be `onClick={onClose}`? Unless this is a best-practice thing about not allowing event object to bleed through?

(similarly below for `restoreBackup`)

packages/design-system/src/utils/sessionStore.js
```diff
@@ -0,0 +1,58 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * External dependencies
+ */
+import { trackError } from '@googleforcreators/tracking';
+
+export const SESSION_STORAGE_PREFIX = {
+  LOCAL_AUTOSAAVE_PREFIX: 'wp_stories_autosave_story',
```
barklund:
```suggestion
  LOCAL_AUTOSAVE_PREFIX: 'wp_stories_autosave_story',
```

packages/design-system/src/index.js
```diff
@@ -23,12 +23,14 @@ import Modal from 'react-modal';
...
  */
 import localStore, { LOCAL_STORAGE_PREFIX } from './utils/localStore';
+import sessionStore, { SESSION_STORAGE_PREFIX } from './utils/sessionStore';
 import * as Icons from './icons';
 
 const { setAppElement } = Modal;
 
 export { Icons };
 export { localStore, LOCAL_STORAGE_PREFIX };
+export { sessionStore, SESSION_STORAGE_PREFIX };
```
barklund, line 33: Any particular reason this uses `sessionStorage` rather than `localStorage`? Is it a security concern? Ping @swissspidy

barklund (approved):
Looks good and works great.

This does not have any actual detection of offline vs online nor any information about "you are now online again, do you want to save your story", but I assume that's also not part of this? Is this something we want to do?

packages/story-editor/src/components/localAutoSaveHandler/localAutoSave.js
miina: > Unless this is a best-practice thing about not allowing event object to bleed through?

Nope, this was a leftover from previously some logic being inline.

packages/design-system/src/index.js
miina: Probably because there's no reason for keeping the backup persistently since this is mainly thought for unexpected happenings and becomes irrelevant.

swissspidy: It matches WordPress behavior and also prevents leaking data on a shared computer for example

swissspidy:
> This does not have any actual detection of offline vs online nor any information about "you are now online again, do you want to save your story", but I assume that's also not part of this? Is this something we want to do?

Correct. And maybe in the future, can be its own ticket

packages/story-editor/src/app/story/effects/useLoadStory.js
swissspidy: > in our editor, however, `auto-draft` is overwritten to `draft`

Can you remind me where this happens?

I don't think that's right and might be why the "All Stories" page is easily flooded with draft stories instead of auto-draft ones (which will cleaned up by WordPress automatically):

Screenshot 2022-08-22 at 16 43 24

miina: Hm, could be. Not sure when it was added exactly:
https://github.com/GoogleForCreators/web-stories-wp/blob/96f067b0283184f672fbf0268b13285f945a2918/packages/story-editor/src/app/story/effects/useLoadStory.js#L76

swissspidy: Apparently in https://github.com/ampproject/amp-wp/pull/4122, which is a looong time ago. Unfortunately I don't see why it was added nor do I recall the reason.

It'd be great if we could get rid of that. Maybe in a fast follow up PR?

miina: Sure!

PR #5889, opened 2021-01-12
Page Layouts: Adds page type based naming and filtering

zachhale:
## Summary

Changes Page Layouts filter pills to being based on page type instead of template name. All templates have been categorized by the types specified in the [figma](https://www.figma.com/file/NHHMwIfxMnz39NxqkrFb7R/Stories---Post-Stable?node-id=2%3A98474).

This also fixes page names when filtering and changes titles/hover text to be the page name instead of just "Page layouts".

## Relevant Technical Choices

* Karma tests are still pending for page layouts so I was unable to update them for this change in functionality, but will add karma tests for these in that work.

## Testing Instructions

Filter page layouts by each of the pills and verify that all items in each filter match the ones specified in [figma](https://www.figma.com/file/NHHMwIfxMnz39NxqkrFb7R/Stories---Post-Stable?node-id=2%3A98474)

---

Fixes #5872

PR #12131, opened 2022-08-19
Create initial story JSON schema

### LINKED ISSUE #12084: Document data model as JSON schema
This is basically #11323 but for [JSON schema](https://json-schema.org/)

Documenting our data model this way would make maintenance easier and increase DX due to the added tooling support.

For example, we could make all our templates and text sets reference this schema.

It's also possible to automatically [create TypeScript typings from JSON schema](https://www.npmjs.com/package/json-schema-to-typescript), which might be useful as well. Though we probably want to create those manually for now since everything is split up across multiple packages.

swissspidy:
## Context

See #12084

## Summary

Found some bugs thanks to this:

343f3551b38211e833de210901a3ab8f3d26656c

## Relevant Technical Choices

* Storing schema in `bin/` for now
* Using `ajv-cli` for JSON schema validation
* USing `ajv-formats` to add support for things like `uuid` and `date-time` types

## To-do

* Maybe add data migrations to remove `backgroundOverlay` and `fontWeight` from existing stories.
  Not sure how likely these exists on existing stories though.
* Stickers: avoid properties like `showInLibrary` from ending up in the JSON data. Add a data migration accordingly & remove from schema.
* Schema: Add schema for resource sizes
* Data structure: consider storing all existing fonts at the story level instead of on each element to reduce redundancy
* Pages: pages have `type: page` and we register a special "page" element type. Is that still relevant or outdated?
* Story: Do not persistently store global properties things like `current`, `selection`, and `story`
* Elements: Do not persistently store the `basedOn` property. Seems useless to keep, since this is only used when duplicating elements.

## User-facing changes

N/A

## Testing Instructions

1.

No

No

No

## Checklist

---

Fixes #12084


bin/schemas/story.json
```diff
@@ -0,0 +1,839 @@
...
+          "default": "",
+          "title": "The zoomDirection Schema",
+          "examples": ["scaleIn"]
+        }
+      }
+    },
+    "element": {
+      "type": "object",
+      "title": "Base element",
+      "required": ["x", "y", "width", "height", "type", "id"],
```
miina: `rotationAngle` should also be required, I believe.


**❓question:**

There are probably more props that unless set, would cause errors, e.g. `isBackground`, should those be considered as well?

Edit: Although this specific one would cause errors in case of media only actually.

bin/schemas/story.json
```diff
@@ -0,0 +1,839 @@
...
+        },
+        "x": {
+          "type": "number",
+          "title": "X position"
+        },
+        "y": {
+          "type": "number",
+          "title": "Y position"
+        },
+        "width": {
```
miina, line 161: **💬 suggestion:**

Maybe this should have a min value, and height, too? At least in the editor I believe we don't allow either of these below 1.

bin/schemas/story.json
```diff
@@ -0,0 +1,839 @@
...
+            "supportsBorder": {
+              "type": "boolean",
+              "$comment": "TODO: Remove / do not store"
+            }
+          }
+        },
+        "type": {
+          "type": "string",
+          "title": "The element type",
+          "examples": ["image", "sticker", "text", "video", "shape"]
```
miina: **💬 suggestion:**

Maybe add `gif` since you're listing all as examples?

bin/schemas/story.json
```diff
@@ -0,0 +1,839 @@
...
+          "type": "number",
+          "title": "Y axis focal point",
+          "minimum": 0
+        },
+        "padding": {
+          "type": "object",
+          "title": "The padding Schema",
+          "required": [],
+          "properties": {
+            "horizontal": {
```
miina: **💬 suggestion:**
Should these have min values?

bin/schemas/story.json
```diff
@@ -0,0 +1,839 @@
...
+        },
+        "isBackground": {
+          "type": "boolean",
+          "title": "Whether it is the background element"
+        },
+        "isDefaultBackground": {
+          "type": "boolean",
+          "title": "Whether it is the default background"
+        },
+        "groupId": {
```
miina, line 308: **❓question:**

I see you're adding examples in some places but not everywhere. Curious in which cases are you adding examples and in which not? Should all props have examples if some have? 

Same question about min values, e.g. `rotationAngle` has min-max values defined here but padding doesn't have the min value, even though we don't allow it below 0 in the editor, as far as I know. Thoughts?

bin/schemas/story.json
```diff
@@ -0,0 +1,839 @@
...
+      "type": "object",
+      "oneOf": [
+        { "$ref": "#/$defs/solidPattern" },
+        { "$ref": "#/$defs/linearPattern" }
+      ]
+    },
+    "solidPattern": {
+      "type": "object",
+      "title": "Solid pattern",
+      "required": ["color"],
```
miina, line 682: **💬 suggestion:**

"type" is required here, too, AFAIK.

bin/schemas/story.json
```diff
@@ -0,0 +1,839 @@
...
+        "elements": {
+          "type": "array",
+          "minItems": 1,
+          "title": "The page's elements",
+          "items": {
+            "anyOf": [
+              { "$ref": "#/$defs/element" },
+              { "$ref": "#/$defs/mediaElement" },
+              { "$ref": "#/$defs/shapeElement" },
+              { "$ref": "#/$defs/stickerElement" }
```
miina: **💬 suggestion:**

Add textElement, too?

bin/schemas/story.json
```diff
@@ -0,0 +1,839 @@
...
+        "backgroundColor": {
+          "type": "object",
+          "title": "Background color",
+          "$ref": "#/$defs/pattern"
+        },
+        "defaultBackgroundElement": {
+          "type": "object",
+          "title": "The default background element",
+          "oneOf": [
+            { "$ref": "#/$defs/mediaElement" },
```
miina: **📝 note:**

The default background element can only be a shape, if it's not a shape, it's not the default background element.

packages/migration/src/migrate.js
```diff
@@ -129,4 +129,21 @@ export function migrate(storyData, version) {
   return result;
 }
 
+export function migrateSingle(storyData, version) {
```
miina: **💬 suggestion:**

Would be nice to have a test for this, too.

bin/schemas/story.json
swissspidy: Minimum 0 works yeah 👍

bin/schemas/story.json
swissspidy: I auto-generated the initial schema from an existing template JSON, which automatically added tons of examples everywhere. I actually want to remove them from most places again unless they are really useful.

As for min/max, just need to go through all the fields and add where it makes sense.

bin/schemas/story.json
swissspidy: In the templates most text elements have bg color defined like this:

`"backgroundColor": { "color": { "r": 196, "g": 196, "b": 196 } },`

Without `type`.

Is that a different type of pattern? Otherwise I guess it's optional.

bin/schemas/story.json
swissspidy: Good catch!

And should probably be an `enum` anyway,

miina: Unless someone registers a new element type in the future? I think I remember discussion about it somewhere related to creating all the separate packages 🤔

bin/schemas/story.json
miina: Ah, you're right, looks like 'solid' is the default if not set.

bin/schemas/story.json
swissspidy: Yup that will/should be possible in the future.

If we make the schema public this becomes relevant and then we'd need to switch to something more extensible here.

For now it adds a little bit more safety for our own use case :)

bin/schemas/story.json
swissspidy: I should list that property though for `solidPattern`, so good call

packages/migration/src/migrate.js
swissspidy: Currently actually pondering how useful this new util is.

timarney (approved):

packages/templates/src/raw/12-hours-in-barcelona/template.json
```diff
@@ -373,39 +289,21 @@
           "groupId": "68836b5c-a442-4407-a938-a790f564d9de"
         }
       ],
-      "backgroundOverlay": "none",
```
swissspidy, line 376: Here's an example of the lingering `backgroundOverlay` property.
`v0023_convertOverlayPattern.js` should have been taking care of this.

packages/templates/src/raw/12-hours-in-barcelona/template.json
```diff
@@ -344,27 +271,16 @@
...
           "rotationAngle": 0,
           "lockAspectRatio": true,
-          "backgroundColor": {
-            "color": {
-              "r": 255,
-              "g": 199,
-              "b": 0
-            }
-          },
+          "backgroundColor": { "color": { "r": 255, "g": 199, "b": 0 } },
```
swissspidy: No idea why the formatting here suddenly changed... But Prettier thinks it's fine like this.

swissspidy:
@miina @barklund @merapi Would appreciate feedback here, especially on the thoughts in the to-do section above.

miina:
@swissspidy On not storing the said values on elements (e.g. `basedOn` etc.): makes sense, I've also thought about it quite a few times when adding some values that are needed only temporarily. Did you have anything in mind for it? Perhaps we could create a hardcoded list of values that should not get saved, e.g. something similar to how some values don't create a new history entry with `ELEMENT_PROPS_TO_IGNORE`. 

> Pages: pages have type: page and we register a special "page" element type. Is that still relevant or outdated?

Having a page element type is still relevant as it currently is. The same utils are used for creating a page as for other elements, it has its default element attributes etc. Not sure if storing `type: 'page'` is really needed for it but the general logic for creating a page depends on it being handled as an element.

bin/schemas/story.json
```diff
@@ -0,0 +1,846 @@
...
+          "minimum": 0,
+          "maximum": 100,
+          "default": 100,
+          "title": "Element opacity",
+          "$comment": "TODO: Set by default and make required"
+        },
+        "flip": {
+          "type": "object",
+          "title": "Flip setting",
+          "$comment": "TODO: Set by default and make required",
```
miina: **📝 note:**
Text elements can't be flipped though.

bin/schemas/story.json
```diff
@@ -0,0 +1,846 @@
...
+        },
+        "marginOffset": {
+          "type": "number",
+          "title": "Margin offset"
+        },
+        "basedOn": {
+          "type": "string",
+          "default": "",
+          "title": "Reference to the original element if this one was duplicated/copied",
+          "$comment": "TODO: Remove?"
```
miina, line 301: Yep!

bin/schemas/story.json
```diff
@@ -0,0 +1,846 @@
...
+        },
+        "isBackground": {
+          "type": "boolean",
+          "title": "Whether it is the background element"
+        },
+        "isDefaultBackground": {
```
miina: Hm, only a shape supports this property, I think.

bin/schemas/story.json
```diff
@@ -0,0 +1,846 @@
...
+          "type": "number",
+          "title": "Margin offset"
+        },
+        "basedOn": {
+          "type": "string",
+          "default": "",
+          "title": "Reference to the original element if this one was duplicated/copied",
+          "$comment": "TODO: Remove?"
+        },
+        "isBackground": {
```
miina, line 303: This is never assigned to a text element, AFAIK.

bin/schemas/story.json
```diff
@@ -0,0 +1,846 @@
...
+              "minimum": 0
+            },
+            "bottomLeft": {
+              "type": "integer",
+              "title": "Bottom left radius",
+              "minimum": 0
+            }
+          }
+        },
+        "marginOffset": {
```
miina: This is only used for a text element (related to fonts).

bin/schemas/story.json
```diff
@@ -0,0 +1,846 @@
...
+          "type": "number",
+          "title": "X axis focal point",
+          "minimum": 0
+        },
+        "focalY": {
+          "type": "number",
+          "title": "Y axis focal point",
+          "minimum": 0
+        },
+        "padding": {
```
miina: Hm, this is also used only for texts, I believe.

bin/schemas/story.json
```diff
@@ -0,0 +1,846 @@
...
+          "type": "number",
+          "title": "Scaling",
+          "minimum": 0
+        },
+        "focalX": {
+          "type": "number",
+          "title": "X axis focal point",
+          "minimum": 0
+        },
+        "focalY": {
```
miina: `scale`, `focalX`, `focalY` are used by media only currently.

bin/schemas/story.json
```diff
@@ -0,0 +1,846 @@
...
+        },
+        "alpha": {
+          "type": "number",
+          "minimum": 0,
+          "maximum": 1
+        },
+        "rotation": {
+          "type": "number",
+          "minimum": -360,
+          "maximum": 360
```
miina: I think we're restricting it from -359 to 359 instead in the editor.

bin/schemas/story.json
```diff
@@ -0,0 +1,846 @@
...
+        }
+      }
+    },
+    "pageTemplateType": {
+      "type": "string",
+      "title": "The pageTemplateType Schema",
+      "enum": [
+        "",
```
miina: Hm, is `""` for the custom page templates?

bin/schemas/story.json
```diff
@@ -0,0 +1,846 @@
...
+        "backgroundColor": {
+          "type": "object",
+          "title": "Background color",
+          "$ref": "#/$defs/pattern"
+        },
+        "defaultBackgroundElement": {
+          "type": "object",
+          "title": "The default background element",
+          "$ref": "#/$defs/shapeElement",
+          "$comment": "TODO: Move to story element or remove?"
```
miina, line 796: Why remove? :)

miina:
> Having a page element type is still relevant as it currently is. The same utils are used for creating a page as for other elements, it has its default element attributes etc. Not sure if storing type: 'page' is really needed for it but the general logic for creating a page depends on it being handled as an element.

I'm checking now that the `defaultAttributes` the page has is actually just empty 😄 Thought for some reason that it included `elements` array as well but looks like not.

bin/schemas/story.json
swissspidy: That's good to know. I couldn't find a place where this is done.

miina: https://github.com/GoogleForCreators/web-stories-wp/blob/7ffe80de4a77b34d786cd4907b725688fcbc2970/packages/story-editor/src/components/panels/design/sizePosition/usePresubmitHandlers.js#L86

bin/schemas/story.json
swissspidy: No, for the default templates. The template pages that shouldn't be listed as page templates currently have empty strings as the `pageTemplateType`. There's also a Jest test that enforcing that all templates have this property. See #5889 and discussion on that PR.

I think we can change this from `""` to `null` though.

bin/schemas/story.json
swissspidy: I was wondering whether it makes sense to store the default background element in the story data JSON in the first place, since this could in theory just be a const somewhere in the editor since it's static and never changes.

From what I can see we mainly need this information about the default background element when removing the background element (e.g. an image or video) and need to restore the default.

We can still have the `isDefaultBackground` property to identify the default in the story JSON, should we ever need a data migration for it.

Does that make sense?

swissspidy:
> On not storing the said values on elements (e.g. `basedOn` etc.): makes sense, I've also thought about it quite a few times when adding some values that are needed only temporarily. Did you have anything in mind for it? Perhaps we could create a hardcoded list of values that should not get saved, e.g. something similar to how some values don't create a new history entry with `ELEMENT_PROPS_TO_IGNORE`.

I didn't have anything in mind, but I can create a ticket to further discuss this idea 👍 

The risk with a hardcoded list is that we forget to maintain it, but I think that becomes easier once we have strong TypeScript types for elements. Then one will get yelled at when adding a new unknown property anyway :-)

bin/schemas/story.json
swissspidy: Aside: while we need to store `pageTemplateType` for the default templates, there's no need to retain that when actually inserting the template in the editor.

swissspidy: Removing from schema now.

swissspidy:
Follow-ups:

* #12258 
* #12259
* #12260 
* #12261 
* #12262

PR #12156, opened 2022-08-23
Media: Video Cropping Prototype

### LINKED ISSUE #12093: Media: Video Cropping Prototype
As per #5985, let's see if we can implement video cropping to cut off-screen parts of videos

Example use cases:

* A video element is partially off canvas
* Video is scaled/zoomed in the editor

For this prototype we can create a new feature flag and some simple button somewhere (maybe in the quick actions menu or the right click menu; doesn't matter much for the prototype) that triggers this action.

Then, I'd imagine we need to do some calculation of the video position and appearance to see which parts need to be cropped, then pass that on to FFmpeg. After cropping, the video won't have the same dimensions anymore, so we'd need to update it accordingly on the canvas.

Accommodating for scaled/zoomed videos might be tricky, so perhaps we can focus on the off-canvas use case for now.

`useProcessMedia` is a good fit for the cropping logic I guess.

timarney:
## Context

As per https://github.com/GoogleForCreators/web-stories-wp/issues/5985,  implement video cropping to cut off-screen parts of videos

## Summary

Adds cropping for off canvas top, right, bottom, left i.e. hidden or off-canvas video

https://www.youtube.com/watch?v=P9Co7r7Droo

## Testing Instructions

Created [this video](https://user-images.githubusercontent.com/62242/187041498-8d192f78-e376-4761-8ae1-0a85f364a9b0.mp4) - each square is 80px X 80px

1. Enable the experiment for `Enable support for cropping cut off-screen parts of videos`
2. Add the video above to the canvas
3. Position the video partially off canvas
4. Select `Crop off-screen video parts` from the Right Click Menu
5.  The video should now be cropped with the off cavas part of the video removed.

## Checklist

---

Fixes #12093


packages/story-editor/src/app/rightClickMenu/menus/foregroundMediaMenu.js
```diff
@@ -66,10 +66,14 @@ function ForegroundMediaMenu({ parentMenuRef }) {
     handleSendToBack,
     handleBringForward,
     handleBringToFront,
+    handleCropHidden,
   } = useLayerActions();
 
   const { hasShapeMask, removeShapeMask } = useShapeMask(selectedElement);
 
+  const isOffCanvas =
```
timarney: Need to update this is handle any direction on the canvas (L,R,T,B)

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -478,11 +481,84 @@ function useProcessMedia({
...
+        updateExistingElementsByResourceId(resourceId, {
+          ...oldResourceWithId,
+        });
+      };
+
+      const process = async () => {
+        let file = false;
+        try {
+          const originalFile = await fetchRemoteFile(url, mimeType);
+          // calculations need to be worked on here:
```
timarney: Need `proper` calculations  here

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -478,11 +481,84 @@ function useProcessMedia({
...
+        // onUploadSuccess is also called with previousResourceId,
+        // for which we don't need to run this.
+        if (id === resource.id) {
+          postProcessingResource(resource);
+        }
+      };
+
+      const onUploadProgress = ({ resource }) => {
+        const oldResourceWithId = { ...resource, id: oldResource.id };
+        updateExistingElementsByResourceId(resourceId, {
```
timarney, line 505: Need to update the width / height to the new dimensions

timarney:
`ffplay`  can be used to preview a crop with a local install of FFMEG:

https://user-images.githubusercontent.com/62242/187041498-8d192f78-e376-4761-8ae1-0a85f364a9b0.mp4

```
ffplay -i squares.mp4 -vf "crop=80:80:0:0" = (square 1) top right 80x80

ffplay -i squares.mp4 -vf "crop=80:80:0:400" = (square 41) bottom left 80x80

ffplay -i squares.mp4 -vf "crop=80:80:560:0" = (square 8) top right 80x80

ffplay -i squares.mp4 -vf "crop=80:80:560:480" = (square 48) bottom right 80x80
```

timarney:
@miina can you take a look at this ?

I believe the cropping / sizing of the element resource seems to be working - but for sure replacing the element on the canvas (sizing) isn't correct.  Feel free to push commits.

packages/story-editor/src/app/rightClickMenu/hooks/useLayerActions.js
```diff
@@ -22,6 +22,8 @@ import { trackEvent } from '@googleforcreators/tracking';
  * Internal dependencies
  */
 import { useStory } from '../..';
+import { useLocalMedia } from '../../media';
+import useisPartiallyOffCanvas from './useIsPartiallyOffCanvas';
```
miina:
```suggestion
import useIsPartiallyOffCanvas from './useIsPartiallyOffCanvas';
```

packages/story-editor/src/app/rightClickMenu/hooks/useIsPartiallyOffCanvas.js
```diff
@@ -0,0 +1,101 @@
...
+  const offCanvasX = Math.floor((offCanvasLeft + offCanvasRight) * multiplier);
+  const offCanvasY = Math.floor((offCanvasTop + offCanvasBottom) * multiplier);
+  const cropWidth = selectedElement.resource.width - offCanvasX;
+  const cropHeight = selectedElement.resource.height - offCanvasY;
+  const cropX = Math.floor(offCanvasLeft * multiplier);
+  const cropY = Math.floor(offCanvasTop * multiplier);
+
+  return {
+    isOffCanvas,
+    cropParams: {
```
miina: **💬 suggestion:**

Should we perhaps instead add something like `getCropParams` callback inside this hook? Otherwise the calculation runs each time when rendering the right click menu, not just when actually needing to crop.

Maybe the same for `isOffCanvas`. Thoughts?

packages/story-editor/src/app/rightClickMenu/hooks/useLayerActions.js
```diff
@@ -43,6 +51,8 @@ const useLayerActions = () => {
     })
   );
 
+  const { cropParams } = useisPartiallyOffCanvas(selectedElement);
```
miina:
```suggestion
  const { cropParams } = useIsPartiallyOffCanvas(selectedElement);
```

packages/story-editor/src/app/rightClickMenu/hooks/useIsPartiallyOffCanvas.js
timarney: Great idea - I make an update

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -478,11 +482,88 @@ function useProcessMedia({
...
+   */
+  const cropHiddenVideo = useCallback(
+    ({ id: elementId, resource: oldResource }, cropParams) => {
+      const { id: resourceId, src: url, mimeType } = oldResource;
+      const { newWidth, newHeight, cropElement } = cropParams;
+
+      const onUploadSuccess = ({ id, resource }) => {
+        copyResourceData({ oldResource, resource });
+        updateOldTranscodedObject(oldResource.id, resource.id, 'source-image');
+        deleteMediaElement({ id: oldResource.id });
```
miina: **📝 note:**

If the original resource is deleted and the user clicks "Undo", and then tries to crop again, I assume it would cause errors since the original resource doesn't exist anymore? Since the user can undo the action then not sure if we can permanently delete the original resource 🤔

timarney: Perhaps in this case the original resource should be kept - most of that part of the code was copied over from the trim functionality.

timarney: removed the delete

packages/story-editor/src/app/rightClickMenu/hooks/useIsPartiallyOffCanvas.js
timarney: Updated here:
https://github.com/GoogleForCreators/web-stories-wp/pull/12156/commits/4a9dafffc36e6b8670c0f47b205c877f7a1445ba

timarney:
Testing locally 

Looks like , left, right and top work but bottom isn't ending up with the right height.

Screen Shot 2022-08-30 at 6 11 45 PM

miina:
> Testing locally
> 
> Looks like , left, right and top work but bottom isn't ending up with the right height.
> 
> Screen Shot 2022-08-30 at 6 11 45 PM
> 
> offCanvasBottom is happening too early Screen Shot 2022-08-30 at 8 37 04 PM

I guess you need to consider the `DANGER_ZONE_HEIGHT` there as well. I'll be working later today, let me know if you prefer I take a look.

packages/story-editor/src/app/media/utils/test/useProcessMedia.js
```diff
@@ -110,12 +111,21 @@ function setup() {
   const storyContextValue = {
     actions: { updateElementsByResourceId },
   };
+
+  const configState = {
+    capabilities: {
+      hasUploadMediaAction: true,
+    },
+  };
```
swissspidy: Why is this change to the test now needed? I don't see any use of the config provider in the `useProcessMedia` hook.

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -478,11 +482,86 @@ function useProcessMedia({
...
+        if (id === resource.id) {
+          postProcessingResource(resource);
+        }
+      };
+
+      const onUploadProgress = ({ resource }) => {
+        const oldResourceWithId = { ...resource, id: oldResource.id };
+        updateExistingElementsByResourceId(resourceId, {
+          ...oldResourceWithId,
+        });
```
swissspidy: This might be better done in the `onUploadSuccess` callback.

Otherwise what can be observed is that updating the resource & updating the element's dimensions happens in two steps, which is odd.

Of course we can also update the element's dimensions in the `onUploadProgress` callback for faster feedback loop.

But then we also need to restore the original resource & dimensions in `onUploadError`. Otherwise when there's an error with video cropping/uploading, it will lead to a broken appearance on canvas:

Screenshot 2022-08-31 at 12 34 53

-----

Aside: we only want to update this specific element, not all elements with this resource. So we shouldn't use `updateExistingElementsByResourceId`.

packages/story-editor/src/app/rightClickMenu/hooks/useIsPartiallyOffCanvas.js
```diff
@@ -0,0 +1,119 @@
...
+  DANGER_ZONE_HEIGHT,
+  PAGE_HEIGHT,
+  PAGE_WIDTH,
+} from '@googleforcreators/units';
+import { useCallback } from 'react';
+
+function useIsPartiallyOffCanvas(selectedElement) {
+  const { x, y, width, height } = selectedElement;
+
+  const isOffCanvas = useCallback(() => {
```
swissspidy: Why is this a callback and not just a const (e.g. with `useMemo`)?

That said, since this custom hook here does not really use any other hooks (besides wrapping in `useCallback`, we could also just keep things simple and turn this logic into 2 regular functions, not using React at all.

Then we could move them to `src/utils` or elsewhere (since it's not tightly coupled to right click menu anyway)

packages/story-editor/src/app/rightClickMenu/hooks/useIsPartiallyOffCanvas.js
```diff
@@ -0,0 +1,119 @@
...
+ * External dependencies
+ */
+import {
+  DANGER_ZONE_HEIGHT,
+  PAGE_HEIGHT,
+  PAGE_WIDTH,
+} from '@googleforcreators/units';
+import { useCallback } from 'react';
+
+function useIsPartiallyOffCanvas(selectedElement) {
```
swissspidy: Would be nice to cater for element rotation here as well. Maybe in a follow-up, but I think we can avoid using custom checks for each side of the element here. Basically calculating the area of two overlapping rectangles.

Adding tests for this functionality would be an excellent way to prevent regressions when doing such a migration.

includes/Experiments.php
```diff
@@ -372,6 +372,18 @@ public function get_experiments(): array {
...
 				'group'       => 'editor',
 			],
+			/**
+			 * Author: @timarney
+			 * Issue: #12093
+			 * Creation date: 2022-08-18
+			 */
+			[
+				'name'        => 'cropHidden',
+				'label'       => __( 'Crop Hidden', 'web-stories' ),
```
swissspidy: Nit pick, but can we rename this here and everywhere else? "Crop hidden" does not really convey what this does.

Maybe "offScreenVideoCropping" / "Crop off-screen video parts" or something.

packages/story-editor/src/app/media/utils/useFFmpeg.js
```diff
@@ -342,6 +342,55 @@ function useFFmpeg() {
...
+      const { cropWidth, cropHeight, cropX, cropY } = cropParams;
+
+      try {
+        ffmpeg = await getFFmpegInstance(file);
+
+        const type = file?.type || MEDIA_TRANSCODED_MIME_TYPE;
+        const ext = getExtensionFromMimeType(type);
+        const tempFileName = uuidv4() + '.' + ext;
+        const outputFileName = getFileBasename(file) + '-cropped.' + ext;
+        const crop = `crop=${cropWidth}:${cropHeight}:${cropX}:${cropY}`;
```
swissspidy, line 361: Neat 👍

packages/story-editor/src/app/rightClickMenu/menus/foregroundMediaMenu.js
```diff
@@ -203,6 +210,13 @@ function ForegroundMediaMenu({ parentMenuRef }) {
       <ContextMenuComponents.MenuButton onClick={handleOpenScaleAndCrop}>
         {scaleLabel}
       </ContextMenuComponents.MenuButton>
+
+      {cropHidden && isOffCanvas() && (
+        <ContextMenuComponents.MenuButton onClick={handleCropHidden}>
+          {RIGHT_CLICK_MENU_LABELS.CROP_HIDDEN}
+        </ContextMenuComponents.MenuButton>
+      )}
```
swissspidy: We should only display this for media elements that can be cropped

(Aside: super nice that this works for images too!)

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -478,11 +482,86 @@ function useProcessMedia({
...
+          elementId,
+          properties: {
+            x: cropElement.x < 0 ? 0 : cropElement.x,
+            y:
+              cropElement.y < DANGER_ZONE_HEIGHT
+                ? -DANGER_ZONE_HEIGHT
+                : cropElement.y,
+            width: newWidth,
+            height: newHeight,
+          },
```
swissspidy: I just cropped an image and the image jumped a bit on the y axis afterwards:


https://user-images.githubusercontent.com/841956/187660800-ab5d7bbe-c646-4363-a4c9-756172fc0893.mov

swissspidy (commented):
This is already super cool! Nice work so far 🎉 🚀

packages/story-editor/src/app/media/utils/test/useProcessMedia.js
timarney: I believe it's because I imported and am using `useFFMEG` inside of useProcessMedia.

packages/story-editor/src/app/media/utils/useProcessMedia.js
timarney: Okay that makes sense --- will update using `onUploadSuccess`

swissspidy:
Another thing I forgot to mention: it seems like after cropping the video play button is broken; I cannot play the video anymore.

packages/story-editor/src/app/media/utils/test/useProcessMedia.js
swissspidy: Oh, I see! 💡 

That's a great point actually. Currently all our FFmpeg usage is centralized in `useMediaUploadQueue` so we can prevent simultaneous transcoding processes. (Otherwise FFmpeg processes could exhaust your CPU easily)

See how `useMediaUploadQueue` handles `muteVideo` and `trimData` today. We should do the same for cropping.

packages/story-editor/src/app/rightClickMenu/hooks/useIsPartiallyOffCanvas.js
timarney: Good call - ya that code turned into `just regular functions` :)

Moved the functions out to `src/utils` and added tests.

packages/story-editor/src/app/rightClickMenu/hooks/useIsPartiallyOffCanvas.js
timarney: Added: https://github.com/GoogleForCreators/web-stories-wp/issues/12203 to cover element rotation.

Also added unit tests for the current off-screen detection.

packages/story-editor/src/app/rightClickMenu/menus/foregroundMediaMenu.js
timarney: Added isVideo check.

timarney:
> Another thing I forgot to mention: it seems like after cropping the video play button is broken; I cannot play the video anymore.

 🤔  just tested on my local and seems to be working --- can still re-create ? 

https://user-images.githubusercontent.com/62242/187905201-d7adced2-29c4-4547-b780-f8409ac66689.mp4

timarney:
@swissspidy FYI I'm refactoring this PR to use `useMediaUploadQueue`

It's currently in a broken state - will dig in but if you have any quick thoughts as to why I'm getting this error that would be great.

Screen Shot 2022-09-02 at 9 33 19 AM

timarney:
@swissspidy this PR is now updated to use `useMediaUploadQueue`

I'm missing something as I can see it's not getting past this line 

```javascript
if (ITEM_STATUS.PENDING_TRANSCODING !== itemState) {
          return;
}
```

To handle the actual crop.

spacedmonkey:
Can you explain how a new video is created? Is a new poster / base color generated?

swissspidy:
> I'm missing something as I can see it's not getting past this line

You might have found a bug here...

That state is set by `prepareForTranscoding()` here at the very end:

https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js#L453-L492

As you can see, we bail early if the video is considered optimized already. But we don't wanna bail early if we need to crop or mute or trim a video.

So this might need some changes. I can take a closer look tomorrow.

> Can you explain how a new video is created? Is a new poster / base color generated?

Great question. Everything goes through our regular media upload queue, so this should work as usual. If not, we can create follow-up tickets to look into it.

timarney:
@swissspidy 

> ... I can take a closer look tomorrow if you'd like.

That would be helpful thanks

swissspidy:
@timarney e827a1c623d9822e067752a6cd9d13a34af1ec24 should do the trick

It's quite a critical bugfix actually. Without this, we would mark videos as muted but the audio channel is never actually removed.

timarney:
With https://github.com/GoogleForCreators/web-stories-wp/commit/e827a1c623d9822e067752a6cd9d13a34af1ec24 now in place --- looks like the crop happens + finishCropping gets called.  I'm trying to follow the code path from that point as the on `onUploadSuccess` handler doesn't get called.

https://github.com/GoogleForCreators/web-stories-wp/pull/12156/files#diff-5c53ed1d367cabc5dc555d27b2846e0500b3def788cd35a88a0d08c7212e1e51R493

spacedmonkey:
Other video modifications have the id of the original video in post meta. This way there is a link to the original for records and the orignal video can be restored easily. We should do the same here.

timarney:
Cropping is working again now using the `useMediaUploadQueue`

Some issues with the poster image

https://www.youtube.com/watch?v=OlWr6U3iwdQ

timarney:
Looks like the full size image is still getting used for the image. 

@spacedmonkey perhaps you have more context as to what needs to be called to re-gen the poster image after the crop?

Screen Shot 2022-09-08 at 7 44 03 AM

timarney:
Made some more adjustments --- the on canvas video / poster image now works properly but the poster image for the resource in the media library doesn't. 

https://www.youtube.com/watch?v=SIk1Faf1Px0

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -478,11 +479,78 @@ function useProcessMedia({
...
+          elementId,
+          properties: {
+            x: cropElement.x < 0 ? 0 : cropElement.x,
+            y:
+              cropElement.y < DANGER_ZONE_HEIGHT
+                ? -DANGER_ZONE_HEIGHT
+                : cropElement.y,
+            width: newWidth,
+            height: newHeight,
+            resource: { ...resource, id: oldResource.id, poster: null },
```
spacedmonkey:
```suggestion
            resource: { ...resource, id: oldResource.id, poster: null, posterId: 0 },
```

packages/story-editor/src/app/media/utils/test/useProcessMedia.js
```diff
@@ -110,6 +110,7 @@ function setup() {
   const storyContextValue = {
     actions: { updateElementsByResourceId },
   };
+
```
spacedmonkey: ?

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -677,6 +732,15 @@ function useMediaUploadQueue() {
...
 
+    /**
+     * Whether any video cropping is currently in progress.
+     *
+     * @type {boolean} Whether we're muting.
+     */
+    const isCropping = state.queue.some(
+      (item) => item.state === ITEM_STATUS.CROPPING
+    );
+
```
spacedmonkey, line 750: Need to add ITEM_STATUS.CROPPING to `isCurrentResourceProcessing` and `isNewResourceProcessing`

spacedmonkey:
The progress bar does not show on the cropping video. 

Also, the poster image / dimisions should be forced to regenerate earlier. It is weird it jumps in later.

spacedmonkey:
Will there ever be away to restore the original image?

packages/story-editor/src/app/media/utils/test/useProcessMedia.js
timarney: Leftover from some previous work on this PR --- cleaned up now.

timarney:
> The progress bar does not show on the cropping video.

This is updated

> Also, the poster image / dimensions should be forced to regenerate earlier. It is weird it jumps in later.

Will look into that but still trying to get the poster image(s) to work properly...

Still seeing this issue with the "panel" item 

panel

spacedmonkey:
> Will look into that but still trying to get the poster image(s) to work properly...

I wonder, if we should take the poster image file and crop that as well in javascript. Once you done this, just add the poster file into the media process queue and should upload. A quick google [shows](https://stackoverflow.com/questions/53810434/crop-the-image-using-javascript) that images can be cropped in javascript.

timarney:
> > Will look into that but still trying to get the poster image(s) to work properly...
> 
> I wonder, if we should take the poster image file and crop that as well in javascript. Once you done this, just add the poster file into the media process queue and should upload. A quick google [shows](https://stackoverflow.com/questions/53810434/crop-the-image-using-javascript) that images can be cropped in javascript.

Okay that's an option ... I noticed if I remove the featured image via the WP media the video displays as I would expect.

https://www.youtube.com/watch?v=htmZn4vfoUk

spacedmonkey:
Screenshot 2022-09-12 at 12 16 57

This says uploading when it should say transcoding.

timarney:
Did a quick test forcing the poster to regenerate and it works as expected --- with some commented out lines for testing
     
    ```javascript
      //if (!resource.poster && !posterFile) {
        try {
          posterFile = await getFirstFrameOfVideo(file);
        } catch {
          // Not interested in errors here.
        }
     // }
     ```

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -298,6 +302,35 @@ function useMediaUploadQueue() {
...
+      try {
+        const newFile = await cropResource(file, additionalData.cropParams);
+
+        if (!isMounted.current) {
+          return;
+        }
+
+        additionalData.isCropped = true;
+        finishCropping({ id, file: newFile, additionalData });
```
spacedmonkey:
```suggestion
         const newFile = await cropResource(file, additionalData.cropParams);
         const posterFile = await getFirstFrameOfVideo(newFile);

        if (!isMounted.current) {
          return;
        }

        additionalData.isCropped = true;
        finishCropping({ id, file: newFile, posterFile, additionalData });
```

The following + updating to finishCropping to accept posterFile worked well for me.

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -478,11 +479,95 @@ function useProcessMedia({
...
+                ? -DANGER_ZONE_HEIGHT
+                : cropElement.y,
+            width: newWidth,
+            height: newHeight,
+            resource: {
+              ...resource,
+              id: oldResource.id,
+              poster: null,
+              posterId: 0,
+            },
```
spacedmonkey: this is incorrect. Updating the resource in the element here is not going to retrigger the poster generation.

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -478,11 +479,95 @@ function useProcessMedia({
...
+          await uploadMedia([file], {
+            onUploadSuccess,
+            onUploadProgress,
+            cropVideo: true,
+            additionalData: {
+              original_id: resourceId,
+              cropParams,
+              mediaSource: isOptimized ? 'video-optimization' : 'editor',
+            },
+            originalResourceId: resourceId,
```
spacedmonkey, line 547:
```suggestion
            originalResourceId: resourceId,
             resource: {
              ...oldResource,
              poster: null,
              posterId: 0,
              width: newWidth,
              height: newHeight,
            },
```

I think we need to the poster image workaround here and maybe even the new height and width.

spacedmonkey:
One of the issues, if poster's are generated here. 

https://github.com/GoogleForCreators/web-stories-wp/blob/98375658419daa4d9789c8a51e51f200e854ca50/packages/story-editor/src/app/media/useUploadMedia.js#L299-L310

We may have skip this poster generate here.

timarney:
> We don't want to do this, as this will force all uploads regenerate poster, even what it is not needed.

Ya --- was just checking to see if it would generate.

The poster image now works passing the posterFile as you mentioned.

timarney:
Also, the poster image / dimisions should be forced to regenerate earlier. It is weird it jumps in later.

@spacedmonkey -- are you still seeing a jump? I've seen it --- but only once in a while.

Will there ever be away to restore the original video?

No that I'm aware of 

This says uploading when it should say transcoding.

Took a quick look --- do we have other instances where that tool tip changes?  I only found the "Uploading media..." code

spacedmonkey:
https://user-images.githubusercontent.com/237508/189891821-10b8f429-a340-47d5-8a25-1609b376907c.mov

Why does the orignal video disappear? We should keep the original video.

spacedmonkey:
I am seeing the video change position after the transcode is finished. 


https://user-images.githubusercontent.com/237508/189916951-9f11fe5f-f83c-4bdd-8fbd-68d569903f9d.mov


https://user-images.githubusercontent.com/237508/189916989-9894a620-cafc-4938-9562-f365761670da.mov

timarney:
@spacedmonkey 

Seems to vary depending on the video ?

https://www.youtube.com/watch?v=5k1dWqXWHWE

swissspidy (approved):
It seems like in good enough shape to just merge this for now. We can do follow up work & bug fixes in new tickets

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -478,11 +479,100 @@ function useProcessMedia({
...
+            resource: {
+              ...oldResourceWithoutId,
+              width: newWidth,
+              height: newHeight,
+            },
+          });
+        } catch (e) {
+          //eslint-disable-next-line no-console -- surface this error
+          console.log(`cropExistingVideo ${e.message}`);
```
spacedmonkey: Let's remove this. We don't console log other errors. Maybe we could use `trackError` and maybe a snackbar message.

packages/story-editor/src/app/rightClickMenu/menus/foregroundMediaMenu.js
```diff
@@ -21,11 +21,12 @@ import {
   ContextMenu,
   ContextMenuComponents,
 } from '@googleforcreators/design-system';
-import { useRef } from '@googleforcreators/react';
+import { useRef, useMemo } from '@googleforcreators/react';
 
 /**
  * Internal dependencies
  */
+import { useFeature } from 'flagged';
```
spacedmonkey: This should be in the external dependencies section.

packages/story-editor/src/app/media/utils/useProcessMedia.js
timarney: Added a trackError for now.

spacedmonkey (approved):
Two small nitpics. But apart from that, this is looking good. Things to follow up. 

- A rethink of how poster generation happens. While the video is processing, the poster looks wrong and the correct poster image kicks in later. 
- Positioning issue after the cropping is finished. 
- Post meta, have a link to the original attachment id. 
- Possibility to restore the original video?
- Improve testing.

swissspidy:
Created a few follow-up tickets:

* #12308
* #12309
* #12310

PR #12167, opened 2022-08-25
Configure Jest for TypeScript

### LINKED ISSUE #12097: TypeScript: Use with Jest
Let's use [ts-jest](https://github.com/kulshekhar/ts-jest) for improved TypeScript support in Jest so that we can also write unit tests with TypeScript —with proper type checking and not just transpilation.

See https://kulshekhar.github.io/ts-jest/docs/babel7-or-ts/
See https://kulshekhar.github.io/ts-jest/docs/27.0/guides/test-helpers/

* [ ] Update ESLint config
* [ ] Add `preset: 'ts-jest'` to Jest config

swissspidy:
## Context

So that we can write tests in TypeScripts if desired. See #12097

## User-facing changes

None

## Testing Instructions

1.

No

No

No

## Checklist

---

Fixes #12097


tsconfig.shared.json
```diff
@@ -28,8 +28,9 @@
     "pretty": true,
     "sourceMap": true,
     "typeRoots": ["./typings", "./node_modules/@types"],
-    "types": []
+    "types": ["node", "jest"]
   },
+  "include": ["typings/**/*"],
   "exclude": [
```
swissspidy, line 33: Excluding `"**/test/**",` here means the type checking with `ts-jest` does not work.

Could change to `"**/test/**/*.js",` so TS files are included in the type checking.

packages/units/src/getBoundRect.ts
```diff
@@ -67,13 +67,13 @@ function getBoundRect(list: ElementBox[]) {
...
 export function calcRotatedObjectPositionAndSize(
-  angle: number,
+  angle = 0,
   x: number,
   y: number,
   width: number,
   height: number
 ) {
-  if (!angle || angle === 0) {
+  if (!angle) {
```
miina, line 82: Hm, now if `angle` is `0`, `!angle` is true. It should still have `angle === 0` check as well, no?

packages/tracking/src/test/trackEvent.ts
```diff
@@ -34,23 +34,23 @@ describe('trackEvent', () => {
...
     config.trackingEnabled = true;
 
-    gtag.mockImplementationOnce((type, eventName, eventData) => {
-      eventData.event_callback();
+    jest.mocked(gtag).mockImplementationOnce((_type, _eventName, eventData) => {
+      (eventData as ControlParams).event_callback?.();
     });
 
-    await trackEvent('name', { foo: 'abc', bar: 'def', baz: 'ghi' });
+    await trackEvent('name', { name: 'abc', value: 'def', event_label: 'ghi' });
```
miina, line 41: Is there a way to still test with unexpected props? In case we want to test that specifically.

packages/media/src/types.ts
```diff
@@ -253,7 +253,17 @@ export interface ResourceSize {
   height: number;
 }
 
-export type ResourceId = string;
+// TODO: Enforce number type for dimensions already in wp-* integration packages.
+// TODO: Make sure all properties are required.
+// For this, move logic like normalizeResourceSizes and parts of calculateSrcSet.
```
miina: Alternatively, could we use `Partial<ResourceSizeInput>` in [normalizeResourceSizes](https://github.com/GoogleForCreators/web-stories-wp/pull/12167/files#diff-314afaadf6da1291d3cc56473d42741919bfb1971eeefc1a0017d452c6db6d0bR29) typing to keep these required already now? And then once the type is complete, we could cast it to `ResourceSizeInput` when [adding to the normalized sizes](https://github.com/GoogleForCreators/web-stories-wp/pull/12167/files#diff-314afaadf6da1291d3cc56473d42741919bfb1971eeefc1a0017d452c6db6d0bR40).`

packages/units/src/getBoundRect.ts
swissspidy: `!angle` does cover the 0 case too though? Or am I missing something?

packages/tracking/src/test/trackEvent.ts
swissspidy: Yeah, by ignoring TypeScript errors.

packages/units/src/getBoundRect.ts
miina: Oh, right, I was thinking something completely off.

packages/media/src/types.ts
swissspidy: Probably. But since that type is only used for that function and it should be moved anyway, I just went ahead and moved it, resolving this todo :-)

tests/js/jest.config.js
```diff
@@ -22,11 +22,25 @@ import { fileURLToPath } from 'url';
...
   transform: {
-    '^.+\\.[jt]sx?$': 'babel-jest',
+    '^.+\\.tsx?$': [
+      'ts-jest',
+      {
+        // Disable type checking but improve performance.
+        // See https://kulshekhar.github.io/ts-jest/docs/getting-started/options/isolatedModules
+        // See https://github.com/kulshekhar/ts-jest/issues/1648
+        // TODO: Look into disabling this again.
+        isolatedModules: true,
```
swissspidy: Not really happy about this but otherwise tests won't work at all since it fails to find some types.

timarney:
Looks like this is falling out of date.

swissspidy:
I know, it‘s why it‘s still a draft

swissspidy:
@barklund Looks like I broke `rich-text` tests here. Could you perhaps take a look? 😅  Not seeing the forest for the trees right now.

packages/wp-story-editor/src/api/utils/normalizeResourceSizes.js
```diff
@@ -15,38 +15,50 @@
...
+  const normalizedSizes = {};
 
   if (!sizes) {
     return normalizedSizes;
   }
 
   for (const size of Object.keys(sizes)) {
     const data = sizes[size];
+
+    if (!data.width || !data.height || !data.sourceUrl) {
```
swissspidy, line 47: This might help with #12485, because that one looks like an issue with trying to do `replaceAll()` an undefined/null `sourceUrl`

packages/rich-text/src/test/styleManipulation.ts
```diff
@@ -17,62 +17,81 @@
...
+  convertFromRaw,
+  EditorState,
+  RawDraftContentState,
+  SelectionState,
+} from 'draft-js';
 
 /**
  * Internal dependencies
  */
+import type { OrderedSet } from 'immutable';
```
spacedmonkey, line 30: Isn't this external dependencies

spacedmonkey (approved):
There is a lot of code here. It looks good to me. I like how the code quality improvements.

packages/rich-text/src/test/styleManipulation.ts
merapi: In tests. Is there any downside to this?

timarney (approved):
Not caused by this PR but while reviewing found these warning in the test run(s) 

Might some quick fixes while your in here (or new issues for later as you see fit):

https://github.com/GoogleForCreators/web-stories-wp/actions/runs/3378308481/jobs/5608342421#step:9:17

https://github.com/GoogleForCreators/web-stories-wp/actions/runs/3378308481/jobs/5608342603#step:9:17

merapi (approved):
LG, I ran some tests locally.

packages/i18n/src/test/transformNode.tsx
```diff
@@ -57,7 +58,11 @@ describe('transformNode', () => {
     node.setAttribute('class', 'bar');
     node.textContent = 'Hello World';
 
-    const AwesomeComponent = ({ children }) => {
+    const AwesomeComponent = ({
+      children,
+    }: {
+      children?: ReactNode | undefined;
```
barklund: If it's optional (`?:`), the `| undefined` is superfluous, no?

packages/i18n/src/translateWithMarkup.tsx
```diff
@@ -131,13 +136,18 @@ function TranslateWithMarkup({
...
   const node = new DOMParser().parseFromString(children, 'text/html').body
     .firstChild as Element;
-  return node
-    ? transform(
-        node,
-        mapping
-        //eslint-disable-next-line react/no-array-index-key -- Order should never change.
-      ).map((element, index) => <Fragment key={index}>{element}</Fragment>)
-    : [];
+  return node ? (
```
barklund: I think it's more clear if we short-circuit the null return like:

```tsx
if (!node) {
  return null;
}
return ...
```

packages/i18n/src/translateWithMarkup.tsx
```diff
@@ -111,7 +116,7 @@ interface TranslateWithMarkupProps {
 function TranslateWithMarkup({
   mapping = {},
   children,
-}: TranslateWithMarkupProps) {
+}: TranslateWithMarkupProps): ReactElement | null {
```
barklund, line 119: Isn't `ReactNode` the proper return type? This should include both `null` and all the primitives as well as JSX elements.

packages/i18n/src/translateWithMarkup.tsx
barklund: (I know you didn't write this way in the first place, but just now that you're already here :P )

packages/media/src/resourceList.ts
```diff
@@ -19,7 +19,7 @@
  */
 import type { ResourceId } from '@googleforcreators/types';
 
-interface ResourceCacheEntry {
+export interface ResourceCacheEntry {
   url: string;
   type: 'cached' | 'fullsize';
```
barklund: Should we make this a proper enum? Would remove the need for types in the `resourceList.test.ts` suite.

packages/moveable/src/test/areEventsDragging.ts
```diff
@@ -31,7 +31,9 @@ describe('areEventsDragging', () => {
       clientY: 1,
       timeStamp: window.performance.now() + 10,
     };
-    expect(areEventsDragging(firstEvent, secondEvent)).toBeFalse();
+    expect(
+      areEventsDragging(firstEvent as MouseEvent, secondEvent as MouseEvent)
```
barklund, line 35: This is a bit dirty, forcing the clearly _not-a-mouse-event_ to be one. Maybe we should create the events correctly through `new MouseEvent`.

I know that doesn't allow us to specify the timestamp, so we would have to use `jest.useFakeTimers()` and `jest.advanceTimersByTime()`. I actually don't know if it'll correctly advance the time inside the generated events, but I think it will?

This is definitely a nit though!

packages/react/src/renderToStaticMarkup.ts
```diff
@@ -35,7 +35,9 @@ import type { ReactElement } from 'react';
  * @param element React element.
  * @return Markup.
  */
-function renderToStaticMarkup(element: ReactElement) {
+function renderToStaticMarkup(
+  element: ReactElement | ReactElement[] | null | string
```
barklund, line 39: `ReactNode` would work again here?

packages/rich-text/src/formatters/test/italic.tsx
```diff
@@ -14,15 +14,22 @@
...
 /**
  * Internal dependencies
  */
 import {
-  togglePrefixStyle,
   getPrefixStylesInSelection,
+  togglePrefixStyle,
 } from '../../styleManipulation';
-import { NONE, ITALIC } from '../../customConstants';
+import { ITALIC, NONE } from '../../customConstants';
```
barklund, line 30: You're using a plugin to alphabetically sort imports, or?

packages/rich-text/src/formatters/test/letterSpacing.tsx
```diff
@@ -35,16 +43,16 @@ const { elementToStyle, stylesToCSS, getters, setters } = formatter;
 
 describe('Color formatter', () => {
   beforeAll(() => {
-    getPrefixStylesInSelection.mockImplementation(() => [NONE]);
+    jest.mocked(getPrefixStylesInSelection).mockImplementation(() => [NONE]);
```
barklund, line 46: I didn't know about [this one](https://jestjs.io/docs/mock-function-api/#jestmockedsource-options). Nice!

packages/rich-text/src/formatters/test/letterSpacing.tsx
```diff
@@ -73,25 +81,33 @@ describe('Color formatter', () => {
 
   describe('stylesToCSS', () => {
     it('should ignore styles without a letter spacing style', () => {
-      const css = stylesToCSS(['NOT-LETTERSPACING', 'ALSO-NOT-LETTERSPACING']);
+      const css = stylesToCSS({
+        toArray: () => ['NOT-LETTERSPACING', 'ALSO-NOT-LETTERSPACING'],
+      } as OrderedSet<string>);
```
barklund: Uh, that's pretty ugly, but I see it's necessary. Maybe a small util function for it?

packages/tracking/src/test/trackEvent.ts
```diff
@@ -34,23 +34,23 @@ describe('trackEvent', () => {
...
     expect(gtag).toHaveBeenCalledWith('event', 'name', {
-      event_callback: expect.any(Function),
-      foo: 'abc',
-      bar: 'def',
-      baz: 'ghi',
+      event_callback: expect.any(Function) as () => void,
```
barklund: Hmm, this is a bit weird. This feels wrong with that cast? Though I can't see how it could be any cleaner than that though - you can't type `expect.any` correctly, can you? Maybe with an overload for every constructor type? That's besides the point though, not much to do here.

patches/@types+draft-js+0.11.9.patch
```diff
@@ -0,0 +1,27 @@
...
++              | 'STRIKETHROUGH'
++              | 'UNDERLINE'
++              | 'NONE'
++              | 'CUSTOM-ITALIC'
++              | 'CUSTOM-UNDERLINE'
++              | 'CUSTOM-WEIGHT'
++              | 'CUSTOM-WEIGHT-700'
++              | 'CUSTOM-WEIGHT-900'
++              | 'CUSTOM-COLOR'
++              | 'CUSTOM-LETTERSPACING'
```
barklund, line 23: It's not limited to these weights either, though. And the custom color and letter spacing can have any of a number of valid prefixes like `CUSTOM-COLOR-BADA55` (IIRC).

And also, custom style types are supported by this package, why do we have to add those types inside that? Well, I guess it's because the types have been tagged on by third-party developers later.

barklund (approved):
I'm loving this. I especially like how many irrelevant tests and slightly off typings (missing union, etc) they reveal!

I only have some minor comments, but I guess most of them are "why aren't things different" 😂

packages/rich-text/src/formatters/test/italic.tsx
swissspidy: Not sure how this one happened

patches/@types+draft-js+0.11.9.patch
swissspidy: This is the only way I've found to be able to extend `DraftInlineStyleType` with our custom types. There's probably a better way, so if you know how to make this work, please let me know.

packages/i18n/src/translateWithMarkup.tsx
swissspidy: It's a bit of a rabbit hole with React's `renderToStaticMarkup` expecting `ReactElement` etc.

There's probably a better way to make things work, any suggestions welcome. Just `ReactNode` (or not specifying a return type) doesn't work out of the box

packages/moveable/src/test/areEventsDragging.ts
swissspidy: When using fake timers, using `jest.advanceTimersByTime()` between two `new MouseEvent()` calls has no effect. At least in my testing. Does it work for you?

packages/rich-text/src/formatters/test/letterSpacing.tsx
swissspidy: Ended up using `OrderedSet.of()` from `immutable` for this

packages/tracking/src/test/trackEvent.ts
swissspidy: Looks like they're completely unnecessary

packages/moveable/src/test/areEventsDragging.ts
barklund: I didn't actually try it out. But if it doesn't work, it's pretty useless then.

PR #12169, opened 2022-08-26
Convert `moveable` package to TypeScript

### LINKED ISSUE #12088: TypeScript: Convert `moveable` package
This should help prevent scenarios like #11136 due to improved typing.

There's already some POC here that can be leveraged.

swissspidy:
## Relevant Technical Choices

* Defined the styled-components theme with TS.
* Added some custom type definitions under `packages/moveable/src/typings` (see to-do below)

## To-do

**Help needed with these:**

- Fix any bugs
- Fix remaining type error when passing a ref to `OriginalMoveable`
- Figure out a solution so we can move `packages/moveable/src/typings/*.d.ts` to top-level `typings/` folder instead. Otherwise it's a bit messy.

## User-facing changes

N/A

## Testing Instructions

1.

## Checklist

---

Fixes #12088


packages/moveable/src/moveable.tsx
```diff
@@ -28,13 +29,20 @@ import InOverlay from './overlay';
...
 const DEFAULT_Z_INDEX = 10;
 
-function MoveableWithRef({ onContextMenu, ...moveableProps }, ref) {
+interface MoveableWithRefProps extends MoveableProps {
+  onContextMenu: MouseEventHandler<HTMLElement>;
+}
+
+function MoveableWithRef(
+  { onContextMenu, ...moveableProps }: MoveableWithRefProps,
+  ref: ForwardedRef<OriginalMoveable>
```
barklund, line 38: I think this fixes it. For some reason, Moveable breaks with regular practices and makes the `ref` a reference to the component itself and not to either an HTML element or some imperative handle. It's probably because it's still written class-based and not converted to hooks.

swissspidy: Thanks for looking into it!

swissspidy:
Last lint error fixed, this should now be ready again.

miina (approved):

PR #12197, opened 2022-08-31
Editor: Add link to revisions

### LINKED ISSUE #85: Editor: Add link to revisions
## Feature description

In Gutenberg there's a link to the revisions comparison screen in the document tab when there are some:

11956-select-revisions-sidebar

It would be nice to do the same in the story editor as well, maybe in the `PublishPanel` component (`packages/wp-story-editor/src/components/documentPane/publish/publish.js`) below the poster image / publisher logo setting

Screenshot 2022-08-22 at 16 47 39

We can use a similar icon as well:

https://fonts.google.com/icons?selected=Material%20Symbols%20Outlined%3Ahistory%3AFILL%400%3Bwght%40400%3BGRAD%400%3Bopsz%4048

timarney:
## Context

In Gutenberg there's a link to the revisions comparison screen in the document tab It would be nice to do the same in the story editor as well, maybe in the PublishPanel component (packages/wp-story-editor/src/components/documentPane/publish/publish.js) below the poster image / publisher logo setting

## Summary

Adds revisions link / icon to publish panel

Screen Shot 2022-09-08 at 6 34 32 AM

## Testing Instructions

0. Enable improved autosaves experiment
1. Create a new story
2. Publish story
3. There should now be a link to the revisions page in the Document tab
4. Verify the link works as expected
5. Make changes to story & save changes
6. Verify the revision count increases and the link points to the revisions page for the latest revision

Note: improvements to the revisions page itself are not part of this ticket.

No

No

No

## Checklist

---

Fixes #85


swissspidy:
For reference, if it helps, here's how WP gets the necessary data to display:

https://github.com/WordPress/gutenberg/blob/7b519e482ca62a186c29d2c00e41c5947712f44f/packages/editor/src/store/selectors.js#L225-L251
https://github.com/WordPress/gutenberg/blob/33d84b036592a5bf31af05b7710f3b2b14163dc4/packages/editor/src/components/post-last-revision/index.js
https://github.com/WordPress/wordpress-develop/blob/69b9ecc54e2b2e1724f10d17dfd681aa8aee8810/src/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php#L2066-L2081

Since we're only using this in `wp-story-editor` right now, getting this information from the REST API is probably a bit overkill. Also, don't want to assemble the `wp-admin/revision.php` link on the client side when we can do it in PHP.
So I think we can just update `\Google\Web_Stories\Admin\Editor::get_editor_settings()` and pass the revision count and revision link there.

timarney:
Added the code to Editor.php for now to move this along a bit.

timarney:
> For this we'd need to pass the revision count & link via the REST API.

@swissspidy is there an existing REST controller I should add a route to ? Or would this be a new controller / route?

swissspidy:
We have Story_Controller with lots of precedence with custom additional fields. We can add these fields there, no need for a new route.

spacedmonkey:
Spent a bit of time on this PR. On the face of it, I don't like the new PHP code added. Adding these new fields is not nessary as this data is already in the REST API in another format. It is currently hidden. So I put together another PR on top of this one that get that using existing fields. It is a WIP, but if you like the idea, I recommend merging into this branch and continuing to work on it. See https://github.com/GoogleForCreators/web-stories-wp/pull/12264

It is worth noting that this data is in the links field. See this screenshot. 

Screenshot 2022-09-08 at 16 56 22

spacedmonkey (dismissed):
Some changes are required here.

timarney:
> This data is already in the REST API in another format. 

Nice okay

I cherry picked your commits into this branch

packages/wp-story-editor/src/components/documentPane/publish/publish.js
```diff
@@ -414,6 +442,33 @@ function PublishPanel({ nameOverride }) {
...
         </HighlightRow>
+        {improvedAutosaves && (
+          <RevisionsWrapper>
+            <LabelWrapper>
+              <Label>
+                <LabelIconWrapper>
+                  <Icons.History width={24} height={24} aria-hidden />
+                </LabelIconWrapper>
+                <RevisionsLabel>
+                  {revisionCount} {__('Revisions', 'web-stories')}
```
spacedmonkey: This should use _n and sprintf.

packages/wp-story-editor/src/components/documentPane/publish/publish.js
```diff
@@ -414,6 +447,41 @@ function PublishPanel({ nameOverride }) {
...
+                <RevisionsLabel>
+                  {sprintf(
+                    /* translators: %d: number of revisions. */
+                    _n(
+                      '%d Revision',
+                      '%d Revisions',
+                      revisionCount,
+                      'web-stories'
+                    ),
+                    sprintf('%s', revisionCount)
```
spacedmonkey:
```suggestion
                    revisionCount
```

This should be fine here.

packages/wp-story-editor/src/components/documentPane/publish/publish.js
```diff
@@ -414,6 +447,41 @@ function PublishPanel({ nameOverride }) {
...
+                  )}
+                </RevisionsLabel>
+              </Label>
+              {revisionLink && revisionId ? (
+                <Row>
+                  <Link
+                    rel="noopener noreferrer"
+                    target="_blank"
+                    href={addQueryArgs(revisionLink, { revision: revisionId })}
```
spacedmonkey: We may want to have a help function to get a revision link from config. Makin this a little more compatible with other cmses. Thoughts @swissspidy

spacedmonkey:
I would love to see some kind of testing here. A unit test or karma test would be great. 

Also, I believe now, we will need to update fixtures, so include the new fields.

packages/wp-story-editor/src/components/documentPane/publish/publish.js
```diff
@@ -414,6 +447,41 @@ function PublishPanel({ nameOverride }) {
...
+                    _n(
+                      '%d Revision',
+                      '%d Revisions',
+                      revisionCount,
+                      'web-stories'
+                    ),
+                    revisionCount
+                  )}
+                </RevisionsLabel>
+              </Label>
```
spacedmonkey: Maybe we don't show this until we have at least one revision. It looks a little weird with 0. 

Screenshot 2022-09-09 at 12 48 00

timarney: Ya --- I had it hidden at one point --- will re-hide

spacedmonkey: Styling nitpic, I think that manage link should be on the same line. Compare. 

Screenshot 2022-09-09 at 13 06 00
Screenshot 2022-09-09 at 13 04 23
Screenshot 2022-09-09 at 13 04 19

spacedmonkey:
🤦 There seems to an issue in core. 

The count of revisions is wrong. After 10 revisions, the number doesn't go up. This issue is fixed in https://github.com/WordPress/wordpress-develop/commit/3506684f2a7516d1e4a272dafd9af7ff6e08c55f ( please note the committer there ). We are going to need to backport this change.

spacedmonkey:
I wonder if we should up the number of revisions from 10. See 
https://github.com/GoogleForCreators/web-stories-wp/blob/e1efbf11ade523a9cd1420f555033e69c9802c8f/includes/Story_Post_Type.php#L388-L399

packages/wp-story-editor/src/components/documentPane/publish/publish.js
swissspidy: The link cannot come from the config because it‘s dynamic. It should come from the REST API, which is why I‘ve originally suggested adding a new REST field to the API, so that we don‘t have to create the link here ourselves.

packages/wp-story-editor/src/components/documentPane/publish/publish.js
timarney: Moved to single line

packages/wp-story-editor/src/components/documentPane/publish/publish.js
spacedmonkey: Why can't it come from a config? How about a function like this.

```getRevisionLink( revisionId )```

Comment (swissspidy) [reply packages/wp-story-editor/src/components/documentPane/publish/publish.js]: The final revision link is `revision.php?revision=<id>` where `<id>` changes every time you save the story in the editor.

Anyway, the current way is OK for now since this is in the `wp-story-editor` package, so other CMSes are not relevant here.

File: packages/wp-story-editor/src/components/documentPane/publish/publish.js [hunk 5]
```diff
@@ -116,6 +123,24 @@ const LogoImg = styled.img`
   max-height: 96px;
 `;
 
+const RevisionsWrapper = styled.div`
+  width: 100%;
+  margin-bottom: 20px;
+  margin-top: 4px;
```
swissspidy: Minor styling suggestion

```suggestion
  margin-bottom: 16px;
  margin-top: 20px;
```

packages/wp-story-editor/src/components/documentPane/publish/publish.js
```diff
@@ -414,6 +448,39 @@ function PublishPanel({ nameOverride }) {
             </LabelWrapper>
           </DropdownWrapper>
         </HighlightRow>
+        {improvedAutosaves && revisionCount >= 1 ? (
+          <RevisionsWrapper>
+            <LabelWrapper>
```
swissspidy: `LabelWrapper` by default has `height: 40px` so that the publisher logo & poster image setting are more nicely aligned.

We don't need this height though for revisions.

Let's create a new `RevisionsLabelWrapper` or so without that height.

swissspidy (approved):
Just two minor nits otherwise looks good

packages/wp-story-editor/src/components/documentPane/publish/publish.js
timarney: Removed the `LabelWrapper` --- not really needed.

PR #12198, opened 2022-08-31
Convert `i18n` package to TypeScript

swissspidy:
## Context

Part of the TS migration, see #1715

## Summary

Converts the `i18n` package to TypeScript.

## Relevant Technical Choices

* Updated ESLint config to reduce redundancy when using TypeScript. No need to declare types in JSDoc when they're already enforced with TS.

## User-facing changes

N/A

## Testing Instructions

- [x] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1.

### Does this PR have a security-related impact?

No

### Does this PR change what data or activity we track or use?

No

### Does this PR have a legal-related impact?

No

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/googleforcreators/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [x] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/googleforcreators/web-stories-wp/tree/main/docs#testing))
- [x] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #12180

PR #12200, opened 2022-08-31
Convert `patterns` package to TypeScript

### LINKED ISSUE #12178: TypeScript: Convert `patterns` package

barklund:
## Context

What is says on the tin.

NB: This depends on the `i18n` package, which hasn't been converted to TS yet, so it has some errors in a single file due to that.

## Testing Instructions

## Checklist

---

Fixes #12178


packages/patterns/src/getOpaquePattern.ts
```diff
@@ -29,8 +34,11 @@ function getOpaquePattern(pattern) {
...
-  return Object.keys(obj)
-    .filter((key) => !propertiesToRemove.includes(key))
-    .reduce((newObj, key) => ({ ...newObj, [key]: obj[key] }), {});
+function objectWithout(
+  obj: Record<string, unknown>,
+  propertiesToRemove: Array<string>
+) {
+  return Object.fromEntries(
+    Object.entries(obj).filter(([key]) => !propertiesToRemove.includes(key))
+  );
```
barklund, line 43: I made a slight improvement here - this is way faster than re-creating objects over and over.

packages/patterns/src/types.ts
```diff
@@ -46,3 +46,50 @@ export const PatternPropType = PropTypes.shape({
...
+  };
+  center?: {
+    x: number;
+    y: number;
+  };
+  rotation?: number;
+  alpha?: number;
+};
+
+export type Pattern = Solid | Linear | Radial;
```
miina: Curious why use types and not interfaces for the objects? As far as I've read so far the general recommendation seems to be using types for functions and interfaces for objects, however, it looks like with the latest TS versions, these two have become more similar. Is there any reason why you prefer types to interfaces here?

packages/patterns/src/getOpaquePattern.ts
```diff
@@ -29,8 +34,11 @@ function getOpaquePattern(pattern) {
 
 export default getOpaquePattern;
 
-function objectWithout(obj, propertiesToRemove) {
-  return Object.keys(obj)
-    .filter((key) => !propertiesToRemove.includes(key))
-    .reduce((newObj, key) => ({ ...newObj, [key]: obj[key] }), {});
+function objectWithout(
+  obj: Record<string, unknown>,
+  propertiesToRemove: Array<string>
```
miina, line 39: **💡 thought:**
Should we try to use the same convention across the files for consistency, .e.g always either `Array<string>` or `string[]`? Or no preference? Not sure if this is usually considered important in TS projects. Thoughts?

packages/patterns/package.json
```diff
@@ -31,6 +31,7 @@
   },
   "main": "dist/index.js",
   "module": "dist-module/index.js",
+  "types": "dist-types/index.d.ts",
   "source": "src/index.js",
```
miina:
```suggestion
  "source": "src/index.ts",
```
Same above as well for `src/index.js` reference.

miina (approved):
🔨  (LGTM pending i18n / test failures)

miina:
Looks like there are some test failures now.

packages/patterns/src/getOpaquePattern.ts
swissspidy: No strong preference, but we could add a lint rule for this: https://typescript-eslint.io/rules/array-type
I think we've mostly used `string[]` so far, also in the PHP code base

packages/patterns/package.json
barklund: Good catch!

packages/patterns/src/getOpaquePattern.ts
barklund: Yeah, let's go for the more compact syntax.

packages/patterns/src/types.ts
barklund: They're mostly interchangeable and I don't think there's any real difference in these cases. Interfaces are easier to extend, but this doesn't require extension ever, so not require. The union type though can only be a type (you can't create a union interface IIRC), so the final thing here has to be type, but the other things could be interfaces.

barklund: I did improve the code around this a bit though, creating a `Gradient` intermediate type. Not sure if this is best-practice, but it works out pretty well.

packages/patterns/src/types.ts
```diff
@@ -46,3 +46,52 @@ export const PatternPropType = PropTypes.shape({
...
+
+export type Hex = {
+  r: number;
+  g: number;
+  b: number;
+  a?: number;
+};
+
+export type Solid = {
+  type?: PatternType.SOLID;
```
barklund, line 66: This is a bit annoying by the way. Yes, it makes solid patterns more compact, because the solid type is implicit, but it makes type narrowing more complex, as we can't just get auto-narrowing by validating the type, but we also have to check for lack of type and then manually cast. But I still think it's worth it.

packages/patterns/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../../tsconfig.shared.json",
+  "compilerOptions": {
+    "rootDir": "src",
+    "declarationDir": "dist-types"
+  },
+  "references": [],
```
swissspidy, line 7: Once the i18n package change is merged, we can add a reference here.

packages/patterns/src/isPatternEqual.ts
```diff
@@ -19,8 +19,13 @@
  */
 import convertToCSS from './convertToCSS';
 import generatePatternStyles from './generatePatternStyles';
+import type { Pattern } from './types';
 
-export default function isPatternEqual(p1, p2, patternType = undefined) {
+export default function isPatternEqual(
+  p1: Pattern,
+  p2: Pattern,
+  patternType?: string
```
swissspidy, line 27: These are the only valid values AFAIK. Could be moved to `types.ts` as well.

```suggestion
  patternType?: 'color' | 'background'
```

packages/patterns/src/generatePatternStyles.ts
```diff
@@ -46,23 +59,26 @@ function truncate(val, pos) {
...
-    ? `ellipse ${truncate(100 * size.w, 2)}% ${truncate(100 * size.h, 2)}%`
-    : '';
-  const centerString = center
-    ? ` at ${truncate(100 * center.x, 2)}% ${truncate(100 * center.y, 2)}%`
-    : '';
-  switch (type) {
-    case 'radial':
+function getGradientDescription(pattern: Gradient) {
+  switch (pattern.type) {
+    case 'radial': {
```
swissspidy: Should we use the `PatternType` enum here for the cases?

packages/patterns/src/generatePatternStyles.ts
```diff
@@ -46,23 +59,26 @@ function truncate(val, pos) {
...
+      const { size, center } = pattern as Radial;
```
swissspidy: Wondering if we could define the gradient types in a way that the assertions here are not necessary, i.e. since `pattern.type === 'radial' we _know_ it's a `Radial`.

swissspidy: Ah, just saw https://github.com/GoogleForCreators/web-stories-wp/pull/12200#discussion_r960552291

packages/patterns/tsconfig.json
barklund: Should I wait for that package to be merged before merging this? To remove the lint errors?

packages/patterns/src/isPatternEqual.ts
barklund: Isn't it used for `border-color` too? But yes, it makes sense to narrow this down as much as possible. I'll look into it.

packages/patterns/src/generatePatternStyles.ts
barklund: Oh yes, my bad.

packages/patterns/src/generatePatternStyles.ts
barklund: Here it's actually because `Gradient` isn't a _supertype_ of `Linear` and `Radial`. Maybe I should change the types a bit so this will work. That probably requires interfaces though. I'll check to see if it can be improved.

packages/patterns/tsconfig.json
swissspidy: I just went ahead and merged #12198, so just needs a merge/rebase now!

packages/patterns/src/generatePatternStyles.ts
barklund: Fixed it in c986cafb77.

packages/patterns/src/isPatternEqual.ts
barklund: It is also used for `fill` and `stroke`.

But there are also other cases making narrowing tricky. [Here](https://github.com/GoogleForCreators/web-stories-wp/blob/512328ada20ed77687817ee479bd09255836ca61/packages/element-library/src/product/display.js#L132) it is used for a CSS variable (`--box-shadow-color`), which isn't an unreasonable usage, and the naming is completely arbitrary.

We could enforce a certain number of properties plus any string that starts with `'--'`, but that seems like overkill?

packages/patterns/src/generatePatternStyles.ts
barklund: Fixed in c986cafb77 (in this and many other places).

PR #12202, opened 2022-09-01
Convert `transform` package to TypeScript

### LINKED ISSUE #12173: TypeScript: Convert `transform` package

miina:
## Context
Converts `transform` package to TypeScript

## Relevant Technical Choices
- Not sure if the typing around the whole state and context is correct here.

## User-facing changes
N/A

## Testing Instructions

1.

## Checklist

---

Fixes #12173


packages/transform/src/types.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export interface HandlersList {
+  [id: string]: TransformHandler[];
+}
+
+export interface TransformsList {
+  [id: string]: Record<string, unknown> | null;
+}
+
+export interface State {
```
miina, line 33: @swissspidy Do you happen to have a good example of creating a context with actions and state that I could check? Not sure exactly how to do this here, had to use [type casting](https://github.com/GoogleForCreators/web-stories-wp/pull/12202/files#diff-ce14352028d036cff60c103838a120678c7279b6ac26f2a9e83a279ba5582d82R44) which doesn't seem right here. Or is this an example of the `useContextSelector` that might still need work (since it's all related to that, and has to match the typing of the context selector)?

swissspidy: So far I only attempted this in `packages/units/src/context.ts` / `packages/units/src/contextProvider.tsx` but yeah it still needs some work to get type support when using `useContextSelector`.

This might do the trick though:

https://github.com/GoogleForCreators/web-stories-wp/tree/try/fix-context-state

This way I get type support when doing something like

```js
useUnits(({ state: { pageSize } }) => pageSize);
```

miina: Hm, not sure if I did something incorrectly but I still have to [assign types](https://github.com/GoogleForCreators/web-stories-wp/pull/12202/files#diff-0408b18c76af2834bb09ccf3f2e26c80643b11f5634f017a02d77b314f822e5eR41) in the provider for it not to throw errors.

packages/transform/src/useTransformHandler.ts
```diff
@@ -23,25 +23,25 @@ import { useEffect } from '@googleforcreators/react';
...
+ * @param handler The transform handler. The argument is
  * the frame object. The `null` value resets the transform.
- * @param {Array} [deps] The effect's dependencies.
+ * @param [deps] The effect's dependencies.
  */
-function useTransformHandler(id, handler, deps = undefined) {
+function useTransformHandler(
+  id: string,
+  handler: TransformHandler,
+  deps: unknown[] | undefined = undefined
```
swissspidy: In other places I've used the `DependencyList` type from the React package for this, e.g.

```suggestion
  deps: DependencyList | undefined = undefined
```

barklund (approved):
🛕

PR #7318, opened 2021-04-28
Enable custom page templates feature

miina:
## Context
Enables the custom page templates feature

## Summary
See [Figma](https://www.figma.com/file/1YM3jA9h4QGc4xLdyaXIJS/Stories-Sprint-1.7?node-id=7524%3A9254)

## Relevant Technical Choices
Currently, only the first 100 page templates are fetched by default. We should define what this number should be and if we can[ fetch all the templates](https://github.com/google/web-stories-wp/pull/7268#discussion_r618416479) or should implement pagination for the virtual list.

## To-do

- [x] Enable feature
- [x] Add Karma tests (still missing some parts to be able to add them all)

### QA

- [ ] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1. Open the Page templates panel in the library
2. Verify the dropdown for choosing between the Default and Saved templates is present
3. Click on the "Save current page as template" button.
4. Verify that "Saved templates" section will be automatically opened and the new template is temporarily shown blue highlight around it.
5. Verify the "Save current page as template" button looks like [in Figma](https://www.figma.com/file/1YM3jA9h4QGc4xLdyaXIJS/Stories-Sprint-1.7?node-id=7524%3A9254).
6. Verify again generally that all the main actions work as expected, including removing a saved template and applying a template (saved or default). Applying a template should add it as a new Page without confirmation, removing a template should ask for confirmation.

### UAT

- [x] UAT should use the same steps as above.

This PR can be tested by following these steps:

1.
## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [x] This PR contains automated tests (unit, integration, and/or e2e) to verify the code works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [x] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #7065

PR #7290, opened 2021-04-23
Page templates: allow removing saved templates

miina:
## Summary

- Adds the option to delete saved templates.

## User-facing changes
N/A

## Testing Instructions
- Enable the custom page templates experiment.
- Switch to saved templates.
- Add an element to the page and Save it as template.
- Hover on the Saved template, verify the Trash icon appears as in Figma
- Verify that clicking on the Trash icon asks for confirmation and then deletes the Saved template.

### QA
See testing instructions above.

- [ ] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1.

### UAT

- [x] UAT should use the same steps as above.

This PR can be tested by following these steps:

1.

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [x] I have verified accessibility to the best of my abilities ([docs](https://github.com/google/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [x] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [ ] This PR contains automated tests (unit, integration, and/or e2e) to verify the code works as intended ([docs](https://github.com/google/web-stories-wp/tree/main/docs#testing))
- [ ] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #7062

PR #12209, opened 2022-09-02
Remove `objectWithout` usage from `elements` package

swissspidy:
## Context

I just noticed this is the only reason for the dependency on the `design-system` package. That util is not used anywhere else (the `story-editor` has its own copy).

## Summary

Remove `objectWithout` usage from `elements` package, removing the `design-system` package.

Fixes a bug in `duplicatePage` too, I think..?

Also removes a similar usage from the `patterns` package that I stumbled upon, making the code a bit more readable

## Testing Instructions

- [X] This is a non-user-facing change and requires no QA

1.

## Checklist

---

Fixes #


packages/elements/src/utils/duplicatePage.js
```diff
@@ -25,18 +21,15 @@ import createNewElement from './createNewElement';
 import duplicateElement from './duplicateElement';
 
 const duplicatePage = (oldPage) => {
-  // Remove title and postId for inserting the page.
+  // Remove title and templateId for inserting the page.
```
miina: I tried to test this part and can't see when would a template / post ID be assigned at all. Thought that perhaps when  inserting a page template and then duplicating it but doesn't seem like it. Were you able to test in which case we need to remove the title and ID when duplicating?

swissspidy: Indeed, doesn't look like that happens at all.

I tried inserting default page templates and custom saved page templates, but there's no `templateId` leaking into this.

Looking at `handlePageClick`, it's specifically omitted:

https://github.com/GoogleForCreators/web-stories-wp/blob/7f3af11a2fc2bc0e35733d1e79462f39918b2101/packages/story-editor/src/components/library/panes/pageTemplates/templateList.js#L72-L76

So yeah, seems like it's safe to remove this comment & check.

For context, originally introduce in #7290 and then adjusted in #7318

miina (approved):

spacedmonkey (approved):

packages/elements/src/utils/createNewElement.js
```diff
@@ -37,9 +36,11 @@ const createNewElement = (type, attributes = {}) => {
     type,
     id: uuidv4(),
   };
+
   // There's an exception for the background shape that should not get all the default attributes.
   if (attributes.isDefaultBackground) {
-    return objectWithout(newElement, ['backgroundColor']);
+    const { backgroundColor: _, ...newElementWithoutBgColor } = newElement;
```
barklund, line 42: Why rename it to `_` if unused anyway?

barklund (approved):
Looks good, makes sense to use the advanced language features.

packages/elements/src/utils/createNewElement.js
swissspidy: ESLint complained otherwise, surprisingly

timarney (approved):

PR #12217, opened 2022-09-05
Media: Remove media frame on close

spacedmonkey:
## Summary

While working on https://github.com/GoogleForCreators/web-stories-wp/pull/11941. I discovered that wp media instance when the wp media instance is closed. This instance may use memory in the browser and requires resetting. Just use remove method. 

See

https://github.com/WordPress/gutenberg/blob/67a652439c3d2f7dd8623d24b0bd12b5aff053f5/packages/media-utils/src/components/media-upload/index.js#L342-L344

## Testing Instructions

1.

## Checklist

---

Fixes #

### DIFF
--- a/packages/wp-story-editor/src/components/mediaUpload/mediaPicker/useMediaPicker.js
+++ b/packages/wp-story-editor/src/components/mediaUpload/mediaPicker/useMediaPicker.js
@@ -140,16 +140,11 @@ function useMediaPicker({
         onSelect(getResourceFromMediaPicker(mediaPickerEl));
       });
 
-      if (onClose) {
-        fileFrame.once('close', onClose);
-      }
-
-      fileFrame.once('content:activate:browse', () => {
-        // Force-refresh media modal contents every time it's opened
-        // to avoid stale data due to media items being upload & updated
-        // through the editor in the meantime.
-        fileFrame.content?.get()?.collection?._requery(true);
-        fileFrame.content?.get()?.options?.selection?.reset();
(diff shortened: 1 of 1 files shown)


swissspidy:
Any recommendations on how to verify this change? Should there be any changes in the DOM?

spacedmonkey:
> Any recommendations on how to verify this change (testing instructions)? Should there be any changes in the DOM?

No dom change here. It is more about reseting state. One way to test this is to open a wp dialog, search for something and close it. Open again. 

Discovery this need while looking at [this](https://github.com/WordPress/gutenberg/blob/e6dd3b19d186c61fa8894dc8fdfb4fa716e3b010/packages/media-utils/src/components/media-upload/index.js#L342-L344).

swissspidy:
> One way to test this is to open a wp dialog, search for something and close it. Open again.

I don't see a difference when doing that. What should be different?

spacedmonkey:
> I don't see a difference when doing that. What should be different?

In my PR this, is no longer required. 

https://github.com/GoogleForCreators/web-stories-wp/blob/f2cf9b8505c535165b7d96e8a555f60762b40bc1/packages/wp-story-editor/src/components/mediaUpload/mediaPicker/useMediaPicker.js#L279-L284

As removing the the wp.media, means state is reset and no chance of state set in the first instance of using the dialog is not used.

swissspidy:
I suppose that explains why there is no difference: there should be none.

swissspidy (approved):

PR #12224, opened 2022-09-05
Code: Converted `date` package to TypeScript

### LINKED ISSUE #12181: TypeScript: Convert `date` package

barklund:
## Context

Fairly complex type conversion requiring adding a mock type for an untyped module and a pretty funky map key extraction.

## Testing Instructions

## Checklist

---

Fixes #12181


packages/date/buildLocalizeFn.d.ts
```diff
@@ -0,0 +1,6 @@
+declare module 'date-fns/locale/_lib/buildLocalizeFn' {
+  type LocalizeFunction = (...args: any[]) => any;
+  type LocalizeFunctionBuilder = (args: any) => LocalizeFunction;
+  const localizeFunctionBuilder: LocalizeFunctionBuilder;
+  export default localizeFunctionBuilder;
+}
```
barklund: @swissspidy, is this best-practice? Not really sure, but it seems to work and is about as "well-typed" as the regular `localize` function from that package.

packages/date/src/convertFormatString.ts
```diff
@@ -211,15 +209,23 @@ const formatMap = {
   U: 't',
 };
 
+type FormatMap = typeof formatMap;
+type DateKey = keyof FormatMap;
+type ObtainKeys<Obj, Type> = {
+  [Prop in keyof Obj]: Obj[Prop] extends Type ? Prop : never;
+}[keyof Obj];
+type DateStringKey = ObtainKeys<FormatMap, string>;
+type DateFunctionKey = ObtainKeys<FormatMap, (date: Date) => string>;
```
barklund: 🤯

packages/date/src/formatDistance.ts
```diff
@@ -19,51 +19,62 @@
...
+  count: number,
+  options: Options
+) {
   options = options || {};
 
   const result = formatDistanceLocale[token](count);
 
   if (options.addSuffix) {
-    if (options.comparison > 0) {
+    if ((options.comparison || 0) > 0) {
```
barklund, line 77: This was required by TypeScript. Not sure why, as `undefined > 0` returns `false`, which is just fine here.

packages/date/src/getOptions.ts
```diff
@@ -124,20 +124,19 @@ function getOptions() {
   const localize = {
     ...originalLocale.localize,
 
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- Required workaround
```
barklund: The _unsafe assignment_ refers to the argument passed to the function. I could update the argument type in the module definition so this passes, but it doesn't seem worth it, as it won't actually represent the true underlying function anyway?

packages/date/src/settings.ts
```diff
@@ -14,15 +14,32 @@
...
 
+type WeekdayIndex = 0 | 1 | 2 | 3 | 4 | 5 | 6 | undefined;
+
 export const DEFAULT_DATE_SETTINGS = {
   dateFormat: 'F j, Y',
   timeFormat: 'g:i a',
   gmtOffset: 0,
   timezone: '',
-  weekStartsOn: 0,
+  weekStartsOn: 0 as WeekdayIndex,
```
barklund, line 24: This is required because 🤷

packages/date/buildLocalizeFn.d.ts
barklund: Maybe we want this file in a different location? Or with a different filename?

packages/date/src/convertFormatString.ts
```diff
@@ -211,15 +209,26 @@ const formatMap = {
...
+type FormatMap = typeof formatMap;
+type DateKey = keyof FormatMap;
+type ObtainKeys<Obj, Type> = {
+  [Prop in keyof Obj]: Obj[Prop] extends Type ? Prop : never;
+}[keyof Obj];
+type DateStringKey = ObtainKeys<FormatMap, string>;
+type DateFormatter =
+  | ((date: Date) => string | number)
+  | (() => string | number);
+type DateFunctionKey = ObtainKeys<FormatMap, DateFormatter>;
```
barklund, line 221: This is quite the mess, but it works! 🤯

packages/date/buildLocalizeFn.d.ts
swissspidy: That is a great question. I don't know anymore why / how I ended up using this function but apparently it's still not public and thus has no types or anything (see https://github.com/date-fns/date-fns/issues/1116)

So this seems like a reasonable workaround.

In other packages I've put this kinda stuff in a dedicated `typings` folder, which should work here as well. E.g. `packages/date/src/typings/date-fns/index.d.ts`. But open to other suggestions.

packages/date/src/settings.ts
```diff
@@ -14,15 +14,32 @@
...
+  gmtOffset: number;
+  timezone: string;
+  weekStartsOn: WeekdayIndex;
+} & Partial<{
+  locale: string;
+  months: string[];
+  monthsShort: string[];
+  weekdays: string[];
+  weekdaysShort: string[];
+  weekdaysInitials: string[];
```
miina: Curious why not just mark all these optional?

packages/date/src/getOptions.ts
miina: Hm, looks like eslint claims this rule was never reported though?
`eslint-comments/no-unused-disable] '@typescript-eslint/no-unsafe-assignment' rule is disabled but never reported.`

packages/date/src/convertFormatString.ts
```diff
@@ -211,15 +209,26 @@ const formatMap = {
...
 };
 
+type FormatMap = typeof formatMap;
+type DateKey = keyof FormatMap;
+type ObtainKeys<Obj, Type> = {
+  [Prop in keyof Obj]: Obj[Prop] extends Type ? Prop : never;
+}[keyof Obj];
+type DateStringKey = ObtainKeys<FormatMap, string>;
+type DateFormatter =
+  | ((date: Date) => string | number)
```
miina, line 219: What does the pipe symbol in front of the line do? 🤓

miina:
Looks like some JS tests are failing now.

packages/date/src/getOptions.ts
barklund: I'll probably fix the mock type so we can avoid this completely.

packages/date/src/settings.ts
barklund: I've been a bit back and forth on the types, in the current iteration `Partial` is a bit overkill yes.

packages/date/src/convertFormatString.ts
barklund: Apparently that's [how TypeScript multilines a union type](https://github.com/prettier/prettier/issues/3986) 🤷. Looks weird, gives me [SML vibes](https://en.wikipedia.org/wiki/Standard_ML#Declarative_definitions).

packages/date/buildLocalizeFn.d.ts
barklund: Moved in b07750287e.

packages/date/src/getOptions.ts
barklund: Fixed in b07750287e.

packages/date/src/settings.ts
barklund: Fixed in b07750287e.

packages/date/src/getTimeZoneString.ts
```diff
@@ -22,23 +22,30 @@ import { getSettings } from './settings';
...
+    ? Math.floor(absoluteOffset)
+    : Math.floor(absoluteOffset / 60);
 
-  return offsetInHours < 10
-    ? `${sign}0${absoluteOffset}`
-    : `${sign}${absoluteOffset}`;
+  const hoursAsString = String(offsetInHours).padStart(2, '0');
+  const minutesAsString = String(offsetInMinutes).padStart(2, '0');
+
+  return `${sign}${hoursAsString}:${minutesAsString}`;
```
barklund: Completely rewrote this to work correctly with both low and high numbers, including or excluding minutes, both negative and positive offsets. See tests for coverage.

miina: @swissspidy Can you confirm the change here just in case? I think we originally took this logic based on how WP shows the UTC timezone difference, it's used in the calendar only (currently), I believe:
Screenshot 2022-09-06 at 13 47 49

miina (approved):
✏️

packages/date/src/getTimeZoneString.ts
barklund: Oh yeah, that looks a bit wrong. We could add an argument about whether minutes are included? Or make sure we can display it including minutes (some actual areas do have non-integer hour offsets).

barklund:
[What do you mean](https://github.com/GoogleForCreators/web-stories-wp/runs/8204136202?check_suite_focus=true#step:13:8) `Error: ESLint issues detected`? Which issues? You're useless!

barklund:
> [What do you mean](https://github.com/GoogleForCreators/web-stories-wp/runs/8204136202?check_suite_focus=true#step:13:8) `Error: ESLint issues detected`? Which issues? You're useless!

Manually running ESLint locally says:

```
<root>/packages/date/src/typings/date-fns/index.d.ts
  1:1  error  missing header  header/header
```

However, it is not highlighted as an error inside the file in neither the editor nor here in Github.

@swissspidy, any idea why ESLint wants this header in this file, but doesn't actually want to tell you that it does? Do we need to add the `.d.ts` extension to some list somewhere?

packages/date/src/getTimeZoneString.ts
swissspidy: Thanks for pointing this out.

So right now I have `GMT+2` in the date picker, and this change here would turn this into `GMT+02:00`?

Including minutes seems reasonable _if there is an offset_. Otherwise `+2` is more compact.

----

Aside: WordPress shows the timezone abbreviation instead of the offset if it exists:

Screenshot 2022-09-06 at 10 06 03

I'll create a ticket to match that.

swissspidy:
That's a great question.

I suspect the annotation script has a bug / troubles adding an annotation when it's not affecting a node in the AST but basically the whole file.

But I do get the warnings in my editor:

Screenshot 2022-09-06 at 08 17 38

packages/date/src/getTimeZoneString.ts
barklund: I've changed it in 27d60baa5a so that minutes are only displayed if non-zero. Can you both check that this looks good in the calendar?

barklund: So now it displays what you've input in the system general settings for your WP installation. See table below

| Timezone Setting in WP | Calendar |
|-|-|
| `"Copenhagen"` (my actual timezone as reported by my computer's clock) | Screen Shot 2022-09-06 at 21 56 16 | 
| `"New York"` | Screen Shot 2022-09-06 at 21 52 20 |
| `"UTC+6"` | Screen Shot 2022-09-06 at 21 59 14 |
| `"UTC+5:45"` | Screen Shot 2022-09-06 at 21 58 41 |
| `"UTC-4"` |  Screen Shot 2022-09-06 at 21 52 37 |
| `"UTC-3:30"`| Screen Shot 2022-09-06 at 21 55 45 |

I guess we can all agree that the last one is a bug, which I'll look at, but otherwise it looks ok, no?

barklund: Hm, can't recreate that last one, must have selected `-04:30` because it works correctly now. Please ignore that 😁.

barklund:
> I suspect the annotation script has a bug / troubles adding an annotation when it's not affecting a node in the AST but basically the whole file.
>
> But I do get the warnings in my editor.

I don't get any ESLint error indication anywhere (in VSCode) for this particular file, but I do for any (other) `*.js`, `*.ts`, or `*.tsx` file.

packages/date/src/getTimeZoneString.ts
swissspidy: We could perhaps ditch the leading zero again, but otherwise looks good 👍

barklund: So, I didn't actually add that leading zero. It was always there. The old code was this: https://github.com/GoogleForCreators/web-stories-wp/blob/5de647c3456ab5cd021a252b6c4873e5996f5c2b/packages/date/src/getTimeZoneString.js#L33-L35

This was pretty broken though, as it didn't take negative offsets into consideration (`-10 < 10` so it would become. `"-010"` 🤷 ).

Your example of `"GMT+2"` has nothing to do with this function actually. If this code was used, it would have said `"UTC"`, not `"GMT"`. So you must have selected `"GMT+2"` in the WP settings somehow as the named timezone of your system, rather than an offset as a number.

swissspidy: Oh, interesting! I must check my settings then.

Then let‘s ship this 👍

Issue #12090, opened 2022-08-09
TypeScript: Create new story package with typings


swissspidy:
These will end up being used by almost all packages and thus cannot really live within `story-editor` because of circular dependency issues.

For example `getStoryMarkup()` in the output package needs it as well (which is a dependency of `story-editor`). See #12089

swissspidy:
That is a good point 👍 

Not sure why I ended up marking this one as being blocked by `animation`  🤔 

I guess I was mostly looking at the `package.json` dependencies and the like and didn't see that the `animation` had so many implicit dependencies on elements, element boxes and resources etc. 

If we only include types in this package, I suppose we could start calling it `@googleforcreators/types` for now and then go from there.

PR #12233, opened 2022-09-06
Migrate `rich-text` package to TypeScript

### LINKED ISSUE #12179: TypeScript: Convert `rich-text` package

miina:
## To-do

- [x] The whole State-Provider-Context part doesn't seem right and only works because of `as State` casting right now. Looks like it generally needs some changes to allow selecting the whole state without a selector as well.
- [x] `renderMap` doesn't have the correct type, for matching the type, we'd need to change the logic to use `Immutable.Map`, however, it works currently as well so I'm wondering if `draft-js` types are not complete.

## Testing Instructions

1.

## Checklist

---

Fixes #12179


packages/rich-text/src/usePasteTextContent.ts
```diff
@@ -32,9 +32,17 @@ import customExport from './customExport';
...
  * in text edit-mode.
  */
-function usePasteTextContent(insertElement, preset) {
+// @todo this really returns the inserted element and props is Partial<Element>, so is preset.
+function usePasteTextContent(
+  insertElement: (
+    type: string,
+    props: Record<string, unknown>,
+    asBackground?: boolean
+  ) => Record<string, unknown>,
```
miina, line 41: Not sure if it makes sense to create a local type `Element` or add a todo and leave it as is until the elements library package to get converted 🤔

packages/rich-text/src/editor.tsx
```diff
@@ -18,22 +18,30 @@
...
+  onChange: (content: string) => void;
+}
+
+function RichTextEditor(
+  { content, onChange }: RichTextEditorProps,
+  ref: ForwardedRef<HTMLElement>
+) {
   const editorRef = useRef(null);
   const {
     state: { editorState },
```
miina, line 47: @barklund Do you have any thoughts on what's the correct typing approach here with the `useRichText`, state and the `useContextSelector`: https://github.com/GoogleForCreators/web-stories-wp/blob/910c86e4114ced8b5e80b5abfec76627e6874e54/packages/react/src/useContextSelector.ts#L47-L51

Since `useContextSelector` returns `T | undefined`, it complains about `state` and `action` not existing in the return value of `useRichText`.
Additionally, using just `useRichText()` is complaining about not having enough arguments, however, making the `selector` optional in `useRichText` in will cause new type errors with using `useContextSelector` inside it, so I'm wondering if `useContextSelector` is still not with correct types.

Let me know if you have any insights here 🤓

barklund: If we change `useContextSelector` line 56 to:

```ts
    if (ref.current && equalityFn(ref.current, selected)) {
```

This will change the return to `S` rather than `S | undefined`. I think that's a reasonable change.

If it's `falsey` (most likely some simple value), we don't actually care about equality anymore. This won't change any functionality as far as I can see, but would make typings easier.

@swissspidy, WDYT?

packages/rich-text/src/utils/getCaretCharacterOffsetWithin.ts
```diff
@@ -64,32 +87,22 @@ function getCaretCharacterOffsetWithin(element, clientX, clientY) {
...
+        element.textContent &&
+        preCaretRange.toString().length <= element.textContent.length
+      ) {
         return preCaretRange.toString().length;
       }
-      return 0;
-    }
-  }
-
-  const sel = doc.selection;
```
miina, line 83: Removed this part since it seemed that it was for IE only that we don't support.

packages/rich-text/src/usePasteTextContent.ts
miina: Edit: probably makes sense to wait for #12090

packages/rich-text/src/editor.tsx
miina: When returning just `S`, it still doesn't allow using the context without props, e.g. just `useRichText()` gives an error of expecting at least 1 argument, making the selector optional, however, generates new type errors. So I'm wondering if `identity` also still needs adjustments, too, to allow using it without `selector`.
https://github.com/GoogleForCreators/web-stories-wp/blob/80a5cc6f593e7d8f37f656dd508759774dc3d5c6/packages/react/src/context.ts#L19

miina:
@barklund There are still 2 type errors in `getPastedBlocks.ts` file that I haven't figured out + the whole provider-context thing might still change but would appreciate if you could review this otherwise already.

packages/rich-text/src/editor.tsx
barklund: Hmm, I'll check it out. It should definitely be possible to use it without a selector defaulting to just returning the entire state in a type-safe way. Just to be clear, you could add in the identity selector manually and it would work, right?

miina: Yes, when adding a selector, it seems to work.
The current version of the PR also works now without TS errors (in this file) but because of casting only.

packages/rich-text/src/getPastedBlocks.ts
```diff
@@ -47,20 +58,34 @@ class CustomBlockRenderMap {
...
+type ConvertFromHTMLReturn = {
+  contentBlocks: ContentBlock[];
+  entityMap: Record<string, unknown>;
+};
+function getPastedBlocks(html: string, existingStyles: string[] = []) {
   const renderMap = new CustomBlockRenderMap();
   const { contentBlocks, entityMap } = convertFromHTML(
     html,
     undefined /* This has to be undefined to trigger default argument */,
     renderMap
```
miina: @barklund Not sure how to fix this TS error, from the `draft-js` [examples](https://draftjs.org/docs/advanced-topics-custom-block-render-map/), I can see that the `renderMap` could be created just by `Immutable.Map(mapObject)` instead of creating a custom class for it, however, I assume that it's not done like this currently not to require `immutabe`.

When using `Immutable.Map`, TS doesn't show any errors and it works, however, since it works currently as well then perhaps the types are just not correct in `draft-js`?

Let me know if you have any thoughts on how to fix this type.

barklund: Technically the types are not correct, as the only method used on the map object is the mapKeys method, so they could relax their types to just require an object with that single method.

However, because draft-js already includes this piece from immutable, we can do so too - I won't increase bundle size (if you only include that thing from immutable). So I do recommend changing this to an Immutable.Map instance instead.

packages/rich-text/src/types.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export interface RichTextProviderProps {
+  children: ReactNode;
+  editingState: EditingState;
+}
+
+export interface RichTextEditorProps {
+  content: string;
+  onChange: (content: string | null) => void;
+}
```
barklund: I feel that props types belong next to the components they're the props types for - unless of course they're used in multiple places. Just for co-location purposes.

packages/rich-text/src/context.ts
```diff
@@ -18,7 +18,16 @@
...
+ */
+import type { State } from './types';
 
-const RichTextContext = createContext({ state: {}, actions: {} });
+// @todo Not sure how to use Partial correctly here to work with empty state and context selector.
+// Should make all state and action values optional as before?
+const RichTextContext = createContext<State>({
+  state: {},
+  actions: {},
+} as State);
```
barklund: I think we should "properly" initialize the state to empty functions and variables, like so:

```ts
const RichTextContext = createContext<State>({
  state: {
    editorState: null,
    hasCurrentEditor: false,
    selectionInfo: undefined,
  },
  actions: {
    setStateFromContent: () => undefined,
    updateEditorState: () => undefined,
    getHandleKeyCommand: () => () => 'not-handled',
    handlePastedText: () => 'not-handled',
    clearState: () => undefined,
    selectionActions: {},
    getContentFromState: () => null,
  },
});
```

Then all would be happy.

packages/rich-text/src/customExport.ts
```diff
@@ -40,13 +41,15 @@ function inlineStyleFn(styles) {
...
+function exportHTML(editorState: EditorState) {
   if (!editorState) {
     return null;
   }
 
   const html = stateToHTML(editorState.getCurrentContent(), {
     inlineStyleFn,
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment -- See the next line for reason.
+    // @ts-ignore Reason: `defaultBlockTag` is optional (undefined | string), however, `undefined` means `p` so we need to use `null` here.
     defaultBlockTag: null,
```
barklund: This has [already been updated upstream](https://github.com/sstur/draft-js-utils/pull/215), so we should probably just update the package here. Ping @swissspidy

packages/rich-text/src/customImport.ts
```diff
@@ -18,22 +18,25 @@
...
-    .map(({ elementToStyle }) => elementToStyle(element))
+    .map(({ elementToStyle }) => elementToStyle(element as HTMLElement))
     .filter((style) => Boolean(style));
 
   if (styleStrings.length === 0) {
     return null;
   }
 
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment -- The defined type is `string` but it actually accepts arrays of strings, too.
+  // @ts-ignore
```
barklund, line 39: I went pretty far down the rabbit hole on this one - pretty hard to track down, and I don't even think I got all they way. But basically: I don't know why this works, but apparently it does. It shouldn't and the types seem correct, but for some reason it does work, so let's just keep this big flag here.

Given that draft-js is slowly dying, it's probably not a problem, but we could potentially face a situation where this is "fixed" upstream and thus not allowed anymore.

But for now 🤷 . This ignore is fine.

packages/rich-text/src/editor.tsx
```diff
@@ -95,7 +100,8 @@ function RichTextEditor({ content, onChange }, ref) {
     <Editor
       ref={editorRef}
       onChange={updateEditorState}
-      editorState={editorState}
+      /* We return before if EditorState is null so we know it's not null here */
+      editorState={editorState as EditorState}
```
barklund: If you just change the definition of `hasEditorState` above to:

```ts

const hasEditorState = editorState !== null;

```


This cast should not be required anymore.


There is a slight difference in the two expressions (`Boolean(editorState)` vs `editorState !== null`), in case the `editorState` has a `valueOf()`-method that returns `false`, but it doesn't so we're good here.

packages/rich-text/src/editor.tsx
barklund: Okay, I figured it out. We have to create useRichText with [manual overloads](https://www.typescriptlang.org/docs/handbook/functions.html#overloads), so we have proper types for the no-argument version. No (further) changes required in the `react` package:

```ts
function useRichText(): State;
function useRichText<T>(selector?: (state: State) => T) {
  return useContextSelector(Context, selector ?? identity<State>);
}
```

That's pretty freaking nice! And it just works. The first line(s) in an overload define(s) the types for what happens when you call it with other arguments (types and/or number of) and the last definition is the actual implementation. The implementation is of course tested against the overloaded definitions to check that they actually match.

Then the cast isn't necessary anymore!

At least I think this works. We don't actually invoke `useRichText` **with** a selector function anywhere in this package, so 🤷 .

packages/rich-text/src/fauxSelection.ts
```diff
@@ -80,12 +89,15 @@ export function useFauxSelection(editorState, setEditorState) {
...
+      setEditorState((oldEditorState: EditorState | null) => {
         try {
+          if (!oldEditorState) {
+            return null;
+          }
           // Get new content with style removed from old selection
           const contentWithoutFaux = Modifier.removeInlineStyle(
             oldEditorState.getCurrentContent(),
-            fauxSelection,
+            fauxSelection as SelectionState,
```
barklund: Again here, the definition of `hasFauxSelection` just needs to be changed to `fauxSelection !== null` rather than `Boolean(fauxSelection)` and this cast isn't required.

packages/rich-text/src/fauxSelection.ts
```diff
@@ -117,12 +129,15 @@ export function useFauxSelection(editorState, setEditorState) {
   }, [fauxSelection, editorState, setEditorState]);
 }
 
-export function fauxStylesToCSS(styles, css) {
+export function fauxStylesToCSS(
+  styles: DraftInlineStyle,
+  css: Record<string, string>
```
barklund: Couldn't this be typed as `CSSProperties` too?

packages/rich-text/src/formatters/convert.ts
```diff
@@ -62,15 +72,15 @@ function convertStyles(contentBlock, blockSelection, contentState) {
...
   UNDERLINE: UNDERLINE,
 };
 
-function getNewStyles(oldStyles) {
+function getNewStyles(oldStyles: string[]): string[] {
   return oldStyles
-    .map((oldStyle) => (oldStyle in styleMap ? styleMap[oldStyle] : null))
+    .map<string>((oldStyle) => styleMap[oldStyle] ?? null)
     .filter(Boolean);
 }
```
barklund: This could be done as:

```ts
const styleMap = {
  BOLD: weightToStyle(700),
  ITALIC: ITALIC,
  UNDERLINE: UNDERLINE,
};
type StyleMapKey = keyof typeof styleMap;

function getNewStyles(oldStyles: string[]): string[] {
  return oldStyles
    .map((oldStyle) =>
      oldStyle in styleMap ? styleMap[oldStyle as StyleMapKey] : ''
    )
    .filter(Boolean);
}
```

I don't know if I like it more, but it does have less funny stuff going on, so perhaps a bit cleaner? I don't know why that last cast is required, because if it's in the object, it must match one of the keys, so not too sure about that.

packages/rich-text/src/formatters/color.ts
```diff
@@ -40,11 +42,13 @@ import { isStyle, getVariable } from './util';
...
-const styleToColor = (style) => getSolidFromHex(getVariable(style, COLOR));
+const styleToColor = (style: string): Pattern =>
+  getSolidFromHex(getVariable(style, COLOR));
 
-const colorToStyle = (color) => `${COLOR}-${getHexFromSolid(color)}`;
+const colorToStyle = (color: Solid): string =>
+  `${COLOR}-${getHexFromSolid(color)}`;
 
-function elementToStyle(element) {
+function elementToStyle(element: HTMLElement): string | null {
```
barklund, line 51: (adding this comment here because I can't in `formatters/index.ts`)

I feel like the whole formatters array needs a proper type - something like:

```ts
const formatters: Formatter[] = [
  ...
];
```

That will make using this array a bit cleaner, as it currently is a weird inferred intersection type, that doesn't really guarantee anything about the members in general, only the specific members that are extracted.

The `Formatter` interface would be something like:

```
interface Formatter {
  elementToStyle: (element: HTMLElement) => string | null;
  stylesToCSS: (styles: DraftInlineStyle) => CSSProperties | null;
  getters: Record<string, Getter>;
  setters: Record<string, Setter>;
  autoFocus: boolean;
}
```

This does require extra work of course to make sure it complies with that, but it makes several of the instances of using the `formatters` array a lot cleaner.

Note that the `stylesToCSS` function inside `uppercase.ts` has to be amended to `{ textTransform: 'uppercase' as const }` to pass as a valid CSSProperties object.

tsconfig.shared.json
```diff
@@ -6,7 +6,7 @@
     "jsx": "preserve",
     "target": "esnext",
     "module": "esnext",
-    "lib": ["dom", "esnext"],
+    "lib": ["dom", "dom.iterable", "esnext"],
```
barklund, line 9: @swissspidy, this seems logical. Do you see any reason why it shouldn't be there?

barklund (commented):
Various comments and suggestions for optimization. Most importantly I feel is typing the generic formatter, so they all conform to a common interface.

packages/rich-text/src/customExport.ts
swissspidy: Looks like there hasn't been a new release of `draft-js-export-html` [since 2019](https://www.npmjs.com/package/draft-js-export-html), so we can't really update it. But we can fork it, like I forked https://github.com/swissspidy/draft-js-import-element

tsconfig.shared.json
swissspidy: Haven't fully looked into [the `lib` option](https://www.typescriptlang.org/tsconfig#lib) but yeah this seems reasonable

packages/rich-text/src/formatters/convert.ts
miina: > I don't know why that last cast is required, because if it's in the object, it must match one of the keys, so not too sure about that.

Yep, for some reason it's required when mapping, just saw this before elsewhere, too.

Sure, I'll go with your version, perhaps a bit cleaner indeed.

packages/rich-text/src/editor.tsx
miina: Uh, neat! I just read about the function overloads a few days ago but didn't make the connection with this issue at all. Thanks! 🧙

packages/rich-text/src/customExport.ts
barklund: Damn, that's a long time ago. A fork sounds reasonable until we can move on to [facebook/lexical](https://github.com/facebook/lexical), when it goes to stable at some point in the hopefully _not-too-distant_ future.

packages/rich-text/src/editor.tsx
miina: @barklund I've added the function overload, now there's no need for casting `useRichText()` 🎉 
However, using
```
function useRichText<T>(selector?: (state: State) => T) {
  return useContextSelector(Context, selector ?? identity<State>);
}
```
still complains:
Screenshot 2022-09-12 at 17 44 29

Let me know if you changed anything else when looking into this since it worked for you.

barklund: Hmm, yeah I didn't get this error. I'll check it out and see what I perhaps did differently?

barklund: Hmm, I can't get it to compile now either. I swore that it worked (or at least seemed to work), but that probably says more about VSCode's flaky TS compilation than the code being correct. I don't know how to correctly type this then, but I'm sure it's possible! I'll ask around.

In the meantime, I guess we have to bite the bullet and make two different functions:

```ts
export function useFullRichText(): State {
  return useContextSelector(Context, identity);
}

export function useRichText<T>(selector: (state: State) => T) {
  return useContextSelector(Context, selector);
}
```

It's a lot less elegant, but for now I guess it'll do...?

barklund: The whole `Partial<State> | State` return type won't work, because we don't always return partial states from our selectors. We might select a single property or even computed properties, e.g. just `useRichText(({ state: { something } }) => something);`.

packages/rich-text/src/formatters/color.ts
miina: Added the `Formatter` interface but with `setters` as `Record<string, unknown>`. Couldn't figure out a better type currently since the setters accept different types as args.

packages/rich-text/src/customExport.ts
miina: Hm, looks like the `inlineStyleFn` used above also gives a TS error, it doesn't seem to be among the accepted options at all 🤔 

Have to check why / how / if it works as expected currently.

miina: @barklund / @swissspidy If you have a moment, could one of you try if you use [this forked version of draft-js-export-html](https://github.com/miina/draft-js-export-html) locally if the `npm install` runs OK and the module is imported fine? 
It fails with deps errors for me and when running with `force`, the module is not resolved. Not sure if it's just a local problem or not. I saw the same error with the repo forked by Pascal before as well so maybe it's me 🤷

packages/rich-text/src/editor.tsx
barklund: I think this might actually work? It compiles and the no-parameter version in `editor.tsx` is working.

```ts
function useRichText(): State;
function useRichText<T>(selector: (state: State) => T): T;
function useRichText<T>(selector: (state: State) => T | State = identity) {
  return useContextSelector(Context, selector);
}
```

It also seems to work in [TS Playground](https://www.typescriptlang.org/play?ssl=32&ssc=2&pln=28&pc=1#code/JYOwLgpgTgZghgYwgAgMpjpZBvAUMg5OALmRAFcBbAI2gG59DrSBnMKUAcwYF9dRIsRCgDCAe3AQAHmAA8qAHw5GBAG5wANuQilUvXLgQS2ySgE9xkmaUuQZ8jJCUBeHMnVadbksgCMAGmRmZAAiACYQnh4GXBhyEAQwYAlkchYIW2kwVAgNCESxKFkAFUDFAAojKzAbCTs5YoVA9LyCqFJyj21SYoBKZGclVF7dZUJkKAgwcigQZBb8sELKuqyAOi6IXv1Y+MTkueAAEwhwYDAzEoVOnpHkYrHCSenZ9x2Aenf71GRQZDAAO5iZBxBJJYy7MEHVLpABi5A0GgASsAEAALYpZcp3dCYFB4AjPGZzNIZVYyHKtJZQcrmTIyQLHU5JC7bXB8UH7FKkgAKcCgSU0KPRmPsjXKCzaHTYeN0ji2AyUfUeRNepPp2Vyi2WdPJYGaWrabL4uE+31+cwkKDEqmgGjEcBORxBe3BIEhXJJ6WFGKxOPlDE5bphEB9ooa10l1Ol8rleP6g3ud2Kgdd0NJYayVwlhujyAlsbQ8oTSuQAB8i3iBr8TmdWSqpsSQxrKdqabrqgaqYVjQYzQARCCUYFpOCcFAA85o+a5wqGYxgIi+au8-mC5Go30ycrlbBEZA8EtENlVEzUZeuDOb8M7vfUA9H6hs01fQfDmFjidTsTkRdRuenoucBhCucIIhuIp+mscAMIBQQgZe3rXlB1B0EAA).

packages/rich-text/src/customExport.ts
swissspidy: The GitHub repository needs to contain the built files, not just the source files.

Briefly did that in https://github.com/swissspidy/draft-js-export-html and 54e12c6074332e01d28e1a1f5d6b62b15c454f18

Let's see how that goes.

packages/rich-text/src/editor.tsx
miina: TY! Using that for now.

barklund (approved):
Looks good to me. It's not perfect, but we're still learning, and the suboptimal bits are only internal to this package, so they shouldn't leak through to the rest of the app anyway.

Issue #11973, opened 2022-07-20
Stickers: Improve bundle size


swissspidy:
## Feature Description

The stickers package contains a ton of stickers from all our templates, which is also reflected in bundle size:

<img width="881" alt="Screenshot 2022-07-20 at 17 43 26" src="https://user-images.githubusercontent.com/841956/180037097-ac992445-12ca-4ae9-ad50-dd5931948b5b.png">

It would be nice if we could somehow find a way to improve that, either by splitting things up into more chunks or even only lazy-loading stickers with dynamic imports / `React.lazy`

## Additional Context

* #11972
* #9858
* #11615

Issue #11972, opened 2022-07-20
Masks: Improve bundle size


swissspidy:
## Feature Description

The `constants.js` file in the masks package is a whopping 300KB large. This also shows in bundle analyzer:

<img width="1013" alt="Screenshot 2022-07-20 at 17 43 18" src="https://user-images.githubusercontent.com/841956/180034599-dea51364-7f99-4333-bd78-fa4d9b52ef3f.png">

That makes sense, because it contains a ton of clip paths.

I think the way the masks package is structured can be improved though to make this more efficient.

For example, in most places we only use functions like `canSupportMultiBorder`, `generateMaskId` or simple constants like `MaskTypes`

The full list of clip paths is only used in the shapes pane in the library:

https://github.com/GoogleForCreators/web-stories-wp/blob/22cf54d030e407f469350f21273f6d1c223aa99b/packages/story-editor/src/components/library/panes/shapes/shapesPane.js#L24

We should be able to improve performance here by reducing the bundle size using the following to steps:

1. Create a separate export in the masks package just for the clip paths
2. Dynamically import that separate definition in the shapes pane

swissspidy:
> We should be able to improve performance here by reducing the bundle size using the following to steps:
> 
> 1. Create a separate export in the masks package just for the clip paths
> 2. Dynamically import that separate definition in the shapes pane

That's easier said than done, because utils like `getMaskByType()` and `getElementMask` obviously rely on `MASKS` as well.

PR #12237, opened 2022-09-06
Dashboard: Remove elements dependency

swissspidy:
## Context

We do not render stories or templates in the dashboard anymore. We just display preview images. So there's no need for us to actually register elements.

## Summary

Removes dependencies on `elements` and `element-library` from the dashboard package.

This should improve bundles a bit.

## User-facing changes

N/A

## Testing Instructions

1.

No

No

No

## Checklist

---

Fixes #

### DIFF
--- a/packages/wp-dashboard/package.json
+++ b/packages/wp-dashboard/package.json
@@ -30,8 +30,6 @@
     "@googleforcreators/dashboard": "*",
     "@googleforcreators/date": "*",
     "@googleforcreators/design-system": "*",
-    "@googleforcreators/element-library": "*",
-    "@googleforcreators/elements": "*",
     "@googleforcreators/fonts": "*",
     "@googleforcreators/i18n": "*",
     "@googleforcreators/media": "*",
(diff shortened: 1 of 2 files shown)


miina:
Screenshot 2022-09-07 at 13 27 54

I wonder why the increase 🤔

spacedmonkey:
> I wonder why the increase 🤔

Just looking at this myself. I think it because `assets/js/1303.js` was removed and now is part of the story editor package.

spacedmonkey:
assets/js/wp-story-editor.js being 1.34MB seems like very big. I wonder if we should look into breaking this up. I am little bit worried about the editor loading.

spacedmonkey (approved):
Seems good to me. But I would like to follow up and look into the bundle size of the editor, which at 1.34mb seems a little large.

swissspidy:
Dashboard and editor have a common bundle for shared code.

The main reason is the stickers package, which is a dependency of `element-library`.

With this PR, stickers are no longer included for the dashboard and thus no longer in the common bundle but only in the editor bundle.

So that all makes sense.

There is already a ticket to optimize the way stickers are included, see #11973 

See also #11972 for the masks package

PR #12244, opened 2022-09-07
Videos: Add option to change volume

### LINKED ISSUE #12211: Media: Allow changing video volume
## Feature Description

Allow changing video volume, perhaps with a simple range slider.

Volume can be passed via the `volume=""` attribute on `video`/`amp-video` (number between 0.1 and 1.0.)

Let's put something together behind a flag & then get UX feedback on it.

Of course with unit & Karma tests etc.

## Additional Context

* `1.0` is the default value, so in other words this can only be used to _reduce volume_ of a video. For increasing volume we'd need to use FFmpeg, which is a bit overkill for now.

spacedmonkey:
## Summary

Created behind a feature flag. 

Create number input, so you can set the volume. In future, I recommend a slider or something for UX. 

## User-facing changes

Screenshot 2022-09-08 at 14 03 13

## Testing Instructions

- Enable experiment. 
- Create story. 
- Insert a video. 
- Select style tab. 
- Use slider to set volume. 
- Save story. 
- Preview story. 
- Play video with audio. 
- Hear sound. 

1.

## Checklist

---

Fixes #12211


swissspidy:
> Create number input, so you can set the volume. In future, I recommend a slider or something for UX.

Let's use a slider already now, we have a slider component in the design system.

packages/element-library/src/video/display.js
```diff
@@ -46,6 +46,7 @@ function VideoDisplay({
     focalX,
     focalY,
     loop,
+    volume = 1.0,
```
swissspidy: We don't need this default value since `1.0` is the default anyway in HTML.

packages/element-library/src/video/output.js
```diff
@@ -32,7 +32,7 @@ function defaultForUndefined(value, def) {
 }
 
 function VideoOutput({ element, box, flags }) {
-  const { resource, loop, tracks } = element;
+  const { resource, loop, tracks, volume = 1.0 } = element;
```
swissspidy: We don't need this default value since `1.0` is the default anyway in HTML.

packages/element-library/src/video/trim.js
```diff
@@ -63,7 +63,16 @@ function VideoTrim({
...
-  const { poster, tracks, isBackground, scale, flip, focalX, focalY } = element;
+  const {
+    poster,
+    tracks,
+    isBackground,
+    scale,
+    flip,
+    focalX,
+    focalY,
+    volume = 1.0,
```
swissspidy: We don't need this default value since `1.0` is the default anyway in HTML.

packages/story-editor/src/components/panels/design/videoOptions/test/videoOptions.js
```diff
@@ -45,14 +52,24 @@ jest.mock('../../../../../app/currentUser', () => ({
...
     type: 'video',
-    resource: { posterId: 0, poster: '', alt: '' },
+    volume: 1.0,
+    resource: { posterId: 0, poster: '', alt: '', isMuted: false },
   };
   function arrange(...args) {
     return renderPanel(VideoOptions, ...args);
   }
 
+  beforeEach(() => {
```
swissspidy: Lint failure: this should come after `beforeAll`

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
```diff
@@ -136,6 +140,20 @@ function VideoOptionsPanel({ selectedElements, pushUpdate }) {
       </Row>
       {canMute && (
         <>
+          {enableVideoVolume && (
+            <Row>
+              <NumericInput
+                value={volume}
+                onChange={onChangeVolume}
+                min={0}
```
swissspidy: As per the ticket description, the min should be 0.1. Why? Because 0.0 technically means muted, but we mute videos separately using FFmpeg and the `muted` attribute.

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
```diff
@@ -136,6 +140,20 @@ function VideoOptionsPanel({ selectedElements, pushUpdate }) {
...
+                max={1.0}
+                isFloat
+                suffix={__('Volume', 'web-stories')}
+                aria-label={__('Volume', 'web-stories')}
+                data-testid="volume-input"
```
swissspidy: Let's not use data-testid if we can target elements differently, e.g. by aria-label

packages/story-editor/src/components/panels/design/videoOptions/test/videoOptions.js
```diff
@@ -74,4 +91,24 @@ describe('Panels/VideoOptions', () => {
     const panel = screen.getByText('Video Settings');
     expect(panel).toBeInTheDocument();
   });
+
+  it('should render volume input', () => {
```
swissspidy, line 83: A karma test would be great for this new change

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
spacedmonkey: No we can't sadly. aria-label / labels do not work with numberic inputs.

packages/story-editor/src/components/panels/design/videoOptions/test/videoOptions.js
spacedmonkey: What is the point of this on a none finalized version of UX?

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
spacedmonkey: I force the min value to be elsewhere.

packages/story-editor/src/components/panels/design/videoOptions/test/videoOptions.js
swissspidy: It will most likely remain a slider, so the test should not really change much.

packages/element-library/src/video/display.js
swissspidy: I see the default values are still there, what's the rationale for that? We should be able to just omit this unless I am missing something.

spacedmonkey:
Karma tests added.

packages/element-library/src/video/constants.js
```diff
@@ -49,6 +49,7 @@ export const defaultAttributes = {
   controls: false,
   loop: false,
   autoPlay: true,
+  volume: 1.0,
```
spacedmonkey:
```suggestion
  volume: 1,
```

packages/element-library/src/video/display.js
swissspidy: @spacedmonkey why was this comment resolved? It‘s still valid and I‘d appreciate a reply.

spacedmonkey: I changed it from 1.0 to 1 as requested.

swissspidy: My request was to remove it entirely.

packages/element-library/src/video/display.js
```diff
@@ -46,6 +46,7 @@ function VideoDisplay({
     focalX,
     focalY,
     loop,
+    volume = 1,
```
swissspidy:
```suggestion
    volume,
```

As per other discussion, this is basically what I meant. For a browser, `1` is the default volume anyway, so we don't need to specify it.

packages/story-editor/src/components/panels/design/videoOptions/test/videoOptions.js
```diff
@@ -74,4 +91,24 @@ describe('Panels/VideoOptions', () => {
...
+    arrange([defaultElement]);
+    expect(screen.getByLabelText('Volume')).toBeInTheDocument();
+  });
+
+  it('should render volume input as muted', () => {
```
swissspidy:
```suggestion
  it('should not render volume input if video is muted', () => {
```

packages/story-editor/src/components/panels/design/videoOptions/test/videoOptions.js
```diff
@@ -18,13 +18,20 @@
...
 import { screen } from '@testing-library/react';
-
+import { useFeature } from 'flagged';
 /**
  * Internal dependencies
  */
 import VideoOptions from '../videoOptions';
 import { renderPanel } from '../../../shared/test/_utils';
 import { useLocalMedia } from '../../../../../app/media';
+import mockUseVideoElementTranscoding from '../../../../../app/media/utils/useVideoElementTranscoding';
```
swissspidy: Unrelated, but why does `useVideoElementTranscoding` exist twice in the code base? We can delete one of the two.

packages/story-editor/src/components/panels/design/videoOptions/test/videoOptions.js
```diff
@@ -74,4 +91,24 @@ describe('Panels/VideoOptions', () => {
...
+  it('should render volume input', () => {
+    arrange([defaultElement]);
+    expect(screen.getByLabelText('Volume')).toBeInTheDocument();
+  });
+
+  it('should render volume input as muted', () => {
+    mockUseVideoElementTranscoding.mockImplementationOnce(() => ({
+      state: { canMute: false },
+      actions: {},
+    }));
```
swissspidy:
```suggestion
```

This is not really needed because the component only checks for `resource.isMuted`, not for `canMute`.

The test passes without this.

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
```diff
@@ -134,6 +152,27 @@ function VideoOptionsPanel({ selectedElements, pushUpdate }) {
...
+            htmlFor={slideId}
+          >
+            {__('Volume', 'web-stories')}
+          </Text>
+          <StyledSlider
+            value={Math.round(volume * 100)}
+            handleChange={onChangeVolume}
+            minorStep={5}
+            majorStep={10}
+            min={0}
```
swissspidy, line 169:
```suggestion
            min={10}
```

Otherwise it's impossible to get more to the left:

Screenshot 2022-09-13 at 10 26 06

packages/story-editor/src/components/panels/design/videoOptions/test/videoOptions.js
spacedmonkey: No idea. Out of scope of this work.

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
spacedmonkey: Yeah, I tried this and sadly it did not work, as it generated some weird UX. 

https://user-images.githubusercontent.com/237508/190159911-1d33addd-b725-4f5c-b05e-67b944c01498.mov


The slider is at the very end, but is still 10, that is confusing. If the slider is at the start, it should be 0. The way it was written, it would not let you set it to 0, but you could see that a little bit of the slider was across. 

Maybe we should just allow 0 volume. If set, either show a tooltip recommending you remove the audio or just trigger the removal of audio automatically.

packages/story-editor/src/components/panels/design/videoOptions/test/videoOptions.js
spacedmonkey: It was needed for an earlier version of the code. Remove now.

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
swissspidy: Either way is confusing right now, yeah, but I we'll probably replace the numbers with something else anyway. So don't worry about it for now. It's just the first PR of many. We'll finalize UX later.

spacedmonkey: I personally think that the way that it is now, is the best way. We can change it in a follow up PR.

swissspidy (approved):

packages/story-editor/src/components/panels/design/videoOptions/videoOptions.js
swissspidy: We'll definitely change this in a followup.

PR #12247, opened 2022-09-07
TypeScript: Convert `wp-utils` package

### LINKED ISSUE #12235: TypeScript: Convert `wp-utils` package

spacedmonkey:
## Summary

Convert wp-utils to typescript.

## Testing Instructions

1.

## Checklist

---

Fixes #12235


packages/wp-utils/src/bindToCallbacks.ts
```diff
@@ -13,11 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-const bindToCallbacks = (callbacks, config) => {
+
+function bindToCallbacks(callbacks: Record<string, unknown>, config: object) {
   return Object.entries(callbacks).reduce((_callbacks, [name, callback]) => {
     _callbacks[name] = callback.bind(null, config);
     return _callbacks;
   }, {});
```
barklund: I'd also rewrite this to `Object.fromEntries` as that's a much better and cleaner way to create dynamic objects rather than reduce. Something like:

```suggestion
  return Object.fromEntries(Object.entries(callbacks).map(
    ([name, callback]) => [name, callback.bind(null, config)]
  ));
```

packages/wp-utils/src/bindToCallbacks.ts
```diff
@@ -13,11 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-const bindToCallbacks = (callbacks, config) => {
+
+function bindToCallbacks(callbacks: Record<string, unknown>, config: object) {
```
barklund: Here I think generics are needed for the config object. We need the callbacks to be typed so that they are functions taking a parameter of the type that the config object has, hence the generic type. Something like this should work:

```suggestion
function bindToCallbacks<T>(callbacks: Record<string, (config: T) => unknown>, config: T):Record<string, unknown> {
```

Additionally, if this is only used in places, where all the callbacks return an object of the same type, the `unknown` can be replaced by another generic, e.g. just `S`.

packages/wp-utils/src/snakeToCamelCase.ts
```diff
@@ -37,11 +37,14 @@ export function snakeToCamelCase(string = '') {
...
+ * @param obj Object to be transformed.
+ * @param ignore Array of keys to be ignored for camelcase.
+ * @return Transformed object.
  */
-export function snakeToCamelCaseObjectKeys(obj, ignore = []) {
+export function snakeToCamelCaseObjectKeys(
+  obj: unknown,
+  ignore: Array<string> = []
+) {
   const isObject = (val) =>
```
barklund: This just need a type of `unknown`.

packages/wp-utils/package.json
```diff
@@ -26,5 +26,7 @@
   },
   "type": "module",
   "main": "./src/index.js",
```
barklund: This should point to `index.ts` now.

barklund (commented):
Great start, adding some pointer to getting the last bits over the line.

packages/wp-utils/package.json
spacedmonkey: Other modules do not do this.

packages/wp-utils/src/snakeToCamelCase.ts
spacedmonkey: I am not sure what you mean here?

packages/wp-utils/package.json
spacedmonkey: Other packages don't do this?

packages/wp-utils/src/snakeToCamelCase.ts
spacedmonkey: Added.

packages/wp-utils/src/bindToCallbacks.ts
spacedmonkey: Seems to work.

spacedmonkey:
I have one last error, than I think we are good.

packages/wp-utils/package.json
swissspidy: This package is a bit different from the others since it's private, and not published to npm. That's why it has different fields in `package.json`.

The published packages all point to `dist/index.js`, but here we don't have a `dist` folder and we don't have `index.js` anymore since the file extension is now `.ts`. It makes sense to change this to a file name that exists.

And since this is a private package only really used by webpack I don't think we even need to specify the `source` field.

packages/wp-utils/src/snakeToCamelCase.ts
```diff
@@ -37,13 +37,15 @@ export function snakeToCamelCase(string = '') {
...
+ * @return Transformed object.
  */
-export function snakeToCamelCaseObjectKeys(obj, ignore = []) {
-  const isObject = (val) =>
-    val && 'object' === typeof val && !Array.isArray(val);
+export function snakeToCamelCaseObjectKeys(
+  obj: any,
+  ignore: Array<string> = []
+) {
+  const isObject = (obj: unknown) => obj && 'object' === typeof obj && !Array.isArray(obj);
```
swissspidy: `obj` is already used as the function parameter name for `snakeToCamelCaseObjectKeys`, so that name is already taken.

I'd just move this `isObject` helper to outside this function to resolve this lint error.

packages/wp-utils/src/snakeToCamelCase.ts
```diff
@@ -17,10 +17,10 @@
  * Takes a string that is in snake case and returns the same
  * term in camel case.
  *
- * @param {string} string The key in snake case
- * @return {string} The key in camel case
+ * @param string The key in snake case
+ * @return The key in camel case
  */
-export function snakeToCamelCase(string = '') {
+export function snakeToCamelCase(string = '') : string{
```
swissspidy: `group1` and `group2` in the regex callback below should be typed.

```diff
-      (_match, group1, group2) =>
+      (_match, group1: string, group2: string) =>
```

swissspidy:
```suggestion
export function snakeToCamelCase(string = '') : string {
```

packages/wp-utils/src/snakeToCamelCase.ts
```diff
@@ -37,13 +37,15 @@ export function snakeToCamelCase(string = '') {
...
- * @return {any} Transformed object.
+ * @param obj Object to be transformed.
+ * @param ignore Array of keys to be ignored for camelcase.
+ * @return Transformed object.
  */
-export function snakeToCamelCaseObjectKeys(obj, ignore = []) {
-  const isObject = (val) =>
-    val && 'object' === typeof val && !Array.isArray(val);
+export function snakeToCamelCaseObjectKeys(
+  obj: any,
```
swissspidy: We should not explicitly use `any`, it's a lint error.

swissspidy (commented):
This should resolve the remaining type errors in `snakeToCamelCase.ts`:

```ts
/**
 * Takes a string that is in snake case and returns the same
 * term in camel case.
 *
 * @param string The key in snake case
 * @return The key in camel case
 */
export function snakeToCamelCase(string = ''): string {
  if (!string.includes('_') && !string.includes('-')) {
    return string;
  }

  return string
    .toLowerCase()
    .replace(
      /([a-z])([_|-][a-z])/g,
      (_match, group1: string, group2: string) =>
        group1 + group2.toUpperCase().replace('_', '').replace('-', '')
    );
}

type ObjectOrPrimitive = SnakeOrCamelCaseObject | unknown;

interface SnakeOrCamelCaseObject {
  [key: string]: ObjectOrPrimitive;
}

const isObject = (obj: unknown) =>
  obj && 'object' === typeof obj && !Array.isArray(obj);

/**
 * Transform a given object keys from snake case to camel case recursively.
 *
 * @param obj Object to be transformed.
 * @param ignore Array of keys to be ignored for camelcase.
 * @return Transformed object.
 */
export function snakeToCamelCaseObjectKeys(
  obj: ObjectOrPrimitive,
  ignore: string[] = []
): ObjectOrPrimitive {
  if (!isObject(obj)) {
    return obj;
  }

  return Object.fromEntries(
    Object.entries(obj as SnakeOrCamelCaseObject).map(
      ([key, value]: [string, ObjectOrPrimitive]) => {
        return [
          snakeToCamelCase(key),
          isObject(value) && !ignore.includes(key)
            ? snakeToCamelCaseObjectKeys(value)
            : value,
        ] as [string, SnakeOrCamelCaseObject];
      }
    )
  );
}
```

miina:
There are some non-TypeScript lint issues to fix.

packages/wp-utils/package.json
```diff
@@ -25,6 +25,8 @@
     "npm": ">= 7.3"
   },
   "type": "module",
-  "main": "./src/index.js",
+  "main": "./src/index.ts",
+  "types": "dist-types/index.d.ts",
+  "source": "src/index.ts",
```
swissspidy, line 30: Nit: this shouldn't be needed for this package

swissspidy (approved):

Issue #12176, opened 2022-08-29
TypeScript: Convert `templates` package


swissspidy:

barklund:
@swissspidy, this is (kind of) blocked by the `migration` package, so should wait until after that (and that package includes Rollup, so it's a bit trickier to convert, I think?).

swissspidy:
That's a great question.

Since it only uses `DATA_VERSION, migrate` from the `migration` package, it should be possible to only migrate these two things and not the whole `migration` package.

See for example the `design-system` package which isn't fully migrated yet but still can be referenced.

So it should not be a complete blocker I think. Let's give it a try.

And if that doesn't work, yeah I guess we'd have to migrate the `migration` package first.

PR #12274, opened 2022-09-12
TypeScript: Convert `text-sets` package

### LINKED ISSUE #12175: TypeScript: Convert `text-sets` package

spacedmonkey:
## Testing Instructions

1.

## Checklist

---

Fixes #12175


spacedmonkey:
I think this PR is blocked by `@googleforcreators/migration`. I also have some issues with I could not resolve in this PR around typing.

swissspidy:
It is more or less, see https://github.com/GoogleForCreators/web-stories-wp/issues/12176#issuecomment-1237469289

spacedmonkey:
> It is more or less, see [#12176 (comment)](https://github.com/GoogleForCreators/web-stories-wp/issues/12176#issuecomment-1237469289)

So what does that mean for this PR?

miina:
> So what does that mean for this PR?

I´m currently working on the migrations package so we could resolve the other issues in this PR, leave the migration package dependency for the last and finish this PR once it gets unblocked.

swissspidy:
As per https://github.com/GoogleForCreators/web-stories-wp/issues/12176#issuecomment-1238416530 you could even start working on that without a fully converted `migration` package. The `design-system` package is a great prior art for reference.

packages/text-sets/src/types.ts
```diff
@@ -0,0 +1,64 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * External dependencies
+ */
+import type { Page, Story, Element } from '@googleforcreators/types';
+
+export interface TemplateData extends Omit<Story, 'pages'> {
```
swissspidy: TemplateData -> TextSetData

spacedmonkey: Fixed.

packages/text-sets/src/loadTextSets.ts
```diff
@@ -84,10 +88,10 @@ export default async function loadTextSets() {
   ];
 
   const results = await Promise.all(
-    textSets.map(async (name) => {
+    textSets.map(async (name: string): Promise<Array<string | TextSet[]>> => {
```
miina: Curious, doesn't TypeScript infer this type without having to define it manually?

miina (approved):
LGTM

packages/text-sets/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../tsconfig.shared.json",
+  "compilerOptions": {
+    "rootDir": "src",
+    "declarationDir": "dist-types"
+  },
+  "include": ["src/**/*"]
```
barklund, line 8: Doesn't this config need `references` to both the react and types packages?

```suggestion
  "references": [{ "path": "../react" }, { "path": "../types" }],
    "include": ["src/**/*"]
```

I thought it was required to make the types resolve correctly between packages? I do see that the tsc bundler runs just fine without it, so I'm wondering why we have to do it elsewhere actually? @swissspidy, any idea?

packages/text-sets/src/loadTextSets.ts
```diff
@@ -31,21 +36,20 @@ function updateMinMax(minMax, element) {
...
       minX: Infinity,
       maxX: 0,
       minY: Infinity,
       maxY: 0,
     };
 
-    const textElements = page.elements.filter((element) => {
+    const textElements = page.elements.filter((element: Element) => {
       return !element.isBackground && Boolean(updateMinMax(minMax, element));
     });
```
barklund: I didn't really understand this sequence at all at first glance, but it's also doing a lot. We filter for text elements by updating `minMax`, which always returns something _truthy_ anyway? I reckon the goal is to filter out the background element, and also update `minMax` based on all the elements, but those two things have nothing to do with each other.

Let's do a type guard for the former and a simple `forEach` for the latter:

```ts
function isTextElement(e: Element): e is TextElement {
   return 'content' in e; // seems like a decent test?
}
...

const textElements = page.elements.filter(isTextElement);
textElements.forEach((element) => updateMinMax(minMax, element));
```

packages/text-sets/src/loadTextSets.ts
```diff
@@ -31,21 +36,20 @@ function updateMinMax(minMax, element) {
   return minMax;
 }
 
-async function loadTextSet(name) {
-  const data = await import(
+async function loadTextSet(name: string): Promise<TextSet[]> {
```
barklund: This return value will be inferred.

packages/text-sets/src/loadTextSets.ts
```diff
@@ -71,7 +75,7 @@ async function loadTextSet(name) {
   }, []);
 }
 
-export default async function loadTextSets() {
+export default async function loadTextSets(): Promise<TextSets> {
```
barklund: This return value will also be inferred.

barklund (approved):
Overall looks great, just a few nits, mostly on overtyping where the type would be inferred by the compiler anyway.

packages/text-sets/tsconfig.json
swissspidy: I'd have to look into it, but see https://www.typescriptlang.org/docs/handbook/project-references.html#what-is-a-project-reference for what this is supposed to do.

packages/text-sets/src/loadTextSets.ts
spacedmonkey: Should I remove it this then? @barklund

packages/text-sets/src/loadTextSets.ts
spacedmonkey: Should I remove?

packages/text-sets/src/loadTextSets.ts
spacedmonkey: It doesn't. Typescript seems to not like two types. So you have to define it as both.

miina: Ok, it seemed to infer it correctly when I tried locally but perhaps it was just a local thing.

packages/text-sets/src/loadTextSets.ts
spacedmonkey: Yeah, you are right, this code is really confusing. 

I have refactor to be cleaner.

packages/text-sets/src/loadTextSets.ts
barklund: Yes please.

packages/text-sets/src/loadTextSets.ts
barklund: Yes please.

packages/text-sets/src/loadTextSets.ts
spacedmonkey: Fixed.

packages/text-sets/src/loadTextSets.ts
spacedmonkey: Fixed.

packages/text-sets/tsconfig.json
spacedmonkey: I added in the references. It doesnt seem to hurt anything. If we should remove it then we should remove it other places.

packages/text-sets/src/loadTextSets.ts
```diff
@@ -89,5 +92,5 @@ export default async function loadTextSets() {
     })
   );
 
-  return Object.fromEntries(results);
+  return Object.fromEntries(results) as TextSets;
```
barklund: `as` is a rather crude tool that should only be applied if no better tool exists, as it will actually ignore and suppress some potential errors. However, moving it to a typed variable or just setting the return type won't do the trick, because `fromEntries` doesn't play nice (typed as returning `any`).

This required updating the type system a bit as:

```ts
export enum TextSetType {
  Cover = 'cover',
  Step = 'step',
  SectionHeader = 'section_header',
  Editorial = 'editorial',
  Contact = 'contact',
  Table = 'table',
  List = 'list',
  Quote = 'quote',
}

export type TextSets = Partial<Record<TextSetType, TextSet[]>>;
```

And then `loadTextSets` becomes:

```ts
export default async function loadTextSets(): Promise<TextSets> {
  const results: [TextSetType, TextSet[]][] = await Promise.all(
    Object.values(TextSetType).map(async (name) => [
      name,
      await loadTextSet(name),
    ])
  );
  return Object.fromEntries(results);
}
```

The `Partial<>` is a bit annoying but I don't see a clean way around it? I feel this is cleaner, but it's probably just a nit really.

spacedmonkey: If I am honest, I don't find your code easier to read or understand. I have pushed up my own version of the code, that is a little easier to follow. See https://github.com/GoogleForCreators/web-stories-wp/pull/12274/commits/ed4be89718d16dba892908a2f1f6afb18539bc63

swissspidy:
Let‘s put this behind a feature flag first before shipping this. Don‘t wanna break sites unexpectedly.

packages/text-sets/src/loadTextSets.ts
spacedmonkey: @barklund Can you take a look ?

swissspidy (approved):

Issue #12261, opened 2022-09-08
Data Structure: reduce redundancy when storing fonts


swissspidy:
## Feature Description

This is a follow-up to #12131.

Right now when adding 10 text elements all with the same font, there will be 10 font objects being stored, all with the same information. That's very redundant.

Instead of this, consider storing all existing fonts at the story level instead of on each element to reduce redundancy.

Add data migration.

PR #12301, opened 2022-09-13
Code: Added new types package for TS

### LINKED ISSUE #12090: TypeScript: Create new story package with typings

barklund:
## Context

This adds a new TS type-only package.

## Testing Instructions

## Checklist

---

Fixes #12090


packages/patterns/src/getOpaquePattern.ts
```diff
@@ -15,9 +15,13 @@
...
 /**
- * Internal dependencies
+ * External dependencies
  */
-import { Gradient, Pattern, PatternType } from './types';
+import {
+  Gradient,
+  Pattern,
+  PatternType,
+} from '@googleforcreators/types';
```
barklund: @swissspidy, I am getting some very annoying conflicts between various formatters now. I have tried updating everything, but the issues persist.

Basically, when I auto-format this file in VScode using prettier (Opt+Shift+F), these three imports are inlined correct as:

```ts
import { Gradient, Pattern, PatternType } from '@googleforcreators/types';
```

However, when I save the file, some other formatter runs and re-formats the file as:

```ts
import {
  Gradient,
  Pattern,
  PatternType,
} from '@googleforcreators/types';
```

And this is just one of the minor problems caused by these unwanted reformats. Sometimes whole lines of imports (almost always types) are moved around in the file, causing imports to occur before `/** External dependencies */` causing other lint errors.

Do you have any idea what I could have configured wrong? Or do we have some new formatting rules for TS that isn't correctly synced with prettier and/or ESLint?

barklund: Just checked and VSCode's built-in "format on save" setting was disabled. Enabling that causes both prettier format and the unknown formatter to run, prettier first and then the other one second.

It seems like some rogue watcher doing things wrong, but I definitely don't have anything running in my terminal. Closing all terminals and relaunching VSCode still has the problem persisting, so not really sure what could be going on here?

barklund: Hah, I figured it out. I had installed a small package to do import sorting on request, but apparently it also did so on save, but it was an idiot at it, so...!

barklund:
> What‘s the advantage of also moving things like the pattern types to this new package?

Mostly to avoid circular references. The types package should be at the bottom of the dependency tree IMO and thus not depend on anything else, but it needs the patterns for various color properties, so this made sense to me. This can of course be reversed.

swissspidy:
I see. There’s definitely a lot to untangle with all the element types and such. Patterns don‘t have any other dependencies, so there should be no circular references in that instance. But let‘s look at it once this progresses further 👍

packages/types/src/element/elementBox.ts
```diff
@@ -14,22 +14,6 @@
...
-  isBackground: boolean;
-  border: {
-    top: number;
-    right: number;
-    bottom: number;
-    left: number;
-  };
-}
-
 export interface ElementBox {
```
miina, line 17: Should this also be removed from here now?

packages/types/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "main": "dist/index.js",
+  "module": "dist-module/index.js",
+  "types": "dist-types/index.d.ts",
+  "source": "src/index.ts",
+  "publishConfig": {
+    "access": "public"
+  },
+  "sideEffects": false,
+  "dependencies": {},
+  "devDependencies": {}
```
miina: Are these needed here if empty?

packages/types/README.md
```diff
@@ -0,0 +1,3 @@
+# Types
+
+Utility package with core TypeScript types used across many other packages.
```
miina, line 3: Should this state that it's for Web Stories packages?

packages/patterns/src/hasGradient.ts
```diff
@@ -15,12 +15,12 @@
...
 /**
- * Internal dependencies
+ * External dependencies
  */
-import { Pattern, PatternType } from './types';
+import { Pattern, PatternType } from '@googleforcreators/types';
 
 function hasGradient(pattern: Pattern) {
-  return Boolean(pattern?.type) && PatternType.SOLID !== pattern.type;
+  return Boolean(pattern?.type) && PatternType.Solid !== pattern.type;
```
miina, line 23: Curious why change from uppercase to camelcase? I can see it's still uppercase for `ResourceType` for example, should we do just one or the other for consistency?

packages/patterns/src/convertToCSS.ts
```diff
@@ -14,13 +14,18 @@
...
  */
 
-function convertToCSS(style: Record<string, string>) {
+/**
+ * External dependencies
+ */
+import type { CSSProperties } from 'react';
+
+function convertToCSS(style: CSSProperties) {
   return Object.entries(style).reduce((str, [key, val]) => {
```
miina, line 23: Could also add the type `string` here for `val` instead, maybe it's better than casting?

packages/media/src/createResource.ts
```diff
@@ -57,6 +62,56 @@ function createResource({
   trimData,
   needsProxy = false,
 }: ResourceInput): Resource | VideoResource | GifResource {
+  type = type || getTypeFromMime(mimeType);
+  if (type === ResourceType.VIDEO) {
```
miina: Hm, these two are almost identical, could we create a general resource object instead and just add the unique props inside `if` clause? E.g. `resource = { ...sharedProps }; if ( Video ) return { ...sharedProps, ... videoProps };` etc.? Would be more visible as well then which props are shared and which unique. Thoughts?

packages/patterns/src/hasGradient.ts
barklund: [TypeScript enums are generally written in `PascalCase`](https://basarat.gitbook.io/typescript/styleguide#enum), so let's try to stick to that. I'll update `ResourceType` as well.

packages/types/README.md
barklund: I think that's implicit, other packages also imply this.

packages/types/package.json
barklund: I think so?

packages/types/src/element/elementBox.ts
barklund: The file has been "moved" so it's no longer where it was.

packages/media/src/createResource.ts
barklund: Yeah, it could be cleaned up a bit. I'll look into it.

packages/patterns/src/convertToCSS.ts
barklund: I wasn't happy about this, but didn't see a better way. I'd have to type the whole array here, so it would become:

```ts
  return Object.entries(style).reduce((str, [key, val]: [string, string]) => {
```

Which doesn't look that good IMO.

packages/types/package.json
swissspidy: Should be safe to omit I think

swissspidy:
> I see. There’s definitely a lot to untangle with all the element types and such. Patterns don‘t have any other dependencies, so there should be no circular references in that instance. But let‘s look at it once this progresses further 👍

To iterate on my previous comment, the main things standing out:

1. Pattern types -> more suitable for `patterns` package
2. Specific element types (`ProductElement`, `GifElement`, etc.) -> more suitable for `element-library`
3. Resource types -> `media` package

There shouldn't be any circular dependencies that way.

I can see the appeal of a catch-all types package at the bottom of the dependency tree though. So not sure what works best...

packages/types/src/element/element.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export interface BorderRadius {
+  topLeft: number;
+  topRight: number;
+  bottomLeft: number;
+  bottomRight: number;
+  locked: boolean;
+}
+
+export interface Element extends ElementBox {
```
miina, line 55: Maybe it would be good to add a union type of all the elements as well for using in other packages? We're very often using `element` as a prop without knowing which element type it is exactly. Maybe `element` could be a union type and `baseElement` (or sth similar) what it currently is?

packages/types/src/story/page.ts
```diff
@@ -0,0 +1,38 @@
...
+export interface Group {
+  name: string;
+  isLocked: boolean;
+  isCollapsed?: boolean;
+}
+
+export type Groups = Record<string, Group>;
+
+export interface Page {
+  elements: Element[];
```
miina, line 32: Should this specify that we have an array that can consist of text / shape / image etc. element types instead of just the general `Element[]` array? 🤔

packages/types/src/element/elementBox.ts
barklund: > To iterate on my previous comment, the main things standing out:
> 
> 1. Pattern types -> more suitable for `patterns` package
> 2. Specific element types (`ProductElement`, `GifElement`, etc.) -> more suitable for `element-library`
> 3. Resource types -> `media` package
> 
> There shouldn't be any circular dependencies that way.

I think it's a lot of work to do it that way, but doable if we're diligent.

Some of the utility functions inside the media package need access to element types (e.g. `generateVideoStrip(e:MediaElement, ...)`), but because the element library package would depend on the media package (the `VideoElement` needs to reference the `VideoResource`, etc), the media package cannot depend in reverse. So the media element would have to be defined in the media package.

That means that the abstract catch-all element type would exist in the types package, the abstract media element type would exist in the media package, and the concrete element types (some extending element, others extending media element) would exist in the element library package.

Sure, it would work, but it would also be a bit confusing in my opinion. Another option would be to move the `generateVideoStrip` function somewhere else, so the media element type can be defined in the element library package. But to where?

I also notice that the element library package depends on a ton of other packages, so none of those can ever depend on any specific element type, which would seem to hamper us a bit. E.g. the animation package, which have methods referring to media elements (e.g. in `mediaPositions.js`). Similarly inside the rich-text package, where `usePastedTextContent` needs to know about the text element interface.

I think the problem with this "types belong in their packages" approach is that we often combine both core functionality and broader utility functions inside a package - those broader utility functions often need types from other packages, but that creates a mess in dependencies.

While having all the types in a single package perhaps isn't best-practice, it would make things a lot simpler.

Would it make a difference if we made the more generic interfaces, of which actual elements would implement one or more? So we have `EditableElement`, `MediaElement`, `TrimmableElement`, `BackgroundableElement`, `ColorableElement`,  etc? Then functions through would accept one of those interfaces, and the actual elements could be passed because they implemented said interface?

Such a solution feels more "computer sciencey", but also more complex, as our actual elements are super concrete and such abstractions would be an almost useless exercise except to make the type system consistent.

WDYT?

Ping @miina and @merapi too.

barklund:
> > I see. There’s definitely a lot to untangle with all the element types and such. Patterns don‘t have any other dependencies, so there should be no circular references in that instance. But let‘s look at it once this progresses further 👍
> 
> To iterate on my previous comment, the main things standing out:
> 
> 1. Pattern types -> more suitable for `patterns` package
> 2. Specific element types (`ProductElement`, `GifElement`, etc.) -> more suitable for `element-library`
> 3. Resource types -> `media` package
> 
> There shouldn't be any circular dependencies that way.
> 
> I can see the appeal of a catch-all types package at the bottom of the dependency tree though. So not sure what works best...

I think it's a lot of work to do _"types belong in the relevant package"_ organisation, but doable if we're diligent.

Some of the utility functions inside the media package need access to element types (e.g. generateVideoStrip(e:MediaElement, ...)), but because the element library package would depend on the media package (the VideoElement needs to reference the VideoResource, etc), the media package cannot depend in reverse. So the media element would have to be defined in the media package.

That means that the abstract catch-all element type would exist in the types package, the abstract media element type would exist in the media package, and the concrete element types (some extending element, others extending media element) would exist in the element library package.

Sure, it would work, but it would also be a bit confusing in my opinion. Another option would be to move the generateVideoStrip function somewhere else, so the media element type can be defined in the element library package. But to where?

I also notice that the element library package depends on a ton of other packages, so none of those can ever depend on any specific element type, which would seem to hamper us a bit. E.g. the animation package, which have methods referring to media elements (e.g. in mediaPositions.js). Similarly inside the rich-text package, where usePastedTextContent needs to know about the text element interface.

I think the problem with this "types belong in their packages" approach is that we often combine both core functionality and broader utility functions inside a package - those broader utility functions often need types from other packages, but that creates a mess in dependencies.

While having all the types in a single package perhaps isn't best-practice, it would make things a lot simpler.

Would it make a difference if we made the more generic interfaces, of which actual elements would implement one or more? So we have EditableElement, MediaElement, TrimmableElement, BackgroundableElement, ColorableElement, etc? Then functions through would accept one of those interfaces, and the actual elements could be passed because they implemented said interface?

Such a solution feels more "computer sciencey", but also more complex, as our actual elements are super concrete and such abstractions would be an almost useless exercise except to make the type system consistent.

WDYT?

Ping @miina and @merapi too.

packages/types/src/story/page.ts
swissspidy: `Element[]` should be fine, since ideally in the future someone would be able to build their own custom story editor with custom element types

swissspidy:
Those are good points, thanks for bringing them up.

This shows that we still have a ways to go to improve our dependencies and avoid circular references.

> Some of the utility functions inside the media package need access to element types
> Another option would be to move the `generateVideoStrip` function somewhere else, so the media element type can be defined in the element library package. But to where?

The media package should not need access to element types.

`generateVideoStrip` is only used by the editor in `useRailBackground()`, so we should just move this to `packages/story-editor/src/components/videoTrim/generateVideoStrip.js`

Are there others utils in `media` besides this one that should be moved? Seems like the only one.

> I also notice that the element library package depends on a ton of other packages, so none of those can ever depend on any specific element type, which would seem to hamper us a bit. E.g. the animation package.

I think we should just get rid of the `animation` dependency in `element-library`. It's only there because of `BG_MIN_SCALE` and `BG_MAX_SCALE`. We can either just copy these constants, or pass them from the `story-editor` package as props to `MediaEdit`.

> E.g. the animation package, which have methods referring to media elements (e.g. in `mediaPositions.js`).

Since `animations` the package already implicitly references elements, we could perhaps add an explicit dependency on the `elements` package instead.

Maybe `getMediaBoundOffsets` is a better fit to be called via `getDefinitionForType()` so it's more element type agnostic.

> Similarly inside the rich-text package, where `usePastedTextContent` needs to know about the text element interface.

Another implicit dependency that we should strive to avoid. In fact, it's the only such dependency on the text element type in the `rich-text` package.

This could be solved by moving the file to `story-editor` or by passing a different `insertElement` so it does not have to know about the interface. e.g. in `useCanvasCopyPaste` we'd do:

```js
const pasteTextContent = usePasteTextContent((content) => {
  insertElement('text', { ...DEFAULT_PRESET, content });
});
```

> I think the problem with this "types belong in their packages" approach is that we often combine both core functionality and broader utility functions inside a package - those broader utility functions often need types from other packages, but that creates a mess in dependencies.

That sounds like a sign to split up packages.

> While having all the types in a single package perhaps isn't best-practice, it would make things a lot simpler.

That is definitely true. And I think it would be OK to do that for now, but we should strive to fix the circular dependency issues so that we no longer have to do this.

miina:
When the packages were being created, I remember it was often not clear what belonged to where, adding types seems to be a great opportunity to improve all that as well.

Agreed that it would be ideal to have the types with the correspondent packages, it would widen the scope of this issue quite a lot though.

Sounds like the simplest way would be using the shared types package as Morten currently has in this PR and then creating follow-up issues for moving the types where they belong together with improving the dependencies and the organization of the packages.

barklund:
Alright, let's get this merged to unblock a number of other issues (which would require a ✅ ), and then I'll create some followups to organize some of the other packages better.

packages/types/src/element/element.ts
barklund: I see why you want that, but it's kind of not the right approach here. Rather, if you're looking through the array of elements for say `TextElement`'s, to manipulate all text elements, you would do so with a [_user-defined type predicate_](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) (we could easily create those along with each element type, but they're also trivial to define when needed):

```ts
function isTextElement(e: Element): e is TextElement {
  return e.type === 'text';
}
const textElements = page.elements.filter(isTextElement);
```

The latter variable here now has `TextElement[]` type. It can also be defined inline:

```
page.elements.filter((e: Element): e is TextElement => e.type === 'text');
```

That `x is Foo` syntax looks weird, but it works, so 🤷.

packages/patterns/src/convertToCSS.ts
barklund: I've changed it to:

```
return `${str}${casedKey}:${String(val)};`;
```

That looks at least a bit better. And would correctly convert numbers (which are the most likely non-strings here).

packages/types/src/element/element.ts
miina: What if it's a function that actually accepts all types, for example:
```
function getLayerName(element) {
  if (element.layerName) {
    return element.layerName;
  }

  if (element.isBackground) {
    return __('Background', 'web-stories');
  }

  return getDefinitionForType(element.type).getLayerText(element);
}
```

If we would assign just `element: Element`, it would complain about `isBackground` not existing (it exists on shapes and media only but not on `Element`).

Thoughts?

packages/types/src/element/textElement.ts
```diff
@@ -0,0 +1,65 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * Internal dependencies
+ */
+import type { Element } from './element';
+import type { Pattern } from './pattern';
+
+export type FontStyle = 'normal' | 'italic';
```
miina: Note: Seeing `regular` in use as well in the codebase.

packages/types/src/element/textElement.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export type FontVariant = [FontVariantStyle, FontWeight];
+
+export interface Font {
+  family: string;
+  service?: string;
+  weights?: FontWeight[];
+  styles?: FontStyle[];
+  variants?: FontVariant[];
+  fallbacks?: string[];
```
miina, line 39: `metrics` seems to be missing from here.

packages/media/src/createResource.ts
barklund: I've overhauled it. PTAL.

packages/patterns/src/hasGradient.ts
barklund: Renamed the ResourceType members too.

packages/types/package.json
barklund: Removed.

packages/types/src/element/textElement.ts
barklund: Oh yeah, we're using both `'regular'` (in text panel and templates) and `'normal'` (in the style panel and design menu). Hmm, what to do? Should we make these aligned or?

packages/types/src/element/textElement.ts
barklund: @merapi, would you happen to have a type handy for metrics, pretty please? 😂

barklund: (it's actually quite the mess that we put the metrics in the story data at all, they definitely don't belong there, and even if they did, they should be summarized somewhere on the top level, not duplicated per text element)

miina: True, it has been discussed previously as well, perhaps @swissspidy remembers why all the font information was saved per element even though it's the same for the same font.

barklund:
I feel it makes sense to "postpone" the moving of types to their relevant packages to when we convert the element library package. Yes, there'll be a lot of references to update, but I think it makes sense to do only then.

In the meantime, we could make another ticket to reduce some of the references between packages? Specifically:

* remove the link from animation package to media (by going through `getDefinitionForType` instead, which would instead add a link to the elements package)
* remove the link from media to element (by moving `generateVideoStrip` to the story-editor package)
* remove the link from element-library to animation package (by passing the relevant properties in the story editor)

packages/types/src/element/textElement.ts
miina: Sounds like another follow-up issue if these are the same things :) Until then I suppose we should add `regular` to the union type.

barklund: Fixed in 0a8dbd2af8. New issue in https://github.com/GoogleForCreators/web-stories-wp/issues/12333.

miina (approved):
LGTM as the initial version 🔨 

(pending the missing `metrics`)

packages/types/src/element/textElement.ts
swissspidy: > (it's actually quite the mess that we put the metrics in the story data at all, they definitely don't belong there, and even if they did, they should be summarized somewhere on the top level, not duplicated per text element)

Yes, absolutely. Which is why I created #12261 to move them to the top level. It would definitely help make things more clean.

I think originally we stored it per-element because it had the advantage of having all the information right away when dealing with text elements, without having to do round trips to the story provider to get font information and pass that along everywhere. Just at the cost of a little redundancy. I don't think we had metrics yet back then.
But yeah especially with the introduction of metrics (and having to keep metrics updated everywhere) this is now a bit too much and the top level storage would be cleaner.

swissspidy: And if you need a type:

```ts
interface FontMetrics {
  upm: number;
  asc: number;
  des: number;
  tAsc: number;
  tDes: number;
  tLGap: number;
  wAsc: number;
  wDes: number;
  xH: number;
  capH: number;
  yMin: number;
  yMax: number;
  hAsc: number;
  hDes: number;
  lGap: number;
}
```

PR #12318, opened 2022-09-14
Editor and output: Custom page advancement

### LINKED ISSUE #12210: Page-Level Page Advancement Setting
## Feature Description

Currently we only support changing the page advancement (manual/automatic, duration) for the whole story for consistency reasons.

We now wanna try to allow this on a per-page basis as well.

barklund:
## Context

This adds a new custom per-page page advancement panel.

## User-facing changes

| Advancement setting | Without override | With override |
|-|-|-|
| Manual | Screen Shot 2022-09-14 at 19 00 12 | Screen Shot 2022-09-14 at 18 58 37 |
| Auto | Screen Shot 2022-09-14 at 18 58 23 | Screen Shot 2022-09-14 at 18 58 33 |

## Testing Instructions

1. Enable the experiment
2. Set any document-wide page advancement setting in the document panel
3. Then for any individual page, override the global page advancement setting locally (or not)
4. Observe that the preview story will correctly used either the local page setting or the global setting if not overridden.

## Checklist

---

Fixes #12210


packages/story-editor/src/components/panels/shared/generalPageAdvancement.js
```diff
@@ -0,0 +1,202 @@
...
+  autoAdvance = DEFAULT_AUTO_ADVANCE,
+  onUpdate,
+  hasOverride = false,
+  isOverridden = false,
+  children,
+  panelName,
+  ...rest
+}) {
+  const [duration, setDuration] = useState(pageDuration);
+  const [override, setOverride] = useState(isOverridden);
```
miina: It's a bit hard to read the `hasOverride`, `isOverridden`, and `override` all together.

Maybe something like `overrideDefaultValue`, `setOverrideDefaultValue` and `isOverridden` => `hasCustomValue`; `hasOverride` => `allowsOverriding`?

Not sure if that's very clear but currently it takes a bit time to understand what's going on :D

miina (approved):

packages/story-editor/src/components/panels/shared/generalPageAdvancement.js
barklund: Yeah, I know... I got confused as well. I'll try rewriting to some longer and more reasonable names.

barklund: Done in b19b91e.

kkalarickal:
Verified this feature via web-story: [Custom page advancement#12318](https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/post.php?post=19697&action=edit#page=%2522718992e3-08ff-43de-9f27-cf093d9c3d7e%2522)

Status: 
- Functionality : OK
- Usability : see UX Suggestions for improvement

Scenario: Settings and Defaults

1. With _Experimental > Enable detailed page advancement settings on a per-page basis_ turned OFF, the global page adv works as expected
2. With the experiment turned ON, each page has a Style > Page Advancement panel that controls page adv for individual page
3. When Global Page Adv = AUTO, then by default Individual Page Adv = AUTO. 
4. Value of Duration is updated from Global to individual when there is no override
5. Similarly, when Global Page Adv = MANUAL, then by default the Individual Page Adv = MANUAL

**UX Suggestion**
The title of the panel - Page Advancement -  is identical for both Global and Individual. This is in contrast to the title for the Background Audio panel ... which is called "Page Background Audio" (individual) vs "Background Audio" (global). It might be useful to have a different title for the subpanel so that use does not mistakenly think they are changing the global when they are in fact changing the individual (or vice versa).

kkalarickal:
Scenario: Overriding and Duration value

1. When Individual Page Adv has OVERRIDE turned ON, it is possible to change the option to Auto or Manual for the page.
2. If the Global Page Adv Auto Duration is changed, then the Individual Page Adv Auto Duration remains unaffected.
3. Similarly, if the Global setting is changed to Manual, then the Individual setting remains the overridden value
4. Resetting Duration when going from Manual to Auto
- Set Global Page Adv to Auto and change Duration (from default 7) to another number, 
- Change the Global Page Adv from Auto to Manual and back to Auto
- Notice that the Duration is reset to default 7 sec
- If you do the same sequence for Individual Page Adv settings, then the Duration is NOT reset.
5. Toggling Override from ON to OFF and back to ON, resets the Duration value to the Global Page Adv Auto Duration as expected

kkalarickal:
Scenario: Using the page adv settings in Preview & Publish

1. Confirmed that setting a mixed configuration of Override > Auto & Default > Auto, each page maintains the appropriate Auto: Duration
2. Confirmed that setting a mixed configuration of Override > Manual & Default > Manual, each page needs to be advanced manually
3. Confirmed that setting a mixed config of Override > Auto, Default > Manual ... as well as Override > Manual, Default > Auto, each page advances as per its specification.

kkalarickal:
Scenario: Corner Cases and Negative Testing

1. A Global Page Advancement option is shown in the Publish overlay that occurs when the story is moved from Draft -> Publish. Confirmed that this Global Page Adv setting matches the behavior of the Global Page Adv setting in the editor.

2. Setting the Auto > Duration to be less than the duration of an embedded video results in the page remaining open beyond the Duration so that the video is played fully. But if video is on a Loop, the Duration value is used for keeping page open.

Example - Page Adv Duration is 6 seconds, but embedded video is of 11 sec. Then the page will remain open for 11 sec till video auto plays to completion. However, if the Loop option is turned on for the video, then the page remains open for the Duration of 6 seconds. 

@swissspidy -- is it valid to have the page adv auto duration ignored if there is a video playing in the page that is longer duration ?

swissspidy:
Yes this is expected behavior.

miina:
> UX Suggestion
The title of the panel - Page Advancement - is identical for both Global and Individual. This is in contrast to the title for the Background Audio panel ... which is called "Page Background Audio" (individual) vs "Background Audio" (global). It might be useful to have a different title for the subpanel so that use does not mistakenly think they are changing the global when they are in fact changing the individual (or vice versa).

Thanks for thinking about this and for the suggestion! I believe @aaskedall recommended using something like "Default Page Advancement" in the Document settings, we're currently still waiting for the mocks though.

PR #12320, opened 2022-09-15
TypeScript: Migrate `migration` package

### LINKED ISSUE #12239: TypeScript: Convert `migration` package

miina:
## To-do

- [x] Should we carry _all_ used types in each file? What if they're used (changed / migrated) just once?
- [ ] `migrate.ts` file's `migrationFn` needs figuring out
- [x] Depth error with `PageV[X]` and `StoryV[X]` types.

## Testing Instructions

1.

## Checklist

---

Fixes #12239


packages/migration/src/migrations/v0042_removeTrackName.ts
```diff
@@ -14,7 +14,38 @@
  * limitations under the License.
  */
 
-function removeTrackName(storyData) {
+/**
+ * Internal dependencies
+ */
+import type { ElementV43, PageV43, StoryV43 } from './v0043_removeTagNames';
```
miina: @barklund @swissspidy Before I continue with all the files, wanted to ask for feedback for this approach (see this file and the next, `v0043_` file).

The idea is that the last migration returns the current `Story` type, however, the migration before that returns `StoryV43` type, and the one before that `StoryV42` type, etc. Each migration will then extend the returned story type, e.g. the last migration returns `Story` as it currently is but it applies the migration script on an extended `Story` type -- as it was before migration.

Typed these two files as an example for now. Thoughts?

barklund: I like it. It's super clean and type-safe. I've checked around, but don't see any recommendations as to what to do in general, so this is as good as any approach IMO.

But let's see what @swissspidy says too.

swissspidy: Hmm the downside with that is whenever you add a new migration, you have to edit the previous one as well.

Looking at this file here (`v0042_removeTrackName`), intuitively I would think that `removeTrackName` takes a `StoryV41` type (from the previous migration) and returns a new `StoryV42` type (the resulting version after this migration).

The `migrate()` function then is the one returning the current `Story` type.

WDYT?

Note: Not sure about `v0004`, which has two migrations (`squareToShape` and `mediaElementToResource`)

miina: > Hmm the downside with that is whenever you add a new migration, you have to edit the previous one as well.

Yep, unfortunately. Not sure if there's a better option though.

> Looking at this file here (v0042_removeTrackName), intuitively I would think that removeTrackName takes a StoryV41 type (from the previous migration) and returns a new StoryV42 type (the resulting version after this migration).
The migrate() function then is the one returning the current Story type.

Sure, sounds good. The last migration would still also return the current Story type. Although we could still create a new type for clarity / to have to make less changes for the next migration.

> Note: Not sure about v0004, which has two migrations (squareToShape and mediaElementToResource)

Could have StoryV4_0 and StoryV4_1 based on the index in the array of migrations for example.

miina: @swissspidy I tried so that StoryV40 is defined in the `v40` file and then `v41` migration imports it from `v40` to take that type as the input, however, since the changes that happen in V41 are then type in V40 file this way, it gets quite confusing to read this -- the changes happening in each file have nothing to do with the type that's defined in that file. It might be correct by the numbers but it's quite confusing to read.

I currently changed the code so that now the input version is defined in the file that actually uses it, PTAL. E.g. V39 is defined in V40 file and used there as well and relates to the code there. It's much easier to read this way but now importing V39 is from V40 file, which is also a bit confusing. It's still easier to read though, so 🤷 

Let me know if you still think it's better to have it correct in terms of numbers (defining `StoryV39` in `V39` file but the relevant code actually being in V40).

packages/migration/src/migrations/v0043_removeTagNames.ts
```diff
@@ -14,26 +14,45 @@
...
+ * External dependencies
+ */
+import type { Story, Page, TagName } from '@googleforcreators/types';
+
+/**
+ * Internal dependencies
+ */
+import type { StoryElement } from '../types';
+
+export type ElementV42 = StoryElement & {
```
barklund: This would specifically extend TextElement rather than element, and you should use the custom type guard to filter the element list down to only text elements first.

I can see the issues though with this approach, because then you're creating a `TextElementV42` because that element type changed in the version, but what about the other element types, that don't change. It's a lot of work to find all the proper types to extend then.

miina: Hm, so you're saying that it should be something like:
```
const ElementV42 = 
  | GifElement
  | ImageElement
  | ProductElement
  | ShapeElement
  | TextElementV42
  | VideoElement;
```

where only the text element is extended.
And then when some other type of Element is changed in another migration file, we should again extend only that element, e.g. 
```
const ElementV39 = 
  | GifElement
  | ImageElementV39
  | ProductElement
  | ShapeElement
  | TextElementV42
  | VideoElement;
```

where the image element is the new type and all the others should use the closest previously extended type (e.g. like `TextElementV42` which was extended for V42?

Do you have a simpler alternative in mind? 🤔

barklund: On the other hand, the above is saying that all elements could have had a `tagName` property in V42, which is definitely not true. Only text elements could.

miina: Yep, but the following would say that only text elements could have `tagName`, no?

```
interface TextElementV42 extends TextElement {
  tagName?: string;
}

const ElementV42 = 
  | GifElement
  | ImageElement
  | ProductElement
  | ShapeElement
  | TextElementV42
  | VideoElement;
```

Let me know if you have a simpler alternative.

miina: @barklund I [updated the PR again](https://github.com/GoogleForCreators/web-stories-wp/pull/12320/commits/f145e40d322ae2fcfb6403adce1c2eb90ce64193) to show what I meant by the code above. So basically depending on which element type is being modified, we extend only that one and leave the others as they were. It shouldn't be much more complicated than the previous version. Thoughts?

packages/migration/src/migrations/v0042_removeTrackName.ts
swissspidy: > Although we could still create a new type for clarity / to have to make less changes for the next migration.

👍 💯  

-----

> I tried so that StoryV40 is defined in the `v40` file and then `v41` migration imports it from `v40` to take that type as the input, however, since the changes that happen in V41 are then type in V40 file this way, it gets quite confusing to read this

I am definitely confused now. Did you push these changes? Because that's not what I am seeing.

And why are V41 types in V40? V40 should not know anything about future migrations, because they don't exist yet at the time of writing.

> I currently changed the code so that now the input version is defined in the file that actually uses it, PTAL. E.g. V39 is defined in V40 file and used there as well and relates to the code there.

Why is V39 defined in V40 and not in the actual V39 file? That seems like the same setup as yesterday, where one has to modify old migrations whenever you need a new one.

----

What I had in mind when I wrote my comment yesterday:

A V39 migration:

1. Imports V38 type from V38 file
2. Defines a new V39 type
3. Takes V38 data and returns V39 data
4. Exports V39 type so it can be used by the next migration

This way, V39 builds upon the previous migration and has no awareness of any future migrations like V40

miina: > This way, V39 builds upon the previous migration and has no awareness of any future migrations like V40

Yep, I understood what you meant -- I tried it out and it was just quite hard to read + one still has to modify old migrations whenever we need one because we don't know what will change in the future. The last migration's types are the same as the current type but if anything changes in the current type, we'll still need to change the previous migration's types as well.

Anyway -- I can definitely still do as you recommend, will update on Monday.

miina: > This way, V39 builds upon the previous migration and has no awareness of any future migrations like V40

Actually there's a problem with this as well -- we don't know how the original Story type looked like in the first migration and we don't know which properties were added while each next migration version was added. Migration doesn't migrate ALL the changes, just the ones that need migration.

> And why are V41 types in V40? V40 should not know anything about future migrations, because they don't exist yet at the time of writing.

Because we need a base story type to take as the reference type. Since we don't have the very original Story of version 0 as the reference and we don't know which properties were added meanwhile to fully write out every type, we can't take the version 0 Story as the reference and start extending from Version 0. We can instead do the other way around -- the last migration takes the current Story type as the reference every time and start extending from the last migration. When a new migration is added, the previously last one takes the new migration's story type as the reference.

If V39 builds upon the previous migration then we would need the Story type of V0 to start from there.

Let me know if I explained this clearly, perhaps you have an easy way in mind to get the Story type for V0 to take as the basis.

swissspidy: Sounds like I might need to look into the code myself to better understand the drawbacks. Definitely want to avoid unnecessary back and forth here, so let's first make sure we are all on the same page.

From the looks of it, it definitely sounds like we need to work backwards and create a V0 type that is then used by the first migration.

miina: > From the looks of it, it definitely sounds like we need to work backwards and create a V0 type that is then used by the first migration.

We could do it but probably just based on the changes in the migration and not the "real" story type that actually existed in the beginning since when adding new props we don't do migration.

However, when taking the last Story as the base, we would end up with the same result since every version type is considering just the changes reflected in the migration files. 

If that's acceptable then we could definitely do that, just take the current Story type and work back through the migrations to create the initial Story type (which would not be fully accurate but reflect the migrations correctly). Thoughts?

packages/migration/src/types.ts
```diff
@@ -0,0 +1,222 @@
...
+  BorderRadius,
+  MediaElement,
+  VideoTrack,
+  TrimData,
+} from '@googleforcreators/types';
+
+type TextAlign = 'left' | 'center' | 'right' | 'justify';
+type TagName = 'h1' | 'h2' | 'h3' | 'p' | 'auto';
+
+export interface FontMetrics {
```
miina: @swissspidy I've updated this file to work back to the Story V0 type (currently unorganized and still depending on the global current types at this point).

A few things to consider
- There are some props that we know did not exist in the V0, such as the font metrics or all the trimming related data etc. However, it's not touched by the migration script at all. Should we remove these and try to track down when they were added, or is it ok to just keep them from V0 (since they're not touched by the migration script)? Thoughts?
- Same goes for the GIF element for example. It was added at some point -- is it ok to assume it existed since V0 even though it didn't?
- Since the current types are changing, I guess we should remove all dependency on the `types` packages and port all the types over to here for V0. Currently there are a few imports still in this file.

swissspidy: I think it would be too much effort to track down when these were added, so should be ok to just leave it. Maybe with an explanatory comment to avoid confusion in the future.

> * Since the current types are changing, I guess we should remove all dependency on the `types` packages and port all the types over to here for V0. Currently there are a few imports still in this file.

Makes sense. `types` is ever-changing so we can't rely on it here via import.

packages/migration/src/migrations/v0003_fullbleedToFill.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export interface GifElementV3 extends Omit<GifElementV0, 'isFullbleed'> {
+  isFill?: boolean;
+}
+
+export interface ImageElementV3 extends Omit<ImageElementV0, 'isFullbleed'> {
+  isFill?: boolean;
+}
+
+export type UnionElementV3 =
```
miina, line 48: @barklund @swissspidy This PR now uses the initial StoryV0 as the base type to extend for each following migration and has a few first files migrated, PTAL.

Let me know what you think of this system with the UnionElement type here. Currently, it's using the last extended version of the specific element, e.g. the ShapeElement has not been changed meanwhile so it still uses `V0`, the VideoElement type has already been changed so it uses the `VideoElementV3`.

An alternative would be creating a new version for each type in each migration file, even if it doesn't change. E.g. each migration file exports a new type for each element type even if the element type doesn't get extended. It would perhaps be more clear to read and easier for the next version even though it would mean adding a bunch of types just for the purposes of having a type for each version.

Also, if we'd do that, should we do it for the "main" types that are most likely used in most of the migration files (different element types, page, etc.), or should we do it for all types (e.g. resource, font metrics, etc. which likely change less or not change at all).

Thoughts?

packages/migration/src/migrations/v0002_dataPixelTo1080.ts
```diff
@@ -14,6 +14,12 @@
  * limitations under the License.
  */
 
+/**
+ * Internal dependencies
+ */
+import type { PageV0, UnionElementV0 } from '../types';
+import type { StoryV1 } from './v0001_storyDataArrayToObject';
```
barklund: This is the main problem I see happening here. In migration 30 (just an example), you'd have to import types from 10 different previous migrations and you have to check which one has the most recent text element, page, etc. It seems like a lot of work to add a migration suddenly, because you have to track down when the involved types changed last time?

packages/migration/src/migrations/v0002_dataPixelTo1080.ts
```diff
@@ -22,41 +28,51 @@ const OLD_PAGE_HEIGHT = 732;
 const SCALE_X = NEW_PAGE_WIDTH / OLD_PAGE_WIDTH;
 const SCALE_Y = NEW_PAGE_HEIGHT / OLD_PAGE_HEIGHT;
 
-function dataPixelTo1080({ pages, ...rest }) {
+export type StoryV2 = StoryV1;
+export type PageV2 = PageV0;
+export type UnionElementV2 = UnionElementV0;
```
barklund: This makes it easier, because you always just extend the definition from the previous migration, but you'd have to do so for all the types. not just these. Also for each separate element type and potentially also all the resource types, etc.

packages/migration/src/migrations/v0002_dataPixelTo1080.ts
```diff
@@ -22,41 +28,51 @@ const OLD_PAGE_HEIGHT = 732;
...
+
+function dataPixelTo1080({ pages, ...rest }: StoryV1): StoryV2 {
   return {
     pages: pages.map(reducePage),
     ...rest,
   };
 }
 
-function reducePage({ elements, ...rest }) {
+function reducePage({ elements, ...rest }: PageV0): PageV0 {
```
barklund: `): PageV2` no? The same for element below.

packages/migration/src/migrations/v0002_dataPixelTo1080.ts
```diff
@@ -22,41 +28,51 @@ const OLD_PAGE_HEIGHT = 732;
...
   const element = {
     x: dataPixels(x * SCALE_X),
     y: dataPixels(y * SCALE_Y),
     width: dataPixels(width * SCALE_X),
     height: dataPixels(height * SCALE_Y),
     ...rest,
   };
-  if (typeof fontSize === 'number') {
-    element.fontSize = dataPixels(fontSize * SCALE_Y);
+  if ('fontSize' in element && typeof element.fontSize === 'number') {
```
barklund, line 88: I feel like this should be done with type guard instead:

```ts
function isTextElement(e:ElementV1):e is TextElementV1 {
  return 'font' in e;
}
...
if (isTextElement(element) && typeof element.fontSize === 'number') {
```

This would also mean that the union type isn't required at all. I actually really want to avoid that union type. I feel like the types here shouldn't be exhaustive but minimalistic - so interfaces over types. That does require these guards all over though, but that's also kind of required for narrowing a union type.

barklund (commented):
Some more thoughts on this.

I also found [this article](https://blog.saleae.com/versioning-typescript-types/), that has some ideas about how to go about this (recognizing some of the same issues, that we run into here). They made the version a generic, which I really like. I don't know if we can scale this concept to our solution though?

packages/migration/src/migrations/v0002_dataPixelTo1080.ts
miina: @barklund See [this comment as well](https://github.com/GoogleForCreators/web-stories-wp/pull/12320#discussion_r975099978), we could define the new types in each migration regardless if there are changes or not, then only the latest could be used.

packages/migration/src/migrations/v0002_dataPixelTo1080.ts
miina: Hm, but it's not correct like this though? Since the function actually can take all types, not just the text element.

What happens if there's a prop that's being used in multiple element types, like `resource`? Or `flip`? Also, the returned element can still be any element type, this would mean a lot of casting everywhere.

miina:
> I also found [this article](https://blog.saleae.com/versioning-typescript-types/), that has some ideas about how to go about this (recognizing some of the same issues, that we run into here). They made the version a generic, which I really like. I don't know if we can scale this concept to our solution though?

Nice find!

As far as I can see, it would still require defining a new type in each file and using that as the new base?

packages/migration/src/migrations/v0002_dataPixelTo1080.ts
miina: Also, I understand that generally we're using just the `Element` type for `Page` since potentially there might be other elements added so better to not use the union type, however, in these migration files we had those specific element types so it seems incorrect to use a generic base `Element`. Not sure what would be the benefit of using a minimalistic type in this case if these migration files were created for very specific element types. Thoughts?

miina: Note: we discussed this in the backlog grooming and decided to convert a few files using the type guarding to see how it seems.

@barklund PTAL

miina: Conclusion: still going to use the union type as before since otherwise TS won't really recognize nor accurately alert about types not matching and we'd have to do a lot of casting for getting rid of the TS errors.

miina:
@swissspidy When you have a moment, can you also check if all seems ok to you with the current approach? I'm at V10 right now, wouldn't want to get to the end to then start changing all the files again 😅

miina:
Some remaining topics:
- Getting the following TS error now starting from V35 -- due to extending the Story and Page in every file, it complains about the depth of it... Should we disable the rule for this package? An alternative could be re-defining the types fully after each 30 migrations for example. Thoughts?
Screenshot 2022-09-28 at 18 14 20

- Each file now includes a versioned type for `Story`, `Page`, all element types. However, there are some more types that are migrated, for example, `Resource` is a quite common one but there are some others that are changed just once or twice. I'm currently thinking that perhaps for more common types (such as `Resource` per each element), we could define a type per each file but only after the type is changed for the first time. E.g. the `resource` is introduced in V4 but it's modified the first time for Video in V13, so perhaps it would make sense to start adding the VideoResource type to each file starting from V13?
- Not sure what to do with types that are rarely touched, for example `Pattern` is changed once. Should we still add a PatternV[X] to each migration file?
- Let me know if you have a good idea for typing the [migration functions](https://github.com/GoogleForCreators/web-stories-wp/pull/12320/files#diff-8bc4f7b04431de4a5f74ffff9c1544839e91ad032b855cf357aa59bb384e1585R70) in `migrate.ts` file (each migration fn accepts a different version of the story) -- still figuring out that one.

packages/migration/src/migrations/test/v0005_setOpacity.js
```diff
@@ -55,7 +55,7 @@ describe('setOpacity', () => {
           elements: [
             {
               _test: 'element1',
-              opacity: 100,
+              opacity: 0,
```
miina, line 58: Opacity was set but `0` so it should stay as it was.

packages/migration/src/migrations/test/v0012_setBackgroundTextMode.js
```diff
@@ -32,15 +32,17 @@ describe('setBackgroundTextMode', () => {
                 _test: 'element1',
                 type: 'text',
                 backgroundColor: '#000',
+                content: '',
               },
               {
                 _test: 'element2',
                 type: 'text',
-                backgroundColor: 'transparent',
```
miina, line 39: Background color can't be `transparent` at this point, it was migrated to `Pattern` in V6

packages/migration/src/migrations/test/v0042_removeTrackName.js
```diff
@@ -23,9 +23,11 @@ describe('removeTrackName', () => {
     expect(
       removeTrackName({
         backgroundAudio: {
-          src: 'https://example.com/audio.mp3',
-          id: 123,
-          mimeType: 'audio/mpeg',
+          resource: {
```
miina, line 26: Story background audio was migrated to have `resource` in V39.

packages/migration/src/migrations/v0007_setFlip.ts
```diff
@@ -0,0 +1,98 @@
...
+function reducePage({ elements, ...rest }: PageV6): PageV7 {
+  return {
+    elements: elements.map(updateElement),
+    ...rest,
+  };
+}
+
+function updateElement(element: UnionElementV6): UnionElementV7 {
+  // If it's a text element or already has flip set, return as is.
+  if ('content' in element || 'flip' in element) {
```
miina, line 86: Not sure if there's a better way for TS to understand it's a text element, so using `'content' in element` since that's unique to text element at this point.

packages/migration/src/migrations/v0030_mark3pVideoAsMuted.ts
```diff
@@ -0,0 +1,106 @@
...
+  UnionElementV29,
+  VideoElementV29,
+} from './v0029_unifyGifResources';
+
+export type TextElementV30 = TextElementV29;
+export type ProductElementV30 = ProductElementV29;
+export type ShapeElementV30 = ShapeElementV29;
+export type ImageElementV30 = ImageElementV29;
+
+// @todo Should we drag resource versions along in each file? Currently, basing on ResourceV0 and extending fully per file.
```
miina: Waiting for thoughts about this... Currently creating a new Resource for each element is done starting from V31, should we do it since V1 for each file even though it's not used much there?

packages/migration/src/migrations/v0042_removeTrackName.ts
```diff
@@ -0,0 +1,112 @@
...
+  ImageElementV41,
+  PageV41,
+  ProductElementV41,
+  ShapeElementV41,
+  StoryV41,
+  TextElementV41,
+  VideoElementV41,
+} from './v0041_removeFontProperties';
+
+// @todo Should we now add Track version to each file instead?
```
miina, line 34: Waiting for feedback here, too -- `Track` is changed twice within all files, should we still version it in each file?

packages/migration/src/migrate.ts
```diff
@@ -57,12 +62,12 @@ import changeBaseColorToHex from './migrations/v0036_changeBaseColorToHex';
...
-import andadaFontToAndadaPro from './migrations/v0040_andadaFontToAndadaPro.js';
-import removeFontProperties from './migrations/v0041_removeFontProperties.js';
-import removeTrackName from './migrations/v0042_removeTrackName.js';
+import andadaFontToAndadaPro from './migrations/v0040_andadaFontToAndadaPro';
+import removeFontProperties from './migrations/v0041_removeFontProperties';
+import removeTrackName from './migrations/v0042_removeTrackName';
 import removeTagNames from './migrations/v0043_removeTagNames';
 
-const MIGRATIONS = {
+const MIGRATIONS: Record<number, any[]> = {
```
miina: Still don't have a solution for a generic `MigrationFn` type (instead of `any[]`). Any (non-`any` 😛 ) ideas are appreciated.

packages/migration/src/migrations/v0030_mark3pVideoAsMuted.ts
swissspidy: Seems like a lot of repetition. If we can avoid that, why not :)

packages/migration/src/migrations/v0042_removeTrackName.ts
swissspidy: Seems like a lot of repetition. If we can avoid that, why not :)

packages/migration/src/migrate.ts
swissspidy: It's a tough one. Unfortunately [the article Morten shared before](https://blog.saleae.com/versioning-typescript-types/) doesn't really cover that for a case like ours.

I've briefly looked into using union types here or something else clever, but couldn't find a nice way for it.

Soo... maybe we just use `any` for now until we find something better? 🙈

barklund (approved):
For now, this has to do. I have a weird vision about how we might make a properly typed migration system, but I'll have to play around with it. Great effort so far!

miina:
Merging with failing Karma tests since those are unrelated.

PR #12321, opened 2022-09-15
Media: Clean up linked post meta on attachment deletion

### LINKED ISSUE #12240: Media: `web_stories_muted_id` not cleaned up on attachment deletion
## Bug Description

When muting a video, we set the `web_stories_muted_id` post meta on the original pointing to the muted version.

When deleting the muted version of a video, we do not reset/delete `web_stories_muted_id` accordingly, causing an invalid ID to be lingering in post meta.

## Expected Behaviour

**Cleanup**

Delete `web_stories_muted_id` post meta on the original when the muted version is deleted.

Similar to how we delete video posters when deleting an attachment in `\Google\Web_Stories\Media\Video\Poster::delete_video_poster()`

**Hardening**

In addition to that, we should update the `getMutedMediaById` API callback to catch errors from calling `getMediaById()` and return `null` in that case.

## Steps to Reproduce

1. Add video
2. Mute video
3. Delete muted video
4. Try to mute video again
5. Get error

## Additional Context

- Plugin Version: 1.24.0
- WordPress Version: 6.0
- Operating System: macOS
- Browser: Chrome

spacedmonkey:
## Context

If you delete a muted / optimized version of a video, also delete the post meta to link to the original video. 

## Relevant Technical Choices

I created a util class so that the code can be shared. I thought about an abstract and traits to share code. A util class is what I needed up with. 

I would like to avoid doing a meta query, but I can't think of a way around it, other than a raw database query.   

## Testing Instructions

1. Add video
2. Mute video
3. Permanently delete muted video from WP media library
4. Try to mute video again
5. See no error

No

No

No

## Checklist

---

Fixes #12240


swissspidy:
Can we do `delete_metadata( 'post', null, 'web_stories_optimized_id', $attachment_id, true );` instead?

The only difference is that this works on all post types, not just attachments. But since these are prefixed meta keys that ar only used by us and only for attachments, that's OK.

This will greatly simplify the code and we don't have to establish a `Utils` class or similar (which I am not keen on adding)

spacedmonkey:
> Can we do delete_metadata( 'post', null, 'web_stories_optimized_id', $attachment_id, true ); instead?

I always for forget about that with the metadata api. That is way simpler. I have pushed a change to be do this.

swissspidy (approved):

kkalarickal:
Error Reproduction: 

I recreated the issue in `main-branch` and got the following error on console:


Bug Fix:

I switched over to `#12321` branch and confirmed via [test-web-story](https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/post-new.php?post_type=web-story#page=%2522a6576e4a-3a7e-4611-98a9-0f6051339c8f%2522) the following sequence worked:

- create a story and adds a page with video embedded
- confirm that story plays with sounds
- press remove-audio button and confirm that a new muted video has been uploaded to media library
- delete that muted video from the media library
- create another page and embed a different video with audio into the page
- now you can play audio, but if you try to "reduce audio" once again, there is NO console error and its possible to reduce audio a second time.

kkalarickal:
Permutations: 

In order to test around other media clean up events, I added background audio and then deleted the original file, to confirm there are no side effects.

PR #12330, opened 2022-09-15
Fix Floating Menu failing tests / random failing tests

### LINKED ISSUE #12334: Fix Floating Menu failing tests, random failing tests

merapi:
## Testing Instructions

1.

## Checklist

---

Fixes #12334


.github/workflows/tests-karma-editor.yml
```diff
@@ -101,9 +101,9 @@ jobs:
 
       - name: Run integration tests with Percy
         run: |
-          npx percy exec --parallel --config=percy.config.karma.yml -- \
+          npx percy exec --quiet --parallel --config=percy.config.karma.yml -- \
             npm run test:karma:story-editor -- --headless --viewport=1600:1000 --coverage --shard=$SHARD || \
-          npx percy exec --parallel --config=percy.config.karma.yml -- \
+          npx percy exec --quiet --parallel --config=percy.config.karma.yml -- \
```
swissspidy, line 107: 👍

merapi: Now it will just print errors and skip:
```
2022-09-19T08:21:26.5310152Z .[percy] Snapshot taken: Inline style override should keep formatting when all text is replaced: "A new text" in bold
2022-09-19T08:21:33.1766490Z .[percy] Snapshot taken: Inline style override should have correct formatting deleting text with one formatting, ending up in different formatting: "Full in some text" in bold
2022-09-19T08:21:36.8470745Z .[percy] Snapshot taken: Inline style override CUJ: Creator Can Style Text: Apply B, Apply I should have correct formatting when already italic, then inline removing italic and adding bold, then inserting text: "Fifooll in some text" in mixed formatting
2022-09-19T08:21:44.4785788Z .[percy] Snapshot taken: Inline style override when cursor is placed after second character CUJ: Creator Can Style Text: Apply Uppercase should have correct formatting when pressing uppercase toggle, then inserting text: "Fifooll in some text" in mixed formatting
2022-09-19T08:21:49.4178276Z .[percy] Snapshot taken: Inline style override when cursor is placed after second character CUJ: Creator Can Style Text: Select weight should have correct formatting when selecting font weight, then inserting text: "Fifooll in some text" in mixed formatting
2022-09-19T08:21:53.0389335Z .[percy] Snapshot taken: Inline style override when cursor is placed after second character CUJ: Creator Can Style Text: Apply B should have correct formatting when pressing bold toggle, then inserting text: "Fifooll in some text" in mixed formatting
2022-09-19T08:21:55.9117637Z .[percy] Snapshot taken: Inline style override when cursor is placed after second character CUJ: Creator Can Style Text: Apply B should have correct formatting when pressing mod+b for bold, then inserting text: "Fifooll in some text" in mixed formatting
2022-09-19T08:21:58.7617807Z .[percy] Snapshot taken: Inline style override when cursor is placed after second character CUJ: Creator Can Style Text: Apply U, Apply I should have correct formatting when pressing underline toggle, then mod+i, then inserting text: "Fifooll in some text" in mixed formatting
2022-09-19T08:22:01.4889864Z .[percy] Snapshot taken: Inline style override CUJ: Creator can Add and Write Text: Select all text inside a textbox, Write/edit text should keep formatting when all text is removed, then replaced: "Even more text" in bold
```

merapi:
Looks like scrollbar styles are not applied when testing locally (`overrideScrollbarCSS.js` is not even fired), causing issues like:
```diff
- Chrome 105.0.0.0 (Mac OS 10.15.7) Zoom selector should allow setting zoom to 200% and work correctly FAILED
	Expected element to not be horizontally scrollable, but overflows -2 pixels
```

merapi:
Fixed the above.
Looks like when converting from `injectGlobal` to `createGlobalStyle` the returned `OverrideScrollbarCSS` component was not included anywhere.

image

packages/story-editor/src/components/panels/design/textStyle/karma/textStyle.other.karma.js
```diff
@@ -290,8 +290,9 @@ describe('Text Style Panel', () => {
       const DEFAULT_CUSTOM_FONTS = 2;
       // Timeout used for submitting / search update + 250ms (250 + 250).
       const TIMEOUT = 500;
-      const openFontPicker = async () => {
-        const input = fixture.screen.getByLabelText('Font family');
+      const openFontPicker = async (i = 0) => {
```
miina, line 293: Maybe add it to the `TextStyle` class in the `karma` folder instead? To make it available via `fixture.editor.sidebar.designPanel.textStyle.fontFamily` or something similar.

karma-story-editor.config.cjs
```diff
@@ -149,6 +149,7 @@ module.exports = function (config) {
         snapshots: config.snapshots || false,
         snapshotsDir: '.test_artifacts/karma_snapshots',
         defaultViewport: getViewport(config.viewport),
+        args: config.headless ? undefined : ['--window-size=1600,1000']
```
swissspidy, line 152: Can you elaborate on this change and why we need this now? Also, wouldn't `args: ['--window-size=1600,1000']` work just as well?

merapi: Without this the browser window is smaller than the viewport when testing locally (you cannot see the whole editor).
`args: ['--window-size=1600,1000']` will probably work just fine, I just didn't want to set it when it's not needed (on CI).

packages/story-editor/src/components/panels/design/textStyle/karma/textStyle.other.karma.js
merapi: But we don't need font family in a fixture variable. I did this change because:
`openFontPicker` is used to click on the `Font family` button (so then you can test search etc), the issue occurs when we have multiple Font Pickers on the screen (Style tab + Floating Menu), most of the time we will have just one Font Picker on the screen and just in Floating Menu tests we will have two.

swissspidy (approved):

packages/karma-puppeteer-launcher/src/index.cjs
```diff
@@ -35,7 +35,8 @@ function puppeteerBrowser(baseBrowserDecorator, config) {
     const defaultPuppeteerOptions = {
       product: 'chrome',
       slowMo: 0,
-      dumpio: true,
+      // Only enable for debugging
+      // dumpio: true,
```
barklund, line 39: How do we enable that for debugging? Is it something we use when running tests locally or is it when debugging the test setup itself? If used regularly, should we make it a CLI flag?

packages/story-editor/src/components/canvas/karma/canvasKeys.karma.js
```diff
@@ -154,6 +154,7 @@ describe('Canvas Keyboard Shortcuts', () => {
     await fixture.events.keyboard.shortcut('mod+enter');
 
     // check that 'Play Animation' toggle button is there
+    await fixture.events.sleep(1000);
```
miina: Could we do this with `waitFor` instead?
E.g. something like
```
waitFor(() => {
    if (!fixture.screen.getByLabelText('Play Page Animations')) throw new Error('node not ready');
  })
´´´

File: packages/story-editor/src/components/floatingMenu/karma/video.karma.js [hunk 6]
```diff
@@ -28,6 +28,8 @@ describe('Video Design Menu: Keyboard Navigation', () => {
     fixture = new Fixture();
     await fixture.render();
     await fixture.collapseHelpCenter();
+    await fixture.showFloatingMenu();
+    await fixture.events.sleep(500);
```
miina: Here as well, any chance we could use `waitFor` and throwing an error instead of `sleep`?

miina (approved):
Thanks for all the work on the Karma tests!

packages/story-editor/src/karma/fixture/containers/canvas.js
```diff
@@ -72,7 +72,7 @@ export class Canvas extends Container {
 
   get quickActionMenu() {
     return this._get(
-      this.getByRole('menu'),
+      this.getByTestId('Element quick actions'),
```
barklund: Not a big fan of this one - we should avoid using test-ids completely (as they're a "fake" way to recognize elements).

I do see that the quick-action menu element applies the label to the wrong element. So if we move [this line](https://github.com/GoogleForCreators/web-stories-wp/blob/7722b81051310337bfd0cb186f14a21a143438e6/packages/design-system/src/components/contextMenu/contextMenu.js#L61) to the `Menu` a few lines below (or maybe just duplicate it to avoid breaking something else), the following should work:

```js
      this.getByRole('menu', { name: /Group of available options/ }),
```

barklund (approved):
🎉

packages/story-editor/src/components/canvas/karma/canvasKeys.karma.js
merapi: We enter `duration=1000` in input and then wait for that amount in another place too, so I just made it a variable.

packages/karma-puppeteer-launcher/src/index.cjs
merapi: If you do `dumpio: true` then you will see more logs (including console.logs) it is used rarely on CI and either way you need to commit this change to CI and wait for results (and then remove it) so with the CLI flag it would be the same.

PR #12335, opened 2022-09-19
Media: Video Segmentation Prototype

### LINKED ISSUE #12164: Media: Video Segmentation Prototype
## Feature Description

Related: #12093

It would be great if we could offer users an option to automatically split videos into smaller parts.

Example: you upload a 2 minute video and we allow you to split it into 15 seconds long segments (so 8 parts in total), putting every segment on its own page.

Imagine doing this already on initial upload, before we actually upload to the server. That would be super cool.

For this prototype we can create a new feature flag and some simple button (and maybe a range slider to set the segmentation length) somewhere that triggers this action.

## Additional Context

[FFmpeg segment muxer](https://ffmpeg.org/ffmpeg-formats.html#segment_002c-stream_005fsegment_002c-ssegment) seems to be what we need for this. Best to test this locally:

`-f segment -segment_time <duration>`

For the ffmpeg.wasm implementation, handling multiple outputs:

https://github.com/ffmpegwasm/ffmpeg.wasm/blob/9c14cca3cdf1fc0f2328f05586b48b74d1150816/examples/node/multiple-output.js?rgh-link-date=2020-05-05T09%3A58%3A47Z

timarney:
## Context

Automatically split videos into smaller parts.

Example: you upload a 2 minute video and we allow you to split it into 15 seconds long segments (so 8 parts in total), putting every segment on its own page.

For this prototype we can create a new feature flag and some simple button (and maybe a range slider to set the segmentation length) somewhere that triggers this action.

## Summary

Adds a segment video panel under styles when you select a video

Screen Shot 2022-09-19 at 7 18 11 AM

Pressing segment will split the selected video into parts based on the number of seconds selected

## Testing Instructions

QA Notes added some known issues for follow-up

https://github.com/GoogleForCreators/web-stories-wp/issues/12463
https://github.com/GoogleForCreators/web-stories-wp/issues/12464
https://github.com/GoogleForCreators/web-stories-wp/issues/12465

👉 Note: the video is segmented based on the videos [keyframes](https://en.wikipedia.org/wiki/Key_frame#:~:text=Whenever%20a%20drastic%20change%20to%20the%20image%20occurs%2C%20such%20as%20when%20switching%20from%20one%20camera%20shot%20to%20another%2C%20or%20at%20a%20scene%20change%2C%5B1%5D%20a%20key%20frame%20must%20be%20created) i.e. using a 30 sec segment length will not result in exactly 30 secs segments but rather the segments will be based on the closet keyframe.

> Note that splitting may not be accurate, unless you force the reference stream key-frames at the given time. See the introductory notice and the examples below.

- Enable the segment video feature
- upload a video to be segmented
- select the video from the canvas
- visit the style tab / video segment panel
- select the # of seconds to segment the video into
- press segment
- the video should get segmented + split onto multiple pages  

No

No

No

## Checklist

---

Fixes #12164


timarney:
Going to re-factor some pieces based on conversation with @swissspidy .

timarney:
@spacedmonkey @swissspidy --- adding this as ready for review now to get some feedback --- still testing on my end.

1) I think the UI (to let the user know something is happening) needs more work --- currently just using a disabled button and the snackbar.

2) I'm seeing some cases where the segmented videos end up out of order --- typically seeing this error when that happens.

Screen Shot 2022-09-21 at 4 59 32 AM

packages/story-editor/src/app/media/utils/useFFmpeg.js
```diff
@@ -282,6 +282,65 @@ function useFFmpeg() {
...
+   */
+  const segmentVideo = useCallback(
+    async (file, segmentTime) => {
+      let ffmpeg;
+      try {
+        ffmpeg = await getFFmpegInstance(file);
+        const type = file?.type || MEDIA_TRANSCODED_MIME_TYPE;
+        const ext = getExtensionFromMimeType(type);
+        const outputFileName = getFileBasename(file) + '_%03d.' + ext;
+        await ffmpeg.run(
```
spacedmonkey, line 308: Should we optimize this video at the same time?

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -567,12 +571,38 @@ function useProcessMedia({
...
+    async ({ resource: oldResource, segmentTime }, handleSegmented) => {
+      const { src: url, mimeType } = oldResource;
+      const segmentedFiles = await ffSegmentVideo(
+        await fetchRemoteFile(url, mimeType),
+        segmentTime
+      );
+
+      const addElement = handleSegmented(segmentedFiles);
+
+      for (const file of segmentedFiles) {
```
spacedmonkey: Instead of a for loop, can we use `Promise.all` here.

packages/story-editor/src/app/media/utils/useProcessMedia.js
```diff
@@ -567,12 +571,38 @@ function useProcessMedia({
...
 
+  /**
+   * Segment video using FFmpeg.
+   *
+   * @param {import('@googleforcreators/media').Resource} resource Resource object.
+   * @param {Function} handleSegmented function to handle segmentedFile processing.
+   */
+  const segmentVideo = useCallback(
+    async ({ resource: oldResource, segmentTime }, handleSegmented) => {
+      const { src: url, mimeType } = oldResource;
```
spacedmonkey: Should we wrap this in a try catch and handle the error.

packages/story-editor/src/app/media/utils/useFFmpeg.js
swissspidy: Related Slack discussion: https://xwp.slack.com/archives/C013J2EGH8R/p1663840238602189?thread_ts=1663837781.331869&cid=C013J2EGH8R

Since the same can be asked for trimming and cropping as well, I'd suggest creating a new ticket to discuss this.

packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js
```diff
@@ -0,0 +1,204 @@
...
+          addPageAt({
+            page: createPage(),
+            position: originalPageIndex + processedFiles,
+          });
+        } else {
+          // remove the original element
+          deleteElementById({ elementId: originalElementId });
+        }
+
+        insertElement('video', element);
```
spacedmonkey:
```suggestion
        insertElement(ELEMENT_TYPES.VIDEO, element);
```

packages/story-editor/src/app/media/utils/useFFmpeg.js
```diff
@@ -282,6 +282,65 @@ function useFFmpeg() {
...
+          .forEach(async (outputFile) => {
+            const data = await ffmpeg.FS('readFile', outputFile);
+            files.push(
+              blobToFile(new Blob([data.buffer], { type }), outputFile, type)
+            );
+          });
+        return files;
+      } catch (err) {
+        // eslint-disable-next-line no-console -- We want to surface this error.
+        console.error(err);
```
spacedmonkey, line 344: Not console, track error.

packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js
```diff
@@ -0,0 +1,204 @@
...
+    });
+
+    const originalElementId = elementId;
+    const pageIds = pages.map(({ id }) => id);
+    const originalPageIndex = pageIds.indexOf(currentPage.id);
+
+    await segmentVideo({ resource, segmentTime }, (segmentedFiles) => {
+      // return the "addElement" function
+      // called via "onUploadSuccess" of the segmented file
+      return (element) => {
```
spacedmonkey: Why is this await? Why are we returning a function here?

packages/story-editor/src/app/media/utils/useProcessMedia.js
timarney: If we use `Promise.all` the files could get out of order correct?

Right now the segmentedFiles are ordered ---the loop calls addElement which adds the new page / element  --- based on the order of the array.

spacedmonkey: Is there no to upload the files the add the elements in the correct order?

spacedmonkey:
Screenshot 2022-09-22 at 11 11 27

This UI should check the length of the original video and not give option to break in to segments longer than the video. I could not be only able to select valid options. This UI should be hidden if shorter than 20 seconds.

packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js
timarney: This gets us "early access" to the segmented files array `before` they get uploaded.  

The function passed get called here https://github.com/GoogleForCreators/web-stories-wp/blob/addcf91b3ca815534c89c0dc1707ad268065ccab/packages/story-editor/src/app/media/utils/useProcessMedia.js#L588

It gets called with the segmentedFiles  as a param so which we need to check the length in addElement to see if all the uploads are done.

Calling that function returns the " addElement function" which gets called per upload one at a time --- and we check the  segmentedFiles length.

packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js
```diff
@@ -0,0 +1,204 @@
...
+`;
+
+function VideoSegmentPanel({ pushUpdate, selectedElements }) {
+  const resource = getCommonValue(selectedElements, 'resource');
+  const elementId = getCommonValue(selectedElements, 'id');
+  const enableSegmentVideo = useFeature('segmentVideo');
+
+  const onChangeSegmentTime = (value) => {
+    const newSegmentTime = value;
+    pushUpdate({ segmentTime: newSegmentTime }, true);
```
spacedmonkey:
```suggestion
    pushUpdate({ segmentTime: value }, true);
```

packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js
timarney: The `await ` is there so we can call  
```
showSnackbar({
      message: __('Inserting video segments', 'web-stories'),
      dismissible: false,
    });
    
 ```
 
 After the ` segmentVideo` returns ...

File: packages/story-editor/src/app/media/utils/useProcessMedia.js (hunk 2 above)
Comment (timarney) [reply packages/story-editor/src/app/media/utils/useProcessMedia.js]: Outside of the order of the array --- it would be possible to check the filenames as they get appended with 001, 002 etc... but we would need logic to add pages / elements between items i.e. file 4 is ready but 2, 3 haven't finished so we need to insert pages 2,3,4 to add the `4` element.


I think it ends up being cleaner / simpler to use the array order.

**Comment by spacedmonkey** [comment]
https://user-images.githubusercontent.com/237508/191723828-1c28a6b7-8394-48c9-bc2c-011bf748ab80.mov

In my testing, I found two things. 
- It seems like some of the segments are 19 / 21 seconds. This seems odd, worth looking into. 
- When the page is created, it automatically changes to the new page and starts playing the video. This makes a lot of noise.

File: packages/story-editor/src/app/media/utils/useProcessMedia.js (hunk 2 above)
Comment (spacedmonkey) [reply packages/story-editor/src/app/media/utils/useProcessMedia.js]: As noted [here](https://github.com/GoogleForCreators/web-stories-wp/pull/12335#issuecomment-1254836974), I think it might be best to upload all the video first before creating the pages. What if one or more of the videos fail to upload, that could make a mess, where page 2 of 5 is missing.

File: packages/story-editor/src/app/media/utils/useFFmpeg.js (hunk 5 above)
Comment (timarney) [reply packages/story-editor/src/app/media/utils/useFFmpeg.js:333]: > Screen.Recording.2022-09-22.at.11.25.31.mov 
> In my testing, I found two things.
> 
> * It seems like some of the segments are 19 / 21 seconds. This seems odd, worth looking into.
> * When the page is created, it automatically changes to the new page and starts playing the video. This makes a lot of noise.


> It seems like some of the segments are 19 / 21 seconds. This seems odd, worth looking into.

I added a note about that under testing in the summary above

 👉 Note: the video is segmented based on the videos keyframes -- we can force keyframes to happen at more exact intervals but this will result in larger videos --- as the keyframes we would be forcing to ensure the segment length are for our purposes vs the actual file compression (more keyframes = larger file).  In this case the keyframe are frame `diff` information not "animation" keyframes.
 
> When the page is created, it automatically changes to the new page and starts playing the video. This makes a lot of noise.

Thinking the same --- I'm using the existing logic to add pages .... do we have a way to insert a new page without moving to it?

**Comment by timarney** [comment]
> Screen.Recording.2022-09-22.at.11.25.31.mov 
> In my testing, I found two things.
> 
> * It seems like some of the segments are 19 / 21 seconds. This seems odd, worth looking into.
> * When the page is created, it automatically changes to the new page and starts playing the video. This makes a lot of noise.


> It seems like some of the segments are 19 / 21 seconds. This seems odd, worth looking into.

I added a note about that under testing in the summary above

 👉 Note: the video is segmented based on the videos keyframes -- we can force keyframes to happen at more exact intervals but this will result in larger videos --- as the keyframes we would be forcing to ensure the segment length are for our purposes vs the actual file compression (more keyframes = larger file).  In this case the keyframe are frame `diff` information not "animation" keyframes.
 
> When the page is created, it automatically changes to the new page and starts playing the video. This makes a lot of noise.

Thinking the same --- I'm using the existing logic to add pages .... do we have a way to insert a new page without moving to it?

File: packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js [hunk 8]
```diff
@@ -0,0 +1,203 @@
...
+        <StyledSlider
+          value={segmentTime}
+          handleChange={onChangeSegmentTime}
+          minorStep={1}
+          majorStep={2}
+          min={20}
+          max={120}
+          aria-label={__('Segment length', 'web-stories')}
+        />
+        {segmentTime} {__('sec', 'web-stories')}
```
swissspidy: Never concatenate strings that need to be translated.

Use `sprintf` with placeholders & translator comments

includes/Experiments.php
```diff
@@ -347,6 +347,18 @@ public function get_experiments(): array {
...
+			 * Author: @timarney
+			 * Issue: #12164
+			 * Creation date: 2022-09-19
+			 */
+			[
+				'name'        => 'segmentVideo',
+				'label'       => __( 'Segment video', 'web-stories' ),
+				'description' => __( 'Enable support for segmenting video files', 'web-stories' ),
+				'group'       => 'editor',
+				'default'     => true,
```
swissspidy: Uhm, no? :-)

packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js
```diff
@@ -62,6 +62,8 @@ const HelperText = styled(Text).attrs({
 `;
 
 function VideoSegmentPanel({ pushUpdate, selectedElements }) {
+  const MIN_SEGMENT_LENGTH = 20;
+  const MAX_SEGMENT_LENGTH = 120;
```
swissspidy: We recommend videos that are less than 15 seconds per page, or 60 seconds maximum.

Let's reflect that here accordingly.

```suggestion
  const MIN_SEGMENT_LENGTH = 10;
  const MAX_SEGMENT_LENGTH = 60;
```

swissspidy:
> > It seems like some of the segments are 19 / 21 seconds. This seems odd, worth looking into.
> 
> I added a note about that under testing in the summary above

After thinking a bit more about it, I am definitely not a fan. Let's force the segments to be accurate.

> Thinking the same --- I'm using the existing logic to add pages .... do we have a way to insert a new page without moving to it?

I think we can add a new option to the `addPage` reducer to accomplish that.

packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js
```diff
@@ -0,0 +1,209 @@
...
+    <SimplePanel
+      name="videoOptions"
+      title={__('Video Segment Settings', 'web-stories')}
+    >
+      <Row spaceBetween>
+        <StyledSlider
+          value={segmentTime}
+          handleChange={onChangeSegmentTime}
+          minorStep={1}
+          majorStep={2}
```
swissspidy: Maybe 5?

packages/story-editor/src/app/media/utils/useFFmpeg.js
swissspidy: This is in line with the rest of the functions in this file.

includes/Experiments.php
timarney: bad commit --- I change this file when working with `Local WP` is there a better way to enable experiments when using `Local WP` ?

swissspidy: > is there a better way to enable experiments when using Local WP ?

Yes, see https://github.com/GoogleForCreators/web-stories-wp/blob/main/docs/feature-flags.md#turning-onoff-feature-flags-with-the-experiments-tab

packages/story-editor/src/app/media/utils/useFFmpeg.js
```diff
@@ -323,11 +387,7 @@ function useFFmpeg() {
           type
         );
       } catch (err) {
-        // eslint-disable-next-line no-console -- We want to surface this error.
-        console.log(err);
```
swissspidy: Why remove that here?

packages/story-editor/src/app/media/utils/useProcessMedia.js
swissspidy: `uploadMedia` accepts multiple files.

So instead of this for loop we can just do `await uploadMedia(segmentedFiles, { onUploadSuccess });`

packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js
swissspidy: This whole `handleSegmentation` callback, as well as the `handleSegmented` callback param passed to `segmentVideo` (which itself returns a callback) are very confusing. It would be good to refactor that to make it more understandable.

Could you perhaps write down the thoughts/ideas here so that we can figure this out together?

packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js
```diff
@@ -0,0 +1,209 @@
...
+    });
+
+    const originalElementId = elementId;
+    const pageIds = pages.map(({ id }) => id);
+    const originalPageIndex = pageIds.indexOf(currentPage.id);
+
+    await segmentVideo({ resource, segmentTime }, (segmentedFiles) => {
+      // return the "addElement" function
+      // called via "onUploadSuccess" of the segmented file
+      return (element) => {
```
swissspidy: NB: this function (which is an `onUploadSuccess` handler) receives `{ id, resource}`, not an "element".

packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js
```diff
@@ -0,0 +1,209 @@
...
+        if (!element.id.length) {
```
swissspidy: When does this happen and why do we need it?

packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js
spacedmonkey: >  It would be good to refactor that to make it more understandable.

+1 I find this code hard to follow. I am not sure of the benefit of doing it this way. I am personally not a massive fan of callback / return functions pattern in javascript. With `await` I don't think it is really needed anymore.

packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js
timarney: Will re-check --- might not be happening now given the re-factoring but was seeing this when using the Media Upload  Queue --- function was getting called twice (I think that was it.)

packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js
timarney: Okay to hash out some ideas 

We call ffMpeg segment --- which gives us back an array of files 

At this point we have the files in the correct order
```
[001,002,003]
```

We can call 

```
await uploadMedia(segmentedFiles, { onUploadSuccess }); // I started by using this
```

Which works well for the upload 

Once those are all done 

We can do the page insert(s) ...  we know how many pages / files to insert

> This is where I think it gets slightly tricky as we need to get the uploaded assets into the correct pages or if not tricky at least where I see the issue to solve.

What's the best way to do that?  --- The resource names files or maybe it was alt text have the input001, input002, input003 as names so might be able to do something with that ?  Or hopefully there's another way I haven't thought of.

=====

When using  uploadMedia([]) we can an onSuccess for each file correct? i.e. not when all the files are done

packages/story-editor/src/app/media/utils/useFFmpeg.js
timarney: Re-added

https://github.com/GoogleForCreators/web-stories-wp/pull/12335/files/4de45e185d683c1b89da934e9685baf6e095bea8#diff-8383cf154a6a03f61cb68645b7fd4267892b7df793c6b3cceb999e40bcf2522bR333

packages/story-editor/src/app/media/utils/useProcessMedia.js
timarney: Can follow up this conversation here:

https://github.com/GoogleForCreators/web-stories-wp/pull/12335#discussion_r978162303

packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js
swissspidy: Yeah that's a great question on how to do that. First time we really have to care about the upload/insertion order.

Yes, the `onUploadSuccess` is called for each file individually. And all it receives is the final `Resource` object.

I wouldn't wanna rely on individual properties of the resources (like alt text) to determine the order, it's fragile and can change later without realizing that it breaks something.

Off the top of my head, I think we can do something like this to ensure correct order:

1. In `useUploadMedia.js`, update the `.map()` callback in `uploadMedia()` so that we pass the `file`'s index in the array to `additionalData`
  Something like `additionalData.batchPosition = files.length - 1 - index;` (because the `files` array is actually reversed)
2. Then, in the same `useUploadMedia.js` file, where the `onUploadSuccess()` callback is called, pass `batchPosition` to it
3. This way, we know at which position to insert the video
4. In `onUploadSuccess` callback, insert uploaded video at `currentPageIndex + 1 + batchPosition`

I hope that works :-)

swissspidy:
> 2\. I'm seeing some cases where the segmented videos end up out of order --- typically seeing this error when that happens.
> 
> Screen Shot 2022-09-21 at 4 59 32 AM

Sounds like we can add some hardening there. Mind filing a ticket?

packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js
spacedmonkey: There is currently no error handling. Error handling is important. If you have say 5 videos to upload in one go, then the likelihood one or more files will fail in higher. This would result in some weird side effects, like missing middle pages etc. For this prototype to be merged, I feel like we need to solve these problems. 

Regarding knowing the order, how about adding to the additionalData then saving this as part of the post meta. From there we can work on the order, maybe with a new attribute on the resource object.

spacedmonkey: It is also worth noting that, file name is a good way to get position. If for example, file001.mp4 already existing on upload, is can result in a file name like this, file001-1.mp4.

timarney: Pushed up a WIP - commit

https://github.com/GoogleForCreators/web-stories-wp/pull/12335/commits/f4b625b794a12c6c71989b91505ac5402c1ffcbd

1) added the index to the  additionalData 
2) added batchCount to the  additionalData
3) pushing those into a segmentedFiles array on upload success callback--- 

```
 segmentedFiles.push({ batchPosition, newResource });
```

For each uploadSuccess callback I'm checking the segmentedFiles.length against the  additionalData.batchCount  length to see if we're done (handles ordering and finish uploading all files --- cases)

4) If we're done I'm sorting the segmentedFiles based on the batchPosition to get the array into the original order

That's all working

======

Next steps --- error handling + inserting the pages and elements

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
```diff
@@ -421,6 +421,7 @@ function useMediaUploadQueue() {
       finishUploading({
         id,
         resource: newResource,
+        additionalData
```
swissspidy: This (and the change to the reducer) seem a bit pointless. It just passes through `additionalData` without modifying it.

packages/story-editor/src/app/media/useUploadMedia.js
```diff
@@ -335,7 +347,12 @@ function useUploadMedia({
             onUploadProgress,
             onUploadError,
             onUploadSuccess,
-            additionalData,
+            additionalData:
+            {
+              batchPosition: files.length - 1 - index,
+              batchCount: files.length,
+              ...additionalData
```
swissspidy:
```suggestion
              ...additionalData
              batchPosition: files.length - 1 - index,
              batchCount: files.length,
```

Also, `additionalData` might not be `undefined`, so let's initialize it to an empty object in the function params.

packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js
```diff
@@ -0,0 +1,209 @@
...
+    const pageIds = pages.map(({ id }) => id);
+    // const originalPageIndex = pageIds.indexOf(currentPage.id);
+
+    let segmentedFiles = [];
+
+    await segmentVideo({ resource, segmentTime }, ({ resource: newResource, batchPosition, batchCount }) => {
+
+      const exists = segmentedFiles.find(item => item.batchPosition === batchPosition);
+      if (!exists) {
+        segmentedFiles.push({ batchPosition, resource: newResource });
```
swissspidy: Can just do `segmentedFiles[ batchPosition ] = newResource`. Then there's no need for sorting.

packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js
swissspidy: > There is currently no error handling. [..] For this prototype to be merged, I feel like we need to solve these problems.

I'd disagree with this, since proper error reoslution very much depends on UX feedback.

> It is also worth noting that, file name is a good way to get position.

Let's not rely on things like alt text or file names. It's fragile.

> For each uploadSuccess callback I'm checking the segmentedFiles.length against the additionalData.batchCount length to see if we're done

Let's find a way to avoid mixing the logic for an individual item's success & the success of the whole batch operation.

Off the top of my head:

1. `uploadMedia()` could do something like `const batchId = uuidv4(); additionalData.batchId = batchId;` and then _return_ the `batchId` to the caller, so that we can store that in state.
2. Add a new `isBatchUploading(batchId)` selector to `useMediaUploadQueue`
3. In `videoSegment`, add a new effect that checks `isBatchUploading` and then creates the pages & inserts elements
4. For that to work, `segmentedFiles` can be in state

WDYT?

packages/story-editor/src/app/media/utils/useFFmpeg.js
spacedmonkey: I think I was wrong. Console error are fine in this file.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
timarney: Will re-check ... at one point I wasn't seeing the data I added coming in.

packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js
timarney: Sounds good -- will pick this back up next week and move in the isBatchUploading direction

timarney: Had an issue pushing to this branch ... will figure that out later

In the meantime I started moving things to an isBatchUploading check 

The queue seems to always be empty any idea why?

https://github.com/GoogleForCreators/web-stories-wp/pull/12370/files#diff-cf3c49123af8e30828deaed0e5d68b9c0c4010945833512d58aa63e85c0e0161R876

timarney: @swissspidy 

2 questions

1) Thoughts on best way to get the batchId back to videoSegment ... to check using the useEffect
2) ~I added the isBatchUploading --- the queue is always empty so I must not be using it correctly or something any ideas?~!

I see the issue is when it's checking the queue --- I added isUploading to the useEffect and now I see the the check happening.

packages/story-editor/src/app/media/utils/useProcessMedia.js
timarney: https://github.com/GoogleForCreators/web-stories-wp/pull/12335/commits/d7cd778051cea8e80aa5adc22d4f35e0e6a3aef3

timarney:
RE:

https://github.com/GoogleForCreators/web-stories-wp/pull/12335#issuecomment-1254883207

> After thinking a bit more about it, I am definitely not a fan. Let's force the segments to be accurate.

Added force key frames , segment times and segment delta 

Seeing times off by 1, 2 secs 

> Thinking the same --- I'm using the existing logic to add pages .... do we have a way to insert a new page without moving to it?

Reducer has been updated.

packages/story-editor/src/app/media/utils/useMediaUploadQueue/index.js
timarney: Seems fine without 

https://github.com/GoogleForCreators/web-stories-wp/pull/12335/commits/b3531bc0f5607b9ca26cfea30cf4631ed137a124

https://github.com/GoogleForCreators/web-stories-wp/pull/12335/commits/b2ea3a589d676309ed6e5821dec23423552a8af2

timarney:
@swissspidy ---

Thoughts on best way to get the batchId back to videoSegment ... to check using the useEffect ?

packages/story-editor/src/components/panels/design/videoSegment/videoSegment.js
timarney: Resolving this as the code has been re-factored quite a bit now 

Picking up here

https://github.com/GoogleForCreators/web-stories-wp/pull/12335#issuecomment-1260832981

packages/story-editor/src/app/story/useStoryReducer/actions.js
```diff
@@ -63,13 +63,16 @@ const setCurrentPage =
 
 const addElements =
   (dispatch) =>
-  ({ elements }) =>
-    dispatch({ type: types.ADD_ELEMENTS, payload: { elements } });
+  ({ elements, pageId }) =>
+    dispatch({ type: types.ADD_ELEMENTS, payload: { elements, pageId } });
```
spacedmonkey: Why is this needed?

packages/story-editor/src/app/story/useStoryReducer/actions.js
```diff
@@ -27,8 +27,8 @@ const addPage =
 
 const addPageAt =
   (dispatch) =>
-  ({ page, position }) =>
-    dispatch({ type: types.ADD_PAGE, payload: { page, position } });
+  ({ page, position, select }) =>
+    dispatch({ type: types.ADD_PAGE, payload: { page, position, select } });
```
spacedmonkey: Why is this needed?

packages/story-editor/src/app/story/useStoryReducer/reducers/addElements.js
```diff
@@ -45,13 +45,20 @@ const isNotProduct = ({ type }) => type !== ELEMENT_TYPES.PRODUCT;
...
   }
 
-  const page = draft.pages.find(({ id }) => id === draft.current);
+  let page = draft.pages.find(({ id }) => id === draft.current);
+
+  if (pageId) {
+    // insert elements into specific page
+    page = draft.pages.find(({ id }) => id === pageId);
+  }
+
```
swissspidy: Let's add tests for all changes to this reducer.

spacedmonkey:
I am seeing

https://user-images.githubusercontent.com/237508/192821714-515417de-385d-4336-bbe8-87893cef106a.mov

I am seeing this work with a 7 minute video. But after the videos are done processing, it then pages are not created. Any idea why this is happening?

packages/story-editor/src/app/story/useStoryReducer/reducers/addPage.js
```diff
@@ -39,8 +39,9 @@ import { isInsideRange } from './utils';
  * @param {Object} payload Action payload
  * @param {Object} payload.page Object with properties of new page
  * @param {Object} payload.position Position at which to insert the new page. If null, insert after current
+ * @param {boolean} payload.select Whether or not to select (give focus to) the newly created page
  */
-export const addPage = (draft, { page, position }) => {
+export const addPage = (draft, { page, position, select = true }) => {
```
swissspidy: Let's add tests for all changes to this reducer.

packages/story-editor/src/app/story/useStoryReducer/reducers/addPage.js
```diff
@@ -51,11 +52,17 @@ export const addPage = (draft, { page, position }) => {
   const currentPageIndex = draft.pages.findIndex(
     ({ id }) => id === draft.current
   );
-  const insertionPoint = isWithinBounds ? position : currentPageIndex + 1;
+
+  const insertionPoint =
+    isWithinBounds || draft.pages.length === position
```
swissspidy: So this now allows inserting a new page at `draft.pages.length`?

If so, why not just change the `isWithinBounds` check above?

```diff
  const isWithinBounds =
-  position !== null && isInsideRange(position, 0, draft.pages.length - 1);
  const isWithinBounds =
+  position !== null && isInsideRange(position, 0, draft.pages.length);
```

Easier to comprehend

packages/story-editor/src/app/story/useStoryReducer/actions.js
swissspidy: By default when adding a new page the editor jumps to that page (selects that page). But we want to avoid that for video segmentation.

packages/story-editor/src/app/story/useStoryReducer/actions.js
swissspidy: By default when adding a new element the editor adds it to the current page. But we want to avoid that for video segmentation, since we need to add each video on a separate page.

swissspidy:
> In my testing, I was seeing a video uploading with 0:00. It seems to be less than 1 second. Maybe we should not upload the video unless it more than 1 second?

Not uploading would possibly mean losing parts of the video.

One of many UX questions to resolve after prototype.

> But after the videos are done processing, it then pages are not created. Any idea why this is happening?

 I think that's simply because of the open to-do Tim mentioned above & in our sync

packages/story-editor/src/app/story/useStoryReducer/actions.js
timarney: Given we don't move to a page --- we need a way to insert elements into a page your not on.

https://github.com/GoogleForCreators/web-stories-wp/pull/12335/files#diff-650d4464ec19595eabe96bbb3dff9ad45b36e53ded1c75a56080e84826a78021R58

packages/story-editor/src/app/story/useStoryReducer/actions.js
timarney: This is whether we want to select the page (as the current page) i.e. have focus move to it or not --- but default it will select / focus 


We don't want that in this case --- we want to create a page and not move to it.

See:
https://github.com/GoogleForCreators/web-stories-wp/pull/12335/files#diff-f4cbc92ce47cdf5411446ab37de62b7987e2ecf34d72336d0f83efc73852c00aR62

timarney:
> I am seeing this work with a 7 minute video. But after the videos are done processing, it then pages are not created. Any idea why this is happening?

A few things --- I've been testing mostly with 2, 4 min videos to date

- Any chance your using local WP can't quite tell in your video    --- seems to drop the connection some times
- Any ffMpeg errors ?--- should have some log output around exit() ... there's a lot of noise with frame creation so hard to spot
- The hard coded batchId could mess things up for sure
- Looks like you ran into this issue https://github.com/GoogleForCreators/web-stories-wp/issues/12377 when can trip things up

packages/story-editor/src/app/media/utils/useFFmpeg.js
```diff
@@ -282,6 +282,81 @@ function useFFmpeg() {
...
+          '-map',
+          '0',
+          '-force_key_frames',
+          `${segmentTimes}`,
+          '-f',
+          'segment',
+          '-segment_times',
+          `${segmentTimes}`,
+          '-segment_time_delta', //account for possible roundings operated when setting key frame times.
+          `${(1 / (2 * FFMPEG_CONFIG.FPS[1])).toFixed(2)}`, //
```
swissspidy: Empty comment?

packages/story-editor/src/app/story/useStoryReducer/reducers/addPage.js
timarney: Sure - my mind got a little hung up on the `Within` naming

packages/story-editor/src/app/media/utils/useFFmpeg.js
timarney: will remove --- was going to comment the resulting value as might be easier to read that line

packages/story-editor/src/app/story/useStoryReducer/reducers/addElements.js
timarney: Added a couple of tests here:
https://github.com/GoogleForCreators/web-stories-wp/pull/12335/commits/ac559dc37dbb6f4d310f3f8641bd65e1c4b88e0c

packages/story-editor/src/app/story/useStoryReducer/reducers/addPage.js
timarney: Added test here:
https://github.com/GoogleForCreators/web-stories-wp/pull/12335/commits/f35929b3f1993f5f7248fca1770a8cc8e867f77d

It's currently failing on the result.current

packages/story-editor/src/app/media/useUploadMedia.js
```diff
@@ -335,7 +347,11 @@ function useUploadMedia({
             onUploadProgress,
             onUploadError,
             onUploadSuccess,
-            additionalData,
+            additionalData: {
+              ...additionalData,
+              batchPosition: files.length - 1 - index,
+              batchId,
+            },
```
spacedmonkey, line 354: So every upload gets batch id?

swissspidy: Yes.

packages/story-editor/src/app/story/useStoryReducer/reducers/addPage.js
swissspidy: So... there's a bug then? :)

timarney:
@swissspidy --- taking another look / testing this.

We now end up with a blank page at the start --- is this what we want?

Screen Shot 2022-09-30 at 5 26 05 AM

The previous code created pages new pages "after" the first element -- for the first page I was inserting the new element (also selected it) + deleted the old one.

swissspidy:
> We now end up with a blank page at the start --- is this what we want?

It's not necessarily _blank_, it's just the current page you were on. There could have been other elements on it than just the video.

But in any case, yes, this is good enough for the prototype. It's one of the many UX questions we need to figure out later.

packages/story-editor/src/app/story/useStoryReducer/reducers/addPage.js
timarney: Looking now.

spacedmonkey:
> Not uploading would possibly mean losing parts of the video.

What could you miss in less than a second? A video of less than a second is not useful to a user. Maybe the cut off should be at least 75% of a second.

spacedmonkey:
I am still not seeing the pages create at all. Not seeing any errors. 


https://user-images.githubusercontent.com/237508/193244275-f7c6b347-16c4-4ba7-827f-84633286ff13.mov

swissspidy:
Let's figure out this scenario later as part of UX rounds. It's a good point.

As mentioned on Wednesday, I'd like us to write down things like this for the future.

packages/story-editor/src/app/story/useStoryReducer/reducers/addPage.js
timarney: > So... there's a bug then? :)

Looks like this was the issue

https://github.com/GoogleForCreators/web-stories-wp/pull/12335/commits/6a73ca119f4883f0655758001e78f88543ea9a94

timarney:
> What could you miss in less than a second? A video of less than a second is not useful to a user. Maybe the cut off should be at least 75% of a second.

For a later PR we could account for this type of thing when we calc the segment times 

https://github.com/GoogleForCreators/web-stories-wp/pull/12335/files#diff-8383cf154a6a03f61cb68645b7fd4267892b7df793c6b3cceb999e40bcf2522bR306

Check the length of the last segment time and if <x tag it onto the previous segment time --- something like that would prevent videos of x length when doing the split.

timarney:
This test is timing out 

https://github.com/GoogleForCreators/web-stories-wp/actions/runs/3158051440/jobs/5139649587#step:9:19

Any thoughts?

timarney:
> I am still not seeing the pages create at all. Not seeing any errors.

Re-tested with a few different videos

4-5 times worked ... the other time I didn't see any errors and no pages created.  Possibly hitting the early return on the add pages code.

 For sure need better UX feedback to show what's happening.

spacedmonkey:
@swissspidy @timarney 
I spent a lot time debugging this. The answer was straight in front of me. In my testing, after starting the segmenting running, I change to the media tab in the sidebar. This destorys all the state in `VideoSegmentPanel` and then the hook never runs and never creates the page. 

I believe we should start by moving everything into a hook, the click is a light length as is. We can use trimming component as a base to copy from.

swissspidy:
Ah, good to know. Yeah we can do such a refactoring as a follow-up item.

timarney:
Anything else needed on this one given it's behind the feature flag?

spacedmonkey:
Let's discuss this in our next meeting.

spacedmonkey:
I am not sure I want to approve this PR. Change the sidebar to another tab breaks it. At the very least, I would expect todos in the code, a callout in the testing / QA notes to say not to do this and a ticket explain how this functionality will be fixed in the future. Prototype that breaks this easily is going to make problems for those testing it and evaluating the functionality. I already used a lot of time debugging.

timarney:
@spacedmonkey 

Added a QA note with the following issues for follow-up

https://github.com/GoogleForCreators/web-stories-wp/issues/12463
https://github.com/GoogleForCreators/web-stories-wp/issues/12464
https://github.com/GoogleForCreators/web-stories-wp/issues/12465

swissspidy:
Thanks for creating the follow-up tickets 👍 

I think we're good to proceed here for now.

swissspidy (approved):

Issue #12126, opened 2022-08-17
TypeScript: Convert `elements` package


swissspidy:
## Task Description

The elements package must be converted to TypeScript.

The types for the specific elements as well as for general elements (and classes of elements, such as media elements) should be moved from the types package to the relevant package and refined once here.

At the same time, other types will move to where they belong and correctly referenced:

* resource types will be moved to the media package
* pattern types will be moved to the patterns package

All references to the types should be updated.

swissspidy:
See also #11257

swissspidy:
@barklund What's the current status on this one?

swissspidy:
I went ahead and converted only the `elements` package in #12508 to help unblock work on some other packages.

PR #12344, opened 2022-09-21
Code: Improved codebase by minimizing package interdependency

### LINKED ISSUE #12332: Reduce package interdependency
## Description

Reduce some references between packages (not all references are "hard" at this point, but would need to be once TypeScript is introduced, so we will update this ahead of time):

* [x] remove the link from animation package to media (by going through `getDefinitionForType` instead, which would instead add a link to the elements package in `getMediaBounds` function)
* [x] remove the link from media to element (by moving `generateVideoStrip` to the story-editor package)
* [x] remove the link from element-library to animation package (by passing the relevant `BG_SCALE_*` properties to `MediaEdit` from the story editor)
* [x] remove the link from rich-text package to element library package (by passing a "simpler" callback `(content: string) => void` to `usePasteTextContent`)

barklund:
## Context

This moves some code and/or responsibility from one package to another in four areas:

* [x] remove the link from animation package to media (by going through `getDefinitionForType` instead, which would instead add a link to the elements package in `getMediaBounds` function)
* [x] remove the link from media to element (by moving `generateVideoStrip` to the story-editor package)
* [x] remove the link from element-library to animation package (by passing the relevant `BG_SCALE_*` properties to `MediaEdit` from the story editor)
* [x] remove the link from rich-text package to element library package (by passing a "simpler" callback `(content: string) => void` to `usePasteTextContent`)

## Testing Instructions

To make sure no regressions are introduced, please thoroughly test these bits of functionality:

#### Video strip
1.  Add a local video to a page
2. Click the scissor icon in the design mode to enter trim mode
3. Observe that the video strip is correctly generated

#### Media pan animation
1. Add a landscape image as background element on a page
2. Open the animation dropdown
3. Verify that the element animation can be set to pan left and right, but not up and down
4. Double-click the image and scale it up a bit
5. Open the animation dropdown
6. Verify that the element animation can be set to pan in all directions

#### Media scale
1. Add an image or video to a page
2. Double click the element and verify that the element can be scaled between 100% to 400%.

### Paste text
1. Select some text anywhere on you computer and copy it (the raw text, not a file or anything)
2. Click anywhere outside the canvas in the editor and make sure nothing is selected
3. Paste using either keyboard shortcut or browser edit menu
4. Observe that the copied text is inserted correctly as a new text element with default style

## Checklist

---

Fixes #12332


packages/elements/src/types.js
```diff
@@ -63,7 +62,9 @@ StoryPropTypes.box = PropTypes.exact({
 
 StoryPropTypes.page = PropTypes.shape({
   id: PropTypes.string.isRequired,
-  animations: PropTypes.arrayOf(PropTypes.shape(AnimationProps)),
+  // Temporary solution for animations. Better would be to move this
+  // prop type to the types package, but really?
+  animations: PropTypes.arrayOf(PropTypes.object),
```
barklund, line 67: As noted, this isn't great, but okay until we convert this package to TypeScript.

packages/elements/src/utils/getElementOffsets.js
```diff
@@ -42,17 +42,17 @@ import { getMediaSizePositionProps } from '@googleforcreators/media';
 
 const PRECISION = 1;
 
-export function getMediaBoundOffsets({ element }) {
+export function getElementOffsets(element) {
```
barklund, line 45: So, I know that this function only makes sense for media elements (that have scale and focal values), but I've made it general anyway (just returning all zeros for non-media).

packages/story-editor/src/components/panels/design/animation/animation.js
```diff
@@ -191,12 +195,17 @@ function AnimationPanel({
...
-      const normalizedScale = progress(selectedElements[0]?.scale || 0, {
+    if (!selectedElement) {
+      return {};
+    }
+    if (selectedElement.isBackground) {
+      const { isMedia } = getDefinitionForType(selectedElement.type);
+      const hasOffset = isMedia
+        ? getHasElementOffsets(selectedElement)
+        : DEFAULT_HAS_OFFSETS;
+      const normalizedScale = progress(selectedElement.scale || 0, {
```
barklund, line 208: Not really sure if the `isMedia` guard is even needed here. The function will work on background shapes just fine.

packages/story-editor/src/components/videoTrim/generateVideoStrip.js
```diff
@@ -17,31 +17,13 @@
...
-
-const CACHE: Map<string, string> = new Map();
-
-interface FlipSettings {
-  vertical: boolean;
-  horizontal: boolean;
-}
-
-// TODO(#12126): Use improved Element type from shared package.
-interface Element {
```
swissspidy, line 37: Kinda sad having to remove the types again 😢

packages/story-editor/src/components/canvas/editElement.js
```diff
@@ -104,6 +105,8 @@ const EditElement = memo(
           deleteSelectedElements={deleteSelectedElements}
           maybeEnqueueFontStyle={maybeEnqueueFontStyle}
           zIndexCanvas={Z_INDEX_CANVAS}
+          min={BG_MIN_SCALE}
+          max={BG_MAX_SCALE}
```
swissspidy: Maybe rename the props to `scaleMin` and `scaleMax`?

packages/elements/src/types.js
swissspidy: > Better would be to move this prop type to the types package

Is this referring to just the animation prop type? Or all of `StoryPropTypes` in general?

Technically it would make sense for all of the `StoryPropTypes.*` prop types to live next to their respective TS counterparts, e.g. `StoryPropTypes.page` in the `story` package, `StoryElementPropTypes` in `elements`, and so on.

But yeah I guess with the conversion to TypeScript the prop types usage will slowly fade away anyway (unless we see a lot of value in keeping it around)

packages/story-editor/src/components/videoTrim/generateVideoStrip.js
barklund: Yes, that did hurt a bit. But they weren't that precise anyway 🤷. In the future we could move this to a `story-editor-util` library, that only the story editor depends on. It could have a ton of functionality that we'd want to typify but can't otherwise move too far out of the main package.

packages/story-editor/src/components/canvas/editElement.js
barklund: Yes, that's probably more reasonable.

barklund: Renamed in 0885cc7255.

packages/story-editor/src/components/videoTrim/generateVideoStrip.js
swissspidy: It should be possible to just use TypeScript for individual files in `story-editor`, no need to create a separate package for that. Just need to adjust `"include": ["src/**/*"]` in `tsconfig.json` to list individual files. Great for migrating step by step. Or is there another benefit of having such a separate package?

barklund: Oh, that sounds awesome. I thought it was an _all-or-nothing_ deal. Let's explore that somewhere closer to the end of these migrations 😄

packages/story-editor/src/components/panels/design/animation/animation.js
swissspidy: Feels like leaving it helps better understand the intent here

packages/elements/src/types.js
barklund: It's just this single prop type, because the dependency between the animation and the elements package have been reversed. Yes, it would make sense to move them all to the types package, I'll look at that in #12126.

merapi (approved):
LG, I saw no regressions.

kkalarickal:
Thank you for the detailed areas that need testing @barklund .

**Video strip generation**

Test web-story: https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/post.php?post=19878&action=edit#page=%25225f845a7e-3e7b-4d18-8d75-2d6fb062f67a%2522

Confirmed that for videos that have enough frame information, it is possible to hit the :scissors icon and trim the video. It is also possible to re-trim back from the original source on subsequent attempts.



Test case permutations:
- different formats of video
- uploaded, recorded, linked source of videos
- video from 3rd party media tab
- adding video as an element, background, within another shape
- animation + video that is trimmed
- video with non-default style (mirrored, borders, turned, tinted, etc)

Negative test cases:
- uploading videos with error
- reloading the editor while trimming
- testing on multiple browsers/os (Safari, Opera, Edge, Firefox)

kkalarickal:
**Paste text**

Confirmed that copying text from another source and pasting into the canvas creates a text element that has default font/styling : left justified, Roboto font, Regulat Size 18 with 1.2 line spacing black & white.

Test Case Variations:
- Tried with arabic (right-to-left) and other international characters (accepts and symbols) in the copied text.
- Copy and paste text that was added to the canvas once again
- Copy & paste into an existing text element

kkalarickal:
**Media scale**

Confirmed scaling for images and video elements from 100% to 400%

Test Web Story: https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/post.php?post=19942&action=edit#page=%2522ee551669-8fb8-4983-8fc6-a5d6836ba7f7%2522

Test Case Variations:
- tried with different image formats
- tried with different video formats
- tried adding media into a boundin shape
- tried with multiple styles applied to the media asset

kkalarickal:
**Media Pan Animation**

Confirmed that without scaling  that the image is too small to allow PAN up or down. But side PAN left and right are supported.



Confirmed that after scaling up from 100%, immediately the PAN up and down are available depending on how the scaling was done.



Test Case Permutations:
- tried image and video as background media
- confirmed that other animations are also working
- verified with different media file formats

kkalarickal:
@barklund -- here is the set of inconsistencies that I wanted to bring to your attention:

_Video strip generation_

1. the trimming did not work when I tried it on Opera/Windows10


_Paste text_
2. I copied and pasted text from within the web story which had a different font/styling.  However, the pasted text was still the default. As  an editor, I was expecting the fonts/styles to carry over from the text that was cut from within the story itself (not external).

_Media Pan Animation_
3. I got an error while trying the following test scenario:

- set an image as background
- without changing scale set the animation to PAN left
- in the LHS animations settings you will see the duration/delay values and four arrows. The left arrow will be highlighted and the right arrow is not. The top and bottom arrows are not clickable
- now click the right arrow

Issue: The animation becomes None and there is a console error 

Expected: I was expecting the PAN left to become PAN right when I clicked the right arrow. Instead I get a console error and the animation is removed.

4. Another issue with Media Pan Animate is the following scenarion:
- insert a media asset and scale it
- now you can set the PAN up or down
- the LHS settings will show duration/delay and the arrow heads highlighted appropriately
- now unscale the asset
- Issue is that the animation still shows PAN down/up even though the arrows are greyed out. This can be confusing

barklund:
> @barklund -- here is the set of inconsistencies that I wanted to bring to your attention:

That @kkalarickal for the thorough report. All of the 4 highlighted issues also occur on main, so these are not regressions (though still bugs). I will file them independently as such and merge this PR.

PR #12346, opened 2022-09-21
Dashboard: Fix template insertion

### LINKED ISSUE #12024: E2E Tests: Fix flakey Explore Templates test
## Feature Description

Test is failing looking for `Add title`

```
FAIL packages/e2e-tests/src/specs/dashboard/templates/useTemplate.js (30.295 s)
  ● Template › should be able to use existing template for new story

    TimeoutError: waiting for selector `input[placeholder="Add title"]` failed: timeout 3000ms exceeded

      65 |
      66 |     // Wait for title input to load before continuing.
    > 67 |     await page.waitForSelector('input[placeholder="Add title"]');
         |                ^
      68 |
      69 |     // Wait for skeleton thumbnails in the carousel to render which gives footer time to also render
      70 |     await page.waitForFunction(
```

### LINKED ISSUE #12345: Dashboard: Template insertion broken
## Bug Description

PR #12237 removed the elements registration from the dashboard app because it was deemed unused and thus redundant.

Turns out the elements are in fact needed when using the "Create story from template" functionality, because there we generate the AMP HTML output (via the `output` package) which in turn requires elements to be registered.

While we would have had an e2e test for this scenario that should have caught this right away, it has been disabled because of flakiness (see #12024)

The quick solution here is to revert #12237, the best solution would be to not use the `output` package at all since we don't actually need it.

## Expected Behaviour

Creating a new story from a template should be possible again.

## Steps to Reproduce

1. Go to "Explore Templates"
2. Try to create a new story from a template.

## Additional Context

- Plugin Version: 1.25.0
- WordPress Version: 6.0.2
- Operating System: macOS
- Browser: Chrome

swissspidy:
## Context

#12237 accidentally broke the "Use template" functionality on the dashboard as it turns out the element types were in fact used after all. It wasn't caught because of a missing (disabled) e2e test.

But now no more!

## Summary

Instead of just reverting #12237, this PR removes usage of the `output` package from the dashboard to fix the template insertion functionality.

## Relevant Technical Choices

Fixes and re-enables the relevant e2e test.

## User-facing changes

No visual changes, just a fixed user flow

## Testing Instructions

1. Go to "Explore Templates"
2. Create a new story from a template
3. It should succeed

No

No

No

## Checklist

---

Fixes #12345 
Fixes #12024


includes/REST_API/Stories_Base_Controller.php
```diff
@@ -127,11 +127,23 @@ protected function prepare_item_for_database( $request ) {
 		if ( ! empty( $schema['properties']['content'] ) ) {
 
 			// Ensure that content and story_data are updated together.
```
swissspidy, line 129: Considering just removing this whole bit. Not sure if this has ever been useful.

miina: Do you mean the check + throwing the error if not updated together to allow updating separately?

packages/wp-dashboard/src/api/story.js
```diff
@@ -165,12 +163,7 @@ export const createStoryFromTemplate = async (config, template) => {
     title: '',
   };
 
-  const content = getStoryMarkup(story, pages, {
-    publisher: createdBy,
```
miina, line 169: Does this mean that we're not using the `createdBy` field of templates anymore?

miina (commented):
> But now no more!

Sounds like a radio ad 😄

includes/REST_API/Stories_Base_Controller.php
swissspidy: Yes

packages/wp-dashboard/src/api/story.js
swissspidy: We use it on the "Explore Templates" page to display who created the template.

timarney (approved):
Tested locally fixes the error currently on main

spacedmonkey (approved):

Issue #1391, opened 2020-04-30
Custom Calendar Styling


jauyong:
## Feature Description

This is a followup to #789 which is a short term solution for the Beta Release. 
Once a calendar is implemented we will need to style it according to the [designs](https://www.figma.com/file/Cd9iacVknHK62RO4pRwb68/Stories-Editor?node-id=6113%3A618)

---

_Do not alter or remove anything below. The following sections will be managed by moderators only._

## Acceptance Criteria
1. The calendar should match the [designs in Figma ](https://www.figma.com/file/bMhG3KyrJF8vIAODgmbeqT/Design-System?node-id=3408%3A85035)

miina:
Note: It would be good to confirm if we're going to reuse the WordPress calendar style again (as it is in Figma currently), or use something custom -- since we're going to use a custom calendar anyway now and not a copy of WP implementation.

swissspidy:
Pinging @samitron7 to confirm, but I don't think it makes sense to use WordPress styling anywhere in the editor.

samitron7:
Yeah, no WP styling but this is probably a low priority thing. Adding to visual design request.

samitron7:
Figma file is attached in description and [here](https://www.figma.com/file/bMhG3KyrJF8vIAODgmbeqT/Design-System?node-id=3408%3A85035) as well

bmattb:
@samitron7 after review the eng team requests that the other views to also be included:
1. Year view
2. Month view 
3. Century view

barklund:
Here are the 4 views of the current calendar component, that all need a design. Note that the first is designed at the three latter are very similar:

| | Month | Year | Decade | Century |
|-|-|-|-|-|
[...]

PR #12353, opened 2022-09-21
Date Picker: show timezone abbreviation if possible

### LINKED ISSUE #12234: Date Picker: show timezone abbreviation if possible
## Feature Description

In the WordPress block editor, the date picker shows the timezone abbreviation instead of the GMT offset if available:

Screenshot 2022-09-06 at 10 10 23

https://github.com/WordPress/wordpress-develop/blob/9c45c9067903d9a80100d67db984bf25e5825865/src/wp-includes/script-loader.php#L379-L386

https://github.com/WordPress/gutenberg/blob/43d9c82922619c1d1ff6b454f86f75c3157d3de6/packages/components/src/date-time/time/timezone.tsx

It would be nice to do that as well in the story editor.

Should be trivial to adapt the code accordingly.

spacedmonkey:
## Summary

Have a more human readable time zone in date picker. 

## User-facing changes

Screenshot 2022-09-21 at 16 40 39

## Testing Instructions

1. Go to Settings > General. 
2. Change time zone to berlin. 
3. Create a new story. 
4. Click on the documents tab. 
5. Click on publish date ( immediately ). 
6. See time zone. 
7. Hover over time zone. 

## Checklist

---

Fixes #12234


packages/story-editor/src/components/form/dateTime/timeZone.js
```diff
@@ -53,15 +53,15 @@ function TimeZone({ date }) {
...
   }
 
-  const { timeZone: timeZoneString } = getOptions();
-  const zoneAbbr = timezone?.length
-    ? format(date, 'T')
-    : `UTC${timeZoneString}`;
-
+  const offsetSymbol = Number( gmtOffset ) >= 0 ? '+' : '';
+  const zoneAbbr =
+    '' !== timezoneAbbr && isNaN( Number( timezoneAbbr ) )
```
swissspidy: Why the `isNaN( Number() )` check? 🤔

spacedmonkey: timezoneAbbr can be +04 or GMT. This check ensures that value is a string and not a int.

swissspidy: Got it. I see that we do the same thing also in `getTimeZoneString()` (which is the utility providing `timeZone` from `getOptions()` in the `date` package.

I recommend using `Number.isNaN` though instead of `isNaN`.

swissspidy: That said, this kinda repeats parts of `getTimeZoneString`. Would be nice if we could reuse that somehow.

For example, we could create a new `getTimeZoneAbbr` util in the date package that takes the abbreviation from the date settings

swissspidy (changes requested):
Lint error

spacedmonkey:
Screenshot 2022-09-23 at 14 30 23
 
If you set the time zone to christmas, then you get some weird styling issues. 

See gutenberg. 

Screenshot 2022-09-23 at 14 30 30

Any thoughts on this @swissspidy ?

swissspidy:
Perhaps easiest to write some Jest tests with a data provider so we can discuss expected behavior for a case like this.

swissspidy:
The output in the Gutenberg calendar looks more reasonable.

`Indian/Christmas` has no standard abbreviation like CEST OR IST, so we can show the UTC offset instead. And in the tooltip we show the offset + the full timezone. So like in the Gutenberg screenshot.

`UTC` should only be followed by a numeric offset, not by a timezone name. So no `UTCIndian/Christmas`.

spacedmonkey:
@swissspidy The logic is now exactly the same as gutenberg. In my testing it seems to work well and is inline with what users would expect, if they have used the gutenberg date picker.

swissspidy (approved):

kkalarickal:
@spacedmonkey - i found one place where the new time zone information is only partially visible: within the Publish / Story Details popup that you get when user first publishes the story:



Is this something that we should fix ?

kkalarickal:
Verified via [stories-qa test story](https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/post.php?post=19985&action=edit#page=%25221efea0dd-fe6e-4604-85e9-1286ce7b537c%2522):

- updated timezone setting within [wp-admin > General](https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/options-general.php)
- observed the timezone information in the Publish time picker

Here are a few of the timezones that I tried:
- [CEST/Berlin](url)
- [Indian/Anantarivo](url)
- [Antarctica/Dumont](url)

swissspidy:
> i found one place where the new time zone information is only partially visible: within the Publish / Story Details popup that you get when user first publishes the story:
> Is this something that we should fix ?

The tooltip is not newly added in this PR. We can deal with this in #1391

PR #12364, opened 2022-09-22
Lightbox: ensure story player always has dimensions

### LINKED ISSUE #12363: Block: Story player not immediately loading in lightbox
## Bug Description

Original report: https://github.com/ampproject/amphtml/issues/35959

Stories being opened in a lightbox (using amp-story-player) are not immediately visible until some user interaction (i.e. press esc or click on the element).

We've finally figured out that the player is not properly loading due to not having dimensions because of the `display: none` on the ligthbox wrapper.

## Expected Behaviour

When opening a story in a lightbox they should always be immediately visible.

## Steps to Reproduce

See https://github.com/ampproject/amphtml/issues/35959

## Screenshots

See https://github.com/ampproject/amphtml/issues/35959

## Additional Context

- Plugin Version: 1.25.0
- WordPress Version: 6.0
- Operating System: macOS
- Browser: Chrome

swissspidy:
## Context

See #12363

## Summary

The solution for this story player bug is to ensure that it always has dimensions

## User-facing changes

Player in lightbox should load right away

## Testing Instructions

1. Add new Web Stories carousel block to site, ideally far down in the footer
2. View website
3. Click on stories in carousel
4. Verify that the player loads right away

No

No

No

## Checklist

---

Fixes #12363

### DIFF
--- a/packages/stories-block/src/css/lightbox.css
+++ b/packages/stories-block/src/css/lightbox.css
@@ -15,7 +15,6 @@
   height: 100%;
   opacity: 0;
   transform: translate(0, -100vh);
-  display: none;
 }
 
 /* amp-lightbox needs to have z-index to render on top of other elements in the page. */
@@ -27,7 +26,6 @@
 .web-stories-list__lightbox.show {
   transform: translate(0, 0);
   opacity: 1;
-  display: block;
 }
 
 .web-stories-list__lightbox amp-story-player {


spacedmonkey (approved):
Seems to work well in my testing.

kkalarickal:
@swissspidy - I could NOT recreate the issue via the external web story  https://cai2r.net/a-new-way-to-image-myelin/ story since they seem to have put in the work around you suggested.

So I tried to create the problem within https://stories-qa-wordpress-amp.pantheonsite.io/lightbox-ensure-story-player-always-has-dimensions-12364/ by adding a web story carousel into a wordpress post.

When the code base was `main branch`, there was a `display: none` directive associated with the class ` .web-stories-list__lightbox `.



When the code base was changed to this PR, the display directive had changed to `block`



This proves to me that the display directive is properly set with this change.

I also tried adding different types of carousels to the wp post.

Question: Is this test good enough for this ticket ?

If you could point me to how to recreate the original problem in the stories-qa environment, then it would be possible to confirm the front end effects of this css fix.

swissspidy:
No worries. A couple of users have now used this workaround successfully, so this should be fine.

PR #12357, opened 2022-09-22
Code Quality: Remove `improvedAutosaves` feature flag code

timarney:
## Context

Now that this has been enabled by default in #12272  in the last release we can remove the relevant code bits.

## Summary

Removed `improved_autosaves` feature flag

## Testing Instructions

- [ ] This is a non-user-facing change and requires no QA

This PR can be tested by following these steps:

1. Ensure Save and AutoSave functionality still works as expected 

### Does this PR have a security-related impact?

No

### Does this PR change what data or activity we track or use?

No

### Does this PR have a legal-related impact?

No

## Checklist

- [x] This PR addresses an existing issue and I have linked this PR to it in ZenHub
- [x] I have tested this code to the best of my abilities
- [ ] I have verified accessibility to the best of my abilities ([docs](https://github.com/googleforcreators/web-stories-wp/blob/main/docs/accessibility-testing.md))
- [ ] I have verified i18n and l10n (translation, right-to-left layout) to the best of my abilities
- [ ] This code is covered by automated tests (unit, integration, and/or e2e) to verify it works as intended ([docs](https://github.com/googleforcreators/web-stories-wp/tree/main/docs#testing))
- [ ] I have added documentation where necessary
- [x] I have added a matching `Type: XYZ` label to the PR

---

Fixes #12307

PR #12373, opened 2022-09-26
Story Locking: Autosave on takeover

### LINKED ISSUE #12325: Story Locking: Autosave on takeover
When editing is taken over from you, we want to autosave the story before displaying the dialog, so that your changes are not lost.

Compare to how Gutenberg is doing that here:

https://github.com/WordPress/gutenberg/blob/61e41f0a222544a9c24c5412573cf641f7c48433/packages/editor/src/components/post-locked-modal/index.js#L86-L114

spacedmonkey:
## Summary

Trigger an autosave when user has their story taken over.

## Relevant Technical Choices

 Only trigger if they are changes in the story. Otherwise, it will trigger a REST API call that returns a 400 error. 

## Testing Instructions

0. Enable autosaving and take over post locking experiment.
1. Create a story. 
2. Add some content. 
3. Save as draft. 
4. Open another browser. 
5. Login, as a differnet user. 
6. Open story ( from previous ) in editor. 
7. Click Take over story. 
8. In first browser, make changes to story. DO NOT SAVE. 
9. Wait 1 minute 30 seconds. 
10. Go back to second browser. 
11. Refresh browser. 
12. See message that there is a newer autosave. 

## Checklist

---

Fixes #12325


spacedmonkey:
Should I remove references to `improvedAutosaves`, considering #12357. CC @timarney

swissspidy (approved):
Change looks OK but eyes from Prometheus pod would be great

timarney:
> Should I remove references to improvedAutosaves, considering https://github.com/GoogleForCreators/web-stories-wp/pull/12357. CC @timarney

Re-tested my other PR further last night --- seems to be working as expected.

packages/wp-story-editor/src/components/postLock/postLock.js
```diff
@@ -160,6 +167,27 @@ function PostLock() {
...
   }, [postLockInterval, currentUserLoaded]);
 
+  useEffect(() => {
+    if (
+      enablePostLockingTakeOver &&
+      improvedAutosaves &&
+      showLockedDialog &&
+      user &&
+      hasNewChanges &&
+      !isFirstTime
```
miina, line 176: Curious why only when `!isFirstTime`?

miina (approved):

packages/wp-story-editor/src/components/postLock/postLock.js
spacedmonkey: The take over message should show until this is not the first time.

swissspidy: To Miina's point, the variable wording is not the most intuitive, but at this point I guess it is what it is :-)

swissspidy:
#12357 is merged now, so this can be updated here

spacedmonkey:
> #12357 is merged now, so this can be updated here

Done in https://github.com/GoogleForCreators/web-stories-wp/pull/12373/commits/6ceb6933b53f2163621a63f28fa559cc3cbc22ac

timarney (approved):

kkalarickal:
Tested this feature by logging in as an admin user and an editor user on two different  browsers , and within the same browser, via two different sessions.

Confirmed that after take over by either admin or editor user, when the previous session is updated, but NOT saved, there is an auto-draft that is made. This is reflected in the other browser/session upon refresh so that changes made in the taken-over editor now reflect on the new owner of the editor session.



Test Variations:
- use Chrome-Firefox, Chrome-Edge, Chrome-Safari as the old and new sessions
- use Chrome on win10-OSX

kkalarickal:
@spacedmonkey - wanted to run some scenarios by you to confirm this is expected:

Scenario 1: 

When Admin user creates a story, but its taken over by Editor user, there is a notification sent to Admin user saying someone else has taken over.

But the Admin user is able to click outside that notification and continue editing the story. After some time, if the Admin user now reloads the editor, there are two overlays one-on-top-of-the-other



- one is a backup notification
- another is the take over notification

Isn't it confusing that the user has to deal with two notifications one on top of the other ?

Cc: @swissspidy

kkalarickal:
Scenario 2:

After Editor user takes over a story Admin is editing, because the Admin can click outside the notification about someone taking over the story, and she can continue to edit her version of the story.

Later, if the Editor refreshes the open story that was taken over, the story is updated with the changes that the Admin made after takeover !

There seems to be a delay between when the story take over notice is sent to previous editor sessions, and how the local edits are auto-saved. As a result, both parties can to continue editing by skipping over the take over notification, leading to auto-saved edits being transferred between the two open editor instances.

swissspidy:
> But the Admin user is able to click outside that notification and continue editing the story.

This should not happen.

> After some time, if the Admin user now reloads the editor, there are two overlays one-on-top-of-the-other

This should not happen.

Both would require new tickets to be opened.

----

@kkalarickal Could you test the same scenarios in the WordPress post editor so that we have something to compare against?

kkalarickal:
@swissspidy - i created #12430 to track the issue with "user is able to click outside that notification and continue editing the story". This ticket also has the double overlay in it. I was not able to re-create the double overlay consistently, but thought I would include it in the ticket.

PR #12405, opened 2022-09-30
Editor: fix incorrect poster dimensions being returned

### LINKED ISSUE #12354: Editor: incorrect poster dimensions being returned
## Bug Description

Lucky found this obscure bug where the checklist in the editor shows a warning about incorrect poster dimensions, even though the poster definitely has the right size.

I tracked this down to some unexpected behavior in WordPress that boils down to honoring `$content_width`.

This only happens in the admin and not when accessing the REST API in an other way.

Here's how it works:

1. In the editor we preload REST API responses, such as the one for the current story (which is handled by `Stories_Controller`)
2. `\Google\Web_Stories\REST_API\Stories_Controller::get_story_poster()` calls `wp_get_attachment_image_src()`
3. `wp_get_attachment_image_src()` in turn calls `image_downsize`
4. `image_downsize` in turn calls `image_constrain_size_for_editor`
5. `image_constrain_size_for_editor` then reduces the width and height because we're in the admin (`is_admin()` is true because of the REST API preloading)

To work around this, I think we have to call `image_get_intermediate_size()` to get the actual size ourselves, and if that returns `false` (because the size does not exist), get the original size from  `wp_get_attachment_metadata()`.

But there might be other options.

**Notes:**

* This same issue theoretically exists in `\Google\Web_Stories\Model\Story::load_from_post()`
* We could just use the `Story` model in `Stories_Controller`, so we can delete the redundant `get_story_poster` method there

**Aside:**

I find it unexpected that `wp_get_attachment_image_src()` honors `$content_width` like this. Maybe worth opening a Trac ticket?

## Expected Behaviour

There should be no incorrect warning.

## Steps to Reproduce

1. Activate Twenty Twenty theme
7. Create a new story and add a few pages
8. Add a poster image by uploading a new image that needs to be cropped & crop it
9. Save the story
10. Reload the page
11. Notice the checklist warning about incorrect poster dimensions

## Screenshots

Screenshot 2022-09-21 at 17 41 01

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

spacedmonkey:
## Summary

Fix incorrect image sizes in editor by filtering `editor_max_image_size` only in our editor. 

## Relevant Technical Choices

Copy the logic `image_constrain_size_for_editor`and using same logic without content_width checks. 

## Testing Instructions

1.

## Checklist

---

Fixes #12354

### DIFF
--- a/includes/Admin/Editor.php
+++ b/includes/Admin/Editor.php
@@ -486,41 +486,80 @@ class Editor extends Service_Base implements HasRequirements {
 	 *
 	 * @since 1.26.0
 	 *
-	 * @param int[] $max_size {
-	 *     An array of width and height values.
+	 * @param int[]        $max_size {
+	 *            An array of width and height values.
 	 *
 	 *     @type int $0 The maximum width in pixels.
 	 *     @type int $1 The maximum height in pixels.
 	 * }
 	 * @param string|int[] $size     Requested image size. Can be any registered image size name, or
 	 *                               an array of width and height values in pixels (in that order).
-	 *
-	 * @return array
+	 * @return array<int, int>
 	 */
-	public function editor_max_image_size( $max_size, $size ) {
+	public function editor_max_image_size( $max_size, $size ): array {
 		$_wp_additional_image_sizes = wp_get_additional_image_sizes();
 
-		if ( is_array( $size ) ) {
+		if ( \is_array( $size ) ) {
 			$max_width  = $size[0];
 			$max_height = $size[1];
 		} elseif ( 'thumb' === $size || 'thumbnail' === $size ) {
-			$max_width  = (int) get_option( 'thumbnail_size_w' );
-			$max_height = (int) get_option( 'thumbnail_size_h' );
+			/**
+			 * Thumbnail size width.
+			 *
+			 * @var int $max_width
+			 */
+			$max_width = get_option( 'thumbnail_size_w' );
+			/**
+			 * Thumbnail size height.
+			 *
+			 * @var int $max_height
+			 */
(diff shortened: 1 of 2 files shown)


includes/Admin/Editor.php
```diff
@@ -479,4 +480,93 @@ protected function setup_lock( int $story_id ): void {
 			wp_set_post_lock( $story_id );
 		}
 	}
+
+	/**
+	 * Filter the editor_max_image_size.
```
swissspidy, line 485: The doc block here absolutely should mention that this is an exact copy of the core function with certain tweaks.

includes/Admin/Editor.php
```diff
@@ -232,6 +232,7 @@ public function replace_editor( $replace, $post ) {
 			// Registering here because the script handle is required for wp_add_inline_script in edit-story.php.
 			$this->assets->register_script_asset( self::SCRIPT_HANDLE, $script_dependencies, false );
 
+			add_filter( 'editor_max_image_size', [ $this, 'editor_max_image_size' ], 99, 2 );
```
swissspidy, line 235: The filter should ideally also be removed again later on

swissspidy:
Not really a fun of having to copy all logic from core :/

Is there perhaps another way we could solve this? For example by modifying the `$content_width` global instead? Modifying globals is usually not great, but it's all WP offers and it's still better than maintaining such a copy.

Also, would be great if we could have some integration test that verifies that this fixes the actual bug with the incorrect poster size in the preloaded request. The current test just covers this copied function itself.

spacedmonkey:
Closing this PR in favour of #12418 as it is a cleaner solution.

PR #12406, opened 2022-09-30
Run local Karma tests in batches

merapi:
## Summary

AFAIK you cannot run Karma tests on selected files via CLI params (you can just select tests), so I used our custom sharding strategy for that. Another option is to move/rename files temporarily in batches.

image

## Testing Instructions

1.

## Checklist

---

Fixes #


package.json
```diff
@@ -99,6 +99,7 @@
     "babel-plugin-transform-react-remove-prop-types": "^0.4.24",
     "browserslist": "^4.21.3",
     "circular-dependency-plugin": "^5.2.2",
+    "chalk": "*",
```
merapi, line 102: `chalk` is already used by other packages, so we need any version that already exists.

PR #12413, opened 2022-10-03
Story Locking: Enable by default

### LINKED ISSUE #12326: Story Locking: Enable by default
Enable `enablePostLockingTakeOver` feature by default.

timarney:
## Context

Enable enablePostLockingTakeOver feature by default.

## Summary

Sets feature flag to default

## Testing Instructions

Test instructions via https://github.com/GoogleForCreators/web-stories-wp/pull/12373 

1. Create a story.
2. Add some content.
3. Save as draft.
4. Open another browser.
5. Login, as a differnet user.
6. Open story ( from previous ) in editor.
7. Click Take over story.
8. In first browser, make changes to story. DO NOT SAVE.
9. Wait 1 minute 30 seconds.
10. Go back to second browser.
11. Refresh browser.
12. See message that there is a newer autosave.

No

No

No

## Checklist

---

Fixes #12326


spacedmonkey (approved):

kkalarickal:
@timarney @swissspidy - I tested this PR by comparing the Stories > Experiments page, on `main branch` and again on `#12413 branch`

With this change, the "Story locking takeover" flag is greyed out and ticked by default.

Next I tested the autosaving feature within the new PR: 

NOTE: I encountered the Uncaught Type Error message on console within the `12413 branch`

This is the scenario from the QA instructions above:

1. User A creates a new story and makes some edits without saving
2. An `wp_stories_autosave_story_auto-draft` is created in session storage with a key containing the unsaved story json
3. User A saves draft - the autosave entry in session storage is deleted
4. Next User B logs in to another browser and opens the same story as in step 1.
5. There is a take over notification given to User B
6. Upon taking over the story _it is possible for both user A an user B to edit the same story albeit in their own local session_
7. After 1 min 30 sec, user A is given the message that "someone else has taken over the story. But don't worry changes up to this moment have been saved"
8. If user B (in second browser) does a refresh of the editor, the changes from user A are visible within User B's session along with a message

So as per acceptance criteria, this ticket passes QA.

@swissspidy - However, I also tested the scenario where in step 6 the user B makes some changes in return and there are two different autosaves present in each browsers session storage. This yields a overlapping set of overlays for user A that is very confusing

It seems to me that the window of opportunity where both sessions can be edited simultaneously is different from the workflow for the WP Post, where the momment user B takes over, user A can no longer edit.

swissspidy:
> It seems to me that the window of opportunity where both sessions can be edited simultaneously is different from the workflow for the WP Post, where the momment user B takes over, user A can no longer edit.

I think we're tracking this in #12435

> However, I also tested the scenario where in step 6 the user B makes some changes in return and there are two different autosaves present in each browsers session storage. This yields a overlapping set of overlays for user A that is very confusing

If we fix the above, this should no longer be possible, so we can track this in the same ticket.

PR #12418, opened 2022-10-03
Story Poster: Use dimensions array instead of size name

### LINKED ISSUE #12354: Editor: incorrect poster dimensions being returned

spacedmonkey:
## Summary
Another solution to #12354. If you pass an array of dimensions instead of passing size name it bypasses the content width check. [See](https://github.com/WordPress/wordpress-develop/blob/5eddc8f18c97d2eb909d0f70f4e20732398a7bda/src/wp-includes/media.php#L69-L71)

## Testing Instructions

0. Activate Twenty Twenty theme
0. Create a new story and add a few pages
0. Add a poster image by uploading a new image that needs to be cropped & crop it
0. Save the story
0. Reload the page
0. View checklist, see no incorrect poster dimensions.

## Checklist

---

Fixes #12354


swissspidy:
Does it work?

As discussed, would be nice to have some sorts of tests verifying that this actually solves the reported bug. With this approach here it should be doable to that with PHP tests alone, no E2E tests needed (though that would be the last resort).

Alternatively, if we can't have tests, at the very least we should add comments explaining why we do it that way.

spacedmonkey:
@swissspidy Unit tests added. Closing other PR in favour of this.

tests/phpunit/integration/tests/REST_API/Stories_Controller.php
```diff
@@ -533,22 +533,71 @@ public function test_get_item_with_featured_image(): void {
...
 		$this->assertArrayHasKey( 'story_poster', $data );
+		$this->assertSame( Image_Sizes::POSTER_PORTRAIT_IMAGE_DIMENSIONS[0], $attachment_src[1] );
+		$this->assertSame( Image_Sizes::POSTER_PORTRAIT_IMAGE_DIMENSIONS[1], $attachment_src[2] );
 		$this->assertEqualSetsWithIndex(
 			[
 				'id'         => $attachment_id,
 				'url'        => $attachment_src[0],
-				'height'     => $attachment_src[1],
-				'width'      => $attachment_src[2],
+				'height'     => $attachment_src[2],
```
timarney: Nit: can we flip these so it's width before height 
'width'     => $attachment_src[1],
'height'     => $attachment_src[2],

Little better readability IMO

tests/phpunit/integration/tests/REST_API/Stories_Controller.php
```diff
@@ -533,22 +533,71 @@ public function test_get_item_with_featured_image(): void {
...
+		$data     = $response->get_data();
+
+		$this->assertArrayHasKey( 'story_poster', $data );
+		$this->assertSame( Image_Sizes::POSTER_PORTRAIT_IMAGE_DIMENSIONS[0], $attachment_src[1] );
+		$this->assertSame( Image_Sizes::POSTER_PORTRAIT_IMAGE_DIMENSIONS[1], $attachment_src[2] );
+		$this->assertEqualSetsWithIndex(
+			[
+				'id'         => $attachment_id,
+				'url'        => $attachment_src[0],
+				'height'     => $attachment_src[2],
```
timarney: Same Nit as above 

```
'width' => $attachment_src[1],
'height' => $attachment_src[2],
```

Would be more readable

tests/phpunit/integration/tests/REST_API/Stories_Controller.php
```diff
@@ -645,8 +695,8 @@ public function test_get_item_with_featured_image_and_hotlinked_poster(): void {
 			[
 				'id'         => $attachment_id,
 				'url'        => $attachment_src[0],
-				'height'     => $attachment_src[1],
-				'width'      => $attachment_src[2],
+				'height'     => $attachment_src[2],
```
timarney: Same again here:

```
'width' => $attachment_src[1],
'height' => $attachment_src[2],
```

swissspidy:
The tests don't work, as in, they demonstrate the bugfix. If I undo the changes to `includes/REST_API/Stories_Controller.php` and `includes/Model/Story.php` they still pass.

----

Note that for the QA instructions you can reuse the reproduction steps from the issue.

spacedmonkey:
> The tests don't work, as in, they demonstrate the bugfix. If I undo the changes to includes/REST_API/Stories_Controller.php and includes/Model/Story.php they still pass.

The fix does work, I just used the wrong test data. I have updated the fixture data and tests pass and you can see them pass.

swissspidy:
When I undo the changes to `includes/REST_API/Stories_Controller.php` I expect the `Stories_Controller` tests to fail, but they don't.

swissspidy:
Also, running the tests creates a bunch of untracked files in the tests directory:

Screenshot 2022-10-13 at 08 38 11

swissspidy (changes requested):
A couple of adjustments needed, as per my recent comments

swissspidy:
QA instructions would be needed too here

spacedmonkey:
> Also, running the tests creates a bunch of untracked files in the tests directory:
> 
> Screenshot 2022-10-13 at 08 38 11

I added delete attachment calls to all places where image sizes are generated.

swissspidy:
> When I undo the changes to `includes/REST_API/Stories_Controller.php` I expect the `Stories_Controller` tests to fail, but they don't.

This still needs to be addressed AFAICT. After that this should be ready.

spacedmonkey:
> This still needs to be addressed AFAICT. After that this should be ready.

I believe I fixed these issues with new test data.

swissspidy:
Just verified that it's still an issue.

spacedmonkey:
> Just verified that it's still an issue.
> 
> [3be82da](https://github.com/GoogleForCreators/web-stories-wp/commit/3be82daae767ee8bfb53c4776a33942d4435f4b0) was the only change since I flagged this.

Fix the unit test here. https://github.com/GoogleForCreators/web-stories-wp/pull/12418/commits/21de72a614073c7fd5a4c82ee3406c7b43602931

swissspidy:
Excellent! 🎉

swissspidy (approved):

kkalarickal:
Reproduced this problem (thanks to @LuckynaSan and @spacedmonkey for your help) by following these steps:

1. Switch to an older theme (I tried Twenty-Twenty, Twenty-Nineteen, Twenty-8teen, Twenty-Seventeen)
2. Create a sample web story and save to draft
3. Hit publish and update the poster image by selecting a larger image and cropping it
4. Exit from editor and reopen the same published story -- Now you will see the incorrect checklist error saying poster image size is not up to expected 640x853. This is in spite of the image being cropped properly.

Verified the fix by loading the `12418 branch` in the same wp environment with the older theme and repeating the steps above to confirm that the checklist does NOT report a non-existent problem.

PR #12431, opened 2022-10-05
Uninstall: Improve logic of uninstall code

### LINKED ISSUE #12361: Uninstall: Improve logic of uninstall code
## Feature Description

Create a new interface, called PluginUninstallAware. Similar to PluginActivationAware. In Uninstall.php, just loop around all the registered services that are `PluginUninstallAware . This makes testing unistall methods easier and make easier to spot when post type / taxonomy /  meta does not. have a uninstall method.

spacedmonkey:
## Summary

Move the logic to delete data on uninstall to service classes. On uninstall, simply run the on_plugin_uninstall method on uninstall plugin. 

## Testing Instructions

1. Create a new site, for example on [Local](https://localwp.com/) environment or a service like tastewp.
2. Install built version of this plugin (see comment below for downloading production zip file). 
3. Activate plugin. 
5. On the settings page, toggle the option remove data on uninstall. 
6. Create some stories. 
7. Upload some media. 
8. Create web stories categories. 
9. Uninstall plugin. 
10. Reinstall plugin. 
11. See all data has been removed. 

## Checklist

---

Fixes #12361


includes/Settings.php
```diff
@@ -383,4 +383,29 @@ public function get_setting( string $key, $default = false ) {
 	public function update_setting( string $key, $value ) {
 		return update_option( $key, $value );
 	}
+
+	/**
+	 * Act on plugin uninstall.
+	 *
+	 * @since 1.26.0
+	 */
+	public function on_plugin_uninstall(): void {
```
swissspidy, line 395: Note that this class is not yet implementing `PluginUninstallAware`

spacedmonkey: Fixed.

includes/Uninstaller.php
```diff
@@ -0,0 +1,174 @@
...
+ */
+
+namespace Google\Web_Stories;
+
+use Google\Web_Stories\Infrastructure\PluginUninstallAware;
+
+/**
+ * Uninstaller class.
+ */
+class Uninstaller extends Plugin {
```
swissspidy: This class now does two very different things:

1. Instantiate services
2. Delete transients/optins on uninstall

Deleting options should be done in its own service class.

includes/Uninstaller.php
```diff
@@ -0,0 +1,174 @@
...
+	 * This takes into account whether the service registration needs to be delayed or not.
+	 *
+	 * @since 1.26.0
+	 *
+	 * @param string       $id    ID of the service to register.
+	 * @param class-string $class Class of the service to register.
+	 */
+	protected function schedule_potential_service_registration( $id, $class ): void {
+		$this->maybe_register_service( $id, $class );
+	}
```
swissspidy: So this is the part where we always register services even they're otherwise `Delayed`?

Would be good to update the doc block here to clarify the difference between this one and the parent method

uninstall.php
```diff
@@ -75,28 +75,7 @@
...
-
-		\Google\Web_Stories\delete_site();
-	}
-
-	restore_current_blog();
-} else {
-	\Google\Web_Stories\delete_site();
-}
-\Google\Web_Stories\delete_stories_user_meta();
+$uninstaller = \Google\Web_Stories\UninstallerFactory::create();
```
swissspidy: We no longer need the `\Google\Web_Stories\PluginFactory::create()->register();` call further above, right?

uninstall.php
```diff
@@ -75,28 +75,7 @@
...
+$uninstaller->register();
+$uninstaller->on_site_uninstall();
+$uninstaller->remove_options();
```
swissspidy: As per my other comment, removing options should be its own service. Then this call is not needed here.

swissspidy (commented):
Overall approach looks good

uninstall.php
spacedmonkey: No, I think we need that, so the settings class registers.

swissspidy: And what is registering the Settings class needed for?

Feels a bit weird to register the plugin here _and_ the uninstaller.

includes/Uninstaller.php
spacedmonkey: Move to it's own class now.

includes/Remove_Options.php
```diff
@@ -0,0 +1,137 @@
...
+		$options = $wpdb->get_col(
+			$wpdb->prepare(
+				"SELECT option_name FROM $wpdb->options WHERE option_name LIKE %s",
+				self::PREFIX
+			)
+		);
+
+		if ( ! empty( $options ) ) {
+			array_map( 'delete_option', (array) $options );
+		}
```
swissspidy: I wonder whether we still need this, because theoretically each service should already be deleting its own options, and they all have this prefix too.

uninstall.php
swissspidy: @spacedmonkey Would love to get an answer here before resolving.

Still not sure why we still need `\Google\Web_Stories\PluginFactory::create()->register();` in this file and why the `Settings` is needed here.

uninstall.php
```diff
@@ -75,28 +75,15 @@
...
+ *
+ * @since 1.26.0
+ */
+function web_stories_uninstall(): void {
+	$ws_plugin = \Google\Web_Stories\PluginFactory::create();
+	$ws_plugin->register();
+	$ws_plugin->on_site_uninstall();
 }
-\Google\Web_Stories\delete_stories_user_meta();
+add_action( 'shutdown', 'web_stories_uninstall' );
```
swissspidy, line 88: What's the purpose of this? Why only do the uninstall on shutdown?

timarney:
Overall code looks good groups things nicely for cleanup less chance keys will be meta keys will be forgotten.


Per above comment also wondering https://github.com/GoogleForCreators/web-stories-wp/pull/12431/files#r994958649

uninstall.php
spacedmonkey: @swissspidy This change was added in by @timarney in https://github.com/GoogleForCreators/web-stories-wp/commit/aa324c02b273e2726a424c8521d1db33bcfe7750. 

It is because of the default value of `$erase = (bool) get_option( \Google\Web_Stories\Settings::SETTING_NAME_DATA_REMOVAL );`

I did not add this change. But either way, it is needed now.

uninstall.php
spacedmonkey: > Defer running uninstall until every service is registered.

See comment above. Sadly I could not find a clean way to only register services without delays or register methods. The cleanest way was to just defer to the latest hook and run the uninstall methods there.

includes/Remove_Options.php
spacedmonkey: This would be used to clean up any transients still in the database. For the most part, it should not be needed, but just being overly careful and clear out options. Same with network options.

spacedmonkey: Good call out. I have remove options logic and now it only removes transient.

tests/phpunit/integration/tests/Remove_Transients.php
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * @runInSeparateProcess
+ * @preserveGlobalState disabled
+ * @coversDefaultClass \Google\Web_Stories\Remove_Transients
+ */
+class Remove_Transients extends DependencyInjectedTestCase {
+	protected static $attachment_ids;
+
+	protected static $user_id;
+
```
spacedmonkey:
```suggestion
```

uninstall.php
swissspidy: I don't think this was because of the default value. The default value is `false` anyway.

IIRC `\Google\Web_Stories\PluginFactory::create()->register();` was added because we needed to register the `Capabilities` service in order to be able to remove caps.

But now the registration is done by the `UninstallerFactory`. It doesn't really make sense to have both the `Plugin` _and_ the `Uninstaller` register the same services twice. For uninstall, only the uninstaller should be registering services.

To put this another way: does the uninstall break if you remove this line? I tend to say _no_.

uninstall.php
swissspidy: Deferring until shutdown seems like a hacky workaround to me though.

I tend to think that it must be possible to adjust the `Uninstaller` so that all services are registered straight away, ignoring `Delayed` and such.

uninstall.php
spacedmonkey: I have removed Uninstaller. A while ago. Please review again.

uninstall.php
spacedmonkey: Uninstaller was removed as I could not get it to work.

spacedmonkey: It is a hacky workaround for our setup. Uninstall is an uncommon use case. As it is rarely used, this workaround to bootstrap works surprisingly well and is only for this one place.

swissspidy: > Uninstaller was removed as I could not get it to work.

Hmm I see. Would have been nice. Too bad.

swissspidy (approved):

includes/Remove_Transients.php
```diff
@@ -0,0 +1,116 @@
+<?php
+/**
+ * Uninstaller class.
```
swissspidy: Wrong class name

includes/Remove_Transients.php
```diff
@@ -0,0 +1,116 @@
...
+ * limitations under the License.
+ */
+
+namespace Google\Web_Stories;
+
+use Google\Web_Stories\Infrastructure\PluginUninstallAware;
+use Google\Web_Stories\Infrastructure\Service;
+
+/**
+ * Uninstaller class.
```
swissspidy: Wrong class name

spacedmonkey: Fixed.

includes/Remove_Transients.php
spacedmonkey: Fixed.

kkalarickal:
Confirmed via setting up a test environment on https://dev-test-google-web-stories.pantheonsite.io/ that had the `12431 branch` code base

- It is possible to mark the checkbox in Stories > Settings  > Plugin Uninstall "Remove all data including stories and saved templates when uninstalling the Web Stories plugin" 

- Created a few categories/ tags and added assets in media library that were used in two or three stories (published, private, and draft modes)
- Took a back up of the database to check the size
- Deactivated and Deleted (uninstalled) the Google Web Stories Plugin
- Re-installed the latest google web stories plugin and confirmed that :
-- stories created before were no longer present in All Stories
-- the categories/tags created previously were removed
-- media assets from previous install continued to exists
- Took another backup of the database to compare the reduced size


QA Notes: I noticed that when the uninstall process happened, the following warnings were issued on the console:
Cc: @spacedmonkey @swissspidy

swissspidy:
@kkalarickal These warnings/errors are coming from a Chrome extension or something. It's impossible for these to be coming from the plugin because at the time of uninstallation the plugin is not even active anymore.

PR #12445, opened 2022-10-10
Data Structure: reduce redundancy when storing fonts 

### LINKED ISSUE #12261: Data Structure: reduce redundancy when storing fonts

timarney:
## Context

This is a follow-up to https://github.com/GoogleForCreators/web-stories-wp/pull/12131.

Right now when adding 10 text elements all with the same font, there will be 10 font objects being stored, all with the same information. That's very redundant.

Instead of this, consider storing all existing fonts at the story level instead of on each element to reduce redundancy.

Add data migration.

## Summary

This PR aim to pull all the font data off elements outside of font `family`

In order to do that font data is now stored at the story level.  When updating or inserting fonts data gets pushed to story level and cleaned up when saving to only use "in use" fonts.

In most cases outside of "output" font data is being pulled from the Font Provider --- using the font family to do a look up.  The data is typically passed via a new prop `elementFontData`.

## Testing Instructions

1.

## Checklist

---

Fixes #12261


packages/story-editor/src/app/story/useStoryReducer/reducers/utils.js
```diff
@@ -105,6 +112,26 @@ export function removeAnimationsWithElementIds(animations = [], ids = []) {
...
+  });
+  return draftFonts;
+}
+
+export function pickElementFontProperties(element) {
+  if (element?.font) {
+    const { upm, asc, des } = element.font.metrics;
+    element.font = {
+      family: element.font.family,
+      metrics: { upm, asc, des },
```
timarney:
```javascript
element.font = {
      family: element.font.family,
      metrics: { upm, asc, des },
```

Leave a minimal set of font metrics for now -- these are used by calcFontMetrics.  Later these should be pulled off as well.

packages/story-editor/src/app/story/useStoryReducer/reducers/updateElements.js
```diff
@@ -62,25 +67,41 @@ export const updateElements = (
...
 
   const isAnimationUpdate = Object.keys(animationLookup).length > 0;
 
   if (isAnimationUpdate) {
     page.animations = updateAnimations(page.animations || [], animationLookup);
   }
+
+  if (isFontUpdate) {
+    // store font props at the story level
+    draft.story.fonts = updateFonts(draft.story.fonts, page.elements);
```
timarney: Storing data here but it's not in use yet.

timarney:
@swissspidy 

Made a few more updates to see if I can just pull data from useFont get font getFontByName --- 

That works great for code that can access useFont.

That said --- we get into some utility functions which don't really have access to state outside of the `element` being passed in i.e. resize events etc... use selected elements.

https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/element-library/src/text/output.js#L60

https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/element-library/src/text/updateForResizeEvent.js#L35

<hr>

👉 👉 👉  I think we need to be able look at an element in any of the utility functions `calcFontMetrics` , `generateParagraphTextStyle` etc... and use the font family to pull data --- but not sure from where or how to access


<hr>

packages/story-editor/src/app/story/useStoryReducer/reducers/addElements.js
```diff
@@ -68,6 +68,14 @@ export const addElements = (draft, { elements }) => {
     addedIds.push(...nonProducts.map(({ id }) => id));
   }
 
+  // store font props at the story level
+  draft.story.fonts = updateFonts(draft.story?.fonts, page.elements);
+
+  // pull font props off the individual elements
+  page.elements = page.elements.map((element) =>
+    pickElementFontProperties(element)
+  );
```
spacedmonkey:
```suggestion
  page.elements = page.elements.map(pickElementFontProperties);
```

packages/element-library/package.json
```diff
@@ -49,6 +49,7 @@
     "@googleforcreators/stickers": "*",
     "@googleforcreators/transform": "*",
     "@googleforcreators/units": "*",
+    "@googleforcreators/story-editor":  "*",
```
spacedmonkey: This seems the wrong way around. Maybe it makes more sense to move useFont to it's own package and load that?

timarney: Maybe I can use it higher up or something --- it would be nice to not pull it into the element-library

timarney: So I think the question becomes for anything that pull props from the element 

i.e. in this case back up to 

```<DisplayElement```

Do we want to manipulate the element on the way in and or add font props with the extra data and follow that through the other code ... 

Or --- try to get the needed data at the part of the code that needs it for instance `calcFontMetrics` and `generateParagraphTextStyle`.

Comment (spacedmonkey) [reply packages/element-library/package.json]: @swissspidy Thoughts?

Comment (swissspidy) [reply packages/element-library/package.json]: Could you please elaborate on why this dependency was added/needed? What are we trying to achieve?

Theoretically adding it would be OK, but typically we've been passing additional information via props. See `getProxiedUrl`, `renderResourcePlaceholder`, `siblingCount`, `previewMode` as prior art.

Comment (timarney) [reply packages/element-library/package.json]: The original intention was looking for a way to get access to the data without having to pass it through several components ...

~👉  I think for `edit.js` and `display.js` we can pass as props and avoid the dependency.~
**Update:** Pushed an update removing the dependancy for edit and display https://github.com/GoogleForCreators/web-stories-wp/pull/12445/commits/85a9c2ad814c577c6cb5d8db46a1d2f40b031def. ☝️

**Comment by swissspidy** [comment]
cc @merapi @barklund @miina to review the approach here at an early stage

File: packages/story-editor/src/app/story/useStoryReducer/reducers/addElements.js [hunk 5]
```diff
@@ -76,6 +76,14 @@ export const addElements = (
     addedIds.push(...nonProducts.map(({ id }) => id));
   }
 
+  // store font props at the story level
+  draft.story.fonts = updateFonts(draft.story?.fonts, page.elements);
```
miina: It's a bit confusing that when an `element` is added, it has a fully defined font and then it's manually removed. Shouldn't it just be without that data from the start then? 🤔 Or is this still todo?

packages/migration/src/migrations/v0044_removeElementFontProperties.ts
```diff
@@ -0,0 +1,92 @@
...
+  ShapeElementV43,
+  StoryV43,
+  TextElementV43,
+  VideoElementV43,
+  UnionElementV43,
+} from './v0043_removeTagNames';
+
+export interface TextElementV44 extends Omit<TextElementV43, 'font'> {
+  font: {
+    family: string;
```
miina: 🎉

packages/story-editor/src/app/story/useStoryReducer/reducers/addElements.js
timarney: Yes --- I need to get back to the font storage piece at the story level (draft.story.fonts).

I need to update that code to pull font data (not from the element).

After that we should be able to update the default font `const` to remove everything but font family. I tried already but holding off for now -- as a bunch of Karma test use that data.  By updating the default font --- the text element shouldn't have those props at all.

timarney: I ended up adding a reducer for the font data vs modifying addElements to add the fonts to the state https://github.com/GoogleForCreators/web-stories-wp/pull/12445/files#diff-37271db74ac066b7cf0d328988fbd657676d6a9acdbe12277c0a8d6a17e5e1c8R77

timarney:
@merapi @barklund @swissspidy @miina I updated the summary for this PR to hopefully make more sense of the changes as there are a lot of files modified.

merapi:
@timarney  I started looking at this (sorry for being late) and found this bug:

Create a new story, insert Title1, and change the font to Anton. Looks like we need to initialize the object.
image

timarney:
> Create a new story, insert Title1, and change the font to Anton. Looks like we need to initialize the object.
Maybe something like:

Will take a look ... might be related to this

https://github.com/GoogleForCreators/web-stories-wp/pull/12445/files#diff-37271db74ac066b7cf0d328988fbd657676d6a9acdbe12277c0a8d6a17e5e1c8R74

merapi:
I pushed a fix so I can test it more.
It seems that font metrics are not used. Check the bounding box (height jump) of the text element on this branch vs main:

https://user-images.githubusercontent.com/21036224/198297303-750dc7f7-bea4-4541-acd6-8b07f4432c63.mov

timarney:
Thanks for taking a look @merapi --- do you think the overall approach is the right way to go for this PR ?  If yes I'll pick this up again next week and start fixing more bugs.

merapi:
> Thanks for taking a look @merapi --- do you think the overall approach is the right way to go for this PR ? If yes I'll pick this up again next week and start fixing more bugs.

Yes, one metrics instance per story and look up by font family 👍🏻

timarney:
Added a console log --- looks like it's sometimes calling with the metrics data included and later drops it --- will take a look at what's calling without passing the data.

Screen Shot 2022-11-03 at 5 42 28 AM

timarney:
@merapi @barklund @miina @swissspidy 

The current PR is having issues re-pulling the font data off the Font provider for some files the provider isn't available or we  ended up with some delayed loading etc... i.e. noticing calc font metrics ends up with or without data all in the same render cycle.

Re-thinking this PR a new approach to simplify get the font data back + less changes to the existing code 

**Save font data at story level**
- When inserting / updating fonts store that data at the Story level  --- don't worry about in-use fonts when inserting or updating

https://github.com/GoogleForCreators/web-stories-wp/pull/12625/files#diff-483babc0aedc4dd5e4061b9ffc4ed820599584aaa912a19b9ef4e711cffbaa4a

**When we save**
- Remove element font data outside of font family when we save 
- Clean up Story level font data to only use "in-use" fonts

**At this point we should have clean / reduced font storage**

https://github.com/GoogleForCreators/web-stories-wp/pull/12625/files#diff-ef719e5463b1b5a0cf15d71c9c66511747ac6966a37d8af84f5651b3e6be5932R25

**When we re-load**

At this point the code should operate the same as current main

-  Re-populate the element font data pulling from story level font data https://github.com/GoogleForCreators/web-stories-wp/pull/12625/files#diff-ea54c3a9b6ee8086ade8e30de7af7c6204697eac9c91d7c5173271c967c53975R87


https://github.com/GoogleForCreators/web-stories-wp/pull/12625  -- WIP

timarney:
Closing this for now --- picked up work in https://github.com/GoogleForCreators/web-stories-wp/pull/12625

PR #12460, opened 2022-10-11
Dashboard: Add global Page Advancement setting

### LINKED ISSUE #11528: Provide way to set manual page advancement as default
## Feature Description

Each new story has a default page advancement set to “Auto” with 7 seconds. [One user in the support forum](https://wordpress.org/support/topic/disable-autoplay-page-advancement-to-manual-by-default/) has asked if it would be possible to allow users to be able to disable the auto page advancement option so that each new story created has a manual page advancement set.

Current default setting for each story:

image

## Alternatives Considered

Rather than allow users to disable the auto option, there could be a setting to change what is used as the default setting for page advancement. 

This setting could be solely something for developers to change, **not** a UI option.

miina:
## Context
Adds Page Advancement global option to the Dashboard settings which will be used if the setting hasn't been set per story.

## To-do

- [x] Style
- [x] Tests
- [x] What to do with the migration V11?

## User-facing changes
Screenshot 2022-10-13 at 12 58 37

## Testing Instructions

1. Go to Dashboard settings
2. Change the Page Advancement settings
3. Verify that when adding a new story, the settings configured in the Dashboard are now used
4. Verify existing stories will not be influenced.

Two new tracking values: `change_auto_advance` and `change_default_page_duration`

## Checklist

---

Fixes #11528


packages/wp-dashboard/src/components/editorSettings/pageAdvancement/index.js
```diff
@@ -0,0 +1,130 @@
...
+        <SettingSubheading size={THEME_CONSTANTS.TYPOGRAPHY.PRESET_SIZES.SMALL}>
+          {TEXT.CONTEXT}
+        </SettingSubheading>
+      </div>
+      <InputsWrapper>
+        <InlineForm>
+          <Switch
+            groupLabel={TEXT.SWITCH_LABEL}
+            name="page-advancement-switch"
+            value={autoAdvance}
```
miina: Probably should store the value in a local state as well since it currently changes only after receiving new value after saving which can take some time.

packages/migration/src/migrations/v0011_pageAdvancement.ts
```diff
@@ -45,22 +45,15 @@ export type UnionElementV11 =
...
 function pageAdvancement(story: StoryV10): StoryV11 {
-  if ('autoAdvance' in story && 'defaultPageDuration' in story) {
-    return story as StoryV11;
-  }
-  return {
-    autoAdvance: true,
-    defaultPageDuration: 7,
-    ...story,
-  };
+  return story as StoryV11;
```
miina: Have to remove always setting the auto-advance and page duration, otherwise we can't really detect if it needs to be overridden by global settings.

miina: @merapi FYI I stepped through the code locally and looks like the `version` is `undefined` for a new story, meaning that every new story will go through the full migration script initially 🤷 Not sure if that's the intended way or a bug.

miina: @swissspidy Perhaps you have thoughts about this?
Looks like new stories will always go through all migrations. Since V11 adds auto-advance props but we shouldn't have those by default on new stories to use the global ones instead, we don't need V11 migration anymore. Should we just remove it? Is it a bug that the migration runs through new stories?

includes/Settings.php
```diff
@@ -349,6 +359,28 @@ public function register(): void {
 				'show_in_rest' => true,
 			]
 		);
+
+		register_setting(
```
spacedmonkey, line 366: Should really update unit tests [here](https://github.com/GoogleForCreators/web-stories-wp/blob/d87cf2a2f1a613c5a1a0e2f1d6ce8746b305ec87/tests/phpunit/integration/tests/Settings.php#L44).

includes/Admin/Editor.php
```diff
@@ -433,6 +437,8 @@ public function get_editor_settings(): array {
 			'version'                 => WEBSTORIES_VERSION,
 			'nonce'                   => $nonce,
 			'showMedia3p'             => true,
+			'globalAutoAdvance'       => (bool) $auto_advance,
+			'globalPageDuration'      => $page_duration,
```
spacedmonkey: Should this be cast to an int?

packages/migration/src/migrations/v0011_pageAdvancement.ts
```diff
@@ -45,22 +45,15 @@ export type UnionElementV11 =
...
-  autoAdvance: boolean;
-  defaultPageDuration: number;
+  autoAdvance?: boolean;
+  defaultPageDuration?: number;
 }
 export interface PageV11 extends Omit<PageV10, 'elements'> {
   elements: UnionElementV11[];
 }
 
 function pageAdvancement(story: StoryV10): StoryV11 {
```
spacedmonkey: So this does nothing now?

As crazy is this sounds, I think it might be cleaner to add a new migration to remove default values.

packages/wp-dashboard/src/components/editorSettings/editorSettings.js
```diff
@@ -401,6 +408,11 @@ function EditorSettings() {
                   vendors={vendors}
                   woocommerce={woocommerce}
                 />
+                <PageAdvancement
```
spacedmonkey, line 411: Stupid question. Is this a WordPress only feature? Should this code only live in wp-dashboard or should be in the dashboard package?

spacedmonkey:
Couple of questions and I think we need to work out what is going on with the migration, otherwise this is looking good to go.

includes/Admin/Editor.php
swissspidy: Isn't it guaranteed to be an int since we use `register_setting()` with `type => number`?

packages/migration/src/migrations/v0011_pageAdvancement.ts
swissspidy: > Looks like new stories will always go through all migrations. Is it a bug that the migration runs through new stories?

AFAIK it's intentional because this way a new story automatically gets the right and up-to-date data structure.

We shouldn't modify existing migrations though, since this won't work for existing stories.

It'd be better to add a new migration, e.g. to delete `defaultPageDuration` if it's `7`.

includes/Admin/Editor.php
spacedmonkey: Sadly no. That type does not mean that WordPress casts it for you. That type is used for the REST API, where type is important. 

I var_dumped it and returned a string. 
Screenshot 2022-10-13 at 18 40 46

WordPress 🙃

swissspidy: Thanks for testing! Let's add a cast then.

packages/migration/src/migrations/v0011_pageAdvancement.ts
miina: > It'd be better to add a new migration, e.g. to delete defaultPageDuration if it's 7.

Yep, I wasn't sure about what's the best way to proceed with that -- removed the current migration to make it work functionally.

If we add a new migration, the existing stories with the default values would be changed and this would mean the new global values would influence those stories, too, not just the new stories. If that's ok then this sounds like a good plan.

miina:
> Couple of questions and I think we need to work out what is going on with the migration, otherwise this is looking good to go.

Yep, that's still a todo in the description, too :) Pascal also recommended adding a new migration. If it's ok that all the previous stories with default values be influenced as well by the new global values, then sounds like a good plan.

packages/migration/src/migrations/v0011_pageAdvancement.ts
swissspidy: The new global value will default to the same number as before, right? So a story would still behave the same.

miina: Not necessarily. The new global value will be default indeed, however, if the user first changes the global value, and then opens an existing story with previous default values, it will use the new global values for the story -- since the migration runs only if a story is opened, and it can be opened after the global value has been changed.

barklund (approved):
Looks and works perfect!

spacedmonkey (approved):
LGTM

kkalarickal:
@swissspidy @miina 

I can confirm that changing the value of auto page advancement time is saved. When I create a new story, the global value is applied to the new story only (does not affect previously created stories).

However, when I publish the story and look at the source, it looks like the updated value is NOT passed on to the amp

Example:

- I update the global page adv to auto-15sec



- The new story that I created had default auto adv se to 15 sec

Test Story URL: https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/post.php?post=20421&action=edit#page=%2522e538f99e-de42-4385-987d-d02a3d867645%2522



- However, the AMP that was generated still showed 7 sec



- Also, if you publish and view the story, the time delay between auto page adv is actually 7 sec .. not the longer 15 sec.

kkalarickal:
I feel this ticket needs some more work.

1. Ensure that the new global value is actually reflected in the amp and the timing meets the new global number.

2. Either disallow decimals in the global page adv setting, or ensure that any fractions in the setting are transferred to the individual story. At the moment, if I set global ppage adv to 2.5 sec, the story shows 2 sec page adv value.

miina:
> * Ensure that the new global value is actually reflected in the amp and the timing meets the new global number.

Good catch! I saved the story before publishing so the values were correct in the front-end, too. Should be fixed now without saving, too.

> * Either disallow decimals in the global page adv setting, or ensure that any fractions in the setting are transferred to the individual story. At the moment, if I set global ppage adv to 2.5 sec, the story shows 2 sec page adv value.

Fixed, thanks!

kkalarickal:
@miina -- now the global value for page adv is applied by directly to the new story and within the amp (both draft and published version) as expected. I also tested other ways of creating a story (duplicate existing story, New > Story from top menu, Stories > Add new) and the page adv value matches the expected.

However, when I create a web-story from template, the updated global value is NOT applied

Test:
1. Set the global auto page adv to 15 sec or manual
2. Go to Explore Templates > Use template to create a web-story from template
3. Verify the Document > Page adv setting

Expected: the new global age adv 15 sec or manual
Observed: the old 7 sec default is found.

I think the path to create a new story from template should honor the global page adv settings.

Cc: @swissspidy

kkalarickal:
@swissspidy @miina - I think this ticket should be in-progress since creating the story via templates does not pick up the global page adv settings. So I am going to move this ticket back so that it reflects the current status.

merapi:
Good catch @kkalarickal but I think we should proceed with what we have here and create a follow-up to solve the template case because for me it's not so obvious that is should be taken from the global settings and not from the template settings.
Let's say I create a template for a recipe and set the auto-advancement to manual. When I create the story from that template, I still want it to be manual (and not global 20s) because it's a recipe and different people need a different amount of time to process/prepare the ingredients or follow the steps on a given page.

swissspidy:
> Let's say I create a template for a recipe and set the auto-advancement to manual.

@merapi Are you referring to custom saved page templates? Or the story templates from the dashboard?

merapi:
> @merapi Are you referring to custom saved page templates? Or the story templates from the dashboard?

Templates from the dashboard. So the question is - whether the template creator would like the new story pages to have a specific auto-advance rather than the global one. Is there a use case for that?

kkalarickal:
> Is everything else fine here?

Yes @merapi -- the global auto advance settings are working fine on all other paths for creating a new story. If the templates from dashboard have their own auto adv setting that is not over-riden by the global setting, then we should probably mention that in the Default Page Advancement blurb "All new stories will be set to this page advancement setting, except when created from default templates...."

merapi:
> we should probably mention that in the Default Page Advancement blurb

💯 

> I think the path to create a new story from template should honor the global page adv settings.

My biggest issue with this is:
When those templates were created the timeout was 7s, if somebody changes the global timeout to 3s or 20s most of those stories will be unreadable - pages will switch too fast and you will not able to read them fully, or they will be annoyed by how long they need to wait for the next page.

miina:
For visibility: it was decided that we'll use global value for the templates. Will create a follow-up PR for that.

PR #12467, opened 2022-10-13
Compatibility with hotlinked story posters

### LINKED ISSUE #12028: Theme/Plugin compatibility with hotlinked story posters
## Feature Description

When using a hotlinked story poster image, we save the image information in a separate `web_stories_poster` post meta field. There's no `_thumbnail_id` meta field because there's no actual attachment.

The downside is that themes relying on `the_post_thumbnail()` or `has_post_thumbnail()` won't pick this up, potentially causing an odd display in the default story archive, or causing issues with schema data generated by other plugins like Yoast SEO.

We could potentially filter `has_post_thumbnail` and the like, but that won't help for cases like `if ( has_post_thumbnail() ) { the_post_thumbnail(); }` which would just print nothing or could even lead to the theme printing some unnecessary markup.

## Additional Context

* https://wordpress.org/support/topic/story-archive-broken/#post-15857646

spacedmonkey:
## Summary

Hook into `has_post_thumbnail` and `post_thumbnail_html` filters to support, hotlinked poster images. 

## Relevant Technical Choices

Force the height and width to be the request height and width. As the hotlinked image will not have different sizes, we can only use the full size. 

## User-facing changes

Screenshot 2022-10-13 at 15 55 30

## Testing Instructions

1. Create a story. 
2. Add a hotlinked story image. Example. https://assets.computingcorner.co.uk/uploads/2022/10/example-3.png
3. Publish story. 
4. Go to FE of site. <site_url>/web-stories
5. See poster image used on as featured image. 

Try this on core themes, by changing site theme to 2010-2023 themes. 

## Checklist

---

Fixes #12028


swissspidy:
> Force the height and width to be the request height and width. As the hotlinked image will not have different sizes, we can only use the full size.

So potentially if a theme wants to display the featured image in a small size like 50x50, we still force the full size?

includes/Story_Post_Type.php
```diff
@@ -252,6 +254,79 @@ public function register_meta(): void {
...
+	 * @param int                          $post_id           The post ID.
+	 * @param int                          $post_thumbnail_id The post thumbnail ID, or 0 if there isn't one.
+	 * @param string|int[]                 $size              Requested image size. Can be any registered image size name, or
+	 *                                                        an array of width and height values in pixels (in that order).
+	 * @param string|array<string, string> $attr              Query string or array of attributes.
+	 * @return string The post thumbnail image tag.
+	 */
+	public function post_thumbnail_html( $html, $post_id, $post_thumbnail_id, $size, $attr ): string {
+		$post = get_post( $post_id );
+		if ( ! $post || $html || $post_thumbnail_id || $this->get_slug() !== get_post_type( $post ) ) {
```
timarney, line 301: Is there a case is where the html could be empty (null etc...) but there's a  $post_thumbnail_id ? --- I haven't look at the code that consumes this --- could that cause issues?

spacedmonkey:
> So potentially if a theme wants to display the featured image in a small size like 50x50, we still force the full size?

Yes. Because, what else could we do here? We only have one size.

includes/Story_Post_Type.php
spacedmonkey: `$html` can be filtered, so it could be an empty string or something.

swissspidy:
Let‘s put this behind a feature flag first before shipping this. Don‘t wanna break sites unexpectedly.

includes/Story_Post_Type.php
swissspidy: If it can be `null` or something other than a string (because it can be filtered), then we should remove the return type declaration and change the type to `mixed` just to be safe.

includes/Story_Post_Type.php
```diff
@@ -252,6 +265,79 @@ public function register_meta(): void {
...
+			'height'   => $height,
+		];
+
+		$attr = wp_parse_args( $attr, $default_attr );
+
+		$attr = array_map( 'esc_attr', $attr );
+		$html = '<img';
+
+		foreach ( $attr as $name => $value ) {
+			$html .= " $name=" . '"' . $value . '"';
```
swissspidy, line 333: Maybe use `esc_attr` on the value here?

spacedmonkey: This is done 3 lines above.

swissspidy: Hmm I see. It's not very obvious. Seems very easy to remove this by accident again in the future. Plus a bit wasteful and slow to loop over `$attr` twice.

I tend to prefer doing the `esc_attr` here instead of using `array_map`

swissspidy:
As per https://github.com/googleforcreators/web-stories-wp/issues/12028#issuecomment-1287419957, closing as a_maybelater_ for now.

PR #12476, opened 2022-10-14
Convert `activation-notice` to TypeScript

swissspidy:
## Context

This wasn't on the agenda, but I wanted to get more practice with converting code that uses styled-components.

## Summary

Converts the `activation-notice` package to TypeScript

## Testing Instructions

1.

## Checklist

---

Fixes #


packages/tracking/src/trackClick.ts
```diff
@@ -14,6 +14,11 @@
  * limitations under the License.
  */
 
+/**
+ * External dependencies
+ */
+import type { MouseEvent } from 'react';
```
spacedmonkey, line 20: Where is this used?

package-lock.json
```diff
@@ -59403,7 +59402,6 @@
         "@wordpress/i18n": "^4.17.0",
         "flagged": "^2.0.6",
         "polished": "^4.2.2",
-        "prop-types": "^15.8.1",
```
spacedmonkey, line 59406: Why is this being removed here?

packages/activation-notice/src/app/components/step1.tsx
```diff
@@ -87,13 +91,7 @@ function Step1() {
         target="_blank"
         rel="noreferrer"
       >
-        <Image
-          name="tips.png"
-          name2x="tips-2x.png"
-          width={150}
-          height={245}
-          $marginTop={60}
```
spacedmonkey, line 95: Are we getting rid of the margin top here?

spacedmonkey (approved):
Couple of nitpics and questions, but otherwise, looking good.

packages/activation-notice/src/app/components/step1.tsx
swissspidy: Getting rid of this prop because it wasn‘t actually used anywhere and does nothing.

package-lock.json
swissspidy: Because I‘ve gotten rid of prop-types in this package in favor of TS

packages/tracking/src/trackClick.ts
swissspidy: 30 lines below.

It was mistakenly using the wrong global `MouseEvent` type instead of the React-provided one.

packages/activation-notice/src/typings/styled-components.d.ts
```diff
@@ -0,0 +1,56 @@
...
+
+interface Font {
+  family: string;
+  size: string;
+  lineHeight: string;
+  fontWeight: number;
+}
+
+declare module 'styled-components' {
+  export interface DefaultTheme {
```
miina, line 27: I wonder if this should instead be just in the `design-system` package and imported from there?

swissspidy: `activation-notice` does not use the `design-system`, it has its own separate theme.

But yes eventually we need this for the `design-system` too :-)

PR #12482, opened 2022-10-17
Composer: Temporarily disable platform check

### LINKED ISSUE #12480: Composer: PHP Unit Tests not running on PHP 7.2
See discussion starting at https://xwp.slack.com/archives/C013J2EGH8R/p1665761420101459

Example run:

https://github.com/GoogleForCreators/web-stories-wp/actions/runs/3250983452/jobs/5335377179#step:17:19

Might have to do with a Composer version update

spacedmonkey:
## Summary

Fix composer install in CI, by setting platform check in composer.

[Context](https://php.watch/articles/composer-platform-check). 

## Relevant Technical Choices

In my testing, composer install / phpunit do not run locally in PHP 7.2/7.3 because of this requirement. 

I believe this bug came around, because the composer vendor cache was empty and build happened in a PHP 7.2 or 7.3 and this resulted in platform check.php file being generated. 

## Testing Instructions

1.

## Checklist

---

Fixes #12480

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -78,7 +78,8 @@
     "platform": {
       "php": "7.2.24"
     },
-    "sort-packages": true
+    "sort-packages": true,
+    "platform-check": false
   },
   "extra": {
     "downloads": {
--- a/composer.json
+++ b/composer.json
@@ -78,8 +78,8 @@
     "platform": {
       "php": "7.2.24"
     },
-    "sort-packages": true,
-    "platform-check": false
+    "platform-check": false,
+    "sort-packages": true
   },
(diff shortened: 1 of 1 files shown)


swissspidy:
If the cache is an issue, would using `composer install --no-cache` help?

Disabling the platform check seems dangerous because we could include newer code without knowing.

What I don't understand is why the platform check now suddenly thinks the required version is PHP 7.4 and not 7.2. Where does the 7.4 come from?

swissspidy (approved):
LGTM as a temporary workaround

PR #12507, opened 2022-10-18
Do not print title tags for classic themes

### LINKED ISSUE #12487: SEO: <title> tag duplicated when using SEO plugin
## Bug Description

When using the latest Web Stories and the latest released Yoast SEO plugin (v19.8) I noticed that the `<title>` tag behaves incorrectly.
The page has a single `<title>` only when using Twenty Twenty Two theme, while in these others: Twenty Twenty, Twenty Twenty One, Hello Elementor, Astra, GeneratePress, OceanWP - I see **two** titles in the page source.

This is a continuation of #12139, which was fixed in #12162 but (according to test instructions in PR) has been tested only with the 2022 theme.

## Expected Behaviour

There should always be only one `<title>` tag in the page source, regardless of plugins/themes activated.

## Steps to Reproduce

1. Activate Yoast SEO plugin
1. Activate not a Twenty-Twenty Two theme, for example: Astra, OceanWP, GeneratePress, etc.
1. View a single story
1. Verify the <title> tag is added

## Screenshots

Note the two search results when searching for `<title>`:
image

## Additional Context

- Plugin Version: 1.25.1
- Yoast SEO: 19.8 (default settings)
- WordPress Version: 6.0.2
- Operating System: MacOS Monterey
- Browser: Safari/Chrome/Firefox

swissspidy:
## Context

Stories can end up having duplicate title tags under some conditions.

## Summary

Disables default title tag output for classic themes

## To-do

- [x] Tests

## Testing Instructions

1. Activate Yoast SEO plugin
2. Activate a theme like Astra, OceanWP, GeneratePress, etc.
3. View a single story
4. Verify there is only one `<title>` tag in the source code
5. Activate a theme like Twenty Twenty Two
6. Verify there is  _still_ only one `<title>` tag in the source code

Repeat with a plugin like Rank Math SEO

## Checklist

---

Fixes #12487

### DIFF
--- a/includes/Discovery.php
+++ b/includes/Discovery.php
@@ -101,23 +101,26 @@ class Discovery extends Service_Base implements HasRequirements {
 	/**
 	 * Prints document title for stories.
 	 *
-	 * Works both for classic themes and block themes without any conditionals.
+	 * Works both for classic themes and block themes.
 	 *
 	 * @since 1.25.0
 	 *
 	 * @link https://github.com/GoogleForCreators/web-stories-wp/issues/12139
+	 * @link https://github.com/GoogleForCreators/web-stories-wp/issues/12487
 	 * @see _wp_render_title_tag()
 	 * @see _block_template_render_title_tag()
 	 */
 	public function print_document_title(): void {
+		$enable_metadata = function_exists( 'wp_is_block_theme' ) && wp_is_block_theme();
+
 		/**
 		 * Filters whether to print the document title.
 		 *
 		 * @since 1.25.0
 		 *
-		 * @param bool $enable_open_graph Whether to print the document title. Default to true.
+		 * @param bool $enable_open_graph Whether to print the document title. Defaults to true for block themes, false otherwise.
 		 */
-		$enable_metadata = apply_filters( 'web_stories_enable_document_title', true );
+		$enable_metadata = apply_filters( 'web_stories_enable_document_title', $enable_metadata );
 		if ( ! $enable_metadata ) {
 			return;
 		}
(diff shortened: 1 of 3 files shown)


includes/Discovery.php
```diff
@@ -101,23 +101,26 @@ public function register(): void {
 	/**
 	 * Prints document title for stories.
 	 *
-	 * Works both for classic themes and block themes without any conditionals.
+	 * Works both for classic themes and block themes.
```
spacedmonkey, line 104: Should this comment be updated.

includes/Discovery.php
```diff
@@ -101,23 +101,26 @@ public function register(): void {
...
 	 *
 	 * @since 1.25.0
 	 *
 	 * @link https://github.com/GoogleForCreators/web-stories-wp/issues/12139
+	 * @link https://github.com/GoogleForCreators/web-stories-wp/issues/12487
 	 * @see _wp_render_title_tag()
 	 * @see _block_template_render_title_tag()
 	 */
 	public function print_document_title(): void {
+		$enable_metadata = function_exists( 'wp_is_block_theme' ) && wp_is_block_theme();
```
spacedmonkey, line 114: How does this work with [Title Tag](https://codex.wordpress.org/Title_Tag).

includes/Discovery.php
swissspidy: I updated it already, do you have a recommendation to further update/clarify it?

includes/Discovery.php
swissspidy: Since we're not relying on theme support and don't use built-in functions like `_wp_render_title_tag` or `_block_template_render_title_tag`, this shouldn't really affect us at all.

The problem with block themes is just that plugins like Yoast SEO behave differently for them, which is why we're having these issues.

You can verify this by testing a theme like Twenty Ten (does not add `title-tag` support) and Twenty Fifteen (_does_ add `title-tag` support)

spacedmonkey (approved):

kkalarickal:
Recreated the problem on `main branch` by applying Astra theme and using Yoast SEO version Version 19.8. I set a specific title tag within the story yoast meta box and searched for that title in the published story AMP. Detected two instances of the <title> .... </title>

(Note: I also changed it to the modern Twenty-Twenty theme and confirmed that the problem was NOT occuring on `master branch`)

Verified the fix by shifting to `12507 branch` and applying Astra theme with Yoast SEO plugin active. With this change, there is only one <title>...</title> entry in the published story AMP

Next deactivated the Yoasst SEO and installed & activated Rank Math SEO plugin. Confirmed that only single <title>...</title> is present in the published story AMP.

Finally, I also changed the theme selection to Steam Punk lite, Twenty-TwentyOne, Twenty-Twenty and confirmed that only one <title>...</title> is present in the published story AMP when Yoast SEO plugin is activated.

PR #12508, opened 2022-10-18
TypeScript: Convert `elements` package

swissspidy:
## Context

This helps unblock some other conversion work, like for the `output` and the `masks` packages.

## Summary

Convert the `elements` package to TS.

## Relevant Technical Choices

* Fixed types in the `units` package
* Adjusted types in the `types` package

## User-facing changes

N/A

## Testing Instructions

1.

## Checklist

---

See #12126


packages/elements/src/types.ts
```diff
@@ -17,7 +17,9 @@
 /**
  * External dependencies
  */
-import PropTypes from 'prop-types';
+/* eslint-disable no-restricted-imports -- Still used by other packages. */
```
spacedmonkey, line 20: Maybe add a todo here, to remind us to remove this later.

packages/units/src/useUnits.ts
```diff
@@ -23,10 +23,12 @@ import { identity, useContextSelector } from '@googleforcreators/react';
...
  */
 import Context from './context';
+import type { State } from './types';
 
-// TODO: Fix selector type.
-function useUnits(selector: () => unknown) {
-  return useContextSelector(Context, selector ?? identity);
+function useUnits(): State;
+function useUnits<T>(selector: (state: State) => T): T;
+function useUnits<T>(selector: (state: State) => T | State = identity) {
```
spacedmonkey, line 30: This looks strange.

packages/elements/package.json
```diff
@@ -26,20 +26,21 @@
...
   "main": "dist/index.js",
   "module": "dist-module/index.js",
-  "source": "src/index.js",
+  "types": "dist-types/index.d.ts",
+  "source": "src/index.ts",
   "publishConfig": {
     "access": "public"
   },
   "dependencies": {
-    "@googleforcreators/animation": "*",
```
spacedmonkey, line 39: Not used?

spacedmonkey:
There seems to be three eslint errors.

spacedmonkey (changes requested):
There seems to be three eslint errors.

packages/units/src/useUnits.ts
swissspidy: It‘s called function overloading. Makes the types work nicely with context selectors. We use it in the rich-text package too.

packages/elements/package.json
swissspidy: Correct. This is a leftover from a recent refactoring.

packages/elements/src/types.ts
swissspidy: Doesn't seem needed, we'll get lint errors anyway once these are not used anymore,

spacedmonkey:
There is a lint error again.

swissspidy:
It's an issue with ESLint on `main`, not an issue with this PR.

spacedmonkey (approved):
SGTM

packages/elements/src/utils/isElementBelowLimit.ts
```diff
@@ -18,18 +18,24 @@
...
+function isElementBelowLimit(element: Element, verifyLink = true) {
   if (verifyLink && !element.link?.url?.length) {
     return false;
   }
+
   const limit = FULLBLEED_HEIGHT * 0.8 - DANGER_ZONE_HEIGHT;
   const { x, y, width, height, rotationAngle } = element;
   const points = getCorners(rotationAngle, x, y, width, height);
-  return Boolean(Object.keys(points).find((point) => points[point].y > limit));
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment -- See below.
```
miina: I can't see any explanation below though 😄

packages/elements/src/elementType.ts
```diff
@@ -0,0 +1,56 @@
...
+  hasEditMode: boolean;
+  hasEditModeIfLocked: boolean;
+  resizeRules: {
+    vertical: boolean;
+    horizontal: boolean;
+    diagonal: boolean;
+    minWidth: number;
+    minHeight: number;
+  };
+  getLayerText: (element: Element) => string;
```
miina, line 61: `panels`, too?

packages/elements/src/elementType.ts
```diff
@@ -0,0 +1,56 @@
...
+  Edit: FunctionComponent;
+  Frame: FunctionComponent;
+  Output: FunctionComponent;
+  canFlip: boolean;
+  isMedia: boolean;
+  isMaskable: boolean;
+  hasDesignMenu: boolean;
+  hasDuplicateMenu: boolean;
+  hasEditMode: boolean;
+  hasEditModeIfLocked: boolean;
```
miina, line 50: This seems to be optional.

miina (approved):
Would be great to get a review from Morten, too, as he also started work on this package.

packages/elements/src/utils/isElementBelowLimit.ts
swissspidy: Yeah I think the eslint autofix removed it again

packages/elements/src/elementType.ts
swissspidy: Too, what? What's the context?

packages/elements/src/elementType.ts
swissspidy: Yeah looks like only a few elements have that defined. I think it would make sense to make it required, but have it default to `false`, this way it's more explicit and you don't have any `undefined` vs `false` type difference issues down the road.

packages/elements/src/elementType.ts
miina: Ah, sorry -- should we add `panels` here, too? Looks like that's also needed for registering an element.

packages/elements/package.json
```diff
@@ -26,20 +26,21 @@
   },
   "customExports": {
     ".": {
-      "default": "./src/index.js"
+      "default": "./src/index.ts"
     }
   },
   "main": "dist/index.js",
   "module": "dist-module/index.js",
```
spacedmonkey, line 33:
```suggestion
  "main": "dist/index.ts",
  "module": "dist-module/index.ts",
```

swissspidy: No that‘s incorrect, these build directories only contain JS files

packages/elements/src/elementType.ts
```diff
@@ -0,0 +1,75 @@
...
+interface BaseDefinition {
+  type: string;
+  name: string;
+  defaultAttributes: Record<string, unknown>;
+}
+
+export interface PageElementDefinition extends BaseDefinition {
+  type: 'page';
+  defaultAttributes: Record<string, never>;
+}
```
swissspidy, line 32: This part is not really ideal because of our special `'page'` element type.

PR #12510, opened 2022-10-19
Story Locking: Change interval to 60 seconds

### LINKED ISSUE #12435: Story Locking - delay in notification allows for concurrent editing by more than one user
## Bug Description

If someone takes over your story, you are not informed immediately. But you are allowed to make changes and save them (thus effectively still being in control) for a few minutes. I get that we can’t push the take over message instantly to the client, but we should at least not allow saving and display the message as an error if you do try to save.

## Additional Context

- Plugin Version:
- WordPress Version:
- Operating System:
- Browser:

spacedmonkey:
## Summary

Now that the internal of when post locking is checked, down to 60 seconds over 150 seconds. This will improve when take over dialog is shown. 

60 seconds brings this inline with WordPress cores heartbeat api. 

## Testing Instructions

1. Enable post locking take over. 
2. Create a story. 
3. Add title. 
4. Save draft. 
5. In another window. Login as another user. 
6. Edit draft story. 
7. Take over story. 
8. return to original Window. 
9. Wait 60 seconds for dialog to show. 

## Checklist

---

Fixes #12435

### DIFF
--- a/includes/Admin/Editor.php
+++ b/includes/Admin/Editor.php
@@ -352,9 +352,7 @@ class Editor extends Service_Base implements HasRequirements {
 			admin_url( 'edit.php' )
 		);
 
-		/** This filter is documented in wp-admin/includes/ajax-actions.php */
-		$time_window = apply_filters( 'wp_check_post_lock_window', 150 );
-		$user        = wp_get_current_user();
+		$user = wp_get_current_user();
 
 		/** This filter is documented in wp-admin/includes/post.php */
 		$show_locked_dialog = apply_filters( 'show_post_locked_dialog', true, $post, $user );
@@ -426,7 +424,7 @@ class Editor extends Service_Base implements HasRequirements {
 				'publisher' => $publisher_name,
 			],
 			'postLock'                => [
-				'interval'         => $time_window,
+				'interval'         => 60,
 				'showLockedDialog' => $show_locked_dialog,
 			],
 			'canViewDefaultTemplates' => true,


swissspidy:
> 60 seconds brings this inline with WordPress cores heartbeat api.

Where is that defined in core btw?

swissspidy:
Also, @kkalarickal does this resolve the time discrepancy you've seen between story editor and post editor? Looks like now both should be checking the lock every 60s

spacedmonkey:
> > 60 seconds brings this inline with WordPress cores heartbeat api.
> 
> Where is that defined in core btw?

Good question. This took me a while to find. [Here](https://github.com/WordPress/wordpress-develop/blob/a7a9e7602c51785de5261c15d8a720c535e829bf/src/js/_enqueues/wp/heartbeat.js#L62-L63). But you can change with a filter like [this](https://github.com/wp-media/heartbeat-control/blob/17ad6264e165766d4b792d73481b807219016001/Heartbeat.php#L122).

swissspidy (approved):

kkalarickal:
Confirmed that without the fix, the call to lock happens at a 2-3-2-3-2-3 min interval

After loading `12510 branch` the story locking is happening at 1 min intervals

@swissspidy .. as regards the discrepancy between story editor and post editor, this definitely makes it more snappy repose when someone takes over and the original author tries to edit. I feel this is a good improvement in how the user would expect the story locking to work.

PR #12511, opened 2022-10-19
Remove `fontWeight` and `backgroundOverlay` from element properties

### LINKED ISSUE #12258: Add data migrations to remove lingering properties
## Feature Description

This is a follow-up to #12131.

Despite previous migrations, there can be stories with lingering `backgroundOverlay` and `fontWeight` properties on elements. Let's add two migration functions that remove these.

See ac52897ed23f4b3e3233a495d54637119c970652

miina:
## Summary
Removes `fontWeight` and `backgroundOverlay` from the element properties. These have been removed before in migrations as well, however, looks like were added again via templates.

## Testing Instructions

1.

## Checklist

---

Fixes #12258


packages/migration/src/migrations/v0044_unusedProperties.ts
```diff
@@ -0,0 +1,90 @@
...
+  return {
+    elements: elements.map(updateElement),
+    ...rest,
+  };
+}
+
+function updateElement(element: UnionElementV43): UnionElementV44 {
+  if ('fontWeight' in element) {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment -- Removes lingering properties that came from the templates but don't exist on the type.
+    // @ts-ignore -- Ignore fontWeight not existing on element.
```
miina, line 79: Not sure if there's an alternative to just ignoring the type error here -- the error is correct, the type does not include these props anymore, however, since we still got them from the templates somehow, then these do exist in some cases.

swissspidy (approved):

spacedmonkey (approved):

PR #12519, opened 2022-10-24
TypeScript: Convert `masks` package

### LINKED ISSUE #12208: TypeScript: Convert `masks` package

swissspidy:
## Summary

Converts the `masks` package to TypeScript

## Testing Instructions

1.

## Checklist

---

Fixes #12208


packages/masks/src/typings/styled-components.d.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import 'styled-components';
+
+declare module 'styled-components' {
```
spacedmonkey, line 19: Should this live in the type package?

swissspidy: Eventually this will have to live in the `design-system` package (once we start migrating it), which is where our theme is defined.

The `types` package itself will likely go away at some point, with types being defined in their respective packages again (e.g. pattern types will be in patterns package)

packages/masks/src/frame.tsx
```diff
@@ -164,16 +192,19 @@ function WithDropTarget({
...
+  element: MediaElement;
+  style: CSSProperties;
+  fill: boolean;
+  flip: Flip;
+  children: JSX.Element;
+  draggingResource: Resource;
+  activeDropTargetId: string;
+  isDropSource: (type: string) => boolean;
+  isSelected?: boolean;
+  registerDropTarget: () => void;
```
spacedmonkey: Why is this different from `registerDropTarget: (id: string, pathRef: SVGPathElement | null) => void;
  unregisterDropTarget: (id: string) => void;`

spacedmonkey (approved):
LGTM

packages/masks/src/frame.tsx
swissspidy: Good catch, I need to update this.

PR #12554, opened 2022-10-27
Add "Copy story data" button to error boundary screen 

### LINKED ISSUE #11394: Add "Copy story data" button to error boundary screen
## Feature Description

It would be useful to add an option for web stories users to export a story as a JSON file to provide to support for debugging. 

This would simply be an export function and would not allow users to export/import story data. This would allow us in the support forums to request this data from the user to help with our troubleshooting of their story issues.

The feature would work much like other plugins with an export settings feature.  

## Alternatives Considered

1. This could be included in the Site Health Information rather than adding an option for the user to download a JSON file. 

2. An alternative to this would be to continue using the process we have in place which is less efficient but does work. This improvement would eliminate many steps in the process making everything much more efficient. 

## Additional Context

N/A

timarney:
## Context

It would be useful to add an option for web stories users to export a story as a JSON file to provide to support for debugging.

This would simply be an export function and would not allow users to export/import story data. This would allow us in the support forums to request this data from the user to help with our troubleshooting of their story issues.

## Summary

Adds a `copy story data` to Error Boundary screen

See Video:
https://www.youtube.com/watch?v=AKZv9o0-F0w

Screen Shot 2022-10-27 at 8 19 43 AM

## Testing Instructions

1.  To test I needed to modify the code to throw an error i.e. a non existent callback or you can modify the error state of the ErrorBoundry

```
 state = {
    error: null,
    errorInfo: null,
  };
```

No

No

No

## Checklist

---

Fixes #11394


packages/story-editor/src/components/errorBoundary/copyStoryDataToClipboard.js
```diff
@@ -0,0 +1,63 @@
...
+  const { pages, current, selection, story } = reducerState;
+
+  const copyToClipboard = useCallback(async () => {
+    const jsonStr = JSON.stringify(
+      { pages, current, selection, story },
+      null,
+      2
+    );
+    await navigator.clipboard.writeText(jsonStr);
+    alert(__('Copied to clipboard', 'web-stories'));
```
spacedmonkey: Wrap in a try / catch, just incase.

spacedmonkey (approved):
Works well in my testing.

packages/story-editor/src/components/errorBoundary/index.js
```diff
@@ -46,15 +47,16 @@ class ErrorBoundary extends Component {
...
     return this.props.children;
   }
 }
 
-const shouldDisableErrorBoundaries =
-  typeof WEB_STORIES_DISABLE_ERROR_BOUNDARIES !== 'undefined' &&
-  WEB_STORIES_DISABLE_ERROR_BOUNDARIES === 'true';
-export default shouldDisableErrorBoundaries
-  ? ({ children }) => children
-  : ErrorBoundary;
```
swissspidy: Why was this `WEB_STORIES_DISABLE_ERROR_BOUNDARIES` check removed?

If we wanna remove it, we should do so in a separate PR since this is referenced in many places. Otherwise we end up with dead code.

packages/story-editor/src/components/errorBoundary/copyStoryDataToClipboard.js
```diff
@@ -0,0 +1,67 @@
...
+ * Internal dependencies
+ */
+import { useStory } from '../../app/story';
+
+function CopyStoryDataToClipboard() {
+  const { reducerState } = useStory(({ internal }) => ({
+    reducerState: internal.reducerState,
+  }));
+
+  const { pages, current, selection, story } = reducerState;
```
swissspidy:
```suggestion
  const { pages, current, selection, story  } = useStory(({ internal }) => ({
    pages: internal.reducerState.pages,
    current: internal.reducerState.current,
    selection: internal.reducerState.selection,
    story: internal.reducerState.story,
  }));
```

Generally, we'd want to reduce the number of re-renders, which we can achieve by only returning what we need.

Example: If `reducer.foo` constantly changes, with the current version this component would always re-render. But with this suggestion it won't, because it's not actually used.

packages/story-editor/src/components/errorBoundary/copyStoryDataToClipboard.js
```diff
@@ -0,0 +1,67 @@
...
+ * External dependencies
+ */
+import { __ } from '@googleforcreators/i18n';
+import {
+  Button,
+  BUTTON_SIZES,
+  BUTTON_TYPES,
+  BUTTON_VARIANTS,
+} from '@googleforcreators/design-system';
+import { useCallback } from 'react';
```
swissspidy: We import from `@googleforcreators/react`

packages/story-editor/src/components/errorBoundary/errorActions.js
```diff
@@ -86,6 +91,7 @@ function ErrorActions({ error, errorInfo }) {
         >
           {__('Reload', 'web-stories')}
         </Button>
+        {withStoryData && <CopyStoryDataToClipboard />}
```
swissspidy: Do we actually need this check?

`CopyStoryDataToClipboard` can just `return null` if there's no story data available.

packages/story-editor/src/components/errorBoundary/index.js
timarney: Bad commit on my end --- at one point I created a new error boundary file.  Reverted now.

packages/story-editor/src/components/errorBoundary/errorActions.js
timarney: This is a contrived example but if something doesn't catch higher up I think it will just error out like this. 

Screen Shot 2022-10-31 at 7 40 04 AM

In this case I forced a miss  ☝️ of the higher up boundries.

Thinking it's better to be specific vs potentially have it error out trying to get data isn't available to access.  Thoughts?

swissspidy: My concern is that this is error prone when code gets moved around, and that we can just make `CopyStoryDataToClipboard` more robust instead of adding these`withStoryData` checks everywhere.

Something like:

```js
  // Get story data if it's available, otherwise return early.
  const { pages, current, selection, story } = useStory(({ internal }) => ({
    pages: internal?.reducerState?.pages,
    current: internal?.reducerState?.current,
    selection: internal?.reducerState?.selection,
    story: internal?.reducerState?.story,
  }));

  if (!pages || !current || !selection || !story) {
    return null;
  }
```

WDYT?

timarney: Okay updated --- I don't think the example I showed can happen give the way the boundaries are setup,

PR #12557, opened 2022-10-28
Reworking the `elements` package

### LINKED ISSUE #12126: TypeScript: Convert `elements` package

barklund:
## Context

More work on the elements package, as well as deprecating the types package completely.

(this is slightly WIP - the git merge was a bit muddy to get an overview of, so might make additional changes after creating this PR)

## User-facing changes

_None (I hope!)_

## Testing Instructions

- Full (single-browser) smoke-test is probably a good idea, as this touches on a lot of things through the entire application.

## Checklist

---

Fixes #12126


spacedmonkey:
This PR should be renamed removing the types package.

packages/elements/src/utils/getElementOrigin.ts
```diff
@@ -23,6 +23,7 @@
  * @param offsets Media element offsets
  * @return object containing horizontal and vertical transform origin percentages
  */
+
```
swissspidy: Errant space; now the docblock is no longer actually attached to the function.

packages/elements/tsconfig.json
```diff
@@ -7,8 +7,7 @@
   "references": [
     { "path": "../i18n" },
     { "path": "../media" },
-    { "path": "../patterns" },
-    { "path": "../units" }
```
swissspidy: Since the `units` package is still listed as a dependency in `package.json`, we should keep the reference here as well.

packages/elements/src/utils/isElementBelowLimit.ts
```diff
@@ -33,10 +35,6 @@ function isElementBelowLimit(element: Element, verifyLink = true) {
   const limit = FULLBLEED_HEIGHT * 0.8 - DANGER_ZONE_HEIGHT;
   const { x, y, width, height, rotationAngle } = element;
   const points = getCorners(rotationAngle, x, y, width, height);
-  return (Object.keys(points) as Array<Extract<keyof Corners, Corners>>).some(
-    (point) => {
-      return (points[point] as typeof points[keyof typeof points]).y > limit;
-    }
-  );
+  return Boolean(Object.values(points).find((coord) => coord.y > limit));
```
swissspidy: Thanks for simplifying this one 😄 

But why `Boolean( .find() )` when `.some()` exists?

packages/media/src/types/resource.ts
```diff
@@ -49,7 +49,7 @@ export interface ResourceSize {
   height: number;
 }
 
-export type ResourceId = string | number;
```
swissspidy: This was `string | number` for a reason.

Media3p resources typically have string IDs, whereas WordPress media resources have numeric values.

packages/media/src/types/resource.ts
```diff
@@ -77,20 +77,13 @@ export interface Resource {
   /** Whether the resource externally hosted. */
   isExternal: boolean;
   /** Whether the resource is a placeholder. */
-  isPlaceholder?: boolean;
+  isPlaceholder: boolean;
   /** Whether the resource needs a CORS proxy. */
-  needsProxy?: boolean;
```
swissspidy: IIRC these were made optional because especially in templates not all resources have this property. I think we only added in later and didn't add a data migration because it was kinda unnecessary.

So to reflect that, we should keep them optional, or add a migration.

packages/media/src/types/resource.ts
```diff
@@ -77,20 +77,13 @@ export interface Resource {
...
+  needsProxy: boolean;
   /** Resource creation date. */
   readonly creationDate?: string;
   /** Resource sizes */
-  sizes: { [key: string]: ResourceSize } | never[];
```
swissspidy: This `never[]` was added because of the templates, where images don't have multiple sizes and instead have `"sizes": [],` (check out the template JSON files to see it)

If we change this, we should add a migration & some code to turn this into `sizes: {}` if there are no sizes or making it optional.

packages/media/src/types/resource.ts
```diff
@@ -77,20 +77,13 @@ export interface Resource {
...
+  sizes: { [key: string]: ResourceSize };
   /** Resource author attribution */
   attribution?: Attribution;
-
-  // TODO: This property should probably not end up in a story. Only relevant for the media library.
-  provider?: 'local' | 'unsplash' | 'coverr' | 'tenor' | 'tenor_stickers';
-
-  // TODO: Figure out why sometimes _images_ end up having these properties.
-  posterId?: ResourceId;
-  isOptimized?: boolean;
```
swissspidy: Also added because of the templates IIRC.

If we remove this here, we should also remove these properties from the affected images in the templates.

packages/media/src/types/resource.ts
```diff
@@ -77,20 +77,13 @@ export interface Resource {
...
   /** Resource creation date. */
   readonly creationDate?: string;
   /** Resource sizes */
-  sizes: { [key: string]: ResourceSize } | never[];
+  sizes: { [key: string]: ResourceSize };
   /** Resource author attribution */
   attribution?: Attribution;
-
-  // TODO: This property should probably not end up in a story. Only relevant for the media library.
-  provider?: 'local' | 'unsplash' | 'coverr' | 'tenor' | 'tenor_stickers';
```
swissspidy, line 91: If we remove this here, we should look into fixing this, hence the TODO.

packages/media/src/types/resourceInput.ts
```diff
@@ -0,0 +1,75 @@
...
+import type { ResourceSize, Attribution } from './resource';
+import type { TrimData } from './videoResource';
+import type { Output } from './gifResource';
+
+export interface ResourceInput {
+  /**
+   * The resource ID.
+   * TODO: currently this value is local to the editor's media system.
+   */
+  id: string;
```
swissspidy: Can be a number as well (hence the `ResourceId` type)

swissspidy (commented):
An additional note:

We have separate `elements` and `element-library` packages for a reason, so that these are decoupled.

We should not put type definitions for individual element types back into `elements` package when we have this dedicated package.

packages/elements/src/utils/isElementBelowLimit.ts
barklund: Good question, I'll simplify further.

packages/elements/tsconfig.json
barklund: Yes, that's a mistake.

packages/media/src/types/resource.ts
barklund: Sorry, a lot of the diff from here on down are due to the merge conflicts and not actually because I wanted to change what you had done, I just kept what I had. I'll go over everything here in more detail today.

barklund:
> We have separate `elements` and `element-library` packages for a reason, so that these are decoupled.
> 
> We should not put type definitions for individual element types back into `elements` package when we have this dedicated package.

Yeah, that's why I wanted to convert both packages to put these in the element library, but that's turned out to be too much work (a lot of the components are tricky), so the element types are just here in the elements package until we get to the element library package. With the element library already depending on the elements package, I didn't think this was a big problem.

I we can put them in the element library right away without converting the rest of that package, I'm all for it, but didn't know the right way to do that.

swissspidy:
> I we can put them in the element library right away without converting the rest of that package, I'm all for it, but didn't know the right way to do that.

In `tsconfig.json` you can adjust `"include"` to something like `"include": ["src/types.ts"]` / `"include": ["src/types/**/*"]` so that only this one file/folder is checked by the compiler, and nothing else.

Here's an example:

https://github.com/GoogleForCreators/web-stories-wp/tree/try/element-library-types

This works well, the only problems are `packages/masks/src/borderedMaskedElement.tsx` referencing `ShapeElement` and the templates package referencing all element types (`TemplatePage` using `UnionElement[]` / `packages/templates/src/getTemplates.ts` modifying `element.resource.src`

Since these packages shouldn't really rely on `element-library` just for types, we need to find a better solution there.

For `masks` it probably means referencing `BackgroundableElement` / `DefaultBackgroundElement` instead of `ShapeElement`

For `packages/templates/src/getTemplates.ts` maybe means simply ignoring TS errors?

barklund:
> > I we can put them in the element library right away without converting the rest of that package, I'm all for it, but didn't know the right way to do that.
> 
> In `tsconfig.json` you can adjust `"include"` to something like `"include": ["src/types.ts"]` / `"include": ["src/types/**/*"]` so that only this one file/folder is checked by the compiler, and nothing else.
> 
> Here's an example:
> 
> https://github.com/GoogleForCreators/web-stories-wp/tree/try/element-library-types
> 
> This works well, the only problems are `packages/masks/src/borderedMaskedElement.tsx` referencing `ShapeElement` and the templates package referencing all element types (`TemplatePage` using `UnionElement[]` / `packages/templates/src/getTemplates.ts` modifying `element.resource.src`
> 
> Since these packages shouldn't really rely on `element-library` just for types, we need to find a better solution there.
> 
> For `masks` it probably means referencing `BackgroundableElement` / `DefaultBackgroundElement` instead of `ShapeElement`
> 
> For `packages/templates/src/getTemplates.ts` maybe means simply ignoring TS errors?

Thanks, I'll work through these and see if I can find a good solution!

barklund:
> > > I we can put them in the element library right away without converting the rest of that package, I'm all for it, but didn't know the right way to do that.
>>
> > This works well, the only problems are `packages/masks/src/borderedMaskedElement.tsx` referencing `ShapeElement` and the templates package referencing all element types (`TemplatePage` using `UnionElement[]` / `packages/templates/src/getTemplates.ts` modifying `element.resource.src`
> 
> Thanks, I'll work through these and see if I can find a good solution!

We already had the `BackgroundableElement` for what masks wanted, and for template, `MediaElement` and `SequenceResource` (a resource with a poster) combined seems to fix the day.

Thanks for the hints!

packages/elements/src/types/element.ts
```diff
@@ -40,6 +46,8 @@ export interface Mask {
   name?: string;
   path?: string;
   ratio?: number;
+  iconPath?: string;
+  iconRatio?: number;
   supportsBorder?: boolean;
 }
```
swissspidy, line 52: Wondering: should the `Mask` type better live in the `masks` package?

packages/media/src/types/sequenceResource.ts
```diff
@@ -17,14 +17,14 @@
...
-import type { Resource, ResourceId } from './resource';
+import type { Resource } from './resource';
 
 /** This is an abstract parent type of both video and gif resources */
 export interface SequenceResource extends Resource {
   /** The resource's poster. */
   poster?: string;
   /** The resource's poster ID. */
-  posterId?: ResourceId;
+  posterId?: string;
```
swissspidy: This should continue to use `ResourceId`.

packages/elements/src/types/elementDefinition.ts
```diff
@@ -0,0 +1,96 @@
...
+  box: ElementBox;
+  previewMode?: boolean;
+  renderResourcePlaceholder?: boolean;
+}
+
+export interface DisplayProps<E extends Element> {
+  element: E;
+}
+
+export interface ElementDefinition<E extends Element = Element> {
```
swissspidy, line 64: I like the refactoring here with generics! 👍 

Looking forward to using these types and see how well it works.

swissspidy (commented):
There are some test failures & prop types warnings, but otherwise looks good!

packages/elements/src/types/elementDefinition.ts
```diff
@@ -0,0 +1,96 @@
...
+  type: ElementType;
+  name: string;
+  isMedia?: boolean;
+  getLayerText: (element: E) => string;
+  defaultAttributes: Partial<E>;
+  Edit: React.FC<EditProps<E>>;
```
miina: I've seen in multiple articles / resources mentioned that `React.FC` is better not to use, e.g. because it causes the component to implicitly accept children. I _think_ there were some other reasons as well. Thoughts? You probably have more information about it.

barklund: That has been [changed in React 18](https://solverfox.dev/writing/no-implicit-children/). `React.FC` (or `React.FunctionComponent`) no longer implicitly accepts children.

**BUT** we're _not_ using React 18, and probably never will because of all the migration issues. So you're right, it should probably be changed to a more clean type.

barklund: `React.VoidFunctionComponent` seems to be [the way](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/34237#issuecomment-915178375) to do it pre-18. Thanks!

packages/elements/src/types/element.ts
barklund: The `masks` package depends on the `elements` package, and that dependency is pretty locked-in. There are all the components for display, frame, and output, that have element typings as well as several util functions (`shouldDisplayBorder`, `canMaskHaveBorder`, ...). And the big one is the `BorderedMaskedElement` using many element features.

Some of these could be solved by utilizing `DimensionableElement` from units package (though the border in that package does not include a color which masks need, which would add a dependency on the patterns package, but that's probably okay), others could be solved by also adding a `MaskableElement` to the masks package, but we still also need things like the `flip` and `isBackground` properties from the elements (in the `BorderedMaskedElement`), so that would require making even more element interfaces.

It's doable, but it's getting a bit complex (a naming is getting terrible 😂). Let's move this idea to a new ticket.

packages/elements/src/types/elementDefinition.ts
barklund: Fixed in 1e3577e885.

packages/media/src/types/sequenceResource.ts
barklund: Fixed in 1e3577e885.

packages/elements/src/types/elementDefinition.ts
miina: Nice! Should we add the eslint configuration as well to prefer this to `React.FC` mentioned in the [comment](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/34237#issuecomment-915178375) you shared? cc @swissspidy

miina (approved):
Nice rework! The JS tests are failing, otherwise LGTM 🐘

packages/elements/src/types/elementDefinition.ts
swissspidy: Sure, why not, perhaps in a new PR?

barklund:
I think all the remaining errors come from the fact, that [`htmlToImage.toCanvas` never resolves](https://github.com/GoogleForCreators/web-stories-wp/blob/main/packages/story-editor/src/utils/storyPageToCanvas.js#L39) (neither successfully or with an error). I've tried all sorts of things, but as far as I can tell, it just halts there.

@merapi, do you have any idea how that can be? I really don't get it? It's probably something obvious, like a missing property on some element that has to be there, but why doesn't it throw an error then? Why just nothing?

swissspidy:
Made that same observation here:

https://github.com/GoogleForCreators/web-stories-wp/issues/12513#issuecomment-1288986873

swissspidy:
> The product type returned by the getProducts API method really shouldn't be the product element type

Yeah that sounds wrong, it should return a „ProductData“ type or something.

> Where should the new non-element product interface live though? Also in elements? We need it in both the dashboard and and the story-editor packages.

I don‘t think dashboard does anything with product data (except for testing that the API endpoint works, for which we don‘t really need types).

Since ProductData is used by both the editor and the product element type, moving to elements might be easiest for now. In the future it could even be its own package though

kkalarickal:
@barklund - I spent some time smoke testing the `12557 branch` on stories-qa server and basic sanity is maintained. I have some more scenarios to cover which I will do in the morning. 

Basically, in a single browser instance, I went through the following:
- Organizing web stories via Dashboard/Categories/Tags
- Managing web stories Settings
- Authoring/editing web stories via the stories editor
--Insert various elements (images, videos, stickers, shapes, products, background audio, closed captions)
-- Manipulate the element style (fonts, colors, orientation, borders, filters, background audio, closed captions, accessibility, record video/audio, manipulate video, animations)
-- Working with Preset/Saved styles and templates
- Draft/Preview/Publish/UnPublish pathways, with autosave and backup of half baked stories
- Grid view, Preview, amp verification, poster image, publisher logo

The functionality follows whatever the behavior is on master-branch. There were a bunch of console errors that I wanted to share with you:

- when opening the editor


- when embedding story into a WP Post

barklund:
> The functionality follows whatever the behavior is on master-branch.

Good to hear.

> There were a bunch of console errors that I wanted to share with you:

Hmm, that's weird. I'll check if it means anything.

barklund:
> > There were a bunch of console errors that I wanted to share with you:
> 
> Hmm, that's weird. I'll check if it means anything.

I'm pretty sure they're unrelated (comes from other plugins). Let's proceed.

barklund:
@kkalarickal, merging this to unblock a number of other issues. If you do end up finding something, please alert me and I'll fix it asap!

PR #12559, opened 2022-10-28
Story Locking: Ensure author users and below correctly see information

### LINKED ISSUE #12441: Author users and below not seeing story takeover dialog
## Bug Description

I noticed in testing that when an author user's story is taken over by someone else, they don't see the dialog at all. This is because there's no `user` object since it's absent from the REST API due to lack of capabilities to list users.

## Expected Behaviour

Author users or contributors should still be able to see a takeover dialog.

## Steps to Reproduce

1. Create story as author
2. In a new window, take over editing as an admin
3. As an author, there should be a takeover message, but there isn't

## Screenshots

Screenshot 2022-10-07 at 17 15 25

https://github.com/GoogleForCreators/web-stories-wp/blob/31d01788be3a19732f0fc03085183cb6223260ca/packages/wp-story-editor/src/components/postLock/postLock.js#L189

It works in WordPress:

Screenshot 2022-10-07 at 17 21 28

## Additional Context

- Plugin Version: 1.26.0-alpha
- WordPress Version: 6.1-beta
- Operating System: macOS
- Browser: Chrome

### LINKED ISSUE #12604: Story Locking: missing information on dashboard depending on role
## Bug Description

When a web story is being edited by user-A, it appears with a lock icon and a thumbnail of the user-A when viewed in the dashboard by user-B. While testing the `main branch` for 1.27 bug bash, I see that the lock icon and thumbnail is NOT visible depending on the user profile of the entity logged into the wp-admin.

## Expected Behaviour

Irrespective of the profile of the user who is logged in, the lock icon and thumbnail should appear consistently.

## Steps to Reproduce

1. Create an Admin, Editor and Author user profile within the WP instance
2. Login on separate/independent browser sessions as each and open/edit a webstory within the GWS plugin.
3. Now, create and login to another admin user profile on a separate browser instance.
4. If you go to the dashboard, you will see that the three stories opened in step 2, will be represented as "locked" - with the lock icon and the thumbnail for admin/editor/author visible
5. Now create and login as another editor user profile on a separate browser instance.
6. If you go to the dashboard, you will see that one or more of the entries for the different web stories are NOT showing either locked icon or thumbnail or neither.
7. If you repeat the steps 5 & 6 with a new author user profile, there might be a dashboard without any indication that the stories are locked by different user profiles

## Additional Context

- Plugin Version: 1.27.0-alpha.0+36c4510
- WordPress Version: WP 6.0
- Operating System: win10
- Browser: Chrome

spacedmonkey:
## Summary

Instead of using the embedded user rest api request, just hardcode the user data in the api response. 

## Testing Instructions

1. Create a story as a user with Author role. 
2. Save draft. 
3. Open another browser. 
4. Login as another user. 
5. Open draft story. 
6. See lock message.
7. Take over editing
8. As the Author user, see the other user's name & avatar in the takeover dialog

## Checklist

---

Fixes #12441
Fixes #12604


packages/wp-story-editor/src/components/postLock/postLock.js
```diff
@@ -96,13 +96,17 @@ function PostLock() {
 
   // When async call only if dialog is true, current user is loaded and post locking is enabled.
   const doGetStoryLock = useCallback(() => {
-    if (showLockedDialog && currentUserLoaded) {
-      getStoryLockById(storyId, stories)
-        .then(({ locked, nonce: newNonce, _embedded }) => {
+    (async () => {
```
spacedmonkey, line 99: Converted this to async code is not needed for this, but made it easier to read.

swissspidy (commented):
At first glance looks good 👍  (besides the failing tests of course)

tests/phpunit/integration/tests/REST_API/Stories_Lock_Controller.php
```diff
@@ -69,7 +69,7 @@ public function test_register(): void {
 		$routes = rest_get_server()->get_routes();
 
 		$this->assertArrayHasKey( '/web-stories/v1/web-story/(?P<id>[\d]+)/lock', $routes );
-		$this->assertCount( 3, $routes['/web-stories/v1/web-story/(?P<id>[\d]+)/lock'] );
+		$this->assertCount( 6, $routes['/web-stories/v1/web-story/(?P<id>[\d]+)/lock'] );
```
swissspidy: The number of routes hasn't changed

spacedmonkey:
```suggestion
		$this->assertCount( 3, $routes['/web-stories/v1/web-story/(?P<id>[\d]+)/lock'] );
```

swissspidy (approved):

swissspidy:
Just ocurred to me that the dashboard needs changing here as well. Can you check that?

see #12604

kkalarickal:
@spacedmonkey @swissspidy -- I tested the `12559 branch` to confirm that author-users were seeing the take over dialog.


This take over dialog is different from the one that the Admin-user sees, when Author-user takes over a story that Admin-user was editing:


I am not clear which of these two dialogs should appear in Author-users editor when Admin-user takes over the story that Author-user was editing.

swissspidy:
Both users can see both dialogs.

You see one when trying to open a locked story, the other when someone takes over editing from you.

In both cases, users should at all time see the name & avatar of the other user.

spacedmonkey:
Added a fix for https://github.com/GoogleForCreators/web-stories-wp/issues/12604

swissspidy (approved):

packages/wp-dashboard/src/api/utils/reshapeStoryObject.js
```diff
@@ -32,16 +32,16 @@ export default function reshapeStoryObject(originalStoryData) {
...
+    _embedded: { author, 'wp:lock': lock } = {},
     _links: links = {},
   } = originalStoryData;
   if (!id) {
     return null;
   }
+
+  const { locked = false, user: lockUser = { id: 0, name: '' } } =
+    lock?.[0] || {};
+  const storyAuthor = author?.[0] || { id: 0, name: '' };
```
swissspidy, line 44: Note that js unit tests now need updating

spacedmonkey: Fixed.

packages/wp-dashboard/src/api/constants/index.js
```diff
@@ -35,7 +35,7 @@ export const STORY_FIELDS = [
 export const SEARCH_PAGES_FIELDS = ['id', 'title'];
 export const GET_PAGE_FIELDS = ['title', 'link'];
 
-export const STORY_EMBED = 'wp:lock,wp:lockuser,author';
```
swissspidy, line 38: Need to update the `Dashboard` class too I think

swissspidy:
@kkalarickal Can you re-test #12604 here?

kkalarickal:
Confirmed that non-admin users can:
- see lock icon and thumbnail of user who is currently editing a story within the dashboard
- are given appropriate takeover dialog when then edit a locked story
- if the take over occurs, then the original user is given the other type of dialog to indicate that non-admin user has taken over editing
- thumbnail is now changed on the story being edited, when viewed by admin & non-admin users in their respective dashboards
BEFORE:
AFTER:

kkalarickal:
@swissspidy I noticed that the following and wanted to get your take on if this is ok:

- If a user profile has lower privileges than the owner of a story, then she is only able to view the published story. That user cannot edit the story. In such a case, even if that story is opened by someone else, it will NOT show the thumbnail if it is viewed by a user with lower privileges.

Example: This is the view for Admin user

The same dashboard, when viewed by Author user

- if a user opens a story, then logs into another browser instance with the same credentials and views the story dashboard, the original opened story does NOT show the lock icon or thumbnail. It looks like for a user, the info about lock/thumbnail is not shared if viewed from another login of the same user.

timarney (approved):

kkalarickal:
@swissspidy - thank you for confirming. I agree that if user-profile cannot edit a story authored by someone with higher level of access, then there is no need to show the locked/thumbnail. 

For the scenario where the same user is accessing the Story dashboard from different browser instances, I was thinking we should indicate that another instance is open. 

However, when I looked at the WP Post workflow  - it also does not show the lock-icon or thumbnail for same user login from another browser. So I guess we are at par with WP Post workflow.

PR #12577, opened 2022-10-31
Global Page Advancement: use global values when creating templates

miina:
## Context
Follow-up for #12460 where it was decided that templates should also use global value.

## Testing Instructions

1. Go to Settings and set Global Page Advancement values.
2. Now go to Explore templates and choose any template, click on Use template
3. When a new story is created with the template, go to the Document tab and verify the globally set page advancement values are used there.

## Checklist

---

Follow-up for #12460


includes/Story_Post_Type.php
```diff
@@ -113,6 +113,15 @@ public static function get_requirements(): array {
 		return [ 'settings' ];
 	}
 
+	/**
+	 * Get settings.
+	 *
+	 * @return Settings Settings instance.
+	 */
+	public function get_settings(): Settings {
+		return $this->settings;
```
miina: @spacedmonkey @swissspidy I've added a getter for `Settings` for being able to use it for the dashboard. 
LMK if you prefer some other way, or feel free to amend to the PR directly.

spacedmonkey: We haven't done this in other places. Getters like this can get messy. Let's just pass settings into constructor like other classes.

miina: Good point, I can see now that you're doing that everywhere else. Will change, thanks!

spacedmonkey (approved):
SGTM

barklund (approved):
Looks good to me, but can't speak for the PHP stuff.

swissspidy (approved):

merapi (approved):
Tested, LG i will merge it right now so we can bug bash it today.

Issue #12547, opened 2022-10-25
Restore from autosave does not revert changes in pages (but story data is restored)


kkalarickal:
## Bug Description

Restoring an autosaved webstory that is different from the local version does not work in reverting changes to pages. However, a change in the story data (like title) is reverted !

## Steps to Reproduce
The following steps consistently reproduce the problem:

1. Create a new story with 1 page
2. Publish it with poster and description etc
3. Add another page and move it to first position
4. Update this story
5. Now add a third page, change the title & wait for a min to create the autosave session key
6. exit to dashboard without saving to DB
7. reopen the now published & updated story
8. you will get two overlays - one for restoring from local backup (dismiss this) and another for restoring from saved version ...
9. Here, view the different versions and restore the latest (3 slides)

ISSUE: the version with 2 slides is restored instead of the last saved version

## Screenshots

![image.png](https://images.zenhubusercontent.com/5fc17739d77478a91f8d75e3/1fde2cc5-41a0-4452-9ee8-bcf1fd6871ce)

![image.png](https://images.zenhubusercontent.com/5fc17739d77478a91f8d75e3/ab54c441-4499-427b-8bc8-843c38517a41)
## Additional Context

This issue was observed in #12521 as well.

- Plugin Version: `main branch` as of 10/25/2022
- WordPress Version: 6.0.3
- Operating System: Win10
- Browser: Chrome

kkalarickal:
Confirmed that this issue exists in Google Web Stories Version 1.25.1 as well.

swissspidy:
From Slack:

> Not sure if it's restore function, `loadStory` has wrong data, is it possible that `$preload_data` doesn't take revision into account?

kkalarickal:
@swissspidy - regarding the two dialogs .. Comparing the way autosave/local-backup behaves for WP Post with how its implemented in web stories, it appears that after local changes are made to a draft, if the user exits the editor:
- in the WP post case, the changes are saved to disk and the local-session backup is deleted
- in the WebStories case, the changes are saved to disk but the local-session backup is retained

This is ok so long as the story is not published. Because after published status, the changes made after autosave are not written to db when exiting the editor. So now there is a legitimate difference between the local-session backup and the story in db which is why we have two dialogs (correctly).

![image](https://user-images.githubusercontent.com/3719412/198062179-69816c5a-6d54-40af-9a07-237a5bf8c85b.png)

swissspidy:
[...]

PR #12578, opened 2022-10-31
Autosaves: Delete local backup after auto-saving to DB

miina:
## Summary
This addresses an issue with local auto-saving where local backup is not removed after auto-saving to database happens.

## To-do

- [x] Remove local backup when a story is auto-saved

## Testing Instructions

1. Add a new Story and publish.
2. Add some element
3. Wait for the local backup to be stored.
4. Now wait until the DB auto-saving happens
5. Verify that the local backup is removed once it happens.

## Checklist

---

Partially addresses #12547


spacedmonkey (approved):
Seems to work in my testing. I did not see the two messages anymore.

kkalarickal:
@miina @swissspidy -- i tried to test this via `12578 branch` on storie-qa server. But somehow there was only one autosave call made. So I could not see the session stored backup being deleted.

After making some changes to a published story, I could see the `wp_stories_autosave_story_20665` backup created in session storage.

However, there were no autosave calls made and so I could not check if the local backup was deleted.

kkalarickal:
Pls advice if I should check this fix  in another branch.

miina:
@kkalarickal I've updated the PR to `main` and tested in the QA environment -- more than one autosave happens now, perhaps the PR was out of date and did not include the fix for the previous bug where only one autosave happened. Please test again on the same branch.

kkalarickal:
Thanks @miina .

I gave it another go after loading `12578 branch` on stories-qa. Now I can see the multiple autosaves happening, and if I open the same story in another browser instance, I can see that the new version is saved on disk.

However,  I am NOT seeing the `wp_stories_autosave_story_20665` key from session storage being deleted.

So the following test case is failing:
1. Open a previously published story
2. Make some changes to the story (add some pages, or update the title)
3. Wait for `wp_stories_autosave_story_NNNN` key to be created in session storage - to indicate a backup has been taken
4. Wait some more time for the `https://stories-qa-wordpress-amp.pantheonsite.io/wp-json/web-stories/v1/web-story/20665/autosaves/?_locale=user` call to be made - to indicate that autosave was done.

EXPECTED the local session storage key is deleted when autosave runs
OBSERVED the local session storage key remains even if use goes to dashboard

miina:
@kkalarickal I was just testing both locally and in QA and it works when following exactly the same steps 😕

1. Open a previously published story
2. Make some changes to the story (I added a text element)
3. Wait for wp_stories_autosave_story_NNNN key to be created in session storage - to indicate a backup has been taken
4. Wait some more time for the call to be made - to indicate that autosave was done. 
5. See the `wp_stories_autosave_story_NNNN` is deleted

I tested with [this story in the QA env](https://stories-qa-wordpress-amp.pantheonsite.io/wp-admin/post.php?post=20320&action=edit#page=%25227bead3ce-9b80-46fa-84bc-3908e7f374dc%2522) although it shouldn't make a difference 

Not sure what we're doing differently 🤔

kkalarickal:
Thanks @miina ... i will double check on another instance and confirm the fix.

swissspidy:
I can confirm that the change works for me.

@kkalarickal Any updates, does it work for you on your end?

Try use the ZIP file from https://github.com/GoogleForCreators/web-stories-wp/pull/12578#issuecomment-1297293959 on a site just in case, just to ensure you're using the latest version.

swissspidy (approved):

kkalarickal:
Yesterday I tried with `12578 branch` on another test server instance by changing version via the web-stories tester - but I could NOT see the session key `wp_stories_autosave_story_NNNN` being deleted.

Thanks for the link to the ZIP file @swissspidy - when I load it on my test server and try the steps outlined by @miina I am not seeing any autosave calls after the first one.

Sorry, but I am unable to confirm this fix. Its probably something to do with my env/test steps. Since you are seeing the expected outcome, I recommend we merge this and I can test it out in the release candidate.

Cc: @bmattb

swissspidy:
> but I could NOT see the session key wp_stories_autosave_story_NNNN being deleted.

Can you confirm you **do** see it being deleted now? Or still no?

> I am not seeing any autosave calls after the first one.

Well, did you make any new changes after the first autosave? We don't do any autosaves if there were no new changes, so it's expected to not see any if you don't make any changes.

miina:
> Well, did you make any new changes after the first autosave? We don't do any autosaves if there were no new changes, so it's expected to not see any if you don't make any changes.

It should be possible to test this specific fix with just one autosave as well since the local backup is happening earlier in the QA environment, AFAIK. So the first autosave should already delete the local backup.

kkalarickal:
Thanks for the suggestions @miina @swissspidy - I restested this from scratch and was able to verify the fix for 12578: i.e. local session backup is deleted when there is an autosave event

To do this, I had to first trigger a take over of the story while it was being edited via another browser session/login.

So the first user had the editor open to the story and had made changes which resulted in a local session key backup.

When the second user took over the story, there was an autosave event seen in the first user's session and that resulted in the local session backup key being deleted.

(in turn, the second user got a "restore from autosave version" message etc.).

I feel the 12578 branch that I tested with has some issue with the autosave not triggering .. i did make changes in the story off and on, with the hope that in 30 sec I would get the locla backup updated, and within 1 min or so, an autosave event would trigger.
But when I used `12578 branch` I was only able to see the local back up being updated, not the autosave event.. When I switched to `main branch` I was able to see the subsequent autosave events (but ofcourse, the deletion of the local session backup was not done).

Thank you for helping me go through this testing. I feel the 12578 change is working. I am not sure why the regular autosave is not triggering though.

merapi:
For me, it removes the local backup after an autosave, but then creates it again - it shouldn't, right? there are no new changes.
Check from 1:15.

https://user-images.githubusercontent.com/21036224/200660306-1af0b08e-62bb-4699-8bcf-99d383ce202f.mov

miina:
> For me, it removes the local backup after an autosave, but then creates it again - it shouldn't, right? there are no new changes.
> Check from 1:15.

And then it probably removes it again after auto-save happens.

Yep, there are no new changes since the autosave, however, there are new changes compared to the last saved story, so it would create a backup again, I guess.

Might be good to create a separate issue for that.

swissspidy:
Yes we're seeing the same thing. That indeed shouldn't happen, it's strange. Especially since there's a `hasNewChanges` check in there.

swissspidy:
> Might be good to create a separate issue for that.

💯

miina:
> Yes we're seeing the same thing. That indeed shouldn't happen, it's strange. Especially since there's a `hasNewChanges` check in there.

`hasNewChanges` means having new changes compared to the last Save only. We're not tracking new changes based on the last autosave anywhere currently.

kkalarickal:
Esactly @merapi . Thanks to some debugging commands that @swissspidy added, I am able to confirm that the local session key is in getting deleted, but there is another save local key event right after that which sort of restores it back into the browser.

- make changes
- store local autosave 206
- perform autosave
- delete local autosave 206
- store local autosave 206
- make changes
- store local autosave 206
- make changes
- store local autosave 206
- perform autosave
- delete local autosave 206
- store local autosave 206

so that last store local autosave happens after the delete and it sort of restores the session key back into the browser

merapi:
> And then it probably removes it again after auto-save happens.

If I don't change anything there is no more autosave and it stays like that.

swissspidy:
> > Yes we're seeing the same thing. That indeed shouldn't happen, it's strange. Especially since there's a `hasNewChanges` check in there.
> 
> `hasNewChanges` means having new changes compared to the last Save only. We're not tracking new changes based on the last autosave anywhere currently.

Ah, right, makes sense. Thanks for clarifying/reminding!

swissspidy:
Merging this one for now :)

PR #12606, opened 2022-11-01
E2E Tests: Run against WordPress 5.9+

### LINKED ISSUE #12572: E2E Tests: Run against WordPress 5.9+
## Task Description

Might require some test changes, otherwise this should be like a one line fix.

timarney:
## Context

Update E2E test to run 5.9 as min version

## Summary

Updates workflow file

## Testing Instructions

1. verify e2e test pass as per current main branch 

No

No

No

## Checklist

---

Fixes #12572

### DIFF
--- a/bin/local-env/install-wordpress.sh
+++ b/bin/local-env/install-wordpress.sh
@@ -145,9 +145,6 @@ wp plugin install rtl-tester --activate --force --quiet
 echo -e $(status_message "Installing WordPress importer...")
 wp plugin install wordpress-importer --activate --force --quiet
 
-echo -e $(status_message "Installing WordPress importer...")
-wp plugin install wordpress-importer --activate --force --quiet
-
 echo -e $(status_message "Installing WooCommerce plugin...")
 wp plugin install woocommerce --activate --force --quiet
 
--- a/bin/local-env/install-wordpress.sh
+++ b/bin/local-env/install-wordpress.sh
@@ -158,7 +158,7 @@ echo -e $(status_message "Installing Classic Widgets plugin...")
 wp plugin install classic-widgets --force --quiet
 
 echo -e $(status_message "Activating Twenty Twenty theme...")
-wp theme activate twentytwenty --quiet
+wp theme install twentytwenty --activate --force --quiet
 
 # Set pretty permalinks.
 echo -e $(status_message "Setting permalink structure...")
--- a/bin/local-env/install-wordpress.sh
+++ b/bin/local-env/install-wordpress.sh
@@ -237,3 +237,4 @@ wp plugin deactivate woocommerce
 
 echo -e $(status_message "Deactivating WordPress importer again...")
 wp plugin deactivate wordpress-importer --quiet
+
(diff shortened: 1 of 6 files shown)


timarney:
@swissspidy @spacedmonkey  --- seems like most of the remaining failures for this PR are for the WooCommerce tests any thoughts as to why? 

Bumped `minWPVersionRequired('5.9'); // WooCommerce requires WP 5.7+` but that didn't fix in this case

packages/e2e-tests/src/specs/editor/shopping/productMenu.js
```diff
@@ -27,7 +27,7 @@ import {
 jest.retryTimes(3, { logErrorsBeforeRetry: true });
 
 describe('Shopping product', () => {
-  minWPVersionRequired('5.8');
+  minWPVersionRequired('5.9');
```
swissspidy: The idea on the ticket was to actually completely **remove** all these checks.

We added them because WooCommerce requires 5.8+ but we still tested on 5.7

Now that we test on 5.9+, these checks are superfluous.

spacedmonkey:
What about - https://github.com/GoogleForCreators/web-stories-wp/blob/66f8d1dd74e223a403067cd71365d4a6c8195b62/.github/workflows/tests-unit-php.yml#L69

swissspidy:
@spacedmonkey There's a separate ticket for changing the _unit_ tests.

swissspidy:
> @swissspidy @spacedmonkey --- seems like most of the remaining failures for this PR are for the WooCommerce tests any thoughts as to why?
> 
> Bumped `minWPVersionRequired('5.9'); // WooCommerce requires WP 5.7+` but that didn't fix in this case

@timarney Because WooCommerce is not being installed on the WP 5.9 tests.

In `install-wordpress.sh` you can see it's only being installed when using `latest`. But now we need to change it to always install WooCommerce.

bin/local-env/install-wordpress.sh
```diff
@@ -232,14 +229,11 @@ wp user list --format=yaml
...
-  echo -e $(status_message "Deactivating WooCommerce again...")
-	wp plugin deactivate woocommerce
+echo -e $(status_message "Deactivating WooCommerce again...")
+wp plugin deactivate woocommerce
 
-  echo -e $(status_message "Deactivating WordPress importer again...")
-  wp plugin deactivate wordpress-importer --quiet
-fi
+echo -e $(status_message "Deactivating WordPress importer again...")
+wp plugin deactivate wordpress-importer --quiet
```
swissspidy: Nit: missing newline at end of file

swissspidy:
FYI, looks like the WP latest e2e tests are not passing because the `twentytwenty` theme could not be found, because it's not recent enough.

WP only ships three default themes, so with 6.1 it's `twentytwentyone`, `twentytwentytwo` and `twentytwentythree`

To fix, we need to force-install the theme. So we need to change this line:

https://github.com/GoogleForCreators/web-stories-wp/blob/1512766a370638509d44baeb8b9586639c0e105c/bin/local-env/install-wordpress.sh#L161

Into `wp theme install twentytwenty --activate --force --quiet`

swissspidy (approved):

PR #12611, opened 2022-11-02
Correctly filter revision fields for story revisions

### LINKED ISSUE #12547: Restore from autosave does not revert changes in pages (but story data is restored)

swissspidy:
## Context

When restoring a story from an autosave, the story content was not correctly restored.

This is because when filtering `_wp_post_revision_fields`, the post type can either be `web-story` or `revision`, and we didn't check for the latter.

## Summary

Correctly filter revision fields for story revisions by checking the post parent.

## Testing Instructions

1. Publish a new story.
2. Add some more content to it, then wait for to be autosaved to the database
3. Refresh the page, then click on the (second) dialog CTA to view the autosaves
4. On the revisions comparison screen, restore the latest version
5. Verify that the correct content has been restored

No

No

No

## Checklist

---

Fixes #12547

### DIFF
--- a/includes/Story_Revisions.php
+++ b/includes/Story_Revisions.php
@@ -38,6 +38,12 @@ use WP_Post;
  *   name: string,
  *   diff: string
  * }
+ * @phpstan-type PostData array{
+ *   post_parent: int,
+ *   post_type: string,
+ *   post_content?: string,
+ *   post_content_filtered?: string
+ * }
  */
 class Story_Revisions extends Service_Base {
 
@@ -101,13 +107,22 @@ class Story_Revisions extends Service_Base {
 	 * @param array|mixed         $fields Array of allowed revision fields.
 	 * @param array<string,mixed> $story  Story post array.
 	 * @return array|mixed Array of allowed fields.
+	 *
+	 * @phpstan-param PostData $story
 	 */
 	public function filter_revision_fields( $fields, array $story ) {
 		if ( ! \is_array( $fields ) ) {
 			return $fields;
 		}
 
-		if ( $this->story_post_type->get_slug() === $story['post_type'] ) {
+		if (
+			$this->story_post_type->get_slug() === $story['post_type'] ||
+			(
+				'revision' === $story['post_type'] &&
+				! empty( $story['post_parent'] ) &&
+				get_post_type( $story['post_parent'] ) === $this->story_post_type->get_slug()
+			)
+		) {
 			$fields['post_content_filtered'] = __( 'Story data', 'web-stories' );
 		}
 
--- a/tests/phpunit/integration/tests/Story_Revisions.php
+++ b/tests/phpunit/integration/tests/Story_Revisions.php
@@ -20,7 +20,7 @@ namespace Google\Web_Stories\Tests\Integration;
 use Google\Web_Stories\Story_Post_Type;
 
 /**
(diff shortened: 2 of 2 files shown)


tests/phpunit/integration/tests/Story_Revisions.php
```diff
@@ -102,4 +104,63 @@ public function data_test_revisions_to_keep(): array {
...
+			[
+				'post_type'   => Story_Post_Type::POST_TYPE_SLUG,
+				'post_parent' => 0,
+			]
+		);
+
+		$this->assertArrayHasKey( 'post_content_filtered', $actual );
+	}
+
+	public function test_filter_revision_fields_story_post_type_revision(): void {
```
spacedmonkey, line 157: Add cover notiations.

includes/Story_Revisions.php
```diff
@@ -101,13 +107,22 @@ public function revisions_to_keep( $num ): int {
...
 	public function filter_revision_fields( $fields, array $story ) {
 		if ( ! \is_array( $fields ) ) {
 			return $fields;
 		}
 
-		if ( $this->story_post_type->get_slug() === $story['post_type'] ) {
+		if (
+			$this->story_post_type->get_slug() === $story['post_type'] ||
+			(
+				'revision' === $story['post_type'] &&
```
spacedmonkey, line 121: Should this also include autosaves?

spacedmonkey (approved):
Some nitpic, like phpdocs on tests but otherwise, looking good.

includes/Story_Revisions.php
swissspidy: I don‘t think such a case exists

timarney (approved):

tests/phpunit/integration/tests/Story_Revisions.php
swissspidy: Good call! Just noticed that the `coversDefaultClass` annotation was wrong here.

spacedmonkey (approved):

kkalarickal:
Confirmed the fix by restoring from a saved version and verifying that the proper version was restored.

Comparing revisions

Restored web story with 3 pages

kkalarickal:
Further tests:

Confirmed that going back in the revision history and restoring a previous version was successful.

Also confirmed that comparing two arbitrary versions and restoring between them works as expected.

Verified that restoration works across user-profiles: meaning when one user creates the original story, and another takes over editing, the revision mismatch is caught and restoring someone else's version is possible.

Verified that with each restoration, the revision count within Document panel keeps incrementing as expected !

PR #12612, opened 2022-11-02
Bump `Requires PHP` value

### LINKED ISSUE #12583: Bump `Requires PHP` value
## Task Description

Bump to `Requires PHP: 7.4` in `web-stories.php`

Also bump `WEBSTORIES_MINIMUM_PHP_VERSION` in `web-stories.php`

Also bump `platform.php` in `composer.json` to 7.4

Also adjust PHPUnit compat table in GitHub Actions workflow

spacedmonkey:
## Summary

Update the min requirements to PHP 7.4. Tweak composer. 

## Testing Instructions

1.

## Checklist

---

Fixes #12583

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -107,7 +107,6 @@
     "post-update-cmd": [
       "@prefix-dependencies"
     ],
-    "composer-normalize": "composer-normalize",
     "phpcbf": "phpcbf --severity=1",
     "phpcs": "phpcs --severity=1",
     "phpmd": "phpmd . text phpmd.xml",
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
   "type": "wordpress-plugin",
   "homepage": "https://github.com/googleforcreators/web-stories-wp",
   "require": {
-    "php": "^7.2 || ^8.0",
+    "php": "^7.4 || ^8.0",
     "ext-dom": "*",
     "ext-iconv": "*",
     "ext-json": "*",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -67,39 +69,20 @@
...
     "sort-packages": true
   },
   "extra": {
-    "downloads": {
-      "composer-normalize": {
-        "path": "vendor/bin/composer-normalize",
-        "type": "phar",
-        "url": "https://github.com/ergebnis/composer-normalize/releases/download/2.24.0/composer-normalize.phar"
-      },
-      "php-scoper": {
```
swissspidy, line 89: PHP-Scoper requires PHP 8, so we can't remove this.

swissspidy:
Marking as **blocked** just so we don't merge this before the next release, which is slated for next week, since this change is intended to be in the release after it.

composer.json
spacedmonkey: It required [8.1.](https://github.com/humbug/php-scoper/blob/ee08e096d379100af32275f234951e90a10874a5/composer.json#L20).Using composer to install phpscoper here means we get an older version that is supported. One from September. See https://github.com/humbug/php-scoper/pull/679

swissspidy: An _older_ version, I see. The downside is we don't use the latest version and can't upgrade it anymore.

spacedmonkey: But it requires PHP 8.1, something I am not running locally and we only just support. I think running the older version is a benefit here.

swissspidy (approved):

PR #12615, opened 2022-11-03
Block: Use `getUsers` with capability query

### LINKED ISSUE #9240: Block: `getAuthors()` deprecation notice
## Feature Description

When using the story picker in the web stories gutenberg block, the following noticed is displayed. 

```
react_devtools_backend.js:4049 select( 'core' ).getAuthors() is deprecated since version 5.9. Please use select( 'core' ).getUsers({ who: 'authors' }) instead. 
    at SelectStories 
```

See screenshot for full details. 
Screenshot 2021-10-04 at 16 45 49

This change happened upstream here https://github.com/WordPress/gutenberg/pull/33725

spacedmonkey:
## Summary

Replace `getAuthors` with `getUsers`. 

Followed this example - https://github.com/WordPress/gutenberg/pull/33695

## Testing Instructions

1. Create a post. 
2. Insert a web story block. 
3. Select "selected stories". 
4. Select a template. 
5. Click select stories. 
6. Click author dropdown. See that dropdown still works.

## Checklist

---

Fixes #9240


swissspidy:
So IIUC this replaces the deprecated `getAuthors` function with `getUsers()`, which in itself makes sense.

But in WP 5.9 `who: 'authors'` was also replaced by using `capabilities: 'edit_web-stories'`, so ideally we use that as well.

Because then we can get rid of the workaround in `\Google\Web_Stories\REST_API\Stories_Users_Controller::filter_user_query()`


So something like

```js
  const { authors } = useSelect(
    (select) => {
      const { getUsers, getPostType } = select(coreStore);

      const capabilities = getPostType('web-story')?.capabilities['edit_posts'];

      const query = {
        search: authorKeyword,
        capabilities,
      };
      return {
        authors: getUsers(query),
      };
    },
    [authorKeyword]
  );
```

Uses optional chaining for the `getPostType()` call because it first returns `undefined` when the data is not available yet.

Because of that, probably actually better to do this:

```js
  const { authors } = useSelect(
    (select) => {
      const { getUsers, getPostType } = select(coreStore);

      const capabilities = getPostType('web-story')?.capabilities['edit_posts'];

      const query = {
        search: authorKeyword,
        capabilities,
      };
      return {
        // Only load users when capability has been fetched already.
        authors: capabilities ? getUsers(query) : [],
      };
    },
    [authorKeyword]
  );
```

spacedmonkey:
@swissspidy Your fix seems to work. 

Should I remove this `\Google\Web_Stories\REST_API\Stories_Users_Controller::filter_user_query()` then?

swissspidy:
Yes

spacedmonkey:
> Yes

Done

swissspidy (approved):
Just needs QA instructions for regression testing

kkalarickal:
ReCreated this issue on stories-QA using `main branch`

Confirmed the fix via `12615 branch` - there is no longer a message in the console related to getAuthors().

QA Note: When the author drop down is pressed, the cover image seems to hide the authors list

@swissspidy @spacedmonkey - is it within scope of this ticket to ensure that the select-story :: drop down of the authors list should be visible ?

swissspidy:
No, not at all. Styling fixes are in progress at #12649

